PR #932, opened 2018-06-10
Fix non exhaustive match warnings and other casper warnings

KentShikama:
## Overview

The title says it all.


rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -20,7 +20,10 @@ object Dispatch {
 
   // TODO: Make this function total
   def buildEnv(dataList: Seq[Seq[Channel]]): Env[Par] =
-    Env.makeEnv(dataList.flatten.map({ case Channel(Quote(p)) => p }): _*)
+    Env.makeEnv(dataList.flatten.map({
+      case Channel(Quote(p)) => p
+      case Channel(_)        => Par() // Should never happen
```
rabbitonweb: if should never happen then maybe leave the partial function as it was and add suppres warning to it. if it ever fails we want it to fail here (with a huge exception), rather then swallowing the error

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
```diff
@@ -17,7 +17,10 @@ object StoragePrinter {
           val sends: Seq[Send] = data.flatMap {
             case Datum(as: Seq[Channel], persist: Boolean) =>
               channels.map { channel =>
-                Send(Some(channel), as.map { case Channel(Quote(p)) => p }, persist)
+                Send(Some(channel), as.map {
+                  case Channel(Quote(p)) => p
+                  case Channel(_)        => Par() // Should never happen
```
rabbitonweb: same as above

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
KentShikama: It seems like suppress warnings is not available in Scala https://stackoverflow.com/a/3506583/2750819 ?

PR #935, opened 2018-06-11
CORE-609 restructure comm

rabbitonweb:
## Overview
Reorganize `comm` structure. Rename / move / replace packages, classes, methods and objects. Making `comm` a bit more readable and maintainable.

The idea is simple:
1. comm holds only `TransportLayer` and `NodeDiscovery` - both API and logic (and tests in future)
2. everything else that is currently in `comm` will be moved to either `node` or `shared`  (in CORE-609 - next PR).
3. This unlocks CORE-607 for Sebastian


comm/src/main/scala/coop/rchain/comm/PeerNode.scala
```diff
@@ -0,0 +1,71 @@
...
+
+/*
+ * Inspiration from ethereum:
+ *
+ *   enode://92c9722c6cd441b9d64cc3f365cdd6cda822c7496b8131078a7362b576526092b58531aa2e5914c4e87e0e574bca4497e68f35a7b66c153ee73b6c7dfc958a34@10.0.0.3:30303
+ *
+ * =>
+ *
+ *   rnode://<key>@<host>:<udp-port>
+ */
```
sebtomba: Let's update this comment

PR #940, opened 2018-06-11
Add validation test that block is from a bonded validator

birchmd:
## Overview
Save some compute cycles by catching that a block is not from a bonded validator as early as possible.

https://rchain.atlassian.net/browse/RHOL-397

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -78,6 +78,16 @@ object ProtoUtil {
       mainParent <- blocks.get(parentHash)
     } yield mainParent
 
+  def weightFromValidator(b: BlockMessage,
+                          validator: ByteString,
+                          blocks: collection.Map[ByteString, BlockMessage]): Int =
+    mainParent(blocks, b)
+      .map(weightMap(_).getOrElse(validator, 0))
+      .getOrElse(weightMap(b).getOrElse(validator, 0)) //no parents means genesis -- use itself
+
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -78,6 +78,16 @@ object ProtoUtil {
       mainParent <- blocks.get(parentHash)
     } yield mainParent
 
+  def weightFromValidator(b: BlockMessage,
+                          validator: ByteString,
+                          blocks: collection.Map[ByteString, BlockMessage]): Int =
+    mainParent(blocks, b)
+      .map(weightMap(_).getOrElse(validator, 0))
+      .getOrElse(weightMap(b).getOrElse(validator, 0)) //no parents means genesis -- use itself
```
KentShikama: We'll have to double check our assumptions at the end (especially ordering). Like are we guaranteed at this point to have an otherwise valid block - because it could be an invalid block with no parents.

birchmd: That's why I check separately if the block is genesis and anything else that returns 0 is invalid. https://github.com/rchain/rchain/pull/940/files#diff-eddb46712698d75072b3f453346dd0afR70

KentShikama: Ok then being pedantic this comment should be `// no parents means genesis or invalid` as when this function is called from blockSender it is never the genesis

KentShikama: Eh, seems like you merged. Don't bother if it is a pain to push.

PR #941, opened 2018-06-11
CORE-592: rspace: create trace logs on ISpace#getCheckpoint

henrytill:
## Overview
This PR adds trace logging to RSpace.  

**Basics**:
- A call to `produce` is logged as a `Produce`
- A call to `consume` is logged as a `Consume`
- A COMM Event is a logged as a `COMM`
- Included in each persisted `Datum` is the corresponding `Produce`.
- Included in each persisted `WaitingContinuation` is the corresponding `Consume`.
- A call to `getCheckpoint` now returns an instance of:
  ```scala
  case class Checkpoint(root: Blake2b256Hash, log: trace.Log)
  ```

https://rchain.atlassian.net/browse/CORE-592

### Notes

- `Consume` and `Produce` were missing their `persist` parameter.  This has been fixed in this PR.
- The constructors to `Consume` and `Produce` have been renamed from `apply` to `create`.
- `create` constructors have been added to `Datum` and `WaitingContinuation` to simplify their creation in test suites and elsewhere.


rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -85,6 +100,11 @@ class RSpace[C, P, A, K](val store: IStore[C, P, A, K]) extends ISpace[C, P, A,
       store.withTxn(store.createTxnWrite()) { txn =>
         logger.debug(s"""|consume: searching for data matching <patterns: $patterns>
               |at <channels: $channels>""".stripMargin.replace('\n', ' '))
+
+        val consumeRef    = Consume.create(channels, patterns, continuation, persist)
+        val currentEvents = eventLog.take()
+        eventLog.put(currentEvents :+ consumeRef)
```
dzajkowski: seems like a pattern
maybe:
```
implicit class RichSyncVar[A](sv: SyncVar[Seq[A]]) {
  def update(f: Seq[A] => Seq[A]) {
    val curr = sv.take()
    sv.put(f(curr))
  }
}

eventLog.update(_ :+ Consume.create(channels, patterns, continuation, persist))
eventLog.update(_ => Seq.empty)
```
?

henrytill: It does, doesn't it.

PR #944, opened 2018-06-12
CORE-608 restructure node

rabbitonweb:
## Overview
Restructure Node code.

CORE-608


node/src/main/scala/coop/rchain/node/api/ReplGrpcService.scala
```diff
@@ -0,0 +1,64 @@
+package coop.rchain.node.api
+
+import coop.rchain.node.diagnostics
+import coop.rchain.p2p.effects._
+import io.grpc.{Server, ServerBuilder}
+
+import scala.concurrent.Future
+import cats._, cats.data._, cats.implicits._
+import com.google.protobuf.empty.Empty
+import coop.rchain.casper.{MultiParentCasper, PrettyPrinter}
```
birchmd: Seems like there are a lot of unused imports here, should probably be cleaned up.

rabbitonweb: I'm planning to provide a automated scala-fix for that next sprint. Don't want to do it manually :)

birchmd: Cool :)

PR #938, opened 2018-06-11
Core 596/rspace/refactor/extension nodes prerequisite

lukasz-golebiewski:
## Overview
Prerequisite for extension(skip) nodes support

### Does this PR relate to an RChain JIRA issue? 
https://rchain.atlassian.net/browse/CORE-596

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] You have someone in mind to assign for review. Make this assignment after submitting the PR.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #949, opened 2018-06-12
CORE-640: rspace.history: add Branch

henrytill:
## Overview
To support both the creation of new deployments and the deterministic replay of deployments in blocks, the `Runtime` will contain two RSpaces (and companion `Reducer`s).   

**Breakdown:**
1. For the creation of code, one would use a normal instance of `RSpace` in its own `Reducer`.  
2. For the deterministic replay of deployments, one would use an instance of the upcoming `ReplayRSpace` in its own `Reducer`.  

After discussing it with @birchmd, it seems to make the most sense to have both of these RSpaces share the same history trie.  In order to support this, we need to have the ability for each RSpace to track its own separate root.  Currently the trie implementation does not support this - it tracks a single root for a single client.  This PR adds the ability to store multiple roots at particular "branches".  Each RSpace gets and operates on its own `Branch` in the history trie.

https://rchain.atlassian.net/browse/CORE-640

### Notes
Curious how this interacts with the work in #938.  Haven't looked closely yet...


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -91,7 +92,7 @@ object Runtime {
       LMDBStore
         .create[Channel, BindPattern, Seq[Channel], TaggedContinuation](dataDir, mapSize)
 
-    val space                                            = new RSpace(store)
+    val space                                            = new RSpace(store, Branch("master"))
```
lukasz-golebiewski: "master" appears multiple times in the code. Maybe it would make sense to put it inside the `Branch` object as constant?

henrytill: Good call.

PR #959, opened 2018-06-14
Add validation for justification regressions

KentShikama:
## Overview

Adds a new field to the BlockDag called latestMessagesOfLatestMessages,
which is required to check for justification regressions.

https://rchain.atlassian.net/browse/RHOL-413

### Notes


casper/src/main/scala/coop/rchain/casper/BlockDag.scala
```diff
@@ -8,17 +9,24 @@ import scala.collection.immutable.{HashMap, HashSet}
...
-                          latestMessages: Map[Validator, BlockHash],
+                          latestMessages: LatestMessages,
+                          latestMessagesOfLatestMessages: Map[Validator, LatestMessages],
                           currentId: Int,
                           currentSeqNum: Map[Validator, Int])
 
 object BlockDag {
+  type LatestMessages = Map[Validator, BlockHash]
+  object LatestMessages {
+    def apply(): LatestMessages = Map.empty[Validator, BlockHash]
```
birchmd: Would prefer if this method was also called `empty` as it is just an alias.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -123,11 +157,48 @@ object Validate {
...
+   */
+  def justificationRegressions[F[_]: Applicative: Log](
+      b: BlockMessage,
+      genesis: BlockMessage,
+      dag: BlockDag): F[Either[BlockStatus, BlockStatus]] = {
+    val latestMessagesOfBlock = ProtoUtil.toLatestMessages(b.justifications)
+    val latestMessagesOfLatestMessagesForSender =
+      dag.latestMessagesOfLatestMessages.getOrElse(b.sender, latestMessagesOfBlock)
+    val containsJustificationRegression =
+      latestMessagesOfBlock.zip(latestMessagesOfLatestMessagesForSender).exists {
```
birchmd: Is `zip` guaranteed to associate pairs with the same key? I don't see any such guarantee in the docs and in fact it says "Note: might return different results for different runs, unless the underlying collection type is ordered." I think it's better to play it safe and write something like
```scala
latestMessagesOfBlock.exists {
  case (validator, currentBlockJustificationHash) =>
    //I'm actually not entirely sure what to do if `validator` is not in the other map, you can think it through though
    val previousBlockJustificationHash = latestMessagesOfLatestMessagesForSender.getOrElse(validator, ???)
    ...
}
```

KentShikama: Great catch. It would mean that the previous latest message should have been already caught as invalid for missing a justification pointer or a new validator got rotated in (in which case just set it to the currentBlockJustificationHash as that is the "first block" from that validator).

KentShikama:
@birchmd PTAL

casper/src/main/scala/coop/rchain/casper/BlockDag.scala
```diff
@@ -8,17 +9,24 @@ import scala.collection.immutable.{HashMap, HashSet}
...
-                          latestMessages: Map[Validator, BlockHash],
+                          latestMessages: LatestMessages,
+                          latestMessagesOfLatestMessages: Map[Validator, LatestMessages],
                           currentId: Int,
                           currentSeqNum: Map[Validator, Int])
 
 object BlockDag {
+  type LatestMessages = Map[Validator, BlockHash]
+  object LatestMessages {
+    def empty(): LatestMessages = HashMap.empty[Validator, BlockHash]
```
birchmd: Parentheses are not needed on this method. Typically, you only include the empty argument list if the function has a side-effect. Not a PR blocker, but should fix in a future commit.

KentShikama: Will do in next PR

PR #961, opened 2018-06-14
Core 658 connection bug

rabbitonweb:
## Overview
---> CORE-658 needs to be merge to master as hot bugfix and tagged&released as `0.4.2` <---

Things that were fixed:

1. UpNp was opening UDP ports, not TCP 🤦‍♂️ 
2. Local vs external network interface (+ UpNp & UDP dirty hacks)
* when `--host` flag was used, UpNp was functionality was being used
* when `--host` flag was not set, then two things would happen: UpNp was trying to open ports AND tried to GUESS a network interface (and IP address) which was being broadcast to other nodes during connection. Problem was that IP that was being broadcast was the INTERNAL one (like 192.168.x.y) not the external one! 🤷‍♂️  Why the hell it worked with `UdpTransportLayer` in the past then? Well UDP had a dirty little hack - whenever answering incoming message it would not used the IP address that it learned during connection (handshake protocol), it would use IP address from the incoming datagram 🤣 
3. roundTrip took default timeout 500 milliseconds, which is in most cased not enough. 
* for this release default timeout value was increased to 1 second and `--default-timeout` flag was added so that the timeout can be changed individually 
* for 0.5 we have do some radical changes: remove timeout from `roundTrip` method, so that `TransportLayer` can implement incremental backoff for every `PeerNode` it has a conversation with. This was we will be able to adjust to different topologies.

### DIFF
--- a/scripts/p2p-test-network.sh
+++ b/scripts/p2p-test-network.sh
@@ -68,9 +68,9 @@ EOF
    fi
 
     if [[ $i == 0 ]]; then
-      rnode_cmd="--port 30304 --standalone"
+      rnode_cmd="--port 30304 --standalone --host bootstrap0.${network_name}"
     else
-      rnode_cmd="--bootstrap rnode://23ea7ec9e3e42054c062c879d8c766a111f3ad37@169.254.1.2:30304"
+      rnode_cmd="--bootstrap rnode://23ea7ec9e3e42054c062c879d8c766a111f3ad37@169.254.1.2:30304 --host ${container_name}"
     fi
     sudo docker run -dit --name ${container_name} \
       -v ${var_lib_rnode_dir}:/var/lib/rnode \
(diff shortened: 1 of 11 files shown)


KellyatPyrofex:
Thanks for the work on this, Pawel.

Re: merge to master and tagging a new release. 
- @jeremybusk can do the tagging after merge to master. 
- @rabbitonweb we will need another PR to update the version number in the build.sbt file. See https://rchain.atlassian.net/browse/CORE-664

scripts/p2p-test-network.sh
```diff
@@ -40,7 +40,7 @@ create_test_network_resources() {
     ${network_name}
 
   echo "Creating docker test containers"
-  for i in {0..2}; do
+  for i in {0..1}; do
```
jeremybusk: change this back to for i in {0..2}; do

rabbitonweb:
Closed in dev

PR #962, opened 2018-06-14
Rev blessed contract

birchmd:
## Overview
Introduces the Rev blessed contract into the casper code base.

This is a pretty big PR because of all the new Rholang code coming in. Maybe best to split up the reviewing a little bit. If @pyrocto and @guardbotmk3  want to divide the rholang contracts/tests between the two of you and @KentShikama can take the Scala parts, I think that should make it manageable.

https://rchain.atlassian.net/browse/RHOL-404

### Notes


casper/src/main/resources/logback.xml
```diff
@@ -0,0 +1,18 @@
...
+  <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
+
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <!-- encoders are  by default assigned the type ch.qos.logback.classic.encoder.PatternLayoutEncoder -->
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <logger name="coop.rchain.rspace" level="warn" />
```
KentShikama: I was about to do this too 👍

casper/src/test/scala/coop/rchain/casper/genesis/contracts/BasicWalletSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+    runtime)
+  val tuplespace = StoragePrinter.prettyPrint(runtime.space.store)
+
+  "BasicWallet rholang contract" should tests.head in {
+    TestSetUtil.testPassed(tests.head, tuplespace) should be(true)
+  }
+
+  tests.tail.map(test => {
+    it should test in {
+      //TODO: remove when withdraw tests are working properly
```
KentShikama: I guess we can wait for https://github.com/rchain/rchain/pull/967 ?

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
...
+
+//assumes standard purse API from MakeMint
+contract @"BasicWallet"(@purse, @algorithm, @pk, return) = {
+  new algorithmKnown, cryptoVerifyCh in {
+    match algorithm {
+      "ed25519" => {
+        algorithmKnown!(true) | cryptoVerifyCh!("ed25519Verify")
+      }
+      _ => {
+        algorithmKnown!(false)
```
guardbotmk3: It would probably be a bit more concise to use a tuple here and pass Nil as the verify string.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
...
+        true => {
+          for (cryptoVerify <- cryptoVerifyCh) {
+            new basicWallet, withdraw, verify, nonceCh in {
+              nonceCh!(-1) |
+              contract verify(@nonce, @sig, return) = {
+                for(@prevNonce <- nonceCh) {
+                  match nonce == (prevNonce + 1) {
+                    true => { 
+                      new result in {
+                        cryptoVerify!(nonce.toByteArray(), sig.hexToBytes(), pk.hexToBytes(), *result) |
```
guardbotmk3: I think if we're going to use integers as a nonce, we should have a method that gets a little endian byte array for just the int rather than the process.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
...
+                        for(@r <- result) {
+                          match r {
```
guardbotmk3: In these cases, I think it would be clearer to use if else.

casper/src/main/rholang/BasicWallet.rho
guardbotmk3: Also, are we just signing the nonce? That doesn't seem like enough.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
...
+                        }
+                      }
+                    }
+                    _ => { nonceCh!(prevNonce) | return!(false) }
+                  }
+                }
+              } |
+              contract withdraw(@amount, return) = {
+                new destPurseCh, success in {
+                  @[purse, "sprout"]!(*destPurseCh) | for(@destPurse <- destPurseCh) {
```
guardbotmk3: It seems this would be simpler if there were a version of sprout that took an amount and a success channel, (or returned a tuple)

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
...
+              } |
+              contract @[*basicWallet, "getBalance"](return) = {
+                @[purse, "getBalance"]!(*return)
+              } |
+              contract @[*basicWallet, "deposit"](@amount, @src, success) = {
+                @[purse, "deposit"]!(amount, src, *success)
+              } |
+              contract @[*basicWallet, "withdraw"](@amount, @nonce, @sig, return) = {
+                new verifyResult in {
+                  verify!(nonce, sig, *verifyResult) | for(@r <- verifyResult) {
```
guardbotmk3: It seems to me that we should be signing and verifying the amount, and the return channel to avoid pre-emption.

casper/src/main/rholang/MakeBrandPair.rho
```diff
@@ -0,0 +1,29 @@
...
+//on a blockchain). However, restricting access is enough to ensure security in
+//some situations. For example one could seal the method on a "purse" object which
+//extracts the tokens inside, thus allowing only those with the unsealer to 
+//take coins out of the purse.
+contract @"MakeBrandPair"(return) = {
+  new seal, unseal, key in {
+    contract seal(@data, return) = {
+      new sealedData in {
+        contract sealedData(@input, return) = {
+          bundle+{
```
guardbotmk3: I don't know what you're trying to achieve with this bundle. It doesn't seem to accomplish anything.

guardbotmk3: It seems that what you might want to do instead is to send a list/tuple where the first element is the name, and the second is a bundle0 that uses that name in a linear for.

casper/src/main/rholang/MakeMint.rho
```diff
@@ -0,0 +1,60 @@
...
+          contract internalMakePurse(@balance, return) = {
+            //balance must be a name that NonNegativeNumber contract is listening on
+            new thisPurse in {
+              contract @[*thisPurse, "decr"](@amount, success) = { @[balance, "sub"]!(amount, *success) } |
+              contract @[*thisPurse, "getBalance"](return) = { @[balance, "value"]!(*return) } |
+              contract @[*thisPurse, "sprout"](return) = { @[*thisMint, "makePurse"]!(0, *return) } |
+              contract @[*thisPurse, "getDecr"](return) = { @seal!([*thisPurse, "decr"], *return) } |
+              contract @[*thisPurse, "deposit"](@amount, @src, success) = {
+                new result in {
+                  @[src, "getDecr"]!(*result) | for(@sealedSrcDecr <- result) { //call getDecr function in src
```
guardbotmk3: There's a simpler way to do this without the sealer unsealer. Because we have bundles, we can use unforgeable names in a challenge response protocol.

Mint has a new name as a tag.

getDecr instead of giving a sealed decr would give a new name as a challenge, and an answer channel. If bundle0{*tag | *challenge} is received on the answer channel, decr can be passed.

casper/src/main/rholang/MakeMint.rho
```diff
@@ -0,0 +1,60 @@
...
+            new balanceCh in {
+              @["MakeMint", "int2NN"]!(init, *balanceCh) | for(@balance <- balanceCh) {
+                internalMakePurse!(balance, *return)
+              }
+            }
+          } |
+          contract internalMakePurse(@balance, return) = {
+            //balance must be a name that NonNegativeNumber contract is listening on
+            new thisPurse in {
+              contract @[*thisPurse, "decr"](@amount, success) = { @[balance, "sub"]!(amount, *success) } |
```
guardbotmk3: I think you need `new thisPurse, decr` and then to use `decr` instead of `"decr"` because as it stands, you can just empty a purse on accident, and I don't think that's what you want.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,67 @@
...
+    for(@ak <- algorithmKnown) {
+      match ak {
+        false => { return!([]) }
+        true => {
+          for (cryptoVerify <- cryptoVerifyCh) {
+            new basicWallet, verify, nonceCh in {
+              nonceCh!(-1) |
+              contract verify(@nonce, @sig, return) = {
+                for(@prevNonce <- nonceCh) {
+                  match nonce == (prevNonce + 1) {
```
pyrocto: This design makes it easy to mount a denial-of-service attack: just request a nonce and never use it.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,67 @@
...
+                  nonceCh!(nonce) | return!(nonce)
+                }
+              } |
+              contract @[*basicWallet, "getBalance"](return) = {
+                @[purse, "getBalance"]!(*return)
+              } |
+              contract @[*basicWallet, "deposit"](@amount, @src, success) = {
+                @[purse, "deposit"]!(amount, src, *success)
+              } |
+              contract @[*basicWallet, "withdraw"](@amount, @nonce, @sig, return) = {
```
pyrocto: The only thing a wallet should do is verify the signature on a message and then forward it to the underlying purse.  If you think there should be a withdraw method, please put it on the purse itself.

casper/src/main/rholang/NonNegativeNumber.rho
```diff
@@ -0,0 +1,42 @@
+//scalapackage coop.rchain.casper.genesis.contracts
+
+//This is a contract to explicitly implement the int >= 0 behavioural type
+contract @"NonNegativeNumber"(@init, return) = {
+  new this, valueStore in {
+    contract @[*this, "add"](@x, success) = {
+      match x >= 0 {
+        true => {
+          for(@v <- valueStore){
+            valueStore!(v + x) | success!(true)
```
pyrocto: At the moment, numbers are signed 32-bit integers.  You have to test for overflow.

casper/src/test/rholang/BasicWalletTest.rho
```diff
@@ -0,0 +1,123 @@
...
+      [
+        ["badAlgorithm", []]
+      ]
+    ) |
+  
+    @"BasicWallet"!(purse, "ed25519", pk, "wallet") |
+    for(@[wallet] <- @"wallet") {
+        @"wallet"!([wallet]) |
+        @"getBalance"!(wallet, "walletBalance") |
+        @"deposit"!(wallet, 300, otherPurse, "overdrawDep") |
```
pyrocto: Add test for overflow.

casper/src/test/rholang/NonNegativeNumberTest.rho
```diff
@@ -0,0 +1,92 @@
...
+      contract nnSub(return) = {
+        @[nn, "sub"]!(x, *return)
+      } |
+      return!(*nnSub)
+    }
+  } |
+
+  @"NonNegativeNumber"!(-5, "negBal") |
+  @"NonNegativeNumber"!(15, "properBal") |
+  @"NonNegativeNumber"!(100, "addTest") |
```
pyrocto: Add test for overflow.

casper/src/main/rholang/NonNegativeNumber.rho
odeac: It would be nice to only provide arbitrary precision numbers in Rholang, in order to avoid overflows and provide a safer environment for the developers.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Rev.scala
```diff
@@ -0,0 +1,26 @@
...
+    |    //public contract for making empty rev purses
+    |    contract @["Rev", "makePurse"](return) = {
+    |       @[revMint, "makePurse"]!(0, *return)
+    |    } |
+    |
+    |    //basic wallets which exist from genesis
+    |    ${wallets.map(Wallet.rhoCode).mkString(" |\n")}
+    |  }
+    |}
+  """.stripMargin
```
odeac: Can we put these snippets of code in resources instead of embedding them in Scala code?

casper/src/main/rholang/NonNegativeNumber.rho
```diff
@@ -0,0 +1,42 @@
...
+          }
+        }
+        _ => { success!(false)  }
+      }
+    } |
+    contract @[*this, "sub"](@x, success) = {
+      match x >= 0 {
```
odeac: Using if-then-else instead of match-true-false would be more readable

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,65 @@
...
+    match algorithm {
+      "ed25519" => {
+        cryptoVerifyCh!(["ed25519Verify"])
+      }
+      _ => {
+        cryptoVerifyCh!([])
+      }
+    } |
+    for(@maybeCryptoVerify <- cryptoVerifyCh) {
+      match maybeCryptoVerify {
```
odeac: isn't this a second match on the same condition?  

if at line 11 you just ```return ! ([])``` this match is not needed

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,65 @@
...
+  new cryptoVerifyCh in {
+    match algorithm {
+      "ed25519" => {
+        cryptoVerifyCh!(["ed25519Verify"])
+      }
+      _ => {
+        cryptoVerifyCh!([])
+      }
+    } |
+    for(@maybeCryptoVerify <- cryptoVerifyCh) {
```
odeac: What is the reason behind ```match ... | for (...) {...}```?

casper/src/main/rholang/NonNegativeNumber.rho
pyrocto: @odeac That's the plan.

casper/src/main/rholang/NonNegativeNumber.rho
pyrocto: Yes, here and everywhere else.

casper/src/main/rholang/BasicWallet.rho
birchmd: That's a good point, though it would then also leave the dangling `for(@cryptoVerify <- cryptoVerifyCh)`. Which if fine when we have a garbage collector, but I don't know when we plan on having that.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,79 @@
+//scalapackage coop.rchain.rholang.wallet
+
+//assumes standard purse API from MakeMint
+contract @"BasicWallet"(@purse, @algorithm, @pk, return) = {
```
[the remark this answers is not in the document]
birchmd: Yes. It should always be possible to wrap a purse into a wallet.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Rev.scala
birchmd: It uses Scala String Interpolation, so must be in the Scala code.

birchmd: All pure Rholang is separately given in the `src/main/rholang` and `src/test/rholang` directories

casper/src/test/rholang/BasicWalletTest.rho
birchmd: Done

casper/src/test/rholang/NonNegativeNumberTest.rho
birchmd: Done

birchmd:
@pyrocto @guardbotmk3 Thanks for all the feedback! I think I have addressed all the comments, PTAL

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,65 @@
...
+        cryptoVerifyCh!(["ed25519Verify"])
+      }
+      _ => {
+        cryptoVerifyCh!([])
+      }
+    } |
+    for(@maybeCryptoVerify <- cryptoVerifyCh) {
+      match maybeCryptoVerify {
+        []             => { return!([]) }
+        [cryptoVerify] => {
```
goral09: Can this be written as
```
if(algorithm == "ed25519Verify") {
  …
} else { return!([]) }
```

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,65 @@
...
+                  _ => { nonceCh!(prevNonce) | return!(false) }
+                }
+              }
+            } |
+            contract @[*basicWallet, "getNonce"](return) = {
+              for(@nonce <- nonceCh) {
+                nonceCh!(nonce) | return!(nonce)
+              }
+            } |
+            contract @[*basicWallet, "getBalance"](return) = {
```
dckc: Have you considered how this "method dispatch" idiom interacts with ocap patterns such as [revocable forwarder][1]?

In another idiom I've seen, the `"getBalance"` method name is sent along with the args. It's more clear to me how that would compose with revocable forwarder.

I started a [forum discussion on method dispatch idioms][2].


[1]: https://developer.rchain.coop/tutorial/#revocation
[2]: https://forum.rchain.coop/developer/rholang/ocap-happy-method-dispatch-idioms-in-rholang/

casper/src/main/rholang/BasicWallet.rho
birchmd: But in the future it won't be that simple. The present design easily extends to including other algorithms. In particular we will be include `secp256k1` as soon as you're done adding it to Rholang.

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -0,0 +1,65 @@
...
+                          _    => { nonceCh!(prevNonce) | return!(false) }
+                        }
+                      }
+                    }
+                  }
+                  _ => { nonceCh!(prevNonce) | return!(false) }
+                }
+              }
+            } |
+            contract @[*basicWallet, "getNonce"](return) = {
```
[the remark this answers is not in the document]
birchmd: Yes, that's right. We are defining a contract on the name `@[*basicWallet, "getNonce"]`. The way this gets used is that when you call `@"BasicWallet"` you get back the unforgable name `basicWallet`, which you can then use to construct the "compound names" where the method contracts can be accessed.

casper/src/main/rholang/BasicWallet.rho
birchmd: I don't think this idiom is less composable than sending string method names explicitly. In this model capabilities are  given to a name based on the contracts contained at "compound names" you can form with it. The fully featured object has contracts at all the compound names, but you can create a new name with fewer capabilities easily:
```
contract @"createRevocableForwarder"(baseName, allowedMethods, return) = {
  new isAlive, forwarder, addMethod, kill in {
    isAlive!(true) |
    contract addMethod(@method, success) = {
      //there is a minor problem here that you need to handle different
     //number of arguments somehow
      for(@args <- @[*forwarder, method]; _ <- isAlive) {
        @[*baseName, method]!(args) | 
        success!(true) |
        isAlive!(true) | 
        addMethod!(method, Nil)
      }
    } |
    contract kill(success) = {
      for(_ <- isAlive) {
        //Do not return any data to the `isAlive` channel, so the 
       //join above will no longer trigger.
        success!(true)
      }
    } |
    //initialize
    new methodsAdded in {
      @["LinkedList", "foreach"]!(allowedMethods, *addMethod, *methodsAdded) |
      for(_ <- methodsAdded) {
        return!([*forwarder, *kill])
      }
    }
  }
}
```
The contract above creates a revocable forwarder with any number of capabilities, which are specified at construction of the forwarder. There is a minor hiccup in that if the methods you wanted to forward took multiple arguments then `args` would need to be a list and you would need to deconstruct that list and pass its elements on to the underlying method.

dckc: Thanks for spelling that out in detail.

It seems a little awkward, but maybe that's mostly a matter of what I'm familiar with.

PR #972, opened 2018-06-16
CORE-665: rspace: add ReplayRSpace

henrytill:
## Overview
This PR adds the oft-mentioned `ReplayRSpace`.  

The key method on this new class is `rig` which takes a starting root hash along with a `trace.Log`.  

This resets the state `ReplayRSpace` of to given root and sets up `produce` and `consume` to work deterministically based upon the `trace.Log`.  

https://rchain.atlassian.net/browse/CORE-593
https://rchain.atlassian.net/browse/CORE-665 (sub-task)

### Notes
Apologies in advance to my reviewers for the size of the diff.  It's one of those rare times where it was essentially unavoidable.

Thinking that some of the refactoring I've done to modularize the innards of `produce` and `consume`  in `ReplayRSpace` can also happen in the "normal" `RSpace`.


rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -0,0 +1,305 @@
...
+        def runMatcher(maybeComm: Option[COMM]): Option[Seq[DataCandidate[C, A]]] = {
+          val channelToIndexedData = channels.map { (c: C) =>
+            c -> {
+              maybeComm match {
+                case Some(comm) =>
+                  store.getData(txn, Seq(c)).zipWithIndex.filter {
+                    case (Datum(_, _, source), _) => comm.produces.contains(source)
+                  }
+                case None =>
+                  Random.shuffle(store.getData(txn, Seq(c)).zipWithIndex)
```
KentShikama: You technically don't need to shuffle in this case right? It doesn't hurt to do so but.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -0,0 +1,305 @@
...
+                .map { comm: COMM =>
+                  runMatcher(Some(comm))
+                }
+                .collectFirst {
+                  case Some(x) => x
+                }
+
+            maybeDataCandidates match {
+              case None =>
+                storeWaitingContinuation(replays, consumeRef)
```
KentShikama: Btw do you not need to erase the element from the replay table?

KentShikama: Edit: This might not be necessary as even if your replay table isn't empty at the end, the root hash will probably be the same...but I got to think a bit.

Like if you have a comm with

`x!(5) | for (_ <- x) { Nil }`

this goes in the table as

```
[[ x!(5) ]] => [CommEvent(x!(5) , for (_ <- x) { Nil }]
[[ for (_ <- x) { Nil } ]] => [CommEvent(x!(5) , for (_ <- x) { Nil }]
```

right?

So even if you're still waiting for the other pair of the comm, you need to decrement that entry from the MultiSet in case you have multiple. This way your replay table in the end would be completely empty.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -0,0 +1,305 @@
...
+                .iterator()
+                .asScala
+                .map { comm: COMM =>
+                  runMatcher(Some(comm), produceRef, groupedChannels)
+                }
+                .collectFirst {
+                  case Some(x) => x
+                }
+
+            maybeProduceCandidate match {
```
KentShikama: Can we get the rename "maybeCandidateConsume" or are you planning to do it all at once later?

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+  def getCheckpoint(): Checkpoint = {
+    val root = store.getCheckpoint()
+    Checkpoint(root, Seq.empty)
+  }
+
+  def getReplayData: ReplayData = replayData.get
+
+  def rig(startRoot: Blake2b256Hash, log: trace.Log): Unit = {
```
KentShikama: I guess this name is appropriate but I always pause a bit on this word - what malicious thing are we doing :P

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
henrytill: Thanks for pointing this out.  I will update.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
henrytill: I guess I thought it was better to mimic the "normal" behavior of `RSpace` in this case.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
henrytill: I'd prefer to address renaming later, as it will span across both RSpaces...

KentShikama: Got it

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
KentShikama: Up to you - it seemed like you wanted to squeeze out every bit in terms of performance so. Maybe just add a comment saying it technically isn't necessary if you decide to keep it?

henrytill:
@KentShikama: I believe that we are now correctly removing things from the replay data, so this is ready for re-review.

rspace/src/main/scala/coop/rchain/rspace/history/Branch.scala
```diff
@@ -9,6 +9,8 @@ object Branch {
 
   val master: Branch = Branch("master")
 
+  val replay: Branch = Branch("replay")
```
lukasz-golebiewski: Constants should be in upper camel case by convention. Let's stick to that
https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

rspace/src/main/scala/coop/rchain/rspace/util/package.scala
```diff
@@ -19,4 +21,38 @@ package object util {
...
     t.foreach { case Some((k, data)) => k(data); case None => () }
+
+  /**
+    * Compare to `memcmp` in C/C++
+    *
+    * Based on:
+    * [[https://github.com/OpenTSDB/asynchbase/blob/f6a8ccb7e55ed9bc0aad265345da4c679e750055/src/Bytes.java#L549-L572]]
+    */
+  @SuppressWarnings(Array("org.wartremover.warts.Equals", "org.wartremover.warts.Return"))
+  def memcmp(a: Array[Byte], b: Array[Byte]): Int = {
```
lukasz-golebiewski: Why not `compare`?

rspace/src/main/scala/coop/rchain/rspace/internal.scala
```diff
@@ -84,4 +86,37 @@ object internal {
...
       codecSeq(codecDatum(codecA)) ::
       codecSeq(codecWaitingContinuation(codecP, codecK))).as[GNAT[C, P, A, K]]
+
+  type MultisetMultiMap[K, V] = mutable.HashMap[K, Multiset[V]]
+
+  object MultisetMultiMap {
+    def empty[K, V]: MultisetMultiMap[K, V] = new mutable.HashMap[K, Multiset[V]]()
+  }
+
+  implicit class RichMultisetMultiMap[K, V](msmm: MultisetMultiMap[K, V]) {
```
lukasz-golebiewski: We could spare an allocation here by making this a value class

rspace/src/main/scala/coop/rchain/rspace/history/Branch.scala
henrytill: Oh there's a convention here!  FANTASTIC.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -0,0 +1,321 @@
...
+              store.eventsCounter.registerProduceCommEvent()
+              val commRef = COMM(consumeRef, dataCandidates.map(_.datum.source))
+              assert(comms.contains(commRef), "COMM Event was not contained in the trace")
+              if (!persistK) {
+                store.removeWaitingContinuation(txn, channels, continuationIndex)
+              }
+              dataCandidates
+                .sortBy(_.datumIndex)(Ordering[Int].reverse)
+                .foreach {
+                  case DataCandidate(candidateChannel, Datum(_, persistData, prodRef), dataIndex) =>
```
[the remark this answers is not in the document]
henrytill: Indeed. Will fix.

rspace/src/main/scala/coop/rchain/rspace/util/package.scala
henrytill: Oh it's specifically aping `memcmp` semantics.

rspace/src/main/scala/coop/rchain/rspace/internal.scala
henrytill: Good call.

PR #976, opened 2018-06-17
[RHOL-427]: Add maps methods

goral09:
## Overview

Adds methods working on Rholang `Map`s. Methods added:
- `union` , input: another map, returns: union of two maps removing duplicates,
- `contains` , input: key, returns: true or false
- `get`, input: key, returns: element if key in map, otherwise `Nil`

[JIRA issue](https://rchain.atlassian.net/browse/RHOL-427)

### Notes

I didn't  implement `add` method as I wasn't sure what type should be the input (`map`, `tuple`? ). I've decided that users can leverage `union` method with single element map.


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -664,22 +665,21 @@ object Reduce {
     }
 
     private[this] def diff: MethodType = { (p: Par, args: Seq[Par]) => (env: Env[Par]) =>
+      def locallyFreeUnion(base: Coeval[BitSet], other: Coeval[BitSet]): Coeval[BitSet] =
```
guardbotmk3: This doesn't appear necessary here.

goral09: It is, few lines below. I extracted it with the intent to reuse for `EMap`.

goral09:
@guardbotmk3 Added `diff` method on maps. It's not the same as removing set of keys but will achieve same result.

PR #977, opened 2018-06-17
Moved some functions to the "shared" project. Added unit-tests.

ys-pyrofex:
## Overview
Moved reusable library functions from the RSpace.util to the "shared" project. 

https://rchain.atlassian.net/browse/CORE-621

### Notes
There is a behavior that looks wrong. Please take a look into <<"dropIndex" should "ignore invalid indexes">> unit-test and let me know what do you think about it.


ys-pyrofex:
codecov seems missing something - comparing to previous version - functionality left unchanged, the only new code are unit-tests.

PR #985, opened 2018-06-18
Rholang: Add Blake2b512Random class and tests.

guardbotmk3:
Blake2b512Random is a specific purpose splittable random number generator. It
allows us to split along byte boundaries, and follows the basic design of
tf-random. In our case, we can only extract 256-bit sequences. We will be using
these sequences to generate predictable unforgeable names. All of the test
vectors were created by using a hex-editor to construct appopriate binary files
and then hashing them with b2sum.

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Block.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Block.scala
@@ -1,0 +1,152 @@
+package coop.rchain.rholang.crypto
+
+import coop.rchain.rspace.Serialize
+import org.bouncycastle.util.Pack
+
+/**
+Block oriented Blake2b512 class.
+
+Only supports hashes of data a positive number of whole blocks.
+  */
+class Blake2b512Block {
+  import Blake2b512Block._
+  val chainValue: Array[Long] = new Array[Long](CHAIN_VALUE_LENGTH)
+  private var t0: Long        = 0
+  private var t1: Long        = 0
+
+  // block must be 128 bytes long
+  def update(block: Array[Byte], offset: Int): Unit =
+    compress(block, offset, chainValue, false)
+
+  // gives the output as if block were the last block processed, but does not
+  // invalidate or modify internal state
+  def peekFinal(block: Array[Byte], inOffset: Int, output: Array[Byte], outOffset: Int): Unit = {
+    val tempChainValue: Array[Long] = new Array[Long](8)
+    compress(block, inOffset, tempChainValue, true)
+    Pack.longToLittleEndian(tempChainValue, output, outOffset)
+  }
+
+  private[this] def compress(msg: Array[Byte],
+                             offset: Int,
+                             newChainValue: Array[Long],
+                             peek: Boolean): Unit = {
+    val internalState: Array[Long] = new Array[Long](BLOCK_LENGTH_LONGS)
+    def g(m1: Long, m2: Long, posA: Int, posB: Int, posC: Int, posD: Int): Unit = {
+      def rotr64(x: Long, rot: Int): Long =
+        x >>> rot | (x << (64 - rot))
+      internalState(posA) = internalState(posA) + internalState(posB) + m1;
+      internalState(posD) = rotr64(internalState(posD) ^ internalState(posA), 32);
+      internalState(posC) = internalState(posC) + internalState(posD);
+      internalState(posB) = rotr64(internalState(posB) ^ internalState(posC), 24);
+      internalState(posA) = internalState(posA) + internalState(posB) + m2;
+      internalState(posD) = rotr64(internalState(posD) ^ internalState(posA), 16);
+      internalState(posC) = internalState(posC) + internalState(posD);
+      internalState(posB) = rotr64(internalState(posB) ^ internalState(posC), 63);
+    }
+
+    Array.copy(chainValue, 0, internalState, 0, CHAIN_VALUE_LENGTH)
+    Array.copy(IV, 0, internalState, CHAIN_VALUE_LENGTH, 4)
+    val f0    = if (peek) 0xffffffffffffffffL else 0x0L
+    val newT0 = t0 + BLOCK_LENGTH_BYTES
+    val newT1 = if (newT0 == 0) t1 + 1 else t1
+    internalState(12) = newT0 ^ IV(4)
+    internalState(13) = newT1 ^ IV(5)
+    internalState(14) = f0 ^ IV(6)
+    internalState(15) = IV(7)
+
+    val m: Array[Long] = new Array(BLOCK_LENGTH_LONGS)
+    0.until(BLOCK_LENGTH_LONGS).foreach { i =>
+      m(i) = Pack.littleEndianToLong(msg, offset + i * 8)
+    }
+
+    0.until(ROUNDS).foreach { round =>
+      // columns
+      g(m(SIGMA(round)(0).toInt), m(SIGMA(round)(1).toInt), 0, 4, 8, 12);
+      g(m(SIGMA(round)(2).toInt), m(SIGMA(round)(3).toInt), 1, 5, 9, 13);
+      g(m(SIGMA(round)(4).toInt), m(SIGMA(round)(5).toInt), 2, 6, 10, 14);
+      g(m(SIGMA(round)(6).toInt), m(SIGMA(round)(7).toInt), 3, 7, 11, 15);
+      // diagonals
+      g(m(SIGMA(round)(8).toInt), m(SIGMA(round)(9).toInt), 0, 5, 10, 15);
+      g(m(SIGMA(round)(10).toInt), m(SIGMA(round)(11).toInt), 1, 6, 11, 12);
+      g(m(SIGMA(round)(12).toInt), m(SIGMA(round)(13).toInt), 2, 7, 8, 13);
+      g(m(SIGMA(round)(14).toInt), m(SIGMA(round)(15).toInt), 3, 4, 9, 14);
+    }
+    if (!peek) {
+      t0 = newT0
+      t1 = newT1
+    }
+    0.until(CHAIN_VALUE_LENGTH).foreach { i =>
+      newChainValue(i) = chainValue(i) ^ internalState(i) ^ internalState(i + 8)
+    }
+  }
+
+}
+
+object Blake2b512Block {
+  def apply(): Blake2b512Block = {
+    val result = new Blake2b512Block
+    Array.copy(IV, 0, result.chainValue, 0, CHAIN_VALUE_LENGTH)
+    result.chainValue(0) ^= PARAM_VALUE
+    result
+  }
+
+  def apply(src: Blake2b512Block): Blake2b512Block = {
+    val result = new Blake2b512Block
+    Array.copy(src.chainValue, 0, result.chainValue, 0, CHAIN_VALUE_LENGTH)
+    result.t0 = src.t0
+    result.t1 = src.t1
+    result
+  }
+
+  implicit val serialize: Serialize[Blake2b512Block] = new Serialize[Blake2b512Block] {
+    override def encode(block: Blake2b512Block) = {
+      val result: Array[Byte] = new Array[Byte](80)
+      Pack.longToLittleEndian(block.chainValue, result, 0)
+      Pack.longToLittleEndian(block.t0, result, DIGEST_LENGTH_BYTES)
+      Pack.longToLittleEndian(block.t1, result, DIGEST_LENGTH_BYTES + 8)
+      result
+    }
+
+    override def decode(bytes: Array[Byte]): Either[Throwable, Blake2b512Block] =
+      if (bytes.length != 80)
+        Left(new Error("Invalid decode of Blake2b512Block"))
+      else {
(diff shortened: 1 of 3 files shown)


pyrocto:
Which implementation of blakeb2 did you base this on?

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Block.scala
```diff
@@ -0,0 +1,152 @@
...
+      val result: Array[Byte] = new Array[Byte](80)
+      Pack.longToLittleEndian(block.chainValue, result, 0)
+      Pack.longToLittleEndian(block.t0, result, DIGEST_LENGTH_BYTES)
+      Pack.longToLittleEndian(block.t1, result, DIGEST_LENGTH_BYTES + 8)
+      result
+    }
+
+    override def decode(bytes: Array[Byte]): Either[Throwable, Blake2b512Block] =
+      if (bytes.length != 80)
+        Left(new Error("Invalid decode of Blake2b512Block"))
```
pyrocto: Did you mean "Invalid encoding"?  It's not this function's fault that the given input's length is wrong.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Block.scala
```diff
@@ -0,0 +1,152 @@
+package coop.rchain.rholang.crypto
+
+import coop.rchain.rspace.Serialize
+import org.bouncycastle.util.Pack
+
+/**
+Block oriented Blake2b512 class.
+
+Only supports hashes of data a positive number of whole blocks.
+  */
```
henrytill: You've based this off `Blake2bDigest.java` in bouncycastle, no?

Might be good to link to that code, especially to help ascertain the provenance of your constants like `SIGMA` and `IV`.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Block.scala
```diff
@@ -0,0 +1,152 @@
...
+
+  // gives the output as if block were the last block processed, but does not
+  // invalidate or modify internal state
+  def peekFinal(block: Array[Byte], inOffset: Int, output: Array[Byte], outOffset: Int): Unit = {
+    val tempChainValue: Array[Long] = new Array[Long](8)
+    compress(block, inOffset, tempChainValue, true)
+    Pack.longToLittleEndian(tempChainValue, output, outOffset)
+  }
+
+  private[this] def compress(msg: Array[Byte],
```
henrytill: Delightfully easy to compare with the original.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
+package coop.rchain.rholang.crypto
+
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.LongBuffer
+import org.bouncycastle.crypto.digests.Blake2bDigest
```
henrytill: unused

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.bouncycastle.util.Pack
+
+import coop.rchain.rspace.Serialize
+
+import coop.rchain.crypto.codec._
```
henrytill: unused

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+    val result = new Blake2b512Random(Blake2b512Block(digest), cloneBlock)
+    result.pathView.position(pathView.position)
+    result
+  }
+
+  def hash(): Unit = {
+    digest.peekFinal(lastBlock.array(), 0, hashArray, 0)
+    val low = countView.get(0)
+    if (low == -1) {
+      val high = countView.get(1)
```
henrytill: Currently untested.  Is there a way to test this code path?

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+  def splitByte(index: Byte): Blake2b512Random = {
+    val split = copy()
+    split.addByte(index)
+    split
+  }
+
+  def splitShort(index: Short): Blake2b512Random = {
+    val split  = copy()
```
henrytill: Untested.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+    val result = new Blake2b512Random(Blake2b512Block(), ByteBuffer.allocate(128))
+    val range  = Range(offset, offset + length - 127, 128)
+    range.foreach { base =>
+      result.digest.update(init, base)
+    }
+    val partialBase =
+      if (range.isEmpty)
+        offset
+      else
+        range.last + 128
```
henrytill: As with the above, also as of yet unexercised.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+
+    // If there is any remainder:
+    if (offset + length != partialBase) {
+      val padded = new Array[Byte](128)
```
henrytill: Same as above.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  def peek: ByteBuffer = lastBlock.asReadOnlyBuffer()
+}
+
+object Blake2b512Random {
+  def apply(init: Array[Byte], offset: Int, length: Int): Blake2b512Random = {
+    val result = new Blake2b512Random(Blake2b512Block(), ByteBuffer.allocate(128))
+    val range  = Range(offset, offset + length - 127, 128)
+    range.foreach { base =>
+      result.digest.update(init, base)
```
henrytill: All the tests have a `range` with size 0, so this code path hasn't been exercised yet.

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
```diff
@@ -0,0 +1,138 @@
...
+      Array.copy(init, partialBase, padded, 0, offset + length - partialBase)
+      result.digest.update(padded, 0)
+    }
+    result
+  }
+  def apply(init: Array[Byte]): Blake2b512Random =
+    apply(init, 0, init.length)
+
+  implicit val serialize: Serialize[Blake2b512Random] = new Serialize[Blake2b512Random] {
+    def encode(rand: Blake2b512Random): Array[Byte] = {
```
henrytill: Might be good to write a round-trip property test for these.

I do something along those lines here (using `Codec` instead of `Serialize`, but the pattern is roughly the same):
https://github.com/rchain/rchain/blob/aa8abab6012707efccb3431cda5c3afe3b3880ae/rspace/src/test/scala/coop/rchain/rspace/history/PointerBlockTests.scala#L19-L28


using:
https://github.com/rchain/rchain/blob/aa8abab6012707efccb3431cda5c3afe3b3880ae/rspace/src/test/scala/coop/rchain/rspace/test/package.scala#L45-L46

You'll need to write an instance of `Arbitrary` for `Blake2b512Random`.  We have a bunch of examples of those for our `rspace` types [here](https://github.com/rchain/rchain/blob/aa8abab6012707efccb3431cda5c3afe3b3880ae/rspace/src/test/scala/coop/rchain/rspace/test/ArbitraryInstances.scala).  Some are nicer than others in presentation, but they all get the job done.

henrytill:
@pyrocto 

> Which implementation of blakeb2 did you base this on?

Looks like [this one](https://github.com/bcgit/bc-java/blob/a258da7f9145eacba46c2705be564e7f3581a9b1/core/src/main/java/org/bouncycastle/crypto/digests/Blake2bDigest.java) to me.

henrytill:
@guardbotmk3 

> All of the test vectors were created by using a hex-editor to construct appopriate binary files and then hashing them with b2sum.

It would be cool if you could provide instructions for doing this for ppl who want to follow along at home.

crypto/src/test/scala/coop/rchain/crypto/hash/Blake2b512RandomTest.scala
```diff
@@ -43,4 +77,62 @@ class Blake2b512RandomSpec extends FlatSpec with Matchers {
...
+  "Multi block prefix" should "give a predictable result" in {
+    val b2Random = Blake2b512Random(multiBlockMsg)
+    val res1     = b2Random.next()
+    val res2     = b2Random.next()
+    Base16.encode(res1) should be(
+      "c6b40c41cb9e3c078936c38bddccce3a017de1e4e751799710c9b50a70feb703")
+    Base16.encode(res2) should be(
+      "b6756c6ebe5e1b2f05ab2212e1a7e8aa126db5637a0111d75715267986df6a95")
+  }
+  val multiBlockAndPartialMsg: Array[Byte] =
```
goral09: What's the difference between `multiBlockAndPartialMsg`, `multiBlockMsg`, `blockAndPartMsg` , `singleBlockMsg` and `partialMsg` ? Is it the length of the input? They're all tested in the same manner.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -143,56 +146,73 @@ object Reduce {
...
+    override def eval(par: Par)(implicit env: Env[Par], rand: Blake2b512Random): M[Unit] = {
+      val filteredExprs = par.exprs.filter { expr =>
+        expr.exprInstance match {
+          case _: EVarBody    => true
+          case _: EEvalBody   => true
+          case _: EMethodBody => true
+          case _              => false
+        }
+      }
+      val starts = Vector(par.receives.size,
```
goral09: Can you add a comment explaining why do you do that?

rholang/src/main/scala/coop/rchain/rholang/crypto/Blake2b512Random.scala
guardbotmk3: Done. Found some bugs even.

crypto/src/test/scala/coop/rchain/crypto/hash/Blake2b512RandomTest.scala
guardbotmk3: Yes, it's just the length.
partialMsg is less than a block.
singleBlockMsg is exactly 1 block.
blockAndPartMsg is just over 1 block.
multiblockMsg is 3 (I think) blocks,
multiblockAndPartialMsg is longer than 3 blocks, but not a 4th

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: Why we accumulate the starts vector?

guardbotmk3:
OK, I put some extra things into this PR that didn't belong. I pulled them out and have only the crypto here, but it is the complete crypto that I intend to use.

goral09:
@guardbotmk3 https://travis-ci.org/rchain/rchain/jobs/397973484#L3917

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
```diff
@@ -0,0 +1,250 @@
...
+import com.google.protobuf.ByteString
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.LongBuffer
+import org.bouncycastle.util.Pack
+import org.scalacheck.{Arbitrary, Gen}
+import scalapb.TypeMapper
+
+import scala.annotation.tailrec
+
```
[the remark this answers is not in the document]
pyrocto: Like, just copy your initial PR message in there.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
```diff
@@ -0,0 +1,257 @@
...
+  val pathView: ByteBuffer = lastBlock.duplicate()
+  pathView.limit(112)
+  val countView: LongBuffer = {
+    val lastDuplicate = lastBlock.duplicate()
+    lastDuplicate.position(112)
+    lastDuplicate.slice().order(ByteOrder.LITTLE_ENDIAN).asLongBuffer()
+  }
+
+  val hashArray: Array[Byte] = new Array[Byte](64)
+  var position: Int          = 0
```
goral09: @guardbotmk3 I've just noticed this -- this `var` is not private!

PR #987, opened 2018-06-19
CORE-639 pruneHistory -> collapseTrieUpdates

ys-pyrofex:
## Overview
Good naming -> code that's easy to understand

https://rchain.atlassian.net/browse/CORE-639

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/IStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/IStore.scala
@@ -68,7 +68,8 @@ trait IStore[C, P, A, K] {
 
   val trieBranch: Branch
 
-  private[rspace] def pruneHistory(in: Seq[TrieUpdate[C, P, A, K]]): Seq[TrieUpdate[C, P, A, K]] =
+  private[rspace] def collapseTrieUpdates(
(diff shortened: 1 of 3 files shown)


rspace/src/main/scala/coop/rchain/rspace/IStore.scala
```diff
@@ -68,7 +68,8 @@ trait IStore[C, P, A, K] {
 
   val trieBranch: Branch
 
-  private[rspace] def pruneHistory(in: Seq[TrieUpdate[C, P, A, K]]): Seq[TrieUpdate[C, P, A, K]] =
+  private[rspace] def collapseTrieUpdates(
```
lukasz-golebiewski: This method takes a sequence of `TrieUpdates` as parameter, so I don't think this has to be mentioned additionally in it's name.

ys-pyrofex: I'm afraid "IStore.collapse" will say nothing to anyone who will read function names and he/she will require additional reading to understand what actually this function doing.

henrytill: I'm sympathetic to both sides of this debate.  I might be more concerned about calling it `collapse` if it were part of the public API.   Even then, looking at the input parameter's type does provide roughly the same context we're getting with the explicit name.

Let's go ahead and change it to `collapse`.

PR #989, opened 2018-06-19
[RHOL-435]: Adds support for using secp256k1 crypto library

goral09:
## Overview
Adds support for using secp256k1 crypto library. For now build process just builds library for linux x64. Need to be enriched with build process for other operating systems.

One has to call `secp256k1/secp256k1:generate` (just as one would execute `bnfc:generate`) to build native library and get `*.so` file available in the `crypto` project.

[JIRA issue](https://rchain.atlassian.net/browse/RHOL-435).

### Notes
For now build process just builds library for linux x64. Need to be enriched with build process for other operating systems.


project/Dependencies.scala
```diff
@@ -51,6 +51,7 @@ object Dependencies {
   val scodecBits          = "org.scodec"                 %% "scodec-bits"               % "1.1.5"
   val shapeless           = "com.chuusai"                %% "shapeless"                 % "2.3.2"
   val weupnp              = "org.bitlet"                  % "weupnp"                    % "0.1.+"
+  val apacheLang     = "org.apache.commons" % "commons-lang3" % "3.7"
```
henrytill: alignment

secp256k1Native/src/main/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -0,0 +1,446 @@
+/*
```
henrytill: Are these files not part of the submodule repo? 

If you are just trying to move them into the "right" place for sbt, I would consider keeping them in their current location and changing where sbt looks for them.  This approach seems brittle in the face of upstream changes.

secp256k1Native/src/main/java/org/bitcoin/Secp256k1Context.java
```diff
@@ -0,0 +1,79 @@
+/*
```
henrytill: See above.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "secp256k1Native/secp256k1"]
```
henrytill: Oh.  Has there been a discussion about using submodules?  

I don't mind them, but I generally only use them in team settings where everybody is OK with them and are reasonably comfortable using them.

project/Secp256k1.scala
```diff
@@ -0,0 +1,53 @@
...
+  def cleanDirs(dirs: File*): Unit = dirs.foreach(dir => s"rm -rf $dir".!)
+
+  def cleanProject(projectDir: File, dirs: File*) = {
+    cleanDirs(dirs:_*)
+    val makeDistCleanCmd = s"make -C ${projectDir.getAbsolutePath} distclean"
+    makeDistCleanCmd.!
+  }
+
+  def secp256k1Generate(baseDir: File) = {
+    val autogenCmd: String = "./autogen.sh"
```
henrytill: So now we depend on `autotools`

project/Secp256k1.scala
```diff
@@ -0,0 +1,53 @@
...
+    val autogenProcess = Process(autogenCmd, new File(s"${baseDir.getAbsolutePath}/secp256k1/"))
+    val configureCmd: String = "./configure --enable-jni --enable-experimental --enable-module-ecdh"
+    val configureProcess = Process(configureCmd, new File(s"${baseDir.getAbsolutePath}/secp256k1/"))
+    val makeCmd: String = s"make -C ${baseDir.getAbsolutePath}/secp256k1"
```
henrytill: And `make`

project/Secp256k1.scala
```diff
@@ -0,0 +1,53 @@
...
+    val makeProcess = Process(makeCmd)
+    val binariesSourceFolder = baseDir / "secp256k1" / ".libs"
+    val resourcesFolder = baseDir / "src" / "main" / "resources"
+    val makeResourcesFolderCmd = s"mkdir ${resourcesFolder.getAbsolutePath}"
+    val mkResourcesFolderProcess = Process(makeResourcesFolderCmd)
+    val copyBinariesCmd: String = s"cp libsecp256k1.so ${resourcesFolder.getAbsolutePath}/"
```
henrytill: This doesn't work on macOS or Windows

secp256k1Native/src/main/java/org/bitcoin/NativeSecp256k1.java
goral09: Not exactly the same as original ones, I added few lines that should take care of loading proper binaries depending on the OS.

.gitmodules
goral09: I thought that was the most reasonable thing to do in this case.

project/Secp256k1.scala
goral09: Yeah, that's what the build process requires.

odeac: Do we need to have secp256k1 built during our build process? Can't we just add it to the prerequisites for the dev machine?

goral09:
@odeac 
> Do we need to have secp256k1 built during our build process? Can't we just add it to the prerequisites for the dev machine?

I would prefer to have it that way but I think that general consensus was that we should ship crypto libraries on our own. Is that right @henrytill @guardbotmk3  ?

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -66,15 +66,16 @@ object SystemProcesses {
 
   //TODO(mateusz.gorski): we decided to look into delivering secp256k1 library (https://github.com/bitcoin-core/secp256k1)
   // as separate jar in the future
```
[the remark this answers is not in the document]
goral09: No, it's not. Thanks.

goral09:
Closing in favor of [sec256k1-java library](https://github.com/rchain/secp256k1Java).

PR #991, opened 2018-06-19
Use Base16 for byte address encoding to String

sebtomba:
## Overview
Use `coop.rchain.crypto.codec.Base16` instead of `map("%02x".format(_))`

https://rchain.atlassian.net/browse/CORE-671

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/PeerNode.scala
+++ b/comm/src/main/scala/coop/rchain/comm/PeerNode.scala
@@ -3,10 +3,12 @@ package coop.rchain.comm
 import java.net.InetSocketAddress
 import com.netaporter.uri.Uri
 import scala.util.control.NonFatal
+import coop.rchain.crypto.codec.Base16
 
 // TODO: Add Show instance
 final case class NodeIdentifier(key: Seq[Byte]) {
-  override def toString: String = key.map("%02x".format(_)).mkString
+  private val keyString         = Base16.encode(key.toArray)
+  override def toString: String = keyString
 }
 
 object NodeIdentifier {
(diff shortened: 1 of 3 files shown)


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -89,15 +89,14 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
...
-      }
-      .map("%02x".format(_))
-      .mkString
+    Base16.encode(
+      publicKey
+        .flatMap(CertificateHelper.publicAddress)
+        .getOrElse {
+          println("Certificate must contain a secp256r1 EC Public Key")
+          System.exit(1)
+          Array[Byte]()
```
goral09: Maybe split that to the logic of extracting public key and encoding it in base16. Also I don't like the `System.exit(1)` in `getOrElse` clause. Would the default empty array from line `98` ever by returned if there's a `System.exit(1)` line before that?

comm/src/main/scala/coop/rchain/comm/PeerNode.scala
```diff
@@ -3,10 +3,12 @@ package coop.rchain.comm
 import java.net.InetSocketAddress
 import com.netaporter.uri.Uri
 import scala.util.control.NonFatal
+import coop.rchain.crypto.codec.Base16
 
 // TODO: Add Show instance
 final case class NodeIdentifier(key: Seq[Byte]) {
-  override def toString: String = key.map("%02x".format(_)).mkString
+  private val keyString         = Base16.encode(key.toArray)
```
rabbitonweb: this field is used only here, why dont keep it inlined as it was before?

sebtomba: It's an optimization for performance. We don't want to encode the bytes each time the program calls `toString`.

rabbitonweb: 👍

PR #992, opened 2018-06-19
Update for logical connectives "and" and "or", and to streamline the matching algorithm.

differentialderek:
## Overview
This update does a few things:
(1) It gives syntax/semantics for the logical "and" and "or" connectives for pattern-matching, including correct associativity.
(2) Streamlines, abstracts and generalizes the matching algorithm to allow for different types of matching in different places. This allows multiple types of matching in rholang to use the same basic algorithm. For example, matching channels is not the same as matching a pattern to a process. The way we have constructed this makes it possible to use the same basic "matching engine" to match in both places, with only slight modifications which are given as inputs to the matching function.
(3) Adds a function that matches channels, using this abstraction from (2).
(4) Generalizes and simplifies the par matching algorithm, in particular to allow for nested parallel process checking. For example, the following code should fail to match:
`for(@ 
     for(@
         for(@ @g!(e) | Nil <- @10 ){ Nil }
     | @2!(p) | Nil <- @z ){ @p!(y) }
  | @3!(z) <- @10){ @y!(y) | @x!(10) | @z!(3) }
| @10!(for(@ @2!(w) | Nil | for(@ @20!(d) | Nil <- @10 ){ Nil }  <- @10 ){ @w!(Nil) } )`

N/a

### Notes
In the next PR we will include more streamlining of the matching algorithm to also generalize matching for parallel processes, as well as adding a matching function to match channels (this function matches in a stricter sense than pattern/process matching).


differentialderek:
I meant to break up these commits into separate pull requests. Sorry about that.

rholang/src/main/k/rholang.k
```diff
@@ -577,40 +609,64 @@ rule <k> #(MATCHFAIL) ~> #(MATCHPAUSE) ~> match P:HigherProcOrChan { L:HigherMat
...
-
+rule <thread> ... <k> ... listmatch[ S:String ]( .HigherProcs ; D:Chan, E:Chans )
+                      => #(MATCHFAIL)[S] ... </k> ... </thread>
+rule <thread> ... <k> ... listmatch[ S:String ]( P:HigherProc, Q:HigherProcs ; .Chans )
+                      => #(MATCHFAIL)[S] ... </k> ... </thread>
+
+// We also split things up if there are logical connectives: "and's" or "or's"
+// (i) The "and's" are straightforward, since we need to match with all of them
+rule <thread> ... <k> ... match[S:String]( P:HigherProcOrChan ; C:Chan /\ D:Chan )
+               => match[S](P ; C) ~> match[S]( P ; D ) ... </k> ... </thread>
```
pyrocto: Does matching "short circuit" here?  That is, if `match[S](P ; C)` fails will it still try to `match[S]( P ; D )`?

differentialderek: No, it won't. It will delete the second match checker and say that the patterns did not match.

differentialderek: It rewrites to `#(MATCHFAIL)[ String ]` which triggers that.

rholang/src/main/k/rholang.k
```diff
@@ -130,19 +139,23 @@ syntax CanBeStdMatched ::= Id
...
-              | "DoTheseMatch" "(" HigherProcOrChan ";" HigherProcOrChan ")"
-              | "matchstdform" "(" CanBeStdMatched ";" CanBeStdMatched ")" [strict]
-              | "tfmatchstdform" "(" CanBeStdMatched ";" CanBeStdMatched ")" [strict]
-              | "matchstdform" "(" CanBeStdMatched ";" CanBeStdMatched ";" Bool ")" [strict(3)]
-              | "tfmatchstdform" "(" CanBeStdMatched ";" CanBeStdMatched ";" Bool ")" [strict(3)]
+              | ChanNoVars
+
+syntax Fun  ::=
+              // General matching functions which use the "matching engine" below
+                "match" "[" String "]" "(" HigherProcOrChan ";" HigherProcOrChan ")"
```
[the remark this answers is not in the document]
differentialderek: Good idea, thank you. Just changed it.

PR #999, opened 2018-06-20
Quit with a human readable error message when data dir is not read/write

sebtomba:
## Overview
Quit with a human readable error message when data dir has no read/write access

https://rchain.atlassian.net/browse/CORE-500


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -36,6 +36,13 @@ import scala.util.{Failure, Success, Try}
 
 class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
 
+  // Check if data_dir has read/write access
+  if (!conf.run.data_dir().toFile.canRead
```
rabbitonweb: Make this part of node program itself. We want to have it undercontrol, tested & maintained.  We dont want to add arbitrary pieces of code to `node` runtime.

If you don't see it as part of the "business logic" then allow program to throw an exception and catch it at the end of the world. One way or the other.

PR #1009, opened 2018-06-21
[RHOL-444]: Fix bug in if-else desugaring

goral09:
## Overview
`Par`s from the environment were concatenated inside the `if` condition instead of to the whole `if-else` expression.

[JIRA issue](https://rchain.atlassian.net/browse/RHOL-444).

### Notes
None

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -234,7 +234,7 @@ object ProcNormalizeMatcher {
 
     def normalizeIfElse(valueProc: Proc,
                         trueBodyProc: Proc,
-                        falseBodyProc: Proc): M[ProcVisitOutputs] =
+                        falseBodyProc: Proc, input: ProcVisitInputs): M[ProcVisitOutputs] =
       for {
         targetResult <- normalizeMatch[M](valueProc, input)
         trueCaseBody <- normalizeMatch[M](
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -762,8 +762,18 @@ object ProcNormalizeMatcher {
...
+          .map(normalized => ProcVisitOutputs(
+            input.par ++ normalized.par,
+            normalized.knownFree
+          ))
+      case p: PIfElse =>
+        normalizeIfElse(p.proc_1, p.proc_2, p.proc_3, input.copy(par = VectorPar()))
+          .map(normalized => ProcVisitOutputs(
+            input.par ++ normalized.par,
+            normalized.knownFree
+          ))
```
odeac: You should be able to reduce this code duplication

goral09: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -234,7 +234,7 @@ object ProcNormalizeMatcher {
 
     def normalizeIfElse(valueProc: Proc,
                         trueBodyProc: Proc,
-                        falseBodyProc: Proc): M[ProcVisitOutputs] =
+                        falseBodyProc: Proc, input: ProcVisitInputs): M[ProcVisitOutputs] =
       for {
         targetResult <- normalizeMatch[M](valueProc, input)
```
guardbotmk3: Wouldn't it be cleaner just to adjust the value here rather than passing input?

goral09: Functions in FP should depend only on their input values.

PR #1011, opened 2018-06-21
Add tutorial for createCheckpoint and reset

lukasz-golebiewski:
## Overview
Adds tutorial and an example for createCheckpoint and reset

https://rchain.atlassian.net/browse/CORE-624

### Notes

### DIFF
--- a/rspace/src/main/tut/Tutorial.md
+++ b/rspace/src/main/tut/Tutorial.md
@@ -441,7 +441,7 @@ We can now create a checkpoint and store it's root.
 
 The first `produceAlice` operation should be able to find data stored by the consume.
 ```tut
-def produceAlice = space2.produce(Channel("friends"), alice, persist = false)
+def produceAlice(): Option[(Printer, Seq[Entry])] = space2.produce(Channel("friends"), alice, persist = false)
 assert(produceAlice.isDefined)
 ```
 
--- a/rspace/src/main/tut/Tutorial.md
+++ b/rspace/src/main/tut/Tutorial.md
@@ -434,18 +434,18 @@ val cres =
 assert(cres.isEmpty)
 ```
 
-We can now create a checkpoint and store it's root
+We can now create a checkpoint and store it's root.
 ```tut
     val checkpointHash = space2.createCheckpoint.root
 ```
 
-The first `produceAlice` operation should be able to find data stored by the consume, hence:
+The first `produceAlice` operation should be able to find data stored by the consume.
 ```tut
 def produceAlice = space2.produce(Channel("friends"), alice, persist = false)
 assert(produceAlice.isDefined)
 ```
 
-Running the same operation again shouldn't return anything, as data hasn't been persisted:
+Running the same operation again shouldn't return anything, as data hasn't been persisted.
 ```tut
 assert(produceAlice.isEmpty)
 ```
@@ -454,7 +454,7 @@ Every following repetition of the operation above should yield an empty result.
 assert(produceAlice.isEmpty)
 ```
 
-After re-setting the RSpace to the state from the saved checkpoint the first produce operation should again return an non-empty result:
+After re-setting the RSpace to the state from the saved checkpoint the first produce operation should again return an non-empty result.
 ```tut
 space2.reset(checkpointHash)
 assert(produceAlice.isDefined)
--- a/rspace/src/main/tut/Tutorial.md
+++ b/rspace/src/main/tut/Tutorial.md
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
 println(space.store.toMap)
 ```
 
+### History & rollback
+
+It is possible to save the current state of RSpace in the form of a checkpoint (root hash of a given Merkle Patricia Trie).
+```scala
+val checkpoint = space.createCheckpoint()
+val checkpointHash = checkpoint.root
+```
+
+To rollback the state of the RSpace to a given checkpoint one simply calls the `reset` method with the hash of the root of the checkpoint provided as parameter.
+```scala
+space.reset(checkpointHash)
+```
+
+Let's see how this works in practice. We'll start by creating a new, untouched RSpace followed by a consume operation which should put data and a continuation at given channel.
+```tut
+    val storePath2: Path = Files.createTempDirectory("rspace-address-book-example-")
+    val store2: LMDBStore[Channel, Pattern, Entry, Printer] = LMDBStore.create[Channel, Pattern, Entry, Printer](storePath2, 1024L * 1024L * 100L)
+    val space2 = new RSpace[Channel, Pattern, Entry, Printer](store2, coop.rchain.rspace.history.Branch.MASTER)
+
+    val cres =
+      space2.consume(List(Channel("friends")),
+                    List(CityMatch(city = "Crystal Lake")),
+                    new Printer,
+                    persist = false)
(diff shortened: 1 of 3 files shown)


rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```
+
+To rollback the state of the RSpace to a given checkpoint one simply calls the `reset` method with the hash of the root of the checkpoint provided as parameter.
+```scala
+space.reset(checkpointHash)
+```
+
+Let's see how this works in practice. We'll start by creating a new, untouched RSpace followed by a consume operation which should put data and a continuation at given channel.
+```tut
+    val storePath2: Path = Files.createTempDirectory("rspace-address-book-example-")
```
henrytill: Don't indent like this.  It makes the tut-generated REPL examples look pretty weird.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
 println(space.store.toMap)
 ```
 
+### History & rollback
+
+It is possible to save the current state of RSpace in the form of a checkpoint (root hash of a given Merkle Patricia Trie).
```
henrytill: How about something like:

It is possible to save the current state of RSpace in the form of a `Checkpoint`.  A `Checkpoint` value contains the root hash of a Merkle Patricia Trie built from the contents of RSpace.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```scala
+val checkpoint = space.createCheckpoint()
+val checkpointHash = checkpoint.root
+```
+
+To rollback the state of the RSpace to a given checkpoint one simply calls the `reset` method with the hash of the root of the checkpoint provided as parameter.
```
henrytill: Similar notes about using the proper name `Checkpoint`.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+    val cres =
+      space2.consume(List(Channel("friends")),
+                    List(CityMatch(city = "Crystal Lake")),
+                    new Printer,
+                    persist = false)
+    assert(cres.isEmpty)
+
+```
+
+We can now create a checkpoint and store it's root
```
henrytill: Add a period to the end of this sentence.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```tut
+    val checkpointHash = space2.createCheckpoint.root
+```
+
+The first `produceAlice` operation should be able to find data stored by the consume, hence:
```
henrytill: replace ", hence:" with a period.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```tut
+    def produceAlice = space2.produce(Channel("friends"), alice, persist = false)
+    assert(produceAlice.isDefined)
+```
+
+Running the same operation again shouldn't return anything, as data hasn't been persisted:
```
henrytill: End with a period, not a colon.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```tut
+    assert(produceAlice.isEmpty)
+```
+Every following repetition of the operation above should yield an empty result.
+```tut
+    assert(produceAlice.isEmpty)
+```
+
+After re-setting the RSpace to the state from the saved checkpoint the first produce operation should again return an non-empty result:
```
henrytill: End with a period, not a colon.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```tut
+    space2.reset(checkpointHash)
+    assert(produceAlice.isDefined)
+```
+And again, every following operation should yield an empty result
```
henrytill: End with a period.

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+```
+
+We can now create a checkpoint and store it's root
+```tut
+    val checkpointHash = space2.createCheckpoint.root
+```
+
+The first `produceAlice` operation should be able to find data stored by the consume, hence:
+```tut
+    def produceAlice = space2.produce(Channel("friends"), alice, persist = false)
```
henrytill: This is an effectful operation, so we should use empty parentheses.  

Apologies to Bertrand Meyer and the _Uniform Access Principle_...

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+    assert(produceAlice.isDefined)
```
henrytill: I think it would result in easier to read examples if we didn't use `assert`s here, so that the tut-generated examples showed the output.  

I understand that this approach reinforces our tests, but if we are relying on the tutorial to test how our code works, then we have a problem...

rspace/src/main/tut/Tutorial.md
```diff
@@ -408,9 +408,69 @@ val pres14 = space.produce(Channel("friends"), erin, persist = true)
...
+To rollback the state of the RSpace to a given checkpoint one simply calls the `reset` method with the hash of the root of the checkpoint provided as parameter.
+```scala
+space.reset(checkpointHash)
+```
+
+Let's see how this works in practice. We'll start by creating a new, untouched RSpace followed by a consume operation which should put data and a continuation at given channel.
+```tut
+    val storePath2: Path = Files.createTempDirectory("rspace-address-book-example-")
+    val store2: LMDBStore[Channel, Pattern, Entry, Printer] = LMDBStore.create[Channel, Pattern, Entry, Printer](storePath2, 1024L * 1024L * 100L)
+    val space2 = new RSpace[Channel, Pattern, Entry, Printer](store2, coop.rchain.rspace.history.Branch.MASTER)
```
henrytill: There's something aesthetically less pleasing about using the `2` suffix in a tutorial.  Perhaps we could use a more descriptive name?

rspace/src/main/tut/Tutorial.md
lukasz-golebiewski: Makes sense. I'll add explicit return type as well

rspace/src/main/tut/Tutorial.md
lukasz-golebiewski: Agreed. I would leave the asserts in the example though

rspace/src/main/tut/Tutorial.md
lukasz-golebiewski: Done

PR #1013, opened 2018-06-22
CORE-683 stable hashing of stored collections by sorting

ys-pyrofex:
## Overview
We need stable hashing of stored collections, otherwise checkpoints with same data will have different root hashes.

https://rchain.atlassian.net/browse/CORE-683


crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b256.scala
```diff
@@ -27,4 +31,15 @@ object Blake2b256 {
     }
   }
 
+  def hash(inputs: Seq[ByteVector]): Array[Byte] = {
+    val outStream = new DigestOutputStream(new Blake2bDigest(256))
+    synchronized {
```
lukasz-golebiewski: Why are you synchronizing here?

ys-pyrofex: That's one of question I have too. Usually hash functions shouldn't depend from any shared memory regions, and therefore don't need any synchronization. However, the original code with synchronized { } region was written by Eitan and passed reviews. Therefore I assumed that I don't know something about actual implementation or JVM, and copied that synchronization region.

lukasz-golebiewski: It's not needed. It can be removed from the other function as well.

henrytill: I believe that the `synchronized` in the original code was mistakenly copied from an upstream source where it was protecting a mutable variable.  I agree with @lukasz-golebiewski that it shouldn't be necessary here.

ys-pyrofex: That's good. Removed both.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b256.scala
```diff
@@ -19,12 +23,19 @@ object Blake2b256 {
...
+    digestFn.update(input, 0, input.length)
+    val res = new Array[Byte](32)
+    digestFn.doFinal(res, 0)
+    res
   }
 
+  def hash(inputs: Seq[ByteVector]): Array[Byte] = {
+    val outStream = new DigestOutputStream(new Blake2bDigest(256))
+    for (input <- inputs) {
+      input.copyToStream(outStream)
```
henrytill: 👍

rspace/src/main/scala/coop/rchain/rspace/util/package.scala
```diff
@@ -55,4 +55,24 @@ package object util {
...
     )
   }
+
+  def veccmp(a: ByteVector, b: ByteVector): Int = {
+    val c = a.length - b.length
+    if (c != 0) {
+      c.toInt
+    } else {
+      for (i <- 0L until a.length) {
+        //indexed access of two ByteVectors can be not fast enough,
```
henrytill: Curious - why is this?  What is "fast enough"?

henrytill:
It's also necessary that, in a join, a pattern always remain associated with the proper channel.  

We need to consider this with any channel sorting strategy we are using.  Your current strategy avoids this, but I'm not sure that it will be sufficient to ensure stable hashes in the Trie or the Replay logs.

rspace/src/main/scala/coop/rchain/rspace/util/package.scala
ys-pyrofex: "Fast enough" here just note about possible place of unreasonably expensive computations for such simple operation. ByteVector can contain multiple chunks and indexed access may require additional computations on every call: "index in vector -> chunk + index in chunk". Ideal comparer in such situations should be implemented inside ByteVector class and compare chunks faster. Absolutely perfect comparer may also use AVX or SSE ops.

henrytill:
@ys-pyrofex: ready for re-review?

ys-pyrofex:
I don't know what to add to this commit, so yes, time for re-review.

henrytill:
@ys-pyrofex: you have a conflict

ys-pyrofex:
Updated to latest dev

rspace/src/main/scala/coop/rchain/rspace/StableHashProvider.scala
```diff
@@ -0,0 +1,42 @@
...
+      channels
+        .map(c => codecC.encode(c).get.bytes)
+        .sorted(util.ordByteVector))
+
+  def hash[C, P, K](channels: Seq[C], patterns: Seq[P], continuation: K, persist: Boolean)(
+      implicit
+      serializeC: Serialize[C],
+      serializeP: Serialize[P],
+      serializeK: Serialize[K]) = {
+    val (channelsSeq, patternsSeq) = channels
```
henrytill: how about `encodedChannels` and `encodedPatterns`?

rspace/src/main/scala/coop/rchain/rspace/StableHashProvider.scala
```diff
@@ -0,0 +1,42 @@
...
+      .zip(patterns)
+      .map { case (channel, pattern) => (serializeC.encode(channel), serializeP.encode(pattern)) }
+      .sorted(util.ordByteVectorPair)
+      .unzip
+
+    Blake2b256Hash.create(
+      channelsSeq ++ patternsSeq
+        :+ serializeK.encode(continuation)
```
henrytill: @birchmd pointed out to me that the default `immutable.Seq` is a `List`, so these single-element appends are expensive.

probably better to put them into a two element `Seq` and append that to the others.

rspace/src/main/scala/coop/rchain/rspace/StableHashProvider.scala
```diff
@@ -0,0 +1,42 @@
...
+        :+ (ignore(7) ~> bool).encode(persist).get.toByteVector)
+  }
+
+  def hash[C, A](channel: C, datum: A, persist: Boolean)(implicit
+                                                         serializeC: Serialize[C],
+                                                         serializeA: Serialize[A]) =
+    Blake2b256Hash.create(
+      Seq(serializeC.encode(channel),
+          serializeA.encode(datum),
+          (ignore(7) ~> bool).encode(persist).get.toByteVector))
```
henrytill: nitpick: I prefer

```scala
(ignore(7) ~> bool).encode(persist).map(_.bytes).get
```

rspace/src/test/scala/coop/rchain/rspace/trace/EventTests.scala
```diff
@@ -48,18 +41,34 @@ class EventTests extends FlatSpec with Matchers with GeneratorDrivenPropertyChec
...
 
-      val expectedHash =
-        List(Codec[Seq[String]].encode(channels),
-             Codec[Seq[Pattern]].encode(patterns),
-             Codec[StringsCaptor].encode(continuation),
-             (cignore(7) ~> bool).encode(persist))
-          .sequence[Attempt, BitVector]
-          .map { (vectors: List[BitVector]) =>
-            Blake2b256Hash.create(vectors.toArray.flatMap(_.toByteArray))
+      val (channelsSeq, patternsSeq) =
```
henrytill: See above,

rspace/src/test/scala/coop/rchain/rspace/trace/EventTests.scala
```diff
@@ -48,18 +41,34 @@ class EventTests extends FlatSpec with Matchers with GeneratorDrivenPropertyChec
...
+              (Serialize[String].encode(channel), Serialize[Pattern].encode(pattern))
           }
-          .get
+          .sorted(util.ordByteVectorPair)
+          .unzip
+
+      val expectedHash =
+        Blake2b256Hash.create(
+          channelsSeq ++ patternsSeq
+            :+ Serialize[StringsCaptor].encode(continuation)
```
henrytill: See above.

rspace/src/test/scala/coop/rchain/rspace/trace/EventTests.scala
```diff
@@ -48,18 +41,34 @@ class EventTests extends FlatSpec with Matchers with GeneratorDrivenPropertyChec
...
+            :+ (cignore(7) ~> bool).encode(persist).get.toByteVector)
```
henrytill: See above.

PR #1014, opened 2018-06-22
Add timestamp validation to casper blocks

KentShikama:
## Overview

### Notes


KentShikama:
@birchmd PTAL

casper/src/test/scala/coop/rchain/casper/helper/BlockGenerator.scala
```diff
@@ -1,22 +1,40 @@
...
-import coop.rchain.catscontrib._
-import Catscontrib._
-import cats._
-import cats.data._
-import cats.implicits._
-import coop.rchain.casper.Estimator.{BlockHash, Validator}
-import coop.rchain.casper.protocol._
+
+object BlockGenerator {
+  implicit def idCapture: Capture[Id] = new Capture[Id] {
```
goral09: There's a `Capture[Id]` instance in the `catscontrib` right?

KentShikama: As I posted on the #scala channel, I got the following error when I tried to import that

```
26: ambiguous implicit values:
[error]  both method idCapture in trait CaptureInstances of type => coop.rchain.catscontrib.Capture[cats.Id]
[error]  and method idCapture in trait CaptureInstances of type => coop.rchain.catscontrib.Capture[cats.Id]
[error]  match expected type coop.rchain.catscontrib.Capture[cats.Id]
[error]   implicit val timeEff = new LogicalTime[Id]
```

birchmd: I think if you `import coop.rchain.catscontrib._` then scala figures it out properly.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -87,6 +89,40 @@ object Validate {
...
   }
 
+  // This is not a slashable offence
+  def timestamp[F[_]: Monad: Log: Time](
+      b: BlockMessage,
+      dag: BlockDag): F[Either[RejectableBlock, IncludeableBlock]] =
+    for {
+      currentTime  <- Time[F].currentMillis
+      timestamp    = b.header.get.timestamp
+      beforeFuture = currentTime >= timestamp
```
birchmd: I think this should be `currentTime + drift >= timestamp`, where `drift` is a constant which represents the amount of time another clock could be ahead of yours. We should probably make this configurable, but for now can probably just be a constant in the class. 

We should also probably set a minimum time that we accept; something like `recent = timestamp >= currentTime - delay`, where `delay` is another constant which represents how long we expect a message takes to reach us, i.e. an explicit synchrony constraint.

KentShikama: I'm worried that this allows attacks to intentionally set their block timestamps to be in the future and then that prevents the next validator in the round robin from creating a valid timestamp that is after that future timestamp before their allotted time is up.

KentShikama: Did you mean

`recent = timestamp <= currentTime - delay` ?

KentShikama: Oh wow geth does this

```
allowedFutureBlockTime = 15 * time.Second
...
if header.Time.Cmp(big.NewInt(time.Now().Add(allowedFutureBlockTime).Unix())) > 0 {
			return consensus.ErrFutureBlock
		}
	}
	if header.Time.Cmp(parent.Time) <= 0 {
		return errZeroBlockTime
}
```
from https://github.com/ethereum/go-ethereum/blob/d429a92f09e476c431830cef48690dc3784153c7/consensus/ethash/consensus.go#L234

Essentially strictly checks that block timestamp is before parent timestamp and for checking for future blocks it adds a `drift`.

I'm still a bit worried though; ethereum PoW doesn't suffer the same round robin attack I've outlined above.

KentShikama: I guess we can allow a drift that is half the allotted time for round robin?

casper/src/test/scala/coop/rchain/casper/helper/BlockGenerator.scala
KentShikama: @birchmd Yup that did the trick!

KentShikama:
@birchmd PTAL

PR #1017, opened 2018-06-22
CORE-661: Improve Serialize type class

ys-pyrofex:
## Overview
The existing API isn't well suited to composability. You would have to copy out pieces to de-serialize the sub-parts. ByteVector has fast and rich API with views, slices, etc.

https://rchain.atlassian.net/browse/CORE-661


goral09:
I am not sure our general-purpose serialization type class should depend on the 3rd party library. If it's so easy to go from one to the other why can't the `Array[Byte]` be turned into `ByteVector` if necessary and leave other places as-is.

PR #1021, opened 2018-06-22
Fix RHOL-442 bug

goral09:
## Overview
Fixes a bug when binding a list to a list remainder was not working properly.

Using only a list remainder as a binding pattern should set the `connectiveUsed` flag to `true`. If `connectiveUsed = false` then in the matcher target and pattern are compared using scala equality.

During this bug fix I've simplified some parts of the `normalize.scala` : 

* get rid of explicit `asScala` conversions in favor of implicit
* get rid of redundant `case _ => ` matches when there's only one type possible

[JIRA issue](https://rchain.atlassian.net/browse/RHOL-442).

### Notes

It's best to view the PR in the split diff version.

### DIFF
--- a/models/src/main/scala/coop/rchain/models/rholang/implicits.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/implicits.scala
@@ -404,7 +404,9 @@ object implicits {
   implicit val ReceiveBindLocallyFree: HasLocallyFree[ReceiveBind] =
     new HasLocallyFree[ReceiveBind] {
       def connectiveUsed(rb: ReceiveBind) =
-        ChannelLocallyFree.connectiveUsed(rb.source.get)
+        ChannelLocallyFree.connectiveUsed(rb.source.get) || rb.patterns.exists(
+          ChannelLocallyFree.connectiveUsed(_))
+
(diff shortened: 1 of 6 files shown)


models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -404,7 +404,9 @@ object implicits {
   implicit val ReceiveBindLocallyFree: HasLocallyFree[ReceiveBind] =
     new HasLocallyFree[ReceiveBind] {
       def connectiveUsed(rb: ReceiveBind) =
-        ChannelLocallyFree.connectiveUsed(rb.source.get)
+        ChannelLocallyFree.connectiveUsed(rb.source.get) || rb.patterns.exists(
+          ChannelLocallyFree.connectiveUsed(_))
```
goral09: I am not yet sure about this change, reverting to previous version does not make tests fail but newer versions seems correct.

guardbotmk3: This isn't correct. The patterns may have connectives without the ReceiveBind being a pattern itself.

PR #1022, opened 2018-06-22
Enable Casper to sign blocks

birchmd:
## Overview
Casper will keep the private key of the validator in memory. This allows it to track seq number correctly, and propose blocks on their behalf (e.g. when responding to a slashing condition). In the future we will not pass the private key to the node directly, but instead read from an encrypted file.

https://rchain.atlassian.net/browse/RHOL-448

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -526,6 +526,8 @@ sealed abstract class MultiParentCasperInstances {
                 acc.updated(p, currChildren + hash)
             }
 
+            val newSeqNum = bd.currentSeqNum.updated(block.sender, block.seqNum)
+
             bd.copy(
               blockLookup = bd.blockLookup.updated(hash, block),
               //Assume that a non-equivocating validator must include
@@ -538,7 +540,8 @@ sealed abstract class MultiParentCasperInstances {
               latestMessagesOfLatestMessages =
                 bd.latestMessagesOfLatestMessages.updated(block.sender,
                                                           toLatestMessages(block.justifications)),
-              childMap = newChildMap
+              childMap = newChildMap,
+              currentSeqNum = newSeqNum
             )
           })
         }
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -3,6 +3,8 @@ package coop.rchain.casper
 import cats.{Applicative, Monad}
 import cats.implicits._
 import com.google.protobuf.ByteString
+import coop.rchain.catscontrib.TaskContrib._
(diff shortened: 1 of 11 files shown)


casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -69,6 +74,9 @@ sealed abstract class MultiParentCasperInstances {
       F[_]: Monad: Capture: NodeDiscovery: TransportLayer: Log: Time: ErrorHandler: SafetyOracle](
       storageLocation: Path,
       storageSize: Long,
+      pk: Array[Byte],
```
goral09: `pk` stands for `publicKey` ? it could also be `privateKey`. I know, existence of `sk` below makes for a good heuristic but could you use full name?

goral09: If it's standard then I'm fine. Don't want to be to trying. Let's leave it this way.

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -554,4 +571,29 @@ sealed abstract class MultiParentCasperInstances {
         } yield ()
       }
     }
+
+  def fromConfig[
+      F[_]: Monad: Capture: NodeDiscovery: TransportLayer: Log: Time: ErrorHandler: SafetyOracle](
+      conf: CasperConf)(implicit scheduler: Scheduler, taskLog: Log[Task]): MultiParentCasper[F] = {
+    val genesis = Genesis
+      .fromBondsFile[Task](conf.bondsFile, conf.numValidators, conf.validatorsPath)
+      .unsafeRunSync
```
KentShikama: Hmm throwing in an explicit Monad `Task` and calling `unsafeRunSync` feels against the paradigm. But lets deal with this latter.

casper/src/main/scala/coop/rchain/casper/Casper.scala
KentShikama: I think if @goral09  didn't immediately understand it then we should use "publicKey" and "secretKey". But I don't think it is a blocker either.

casper/src/main/scala/coop/rchain/casper/Casper.scala
birchmd: Yes, you're probably right. Although I'm not sure what else to do about it because I need an actual MultiParentCasper instance at the end, not wrapped in some other effect. So if I used and `F` in this function as we normally do then I would somehow need to unwrap it afterwards. I guess maybe I need to think about using a `Comonad`...

goral09: @birchmd couldn't you `map` over the `Genesis.fromBondsFile[F]` returned value?

PR #1025, opened 2018-06-25
 Improve the TCP transport layer witch Monix Task

sebtomba:
## Overview
Improve the TCP transport layer witch Monix Task. This PR gets rid of the `Await` and implements a timeout in terms of `Task.timeout`. Implementation of a _send and forget_ semantic with `Task.racePair`.

https://rchain.atlassian.net/browse/CORE-678


comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -105,28 +116,25 @@ class TcpTransportLayer[
...
                 })
-                .pure[F]
+                .pure[Task]
     } yield pmErr
 
-  val local: F[PeerNode] = src.pure[F]
+  // TODO: rename to sendAndForget
+  def send(peer: PeerNode, msg: ProtocolMessage): Task[CommErr[Unit]] =
+    Task
+      .racePair(innerSend(peer, TLRequest(msg.proto.some)), Task.unit)
```
goral09: As per my understanding of `Task.racePair` it doesn't make sens to race `Task.unit` . Why did you choose to do that?

sebtomba: Please distinguish between `Task.race` and `Task.racePair`. `Task.race` cancells the losing `Task`.

goral09: Won't `innerSend(…).attempt` achieve the same result? and `Task.timeout` is [not implemented with racing a `Task.unit`](https://github.com/monix/monix/blob/master/monix-eval/shared/src/main/scala/monix/eval/Task.scala#L1803).

rabbitonweb:
@sebtomba I've talked with @birchmd and they are cool with going full `fire and forget` for the `send` in `TransportLayer`. This means that both `send` and `broadcast` will now return `F[Unit]` (not `F[CommErr[Unit]]`

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
sebtomba: No, `Task.attempt` turns a `Task[A]` into `Task[Either[Throwable, A]]`. Such a `Task` can't result in an `Error`.
**Send and forget** means that we are scheduling a message for sending and don't wait for the execution nor the response. We don't care if the message actually gets sent, delivered and handled by the remote peer. However, there are some edge cases, f.e. we need to provide a grace period (not included in this PR) on shutdown to deliver the `Disconnect` message to our peers.

sebtomba:
@birchmd Please review my changes to Casper

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -51,51 +50,63 @@ class TcpTransportLayer[
...
     for {
-      connections <- ConnectionsState[F].get
-      _ = connections
-        .get(remote.id)
-        .foreach(c => Try(c.getChannel.asInstanceOf[ManagedChannel].shutdown()))
-      _ <- ConnectionsState[F].modify(_ - remote.id)
+      cs <- connections.get
+      _ = cs
+        .get(peer.id)
+        .foreach(c => Try(c.shutdown()))
```
rabbitonweb: foreach replace with fold
Try replace with Task

sebtomba: Replaced with `match`. Is IMO more readable in this case.

PR #1031, opened 2018-06-25
Casper takes Runtime as an argument

birchmd:
## Overview
First step towards enabling metrics for casper

### Notes

### DIFF
--- a/node/src/main/scala/coop/rchain/node/node.scala
+++ b/node/src/main/scala/coop/rchain/node/node.scala
@@ -134,7 +134,9 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
   private val address                  = s"rnode://$name@$host:$port"
   private val src                      = PeerNode.parse(address).right.get
   private val storagePath              = conf.run.data_dir().resolve("rspace")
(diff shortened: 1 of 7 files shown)


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -134,7 +134,9 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
   private val address                  = s"rnode://$name@$host:$port"
   private val src                      = PeerNode.parse(address).right.get
   private val storagePath              = conf.run.data_dir().resolve("rspace")
+  private val casperStoragePath        = storagePath.resolve("casper")
   private val storageSize              = conf.run.map_size()
+  private val casperRuntime            = Runtime.create(casperStoragePath, storageSize)
```
henrytill: I think it's probably better if we are doing this in `acquireResources` and putting the Casper `Runtime` into the `Resources` along with the REPL `Runtime`.  

I recognize that this has implications for how you instantiate Casper, but it seems problematic that resources would be acquired and accounted for out of band like this, especially when we should have a unified approach for both acquiring them and closing them.

PR #1039, opened 2018-06-26
[Rhol-265] Implement 'Matches Expression'

odeac:
## Overview
Implement matches expressions in Rholang

https://rchain.atlassian.net/browse/RHOL-265

TODO: 
* add normaliser tests that test the behaviour regarding locallyFree and connectiveUsed


models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -316,6 +320,8 @@ object implicits {
         case EAndBody(EAnd(p1, p2))     => p1.get.connectiveUsed || p2.get.connectiveUsed
         case EOrBody(EOr(p1, p2))       => p1.get.connectiveUsed || p2.get.connectiveUsed
         case EMethodBody(e)             => e.connectiveUsed
+        case EMatchesBody(EMatches(process, pattern)) =>
+          process.get.connectiveUsed || pattern.get.connectiveUsed
```
goral09: You shouldn't have to call `.get` anymore. There're probably couple residues. There must be an implicit that goes from `Par => Option[Par]` that makes compiler accept those constructs.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -487,6 +488,19 @@ object Reduce {
             b1 <- evalToBool(p1.get)
             b2 <- evalToBool(p2.get)
           } yield GBool(b1 || b2)
+
+        case EMatchesBody(EMatches(target, pattern)) =>
+          for {
+            evaledTarget <- evalExpr(target.get)
```
goral09: Here as well, I think those `.get`s are not necessary.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -894,4 +894,50 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+      val reducer      = RholangOnlyDispatcher.create[Task, Task.Par](space).reducer
+
+      val inspectTask =  reducer.evalExpr(EMatches(GInt(1), EVar(Wildcard(Var.WildcardMsg()))))
+
+      Await.result(inspectTask.runAsync, 3.seconds)
+    }
+
+    result.exprs should be (Seq(Expr(GBool(true))))
+    errorLog.readAndClearErrorVector should be(Vector.empty[InterpreterError])
+  }
```
goral09: You should also test for `1 matches x` where `x` is some bound variable.

rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -43,7 +43,7 @@ PLt.          Proc7  ::= Proc7 "<" Proc8 ;
 PLte.         Proc7  ::= Proc7 "<=" Proc8 ;
 PGt.          Proc7  ::= Proc7 ">" Proc8 ;
 PGte.         Proc7  ::= Proc7 ">=" Proc8 ;
-PMatches.     Proc6  ::= Proc7 "matches" Proc7 ;
+PMatches.     Proc6  ::= Proc7 "matches" ProcVar ;
```
guardbotmk3: This is wrong. Why did you change this?

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -424,8 +424,9 @@ object ProcNormalizeMatcher {
       case p: PEq  => binaryExp(p.proc_1, p.proc_2, input, EEq.apply)
       case p: PNeq => binaryExp(p.proc_1, p.proc_2, input, ENeq.apply)
 
-      case p: PAnd => binaryExp(p.proc_1, p.proc_2, input, EAnd.apply)
-      case p: POr  => binaryExp(p.proc_1, p.proc_2, input, EOr.apply)
+      case p: PAnd     => binaryExp(p.proc_1, p.proc_2, input, EAnd.apply)
+      case p: POr      => binaryExp(p.proc_1, p.proc_2, input, EOr.apply)
+      case p: PMatches => binaryExp(p.proc_1, p.proc_2, input, EMatches.apply)
```
guardbotmk3: This doesn't even match the grammar change that you made.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
guardbotmk3: Do you mean `1 matches =x` where `x` is a variable bound to `1`

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -894,4 +894,50 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+      val reducer      = RholangOnlyDispatcher.create[Task, Task.Par](space).reducer
+
+      val inspectTask =  reducer.evalExpr(EMatches(GInt(1), GInt(0)))
+
+      Await.result(inspectTask.runAsync, 3.seconds)
+    }
+
+    result.exprs should be (Seq(Expr(GBool(false))))
+    errorLog.readAndClearErrorVector should be(Vector.empty[InterpreterError])
+  }
```
guardbotmk3: These 2 examples are kind of trivial. It would be nice to see a more complicated example.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
goral09: Yes, my mistake.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: The matching functionality is already tested with the tests for the `match-case` construction. 

What I'm testing here is that the matching algorithm is called when a `matches` expression is reduced.

Since I haven't done any changes to the matching functionality, adding more ReduceSpec tests for the `matches` operator doesn't improve the test coverage.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: Please see 84d3b59b

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: The examples in 84d3b59 should cover the usage of BoundevVars and Connectives

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
odeac: done

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -309,6 +309,8 @@ object Substitute {
             s2(par1.get, par2.get)(EAnd(_, _))
           case EOrBody(EOr(par1, par2)) =>
             s2(par1.get, par2.get)(EOr(_, _))
+          case EMatchesBody(EMatches(target, pattern)) =>
+            s2(target.get, pattern.get)(EMatches(_, _))
```
goral09: you don't need those `.get`s.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -316,7 +320,9 @@ object implicits {
         case EAndBody(EAnd(p1, p2))     => p1.get.connectiveUsed || p2.get.connectiveUsed
         case EOrBody(EOr(p1, p2))       => p1.get.connectiveUsed || p2.get.connectiveUsed
         case EMethodBody(e)             => e.connectiveUsed
-        case ExprInstance.Empty         => false
+        case EMatchesBody(EMatches(process, pattern)) =>
+          process.connectiveUsed || pattern.connectiveUsed
```
guardbotmk3: No, this is still wrong.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -347,7 +353,9 @@ object implicits {
         case EAndBody(EAnd(p1, p2))     => p1.get.locallyFree | p2.get.locallyFree
         case EOrBody(EOr(p1, p2))       => p1.get.locallyFree | p2.get.locallyFree
         case EMethodBody(e)             => e.locallyFree
-        case ExprInstance.Empty         => BitSet()
+        case EMatchesBody(EMatches(target, pattern)) =>
+          target.locallyFree | pattern.locallyFree
```
guardbotmk3: Same here.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
goral09: So the `connectiveUsed` and `locallyFree` should be only calculated from the `process` (target) and not the `pattern` as well?

odeac: what would be a rholang test that will fail because of this?

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
odeac: what would be a rholang test that will fail because of this?

PR #1050, opened 2018-06-27
No ticket: Fixes variable reference substitution

goral09:
## Overview
When variable reference was referencing a variable in the receive pattern the substitution failed.

As part of the ticket I've also added `PrettyPrinter` cases for logical connectives although I wasn't able to write proper case for `VarRef`. I didn't want to postpone pushing the bugfix because of that.

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
@@ -3,6 +3,7 @@ package coop.rchain.rholang.interpreter
 import cats.implicits._
 import cats.{Applicative, Monad}
 import coop.rchain.models.Channel.ChannelInstance._
+import coop.rchain.models.Connective.ConnectiveInstance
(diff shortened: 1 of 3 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -77,7 +78,7 @@ object Substitute {
       env.get(term.index) match {
         case Some(par) => Applicative[M].pure(Right(par))
         case None =>
-          interpreterErrorM[M].raiseError(SubstituteError(s"Illegal VarRef [$term]"))
+          Applicative[M].pure(Left(term))
```
guardbotmk3: Good catch. At the time it seemed to make perfect sense that they would all be available right now, but of course some of them could not be available yet.

goral09: I should have caught that in the review.

PR #1051, opened 2018-06-28
Implement a safe broadcast in the transport layer

sebtomba:
## Overview
Implement a safe broadcast in the transport layer. The safe roundtrip waits for the response of the remote peers and returns if applicable an error. However, one failed message doesn't fail the whole broadcast.

https://rchain.atlassian.net/browse/CORE-751

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
@@ -9,26 +9,28 @@ import coop.rchain.casper.protocol._
 import coop.rchain.comm.PeerNode
 import coop.rchain.comm.protocol.rchain.Packet
 import coop.rchain.crypto.codec.Base16
-import coop.rchain.p2p.effects._
 import coop.rchain.comm.transport._, CommMessages._
 import coop.rchain.comm.discovery._
 import coop.rchain.comm.CommError.ErrorHandler
 import coop.rchain.shared._
 
 import scala.util.Try
+import scala.concurrent.duration._
 
 object CommUtil {
 
   private implicit val logSource: LogSource = LogSource(this.getClass)
+
+  private val defaultTimeout = 1000.milliseconds
 
   def sendBlock[F[_]: Monad: NodeDiscovery: TransportLayer: Log: Time: ErrorHandler](
       b: BlockMessage): F[Unit] = {
     val serializedBlock = b.toByteString
     val hashString      = PrettyPrinter.buildString(b.blockHash)
     for {
-      _ <- Log[F].info(s"CASPER: Beginning send of ${PrettyPrinter.buildString(b)} to peers...")
-      _ <- sendToPeers[F](serializedBlock)
-      _ <- Log[F].info(s"CASPER: Sent $hashString to peers")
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -9,26 +9,28 @@ import coop.rchain.casper.protocol._
...
 import coop.rchain.shared._
 
 import scala.util.Try
+import scala.concurrent.duration._
 
 object CommUtil {
 
   private implicit val logSource: LogSource = LogSource(this.getClass)
 
+  private val defaultTimeout = 1000.milliseconds
```
birchmd: I'm not a fan of this hardcoded timeout, would definitely prefer it to be configurable. And perhaps it makes more sense for it to belong in `TransportLayer` itself and have it be overridable at the call-site

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -77,12 +88,15 @@ object CommUtil {
             local    <- TransportLayer[F].local
             block    = dag.blockLookup.get(r.hash).map(_.toByteString)
             maybeMsg = block.map(serializedMessage => packet(local, serializedMessage))
-            send     <- maybeMsg.traverse(msg => TransportLayer[F].send(peer, msg))
+            send     <- maybeMsg.traverse(msg => TransportLayer[F].roundTrip(peer, msg, defaultTimeout))
```
birchmd: This seems like a strange implementation to me. The communication originates with the other node sending the request, so I would expect the `roundTrip` to be called there and the response to be given directly here, as opposed to starting another `roundTrip`. But maybe it makes sense because then both sides get to know their messages were received?

sebtomba: @birchmd: Do you want to keep the *send and forget* semantic here?
@rabbitonweb ?

birchmd: The logic is as follows. Suppose we have two nodes A and B; A sends a BlockRequest message to B; B recieves the message and checks if it has the block, if so then sends the block to A. Now, what this should look like in terms of communications functions I'm not sure. I could see this as a single `roundTrip` initiated by A, where the block from B comes as a response. Or, as it is right now, with two `roundTrips` where two additional messages are sent, which are effectively just acknowledgements.

I suppose having the extra two acknowledgements isn't a problem really, so maybe it's fine as is.

sebtomba: Let's keep the message passing style with acknowledgements.

goral09: @sebtomba
> I suppose having the extra two acknowledgements isn't a problem really, so maybe it's fine as is.

in the scenario @birchmd  described, what happens when block A don't send the `ack` in time, does B initiates another `roundTrip` ?

sebtomba:
We need a more holistic approach to message passing between nodes on the transport layer. This needs further discussion. Closing for now.

PR #1054, opened 2018-06-28
Core 720 heartbeat

rabbitonweb:
## Overview
Best to follow this PR commit by commit.

Originally this PR was suppose to be about providing heartbeat. Eventually I:

1. upgraded Kademlia's `table.observe` to `Ping` remotes each time `updateLastSeen` is called
2. fixed Kademlia issues (when adding a node that already extis in a tabkle, routing information was not updated, peer was not `Ping`ed

After this PR is merged, when two nodes are connected to each other we have 100% guarantee  that they can exchange messages with each other (both sending and receiving). 

Also: `NotHandled` was given `CommError` to propagate. `InternalServerError` response pushes that back to the caller via `String` - this has to be handled better in a future.

During the work on this issue, some additional Kademlia issues were spotted, like:

1. Kademlia.observe - consider returning an result of action taken (added, updated, older update instead)
2. Potential bug: we ignore ping request to update the kademlai table
3. Potenatial BUG: when disconnecting from network other nodes (even though they've learned about disconnect) continue on sending PING messages

Overall we need to focus on writing unit tests for Kademlia.

CORE-720

### DIFF
--- a/comm/src/main/protobuf/coop/rchain/comm/protocol/routing.proto
+++ b/comm/src/main/protobuf/coop/rchain/comm/protocol/routing.proto
@@ -62,6 +62,7 @@ message TLRequest {
 }
 
 message InternalServerError {
+  bytes error = 1;
 }
 
 message NoResponse {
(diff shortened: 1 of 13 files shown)


rabbitonweb:
special thanks go to @dzajkowski for helping me out with this issue (second set of eyes and one extra brain) :)

PR #1056, opened 2018-06-28
Rspace: Add parameter type for the result of match.

guardbotmk3:
Note that the result match type does not need a Serialize instance, because it
only ever occurs post-deserialize. While not strictly necessary for passing the
random state along with the results, It is conceptually cleaner to separate the
match input type from the match output type.


rspace/src/main/scala/coop/rchain/rspace/ISpace.scala
```diff
@@ -13,9 +13,10 @@ import scala.collection.immutable.Seq
   * @tparam C a type representing a channel
   * @tparam P a type representing a pattern
   * @tparam A a type representing an arbitrary piece of data
+  * @tparam A a type representing a match result
```
henrytill: `@tparam R`

goral09: Can match result be different type that the data?

PR #1064, opened 2018-06-29
Remove Capture in favor of Sync.

goral09:
## Overview
We shouldn't be reinventing the wheel. `Capture` type class provides same functionalities as `cats.Sync` and we had to shim it (`Capture`) to `monix` and `cats`.

None.

### Notes


rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -38,21 +39,19 @@ class RholangOnlyDispatcher[M[_]] private (_reducer: => Reduce[M])(implicit capt
         val env = Dispatch.buildEnv(dataList)
         reducer.eval(par)(env)
       case ScalaBodyRef(_) =>
-        captureM.apply(())
+        Sync[M].unit
```
henrytill: I really don't like this approach of using context bounds sugar and then using some implicit summoner to get a handle on the type class dictionary.  If you didn't use the sugar in the first place, you'd already have the handle in the form of the implicit parameter.  

I've mentioned this before [here](https://rchain.atlassian.net/browse/CORE-455?focusedCommentId=10763&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-10763).

In general, context bounds just seems like more playacting.  An attempt is made to imitate Haskell's syntax and all of its conveniences and instead we get something that is just kinda weird and shitty, and doesn't scale well at all.  

I think that the visibly using implicit parameters in Scala is a cleaner and clearer expression the underlying notion of type class dictionary passing than attempting to engage in this kind of theater.  And it scales better.

henrytill:
> We should be reinventing the wheel.

🤔

goral09:
> >We should be reinventing the wheel.

> :thinking: 

typo of course.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
goral09: Thanks for the comment and opinion on that.

I don't know anything about scaling problems with this approach. AFAIK `scalac` will desugar  it to regular implicit parameters,  and it no longer suffers from the problem where it couldn't have both context bounds and implicit parameters. 

I didn't want to imitate Haskell's syntax or anything like that, I just find calling `Sync[F].delay(…)` cleaner than `s.delay(…)` because `s` doesn't say anything and I have to scroll up to check. 

In general I don't really care so I will listen to your advise and fix the code.

goral09: One more thing,  when the implicit argument is being plumbed through various levels of code to be used somewhere deeper I find it cleaner to use context bounds than implicit parameters. It feels more natural because I won't have to _summon_ instance anyway.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -37,8 +38,8 @@ object Substitute {
 
   def apply[M[_], A](implicit ev: Substitute[M, A]): Substitute[M, A] = ev
 
-  def maybeSubstitute[M[+ _]: InterpreterErrorsM](term: Var)(implicit depth: Int,
-                                                             env: Env[Par]): M[Either[Var, Par]] =
+  def maybeSubstitute[M[+ _]: Sync](term: Var)(implicit depth: Int,
```
goral09: @henrytill I'm keeping context bounds here because even  though `depth` and `env` are marked as `implicit` they are being passed explicitly and passing `Sync` instances implicitly will be a  pain.

goral09:
@henrytill PTAL

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
henrytill: I didn't do a particularly good job of of conveying this, but my general criticism of context bounds was aimed at the designer(s) of this feature.

> I just find calling Sync[F].delay(…) cleaner than s.delay(…)

I actually find the opposite to be true, as `s` is locally defined and scoped, whereas `Sync[F]` is an (unnecessary) companion object almost always defined elsewhere.  The latter really isn't necessary when you have the former.

 > I don't know anything about scaling problems with this approach.

https://github.com/rchain/rchain/blob/62e3c24be496098410dd3d462485a56557599d63/comm/src/main/scala/coop/rchain/comm/connect/Connect.scala#L139-L141

goral09:
@henrytill I've updated the PR as you requested. Is this OK now?

henrytill:
Given the fact that these changes are mostly in `rholang` and touch the error-reporting stuff, it's @guardbotmk3 who should be reviewing here.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
KentShikama: How did you do that code insertion?

KentShikama:
So we should move away from using Capture[F].capture { } to wrap effectful code now?

PR #1065, opened 2018-06-30
Add deterministic unforgeable names.

guardbotmk3:
## Overview
Adds deterministic unforgeable names by using a splittable, mergeable, random number generator.
We take advantage of the sort order on pars to deterministically split the random number generator.
We use the merging so that we can get unique names even on replicated for's without worrying about sequencing.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -71,7 +71,6 @@ class RuntimeManager private (runtimeContainer: SyncVar[Runtime]) {
       implicit scheduler: Scheduler): Option[Throwable] =
     terms match {
       case deploy +: rest =>
-        val deployBytes = DeployString.toByteArray(deploy.raw.get)
         implicit val rand: Blake2b512Random = Blake2b512Random(
           DeployString.toByteArray(deploy.raw.get))
         Try(reducer.inj(deploy.term.get).unsafeRunSync) match {
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -1,7 +1,10 @@
 package coop.rchain.casper.util.rholang
 
 import com.google.protobuf.ByteString
+import coop.rchain.casper.protocol.Deploy
+import coop.rchain.casper.protocol.DeployString
 import coop.rchain.catscontrib.TaskContrib._
+import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Par
 import coop.rchain.rholang.interpreter.{Reduce, Runtime}
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
@@ -17,8 +20,8 @@ import monix.eval.Task
 class RuntimeManager private (runtimeContainer: SyncVar[Runtime]) {
 
   def replayComputeState(log: trace.Log)(
-      implicit scheduler: Scheduler): (StateHash, List[Par]) => Either[Throwable, Checkpoint] = {
-    (hash: StateHash, terms: List[Par]) =>
+      implicit scheduler: Scheduler): (StateHash, Seq[Deploy]) => Either[Throwable, Checkpoint] = {
+    (hash: StateHash, terms: Seq[Deploy]) =>
       {
         val runtime   = runtimeContainer.take()
         val blakeHash = Blake2b256Hash.fromByteArray(hash.toByteArray)
@@ -36,7 +39,7 @@ class RuntimeManager private (runtimeContainer: SyncVar[Runtime]) {
       }
   }
 
-  def computeState(hash: StateHash, terms: List[Par])(
+  def computeState(hash: StateHash, terms: Seq[Deploy])(
       implicit scheduler: Scheduler): Either[Throwable, Checkpoint] = {
     val resetRuntime: Runtime = getResetRuntime(hash)
     val error                 = eval(terms, resetRuntime.reducer)
@@ -64,11 +67,14 @@ class RuntimeManager private (runtimeContainer: SyncVar[Runtime]) {
     }
   }
 
-  private def eval(terms: List[Par], reducer: Reduce[Task])(
+  private def eval(terms: Seq[Deploy], reducer: Reduce[Task])(
       implicit scheduler: Scheduler): Option[Throwable] =
     terms match {
-      case term :: rest =>
-        Try(reducer.inj(term).unsafeRunSync) match {
+      case deploy +: rest =>
+        val deployBytes = DeployString.toByteArray(deploy.raw.get)
+        implicit val rand: Blake2b512Random = Blake2b512Random(
+          DeployString.toByteArray(deploy.raw.get))
+        Try(reducer.inj(deploy.term.get).unsafeRunSync) match {
           case Success(_)  => eval(rest, reducer)
           case Failure(ex) => Some(ex)
         }
--- a/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
+++ b/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
@@ -10,7 +10,6 @@ import coop.rchain.rholang.collection.LinkedList
 import coop.rchain.rholang.unittest.TestSet
 
 import java.nio.file.Files
-import java.security.SecureRandom
 
 import monix.execution.Scheduler
 
@@ -38,9 +37,7 @@ object TestSetUtil {
   def runTests(tests: Par, otherLibs: Seq[Par], runtime: Runtime)(
       implicit scheduler: Scheduler): Unit = {
     //load "libraries" required for all tests
-    val bytes = new Array[Byte](128)
-    new SecureRandom().nextBytes(bytes)
-    val rand = Blake2b512Random(bytes)
+    val rand = Blake2b512Random(128)
     eval_term(LinkedList.term, runtime)(implicitly, rand.splitShort(0))
     eval_term(TestSet.term, runtime)(implicitly, rand.splitShort(1))
 
--- a/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
+++ b/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
@@ -3,12 +3,14 @@ package coop.rchain.casper.genesis.contracts
 import coop.rchain.casper.util.rholang.InterpreterUtil.mkTerm
 import coop.rchain.catscontrib.Capture.taskCapture
 import coop.rchain.catscontrib.TaskContrib._
+import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Par
 import coop.rchain.rholang.interpreter.Runtime
 import coop.rchain.rholang.collection.LinkedList
 import coop.rchain.rholang.unittest.TestSet
 
 import java.nio.file.Files
+import java.security.SecureRandom
 
 import monix.execution.Scheduler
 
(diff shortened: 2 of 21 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -878,7 +934,10 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
 
   "variable references" should "be substituted before being used." in {
     implicit val errorLog = new Runtime.ErrorLog()
-
+    val splitRandResult = rand.splitByte(3)
+    val splitRandSrc = rand.splitByte(3)
+    splitRandResult.next()
```
goral09: I didn't do good job reviewing your previous PR that added `Blake2b512Random` but this is not functional. I understand that mutation is sometimes more performant.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -148,56 +156,77 @@ object Reduce {
...
+        }
+      }
+      val starts = Vector(par.sends.size,
+                          par.receives.size,
+                          par.news.size,
+                          par.matches.size,
+                          par.bundles.size,
+                          filteredExprs.size)
+        .scanLeft(0)(_ + _)
+      def handle[A](eval: (A => (Env[Par], Blake2b512Random) => M[Unit]), start: Int)(
```
goral09: I am having troubles adding and effect for the cost accounting already and adding this won't make my life easier but I understand and let's go with it for now. I will take care of it later.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
```diff
@@ -34,13 +38,18 @@ object TestSetUtil {
   def runTests(tests: Par, otherLibs: Seq[Par], runtime: Runtime)(
       implicit scheduler: Scheduler): Unit = {
     //load "libraries" required for all tests
-    eval_term(LinkedList.term, runtime)
-    eval_term(TestSet.term, runtime)
+    val bytes = new Array[Byte](128)
+    new SecureRandom().nextBytes(bytes)
+    val rand = Blake2b512Random(bytes)
```
goral09: If you add the smart constructor to the `Blake2b512Random` refactor this part as well.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -148,56 +156,77 @@ object Reduce {
...
+          ta: (A, Int)): M[Unit] = {
+        val newRand =
+          if (starts(6) == 1)
+            rand
+          else if (starts(6) > 256)
+            rand.splitShort((start + ta._2).toShort)
+          else
+            rand.splitByte((start + ta._2).toByte)
```
goral09: I know this is not about this PR but can you add a difference when one should use `splitShort` vs `splitByte` ? In case someone needs to use it in the future and you are not here?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -148,56 +156,77 @@ object Reduce {
...
+                          par.receives.size,
+                          par.news.size,
+                          par.matches.size,
+                          par.bundles.size,
+                          filteredExprs.size)
+        .scanLeft(0)(_ + _)
+      def handle[A](eval: (A => (Env[Par], Blake2b512Random) => M[Unit]), start: Int)(
+          ta: (A, Int)): M[Unit] = {
+        val newRand =
+          if (starts(6) == 1)
```
goral09: If I understand correctly `starts(6)` yields the number of rholang exprssions right? Why do you do that in the `handle` method which will never be invoked with `Expr` (I understand that [this part](https://github.com/rchain/rchain/pull/1065/commits/331badc4a6cea1df6b80330cb2235e0959b84f42#diff-2e087015257005bc83676b5d76c6a366R194) is called on expressions).

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -148,56 +156,77 @@ object Reduce {
...
+            rand
+          else if (starts(6) > 256)
+            rand.splitShort((start + ta._2).toShort)
```
goral09: Where does the `start` comes from?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -148,56 +156,77 @@ object Reduce {
...
+        Parallel.parTraverse(par.news.zipWithIndex.toList)(handle(evalExplicit, starts(2))),
+        Parallel.parTraverse(par.matches.zipWithIndex.toList)(handle(evalExplicit, starts(3))),
+        Parallel.parTraverse(par.bundles.zipWithIndex.toList)(handle(evalExplicit, starts(4))),
+        Parallel.parTraverse(filteredExprs.zipWithIndex.toList)(texpr => {
+          val (expr, idx) = texpr
+          val newRand =
+            if (starts(6) == 1)
+              rand
+            else if (starts(6) > 256)
+              rand.splitShort((starts(5) + idx).toShort)
```
goral09: I don't get it why you first check number of expressions (`starts(6)`) but you split on the bundles? I'm lost.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: This shouldn't make your life any harder.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: No, `starts(6)` is the total number of things in the par that need random state.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: Start is an argument to handle. It is the # of things that came before this one and therefore where to start counting.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: You would use splitShort when you needed to split a generator into more than 256 different generators.
I probably need splitInt and splitLong to avoid an overflow attack.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: Starts(0) is always 0, 
Starts(1) is sends.size
Starts(2) is sends.size + receives.size
Starts(3) is sends.size + receives.size + news.size
...
Starts(5) is everything but filtered expressions, and so that's where we start counting.
Starts(6) is everything, so we use that to see if we can skip a split.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -241,8 +272,10 @@ object Reduce {
                                         substituteChannel[M].substitute(pattern)(1, env))
                     } yield (BindPattern(substPatterns, rb.remainder, rb.freeCount), q))
         // TODO: Allow for the environment to be stored with the body in the Tuplespace
-        substBody <- substitutePar[M].substitute(receive.body.get)(0, env.shift(receive.bindCount))
-        _         <- consume(binds, substBody, receive.persistent)
+        substBody <- substitutePar[M].substituteNoSort(receive.body.get)(
```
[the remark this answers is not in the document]
guardbotmk3: Because nothing is gained by sorting the body. We don't need it in canonical form, we just need it in a deterministic form, and so substituting without re-sorting is still deterministic.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -274,6 +279,7 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
 
   it should "verify that bundle is readable if receiving on Bundle" in {
     implicit val errorLog = new Runtime.ErrorLog()
+    val splitRand = rand.splitByte(1)
```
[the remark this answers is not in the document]
guardbotmk3: For the individual tests, no. It shouldn't affect the results in either case.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
guardbotmk3: The functional version of next() would require array copies.
In actual use, we want the overwrite anyway, because we **don't** want to get the same random numbers twice.

It's only for testing that we intentionally break the API requirements so that we can make sure that we're repeatable.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
```diff
@@ -22,10 +24,12 @@ object TestSetUtil {
...
   }
 
-  def eval_term(term: Par, runtime: Runtime)(implicit scheduler: Scheduler): Unit =
+  def eval_term(term: Par, runtime: Runtime)(implicit scheduler: Scheduler,
+                                             rand: Blake2b512Random): Unit =
     runtime.reducer.inj(term).unsafeRunSync
 
-  def eval(code: String, runtime: Runtime)(implicit scheduler: Scheduler): Unit =
+  def eval(code: String, runtime: Runtime)(implicit scheduler: Scheduler,
+                                           rand: Blake2b512Random): Unit =
```
henrytill: Since you are splitting and explicitly wiring it in several places, you might consider making `rand` an explicit parameter for these helpers.

models/src/main/scala/coop/rchain/models/rholang/sort/ScoreTree.scala
```diff
@@ -22,25 +25,55 @@ trait ScoreTree {
...
-      case (Left(_), Right(_))    => -1
-      case (Right(_), Left(_))    => 1
-      case (Right(s1), Right(s2)) => s1.compare(s2)
+  sealed trait TaggedAtom
+  case class IntAtom(i: Int)          extends TaggedAtom
+  case class StringAtom(s: String)    extends TaggedAtom
+  case class BytesAtom(b: ByteString) extends TaggedAtom
+
+  case class ScoreAtom(value: TaggedAtom) {
+    def bsCompare(b1: ByteString, b2: ByteString) = {
```
henrytill: We might think about eventually moving this to be part of an `implicit class RichByteString(...)` in `shared`

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -64,11 +67,14 @@ class RuntimeManager private (runtimeContainer: SyncVar[Runtime]) {
...
   }
 
-  private def eval(terms: List[Par], reducer: Reduce[Task])(
+  private def eval(terms: Seq[Deploy], reducer: Reduce[Task])(
       implicit scheduler: Scheduler): Option[Throwable] =
     terms match {
-      case term :: rest =>
-        Try(reducer.inj(term).unsafeRunSync) match {
+      case deploy +: rest =>
+        val deployBytes = DeployString.toByteArray(deploy.raw.get)
```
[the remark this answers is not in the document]
guardbotmk3: Done

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -89,11 +91,15 @@ object Interpreter {
                  Task.raiseError(new RuntimeException(mkErrorMsg(errors)))
     } yield (result)
 
-  def evaluate(runtime: Runtime, normalizedTerm: Par): Task[Vector[InterpreterError]] =
+  def evaluate(runtime: Runtime, normalizedTerm: Par): Task[Vector[InterpreterError]] = {
+    val bytes = new Array[Byte](128)
+    new SecureRandom().nextBytes(bytes)
+    implicit val rand = Blake2b512Random(bytes)
```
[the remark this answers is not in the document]
guardbotmk3: Done

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
guardbotmk3: Done.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestSetUtil.scala
guardbotmk3: I'll look at these again.

models/src/main/scala/coop/rchain/models/rholang/sort/ScoreTree.scala
guardbotmk3: I would approve that PR

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
goral09: Make sense :thumbsup:

PR #1066, opened 2018-06-30
Nodes look for ApprovedBlocks on boot instead of making their own genesis

birchmd:
## Overview
This PR does a whole lot. Validator identity is now optional, so read-only nodes are possible. Casper does not start on node boot, instead a Casper constructor starts on node boot, which can create a casper instance from an `ApprovedBlock`. This enables node to accept the result of the genesis block signing protocol [described here](https://rchain.atlassian.net/wiki/spaces/CORE/pages/485556483/Initializing+the+Blockchain+--+Protocol+for+generating+the+Genesis+block) as well addresses weak subjectivity in a way which is even less subjective. A standalone node will still create its own genesis block an automatically approve it. Thus, all nodes connecting to this bootstrap will have the same genesis block automatically. In the future this will be substituted out for the proper genesis signing protocol.

https://rchain.atlassian.net/browse/RHOL-409

### Notes


casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -52,41 +56,117 @@ object CommUtil {
...
+                case Right(response) =>
+                  Log[F].info(s"Received response from $peer! Dispatching...")
+                  Connect
+                    .dispatch[F](response)
+                    .flatMap(_ => {
+                      MultiParentCasperConstructor[F].lastApprovedBlock.map {
+                        case Some(_) => () //got one!
+                        case None    => askPeers(rest, local)
+                      }
+                    })
```
KentShikama: @sebtomba So with this we are introducing a synchronous two way communication at the Casper layer.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -52,41 +56,117 @@ object CommUtil {
...
+        case b @ (_: BlockMessage | _: BlockRequest) =>
+          MultiParentCasperConstructor[F].casperInstance match {
+            case Left(ex) =>
+              Log[F]
+                .warn(
+                  "CASPER: a Casper message was received, " +
+                    "however could not be processed due to the following error. " +
+                    ex.getMessage
+                )
+                .map(_ => none[Packet])
```
KentShikama: Is this equivalent to

`Log[F].warn(...) *> Applicative[F].pure(none[Packet])`?

casper/src/main/scala/coop/rchain/casper/CasperConf.scala
```diff
@@ -1,17 +1,88 @@
...
 }
+
+object CasperConf {
+  private implicit val logSource: LogSource = LogSource(this.getClass)
+
+  def parseValidatorsFile[F[_]: Monad: Capture: Log](
+      knownValidatorsFile: Option[String]): F[Set[ByteString]] =
+    knownValidatorsFile match {
+      //TODO: Add default set? Throw error?
+      case None => Set.empty[ByteString].pure[F]
```
KentShikama: I would say just throw an error for now.

casper/src/main/scala/coop/rchain/casper/CasperConf.scala
```diff
@@ -1,17 +1,88 @@
...
+    }
+
+    (maybeInferred, givenPublicKey) match {
+      case (Some(k1), Some(k2)) =>
+        if (keysMatch(k1, k2)) k1
+        else throw new Exception("Public key not compatible with given private key!")
+
+      case (Some(k1), None) => k1
+
+      case (None, Some(k2)) => k2
```
KentShikama: Hmm if we failed to infer the Ed25519 public key from the private key, that is a red flag. I understand with other algorithms this might be ok but I think we at least should add a `// TODO: Think again` comment.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
sebtomba: @KentShikama @birchmd @rabbitonweb I believe this is a bad idea. Let us discuss it because this is important.

sebtomba: I would like to remove the `roundTrip` from the transport layer and use synchronous communication only for Kademlia actions. However, this requires that Casper reacts only to events. The function `askPeers` is pretty heavy. It does a `roundTrip` dispatches the response and calls `askPeers` recursively. This function is then called in `NodeRuntime` every few seconds. There must exist smarter way to propagate approved blocks.

birchmd: @sebtomba The function is only expensive in the case that the node doesn't already have an approved block. If it had an approved block then none of that communication is done. That being said, repeating the check to see if we have an approved block over and over again is unnecessary and I would rather not do it. I am also completely on board with having Casper being entirely event-based. This initialization step where we need to get an approved block to start Casper is a bit of a special case though because we cannot react to any other event until we have an approved block.

birchmd: Having an in-depth discussion about how comm is being re-designed and how Casper should use it most effectively is welcome. Perhaps we should schedule a time to talk about it for this week.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
birchmd: Yes. Though, in general `*>` is different from `map` because the former sequences unrelated effects, where as the latter transforms values within the same effect. From a style perspective I'm not sure which is better, to have an sequence into a `pure` or an unconditional function into the desired value.

goral09: `*>` is eager where `map/flatMap` is lazy with trampolined monads.

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -293,8 +304,17 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
                       Connect.connectToBootstrap[Effect](addr,
                                                          maxNumOfAttempts = 5,
                                                          defaultTimeout = defaultTimeout)))
-      _ <- if (res.isRight) MonadOps.forever(Connect.findAndConnect[Effect](defaultTimeout), 0)
-          else ().pure[Effect]
+      _ <- {
+        implicit val casperEvidence      = resources.casperConstructor
+        implicit val packetHandlerEffect = resources.packetHandler
```
KentShikama: You have a packetHandlerEffect above too but is this still necessary?

KentShikama: Oh nvm. It is in two different functions.

birchmd:
@rabbitonweb PTAL (travis error is unrelated to the code and I've asked @jeremybusk to take a look at it)

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -52,41 +56,127 @@ object CommUtil {
...
 
+  def requestApprovedBlock[
+      F[_]: Monad: Capture: MultiParentCasperConstructor: Log: Time: Metrics: TransportLayer: NodeDiscovery: ErrorHandler: PacketHandler]
+    : F[Unit] = {
+    val request = ApprovedBlockRequest("PleaseSendMeAnApprovedBlock").toByteString
+
+    def askPeers(peers: List[PeerNode], local: PeerNode): F[Unit] = peers match {
+      case peer :: rest =>
+        for {
+          _    <- Log[F].info(s"CASPER: Sending request for ApprovedBlock to $peer")
```
rabbitonweb: this is not a blocker but just so you know, each log has now a `LogSource` where you can define a common prefix for all your logs in given file

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -52,41 +56,127 @@ object CommUtil {
...
+                        case (Some(sender), Some(_)) =>
+                          Connect
+                            .handlePacket[F](sender, maybePacket)
+                            .flatMap(_ => {
+                              MultiParentCasperConstructor[F].lastApprovedBlock.flatMap {
+                                case Some(_) => ().pure[F] //valid ApprovedBlock received
+                                case None    => askPeers(rest, local)
+                              }
+                            })
+                        case _ => Log[F].info(s"CASPER: Response invalid.") *> askPeers(rest, local)
```
rabbitonweb: this should not be an `info`, it shouyld be logged as `error`. also please include what went wrong, if sender was available, what was the mesage type

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -52,41 +56,127 @@ object CommUtil {
...
+        } yield ()
+
+      case Nil => ().pure[F]
+    }
+
+    for {
+      a     <- MultiParentCasperConstructor[F].lastApprovedBlock
+      peers <- NodeDiscovery[F].peers
+      local <- TransportLayer[F].local
+      _ <- if (a.isEmpty) askPeers(peers.toList, local)
```
rabbitonweb: consider `a.fold(askPeers(peers.toList, local))(kp(().pure[F])` instead

PR #1070, opened 2018-07-03
RHOL-436 cost accounting reducer

goral09:
Pull request created for easier tracking progress and as a place for discussion.

## TODO:

- [ ] Adjust costs of operations
- [x] Fix tests
- [ ] Comprehensive code review

## Notes:

1. Invariants of cost accounting model to consider:
  - Empty collection should be cheaper than `1+ element` collection
  - Given the same number of equivalent terms cost of list and tuple should be the same (?)
  - ...

2. Separate interpreting the program from cost accounting. (Possible source of ideas: [tying the knot](https://wiki.haskell.org/Tying_the_Knot)).


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -57,41 +57,7 @@ object Runtime {
         )
     }
 
-  class ErrorLog extends FunctorTell[Task, InterpreterError] {
```
goral09: Moved to separate file.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -355,62 +354,64 @@ object Reduce {
...
-    def eval(neu: New)(implicit env: Env[Par]): M[Unit] = {
+    private def eval(neu: New)(implicit env: Env[Par]): M[Unit] = {
       def alloc(level: Int): Env[Par] =
         (env /: (0 until level).toList) { (_env, _) =>
           val addr: Par = GPrivate()
           _env.put(addr)
         }
 
-      eval(neu.p.get)(alloc(neu.bindCount))
+      // TODO: should cost of `new` construct depend on the # of bindings?
```
guardbotmk3: Yes, new should cost based on the # of bindings. It's one hash evaluation per 2 names, so it seems pretty fair.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -355,62 +354,64 @@ object Reduce {
...
           for {
-            p       <- eval(v.get)
+            p       <- eval(v)
             evaledP <- evalExpr(p)
           } yield evaledP
         case EMethodBody(EMethod(method, target, arguments, _, _)) => {
           val methodLookup = methodTable(method)
           for {
-            evaledTarget <- evalExpr(target.get)
+            _            <- costAccounting.modify(_ + METHOD_CALL_COST)
```
guardbotmk3: Different methods might have different costs, but this is fine for now.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -420,99 +421,122 @@ object Reduce {
...
         } yield result
 
       expr.exprInstance match {
-        case x: GBool      => Applicative[M].pure[Expr](x)
-        case x: GInt       => Applicative[M].pure[Expr](x)
-        case x: GString    => Applicative[M].pure[Expr](x)
-        case x: GUri       => Applicative[M].pure[Expr](x)
-        case x: GByteArray => Applicative[M].pure[Expr](x)
+        case x: GBool =>
+          Applicative[M].pure[Expr](x) <* costAccounting.modify(_ + BOOLEAN_COST)
```
guardbotmk3: The pure cases shouldn't really have costs.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -420,99 +421,122 @@ object Reduce {
...
         case EEqBody(EEq(p1, p2)) =>
           for {
-            v1 <- evalExpr(p1.get)
-            v2 <- evalExpr(p2.get)
+            v1 <- evalExpr(p1)
+            v2 <- evalExpr(p2)
             // TODO: build an equality operator that takes in an environment.
             sv1 <- substitutePar[M].substitute(v1)(0, env)
             sv2 <- substitutePar[M].substitute(v2)(0, env)
+            _   <- costAccounting.modify(_ + COMPARISON_COST)
```
guardbotmk3: Equality is more like string comparison than the other costs.

ArturGajowy:
Would be great to have tests for cost accounting early, as 1. they might be hard to write and need thought, 2. they'd help understanding the cost function we end up with :)

goral09:
@ArturGajowy We don't know yet how the cost will be calculated. This is first take on how to integrate the cost accounting into the reducer.

ArturGajowy:
@goral09 that's what I understood :) I'm just suggesting having the tests long before the costs calculattion impl is finished, so they can aid the implementation and not be brittle. Don't mind me though, that's just an idea :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: I agree, I planned to put method specific costs into the evaluation itself and that was supposed to be constant price of the method call.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -77,29 +62,11 @@ object Reduce {
...
-            } else {
-              dispatcher.dispatch(continuation, dataList)
-            }
-          case None =>
-            Applicative[M].pure(())
-        }
-
+    private def produce(chan: Quote, data: Seq[Channel], persistent: Boolean): M[Unit] =
       for {
-        substData <- data.toList.traverse(
```
goral09: This substitution is now done before call to `produce`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -77,29 +62,11 @@ object Reduce {
       * @param env  An environment marking the execution context.
       * @return  An optional continuation resulting from a match in the tuplespace.
       */
-    override def produce(chan: Quote, data: Seq[Par], persistent: Boolean)(
```
goral09: `produce` and `consume` are now [extracted to `TuplespaceAlg`](https://github.com/goral09/rchain/blob/1ddaf64d9f7ae50c70ef4be7df617ac006d04e3f/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala#L21).

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -170,34 +123,29 @@ object Reduce {
...
+            case _ => s.unit
         })
-      ).parSequence.map(_ => ())
+      ).parSequence.as(Unit)
     }
 
     override def inj(par: Par): M[Unit] =
-      for { _ <- eval(par)(Env[Par]()) } yield ()
-
-    def debug(msg: String): Unit = {
```
goral09: not used.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -330,21 +286,22 @@ object Reduce {
...
 
       for {
-        evaledTarget <- evalExpr(mat.target.get)
-        // TODO(kyle): Make the matcher accept an environment, instead of
-        // substituting it.
+        evaledTarget <- evalExpr(mat.target)
+        // TODO(kyle): Make the matcher accept an environment, instead of substituting it.
         substTarget <- substitutePar[M].substitute(evaledTarget)(0, env)
         _           <- firstMatch(substTarget, mat.cases)
+        _           <- costAccountingAlg.charge(MATCH_EVAL_COST)
```
goral09: @guardbotmk3 I've added constant cost for each term evaluation (`MATCH_EVAL_COST`, `SEND_EVAL_COST`, `METHOD_CALL_COST` etc) but I am not convinced yet about it.

odeac:
This change grows to a size which is difficult to review. 

I understand it's a big story but I've noticed a few trivial refactoring steps (renames, move class, etc) that could easily be extracted as a separate PR and that would make the reviewer's job a bit easier

goral09:
@odeac Some of the changes are already in #1064 which has been up for some time now. I will try to extract more refactoring to separate PRs later.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -13,98 +13,127 @@ import coop.rchain.models._
 import coop.rchain.models.rholang.sort._
 import coop.rchain.rholang.interpreter.errors.SubstituteError
 import coop.rchain.models.rholang.implicits._
-import errors._
-import coop.rchain.models.rholang.sort.ordering._
+import coop.rchain.models.rholang.sort._
+import coop.rchain.rholang.interpreter.accounting.CostAccountingAlg
```
guardbotmk3: I'm uncertain about needing the cost accounting inside of substitution. We should be able to calculate it outside.

goral09: Well, I don't like it because it pollutes the `Substitution` logic. If cost of substitution is:

1. Cost of the substituted `Par`  (if substitution is possible)
2. Cost of the initial `Par` (if no substitution)
3. Cost of the initial `Par` if an user error occurs

Then we would need to perform another equivalence check -- if what goes out from the `Substitute` equals what went it (and handle error cases). I will look at that possibility and see how it compares to the current one.

goral09: Oh no, it's even simpler than that. It's always the _"cost"_ of what goes out of the substitution... Oh my, that's so obvious. Thanks @guardbotmk3 .

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
```diff
@@ -0,0 +1,81 @@
...
+                             TaggedContinuation],
+      dispatcher: => Dispatch[F, ListChannelWithRandom, TaggedContinuation])(
+      implicit F: Sync[F],
+      P: Parallel[F, M]): TuplespaceAlg[F] = new TuplespaceAlg[F] {
+    override def produce(channel: Channel,
+                         data: ListChannelWithRandom,
+                         persistent: Boolean): F[Unit] = {
+      // TODO: Handle the environment in the store
+      def go(res: Option[(TaggedContinuation, Seq[ListChannelWithRandom])]): F[Unit] =
+        res match {
```
odeac: instead of "manual" matching the Option you could use getOrElse. I find it more readable than doing the match but it's a matter of preference.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
```diff
@@ -0,0 +1,81 @@
...
+      binds match {
+        case Nil => F.raiseError(ReduceError("Error: empty binds"))
+        case _ =>
+          val (patterns: Seq[BindPattern], sources: Seq[Quote]) = binds.unzip
+          pureRSpace
+            .consume(sources.map(q => Channel(q)).toList,
+                     patterns.toList,
+                     TaggedContinuation(ParBody(body)),
+                     persist = persistent)
+            .flatMap {
```
odeac: For me it looks cleaner to use for-comprehension instead of flatMap, as you do in the function "produce".

Anyway, for consistency reasons, it would be better to write them both the same way.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccount.scala
```diff
@@ -0,0 +1,10 @@
+package coop.rchain.rholang.interpreter.accounting
+
+case class CostAccount(idx: Int, cost: BigInt) {
```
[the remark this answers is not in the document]
goral09: I've been thinking about it but I have some doubts when `CostAccount` may be modified in the concurrent context, how should `combine` function behave?

1. 
```
combine(x: CostAccount, y: CostAccount): CostAccount = CostAccount(x.idx + y.idx, x.cost + y.cost)
```

2.
```
combine(x: CostAccount, y: CostAccount): CostAccount = CostAccount(math.max(x.idx, y.idx), math.max(x.cost, y.cost))
```
3.
```
combine(x: CostAccount, y: CostAccount): CostAccount = CostAccount(math.max(x.idx, y.idx) + 1, math.max(x.cost, y.cost))
```
4. ???

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -83,19 +84,20 @@ object Interpreter {
   def execute(runtime: Runtime, reader: Reader): Task[Runtime] =
     for {
       term   <- Task.coeval(buildNormalizedTerm(reader)).attempt.raiseOnLeft
-      errors <- evaluate(runtime, term).attempt.raiseOnLeft
+      errors <- evaluate(runtime, term).map(_._2).attempt.raiseOnLeft
```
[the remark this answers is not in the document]
goral09: :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
```diff
@@ -0,0 +1,81 @@
...
+object TuplespaceAlg {
+  def rspaceTuplespace[F[_], M[_]](
+      pureRSpace: PureRSpace[F,
+                             Channel,
+                             BindPattern,
+                             ListChannelWithRandom,
+                             ListChannelWithRandom,
+                             TaggedContinuation],
+      dispatcher: => Dispatch[F, ListChannelWithRandom, TaggedContinuation])(
+      implicit F: Sync[F],
```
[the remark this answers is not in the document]
odeac: Doing a quick search, in both cats and scalaz I can see that they don't really have a convention for this:
https://github.com/typelevel/cats/search?p=3&q=Monoid&unscoped_q=Monoid
https://github.com/scalaz/scalaz/search?q=Monoid&unscoped_q=Monoid

Still it would be good to have a standard for naming typeclass instances, especially when you have more than one typeclass instances

PR #1074, opened 2018-07-04
Include blessed contracts in genesis block -- including Rev and wallets

birchmd:
## Overview
Genesis block contains contracts, including Rev and Wallets. Wallets come from an additional input into the node at start up.

https://rchain.atlassian.net/browse/RHOL-407

### Notes


casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -5,75 +5,169 @@ import cats.implicits._
...
+
+    val blessedTerms = List(
+      LinkedList.term,
+      NonNegativeNumber.term,
+      MakeMint.term,
+      BasicWallet.term,
+      (new Rev(wallets)).term
+    ).map(termDeploy)
+
+    val Right(checkpoint) = runtimeManager.computeState(startHash, blessedTerms)
```
goral09: Does `computeState` return `Right` always?

```scala
scala> def returnLeft(i: Int): Either[Int, String] = Left(i)
returnLeft: (i: Int)Either[Int,String]

scala> val Right(r) = returnLeft(1)
scala.MatchError: Left(1) (of class scala.util.Left)
  ... 36 elided
```

birchmd: It would be `Left` if injecting a term with the reducer were to fail: https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala#L76 . I don't know under what conditions that happens, but I think in that case something is really wrong and having an exception raised probably good. Though I suppose having that error manifest itself in the form of a match error is not very intuitive.

PR #1094, opened 2018-07-09
No ticket rholang refactoring

goral09:
## Overview
Some ground work before adding cost accounting to Rholang interpreter.

None.

### Notes

### DIFF
--- a/shared/src/main/scala/coop/rchain/shared/AtomicRefMonadState.scala
+++ b/shared/src/main/scala/coop/rchain/shared/AtomicRefMonadState.scala
@@ -1,0 +1,41 @@
+package coop.rchain.shared
+
+import java.util.concurrent.atomic.AtomicReference
+
+import cats.Monad
+import cats.effect.Sync
+import cats.mtl.MonadState
+import cats.implicits._
+
+import scala.annotation.tailrec
+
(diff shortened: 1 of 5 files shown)


shared/src/main/scala/coop/rchain/shared/AtomicRefMonadState.scala
```diff
@@ -0,0 +1,41 @@
...
+import cats.Monad
+import cats.effect.Sync
+import cats.mtl.MonadState
+import cats.implicits._
+
+import scala.annotation.tailrec
+
+//Inspired by Ref from  fs2 and cats
+//TODO: Remove once we upgrade to cats 1.0.1
+final class AtomicRefMonadState[F[_], S] private (private val ar: AtomicReference[S])(
```
odeac: A few comments regarding this code:
* is this really needed now? If you intend to use it later I would add it when you actually add the first usage. This way we understand how the addition was motivated
* If you add a new utility class you should also add some motivation document and tests as usage examples or, at least, point to some existing web page with motivation and usage examples
* if you are using Ref from cats as an example you should probably keep the name and the API, or at least a part of it, in order to make the transition to cats smoother.

PR #1096, opened 2018-07-09
The parallelized `strandPool` implementation based fork-join pool.

DoSOfRedRiver:
You can find a high-level overview here https://rchain.atlassian.net/wiki/spaces/ROS/pages/500367442/strandPool+implementation+in+Roscala+parallelism+support


roscala/src/main/scala/coop/rchain/roscala/Vm.scala
```diff
@@ -619,11 +635,14 @@ object Vm {
 
   def getNextStrand(state: State): Boolean =
     if (state.strandPool.isEmpty) {
-      logger.debug("Empty strandPool - exiting VM")
+      logger.debug("Empty strandPool")
       true
     } else {
-      val ctxt = state.strandPool.remove(state.strandPool.size - 1)
-      logger.debug(s"Get next ctxt ($ctxt) and install it")
+      logger.debug(s"Waiting for a Ctxt to complete")
```
tymm: Are we really waiting here? `strandPool` is not empty and when we get here a `Ctxt` already gave up control.

roscala/src/main/scala/coop/rchain/roscala/ob/Ctxt.scala
```diff
@@ -22,20 +22,27 @@ class Ctxt(var tag: Location,
            var monitor: Monitor)
     extends Ob {
 
+  private val _rslt = new AtomicReference[Ob](Niv)
```
tymm: Can you add some comments here why we are doing that?
Also I fear that we'll have to do the same for `argvec` and `env` (see `doRtn` and `Location.store`) since a `Ctxt` can write into its parent's `argvec` as well.

roscala/src/main/scala/coop/rchain/roscala/pools/SimpleStrandPool.scala
```diff
@@ -0,0 +1,17 @@
...
+import coop.rchain.roscala.ob.Ctxt
+
+import scala.collection.mutable
+
+class SimpleStrandPool extends StrandPool {
+  private val queue = mutable.Buffer[Ctxt]()
+
+  override def append(task: (Ctxt, GlobalEnv)): Unit = queue.append(task._1)
+
+  override def prepend(task: (Ctxt, GlobalEnv)): Unit = queue.prepend(task._1)
```
tymm: Why `(Ctxt, GlobalEnv)`?

roscala/src/main/scala/coop/rchain/roscala/Vm.scala
DoSOfRedRiver: There is a possibility when `ParallelStrandPool` will `join` a task on `deque`, and `join` operation is blocking.

roscala/src/main/scala/coop/rchain/roscala/ob/Ctxt.scala
DoSOfRedRiver: I think you're right. I've been thinking about a better solution for atomic fields, maybe a special macro.

roscala/src/main/scala/coop/rchain/roscala/pools/SimpleStrandPool.scala
DoSOfRedRiver: `ParallelStandPool` relies on `GlobalEnv` in it's logic.

roscala/src/main/scala/coop/rchain/roscala/pools/SimpleStrandPool.scala
```diff
@@ -0,0 +1,33 @@
...
+  override def prepend(task: Task): Unit = queue.prepend(task._1)
+
+  override def getNextStrand(state: State): Boolean =
+    if (queue.isEmpty) {
+      logger.debug("Empty strandPool")
+      true
+    } else {
+      val ctxt = queue.remove(queue.size - 1)
+
+      logger.debug(s"Ctxt completed. Install $ctxt")
```
tymm: I still find this irritating since the `Ctxt` that gave up control not necessarily *completed* (see the `(seq (foo actor) (foo actor))` example).
Please change to "Install $ctxt" or "Ctxt gave up control. Install $ctxt".

PR #1102, opened 2018-07-09
Add Secp256k1.newKeyPair

birchmd:
## Overview
Will be convenient to create test secp256k1 keys for demos and tests.

### Notes


crypto/src/main/scala/coop/rchain/crypto/signatures/Secp256k1.scala
```diff
@@ -1,9 +1,45 @@
...
+    * >>> val data = Sha256.hash("testing".getBytes)
+    * >>> val sig = Secp256k1.sign(data, sec)
+    * >>> Secp256k1.verify(data, sig, pub)
+    * true
+    * }}}
+    *
+    * @return (private key, public key) pair
+    *
+    */
+  def newKeyPair: (Array[Byte], Array[Byte]) = {
```
odeac: This signature is prone to errors. I would like to have a wrapper types to make sure the caller doesn't mix them.

I'm thinking something like this:

```
case class PublicKey(bytes:Array[Byte])
case class SecretKey(bytes:Array[Byte])

def newKeyPair : (PublicKey, SecretKey) = {...}
```

I can see that `Array[Byte]`S are used through the whole file but this is a small change with big long term benefits.

goral09: This is consistent with the approach in all the crypto libraries we use. If we decide to refactor here we should do it in all other cryptos as well.

odeac: Consistently unsafe is not good :)

The crypto libraries provide the implementation of the algorithms and, probably, choosed to return `Array[Byte]` for easier integration. I don't question that decision. 

In our case we need type safety more than we need easy integration. For this reason, on top of the actual bytes that represent the keys, we should add strong type information in order to eliminate a certain class of bugs.

odeac:
It depends on the philosophy of the team. One approach is to commit fast and refactor later. The other approach is to do a bit of refactoring with each commit, such that the later refactoring step is less needed.

Choosing one style or another is a decision that should be taken with the whole team. My personal preference is towards the second approach because, in many cases, "later" becomes "never" or "too late".

goral09:
You yourself pointed (correctly) that my PR shouldn't contain things that were not inherent to the problem. Increasing type safety is not inherent to this PR either. I am not saying this shouldn't be done, I am just saying that refactoring ALL crypto methods (which in turn will propagate changes like a ripple throughout rest of the project) should be done as separate PR.

odeac:
First, I think the type safety is always part of the problem.

Second, there is a misunderstanding. I didn't ask to change the whole file.

My suggestion is that, since this `newKeyPair` function is new and not used yet, the change to make it type-safe is small. 

Later we can (and I think we should) refactor the other crypto functions to use the same approach.

crypto/src/main/scala/coop/rchain/crypto/signatures/Secp256k1.scala
sebtomba: I agree with @odeac . Returning a pair of byte arrays is horrible, especially for a public method.

birchmd:
@odeac @sebtomba PTAL
I also included some code to address the conversion that was in discord about `SecureRandom`
Also note that I have created https://rchain.atlassian.net/browse/CORE-844 to address the new TODO items related to refactoring.

crypto/src/main/scala/coop/rchain/crypto/util/SecureRandomUtil.scala
```diff
@@ -0,0 +1,20 @@
...
+  private val instances = List(
+    "NativePRNGNonBlocking",
+    "Windows-PRNG",
+    "SHA1PRNG"
+  )
+
+  def getSecureRandomInstance: SecureRandom =
+    instances.iterator //use iterator for lazy evaluation
+      .map(name => Try(SecureRandom.getInstance(name)))
+      .collectFirst {
```
odeac: This should be the same as find(_.isSuccess) which is a bit more readable

birchmd: `find(_.isSuccess)` would return an `Option[Try[SecureRandom]]` whereas the current version returns `Option[SecureRandom]` since we have already destructured the `Try`.

odeac: Right. Never mind then.

sebtomba: @birchmd I'm wondering if this shouldn't be a `lazy val`?

birchmd: Yes, I like that idea. I'll make that change

birchmd:
@sebtomba PTAL

crypto/src/main/scala/coop/rchain/crypto/util/SecureRandomUtil.scala
birchmd: Done

PR #1109, opened 2018-07-10
Eagerly evaluate genesis block; include genesis contracts in unit tests

birchmd:
## Overview
Prevents untraced comm events related to genesis from occurring. Should hopefully make the integration tests pass. Note that this likely does not solve a deeper issue: https://rchain.atlassian.net/browse/RHOL-508

https://rchain.atlassian.net/browse/RHOL-505

### Notes


birchmd:
In this PR I use regular rspace instead of replay rspace when processing the genesis block. This is not strictly speaking wrong since we have deterministic unforgable names and there are no races, so everyone should get the same answer, but in principle we should be using replay. I don't have time to make this change now, so I will merge it now and either @KentShikama  or myself can make a separate PR with that change in the future.

PR #1123, opened 2018-07-12
CORE-769 Remove Ping from addNode and move explicitly to handleProtocolHandshake

rabbitonweb:
## Overview
Remove Ping from addNode, call heartbeat explicitly in Rchain protocol

Kademlia constantly updates last seen peers. Calling Ping on
updateLastSeen was a mistake. This is a higher protocol concept thus
was moved to RChain protocol.

CORE-769


comm/src/main/scala/coop/rchain/comm/connect/Connect.scala
```diff
@@ -120,25 +116,36 @@ object Connect {
...
-      phr      = protocolHandshakeResponse(local)
-      addedErr <- NodeDiscovery[F].addNode(peer)
-      commResponse <- addedErr.fold(
+      local  <- TransportLayer[F].local
+      _      <- getOrError[F, ProtocolHandshake](maybePh, parseError("ProtocolHandshake"))
+      hbrErr <- TransportLayer[F].roundTrip(peer, heartbeat(local), defaultTimeout)
+      commResponse <- hbrErr.fold(
                        error =>
-                         Log[F].error(s"Could not add $peer, reason: $error").as(notHandled(error)),
+                         Log[F].warn(s"Not adding. Could receive Pong message back from $peer, reason: $error").as(notHandled(error)),
```
sebtomba: Typo. You mean "couldn't"

comm/src/main/scala/coop/rchain/comm/connect/Connect.scala
```diff
@@ -120,25 +116,36 @@ object Connect {
...
+                         NodeDiscovery[F].addNode(peer) *>
+                           Log[F]
+                             .info(s"Responded to protocol handshake request from $peer")
+                             .as(handledWithMessage(protocolHandshakeResponse(local)))
                      )
     } yield commResponse
 
+  private def handleHeartbeat[F[_]: Monad: TransportLayer: ErrorHandler](
+      peer: PeerNode,
+      maybeHeartbeat: Option[Heartbeat]): F[CommunicationResponse] =
```
sebtomba: I don't get it. Why do we get _maybe_ messages? Can't we bounce invalid messages earlier? IMO `Option` method parameters are a code smell.

rabbitonweb: Check how that method is being used. This is the consequence of using `proto.Any` in `routing.proto`. This will eventually go away once that `proto.Any` is gone.

comm/src/main/scala/coop/rchain/comm/connect/Connect.scala
rabbitonweb: 👍

rabbitonweb:
Duplicates https://github.com/rchain/rchain/pull/1170

PR #1131, opened 2018-07-16
[WIP] Add bonds cache validation over proof of stake stub

KentShikama:
## Overview

We enforce that the bonds cache in the block corresponds to the bonds map in the proof of stake contract. This is preparation of the insertion of the proof of stake contract.

https://rchain.atlassian.net/browse/RHOL-405

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -1,11 +1,10 @@
 package coop.rchain.casper.util.rholang
 
 import com.google.protobuf.ByteString
-import coop.rchain.casper.protocol.Deploy
-import coop.rchain.casper.protocol.DeployString
+import coop.rchain.casper.protocol.{Bond, Deploy, DeployString}
 import coop.rchain.catscontrib.TaskContrib._
 import coop.rchain.crypto.hash.Blake2b512Random
-import coop.rchain.models.Par
+import coop.rchain.models.{Channel, _}
 import coop.rchain.rholang.interpreter.{Reduce, Runtime}
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
 import coop.rchain.rspace.{trace, Blake2b256Hash, Checkpoint}
@@ -14,6 +13,11 @@ import monix.execution.Scheduler
 import scala.concurrent.SyncVar
 import scala.util.{Failure, Success, Try}
 import RuntimeManager.StateHash
+import coop.rchain.crypto.codec.Base16
+import coop.rchain.models.Channel.ChannelInstance.Quote
+import coop.rchain.models.Expr.ExprInstance.{GInt, GString}
+import coop.rchain.rspace.internal.Datum
+import coop.rchain.rspace.trace.Produce
 import monix.eval.Task
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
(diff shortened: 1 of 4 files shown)


KentShikama:
@birchmd PTAL

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -56,6 +60,31 @@ class RuntimeManager private (val initStateHash: ByteString, runtimeContainer: S
     result
   }
 
+  def computeBonds(hash: StateHash): Seq[Bond] = {
+    val resetRuntime     = getResetRuntime(hash)
+    val bondsChannel     = Channel(Quote(Par().copy(exprs = Seq(Expr(GString("proofOfStake"))))))
```
birchmd: I'm concerned about this import source of truth just living at a public name. I would prefer if it lived at an unforgeable name inside a contract and there was read-only access to it via a public name.

As things are today, any validator could unbond any other validator by just changing the bonds listing, however I think that validators should need to unbond themselves through a message they sign. We could get that kind of security through the PoS contract if the bond listing is maintained internally.

KentShikama: I'm not quite aware of how to build a read-only name from protobuf-generated case classes. Would a 

// TODO:  Switch to a read only name

do for now?

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -338,4 +350,39 @@ class ValidateTest extends FlatSpec with Matchers with BeforeAndAfterEach with B
...
+    val (_, validators)   = (1 to 4).map(_ => Ed25519.newKeyPair).unzip
+    val bonds             = validators.zipWithIndex.map { case (v, i) => v -> (2 * i + 1) }.toMap
+    val initial           = Genesis.withoutContracts(bonds = bonds, version = 0L, timestamp = 0L)
+    val storageDirectory  = Files.createTempDirectory(s"hash-set-casper-test-genesis")
+    val storageSize: Long = 1024L * 1024
+    val activeRuntime     = Runtime.create(storageDirectory, storageSize)
+    val runtimeManager    = RuntimeManager.fromRuntime(activeRuntime)
+    val initStateHash     = runtimeManager.initStateHash
+
+    val validatorsString = validators.zipWithIndex
```
birchmd: I would prefer if the validator code were generated using your ProofOfStake class so that it gets tested too.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -56,6 +60,32 @@ class RuntimeManager private (val initStateHash: ByteString, runtimeContainer: S
...
+    val resetRuntime = getResetRuntime(hash)
+    // TODO: Switch to a read only name
+    val bondsChannel     = Channel(Quote(Par().copy(exprs = Seq(Expr(GString("proofOfStake"))))))
+    val bondsChannelData = resetRuntime.space.getData(bondsChannel)
+    runtimeContainer.put(resetRuntime)
+    toBondSeq(bondsChannelData)
+  }
+
+  private def toBondSeq(data: Seq[Datum[ListChannelWithRandom]]): Seq[Bond] = {
+    assert(data.length == 1)
```
birchmd: Rather than asserts, these conditions should be included in the validation logic. We don't want the node to crash because there are multiple things on the proof of stake channel or the map is mis-formatted

KentShikama: The failures will be caught by the Try in bondsCache so the node wouldn't crash.

birchmd: Ah, yes, you're right.

PR #1132, opened 2018-07-16
CORE-761 Benchmark MVCEPP

ys-pyrofex:
## Overview
This benchmark allows checking execution performance of any .rho program. The program shouldn't contain infinite loops.

### DIFF
--- a/rspace-bench/src/main/resources/rholang/mvcepp.rho
+++ b/rspace-bench/src/main/resources/rholang/mvcepp.rho
@@ -1,0 +1,4 @@
+new chan in {
+  for (_ <- chan) { Nil } | chan!(Nil)
+}
+
--- a/rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
+++ b/rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
@@ -1,0 +1,64 @@
+package coop.rchain.rspace.bench
+
+import cats.implicits._
+import org.openjdk.jmh.annotations._
+import monix.eval.Task
+import monix.execution.{ExecutionModel, Scheduler}
+import monix.execution.schedulers.{CanBlock, TrampolineScheduler}
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+//for debug/check run from IDEA with
+//rspaceBench/jmh:run EvalBench.replayMvcEppST -i 1 -wi 0 -f 0 -t 1
+//on hyper-threaded machines for standalone check use -f <number of physical cores>
+//for example
+//java -jar target/scala-2.12/rspacebench_2.12-0.1.0-SNAPSHOT.jar EvalBench -i 10 -wi 5 -f 2 -t 2
+class EvalBench {
+
+  import EvalBench._
+
+  def processErrors(errors: Vector[Throwable]): Unit = if (errors.nonEmpty) {
+    throw new RuntimeException(
+      errors
+        .map(_.toString())
+        .mkString("Errors received during evaluation:\n", "\n", "\n"))
+  }
+
+  def createTest(state: MVCEPPBenchState): Task[Vector[Throwable]] = {
+    val par = state.term.getOrElse(throw new Error("Failed to prepare executable rholand term"))
+    state.runtime.reducer
+      .inj(par)(state.rand)
+      .map(_ => state.runtime.readAndClearErrorVector())
+  }
+
+  //if we run multiple tests on a single-threaded scheduler
+  //they will compete on single execution queue
+  //therefore this test always limited to one thread
+  @Benchmark
+  @Threads(1)
+  def reduceMVCEPPST(state: MVCEPPBenchState): Unit = {
+    val runTask = createTest(state).executeOn(state.singleThreadedScheduler, forceAsync = false)
+    processErrors(
+      runTask.runSyncUnsafe(Duration.Inf)(state.singleThreadedScheduler, CanBlock.permit))
+  }
+
+  @Benchmark
+  def reduceMVCEPPMT(state: MVCEPPBenchState): Unit = {
+    implicit val scheduler: Scheduler = monix.execution.Scheduler.Implicits.global
+    val runTask = createTest(state)
+    processErrors(Await.result(runTask.runAsync, Duration.Inf))
+  }
+}
+
+object EvalBench {
+
+  @State(Scope.Benchmark)
+  class MVCEPPBenchState extends EvalBenchStateBase {
+    val singleThreadedScheduler: Scheduler = TrampolineScheduler.apply(
+      Scheduler.singleThread(name = "mvcepp-1"),
+      ExecutionModel.SynchronousExecution)
+
+    override val rhoScriptSource: String = "/rholang/mvcepp.rho"
+  }
+}
--- a/rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBenchStateBase.scala
+++ b/rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBenchStateBase.scala
@@ -1,0 +1,59 @@
+package coop.rchain.rspace.bench
+
+import java.io.{FileNotFoundException, InputStreamReader}
+import java.nio.file.{Files, Path}
+import org.openjdk.jmh.annotations.{Setup, TearDown}
+
+import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.models.Par
+import coop.rchain.rholang.interpreter.{Interpreter, Runtime}
+import coop.rchain.shared.PathOps.RichPath
+
+trait EvalBenchStateBase {
+  private val dbDir: Path = Files.createTempDirectory("rchain-storage-test-")
+  private val mapSize: Long = 1024 * 1024 * 1024
+
+  val rhoScriptSource: String
+  val runtime: Runtime = Runtime.create(dbDir, mapSize)
+  val rand: Blake2b512Random = Blake2b512Random(128)
+  var term: Option[Par] = None
+
+  /**
+    * until we add flag 'delete_lmdb_dir_on_close' for benchmarks and unit-tests
+    * this prevents periodic out of disk space failures
+    */
+  def deleteOldStorages(): Unit =
+    dbDir.getParent.toFile.listFiles
+      .filter(dir => dir.isDirectory && (dir != dbDir))
+      .filter(_.getName.startsWith("rchain-storage-test-"))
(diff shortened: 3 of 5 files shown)


rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+  @State(Scope.Benchmark)
+  class MvcEppBenchState extends BenchStateBase {
+    val singleThreadedScheduler: Scheduler = TrampolineScheduler.apply(
+      Scheduler.singleThread(name = "mvcepp-1"),
+      ExecutionModel.SynchronousExecution)
+
+    override val rhoScriptSource: String = "/rholang/mvcepp.rho"
+  }
+
+  trait BenchStateBase {
```
henrytill: Could we move this to a separate file and call it `EvalBenchStateBase`?

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+
+  def processErrors(errors: Vector[Throwable]): Unit = if (errors.nonEmpty) {
+    throw new RuntimeException(
+      errors
+        .map(_.toString())
+        .mkString("Errors received during evaluation:\n", "\n", "\n"))
+  }
+
+  def prepareRunTask(state: MvcEppBenchState): Task[Vector[Throwable]] = {
+    val par = state.term.getOrElse(throw new Error("Failed to prepare executable rholand term"))
```
henrytill: `s/rholand/rholang/`

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+    state.runtime.reducer
+      .inj(par)(state.rand)
+      .map(_ => state.runtime.readAndClearErrorVector())
+  }
+
+  //if we run multiple tests on a single-threaded scheduler
+  //they will compete on single execution queue
+  //therefore this test always limited to one thread
+  @Benchmark @Threads(1)
+  def replayMvcEppST(state: MvcEppBenchState): Unit = {
```
henrytill: Let's make these a little more consistent:

```scala
  @Benchmark @Threads(1)
  def evaluateMVCEPPSingleThreaded(state: MVCEPPBenchState): Unit = {
    implicit val scheduler: Scheduler = state.singleThreadedScheduler
    implicit val canBlock: CanBlock   = CanBlock.permit
    val eval: Task[Vector[Throwable]] = createTest(state)
    val results: Vector[Throwable]    = eval.runSyncUnsafe(Duration.Inf)
    processErrors(results)
  }

  @Benchmark
  def evaluateMVCEPPMultiThreaded(state: MVCEPPBenchState): Unit = {
    implicit val scheduler: Scheduler = monix.execution.Scheduler.Implicits.global
    val eval: Task[Vector[Throwable]] = createTest(state)
    val results: Vector[Throwable]    = Await.result(eval.runAsync, Duration.Inf)
    processErrors(results)
  }
```

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+    val runTask = prepareRunTask(state).executeOn(state.singleThreadedScheduler, forceAsync = false)
+    processErrors(
+      runTask.runSyncUnsafe(Duration.Inf)(state.singleThreadedScheduler, CanBlock.permit))
+  }
+
+  @Benchmark
+  def replayMvcEppMT(state: MvcEppBenchState): Unit = {
```
henrytill: (See other comment)

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+      ExecutionModel.SynchronousExecution)
+
+    override val rhoScriptSource: String = "/rholang/mvcepp.rho"
+  }
+
+  trait BenchStateBase {
+    private val dbDir: Path   = Files.createTempDirectory("rchain-storage-test-")
+    private val mapSize: Long = 1024 * 1024 * 1024
+
+    val rhoScriptSource: String
```
henrytill: Could we have this be:
```scala
val rholangSourceFile: Path
```

(see how I use the `Path` in other comment)

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+            dir.toPath.recursivelyDelete
+          } catch {
+            case _: Exception =>
+        })
+
+    @Setup
+    def doSetup(): Unit = {
+      deleteOldStorages()
+
+      def resourceFileReader(path: String): InputStreamReader =
```
henrytill: Could we move this to the outermost context, since it's a fairly general helper function?

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+        new InputStreamReader(
+          Option(getClass.getResourceAsStream(path))
+            .getOrElse(throw new FileNotFoundException(path)))
+
+      val rhoScriptReader = resourceFileReader(rhoScriptSource)
+
+      term = Interpreter.buildNormalizedTerm(rhoScriptReader).runAttempt match {
```
henrytill: Just inline `resourceFileReader(rholangSourceFile)` here.

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+    val runtime: Runtime       = Runtime.create(dbDir, mapSize)
+    val rand: Blake2b512Random = Blake2b512Random(128)
+    var term: Option[Par]      = None
+
+    /**
+      * until we add flag 'delete_lmdb_dir_on_close' for benchmarks and unit-tests
+      * this prevents periodic out of disk space failures
+      */
+    def deleteOldStorages(): Unit =
+      dbDir.getParent.toFile.listFiles
```
henrytill: This seems like a really fragile way to be getting our hands on `TMPDIR` and all of the test directories.

Is there a way for us to fix this problem in a general fashion so they we don't have to do this here?

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
```diff
@@ -0,0 +1,119 @@
...
+  import EvalBench._
+
+  def processErrors(errors: Vector[Throwable]): Unit = if (errors.nonEmpty) {
+    throw new RuntimeException(
+      errors
+        .map(_.toString())
+        .mkString("Errors received during evaluation:\n", "\n", "\n"))
+  }
+
+  def prepareRunTask(state: MvcEppBenchState): Task[Vector[Throwable]] = {
```
henrytill: maybe `createTest`?

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
ys-pyrofex: Well, I don't like this way too, but it looks almost perfect after removing these folders 10x times. I found that during benchmark runs all LMDB files are locked and can't be deleted, and if the are not locked - they never accessed anymore.

There is another way. We even discussed it approx one month ago. We have a lot of different storage related unit-tests and now with benchmarks count of "rchain-storage-test-" increases. That's not a problem on build server since it's cleaned before every build. I think we need to add a "test mode" flag to LMDBStore/LMDBTrieStore, and storage folder should delete automatically. Since this refactoring isn't related to benchmark and will affect numerous projects (Rspace, Rholang, Bench, etc). I think it should be extracted to a separate task. Removal of this function can be a final step of this task.

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
ys-pyrofex: Actually that's not a path, but resource key. And function getResourceAsStream doesn't have Path overload. However when a file located in resources it automatically gets a key equal to it's relative path (so it's a path, lol). To avoid cases like "Path constructor 'fixed' incoming value, for example by replacing '/' with '\'" on Windows I decided to leave type String,

rspace-bench/src/main/resources/rholang/mvcepp.rho
```diff
@@ -0,0 +1,3 @@
+new chan in {
+  for (_ <- chan) { Nil } | chan!(Nil)
+}
```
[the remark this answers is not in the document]
ys-pyrofex: Done, seems git diff don't see it.

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
henrytill: I meant to delete this comment before submitting my review, as I realized yr approach makes sense given the fact that you are including the rholang source file as a resource. 

So this is fine.

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBench.scala
henrytill: Understood.  I'm fine with this approach here and now in this PR.  Just was curious to hear your thoughts.

rspace-bench/src/main/scala/coop/rchain/rspace/bench/EvalBenchStateBase.scala
```diff
@@ -0,0 +1,59 @@
...
+import java.nio.file.{Files, Path}
+import org.openjdk.jmh.annotations.{Setup, TearDown}
+
+import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.models.Par
+import coop.rchain.rholang.interpreter.{Interpreter, Runtime}
+import coop.rchain.shared.PathOps.RichPath
+
+trait EvalBenchStateBase {
+	private val dbDir: Path   = Files.createTempDirectory("rchain-storage-test-")
```
henrytill: It looks like indentation was accidentally increased (or perhaps some tabs snuck in?).

PR #1135, opened 2018-07-17
Cleanups on the go

ArturGajowy:
## Overview
Several cleanups I've made while reading the code.

### Which JIRA issue does this PR relate to? If there is not a JIRA issue 
N/A

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,0 +1,3 @@
+-J-Xmx2G
+-J-Xms2G
+-J-XX:MaxMetaspaceSize=512m
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -39,13 +39,6 @@ import monix.eval.Coeval
   */
 trait Reduce[M[_]] {
 
-  def produce(chan: Quote, data: Seq[Par], persistent: Boolean)(implicit env: Env[Par],
-                                                                rand: Blake2b512Random): M[Unit]
-
-  def consume(binds: Seq[(BindPattern, Quote)], body: Par, persistent: Boolean)(
-      implicit env: Env[Par],
-      rand: Blake2b512Random): M[Unit]
-
   def eval(par: Par)(implicit env: Env[Par], rand: Blake2b512Random): M[Unit]
 
   def inj(par: Par)(implicit rand: Blake2b512Random): M[Unit]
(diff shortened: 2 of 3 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -39,21 +39,23 @@ import monix.eval.Coeval
   */
 trait Reduce[M[_]] {
 
-  def produce(chan: Quote, data: Seq[Par], persistent: Boolean)(implicit env: Env[Par],
-                                                                rand: Blake2b512Random): M[Unit]
-
-  def consume(binds: Seq[(BindPattern, Quote)], body: Par, persistent: Boolean)(
-      implicit env: Env[Par],
-      rand: Blake2b512Random): M[Unit]
+  protected[this] implicit def monadM: Monad[M]
```
goral09: We don't need to put this constraint in the trait. It should work with any `M[_]` at this point.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -39,21 +39,23 @@ import monix.eval.Coeval
...
     * Evaluate any top level expressions in @param Par .
     */
-  def evalExpr(par: Par)(implicit env: Env[Par]): M[Par]
+  def evalExpr(par: Par)(implicit env: Env[Par]): M[Par] =
+    for {
+      evaledExprs <- par.exprs.toList.traverse(expr => evalExprToPar(expr))
+      result = evaledExprs.foldLeft(par.copy(exprs = Vector())) { (acc, newPar) =>
+        acc ++ newPar
+      }
+    } yield result
```
goral09: Implementations of those methods were moved out from the `trait` on purpose. I plann on removing `evalExpr`  and `evalExprToPar` from the `trait` alltogether leaving only `inj` and `eval`.

ArturGajowy: Great! :) I've dropped that commit. Good to merge?

.sbtopts
```diff
@@ -0,0 +1,3 @@
+-J-Xmx2G
+-J-Xms2G
```
goral09: Did you try running 
`;clean;reload;update;compile;test` with  this settings? I think it will overflow.

ArturGajowy: It seemed to fix things for me after the overflows I had, but I'll try your scenario as well :) Did it overflow for you? What are your memory settings?

goral09: I usually use `5G` just because I can ;) but I know people had problem with tests at 2 gigs.

ArturGajowy: For the others reading this: the change only affects default settings, so your custom ones should still be applied automatically.

For those without custom settings, I've tested the build several times and had no OOM errors since applying those settings. Will update them if I spot anything OFC :)

PR #1137, opened 2018-07-17
[RHOL-445]: Add convenience methods for String

itegulov:
## Overview
Add convenience methods for String:

- length method
- slice method
- interpolation operator %
- concatenation operator ++

[RHOL-445](https://rchain.atlassian.net/browse/RHOL-445)

### Notes
Is there any mechanism for converting ground terms into Strings (I do not want to override the case class geenrated toString)? Should I use `PrettyPrinter` or maybe implement instances for `cats.Show` typeclass?


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -827,6 +894,44 @@ object Reduce {
...
 
+    private[this] def length: MethodType =
+      (p: Par, args: Seq[Par]) =>
+        (env: Env[Par]) => {
+          def length(baseExpr: Expr): M[Expr] =
+            baseExpr.exprInstance match {
+              case GString(string) =>
+                Applicative[M].pure[Expr](GInt(string.length))
+              case _ =>
+                s.raiseError(ReduceError("Error: method 'length' can only be called on Strings."))
```
goral09: According to latest comment/discussions we should use errors messages like:
> Error: Method `length` is not defined on _the_type_

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -827,6 +894,44 @@ object Reduce {
...
+
+    private[this] def slice: MethodType =
+      (p: Par, args: Seq[Par]) =>
+        (env: Env[Par]) => {
+          def slice(baseExpr: Expr, from: Int, until: Int): M[Par] =
+            baseExpr.exprInstance match {
+              case GString(string) =>
+                Applicative[M].pure[Par](GString(string.slice(from, until)))
+              case _ =>
+                s.raiseError(ReduceError("Error: method 'slice' can only be called on Strings."))
```
goral09: Same comment applies here.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -537,7 +537,74 @@ object Reduce {
...
+                           }
+                       case (_: GString, _) =>
+                         s.raiseError(ReduceError("Error: String can be interpolated only by Map"))
+                       case _ =>
+                         s.raiseError(ReduceError("Error: only Strings can be interpolated"))
+                     }
+          } yield result
+        case EPlusPlusBody(EPlusPlus(p1, p2)) =>
+          for {
+            v1 <- evalSingleExpr(p1.get)
```
goral09: I think you can drop `.get` . There are some residues here and there but thanks to 
```
Par p1 = 1 [(scalapb.field).no_box = true];
```

in proto files it won't wrap in `Option`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -537,7 +537,74 @@ object Reduce {
...
+            v2 <- evalSingleExpr(p2.get)
+            result <- (v1.exprInstance, v2.exprInstance) match {
+                       case (GString(lhs), GString(rhs)) =>
+                         Applicative[M].pure[Expr](GString(lhs + rhs))
+                       case _ =>
+                         s.raiseError(ReduceError("Error: only Strings can be concatenated"))
```
goral09: Same comment about error messages applies.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -537,7 +537,74 @@ object Reduce {
...
+                  ReduceError(s"Error: interpolation doesn't support ${value.getClass}")
+                )
+              case _ =>
+                s.raiseError[(String, String)](
+                  ReduceError("Error: interpolation Map should only contain String keys")
+                )
+            }
+          for {
+            v1 <- evalSingleExpr(p1.get)
+            v2 <- evalSingleExpr(p2.get)
```
goral09: Drop `.get`s .

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -537,7 +537,74 @@ object Reduce {
...
+                                 case (k, v) => current.startsWith("${" + k + "}")
+                               } match {
+                                 case Some((k, v)) =>
+                                   result ++= v
+                                   current = current.drop(k.length + 3)
+                                 case None =>
+                                   result += current.head
+                                   current = current.tail
+                               }
+                             }
```
goral09: Couldn't we just `replace` elements of format `${x}` in the string? 
https://scalafiddle.io/sf/nQz5nW1/0

itegulov: I also thought so (my second commit implementation looked just like yours), but unfortunately it is not so easy. Consider the following situation:
"Hello, ${name1}! Bye, ${name2}!" % {"name1": "Bob ${name2}", "name2": "Alice"}
Your implementation returns "Hello, Bob Alice! Bye, Alice!", while mine returns "Hello, Bob ${name2}! Bye, Alice!"

goral09: Ok, makes sense. Can you interpolating part to the method?

goral09: Also, I would prefer to see this implemented as a regex but it's just my personal taste. We can keep it as is.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
itegulov: Are you sure? Every single case here has them. Should I also delete them as well?

goral09: If it compiles after removing then it will work.

itegulov:
Still working on the tests for normalization and reduction

itegulov:
Wrong target

PR #1145, opened 2018-07-18
[RHOL-504]: Add cost accounting to the spatial matcher

goral09:
## Overview
Adds cost accounting to the spatial matcher.

https://rchain.atlassian.net/browse/RHOL-504

### Notes


rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -0,0 +1,139 @@
...
+package object matcher {
+
+  type FreeMap            = Map[Int, Par]
+  type OptionalFreeMap[A] = StateT[Option, FreeMap, A]
+
+  type OptionalFreeMapWithCost[A] = StateT[OptionT[State[CostAccount, ?], ?], FreeMap, A]
+
+  object OptionalFreeMapWithCost {
+
+    class OptionalFreeMapWithCostOps[A](s: OptionalFreeMapWithCost[A]) {
```
guardbotmk3: This is a nice use of implicits.

There are probably some others we could use

models/src/main/protobuf/RhoTypes.proto
```diff
@@ -42,9 +42,16 @@ message Channel {
...
+message PCost {
+    uint64 cost = 1;
+    int32 iterations = 2;
+}
+
 message ListChannelWithRandom {
     repeated Channel channels = 1;
     bytes randomState = 2 [(scalapb.field).type = "coop.rchain.crypto.hash.Blake2b512Random"];
+    // cost of performing the spatial match
+    PCost cost = 3 [(scalapb.field).no_box = true];
```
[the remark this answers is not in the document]
goral09: Who pays for interacting with the contract? If the client then we should charge for `produce` as well

goral09:
@guardbotmk3 I don't see any places for improvements in this PR.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  }
+
+  type NonDetFreeMap[A] = StateT[Stream, FreeMap, A]
+
+  def emptyMap: FreeMap = Map.empty[Int, Par]
+
+}
+
+final case class StreamT[F[_], A](value: F[Stream[A]]) {
```
ArturGajowy: Let's move this out to a separate file as it's purely generic and should not be mixed with domain code, nor distract from it. Maybe even to a `contrib` or `util` module if we have one, so others can use it.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1053,10 +1053,8 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
 
   "variable references" should "be substituted before being used." in {
     implicit val errorLog = new ErrorLog()
-    val costAccounting =
-      CostAccountingAlg.monadState(AtomicRefMonadState.of[Task, CostAccount](CostAccount.zero))
-    val splitRandResult = rand.splitByte(3)
-    val splitRandSrc    = rand.splitByte(3)
+    val splitRandResult   = rand.splitByte(3)
+    val splitRandSrc      = rand.splitByte(3)
```
ArturGajowy: thanks for ggetting rid of that `val costAccounting` boilerplate in each test :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
```diff
@@ -65,52 +62,47 @@ class VarMatcherSpec extends FlatSpec with Matchers {
...
+    assert(
+      spatialMatchOptionalTest(spatialMatch(target, pattern),
+                               Some(Map[Int, Par](0 -> GInt(9).prepend(GInt(7))))))
   }
   "Matching extras with wildcard" should "work" in {
     val target: Par  = GInt(7).prepend(GInt(8)).prepend(GInt(9))
     val pattern: Par = GInt(8).prepend(EVar(Wildcard(Var.WildcardMsg())))
-    val result       = spatialMatch(target, pattern).runS(emptyMap)
-    result should be(Some(Map.empty[Int, Par]))
+    assert(spatialMatchOptionalTest(spatialMatch(target, pattern), Some(Map.empty[Int, Par])))
```
ArturGajowy: maybe let's go a step further and make it:
```
assertSpatialMatch(target, pattern, Some(Map.empty[Int, Par]))
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -0,0 +1,145 @@
...
+    class OptionalFreeMapWithCostOps[A](s: OptionalFreeMapWithCost[A]) {
+      def modifyCost(f: CostAccount => CostAccount): OptionalFreeMapWithCost[A] =
+        StateT((m: FreeMap) => {
+          OptionT(State((c: CostAccount) => {
+            val (cost, result) = s.run(m).value.run(c).value
+            (f(cost), result)
+          }))
+        })
+
+      def runWithCost: (CostAccount, Option[(FreeMap, A)]) =
```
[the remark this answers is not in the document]
goral09: clash with `run` defined on  `StateT`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -334,42 +234,44 @@ object SpatialMatcher {
...
+                               wildcard: Boolean)(
       implicit lf: HasLocallyFree[T],
-      sm: SpatialMatcher[T, T]): NonDetFreeMap[Unit] = {
+      sm: SpatialMatcher[T, T]): NonDetFreeMapWithCost[Unit] = {
     val exactMatch = !wildcard && varLevel.isEmpty
     val plen       = plist.length
     val tlen       = tlist.length
     if (exactMatch && plen != tlen)
-      StateT.liftF(None)
+      NonDetFreeMapWithCost.emptyMap[Unit].modifyCost(_.charge(COMPARISON_COST))
```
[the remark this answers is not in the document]
goral09: I think `modify` is better name because we supply a function `f: CostAccount => CostAccount`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -129,7 +132,8 @@ object Reduce {
       val rspaceCost                                        = body.storageCost + patterns.storageCost + srcs.storageCost
       for {
         _ <- costAccountingAlg.charge(rspaceCost)
-        _ <- tuplespaceAlg.consume(binds, ParWithRandom(body, rand), persistent)
+        c <- tuplespaceAlg.consume(binds, ParWithRandom(body, rand), persistent)
+        _ <- costAccountingAlg.modify(_.charge(c))
```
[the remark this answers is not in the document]
goral09: We do use `.set` actually. When we start with empty environment in `inj` method on `Reduce`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -106,7 +108,8 @@ object Reduce {
                         rand: Blake2b512Random): M[Unit] =
       for {
         _ <- costAccountingAlg.charge(Channel(chan).storageCost + data.storageCost)
-        _ <- tuplespaceAlg.produce(Channel(chan), ListChannelWithRandom(data, rand), persistent)
+        c <- tuplespaceAlg.produce(Channel(chan), ListChannelWithRandom(data, rand), persistent)
+        _ <- costAccountingAlg.modify(_.charge(c))
```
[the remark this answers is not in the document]
goral09: `charge(c: Cost)` , `modify(f: CostAccount => CostAccount)`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
goral09: Well, it's still there I think  :)

goral09: I will replace with the reference to empty cost accounting.

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
goral09: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
guardbotmk3: Agreed.

models/src/main/protobuf/RhoTypes.proto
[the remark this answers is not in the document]
guardbotmk3: No, I mean that we use ListChannelWithRandom to store the data written for a `produce`, and there's no need to supply a PCost in that case.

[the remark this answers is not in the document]
goral09: Well, we don't need to persist it but to get the cost for performing a spatial match we need to put it in `ListChannelWithRandom` because that's the API of rspace: https://github.com/rchain/rchain/blob/2c231c3fb7c2350ce2939758c3659ad2f4eb7914/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala#L32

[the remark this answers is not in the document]
goral09: I will change it to boxed type but we didn't have to supply it even when it wasn't boxed -- proto was using `PCost.defaultInstance` (probably used default values for both types of the `PCost`) if not passed explicitly.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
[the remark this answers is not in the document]
ArturGajowy: Yeah I've seen that. But all the usages of `modify` are `.modify(_.charge($c))` (for some $c of type Cost), so it's equivalent to just `costAccountingAlg.charge(c)`. Which I'd prefer, together with removing the `modify` method. We can always add it back when it's needed.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
goral09: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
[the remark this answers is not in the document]
ArturGajowy: I think this corresponds with your discussion with @rabbitonweb on the Cell typeclass PR. Can we somehow separate `.set` into a separate interface, so the types guard where it's used? Maybe create a fresh instance of CostAccountingAlg on `reduce.inj` isntead of resetting it?

I won't insist on `set`, but would really like to see `.modify` hidden and `.charge` being used instead.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
ArturGajowy: Yeah :D

why empty? Isn't zero the default state of a cost account?

ArturGajowy: Also, while we're at it - wouldn't defining a class-level `val costAccounting = costAccountingZero` and deleteing the vals inisde tests get rid of (some of) the boilerplate?

goral09: No, because we don't want to share the `CostAccounting` state between  test cases.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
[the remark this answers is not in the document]
goral09: I can go the other way around -- use `modify` instead of `charge` but removing `modify` is not possible.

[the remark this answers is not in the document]
goral09: > so it's equivalent to just `costAccountingAlg.charge(c)`

no, it's not really. `CostAccount` , besides `cost`,  carries also information about _how many times we charged_ . I can't pull this information out of the thin air.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
ArturGajowy: I missed the AtomicRef in the test, so yeah...

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -52,11 +52,11 @@ object SystemProcesses {
                               TaggedContinuation],
                 dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation])
     : Seq[ListChannelWithRandom] => Task[Unit] = {
-    case Seq(ListChannelWithRandom(Seq(arg, ack), rand)) =>
+    case Seq(ListChannelWithRandom(Seq(arg, ack), rand, cost)) =>
       Task(Console.err.println(prettyPrinter.buildString(arg))).flatMap { (_: Unit) =>
         space
           .produce(ack,
-                   ListChannelWithRandom(Seq(Channel(Quote(Par.defaultInstance))), rand),
+                   ListChannelWithRandom(Seq(Channel(Quote(Par.defaultInstance))), rand, cost),
```
guardbotmk3: It seems to me that it should be empty on produce, and that rather than passing it through, we should use an explicit None.

guardbotmk3: Applies through this whole file.

goral09: @guardbotmk3 I decided to merge. We can sort this in the next PR. I will get in touch.

PR #1148, opened 2018-07-18
Add the tests folder

differentialderek:
## Overview
We add a folder of tests to show functionality of the framework and lack thereof. Includes tests on process functionality, pattern matching, sending/receiving and detecting inappropriate free variables.

N/A

### Notes
Each PR will come with new tests that show what works now and what still does not work. We will mention goals with regards to getting a certain test to work.


goral09:
Please change the file names to not include whitespaces.

rholang/src/main/k/rholang/tests/Free-Variables/Finding-free-vars-with-logical-or-between-channels.rho
```diff
@@ -0,0 +1,9 @@
+// Logical connectives are slightly nuanced.
+
+// The first match should not execute due to the
+// free variable in the pattern.
+// Since nothing matches, this will rewrite to Nil.
+match for(@{@{x \/ Nil}!(Nil)} <- @Nil){ Nil }
```
goral09: This won't pass the parser because grammar is as follows:
`_match_ Proc4 "{" [Case] "}"`

and `for( …) { P } ` is Proc2

goral09: This will:
```
@0!(for(x <- @Nil) { Nil }) | for (@y <- @0) { match y { x => x } }
```

and not only this will parse but it will also match.

rholang/src/main/k/rholang/tests/Sending-Receiving/logical-and-within-patterns.rho
```diff
@@ -0,0 +1,24 @@
+// The following program consists of a receive which listens for a pattern
+// which itself is a send over *either* @1 or @2. It grabs whatever is being
+// sent and sends it over @Nil.
+
+// This will nondeterministically either rewrite to:
+//     @Nil!(@1!("first succeeds"))
+//     @Nil!("second succeeds")
+// or
+//     @Nil!(@2!("second succeeds"))
+//     @Nil!("first succeeds")
```
goral09: It will be either
```
@Nil!("first succeeds")
@Nil!(@"second succeeds"!(Nil))
```

or

```
@Nil!("second succeeds")
@Nil!(@"first succeeds"!(Nil))
```

no **@1! …** or **@2! …**

rholang/src/main/k/rholang/tests/Std-Pattern-Matching/matching-with-the-new-construct.rho
```diff
@@ -0,0 +1,8 @@
+// A successful match rewrites to
+// Nil!("success")
+
+match
+  new x in { x!(Nil) }
+{
+  new y in { y!(z) } => { z!("success") }
+}
```
goral09: This won't parse (although it probably should) because of how coercions are ordered.

rholang/src/main/k/rholang/tests/Free-Variables/Finding-free-vars-with-logical-or-between-channels.rho
differentialderek: Thanks for pointing this out; I notice now a slight issue in the way I wrote the framework. I will address this in a future PR, I still need to think about how that affects the framework right now.

rholang/src/main/k/rholang/tests/Std-Pattern-Matching/matching-with-the-new-construct.rho
differentialderek: I'm not sure I understand why this doesn't parse?

goral09: Well, when I tried running this contract it didn't pass the parser phase.

differentialderek: Ah, there is a problem with this example. The `z` is a process variable, not a name variable, so `z!("success")` should be replaced with `@z!("success")`. In addition to that change, wrap both the target and the pattern in braces and that should run.

goral09: Yes, exactly.

differentialderek: You're absolutely right. They are working in the framework but I neglected to run them through the current rholang compiler. I'll do that for future examples.

goral09: :+1: good work nevertheless.

PR #1157, opened 2018-07-20
CORE-850 ISpace.retrieve

ys-pyrofex:
## Overview
ISpace.retrieve gets a GNAT from the history trie at a particular checkpoint and channels hash.

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/package.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/package.scala
@@ -51,16 +51,18 @@ package object history {
       }
     }
 
-  def lookup[T, K, V](store: ITrieStore[T, K, V], branch: Branch, key: K)(
-      implicit codecK: Codec[K]): Option[V] = {
+  private[this] def lookup[T, K, V](txn: T,
+                                    store: ITrieStore[T, K, V],
+                                    branchRootHash: Blake2b256Hash,
+                                    key: K)(implicit codecK: Codec[K]): Option[V] = {
     val path = codecK.encode(key).map(_.bytes.toSeq).get
 
     @tailrec
-    def loop(txn: T, depth: Int, curr: Trie[K, V]): Option[V] =
+    def loop(depth: Int, curr: Trie[K, V]): Option[V] =
       curr match {
         case Skip(affix, pointer) =>
           store.get(txn, pointer.hash) match {
-            case Some(next) => loop(txn, depth + affix.length.toInt, next)
+            case Some(next) => loop(depth + affix.length.toInt, next)
             case None       => throw new LookupException(s"No node at ${pointer.hash}")
           }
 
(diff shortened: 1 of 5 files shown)


rspace/src/test/scala/coop/rchain/rspace/HistoryActionsTests.scala
```diff
@@ -36,6 +36,14 @@ trait HistoryActionsTests
 
   type TestGNAT = GNAT[String, Pattern, String, StringsCaptor]
 
+  def getRootHash(store: IStore[String, Pattern, String, StringsCaptor],
```
henrytill: It might be good to add a comment that this a convenience function for testing.

rspace/src/main/scala/coop/rchain/rspace/history/package.scala
```diff
@@ -51,8 +51,9 @@ package object history {
       }
     }
 
-  def lookup[T, K, V](store: ITrieStore[T, K, V], branch: Branch, key: K)(
-      implicit codecK: Codec[K]): Option[V] = {
+  private[this] def lookup[T, K, V](store: ITrieStore[T, K, V],
+                                    root: Either[Branch, Blake2b256Hash],
```
henrytill: @dzajkowski @lukasz-golebiewski: How do you both feel about this use of `Either`?  

While I tend not to mostly use `Either` for error-reporting, this seems like a reasonable approach to avoiding code duplication while exporting an overloaded definition for the two different ways to look things up.

rspace/src/test/scala/coop/rchain/rspace/HistoryActionsTests.scala
dzajkowski: and `private[rspace]` access level?

rspace/src/main/scala/coop/rchain/rspace/history/package.scala
dzajkowski: alternative would be something along these lines?
```
def lookup[T, K, V](store: ITrieStore[T, K, V], rootHash: Blake2b256Hash, key: K)(
    implicit codecK: Codec[K]): Option[V] = 
store.withTxn(store.createTxnRead()) { (txn: T) => lookup(txn, store, rootHash, key) }

def lookup[T, K, V](store: ITrieStore[T, K, V], branch: Branch, key: K)(
     implicit codecK: Codec[K]): Option[V] = 
store.withTxn(store.createTxnRead()) { (txn: T) => lookup(store, store.getRoot(txn, branch), key) }

```

rspace/src/test/scala/coop/rchain/rspace/HistoryActionsTests.scala
ys-pyrofex: Done.

rspace/src/main/scala/coop/rchain/rspace/history/package.scala
ys-pyrofex: After some thinking I decided to apply changes proposed by @dzajkowski - both implementations will work well, however we also have lookup method that accepts sequence of keys. And I suppose we will save some ticks if all of them will be done inside one transaction.

PR #1167, opened 2018-07-24
CORE-769 Remove Ping from addNode and move explicitly to handleProtocolHandshake

rabbitonweb:
Overview
Remove Ping from addNode, call heartbeat explicitly in Rchain protocol

Kademlia constantly updates last seen peers. Calling Ping on
updateLastSeen was a mistake. This is a higher protocol concept thus
was moved to RChain protocol.

Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please
CORE-769


rabbitonweb:
Duplicates https://github.com/rchain/rchain/pull/1170

PR #1176, opened 2018-07-24
CORE-886 Explicit responses on 'propose' action.

dzajkowski:
## Overview
Additional behavior in "propose" action that allows the integration tests to check if propose was successful.

https://rchain.atlassian.net/browse/CORE-886

### Notes
none


casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -16,9 +16,14 @@ object DeployRuntime {
   def propose[F[_]: DeployService: Monad](): F[Unit] =
     DeployService[F].createBlock().flatMap {
       case Some(block) =>
-        DeployService[F].addBlock(block)
+        println(s"Response: Successfully created block")
```
goral09: println leftover

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -16,9 +16,14 @@ object DeployRuntime {
...
+        for {
+          response <- DeployService[F].addBlock(block)
+          _        = println(s"Response: ${response._2}")
```
goral09: println leftover

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -16,9 +16,14 @@ object DeployRuntime {
...
+        } yield ()
 
       case None =>
+        println(s"Response: Failure! Did not create block")
```
goral09: println leftover

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
dzajkowski: I didn't want to change how the existing communication is done (via println's).
WDYT - would it be better to introduce a logger?
The goal is (AFAI understand it) to communicate with the user via simple messages.

goral09: Oh, I saw usage of `Log` earlier but it was from node. This is probably a question to @rabbitonweb and @sebtomba

casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -16,7 +16,7 @@ trait DeployService[F[_]] {
   def createBlock(): F[Option[BlockMessage]]
   def showBlock(q: BlockQuery): F[String]
   def showBlocks(): F[String]
-  def addBlock(b: BlockMessage): F[Unit] //add a block to Casper internal state
+  def addBlock(b: BlockMessage): F[(Boolean, String)] //add a block to Casper internal state
```
lukasz-golebiewski: Why not `Either` so that it's consistent with casper?

dzajkowski: to keep aligned with `def deploy(d: DeployString): F[(Boolean, String)]` in the same trait.
@birchmd WDYT which should be preferred?

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
KentShikama: How does ._2 work here? What is the type of response?

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
birchmd: The `Log` instance we have is server-side, but we want this information to be shown client-side, so I think `println` is the right call.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -67,7 +67,20 @@ private[api] class DeployGrpcService[
 
   override def createBlock(e: Empty): Future[MaybeBlockMessage] = BlockAPI.createBlock[F].toFuture
 
-  override def addBlock(b: BlockMessage): Future[Empty] = BlockAPI.addBlock[F](b).toFuture
```
birchmd: I think it would be better to update `BlockAPI.addBlock` with the new logic and leave this file unchanged.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
birchmd: Reusing the deploy return type is good with me. Idea was that `_1` is a simple success/failure Boolean and `_2` is a message giving more information. And this is exactly the same as gRPC `DeployServiceResponse`

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
birchmd: `response` is of type `(Boolean, String)`.

The `println` does look a little funny on the other side of an `=` like that though. I think using `Capture` like we do in `deployFileProgram` would be better.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -49,8 +49,8 @@ class GrpcDeployService(host: String, port: Int) extends DeployService[Task] {
     response.toProtoString
   }
 
-  def addBlock(b: BlockMessage): Task[Unit] =
-    Task.delay {
-      blockingStub.addBlock(b)
-    }.void
+  def addBlock(b: BlockMessage): Task[(Boolean, String)] = Task.delay {
+    val response = blockingStub.addBlock(b)
```
KentShikama: @birchmd Should this not return `Either[InvalidBlock, Unit]` though?

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
dzajkowski: should I also move the logic for `doDeploy` to `BlockApi`?

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -46,7 +46,7 @@ import scodec.Codec
 import scodec.bits.BitVector
 
 trait Casper[F[_], A] {
-  def addBlock(b: BlockMessage): F[Unit]
+  def addBlock(b: BlockMessage): F[Either[InvalidBlock, Unit]]
```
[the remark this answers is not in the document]
dzajkowski: should I just return `attempt` from the whole `for`? or should I try to figure out what the status should be after a possible `reAttemptBuffer`?

casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
birchmd: No, the `blockingStub.addBlock` is calling this: https://github.com/rchain/rchain/pull/1176/files#diff-72121b5dc081b75121497cdaf951797dR70

casper/src/main/scala/coop/rchain/casper/Casper.scala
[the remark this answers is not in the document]
birchmd: I think returning `attempt` is fine. Though it is a good point that there is no indication of blocks being added from the buffer as a result of adding the current block. If we wanted that then I guess we would need to return a `Seq[(BlockHash, BlockStatus)]` or something. Maybe that's overkill though? How useful do we think that extra information would be?

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
birchmd: Yes, that could be a good think to do for consistency. `BlockAPI` did not exist when `doDeploy` was first written.

casper/src/main/scala/coop/rchain/casper/Casper.scala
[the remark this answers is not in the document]
dzajkowski: I'm aiming for a maximum value - minimum effort as this is a first draft of checking if `propose` works in integration tests so I'll stick with attempt. If it proves to be not enough there's always a revisit possible :)
Thanks!

scripts/p2p-test-tool.py
```diff
@@ -282,6 +282,19 @@ def test_propose(container):
             r = container.exec_run(['sh', '-c', cmd])
             for line in r.output.decode('utf-8').splitlines():
                 print(line)
+                if "Response: Success!" in line:
+                    print(f"{container.name}: {line}")
+                    if retval != 1:
```
jeremybusk: This will guarantee that at least one of proposes succeeds.  This is better than it was. I'll rewrite it to work for any failure if some successes and some Failures happen on propose loops we still want it to fail.

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -155,16 +155,16 @@ sealed abstract class MultiParentCasperInstances {
                     else if (!validSender) InvalidUnslashableBlock.pure[F]
                     else attemptAdd(b)
           _ <- attempt match {
-                case MissingBlocks         => ().pure[F]
```
birchmd: You don't need to change either of these lines because the return type of `reAttemptBuffer` is `F[Unit]`

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -1,34 +1,62 @@
...
-  def addBlock[F[_]: Monad: MultiParentCasperConstructor: Log](b: BlockMessage): F[Empty] =
+  def addBlock[F[_]: Monad: MultiParentCasperConstructor: Log](
+      b: BlockMessage): F[DeployServiceResponse] = {
+    def doAddBlock(implicit casper: MultiParentCasper[F]): F[DeployServiceResponse] =
+      for {
+        status <- MultiParentCasper[F].addBlock(b)
+      } yield
+        status match {
+          case _: InvalidBlock => DeployServiceResponse(false, s"Failure! Invalid block: $status")
+          case _: ValidBlock   => DeployServiceResponse(true, s"Success! Valid: $status")
```
birchmd: There is only one way for a block to be valid, so the message will always read "Success! Valid: Valid", which seems a little funny.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -1,34 +1,62 @@
...
+          case _               => DeployServiceResponse(false, s"Failure! Unknown status: $status")
```
birchmd: Is this case really needed? `BlockStatus` is a `sealed trait` with just the two subcases you have already matched on.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -1,34 +1,62 @@
...
+        }
+
     MultiParentCasperConstructor
-      .withCasper[F, Empty](_.addBlock(b).map(_ => Empty.defaultInstance), Empty.defaultInstance)
+      .withCasper[F, DeployServiceResponse](doAddBlock(_), DeployServiceResponse(false, "Failure!"))
```
birchmd: This failure message could be more clear to the user. It is failure because the node does not have any Casper instance. Perhaps use the same message as in `deploy` above.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -13,12 +11,17 @@ import scala.util.{Failure, Success, Try}
 
 object DeployRuntime {
 
-  def propose[F[_]: DeployService: Monad](): F[Unit] =
+  def propose[F[_]: Monad: Capture: DeployService](): F[Unit] =
     DeployService[F].createBlock().flatMap {
       case Some(block) =>
-        DeployService[F].addBlock(block)
+        println(s"Response: Successfully created block")
```
birchmd: Please include this under a `Capture` in the for-comprehension as well.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -13,12 +11,17 @@ import scala.util.{Failure, Success, Try}
...
+        for {
+          response <- DeployService[F].addBlock(block)
+          _        <- Capture[F].capture { println(s"Response: ${response._2}") }
+        } yield ()
 
       case None =>
+        println(s"Response: Failure! Did not create block")
```
birchmd: Please put this under a `Capture` and remove the `().pure[F]`. It rather misleading to have pure unit returned simply because the intended side-effect have been placed out of scope of the `F`.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
dzajkowski: missed it, will do.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
dzajkowski: future proofing. old habits die hard...

PR #1177, opened 2018-07-25
CORE-893 kademlia transport abstraction

rabbitonweb:
## Overview

In this PR:
1. `Ping` typeclass is renamed to `KademliaRPC`
2. `lookup` method is moved from `KademliaNodeDiscovery` to `KademliaRPC`

This way `KademliaNodeDiscovery` is *almost* ready to stop relying on `TransportLayer`. It will require to move `Disconnect` to Rchain protocol to completely remove dependency on `TransportLayer`

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-893

### DIFF
--- a/node/src/main/scala/coop/rchain/node/effects/package.scala
+++ b/node/src/main/scala/coop/rchain/node/effects/package.scala
@@ -28,9 +28,10 @@ package object effects {
     }
   }
 
-  def ping[F[_]: Monad: Capture: Metrics: TransportLayer](src: PeerNode,
-                                                          timeout: FiniteDuration): Ping[F] =
-    new Ping[F] {
+  def kademliaRPC[F[_]: Monad: Capture: Metrics: TransportLayer](
+      src: PeerNode,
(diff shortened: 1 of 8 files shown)


node/src/main/scala/coop/rchain/node/effects/package.scala
```diff
@@ -13,6 +13,7 @@ import coop.rchain.comm.discovery._
 import coop.rchain.shared._
 import scala.concurrent.duration.FiniteDuration
 import java.io.File
+import coop.rchain.comm.protocol.routing.{Ping => _, _}
```
sebtomba: Please check if this is required

sebtomba: I mean `Ping` is gone.

rabbitonweb: good catch, i will remove it

PR #1182, opened 2018-07-25
[RHOL-447] Validate no deploy by the same (user, millisecond timestamp) has been…

AbnerZheng:
## Overview
Validate no deploy by the same (user, millisecond timestamp) has been produced in the chain

https://rchain.atlassian.net/browse/RHOL-447

### Notes


casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -151,12 +151,15 @@ sealed abstract class MultiParentCasperInstances {
           validSig    <- Validate.blockSignature[F](b)
           dag         <- blockDag
           validSender <- Validate.blockSender[F](b, genesis, dag)
+          validDepoly <- Validate.repeatDeploy[F](b, genesis, dag)
```
KentShikama: typo: deploy

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -30,6 +30,15 @@ object Validate {
       "ed25519" -> Ed25519.verify
     )
 
+  def childMapIterator(blockHashSet: Set[BlockHash],
+                       dag: BlockDag,
+                       internalMap: Map[BlockHash, BlockMessage]) = new Iterator[BlockMessage] {
+    val undelying: Iterator[BlockHash] = blockHashSet.iterator
```
KentShikama: typo: underlying

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -30,6 +30,15 @@ object Validate {
       "ed25519" -> Ed25519.verify
     )
 
+  def childMapIterator(blockHashSet: Set[BlockHash],
+                       dag: BlockDag,
```
KentShikama: Where is the dag used here?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +150,44 @@ object Validate {
...
+    * Validate no deploy by the same (user, millisecond timestamp)
+    * has been produced in the chain
+    */
+  def repeatDeploy[F[_]: Monad: Log: BlockStore](block: BlockMessage,
+                                                 genesis: BlockMessage,
+                                                 dag: BlockDag): F[Boolean] =
+    BlockStore[F].asMap().flatMap { internalMap: Map[BlockHash, BlockMessage] =>
+      {
+        val deployKeySet = block.getBody.newCode.map(d => (d.getRaw.user, d.getRaw.timestamp)).toSet
+        val iterator = DagOperations.bfTraverse(Some(genesis))(x =>
```
KentShikama: Can we use b (for "block") instead of x?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +150,44 @@ object Validate {
...
+            _.newCode.exists(
+              _.raw.exists(p => deployKeySet.contains((p.user, p.timestamp)))
+            )
+          )
+        })
+        // it is allowed for there to be the same (user, timestamp) transaction in both forks
+        val parents = ProtoUtil.parents(block)
+        val invalid = repeatedBlocks.exists(b1 => {
+          parents.contains(b1.blockHash)
+        })
```
KentShikama: How does this cover forks that are longer than one block? Are you familiar with the syntax "P <- C" meaning block P is the parent of block C

A <- B1 <- C1 <- D1 <- E1
A <- B2 <- C2 <- D2<- E2

Imagine that B1 and D2 have the same (user, timestamp) transaction. That would be ok too.

AbnerZheng: I think it works. When you deploy D2, I will firstly traverse the dag to find these blocks that contain deploy with same (user, timestamp) transaction, which is B1 in your case. But D2's parents is [A, B2, C2, D2], it doesn't contain B2, so it is valid.

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: done

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: done. Sorry about it.

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: Yeah, it is not need anymore.

casper/src/main/scala/coop/rchain/casper/Casper.scala
AbnerZheng: done

casper/src/main/scala/coop/rchain/casper/Validate.scala
KentShikama: `ProtoUtil.parents(D2) == C2`. It doesn't include B2 or A. You'll probably need to write something similar to `ProtoUtil.getMainChain`. Make sure to add a test case that fails otherwise.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -246,6 +246,8 @@ object ProtoUtil {
     ByteString.copyFrom(Base16.decode(string))
 
   def basicDeployString(id: Int): DeployString = {
+    //todo this shoule be removed once we assign the deploy with exact user
```
KentShikama: Can you make this

// TODO:

Later on we want to grep for this pattern exactly when we do clean up

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -265,6 +267,8 @@ object ProtoUtil {
   }
 
   def termDeploy(term: Par): Deploy = {
+    //todo this shoule be removed once we assign the deploy with exact user
```
KentShikama: Same here

KentShikama:
@birchmd Can you take a look too? Besides the nitpick above it LGTM

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -30,6 +32,14 @@ object Validate {
       "ed25519" -> Ed25519.verify
     )
 
+  def childMapIterator(blockHashSet: Set[BlockHash], internalMap: Map[BlockHash, BlockMessage]) =
+    new Iterator[BlockMessage] {
```
birchmd: I don't understand the point of defining your own iterator instance. Is this not the same as `blockHashSet.iterator.map(internalMap)`?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +151,49 @@ object Validate {
...
+    */
+  def repeatDeploy[F[_]: Monad: Log: BlockStore](block: BlockMessage,
+                                                 genesis: BlockMessage,
+                                                 dag: BlockDag): F[Boolean] =
+    BlockStore[F].asMap().flatMap { internalMap: Map[BlockHash, BlockMessage] =>
+      {
+        def parents(b: BlockMessage): Iterator[BlockMessage] =
+          ProtoUtil.parents(b).iterator.flatMap(internalMap.get)
+
+        val deployKeySet = block.getBody.newCode.map(d => (d.getRaw.user, d.getRaw.timestamp)).toSet
```
birchmd: I'm a little worried about these `get*` methods. I know that the technically optional fields they get should always be present, but we don't want to make it trivially easy to crash the node by simply sending it an empty message. I would prefer if you used a for-comprehension like
```scala
(for {
  bd <- block.body
  d  <- bd.newCode
  r  <- d.raw
} yield (r.user, r.term, r.timestamp)).toSet
```

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +151,49 @@ object Validate {
       }
     }
 
+  /**
+    * Validate no deploy by the same (user, millisecond timestamp)
```
birchmd: I think you also need to include `term` in the uniqueness key. The same user can deploy two different pieces of code at the exact same time and that is fine.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +151,49 @@ object Validate {
...
+        })
+
+        if (!invalid) {
+          true.pure[F]
+        } else {
+          for {
+            _ <- Log[F].warn(
+                  ignore(
+                    block,
+                    "found deploy by the same (user, millisecond timestamp) produced in the chain"))
```
birchmd: It would be nice if the log had more information. Like this particular deploy was already included in this particular block.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -141,6 +151,49 @@ object Validate {
...
+  def repeatDeploy[F[_]: Monad: Log: BlockStore](block: BlockMessage,
+                                                 genesis: BlockMessage,
+                                                 dag: BlockDag): F[Boolean] =
+    BlockStore[F].asMap().flatMap { internalMap: Map[BlockHash, BlockMessage] =>
+      {
+        def parents(b: BlockMessage): Iterator[BlockMessage] =
+          ProtoUtil.parents(b).iterator.flatMap(internalMap.get)
+
+        val deployKeySet = block.getBody.newCode.map(d => (d.getRaw.user, d.getRaw.timestamp)).toSet
+        val iterator = DagOperations.bfTraverse(Some(genesis))(b =>
```
birchmd: The algorithm you have seems unnecessary complicated and expensive. You accumulate all of instances of the deploys in the current block which have appeared anywhere in the DAG and then check if those repetitions are in a different fork. Save yourself the trouble and only look on the current fork to begin with.
```scala
val invalid =  bfTraverse[BlockMessage](Some(block))(parents).exists(
  _.body.exists(
       _.newCode.exists(
         _.raw.exists(p => deployKeySet.contains((p.user, p.timestamp)))
      )
  )
)
```

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: Look like we will rRemove term from deploy proto. 
https://rchain.atlassian.net/browse/RHOL-541

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: Done.

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: Fantastic. And we should exclude current block in traversing.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
AbnerZheng: Done

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
AbnerZheng: Done

casper/src/main/scala/coop/rchain/casper/Validate.scala
birchmd: True, but I was talking about the `term` in `DeployString`. It's fine though, we talked today in standup that we will no longer use `term` as part of the uniqueness key for other reasons, which means the PR is fine as is.

casper/src/main/scala/coop/rchain/casper/Validate.scala
AbnerZheng: Done

AbnerZheng:
@birchmd I have resolved the conflicts, could you please help me merge this PR?

PR #1183, opened 2018-07-25
Export / Import improvements

kjmcqueen:
## Overview

- Moved object code collection to the VM execute method to more closely model how JIT compile and execute works.
- Split out Location handling from the ExpandedLocation object
- Generalized the looped reference handling to handle longer loops and for all object types.
- Added support for Export/Import of meta and parent objects.
- Added support for several additional object types

https://rchain.atlassian.net/browse/ROS-425
https://rchain.atlassian.net/browse/ROS-437

### Notes


rosette/src/Export.cc
```diff
@@ -575,17 +621,166 @@ void rblTableObjectHandler( ObjectCodePB::Object * lvob, pOb ob, ObjectCodePB::O
     populateObjectByType(rt->tbl, tbl);
 }
 
+void ctxtObjectHandler( ObjectCodePB::Object * pbob, pOb ob, ObjectCodePB::ObType pbtype) {
```
tymm: What was the reason again for exporting `Ctxt`?

rosette/src/Export.cc
```diff
@@ -575,17 +621,166 @@ void rblTableObjectHandler( ObjectCodePB::Object * lvob, pOb ob, ObjectCodePB::O
...
+
+    ObjectCodePB::Object *extension = imob->mutable_extension();
+    populateObjectByType(im->extension, extension);
+}
+
+void stdOprnObjectHandler( ObjectCodePB::Object * pbob, pOb ob, ObjectCodePB::ObType pbtype ) {
+    ObjectCodePB::StdOprn *soob = pbob->mutable_std_oprn();
+    StdOprn * so = (StdOprn *)ob;
+
+    ObjectCodePB::Object *extension = soob->mutable_extension();
```
tymm: Looks good but as a side note: I haven't seen `StdOprn` making use of its `Actor` capabilities anywhere. Only saw it being used as a key so far. If you know about places where it is not just used as a key, I would be interested in that.

rosette/src/Export.cc
```diff
@@ -575,17 +621,166 @@ void rblTableObjectHandler( ObjectCodePB::Object * lvob, pOb ob, ObjectCodePB::O
...
+    populateObjectByType(rmo->code, code);
+
+    ObjectCodePB::Object *id = rmob->mutable_id();
+    populateObjectByType(rmo->id, id);
+
+    ObjectCodePB::Object *source = rmob->mutable_source();
+    populateObjectByType(rmo->source, source);
+}
+
+void productTypeObjectHandler( ObjectCodePB::Object * pbob, pOb ob, ObjectCodePB::ObType pbtype ) {
```
tymm: You really need this for `boot.rbl`? Can you tell me where this occurs?

rosette/src/Import.cc
```diff
@@ -440,11 +462,111 @@ pOb createRosetteObject(ObjectCodePB::Object * ob) {
...
+        retval = StdOprn::create(INVALID, RBLBOOL(false));
+    }
+
     default:
         warning("Import object type=%d not yet implemented!\nObjectPB=\n%s\n", ob->type(), ob->DebugString().c_str());
         assert(false);
     }
 
+    // Handle meta and parent fields. Not for RblAtom objects.
+    if (retval && TAG(retval) == OTptr) {
```
tymm: `RblAtom`s have `meta` and `parent` objects as well. Is this handled elsewhere?

rosette/src/Export.cc
kjmcqueen: I had thought that a ctxt was required for execution of a Code block in the VM.  I may be mistaken.

rosette/src/Export.cc
kjmcqueen: I don't know of any specific uses. I just know when I started processing meta and parent pointers, this showed up as being needed. Whenever I go to process a pOb, if its type isn't implemented, I log a warning. I then go and add support for it and in some cases I've added support for the Objects from which it is derived.  In my next go around, I plan to better and more fully support object inheritance.

rosette/src/Import.cc
kjmcqueen: The only place that RblAtom objects are fully instantiated that I know of is in RblAtom.cc around line 323. Here a single instance of each is created for use in the BASE() function which calls decodeAtom(). This function copies the meta data into the single instance of the object, and returns a pointer to it.  This seems to be only used to call member functions, that print, display, or otherwise output the value of the Atom.  The rest of the time, as far as I know, RblAtoms are passed around as pseudo objects with their data contained within the pOb itself. It this use, there are no meta or parent objects.

tymm: `Fixnum` is a `RblAtom` and has a `meta` and `parent` for example.

rosette/src/Export.cc
tymm: Yes, that's correct. However I don't think there is a need to export it (especially for the `opLookup` hack we want to do). The `Code` that runs inside of a certain `Ctxt` will change certain fields in the `Ctxt` while executing. The `Ctxt` is just there to store temporary state for the associated `Code` object.

rosette/src/Import.cc
tymm: Yes, but once you called `BASE()` on the pointer and for example got a `Fixnum`, that `Fixnum` has a `meta` and a `parent` we have to keep track of when exporting.

kjmcqueen: The only meta and parent that it has at that point are those of the global "prototypicalFixnum" object. As near as I can tell, those are never set to anything interesting.

There is really only ever one instance of the Fixnum object, and that is the "prototypicalFixnum" created by MODULE_INIT() in RblAtom.cc. As near as I can tell, the meta and parent pointers there are never set to anything different for a specific Fixnum.

Look closely at how decodeAtom() in RblAtom.cc works. You'll see that the only thing that happens when it is called is the pseudo-pointer pOb tag is referenced to determine what type of RblAtom this is, and then the pseudo-pointer is transferred into the atom member variable of the appropriate global Atom instance, and the address of the global is returned.

The bottom line is, every time you use BASE() to access a Fixnum, you will get the same memory address pointing to the same Fixnum instance, with the same meta and parent pointers. The only thing that changes is the value of the "atom" member variable will contain the original pOb contents that were passed to BASE().

tymm: They are of interested. For example the `Fixnum` parent is the `Fixnum SBO` which maps from operations (e.g. the `+` operation/`StdOprn`) to primitives (e.g. the `fx+` primitive).
These mappings get built up while running `boot.rbl`.

I guess you are right in the sense that it's not really important for exporting and importing. All we have to know is the type and the rest can be hardcoded.

tymm: Yes. What you described there sounds fine.

rosette/src/Export.cc
kjmcqueen: Done

PR #1185, opened 2018-07-25
CORE-741 - Kademlia, same peer, different IPs

rabbitonweb:
## Overview
In this PR we fix how Kademlia should behave when it tries to add node that already exists but its IP address has changed.
After investigation the error was found, Kademlia should update the IP address information.

Two tests were added to `KademliaSpec` with following output:

```
[info] A PeertTable with 1 byte addresses and k = 3
(...)
[info]   when adding a peer when that peer already exists but with different IP
[info]   - should replace peer with new entry (the one with new IP)
[info]   - should move peer to begining of the bucket (meaning it's been seen lately)
```

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-741

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/PeerNode.scala
+++ b/comm/src/main/scala/coop/rchain/comm/PeerNode.scala
@@ -31,8 +31,7 @@ final case class PeerNode(id: NodeIdentifier, endpoint: Endpoint) {
   def key  = id.key
   val sKey = id.toString
 
-  override def toString =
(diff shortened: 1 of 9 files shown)


rabbitonweb:
This PR waits for https://github.com/rchain/rchain/pull/1177 to be first merged to master... ping @sebtomba :)

PR #1194, opened 2018-07-26
CORE-903/CORE-774 Cell typeclass

rabbitonweb:
## Overview
In TcpTransport layer we are using `MonadState` to hold open connections. This is not good enough, as it might hit race conditions when multipple threads try to manipulate the state.

This PR introduces a `Cell` typeclass - it is simplified version of `MonadState` that only holds single method `modify`. This way we can provide a thread-safe instance of it based on `monix.MVar`. Each thread will call `modify` but if other thread is running the modifications at the given moment, the `modify` will block until access is released by the second thread.

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-774

### DIFF
--- a/shared/src/main/scala/coop/rchain/shared/Cell.scala
+++ b/shared/src/main/scala/coop/rchain/shared/Cell.scala
@@ -2,6 +2,7 @@ package coop.rchain.shared
 
 import cats._, cats.data._, cats.implicits._
 import coop.rchain.catscontrib._, Catscontrib._
+import monix.eval.{MVar, Task}
 
 trait Cell[F[_], S] {
   def modify(f: S => S): F[Unit]
@@ -17,6 +18,12 @@ object Cell {
 
   class NOPCell[F[_]: Applicative, S] extends Cell[F, S] {
     def modify(f: S => S): F[Unit] = ().pure[F]
+  }
+
+  def mvarCell[S](mvar: MVar[S]): Cell[Task, S] = new Cell[Task, S] {
+    def modify(f: S => S): Task[Unit] = mvar.take >>= {
+      case s => mvar.put(f(s))
+    }
   }
 }
 
--- a/shared/src/main/scala/coop/rchain/shared/Cell.scala
+++ b/shared/src/main/scala/coop/rchain/shared/Cell.scala
@@ -1,0 +1,26 @@
+package coop.rchain.shared
+
+import cats._, cats.data._, cats.implicits._
+import coop.rchain.catscontrib._, Catscontrib._
+
+trait Cell[F[_], S] {
+  def modify(f: S => S): F[Unit]
+}
+
+object Cell {
+  def apply[F[_], S](implicit ev: Cell[F, S]): Cell[F, S] = ev
+
+  def forTrans[F[_]: Monad, T[_[_], _]: MonadTrans, S](implicit C: Cell[F, S]): Cell[T[F, ?], S] =
+    new Cell[T[F, ?], S] {
+      def modify(f: S => S): T[F, Unit] = C.modify(f).liftM[T]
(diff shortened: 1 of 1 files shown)


goral09:
According to the [cats-effect docs](https://typelevel.org/cats-effect/concurrency/mvar.html#use-case-synchronized-mutable-variables): 

>Obviously after the call for take and before the call for put happens we could have concurrent logic that can update the same variable. While the two operations are atomic by themselves, a combination of them isn’t atomic (i.e. atomic operations don’t compose), therefore if we want this sample to be safe, then we need extra synchronization. 

You should take a look at [Ref](https://typelevel.org/cats-effect/concurrency/ref.html) docs and maybe this little read: [things to store in ref](https://olegpy.com/things-to-store-in-a-ref/).

rabbitonweb:
@goral09  I just had this conversation  with Dominik. `Ref` is stupid as it is basically `MonadState` in disguise - nothing new. Plus it has broken contract for multithreaded  access. 

When I have `modify` only typeclass, that is based on `MVar` I can have a guarantee that multi-thread access is safe. You can not have this guarantee with `Ref`.

To give you an example, create instance of `Ref` based on `MVar`. Then write piece of code like:

```scala
for {
s1 <- get
s2 <- get
_ <- put(s2)
_ <- put(s1)
} yield s1 = s2
```

Guess what, it blocks :)

goral09:
@rabbitonweb What do you mean by 

>  Plus it has broken contract for multithreaded access.

Docs say:

> Provides safe concurrent access and modification of its content, but no functionality for synchronisation, which is instead handled by Deferred.

Your implementation  is only safe because it exposes _only_ `modify` method and hides other. The underlying `take` and `put` are still not safe.

goral09:
You would not create an instance of `Ref` based on `MVar` in practice. The `Ref` from `cats-effect` uses `AtomicReference` for holding the data.

rabbitonweb:
@goral09 Oh, I see. So it uses it's own internal implementation, you don't write instance for it?

goral09:
No. Also, won't your example block if you used `MVar` as well?

rabbitonweb:
@goral09 ok, just looked at the code, it has same issues as `Atomic` from monix. It is bsed on `AtomicRefernce` so when you do:

1. `get`
2. some effectful modifications
3. `compareAndSet`

you are responsible to revert the effectful changes. `Atomic` at least informs about it in documentation

rabbitonweb:
@goral09 my typeclass doess NOT have anything else other then `modify`

lukasz-golebiewski:
Can you prove that this code works as expected? Any chance for some tests?

rabbitonweb:
@goral09 I've added the `CellSpec`, maybe this will clarify

rabbitonweb:
@goral09 monix is very clear about this fact in documentation https://monix.io/api/3.0/monix/execution/atomic/Atomic.html#transform(cb:A=%3EA):Unit

```
* Note that the callback will be executed on each iteration of the loop, so it can be called
* multiple times - don't do destructive I/O or operations that mutate global state in it.

```

rabbitonweb:
@goral09 what is not clear in `CellSpec`? how would you implement second test (from that spec), using `Ref` ?

rabbitonweb:
@goral09 or are you asking "why" would I need effects at all while modifing state?

Well, to give you example from our TcpTransportLayer. Imagine your state is map of connections. And you want to modify that map by creating a connection (only if that connection does not exists, so you need to first inspect the content of the map). Creating a connection is an effectful computation.

Details in https://github.com/rchain/rchain/pull/1196/ where you are a reviewer :)

PR #1195, opened 2018-07-26
Framework for Approval protocol

birchmd:
## Overview
This is meant to implement the "bootstrap" half of the protocol described [here](https://rchain.atlassian.net/wiki/spaces/CORE/pages/485556483/Initializing+the+Blockchain+--+Protocol+for+generating+the+Genesis+block). The "Validator" half as well as integration into the node will come in future PRs.

https://rchain.atlassian.net/browse/RHOL-487

### Notes


casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,121 @@
+package coop.rchain.casper.util.comm
```
birchmd: It's occurring to me now that this is not the right package for this and `coop.rchain.casper.genesis` probably makes more sense.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,121 @@
...
+class ApproveBlockProtocol[
+    F[_]: Capture: Monad: NodeDiscovery: TransportLayer: Log: Time: ErrorHandler] private (
+    val candidate: ApprovedBlockCandidate,
+    val start: Long,
+    val duration: Long) {
+  private implicit val logSource: LogSource = LogSource(this.getClass)
+
+  private val u                         = UnapprovedBlock(Some(candidate), start, duration)
+  private val serializedUnapprovedBlock = u.toByteString
+  private val candidateHash             = candidate.block.fold("")(b => PrettyPrinter.buildString(b.blockHash))
```
KentShikama: Shouldn't we just a throw an error if the candidate's blockHash doesn't exist?

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  def run[F[_]: Monad: Capture: Time: ApproveBlockProtocol](waitTime: Long = 5000L): F[Unit] =
+    for {
+      t             <- Time[F].currentMillis
+      maybeApproved <- ApproveBlockProtocol[F].approvedBlock
+      _ <- maybeApproved match {
+            case Some(approvedBlock)
+                if (ApproveBlockProtocol[F].candidate.requiredSigs == 0 || t >= ApproveBlockProtocol[
+                  F].start + ApproveBlockProtocol[F].duration) =>
+              ApproveBlockProtocol[F].sendApprovedBlock(approvedBlock)
```
KentShikama: Can you explain the reasoning behind `t >= ApproveBlockProtocol[F].start + ApproveBlockProtocol[F].duration`?

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,121 @@
...
+  def sendUnapprovedBlock: F[Unit] =
+    for {
+      _ <- Log[F].info(s"APPROVAL: Beginning send of UnapprovedBlock $candidateHash to peers...")
+      _ <- CommUtil.sendToPeers[F](transport.UnapprovedBlock, serializedUnapprovedBlock)
+      _ <- Log[F].info(s"APPROVAL: Sent UnapprovedBlock $candidateHash to peers.")
+    } yield ()
+
+  def sendApprovedBlock(a: ApprovedBlock): F[Unit] = {
+    val serializedApprovedBlock = a.toByteString
+    val hash                    = a.candidate.flatMap(_.block).fold("")(b => PrettyPrinter.buildString(b.blockHash))
```
KentShikama: Should we not just log a warning and then return if the blockhash doesn't exist?

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,121 @@
...
+  private val sigData                   = Blake2b256.hash(candidate.toByteArray)
+
+  private val sigsF = new AtomicMonadState[F, Set[Signature]](AtomicAny(Set.empty[Signature]))
+
+  def addApproval(a: BlockApproval): F[Unit] = {
+    val validSig = for {
+      c   <- a.candidate if c == this.candidate
+      sig <- a.sig if Validate.signature(sigData, sig)
+    } yield sig
+    val sender = a.sig.fold("")(sig => Base16.encode(sig.publicKey.toByteArray))
```
KentShikama: "" could be more descriptive. Maybe "Empty signature"

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
KentShikama: Hmm but we run this protocol occasionally as the chain progresses right?

birchmd: Yeah, true. This also gives a mechanism for creating signed checkpoints in the future.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: We want to wait for the duration to elapse before creating the approved block. This gives an opportunity to collect signatures in excess of the minimum. Say for example there are 10 validators and we only need signatures from 8 to make an approved block. If we get the first 8 signatures within 5 seconds, we could make an approved block right then, but suppose waiting 5 more seconds would give us the remaining 2. Setting a duration of 10-15 seconds would give us all 10 signatures and still not be a terribly long time to wait for the approved block to be created.

KentShikama: So if it hasn't passed that duration you want to call `              ApproveBlockProtocol[F].sendUnapprovedBlock *> IOUtil.sleep(waitTime) *> run[F](
                waitTime)` in an else case like below?

birchmd: That is exactly what gets called. `case _` will match both `None` and `Some(_)` when the `if t >= ...` returns false.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+import coop.rchain.comm.transport.TransportLayer
+import coop.rchain.shared._
+import coop.rchain.catscontrib.{Capture, IOUtil}
+
+import monix.execution.atomic.AtomicAny
+
+import scala.util.Try
+
+class ApproveBlockProtocol[
+    F[_]: Capture: Sync: NodeDiscovery: TransportLayer: Log: Time: ErrorHandler] private (
```
goral09: `Capture` and `Sync` are the same things.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+    val block: BlockMessage,
+    val requiredSigs: Int,
+    val start: Long,
+    val duration: Long,
+    private val sigsF: Ref[F, Set[Signature]]) {
+  private implicit val logSource: LogSource = LogSource(this.getClass)
+
+  val candidate                         = ApprovedBlockCandidate(Some(block), requiredSigs)
```
goral09: If you add `[(scalapb.field).no_box = true]` to field in proto file you won't have them wrapped in `Option`.

goral09: If you want I can add it as a part of the PR.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+      block: BlockMessage,
+      requiredSigs: Int,
+      duration: Long): F[ApproveBlockProtocol[F]] =
+    for {
+      now   <- Time[F].currentMillis
+      sigsF <- Ref.of[F, Set[Signature]](Set.empty[Signature])
+    } yield new ApproveBlockProtocol(block, requiredSigs, now, duration, sigsF)
+
+  def blockApprovalPacketHandler[F[_]: Monad: ApproveBlockProtocol](
+      peer: PeerNode): PartialFunction[Packet, F[Option[Packet]]] =
```
goral09: `peer` is not used. It seems like an oversight?

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+      t             <- Time[F].currentMillis
+      maybeApproved <- ApproveBlockProtocol[F].approvedBlock
+      _ <- maybeApproved match {
+            case Some(_)
+                if (ApproveBlockProtocol[F].candidate.requiredSigs == 0 || t >= ApproveBlockProtocol[
+                  F].start + ApproveBlockProtocol[F].duration) =>
+              ApproveBlockProtocol[F].sendApprovedBlock
+
+            case _ =>
+              ApproveBlockProtocol[F].sendUnapprovedBlock *> IOUtil.sleep(waitTime) *> run[F](
```
goral09: Hmm, I think `IOUtil.sleep(…)` is not a proper way to do sleep. IMO we should use [`cats-effect.Timer#sleep`](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Timer.scala#L154) and [the implementation](https://github.com/typelevel/cats-effect/blob/master/core/jvm/src/main/scala/cats/effect/internals/IOTimer.scala#L47).

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
```diff
@@ -0,0 +1,131 @@
...
+          } yield none[Packet]
+      }
+
+  def run[F[_]: Monad: Capture: Time: ApproveBlockProtocol](waitTime: Long = 5000L): F[Unit] =
+    for {
+      t             <- Time[F].currentMillis
+      maybeApproved <- ApproveBlockProtocol[F].approvedBlock
+      _ <- maybeApproved match {
+            case Some(_)
+                if (ApproveBlockProtocol[F].candidate.requiredSigs == 0 || t >= ApproveBlockProtocol[
```
goral09: Is `ApproveBlockProtocol[F].candidate.requiredSigs == 0` some special case?

KentShikama: Yes. I had the same question originally  too. Probably means we should add a comment saying how it is unnecessary to wait for sigs if we don't require any in the first place.

goral09: @KentShikama What would be an example of the block that requires no signatures?

birchmd: It's for testing. The `requiredSigs == 0` case will function identically to how the node behaves now where it just gets to create the approved block immediately and send it out. In production `requiredSigs` will certainly be positive.

goral09: The other thing is - I thought that this protocol is only for genesis block?

goral09: The other thing is - I thought that this protocol is only for genesis block?

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: Make a PR to change it? I used what was already in `shared` but don't really have strong opinions on this.

goral09: I will, I am just making notes for both you and myself for the future.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: Nope. Not all `PacketHandler`s need to know the peer that the packet came from, but some do, e.g. https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/util/comm/BlockApproverProtocol.scala#L50
For this reason the effect constructor includes the peer https://github.com/rchain/rchain/blob/dev/comm/src/main/scala/coop/rchain/p2p/effects/PacketHandler.scala#L25

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: Oh, neat. Yeah, that sounds useful.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
goral09: But you don't use `peer` here and don't make a use of https://github.com/rchain/rchain/blob/dev/comm/src/main/scala/coop/rchain/p2p/effects/PacketHandler.scala#L25

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: Ah I see, `Sync.delay` is the same as `Capture.capture`. Cool, feel free to remove usage of `Capture` in favour of `Sync` then.

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
KentShikama: It is also for checkpoints along the way under assumption of weak subjectivity. See https://blog.ethereum.org/2014/11/25/proof-stake-learned-love-weak-subjectivity/

casper/src/main/scala/coop/rchain/casper/util/comm/ApproveBlockProtocol.scala
birchmd: Well I hadn't yet, but was planning on it. If you will take a different approach then you can drop the `peer` parameter

PR #1196, opened 2018-07-27
CORE-774/CORE-904 Replace MonadState with Cell in TcpTransportLayer

rabbitonweb:
## Overview
Previously when using `MonadState` we could have race conditions when
creating or removing grpc connections. Example would be two threads
calling `connection` function, creating new ManagedConnection and
storing them to map. During race, only one thread would store its
connection to map, leaving the other with leaking resources.

With Cell we can have thread-safe state modifications while applying
effects (creating connections). This has adventage over using Ref and
its compareAndSet/transform function. While using `Ref` is in general
considered better approach (as it is almost non-blocking), it has its
limits when state modification wants to apply external effect.
Cell typeclass does not have this limitation. Cons of using Cell is
the fact that whole `modify` is a blocking function.

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-774

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
@@ -7,7 +7,7 @@ import coop.rchain.comm.protocol.routing._
 
 import cats._, cats.data._, cats.implicits._
 import coop.rchain.catscontrib._, Catscontrib._, ski._, TaskContrib._
-import coop.rchain.shared.{Log, LogSource}
+import coop.rchain.shared.{Cell, Log, LogSource}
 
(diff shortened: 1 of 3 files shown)


birchmd:
@KentShikama I actually think that we should use `cats.effect.Ref` instead. We don't need the effectful modify provided by `Cell`.

goral09:
@rabbitonweb 
> While using Ref is in general considered better approach (as it is almost non-blocking), it has its
limits when state modification wants to apply external effect.

I got it!

```scala
val cell: Ref[IO, IO[A]] = …
```
which means that the value stored in `Ref` is itself an effect -- in particular it can mean modifying itself :)

PR #1202, opened 2018-07-27
Casper/core 765/integrate lmdb block store

dzajkowski:
## Overview
Introduces LMDB BlockStore to replace InMemBS.
Current implementation does not carry state over.

https://rchain.atlassian.net/browse/CORE-765

### Notes
none


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -181,33 +180,35 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
   implicit val nodeDiscoveryEffect: NodeDiscovery[Task] =
     new KademliaNodeDiscovery[Task](src, defaultTimeout)
 
+  if (Files.notExists(casperPersistencePath)) Files.createDirectories(casperPersistencePath)
+  val config =
+    coop.rchain.blockstorage.LMDBBlockStore.Config(casperPersistencePath, 1024 * 1024 * 1024)
+  val syncEffect: Sync[Effect] = SyncInstances.syncEffect
+  implicit val blockStore: BlockStore[Effect] =
+    LMDBBlockStore.create[Effect](config)(syncEffect, metricsEffect)
```
dzajkowski: not convinced this is a good spot for this init, but did not want to pollute the `Runtime`

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockStore.scala
```diff
@@ -42,11 +42,13 @@ class InMemBlockStore[F[_], E] private ()(implicit
           }
     } yield ()
 
-  private[blockstorage] def clear(): F[Unit] =
+  def clear(): F[Unit] =
     for {
       _ <- metricsF.incrementCounter("block-store-put")
```
lukasz-golebiewski: This metric name doesn't look right

casper/src/test/scala/coop/rchain/casper/CliqueOracleTest.scala
```diff
@@ -24,12 +24,11 @@ import monix.execution.Scheduler.Implicits.global
 
 import scala.collection.immutable.{HashMap, HashSet}
 
-class CliqueOracleTest extends FlatSpec with Matchers with BlockGenerator {
+class CliqueOracleTest extends FlatSpec with Matchers with BlockGenerator with WithBlockStore {
```
lukasz-golebiewski: `with WithBlockStore` sound redundant. `with BlockStoreFixture` maybe?

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -181,33 +180,35 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
   implicit val nodeDiscoveryEffect: NodeDiscovery[Task] =
     new KademliaNodeDiscovery[Task](src, defaultTimeout)
 
+  if (Files.notExists(casperPersistencePath)) Files.createDirectories(casperPersistencePath)
```
birchmd: I would move this directory existence check into the `LMDBBlockStore.create` method itself.

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -256,6 +257,8 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
     resources.runtime.close()
     println("Shutting down Casper runtime ...")
     resources.casperRuntime.close()
+    println("Bringing BlockStore down ...")
```
lukasz-golebiewski: https://www.youtube.com/watch?v=KYh7PwDo3Iw

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -181,33 +180,35 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
   implicit val nodeDiscoveryEffect: NodeDiscovery[Task] =
     new KademliaNodeDiscovery[Task](src, defaultTimeout)
 
+  if (Files.notExists(casperPersistencePath)) Files.createDirectories(casperPersistencePath)
+  val config =
+    coop.rchain.blockstorage.LMDBBlockStore.Config(casperPersistencePath, 1024 * 1024 * 1024)
```
lukasz-golebiewski: `1024 * 1024 * 1024` should be taken from the config

birchmd: The store size should probably also be configurable. I chose to construct the Casper config object from the node config object in [`conf.scala`](https://github.com/rchain/rchain/blob/dev/node/src/main/scala/coop/rchain/node/conf.scala#L247). You could consider doing the same.

node/src/main/scala/coop/rchain/node/node.scala
birchmd: If you make the two changes above then this just becomes a one-liner similar to the other effect instantiations above, so then it should be fine.

casper/src/test/scala/coop/rchain/casper/helper/WithBlockStore.scala
```diff
@@ -0,0 +1,61 @@
...
+          flags: List[EnvFlags] = List(EnvFlags.MDB_NOTLS)): Env[ByteBuffer] =
+    Env
+      .create()
+      .setMapSize(mapSize)
+      .setMaxDbs(8)
+      .setMaxReaders(126)
+      .open(path.toFile, flags: _*)
+
+  def dbDir: Path   = Files.createTempDirectory("casper-block-store-test-")
+  val mapSize: Long = 1024L * 1024L * 4096L
```
birchmd: I doubt the test block store will need to be this big! 100 MB is probably sufficient.

PR #1171, opened 2018-07-24
Core 763/lmdb block storage2

lukasz-golebiewski:
## Overview
Adds LMDB implementation of the Block Storage

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
https://rchain.atlassian.net/browse/CORE-763

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] You have someone in mind to assign for review. Make this assignment after submitting the PR.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1203, opened 2018-07-27
Move Sync[Id] to shared

lukasz-golebiewski:
## Overview
Moves Sync[Id] from block-store to shared so it can be reused as per Kent's request
https://github.com/rchain/rchain/pull/1171#discussion_r205509191

https://rchain.atlassian.net/browse/CORE-763

### Notes

### DIFF
--- a/shared/src/main/scala/coop/rchain/catscontrib/effect/implicits.scala
+++ b/shared/src/main/scala/coop/rchain/catscontrib/effect/implicits.scala
@@ -3,6 +3,7 @@ package coop.rchain.catscontrib.effect
 import cats._
 import cats.implicits._
 import cats.effect.{ExitCase, Sync}
+import scala.util.{Failure, Success, Try}
 
 import scala.util.control.NonFatal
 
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/coop/rchain/catscontrib/effect/implicits.scala
```diff
@@ -0,0 +1,43 @@
...
+
+      def flatMap[A, B](fa: cats.Id[A])(f: A => cats.Id[B]): cats.Id[B] =
+        catsInstancesForId.flatMap(fa)(f)
+
+      def tailRecM[A, B](a: A)(f: A => cats.Id[Either[A, B]]): cats.Id[B] =
+        catsInstancesForId.tailRecM(a)(f)
+
+      def bracketCase[A, B](acquire: A)(use: A => B)(
+          release: (A, ExitCase[Throwable]) => Unit): B = {
+        var maybeErrorCase: Option[ExitCase[Throwable]] = None
```
birchmd: I think you could do this without the `var` using `scala.util.Try` (which only catches `NonFatal` by default)
```scala
Try(use(acquire)) match {
  case Success(result) => 
    release(acquire, ExitCase.Completed)
    result

  case Failure(e) =>
    release(acquire, ExitCase.error(e))
    throw e
}
```

PR #1207, opened 2018-07-27
CORE-911 - Change node.scala so that you can use instances that are provided in effectful way

rabbitonweb:
This solves the issue of creation of `CasperConstructor` and (soon to be merge) `Cell[Connection]`

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-911


rabbitonweb:
Dear reviewers, @dzajkowski  @KentShikama @birchmd ),

`SafetyOracle.scala` seem like to have a lot of changes but those were minimal. Diff is huge becasue somebody forgot to format `SafetyOracle.scala` :) 

With todays PR https://github.com/rchain/rchain/pull/1212 this will never be the case in the future

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -314,13 +301,74 @@ class NodeRuntime(conf: Conf)(implicit scheduler: Scheduler) {
...
+    kademliaRPC = effects.kademliaRPC(src, defaultTimeout)(metricsTask, transport)
+    nodeDiscovery = effects.nodeDiscovery(src, defaultTimeout)(log,
+                                                               time,
+                                                               metricsTask,
+                                                               transport,
+                                                               kademliaRPC)
+    blockStore     = LMDBBlockStore.create[Effect](conf.casperBlockStoreConf)(sync, metrics)
+    _              <- blockStore.clear() // FIX-ME replace with a proper casper init when it's available
+    oracle         = SafetyOracle.turanOracle[Effect](Applicative[Effect], blockStore)
+    runtime        <- Runtime.create(storagePath, storageSize).pure[Effect]
```
birchmd: There is an inconsistency here between `<- ...pure[Effect]` and `=`. Please fix in a future PR.

rabbitonweb: 👍 roger that!

PR #1213, opened 2018-07-30
Enforce sorted input- & expected-result-terms in spatial matching tests

ArturGajowy:
## Overview

Also make the matcher output a sorted term for reminder.

Enforce sorted input- & expected-result-terms in spatial matching tests  …
This includes:
 - rearranging target, pattern, and expected capture terms
   in the spatial matcher's tests (so that they're sorted)
 - fixing two tests that started failing (because previously they were
   enforcing wrong behaviour), by reversing the remainder.
   The tests in question are:
    - "Matching extras with wildcard"
    - "Matching extras with wildcard and free variable"

### JIRA issue:
Discoverd during and needed for [RHOL-533 Use Stable Marriage algo for pattern matching in listMatch](https://rchain.atlassian.net/browse/RHOL-533)

### Checklist

### DIFF
--- a/models/src/main/scala/coop/rchain/models/rholang/sort/Sortable.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sort/Sortable.scala
@@ -1,0 +1,25 @@
+package coop.rchain.models.rholang.sort
+import coop.rchain.models.Expr.ExprInstance
+import coop.rchain.models.Expr.ExprInstance.GBool
+import coop.rchain.models._
+
+trait Sortable[T] {
+  def sortMatch(term: T): ScoredTerm[T]
+}
+
+object Sortable {
+  def apply[T](implicit ev: Sortable[T]) = ev
+
+  implicit val boolSortable: Sortable[GBool]            = BoolSortMatcher
+  implicit val bundleSortable: Sortable[Bundle]         = BundleSortMatcher
+  implicit val channelSortable: Sortable[Channel]       = ChannelSortMatcher
+  implicit val connectiveSortable: Sortable[Connective] = ConnectiveSortMatcher
+  implicit val exprSortable: Sortable[Expr]             = ExprSortMatcher
+  implicit val groundSortable: Sortable[ExprInstance]   = GroundSortMatcher
+  implicit val matchSortable: Sortable[Match]           = MatchSortMatcher
+  implicit val newSortable: Sortable[New]               = NewSortMatcher
+  implicit val parSortable: Sortable[Par]               = ParSortMatcher
+  implicit val receiveSortable: Sortable[Receive]       = ReceiveSortMatcher
+  implicit val sendSortable: Sortable[Send]             = SendSortMatcher
+  implicit val varSortable: Sortable[Var]               = VarSortMatcher
+}
--- a/models/src/main/scala/coop/rchain/models/rholang/sort/BoolSortMatcher.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sort/BoolSortMatcher.scala
@@ -2,7 +2,7 @@ package coop.rchain.models.rholang.sort
 
 import coop.rchain.models.Expr.ExprInstance.GBool
 
-object BoolSortMatcher {
+object BoolSortMatcher extends Sortable[GBool] {
   def sortMatch(g: GBool): ScoredTerm[GBool] =
     if (g.value) {
       ScoredTerm(g, Leaves(Score.BOOL, 0))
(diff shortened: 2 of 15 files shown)


ArturGajowy:
See also the todos in #1201:

https://github.com/rchain/rchain/pull/1201/files#diff-312d2045b0470a95cb7de7187e4fc257R382

https://github.com/rchain/rchain/pull/1201/files#diff-312d2045b0470a95cb7de7187e4fc257R480

This should probably go in before #1201. I've checked that #1201 works with those changes (that is: including the `remainder.reverse` change).

models/src/main/scala/coop/rchain/models/rholang/sort/Sortable.scala
```diff
@@ -0,0 +1,27 @@
+package coop.rchain.models.rholang.sort
+import coop.rchain.models.Expr.ExprInstance
+import coop.rchain.models.Expr.ExprInstance.GBool
+import coop.rchain.models._
+
+trait Sortable[T] {
+  def sortMatch(term: T): ScoredTerm[T]
+
+  implicit val newSrotable: Sortable[New] = NewSortMatcher
```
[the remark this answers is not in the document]
ArturGajowy: human error :) Will push fix in a minute

rholang/src/test/scala/coop/rchain/rholang/interpreter/MatchTest.scala
```diff
@@ -3,61 +3,66 @@ package coop.rchain.rholang.interpreter
...
   import coop.rchain.models.rholang.implicits._
   import SpatialMatcher._
+
+  def assertSpatialMatch[T: Sortable, P: Sortable](
+      target: T,
+      pattern: P,
+      expected: Option[FreeMap])(implicit sm: SpatialMatcher[T, P]): Assertion = {
+    assertSorted(target, "target")
+    assertSorted(pattern, "pattern")
+    expected.foreach(_.values.foreach((v: Par) => assertSorted(v, "expected captured term")))
```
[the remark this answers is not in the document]
ArturGajowy: > Won't two previous assertions already test that?

No, because the expected captures are provided by hand separately, and we can err in there by providing unsorted terms just as well.

In fact, all the changes I've done to tests were guided by one of those three assertions failing and you can see in the diff that some of the expected captures had to change too. If I didn't put this assertion here, those tests would be enforcing unsorted output for sorted input.

See matcher test line 65

models/src/main/scala/coop/rchain/models/rholang/sort/Sortable.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.models.rholang.sort
+import coop.rchain.models.Expr.ExprInstance
+import coop.rchain.models.Expr.ExprInstance.GBool
+import coop.rchain.models._
+
+trait Sortable[T] {
```
goral09: I still don't think we need this type class. You can test that terms are sorted separately from `assertSpatialMatch`. 

```scala
def assertSorted(par: Par): Assertion = ParSortMatcher.sortMatch(par) == par

test("some test") {
assertSorted(target)
assertSorted(pattern)
assertSpatialMatch(target, pattern, expectedResult)
```

I don't like the fact that we are creating a type class only for the tests.

ArturGajowy: Most of `target`-s and `pattern`s aren't Pars, so I'd have to pattern match over types not having a thing in common (on type level), and I'd have to put each one of those into a Par in a different way (a Send to sends field, an Expr to expressions field, etc...) and sort the Par, or just call the relevant `*SortMatcher.sortMatch` - which the typeclass does for me in a way better, cleaner and less error-prone way than both the other approches. 

Additionally, the `Sortable` typeclass captures and solidifies in types a very strong convention we have. I'd dare say we already have that typeclass in `dev`, just on convention level and can't benefit from it fully due to that. E.g.:
 - the compiler doesn't understand that convention and doesn't help
 - there's not a way to do sth for anything sortable, like I need to do above and wanted many times during debugging so far
 - newcomers have to discover the convention, while a typeclass would make things obvious
 - without that typeclass, it's easy to break the convention, because no types guard it

I'm sure replacing all the FooSortMatch.sortMatch with Sortable.sortMatch would make the code more readable, especially to those not knowing of the SortMatcher convention/typeclass yet (I've been that person a month ago). I plan to do that in a later PR.

ArturGajowy: And since #1145 is ready to merge, you go to dev first and I'll update my PR :)

PR #1218, opened 2018-07-31
RHOL-463: Aggregate transaction fees of the deploy.

goral09:
## Overview
Accumulates cost of Rholang programs within single deploy.

https://rchain.atlassian.net/browse/RHOL-463

### Notes


goral09:
@birchmd  @KentShikama  PTAL at https://github.com/rchain/rchain/pull/1218/files especially the last commit which adds cost of the deployment into the block  message. Ignore failing tests - I haven't updated them yet.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -51,9 +51,10 @@ object Genesis {
                     initial: BlockMessage,
                     startHash: StateHash,
                     runtimeManager: RuntimeManager)(implicit scheduler: Scheduler): BlockMessage = {
-    val Right(checkpoint) = runtimeManager.computeState(startHash, blessedTerms)
-    val stateHash         = ByteString.copyFrom(checkpoint.root.bytes.toArray)
-    val reductionLog      = checkpoint.log.map(EventConverter.toCasperEvent)
+    //TODO: ignore `deployCost` of blessed contracts?
```
birchmd: Yeah, the genesis block is created for free to get the network going. So we can track the cost, but it doesn't really mean anything per se.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -66,9 +66,10 @@ object InterpreterUtil {
       internalMap: Map[BlockHash, BlockMessage],
       emptyStateHash: StateHash,
       knownStateHashes: Set[StateHash],
-      computeState: (StateHash, Seq[Deploy]) => Either[DeployError, Checkpoint])
-    : (Checkpoint, Set[StateHash]) = {
-    val (postStateHash, updatedStateHashes) =
+      computeState: (StateHash,
+                     Seq[Deploy]) => Either[DeployError, (Checkpoint, Vector[DeployCost])])
+    : (Checkpoint, Set[StateHash], Vector[DeployCost]) = {
+    val (postStateHash, updatedStateHashes, postStateCost) = //TODO: what to do with the postStateCost -- ignore?
```
birchmd: This is a good question. Ideally it does not actually take any work to deduce the parent post state because the parents need to be independent, so you should be able to just "merge" the tuplepaces. Of course we don't have any way of doing that right now, so we just repeat the calculations on top of each other, which is potentially an intensive task. Ultimately though we do not want it to cost extra to produce multiparent blocks because they are good for the network since they enable a fork-join parallelism model in the blockchain itself (multiparent blocks are a join bringing to previously concurrent forks together).

For now the easiest thing to do is to probably just ignore this `postStateCost` and we should have a TODO item to revisit the way we merge parent states.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -136,11 +136,17 @@ message Header {
   bytes extraBytes = 8;
 }
 
+message DeployCost {
+    bytes deployHash = 1; //Sha256 hash of the DeployString
```
birchmd: When I said tuple I meant the actual `Deploy` proto. This is fine too, but we have been using `Blake2b256` hashing for everything, so we had better stick with that here as well.

goral09: I can change to `Blake2b256`. I didn't want to duplicate the data. If you prefer the actual `Deploy` that's no problem.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -2,9 +2,9 @@ package coop.rchain.casper.util.rholang
 
 import com.google.protobuf.ByteString
 import coop.rchain.casper.util.ProtoUtil
-import coop.rchain.casper.protocol.{Bond, Deploy, DeployString}
+import coop.rchain.casper.protocol.{Bond, Deploy, DeployCost, DeployString}
 import coop.rchain.catscontrib.TaskContrib._
-import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.crypto.hash.{Blake2b256, Blake2b512Random, Sha256}
```
birchmd: No longer a need to import hash functions.

casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
```diff
@@ -250,9 +251,71 @@ class InterpreterUtilTest
...
+    accCostBatch should contain theSameElementsAs (accCostsSep)
+  }
+
+  it should "return cost of deploying even if one of the programs withing the deployment throws an error" in {
+    //reference costs
+    //deploy each Rholang program separately and record its cost
+    val deploy1 = ProtoUtil.termDeploy(mkTerm("@1!(Nil)").toOption.get)
+    val deploy2 = ProtoUtil.termDeploy(mkTerm("@2!([1,2,3,4])").toOption.get)
+
+    def computeSingleDeployCost(deploy: Deploy*): Vector[DeployCost] = {
```
birchmd: This appears to be the same function you defined on L261. Maybe factor out into a function you can call in both places?

goral09:
@birchmd PTAL.

casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
```diff
@@ -250,9 +263,75 @@ class InterpreterUtilTest
...
+    accCostBatch should contain theSameElementsAs (accCostsSep)
+  }
+
+  it should "return cost of deploying even if one of the programs withing the deployment throws an error" in {
+    pendingUntilFixed { //reference costs
+      //deploy each Rholang program separately and record its cost
+      val deploy1 = ProtoUtil.termDeploy(mkTerm("@1!(Nil)").toOption.get)
+      val deploy2 = ProtoUtil.termDeploy(mkTerm("@2!([1,2,3,4])").toOption.get)
+
+      def computeSingleDeployCost(deploy: Deploy*): Vector[DeployCost] = {
```
birchmd: This function seems to be the same as the one on L274. Perhaps you could factor it out and call the same function in both places.

PR #1226, opened 2018-07-31
CORE-796: Channels hashes in trace log

ys-pyrofex:
## Overview
We need channels hashes to allow subscribing to updates on a channel name.

https://rchain.atlassian.net/browse/CORE-796


rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
```diff
@@ -43,7 +43,7 @@ object COMM {
 
 sealed trait IOEvent extends Event
 
-class Produce private (val hash: Blake2b256Hash) extends IOEvent {
+class Produce private (val channelsHash: Blake2b256Hash, val hash: Blake2b256Hash) extends IOEvent {
```
dzajkowski: curious: couldn't this be a `case class`?

rspace/src/test/scala/coop/rchain/rspace/StorageActionsTests.scala
```diff
@@ -1387,6 +1387,55 @@ trait StorageActionsTests
     check(prop)
   }
 
+  "produce and consume" should "store channel hashes" in withTestSpace { space =>
```
dzajkowski: wouldn't it make sense to extend existing tests (eg. `consume twice then createCheckpoint`)? they have the benefit of being generative.

casper/src/main/scala/coop/rchain/casper/util/EventConverter.scala
```diff
@@ -13,34 +13,39 @@ import coop.rchain.rspace.trace.{
...
+
+  implicit def blake2b256HashToByteString(hash: Blake2b256Hash): ByteString =
+    ByteString.copyFrom(hash.bytes.toArray)
+
   def toCasperEvent(event: RspaceEvent): Event = event match {
-    case RspaceProduce(produce) =>
-      Event(Produce(ProduceEvent(ByteString.copyFrom(produce.bytes.toArray))))
-    case RspaceConsume(consume) =>
-      Event(Consume(ConsumeEvent(ByteString.copyFrom(consume.bytes.toArray))))
+    case produce: RspaceProduce =>
```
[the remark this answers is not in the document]
ys-pyrofex: Just wanted to make it seamless with lines below, where we use more complicated patterns. Since compiler/casper teams work with this code much more often - your opinion has top priority. Please let me know if you prefer this style here.

rspace/src/test/scala/coop/rchain/rspace/StorageActionsTests.scala
ys-pyrofex: I initially implemented these checks as part of another test, but the result wasn't look well - a mess of checks that verify non-related functionality. To make it more readable and follow single responsibility principle decided to create a separate test.

casper/src/main/scala/coop/rchain/casper/util/EventConverter.scala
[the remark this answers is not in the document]
goral09: For me it's fine as it is.

rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
ys-pyrofex: I don't know why, but historically Produce and Consume were not case classes, while COMM was. Definitely all of them are good candidates to be case classes. Thanks for pointing this out.

PR #1233, opened 2018-08-01
Dev new urn request resource

guardbotmk3:
## Overview
Implements ability to request a specific resource via new, rather than just create a new one.

Solves the non-ambient part of
https://rchain.atlassian.net/browse/RHOL-499


ArturGajowy:
The build failed due to a bad merge, here's a patch for that:

```
Index: rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala	(revision 48f9bf98a9d1c6a5530adb216526a556c4f8f28c)
+++ rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala	(revision 68b68e6b4201c3c0aad7d3b367f2a1d86f5751fa)
@@ -9,12 +9,12 @@
 import coop.rchain.models._
 import coop.rchain.models.rholang.sort.Sortable
 import coop.rchain.rholang.interpreter.accounting.CostAccount
+import coop.rchain.rholang.interpreter.matcher.SpatialMatcher.spatialMatch
 import org.scalatest._
 
 import scala.collection.immutable.BitSet
 
 class VarMatcherSpec extends FlatSpec with Matchers {
-  import SpatialMatcher._
 
   def assertSpatialMatch[T: Sortable, P: Sortable](
@@ -236,8 +236,8 @@
   "Matching inside bundles" should "not be possible" in {
     val target: Bundle = Bundle(
       Par()
-        .prepend(Send(Quote(GPrivateBuilder("0")), Seq(GInt(43)), persistent = false)))
-        .prepend(Send(Quote(GInt(7)), Seq(GInt(42)), persistent = false))
+        .prepend(Send(Quote(GPrivateBuilder("0")), Seq(GInt(43)), persistent = false))
+        .prepend(Send(Quote(GInt(7)), Seq(GInt(42)), persistent = false)))
     val pattern: Bundle = Bundle(
       Par()
         .prepend(Send(Quote(GInt(7)), Seq(EVar(FreeVar(0))), persistent = false))
@@ -257,8 +257,8 @@
   it should "be possible to match on entire bundle" in {
     val target: Bundle = Bundle(
       Par()
-        .prepend(Send(Quote(GPrivateBuilder("0")), Seq(GInt(43)), persistent = false)))
-        .prepend(Send(Quote(GInt(7)), Seq(GInt(42)), persistent = false))
+        .prepend(Send(Quote(GPrivateBuilder("0")), Seq(GInt(43)), persistent = false))
+        .prepend(Send(Quote(GInt(7)), Seq(GInt(42)), persistent = false)))
 
     val pattern: Par = EVar(FreeVar(0))
 
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -389,14 +390,44 @@ object Reduce {
...
+            case Some(p) => Right(newEnv.put(p))
+            case None    => Left(ReduceError(s"Unknown urn for new: ${urn}"))
+          }
+        // This is non-functional, but it's a self-contained implementation of a
+        // functional early terminating left fold, and the library does the same
+        // thing with vars.
+        def foldLeftEarly[A, B, E](as: Seq[A],
+                                   init: B,
+                                   cata: (A, B) => Either[E, B]): Either[E, B] = {
+          var result: Either[E, B] = Right(init)
```
ArturGajowy: I think the body of this method can be replaced with:
`Foldable[List].foldM[Either[E, ?], A, B](as.toList, init)((x, y) => cata(y, x))`

Please replace it and remove the comment on the code being non-functional :)

We could even use the foldM directly, but then implicit resolution starts fussing and we end up in the 'apologizing' part of FP in Scala.

After we have `SeqInstances.traverseSeq` available from a util module, we can drop the `.toList` - please add a TODO comment for that?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -389,14 +390,44 @@ object Reduce {
...
+        }
+        foldLeftEarly(urns, simpleNews, addUrn) match {
+          case Right(env) => Applicative[M].pure(env)
+          case Left(e)    => s.raiseError(e)
+        }
+      }
 
       costAccountingAlg.charge(newBindingsCost(neu.bindCount)) *>
-        eval(neu.p)(alloc(neu.bindCount), rand)
+        alloc(neu.bindCount, neu.uri).flatMap { newEnv =>
```
ArturGajowy: please add a test for this commit, the production code in here is not secured against regresions

rholang/examples/hello_world_again.rho
```diff
@@ -1,9 +1,9 @@
 // A simple HelloWorld contract demonstrating a contract being rerun
 
-new helloworld in {
+new helloworld, stdout(`rho:io:stdout`) in {
```
ArturGajowy: Please also update rholangtut.md

Also please consider all text occurrences of @"stdout", @"stdoutAck", etc. - I think there's some default testing rholang hardcoded in the p2p test tool, for example...

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -614,11 +614,58 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
         Send(ChanVar(BoundVar(2)), List[Par](GInt(7)), false, BitSet(2))
           .prepend(Send(ChanVar(BoundVar(1)), List[Par](GInt(8)), false, BitSet(1)))
           .prepend(Send(ChanVar(BoundVar(0)), List[Par](GInt(9)), false, BitSet(0))),
+        Vector.empty,
         BitSet()
       )))
     result.knownFree should be(inputs.knownFree)
   }
 
+  "PNew" should "Sort URI's and place them at the end" in {
```
goral09: I don't follow -- we're sorting the uris in lexicographical order in the `normalize.scala` instead of `NameSortMatcher` ?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: One more reason scala sucks: I would have never expected foldM to short circuit.

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
guardbotmk3: We can't sort them in the sorter, than the variable numbering would be messed up.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -124,28 +125,35 @@ object Runtime {
       9L -> SystemProcesses.secp256k1Verify(space, dispatcher)
     )
 
+    def byteName(b: Byte): Par = GPrivate(ByteString.copyFrom(Array[Byte](b)))
+
+    val urnMap: Map[String, Par] = Map("rho:io:stdout" -> byteName(0),
+                                       "rho:io:stdoutAck" -> byteName(1),
+                                       "rho:io:stderr"    -> byteName(2),
+                                       "rho:io:stderrAck" -> byteName(3))
```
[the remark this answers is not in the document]
guardbotmk3: We may do that eventually, but it's less urgent because they don't convey extra authority.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -427,14 +428,33 @@ object Reduce {
...
           _env.put(addr)
         }
+        def addUrn(newEnv: Env[Par], urn: String): Either[ReduceError, Env[Par]] =
+          urnMap.get(urn) match {
+            case Some(p) => Right(newEnv.put(p))
+            case None    => Left(ReduceError(s"Unknown urn for new: ${urn}"))
+          }
+        // This is non-functional, but it's a self-contained implementation of a
+        // functional early terminating left fold, and the library does the same
+        // thing with vars.
```
ArturGajowy: You forgot to delete this comment :)

PR #1234, opened 2018-08-01
Rholang: Substitute the target of toByteArray

guardbotmk3:
Users will expect a fully sorted and substituted term so that they can do
reliable signature checks. Thanks to dckc for finding this.

Fixes 
https://rchain.atlassian.net/projects/RHOL/issues/RHOL-564


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -799,8 +799,9 @@ object Reduce {
         } else {
           for {
             exprEvaled <- evalExpr(p)
-            _          <- costAccountingAlg.charge(toByteArrayCost(exprEvaled))
-            ba         <- s.fromEither(serialize(exprEvaled))
+            exprSubst  <- substituteAndCharge[Par, M](exprEvaled, 0, env, costAccountingAlg)
```
ArturGajowy: 1. Don't we want all method arguments (including "this") be fully substituted before call/execution? Or do we want lazy semantics wherever possible?

2. Do we have in plans some form of method call protocol, and API for it? I'd imagine defining a method should be as easy as *declaring* the signature (including parameter strictness) and body. I'm not saying we should do this in this PR, just asking a question.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1014,6 +1014,41 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+    val splitRand = rand.splitByte(0)
+    val unsubProc: Par =
+      New(bindCount = 1, p = EVar(BoundVar(1)), locallyFree = BitSet(0))
+    val subProc: Par =
+      New(bindCount = 1, p = GPrivate("zero"), locallyFree = BitSet())
+    val serializedProcess         = subProc.toByteString
+    val toByteArrayCall: Par      = EMethod("toByteArray", unsubProc, List[Par]())
+    def wrapWithSend(p: Par): Par = Send(Quote(GString("result")), List[Par](p), false, BitSet())
+
+    val result = withTestSpace { space =>
```
ArturGajowy: I guess it'd be much easier to write such a test by writing inline rholang in a string literal and asserting execution results. I'd gladly work on infrastructure for writing such tests.

ArturGajowy: We could even easily convert selected tests to 'rholang source' style by simply pretty-printing the input source. So it'd not only be a long term investment - it'd give immediate returns in terms of test code readability.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1014,6 +1014,41 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+    implicit val costAccounting =
+      CostAccountingAlg.unsafe[Task](CostAccount.zero)
+    val splitRand = rand.splitByte(0)
+    val unsubProc: Par =
+      New(bindCount = 1, p = EVar(BoundVar(1)), locallyFree = BitSet(0))
+    val subProc: Par =
+      New(bindCount = 1, p = GPrivate("zero"), locallyFree = BitSet())
+    val serializedProcess         = subProc.toByteString
+    val toByteArrayCall: Par      = EMethod("toByteArray", unsubProc, List[Par]())
+    def wrapWithSend(p: Par): Par = Send(Quote(GString("result")), List[Par](p), false, BitSet())
```
goral09: isn't `locallyFree` set a union of `channel`'s `locallyFree` and `locallyFree` of the body?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1014,6 +1014,41 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+
+    val result = withTestSpace { space =>
+      val reducer     = RholangOnlyDispatcher.create[Task, Task.Par](space).reducer
+      val env         = Env.makeEnv[Par](GPrivate("one"), GPrivate("zero"))
+      val task        = reducer.eval(wrapWithSend(toByteArrayCall))(env, splitRand, costAccounting)
+      val inspectTask = for { _ <- task } yield space.store.toMap
+      Await.result(inspectTask.runAsync, 3.seconds)
+    }
+    val channel = Channel(Quote(GString("result")))
```
goral09: Extract `Quote(GSTring("result"))` to variable and use it in https://github.com/rchain/rchain/pull/1234/files?utf8=%E2%9C%93&diff=unified#diff-aa1f7851ad04305bebc084b34b25e16aR1028 .

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: 1. No. Most methods should be fine without fully substituted terms.
2. I don't have any plans to do so.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
guardbotmk3: Good catch.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
guardbotmk3: I've said this before:
No.
That would literally break the point of these tests, which is to test some portion of the interpreter **without** depending on the other parts.

ArturGajowy: I remember you said this. I'm bringing this because it seemed to me you asked if we have sth similar in one of the discord channels a couple days ago. I guess it was about sth else / for a different purpose then.

PR #1247, opened 2018-08-03
Artifacts from CORE-637 (RNode windows native build)

ys-pyrofex:
## Overview
We will be glad if RNode will run on windows (either natively or inside WSL).


ys-pyrofex:
I think for windows we can check if standard windows path shortcuts are supported (and add supports, if they're not). In this case default windows configs will have paths like "%AppData%/Roaming/Rchain/Data", and cover 90% of kick-start cases.

PR #1248, opened 2018-08-03
createBlock gRPC interface also adds the block to the state

birchmd:
## Overview
It is inefficient to have the block passed around internally to cross the gRPC boundary.

As per discussion on discord earlier today,

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -37,26 +37,23 @@ object BlockAPI {
         DeployServiceResponse(false, s"Error: Casper instance not available"))
   }
 
+  def addBlock[F[_]: Monad: MultiParentCasperConstructor: Log](
+      b: BlockMessage): F[DeployServiceResponse] =
+    MultiParentCasperConstructor.withCasper[F, DeployServiceResponse](
+      casper =>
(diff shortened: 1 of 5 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -37,33 +37,28 @@ object BlockAPI {
...
+            case Some(_: InvalidBlock) =>
+              DeployServiceResponse(false, s"Failure! Invalid block: $status")
+            case Some(_: ValidBlock) =>
+              DeployServiceResponse(true, s"Success! Block created and added.")
+            case Some(BlockException(ex)) =>
+              DeployServiceResponse(false, s"Error during block processing: $ex")
+            case Some(Processing) =>
+              DeployServiceResponse(
+                false,
+                "No action taken since other thread if already processing the block.")
```
dckc: typo? should be "other thread _is_ already..."?

dckc:
How do we get the block we just created?

How about returning the block hash so we can get it with `showBlock`?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: +1

PR #1249, opened 2018-08-03
CORE-956 Added 'find' functionality to BlockStore. Moved InMemBlockStore to tests. Import cleanup in Casper.

dzajkowski:
## Overview
A `find by predicate` method added to BlockStore.

https://rchain.atlassian.net/projects/CORE/issues/CORE-956

### Notes
none

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
@@ -15,6 +15,8 @@ trait BlockStore[F[_]] {
 
   def get(blockHash: BlockHash): F[Option[BlockMessage]]
 
+  def find(p: BlockHash => Boolean): F[Seq[(BlockHash, BlockMessage)]]
+
   def put(f: => (BlockHash, BlockMessage)): F[Unit]
 
   def apply(blockHash: BlockHash)(implicit applicativeF: Applicative[F]): F[BlockMessage] =
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/LMDBBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/LMDBBlockStore.scala
@@ -90,7 +90,6 @@ class LMDBBlockStore[F[_]] private (val env: Env[ByteBuffer], path: Path, blocks
       ret <- withReadTxn { txn =>
               blocks
                 .iterate(txn)
-                .iterable()
                 .asScala
                 .withFilter(kv => p(ByteString.copyFrom(kv.key())))
                 .map { kv =>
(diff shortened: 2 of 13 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/LMDBBlockStore.scala
```diff
@@ -84,6 +84,25 @@ class LMDBBlockStore[F[_]] private (val env: Env[ByteBuffer], path: Path, blocks
             }
     } yield ret
 
+  override def find(p: BlockHash => Boolean): F[Seq[(BlockHash, BlockMessage)]] =
+    for {
+      _ <- metricsF.incrementCounter(MetricNamePrefix + "find")
+      ret <- withReadTxn { txn =>
+              blocks
+                .iterate(txn)
+                .iterable()
```
birchmd: Why turn into an `Iterable` here? I'm not too sure about the LMDB API overall, but I would wait until after the filter is applied.

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
```diff
@@ -106,6 +106,23 @@ trait BlockStoreTest
...
+      forAll(blockStoreElementsGen, minSize(0), sizeRange(10)) { blockStoreElements =>
+        val items = blockStoreElements
+        items.foreach(store.put(_))
+        items.foreach {
+          case (k, v) =>
+            val w = store.find(_ == ByteString.copyFrom(k.getBytes()))
+            w should have size 1
+            w.head._2 shouldBe v
+        }
+        store.asMap().size shouldEqual items.size
```
lukasz-golebiewski: Should we use deprecated methods in tests?

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
```diff
@@ -15,6 +15,8 @@ trait BlockStore[F[_]] {
 
   def get(blockHash: BlockHash): F[Option[BlockMessage]]
 
+  def find(p: BlockHash => Boolean): F[Seq[(BlockHash, BlockMessage)]]
```
lukasz-golebiewski: Is it possible for the result to contain a hash that was not already supplied as parameter? If not - why return it?

dzajkowski: I would say that it's deprecated for external use. after it's not needed I would make it package private and let it be for tests. or we can try to replace it with `find`. I'm fine with that also.

block-storage/src/test/scala/coop/rchain/blockstorage/InMemBlockStore.scala
```diff
@@ -30,6 +30,12 @@ class InMemBlockStore[F[_], E] private ()(implicit
       state <- refF.get
     } yield state
 
+  override def find(p: BlockHash => Boolean): F[Seq[(BlockHash, BlockMessage)]] =
+    for {
+      _     <- metricsF.incrementCounter("block-store-find")
+      state <- refF.get
+    } yield state.toSeq.filter { case (k, _) => p(k) }
```
birchmd: I suspect that `state.filterKeys(p).toSeq` would be a little more efficient.

PR #1253, opened 2018-08-05
Include non-body BlockMessage fields into blockhash

odeac:
## Overview
Justifications, sender, seqNum, sigAlgorithm need to be included in the hash such that they cannot be faked

https://rchain.atlassian.net/browse/RHOL-546

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -190,15 +190,14 @@ object ProtoUtil {
         acc.updated(validator, block)
     }
 
-  def protoHash[A <: { def toByteArray: Array[Byte] }](proto: A): ByteString =
-    ByteString.copyFrom(Blake2b256.hash(proto.toByteArray))
+  def protoHash[A <: { def toByteArray: Array[Byte] }](protoSeq: A*): ByteString =
+    protoSeqHash(protoSeq)
 
-  def protoSeqHash[A <: { def toByteArray: Array[Byte] }](protoSeq: Seq[A]): ByteString = {
-    val bytes = protoSeq.foldLeft(Array.empty[Byte]) {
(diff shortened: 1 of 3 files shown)


casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -215,29 +214,51 @@ object ProtoUtil {
...
       .withJustifications(justifications)
+  }
+
+  def hashUnsignedBlock(header: Header, justifications: Seq[Justification]) = {
+    val items = header.toByteArray +: justifications.map(_.toByteArray)
+    hashByteArrays(items: _*)
+  }
+
+  def hashSignedBlock(header: Header, sender: ByteString, sigAlgorithm: String, seqNum: Int) =
+    hashByteArrays(header.toByteArray, sender.toByteArray, sigAlgorithm.getBytes, Array(seqNum.toByte))
```
KentShikama: Is `Array(seqNum.toByte)` Scala specific? Mabye use protobuf's toByteArray instead? We want to allow for non-Scala implementations to be able to reproduce this block hashing process.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -215,29 +214,61 @@ object ProtoUtil {
...
+
     BlockMessage()
-      .withBlockHash(protoHash(header))
+      .withBlockHash(hash)
       .withHeader(header)
       .withBody(body)
       .withJustifications(justifications)
+  }
+
+  def int32ToByteArray(i: Int): Array[Byte] = {
```
birchmd: I'm a little worried about this because our block hashes will now be tied to the way the JVM represents integers, whereas previously everything was based on protobuf serialization. Perhaps it is worth considering making a `BlockHashData` protobuf with fields for everything that needs to be included in the hash, then simply calling hashing the result of `toByteArray`.

PR #1255, opened 2018-08-06
RHOL-483: Add shard identifier to blocks

itegulov:
## Overview
Adds the shard identifier to BlockMessage and BlockInfo

https://rchain.atlassian.net/browse/RHOL-483


node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -45,6 +45,7 @@ object Configuration {
     .parse("rnode://de6eed5d00cf080fc587eeb412cb31a75fd10358@52.119.8.109:40400")
     .right
     .get
+  private val DefaultShardId = "rchain"
```
birchmd: I'm not sure if I agree with this as the default. I think `root` might be better. Perhaps @pyrocto would like to suggest something as well.

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
```diff
@@ -11,7 +11,8 @@ import coop.rchain.shared.{Log, LogSource}
 
 case class ValidatorIdentity(publicKey: Array[Byte],
                              privateKey: Array[Byte],
-                             sigAlgorithm: String) {
+                             sigAlgorithm: String,
+                             shardId: String) {
```
birchmd: I think conceptually the `shardId` does not belong here because read-only nodes (which do not have a validator ID) will still connect to a particular shard. Instead maybe it should be an argument passed to the `Casper` object itself, like the genesis block.

KentShikama:
@birchmd Yeah I agree a block should be invalid if it does not have the shardId corresponding to the Casper instance.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -172,7 +173,8 @@ message BlockMessage {
   int32                  seqNum         = 6; // number of blocks created by the validator
   bytes                  sig            = 7; // signature generated by signing `hash(hash(justification) concat blockHash)`.
   string                 sigAlgorithm   = 8; // name of the algorithm used to sign
-  bytes extraBytes = 9;
+  string                 shardId        = 9; // identifier of the shard where the block was created
```
birchmd: I'm second guessing now whether this belongs here or in the `Header`... It's probably fine here, but maybe logically it belongs in the `Header` or perhaps in `RChainState`...

PR #1259, opened 2018-08-06
[RHOL-541] Remove term from deploy proto

AbnerZheng:
## Overview
Remove term from deploy proto

https://rchain.atlassian.net/browse/RHOL-541

### Notes


AbnerZheng:
@KentShikama PTAL.  
I don't know how to realize the the pre-compiled feature of CompiledRholangSource. Could you please give me some suggestion?

birchmd:
From what I understand the spirit of this PR is to make sure that the source code of the deploy matches the term which is given. If that is the case then I think a better approach would be to simply make the `deploy` method of `Casper` take a `DeployData` and construct the `Deploy` object itself. Totally eliminating the `term` field from the `Deploy` is a bad idea in my opinion because turning source code into a `Par` is a non-trivial amount of work and so we should only do it once. Moreover, an error could result from it (because of bad syntax) so we want to fail fast and have that error caught right away when someone tries to deploy.

AbnerZheng:
@birchmd Yes, the spirit of this PR is to make sure that the source code of the deploy matches the term which is given. I have replaced `deploy` method of `Casper` take a DeployData and construct the `Deploy` object itself as you said.  I think only source deploy is allowed except for genesis, am I right? If I am right, for test case in `RholangBuildTest`, where we wanna support importing of rholang sources into scala code, we should provide the source code in class `CompiledRholangSource`.

rholang-proto-build/src/main/scala/coop/rchain/rholang/build/RholangProtoBuild.scala
```diff
@@ -30,17 +30,22 @@ object RholangProtoBuild {
       case Right(term) => term.writeTo(new FileOutputStream(output.toFile))
     }
 
-  def createScalaArtifact(pkg: Option[String], name: String, proto: Path): String = {
+  def escapeSourceCode(source: Path): String =
+    Source.fromFile(source.toFile).getLines().mkString("\"\"\"\n", "\n|", "\"\"\".stripMargin")
```
birchmd: I think the `\n` in the first argument to `mkString` is not needed.
```
scala> def test(source: String): String = Source.fromFile(source).getLines().mkString("\"\"\"", "\n|", "\"\"\".stripMargin")
test: (source: String)String

scala>  val x = test("./rholang/examples/tut-hello.rho")
x: String =
"""new helloWorld, stdout(`rho:io:stdout`), stdoutAck(`rho:io:stdoutAck`) in {
|  contract helloWorld(@name) = {
|    new ack in {
|      stdoutAck!("Hello, ", *ack) |
|      for (_ <- ack) {
|        stdoutAck!(name, *ack) |
|        for (_ <- ack) {
|          stdout!("\n")
|        }
|      }
|    }
|  } |
|  helloWorld!("Joe")
|}""".stripMargin
```

AbnerZheng: Done. @birchmd  Could you please help me merge this PR?

birchmd: Do you still not have permissions to merge? If not then I will merge on your behalf after the merge conflicts have been resolved.

PR #1260, opened 2018-08-06
CORE-770/CORE-947 remove connections in rchain protocol when near max size

rabbitonweb:
## Overview
RChain protocol connections can not grow infinitely. It needs to be cleared periodically.
Feature is implemented but not yet turned on.

https://rchain.atlassian.net/secure/RapidBoard.jspa?rapidView=7&projectKey=CORE&modal=detail&selectedIssue=CORE-947

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
@@ -1,7 +1,6 @@
 package coop.rchain.comm.rp
 
 import coop.rchain.p2p.effects._
-
 import coop.rchain.comm.discovery._
 import scala.concurrent.duration._
 import com.google.protobuf.any.{Any => AnyProto}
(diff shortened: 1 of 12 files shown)


comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
```diff
@@ -13,24 +12,59 @@ import coop.rchain.metrics.Metrics
...
+      for {
+        numOfConnectionsPinged <- RPConfAsk[F].reader(_.clearConnections.numOfConnectionsPinged)
+        toPing                 = connections.take(numOfConnectionsPinged)
+        local                  <- TransportLayer[F].local
+        timeout                <- RPConfAsk[F].reader(_.defaultTimeout)
+        hb                     = heartbeat(local)
+        results <- toPing.traverse(peer =>
+                    TransportLayer[F].roundTrip(peer, hb, timeout).map(r => (peer, r)))
+        successfulPeers = results.filter(_._2.isRight).map(_._1)
+        failedPeers     = results.filter(_._2.isLeft).map(_._1)
```
sebtomba: Please use `collect` (twice) for better readability

PR #1263, opened 2018-08-07
Fix untraced comm event error

birchmd:
## Overview
Allows multiple deploys without crashing. The bug was actually caused by the deploys being stored in the block in reverse order due to the cost accounting accumulator prepending instead of appending. The fix was therefore very simple. I also added a unit test for the bug to prove it works.

https://rchain.atlassian.net/browse/RHOL-568

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -162,7 +162,7 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: 
             val otherErrors = errorLog.readAndClearErrorVector()
             val cost        = CostAccount.toProto(costAlg.getCost().unsafeRunSync)
             val deployCost  = DeployCost().withDeploy(deploy).withCost(cost)
-            Left((deploy, ex +: otherErrors, accCost :+ deployCost))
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -155,14 +155,14 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
...
+              eval(rest, reducer, errorLog, costAlg, accCost :+ deployCost)
             else
-              Left((deploy, errors, accCost))
+              Left((deploy, errors, accCost :+ deployCost))
           case Failure(ex) =>
             val otherErrors = errorLog.readAndClearErrorVector()
             val cost        = CostAccount.toProto(costAlg.getCost().unsafeRunSync)
             val deployCost  = DeployCost().withDeploy(deploy).withCost(cost)
-            Left((deploy, ex +: otherErrors, deployCost +: accCost))
+            Left((deploy, ex +: otherErrors, accCost :+ deployCost))
```
goral09: Shouldn't the errors be reversed as well?

birchmd: Seems less critical because they all apply to the same deploy. But sure, we can append there too.

PR #1264, opened 2018-08-07
CORE-770 - Granade

rabbitonweb:
## Overview
While working on CORE-770 some changes were necessary and worth extracting them to separate PR. This is mainly refactoring work where:
1. Remove local from TransportLayer and move it to RPConf
2. Add factory method constructing KademliaNodeDiscovery within Effect

CORE-770


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -36,6 +37,9 @@ class NodeRuntime(conf: Configuration, host: String)(implicit scheduler: Schedul
 
   private implicit val logSource: LogSource = LogSource(this.getClass)
 
+  implicit def eiterTrpConfAsk(implicit ev: RPConfAsk[Task]): RPConfAsk[Effect] =
```
ArturGajowy: typo

PR #1265, opened 2018-08-08
RHOL-464: Cross shard transfers

pyrocto:
## Overview
Rholang contracts for transferring money between shards

RHOL-464

Sorry it's a little long.  I tried to comment heavily.


rholang/src/main/rholang/cross-shard.rho
```diff
@@ -0,0 +1,524 @@
...
+              statusOut!(["fail", "Already done."])
+            } else {
+              // check that the pubkey is in the list
+              if (not pubKeySet.has(pubKey)) {
+                canceledSetStore!(canceledSet) |
+                doneStore!(done) |
+                statusOut!(["fail", "Invalid key."])
+              } else {
+                // check that it hasn't already cast its vote to cancel
+                if (canceledSet.has(pubKey)) {
```
goral09: Do we have `has` method? Did you mean `contains` ?

pyrocto: Yes, thanks.

pyrocto:
By the way, I fully expect that this code will need more tweaking and refinement once we get the off-chain stuff working with it, but this should express the general form.  I also ended up not using the lockout stuff, so if you'd like, I can remove that.  I think the pattern's a useful one, so it should end up in a repo of code somewhere.

The registry syntax won't work until Kyle gets the registry implemented, and most of these "classes" should be deployed separately but ask for the latest compatible version of the others from the registry.

rholang/src/main/rholang/cross-shard.rho
```diff
@@ -0,0 +1,538 @@
...
+              new splitPurseOut in {
+                // Create new wallet, transfer funds
+                @[purse, "split"]!(balance, *splitPurseOut) |
+                for (@[] <- splitPurseOut) {
+                  statusOut!(["failure", "Could not deposit funds."])
+                } |
+                // Throw away the resulting purse.
+                for (@[_] <- splitPurseOut) {
+                  if (balance < depositFee) {
+                    statusOut!(["failure", "Not enough funds to pay the fee."])
```
birchmd: I don't really like that the money gets destroyed no matter what. It would be more user friendly to check if the balance was sufficient first.

dckc:
FYI: some testing in progress. Maybe 1% coverage so far. I'm struggling to get a top-side view of what the happy path should be.

https://github.com/dckc/rchain/tree/cross-shard/rholang/src/test/rholang 770d655

rholang/src/main/rholang/cross-shard.rho
```diff
@@ -0,0 +1,538 @@
...
+                // check that it hasn't already cast its vote to cancel
+                if (canceledSet.contains(pubKey)) {
+                  canceledSetStore!(canceledSet) |
+                  doneStore!(done) |
+                  statusOut!(["fail", "Already canceled."])
+                } else {
+                  // check the signature
+                  new ret in {
+                    verify!([*ballot, *statusOut].toByteArray(), sig, pubKey, *ret) |
+                    for (@verifyResult <- ret) {
```
odeac: A pipe operator would be nice to have. With a pipe operator, the construct 

```
new ret in {
   x ! (params, ret)
   for (@result <- ret) {
      computation(result)
   }
}
```

...would become `verify ! (params) |> computation`

Jake-Gillberg:
Is this deprecated? Or is the intent to still use this work for the shards of [RHOL-440](https://rchain.atlassian.net/browse/RHOL-440)?

dzajkowski:
Closing as its not mergeable (easily).
Will keep it for reference.

PR #1270, opened 2018-08-09
Either rholang contract and tests

birchmd:
****
This should allow nicer error handling involving user input. If all the validation/transformation functions can be composed together in one line then we should be able to avoid many nested if statements. This will probably make parts of the sharding code and PoS code easier to read, but a simpler example using the composition is below.
```
contract @"divide"(@x, @y, ret) = {
  if(y == 0) { ret!(("Left", "Div by zero!")) }
  else { ret!(("Right", x / y)) }
} |
contract @"sqrt"(@x, ret) = {
  match x {
    case 1 => ret!(("Right", 1))
    case 4 => ret!(("Right", 2))
    case 9 => ret!(("Right", 3))
    case _ => ret!(("Left", "Unknown square root"))
  }
} |
for(@x, ret <- @"div2"){ @"divide"(x, 2, *ret) } |
@["Either", "compose"]!(18, ["div2", "sqrt"], *stdout)
```

https://rchain.atlassian.net/browse/RHOL-560

### Notes

### DIFF
--- a/casper/src/main/rholang/Either.rho
+++ b/casper/src/main/rholang/Either.rho
@@ -1,0 +1,47 @@
+//scalapackage coop.rchain.rholang.collection
+
+//Requires LinkedList
+//Represents Either[A, B] as ("Left", A) or ("Right", B)
+
+//Right-biased flatMap; f must also return an Either
+contract @["Either", "flatMap"](@either, f, return) = {
+  match either {
+    ("Right", value) => { f!(value, *return) }
+    ("Left", _)      => { return!(either) }
+  }
+} |
+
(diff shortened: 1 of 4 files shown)


guardbotmk3 (approved):
I would be happier if we were removing LinkedList, rather than adding dependencies to it.

birchmd:
I'm happy to change this when we have a native fold over lists 😉

dckc:
Either / Right / Left are pretty strange names for error handling. Please consider Result / Ok / Err as in https://doc.rust-lang.org/std/result/

PR #1272, opened 2018-08-09
Maximum bipartite matching in listMatch

ArturGajowy:
## Overview

Reimplements matching of List[Pattern] with List[Term], making it deterministic and polynomial-time, according to the [Maximum Bipartite Matching algorithm](http://olympiad.cs.uct.ac.za/presentations/camp2_2017/bipartitematching-robin.pdf) (see: "alternate version" in the slides)

Supersedes #1201 due to a counterexample we've missed initially.

### JIRA issue:
https://rchain.atlassian.net/browse/RHOL-533

### Checklist

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/MaximumBipartiteMatch.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/MaximumBipartiteMatch.scala
@@ -1,0 +1,108 @@
+package coop.rchain.rholang.interpreter.matcher
+
+import cats._
+import cats.data.StateT
+import cats.implicits._
+
+object MaximumBipartiteMatch {
+  def apply[P, T, F[_]: Monad](matchFun: (P, T) => F[Boolean]): MaximumBipartiteMatch[P, T, F] = {
+    val fM = implicitly[Monad[F]]
+    new MaximumBipartiteMatch[P, T, F] {
+      private[matcher] override implicit val fMonad: Monad[F] = fM
+      private[matcher] override val matchFunction             = matchFun
+    }
+  }
+}
+
+  /**
+  * Implements the MaximumBipartiteMatch algorithm, according to:
+  * http://olympiad.cs.uct.ac.za/presentations/camp2_2017/bipartitematching-robin.pdf
+  * (mainly the "Alternative approach" section)
+  * <p/>
+  * All the effects produced by calling the provided `matchFunction`
+  * are going to be retained in the final effect via which the algo returns.
+  * See type signatures for `machFunction` and `findMatches`.
+  *
+  * @tparam P type of pattern / the U set
+  * @tparam T type of term / the V set
+  * @tparam F the target Monadic effect this algorithm is to be embedded into
+  */
+trait MaximumBipartiteMatch[P, T, F[_]] {
+
+  private[matcher] implicit val fMonad: Monad[F]
+  private[matcher] val matchFunction: (P, T) => F[Boolean]
+
+  private case class S(matches: Map[T, Pattern], seenTargets: Set[T])
+  type Pattern    = (P, Candidates)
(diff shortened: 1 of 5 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -264,119 +248,130 @@ object SpatialMatcher extends SpatialMatcherInstances {
...
+              Matching a single variable using that mechanism (e.g. [1] matches [x]) was OK so far, since
+              overriding the match state (FreeMap entry) for a given variable never happened in the previous algo.
+
+              We don't have a proof it was impossible. What we know for sure though is:
+              none of the tests added for the new algo - which revealed this problem - have failed on the old algo.
+
+              Since we've started modifying the var assignment during the MBM algorithm, the case of matching a var
+              to a single term must be handled separately, without merging the prior match state.
+              */
+              if (patternsSorted.isEmpty && remainderTargets.length == 1 &&
```
ArturGajowy: @guardbotmk3 This needs your special attention and feedback. I'm not entirely sure if the condition used here is right or why the issue wouldn't surface earlier.

I tried getting rid of the `noFrees` in expressions matching and match the free variables in the expression matcher, but then thought we should discuss if it's the right approach.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -264,119 +248,130 @@ object SpatialMatcher extends SpatialMatcherInstances {
...
+        guard(t == p).modifyCost(_.charge(COMPARISON_COST))
+      } else {
+        spatialMatch(t, p)
       }
-    }
+      matchEffect.attempt.map(_.isRight)
+    })
+
+    for {
+      matchedTargetsOpt <- maximumBipartiteMatch.findMatches(patternsSorted.toList)
```
guardbotmk3: There's no way that this is correct for remainders anyway.
For wildcards you're fine, but for remainders you need #(patterns) - #(targets) pseudo patterns that only match if locallyFree is empty.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
guardbotmk3: > Since we've started modifying the var assignment during the MBM algorithm, the case of matching a var
> to a single term must be handled separately, without merging the prior match state.

You just explained why this is wrong:
> we've started modifying the var assignment during the MBM algorithm

This **should not** be the case. For each (target, pattern) pair you test, you should find the update that results when they are matched, and you should apply the updates from **only** those (target, pattern) pairs in the final match

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/MaximumBipartiteMatch.scala
```diff
@@ -0,0 +1,108 @@
...
+  * See type signatures for `machFunction` and `findMatches`.
+  *
+  * @tparam P type of pattern / the U set
+  * @tparam T type of term / the V set
+  * @tparam F the target Monadic effect this algorithm is to be embedded into
+  */
+trait MaximumBipartiteMatch[P, T, F[_]] {
+
+  private[matcher] implicit val fMonad: Monad[F]
+  private[matcher] val matchFunction: (P, T) => F[Boolean]
```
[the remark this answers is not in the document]
ArturGajowy: I tried to leave that part of the effect stack out of the algo. Wouldn't the match being executed in outer effects be enough for the free-variable capture to happen? Notice that all the tests are passing.

ArturGajowy:
Superseded by #1332

PR #1273, opened 2018-08-09
And connectives to match simple ground types.

guardbotmk3:
## Overview
Adds connectives to match simple ground types.

Couldn't find Jira Link. Will edit when I find it.


guardbotmk3:
OK, examples added to tutorial.

PR #1274, opened 2018-08-10
Add listenForName GRPC endpoint

KentShikama:
## Overview

https://rchain.atlassian.net/browse/RHOL-441

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -12,6 +12,8 @@ import coop.rchain.casper.util.rholang.InterpreterUtil
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.shared.Log
 
+import monix.execution.Scheduler
+
 object BlockAPI {
 
   def deploy[F[_]: Monad: MultiParentCasperConstructor: Log](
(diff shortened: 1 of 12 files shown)


casper/src/test/scala/coop/rchain/casper/api/ListeningNameAPITest.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  private val (validatorKeys, validators) = (1 to 4).map(_ => Ed25519.newKeyPair).unzip
+  private val genesis                     = createGenesis(validators)
+
+  "getListeningNameResponse" should "return listening names across a chain" in {
+    val stake                 = 10
+    val equalBonds            = validators.zipWithIndex.map { case (v, _) => v -> stake }.toMap
+    val genesisWithEqualBonds = buildGenesis(equalBonds)
+    val nodes                 = HashSetCasperTestNode.network(validatorKeys.take(3), genesisWithEqualBonds)
+    implicit val nodeZeroConstructor = MultiParentCasperConstructor
```
birchmd: The node must already have an instance of this. Why create another one instead of accessing the effect like you do for the safety oracle and logs?

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -17,6 +17,7 @@ service DeployService {
   rpc createBlock(google.protobuf.Empty) returns (DeployServiceResponse) {}
   rpc showBlock(BlockQuery) returns (BlockQueryResponse) {}
   rpc showBlocks(google.protobuf.Empty) returns (BlocksResponse) {}
+  rpc listenForName(ListeningNameQuery) returns (ListeningNameResponse) {}
```
birchmd: I think it would be better to take a `Channel` directly, saves you one unnecessary level of `Option`. Also, is `Channel` the right type? That also includes `Var`s, which I don't think we want to be allowed to query for. What if this took a `Par` instead and constructed the `Channel` instance from it?

casper/src/test/scala/coop/rchain/casper/api/ListeningNameAPITest.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  import HashSetCasperTest._
+
+  private val (validatorKeys, validators) = (1 to 4).map(_ => Ed25519.newKeyPair).unzip
+  private val genesis                     = createGenesis(validators)
+
+  "getListeningNameResponse" should "return listening names across a chain" in {
+    val stake                 = 10
+    val equalBonds            = validators.zipWithIndex.map { case (v, _) => v -> stake }.toMap
+    val genesisWithEqualBonds = buildGenesis(equalBonds)
```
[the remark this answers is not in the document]
KentShikama: Copy paste error from a similar test in HashSetCasper

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -44,19 +52,70 @@ object BlockAPI {
...
+                                        block.body.fold(Seq.empty[Event])(_.commReductions)
+                                      val log =
+                                        serializedLog.map(EventConverter.toRspaceEvent).toList
+                                      val listeningNameReduced = log.exists {
+                                        case Produce(channelHash, _) =>
+                                          channelHash == StableHashProvider.hash(
+                                            immutable.Seq(listeningName))
+                                        case Consume(channelHash, _) =>
+                                          channelHash == StableHashProvider.hash(
+                                            immutable.Seq(listeningName))
```
birchmd: Per @guardbotmk3 We should sort the term before we go searching for it in the tuplespace since there is no guarantee that the proto someone gives us is already sorted.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -4,10 +4,11 @@ import cats._
 import com.google.protobuf.empty.Empty
 import coop.rchain.blockstorage.BlockStore
 import coop.rchain.casper.api.BlockAPI
-import coop.rchain.casper.protocol.{DeployServiceGrpc, DeployServiceResponse, DeployData, _}
+import coop.rchain.casper.protocol.{DeployData, DeployServiceGrpc, DeployServiceResponse, _}
```
birchmd: Nit pick, but doesn't having the `_` at the end mean we don't need the other things anyway?

KentShikama:
@birchmd PTAL

dckc:
Any particular reason why this is polling when gRPC supports streaming?

PR #1275, opened 2018-08-10
Initial bonding contract

birchmd:
## Overview
Moves us towards having bonding working. I'll have to write tests and integrate with Genesis block in another PR.

https://rchain.atlassian.net/browse/RHOL-560

### Notes


guardbotmk3 (dismissed):
Looks OK for me. I think the rate limiting may have to be exogenous.

casper/src/main/rholang/MakePoS.rho
```diff
@@ -0,0 +1,111 @@
...
+              if (isBonded) { ret!(("Left", "Public key ${pk} already bonded." %% {"pk" : publicKey})) }
+              else          { ret!(("Right", x)) }
+            }
+          }
+        } |
+        clonePurse!(givenBondPurse, *bondPurseCh) |
+        for(@cloneResult <- bondPurseCh) {
+          match cloneResult {
+            ("Right", (bondPurse, bondAmount)) => {
+              @["Either", "compose"]!(bondPurse, [
```
KentShikama: This is the same thing we do to traverse the Validate logic right? If so, consider renaming to @["Either", "traverse"]

casper/src/main/rholang/MakePoS.rho
```diff
@@ -0,0 +1,111 @@
+contract @"MakePoS"(@purse, @minimumBond, @maximumBond, @initBonds, return) = {
```
KentShikama: So this `return` channel must be public so that anyone can submit a deploy that uses `this` right?

casper/src/main/rholang/MakePoS.rho
```diff
@@ -0,0 +1,111 @@
...
+  new this, bondsCh, clonePurse, depositPurse in {
+    bondsCh!(initBonds) |
+  
+    //The purpose of this contract is to move the contents
+    //of an existing purse into a new purse. This prevents
+    //the owner of the given purse from modifying the balance
+    //during processing.
```
KentShikama: Is this specifically because we don't process deploy sequentially?

casper/src/main/rholang/MakePoS.rho
```diff
@@ -0,0 +1,111 @@
...
+    contract clonePurse(@givenPurse, return) = {
+      new balanceCh, resultCh in {
+        @[givenPurse, "getBalance"]!(*balanceCh) |
+        for(@balance <- balanceCh) {
+          @[givenPurse, "split"]!(balance, *resultCh) |
+          for(@result <- resultCh) {
+            match result {
+              [] => { return!(("Left", "Could not extract funds from given purse")) }
+              [newPurse] => { return!(("Right", (newPurse, balance))) }
```
KentShikama: Do we not have to check that the newPurse is from the `thisMint` of the Rev mint?

casper/src/main/rholang/MakePoS.rho
```diff
@@ -0,0 +1,111 @@
...
+      }
+    } |
+    
+    contract @[*this, "getBonds"](return) = {
+      for(@bonds <- bondsCh) {
+        bondsCh!(bonds) | return!(bonds)
+      }
+    } |
+    
+    contract @[*this, "bond"](@publicKey, @sigVerify, @givenBondPurse, @rewardsForwarder, return) = {
```
KentShikama: Same with `givenBondPurse`

casper/src/main/rholang/MakePoS.rho
birchmd: It's not really a `traverse`. It's more like a `for` comprehension or `do` notation in Haskell, but less powerful because you cannot bind new variables and reference them later. Instead it just pipes the output of one function into the input of the next, terminating on a `Left`. So `compose` seemed like the best name because it is just fancy function composition.

casper/src/main/rholang/MakePoS.rho
birchmd: It'll be registered with the name registry to ensure users can send messages to it, but not listen on the same name, intercepting its messages.

casper/src/main/rholang/MakePoS.rho
birchmd: It's so that we don't check that the purse has the right balance at one step only later to see that the purse balance has changed. Basically, we move the authority from the user to the contract so that know once we verify the purse that it will not be changed suddenly.

casper/src/main/rholang/MakePoS.rho
birchmd: That is a good point. A PoS should be tied to a particular `mint`. We would check the given purse and then just know that the new purse is of the same mint because that is how purses work.

casper/src/main/rholang/MakePoS.rho
KentShikama: >  the purse has the right balance at one step only later to see that the purse balance has changed

So you're saying "yes" to my question right? If we processed deploys sequentially, then there is no way something can come in between the previous and next steps.

casper/src/main/rholang/MakePoS.rho
KentShikama: Sounds good

casper/src/main/rholang/MakePoS.rho
KentShikama: I see yeah it is like a 

```
for {
   ... <- ...traverse(...)
   ... <- ...traverse(...)
   ... <- ...traverse(...)
} yield ...
```

not a traverse itself.

birchmd:
@KentShikama PTAL I have added much more to this PR since you last looked it over. I still don't verify the mint of the purse and just left it as a TODO item until we get the fix on the Mint side of things.

KentShikama (approved):

PR #1276, opened 2018-08-10
CORE-908 InMemoryTrieStore + lmdb-independent InMemoryStore

ys-pyrofex:
## Overview
We need an in-memory implementation for 1) RSpace performance check 2) Possibility to run on WSL until mmap will be fixed 3) Testing 4) etc.

https://rchain.atlassian.net/browse/CORE-908

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
@@ -1,0 +1,117 @@
+package coop.rchain.rspace.history
+import coop.rchain.rspace.{Blake2b256Hash, InMemTransaction, InMemoryOps}
+import kamon.Kamon
+
+import scala.collection.immutable.Seq
+
+case class State[K, V](_dbTrie: Map[Blake2b256Hash, Trie[K, V]],
+                       _dbRoot: Map[Branch, Blake2b256Hash],
+                       _dbPastRoots: Map[Branch, Seq[Blake2b256Hash]]) {
+
+  def changeTrie(newTrie: Map[Blake2b256Hash, Trie[K, V]]): State[K, V] =
+    State(newTrie, _dbRoot, _dbPastRoots)
+
+  def changeRoot(newRoot: Map[Branch, Blake2b256Hash]): State[K, V] =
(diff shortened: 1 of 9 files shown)


rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
```diff
@@ -0,0 +1,117 @@
...
+                       _dbRoot: Map[Branch, Blake2b256Hash],
+                       _dbPastRoots: Map[Branch, Seq[Blake2b256Hash]]) {
+
+  def changeTrie(newTrie: Map[Blake2b256Hash, Trie[K, V]]): State[K, V] =
+    State(newTrie, _dbRoot, _dbPastRoots)
+
+  def changeRoot(newRoot: Map[Branch, Blake2b256Hash]): State[K, V] =
+    State(_dbTrie, newRoot, _dbPastRoots)
+
+  def chnagePastRoot(newPastRoots: Map[Branch, Seq[Blake2b256Hash]]): State[K, V] =
```
dzajkowski: typo

rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
```diff
@@ -0,0 +1,117 @@
...
+      }
+      .orElse {
+        getAllPastRoots(txn)
+          .find(_ == hash)
+          .map { blake: Blake2b256Hash =>
+            putRoot(txn, branch, blake)
+            blake
+          }
+      }
+      .getOrElse(throw new Exception(s"Unknown root."))
```
dzajkowski: I understand the need to keep the implementation close, but don't you think this could be just the last `else`? It's all in memory so it should be good enough.

rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
ys-pyrofex: Fixed

rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
ys-pyrofex: Sure, updated

PR #1282, opened 2018-08-13
CORE-908 Version of Node that uses InMemory{Store,TrieStore}

ys-pyrofex:
## Overview
InMemory RSpace will be useful for performance testing and possible launch of test node on systems where LMDBStore isn't yet supported (WSL on Windows for example)

https://rchain.atlassian.net/browse/CORE-908


rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -344,9 +344,17 @@ object ReplayRSpace {
     implicit val codecA: Codec[A] = sa.toCodec
     implicit val codecK: Codec[K] = sk.toCodec
 
-    val mainStore = LMDBStore.create[C, P, A, K](context, branch)
-
-    val replaySpace = new ReplayRSpace[C, P, A, R, K](mainStore, branch)
+    val (mainStore, replaySpace) = context match {
```
lukasz-golebiewski: The code duplication below is unnecessary

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
```diff
@@ -261,7 +261,7 @@ object InMemoryStore {
     implicit val codecC: Codec[C] = sc.toCodec
     implicit val codecP: Codec[P] = sp.toCodec
     implicit val codecA: Codec[A] = sa.toCodec
-    initialize(trieStore, branch)
+
```
lukasz-golebiewski: Why is this not needed anymore?

docs/windows/rnode.toml
```diff
@@ -10,6 +10,7 @@ bootstrap = "rnode://c61769b39d368cbcbc9499634e030386c79d5b02@52.119.8.108:40400
 standalone = true
 data-dir = "C:/RChain/data"
 map-size = 200000000
+in-memory-store = true
```
lukasz-golebiewski: Why is true default for Windows?

dzajkowski:
Looks good enough to me, please also ask @rabbitonweb for node changes.

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
ys-pyrofex: With latest update InMemoryStore implementation is closer to LMDBStore, which doesn't need this call as well.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
ys-pyrofex: You're right, fixed.

docs/windows/rnode.toml
ys-pyrofex: Mistakenly left from my local benchmarks. Set to false.

docs/windows/rnode.toml
```diff
@@ -10,6 +10,7 @@ bootstrap = "rnode://c61769b39d368cbcbc9499634e030386c79d5b02@52.119.8.108:40400
 standalone = true
 data-dir = "C:/RChain/data"
 map-size = 200000000
+in-memory-store = false
```
rabbitonweb: this configuration file is for windows, should it not be set to true?

rspace/src/main/scala/coop/rchain/rspace/Context.scala
```diff
@@ -3,23 +3,35 @@ package coop.rchain.rspace
...
   def close(): Unit = {
     trieStore.close()
     env.close()
   }
 }
 
+class InMemoryContext[C, P, A, K] private[rspace] (
+    val trieStore: ITrieStore[InMemTransaction[history.State[Blake2b256Hash, GNAT[C, P, A, K]]],
+                              Blake2b256Hash,
+                              GNAT[C, P, A, K]]
```
rabbitonweb: "GNAT C.P.A.K" sounds like a NY rap group :)

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -189,10 +189,16 @@ object RSpace {
       sc: Serialize[C],
       sp: Serialize[P],
       sa: Serialize[A],
-      sk: Serialize[K]): RSpace[C, P, A, R, K] = {
-    val mainStore = LMDBStore.create[C, P, A, K](context, branch)
-    create(mainStore, branch)
-  }
+      sk: Serialize[K]): RSpace[C, P, A, R, K] =
+    context match {
```
rabbitonweb: Just so I understand this correctly (sorry if asking dumb question. We are given here a `context` as parameter and we create one again?

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -344,7 +344,13 @@ object ReplayRSpace {
     implicit val codecA: Codec[A] = sa.toCodec
     implicit val codecK: Codec[K] = sk.toCodec
 
-    val mainStore = LMDBStore.create[C, P, A, K](context, branch)
+    val mainStore = context match {
```
rabbitonweb: same question as above

docs/windows/rnode.toml
ys-pyrofex: On Windows (not WSL) LMDB works fine. The only problem we have now - absent native DLLs in our port of kalium. In memory store should be default for WSL, but we don't have a sample config for WSL yet.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
ys-pyrofex: Replaced "match" with context.createStore method.

PR #1289, opened 2018-08-14
 Track log computed when merging multiple parents

birchmd:
## Overview
The untraced comm event reported in RHOL-591 likely occurred because the fork-choice rule created a multi-parent block in which the parents' post-state was non-trivial (because it included the comm events from the respective parent blocks). The reason this produced an untraced comm event is because, prior to this change, the log reported in the block did not include anything about computing the parents' post-state, therefore when the parents' post-state was computed in replay mode all comm events were untraced. This change makes sure we track the log for the parents' post-state computation and includes it in the block.

https://rchain.atlassian.net/browse/RHOL-591

### Notes


MParlikar:
I tested the fix off of @birchmd's branch on stress-docker with multiple nodes deploying and proposing.  RHOL-591 appears to be fixed as a result of this change.

PR #1299, opened 2018-08-16
serialize private key as even number of hex digits

odeac:
## Overview
Make sure the number of digits are always even because if it's odd the call to Base16.decode fails.

https://rchain.atlassian.net/browse/CORE-981


crypto/src/main/scala/coop/rchain/crypto/signatures/Secp256k1.scala
```diff
@@ -16,6 +15,8 @@ object Secp256k1 {
   private val provider  = new BouncyCastleProvider()
   private val curveName = "secp256k1"
 
+  private def toHex(bytes: Array[Byte]) = bytes.map(byte => "%02X".format(byte)).mkString
```
odeac: We could move this to some utilities file. Where would be a proper location for such a function?

crypto/src/main/scala/coop/rchain/crypto/signatures/Secp256k1.scala
```diff
@@ -37,7 +38,9 @@ object Secp256k1 {
     kpg.initialize(new ECGenParameterSpec(curveName), secureRandomInstance)
     val kp = kpg.generateKeyPair
 
-    val sec = Base16.decode(kp.getPrivate.asInstanceOf[ECPrivateKey].getS().toString(16))
+    val base16Sec = toHex(kp.getPrivate.asInstanceOf[ECPrivateKey].getS.toByteArray)
```
birchmd: No, this is not going to work. `toByteArray` returns a representation of the number which we do not support, see the [scaladocs](https://www.scala-lang.org/api/current/scala/math/BigInt.html#toByteArray:Array[Byte]). We need to give the number's literal value in base 16 and simply add a leading zero if the number of hex-digits is odd.

crypto/src/main/scala/coop/rchain/crypto/signatures/Secp256k1.scala
```diff
@@ -15,7 +14,7 @@ object Secp256k1 {
 
   private val provider  = new BouncyCastleProvider()
   private val curveName = "secp256k1"
-
+
```
birchmd: Additional whitespace?

PR #1304, opened 2018-08-16
CORE-887 resource exhausted

rabbitonweb:
## Overview
The problem orbits around the problem that grpc-java has a limit on maximum size of the message that is being sent and received. In our project grpc-java is used in two places:
1. as an underlaying transport layer that allows sending messages between nodes
2. as an API server

Originally the issue (CORE-887) was about `show-blocks`, thus manifesting itself in the latter (API server). When calling `show-blocks` on a node, the node would send the description of the *whole* blockchain (all blocks). At some point that description would exceed the maximum message size and the server would throw an exception.

Solution to this problem was to send that information in chunks, stream fashion.

While working on this problem, I've realized that there is a bigger issue here at hand - in the transport layer. Blocks themselves can exceed the maximum message size and so making them to big to be sent via transport layer. It is possible to deploy a contract that once proposed will generate enormous amount of comm events. Example of such contract below:

```
new dupe, stdout(`rho:io:stdout`), stdoutAck(`rho:io:stdoutAck`) in {
contract dupe(@depth) = {
  if (depth <= 0) { Nil } else {  dupe!(depth-1) | dupe!(depth-1) | dupe!(depth-1) | dupe!(depth-1) |  dupe!(depth-1) | dupe!(depth-1) | dupe!(depth-1) | dupe!(depth-1) | dupe!(depth-1) }
} |
dupe!(5)
}
```

This generates block of size 8 MB, that is bigger then the default value of maximum message sized allowed by grpc-java (4MB). 
For the testnet (in this PR) we address the issue by extending the maximum message size to be 100MB (and a value that can be configured).
For main-net (not in this PR) we must consider chunking up blocks when sending them between nodes, if the block is bigger then the maximum message size.

In this PR I did the following:
1. created `maxMessageSize` that configures both grpc-java (the transport layer and server api), its default value is 100 MB. Value is hardcoded, we might consider adding a flag in configuration for it.
2. switched grpc-java in server api to explicitly use the Netty server (because only on netty server you can explicitly set that value
3. I've created new version of `show-blocks` that sends them in a streaming fashion to the caller

CORE-887

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
@@ -46,7 +46,7 @@ class GrpcDeployService(host: String, port: Int) extends DeployService[Task] wit
   }
 
   def showBlocks(): Task[String] = Task.delay {
-    val response = blockingStub.showBlocks2(Empty()).toList
+    val response = blockingStub.showBlocks(Empty()).toList
 
     val showResponses = response
       .map {
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
@@ -16,7 +16,6 @@ trait DeployService[F[_]] {
   def createBlock(): F[(Boolean, String)] //create block and add to Casper internal state
   def showBlock(q: BlockQuery): F[String]
   def showBlocks(): F[String]
-  def showBlocks2(): F[String]
   def addBlock(b: BlockMessage): F[(Boolean, String)]
 }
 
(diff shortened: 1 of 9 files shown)


casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -46,8 +46,24 @@ class GrpcDeployService(host: String, port: Int) extends DeployService[Task] wit
...
   def showBlocks(): Task[String] = Task.delay {
-    val response = blockingStub.showBlocks(Empty())
-    response.toProtoString
+    val response = blockingStub.showBlocks(Empty()).toList
+
+    val showResponses = response
+      .map {
+        case bi =>
+          s"""
+------------- block ${bi.blockNumber} ---------------
```
birchmd: This might look nicer if you indented and used the `stripMargin` function

PR #1320, opened 2018-08-20
Limit the amount of threads in scheduler.

dzajkowski:
## Overview
Solve: `java.lang.OutOfMemoryError: unable to create new native thread`

https://rchain.atlassian.net/projects/CORE/issues/CORE-884

### Notes
none

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -1,13 +1,12 @@
 package coop.rchain.node
 
 import java.security.Security
+import java.util.concurrent.{SynchronousQueue, ThreadPoolExecutor, TimeUnit}
 
 import scala.collection.JavaConverters._
 import scala.tools.jline.console._
 import completer.StringsCompleter
-
 import cats.implicits._
-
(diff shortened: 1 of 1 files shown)


goral09:
Is this really a fix? From `monix` docs:
```
    * - the maximum number of threads is unbounded, as recommended for blocking I/O
```
isn't this just hiding the fact that we threads don't yield and the thread pool will kill them?

lukasz-golebiewski:
It is a fix @goral09, before this rnode would crash and now it can execute the contract to the very end.
We need to revisit scheduling in the app, as stated in the TODO

dzajkowski:
@goral09 it is. casper is using this scheduler. as long as it is unbounded there is no hope to finish a large computation. with a limit a contract will finish. this fix does not address issues with threads not being released it addresses how many are being spawned.

pyrocto:
On discord, Pawel said:  I've thought about it but I don't belive you [can] model a contract that [deadlocks]. Each thread eventually enters a lmdb, tries to pattern match, but when it fails to match, it will store data there and leave the lock.

PR #1324, opened 2018-08-20
Add two retries (10 sec each) when closing grpc clients

rabbitonweb:
## Overview
When grpc clients will not have time to properly close connection, they freak out with

```
SEVERE: ~ Channel io.grpc.internal.ManagedChannelImpl-5 for target grpc:30015 was not shutdown properly!!! ~
```

 I've added extended the time to await termination, also added two retries. Should be more then enough. Maybe reconsider the close functionality to try more then twice? Worth considering.

CORE-943

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
@@ -71,5 +71,12 @@ Blockchain length: ${response.length}
     (response.success, response.message)
   }
 
-  override def close(): Unit = channel.shutdown().awaitTermination(3, TimeUnit.SECONDS)
+  override def close(): Unit = {
+    val terminated = channel.shutdown().awaitTermination(10, TimeUnit.SECONDS)
+    if (!terminated) {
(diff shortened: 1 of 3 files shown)


casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -71,5 +71,12 @@ Blockchain length: ${response.length}
     (response.success, response.message)
   }
 
-  override def close(): Unit = channel.shutdown().awaitTermination(3, TimeUnit.SECONDS)
+  override def close(): Unit = {
+    val terminated = channel.shutdown().awaitTermination(10, TimeUnit.SECONDS)
+    if (!terminated) {
+      println(
+        "warn: did not shutdown after 10 seconds, retrbying with additional 10 seconds timeout")
```
AndyLiu0429: typo: retrbying -> retrying

rabbitonweb: thx!

PR #1327, opened 2018-08-20
Add the beginnings of a registry implementation.

guardbotmk3:
## Overview
This adds the beginnings of a registry. Only lookup is implemented, but there's a test and it works.

ref https://rchain.atlassian.net/browse/RHOL-499


rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+import coop.rchain.models.Var.VarInstance.FreeVar
+import coop.rchain.models._
+import coop.rchain.models.rholang.implicits._
+import coop.rchain.rholang.interpreter.storage.implicits._
+import coop.rchain.rspace.{ISpace, IStore}
+import monix.eval.Task
+import scala.collection.immutable.Seq
+import scala.collection.{Seq => RootSeq}
+
+class Registry(private val space: ISpace[Channel,
```
ArturGajowy: Re commit message:
`This is just a patricia trie with byte fanout. At the moment only lookup is
implemented.`

I really can't see the places where it's a patricia tree, maybe b/c of only lookup being implemented :) I've got to take that for a ride and try to understand/restructure the AST work in here. I think that's what's preventing me from seeing the tree structure.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  val lookupCallbackRef: Long = 11L
+  val prefixRetReplacePattern = BindPattern(
+    Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true)),
+        ChanVar(FreeVar(1)),
+        ChanVar(FreeVar(2))),
+    freeCount = 3)
+  val triePattern = BindPattern(
+    Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true))),
+    freeCount = 1)
```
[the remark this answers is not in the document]
ArturGajowy: Adding a comment with the relevant rholang would be a minimum - this way the reader could assume the comment to be true for understanding, and later verify it being true for being sure. However brittle that is, especially when compared to using the parser for building the AST.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+  val testingDispatchTable: Map[Long, Function1[RootSeq[ListChannelWithRandom], Task[Unit]]] =
+    Map(lookupRef -> lookup, lookupCallbackRef -> lookupCallback)
+}
+
+object Registry {
+  val registryRoot = GPrivate(
+    ByteString.copyFrom(
+      Base16.decode("a4fd447dedfc960485983ee817632cf36d79f45fd1796019edfb4a84a81d1697")))
```
pyrocto: How was this number generated?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+    loop(b1.iterator, b2.iterator)
+  }
+
+  def safeUncons(b: ByteString): (ByteString, ByteString) = {
+    val head = if (b.isEmpty()) b else b.substring(0, 1)
+    val tail = if (b.isEmpty()) b else b.substring(1)
+    (head, tail)
+  }
+
+  val lookupRef: Long = 10L
```
goral09: We need to find a way to prevent from `SystemProcesses` overwriting these. Maybe pass initial value that is one bigger that the biggest ref value in `SystemProcesses` ?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+  val lookupPatterns = List(
+    BindPattern(
+      Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true)), ChanVar(FreeVar(1))),
+      freeCount = 2))
```
goral09: (I'm now validating my understanding)

From the name registry document:
```
[r, "lookup"]!(`rho:uuid:<uuid given by registering timesTwo>`, *twoXret)
```

`lookupPatterns` correspond to what is being sent on `[r, "lookup"]` ?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+  val lookupChannels  = List(Channel(Quote(GPrivate(ByteString.copyFrom(Array[Byte](10))))))
+  val insertRef: Long = 12L
+  val insertPatterns = List(
+    BindPattern(
+      Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true)),
+          Quote(Par(exprs = Seq(EVar(FreeVar(1))), connectiveUsed = true)),
+          ChanVar(FreeVar(2))),
+      freeCount = 3
+    ))
```
goral09: And `insertPatterns` corresponds to:
```
[r, "register"]!("uuid", bundle+{*timesTwo}, *print)
``` 
? 

Meaning this pattern won't match the pubkey based registration?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+              case Some(Expr(GByteArray(bs))) =>
+                val head = if (bs.isEmpty()) bs else bs.substring(0, 1)
+                val tail = if (bs.isEmpty()) bs else bs.substring(1)
+                data match {
+                  case Channel(Quote(dataPar)) =>
+                    dataPar.singleExpr match {
+                      case Some(Expr(EMapBody(parMap))) =>
+                        parMap.ps.get(Expr(GByteArray(head))) match {
+                          case None => localFail()
+                          case Some(value) =>
```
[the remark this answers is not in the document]
guardbotmk3: How do you call localFail() in that case?
Also, I don't like writing code I don't understand, and there's magic going on that I don't understand. So you'll have to explain it better.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+  def succeed(ret: Channel, result: Par, rand: Blake2b512Random): Task[Unit] =
+    ret match {
+      case Channel(q @ Quote(_)) =>
+        handleResult(space.produce(q, ListChannelWithRandom(Seq(Quote(result)), rand, None), false))
+      case _ => Task.unit
+    }
+
+  def fail(ret: Channel, rand: Blake2b512Random): Task[Unit] = succeed(ret, Par(), rand)
```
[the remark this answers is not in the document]
guardbotmk3: Are you asking me to copy paste the body of succeed into fail? Because that seems to violate DRY

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: Nodes are maps.
The keys in the map are byte arrays of length [0, 1]
The values in the map are tuples:
(tag, edgeAdditional, data)
tag:
if tag is 0, it is a terminal edge
if tag is 1, data is a name where another map can be read.
edgeAdditional:
this is a bytestring that stores any additional edge labeling in the radix tree.
data:
if tag is 0, this is the stored data.
if tag is 1, this is a name where the process recurs.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: kyle@clinch:~/pyrofex-devel/rchain$ echo -en "rho:io:registry" | b2sum | sed -E 's/.{64}/\0\n/'
a4fd447dedfc960485983ee817632cf36d79f45fd1796019edfb4a84a81d1697
d5fd3d8daf9f295aa590b37b50d5518803b4596ed6940aa42d46b1413a1bb16e  -

We can use something else if you want, this is a minor thing to focus on.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  val lookupRef: Long = 10L
+  val lookupPatterns = List(
+    BindPattern(
+      Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true)), ChanVar(FreeVar(1))),
+      freeCount = 2))
+  val lookupChannels = List(Channel(Quote(GPrivate(ByteString.copyFrom(Array[Byte](10))))))
+
+  // This is the only construction side effect.
+  space.install(lookupChannels, lookupPatterns, TaggedContinuation(ScalaBodyRef(lookupRef)))
```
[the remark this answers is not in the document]
guardbotmk3: See @pyrocto 's comment about centralizing it. Remember this is a WIP.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+                     replaceChan: Channel,
+                     retChan: Channel,
+                     dataRand: Blake2b512Random,
+                     failRand: Blake2b512Random): Task[Unit] =
+    for {
+      _ <- replace(data, replaceChan, dataRand)
+      _ <- fail(retChan, failRand)
+    } yield ()
+
+  def fetchData(dataSource: Quote,
```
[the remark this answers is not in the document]
guardbotmk3: Will do.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,230 @@
...
+    }
+    locallyFree: "\000\000\000\000\000\000\000\000"
+    connective_used: false
+  }
+  persistent: false
+  locallyFree: "\001\000\000\000\000\000\000\000"
+  connective_used: false
+}
+locallyFree: "\000\000\000\000\000\000\000\000"
+connective_used: false""")
```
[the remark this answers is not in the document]
guardbotmk3: What does what correspond to?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+  val lookupChannels  = List(Channel(Quote(GPrivate(ByteString.copyFrom(Array[Byte](10))))))
+  val insertRef: Long = 12L
+  val insertPatterns = List(
+    BindPattern(
+      Seq(Quote(Par(exprs = Seq(EVar(FreeVar(0))), connectiveUsed = true)),
+          Quote(Par(exprs = Seq(EVar(FreeVar(1))), connectiveUsed = true)),
+          ChanVar(FreeVar(2))),
+      freeCount = 3
+    ))
+  val insertChannels = List(Channel(Quote(GPrivate(ByteString.copyFrom(Array[Byte](12))))))
```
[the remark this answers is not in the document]
guardbotmk3: Yes, they don't need to be the same.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
ArturGajowy: > So you'll have to explain it better.

Every assignment in a for-comprehension is a pattern match, or a destructuring assignment to be more precise. The difference is: in a pattern match you get to decide what to do for each case; in a D.A. you just get the values assigned if everything matches or a MatchError otherwise.

Notice the `: Any`-s hiding the type info from the compiler in the example below:

```scala
val key: Any = Some(42)
val v1: Any  = EVar(Var(FreeVar(1)))
val v2: Any  = EVar(Var(FreeVar(2)))
val vs: Any  = v1 :: v2 :: Nil

for {
  _                         <- ().pure[Id] //Any Monad will do just fine
  Some(k)                   = key
  EVar(a) :: EVar(b) :: Nil = vs
  Var(FreeVar(x))           = a
  Var(FreeVar(y))           = b
} yield (k, x, y)

//prints:
//res0: cats.Id[(Any, Int, Int)] = (42,1,2)
```

> How do you call localFail() in that case?

If any of the patterns doesn't match, a MatchError is thrown. Since our handling of non-matches is always the same and doesn't rely on the things matched so far, we can simply:

```scala
try {
  for {
  //...
  } yield result
} catch {
  case x: MatchError => localFail()
}
```

I hate resorting to the `try/catch` too, but I'd say the readability gain justifies that.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
ArturGajowy: :+1: Please paste this into the `Registry`'s javadoc

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
ArturGajowy: If not for the `Task` effect around, we could just do:
```scala
try {
  val Some(k)                   = key
  val EVar(a) :: EVar(b) :: Nil = vs
  val Var(FreeVar(x))           = a
  val Var(FreeVar(y))           = b
  (x, y, k) //return result
} catch {
  case x: MatchError => localFail()
}
```

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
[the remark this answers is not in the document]
goral09: the ASCII you're parings.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: Insert (and likely lookup too) are not the public API. You'll have a much easier time understanding this code if you **ignore** that document.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+                          false))
+      _ <- handleResult(
+            space.consume(
+              Seq[Channel](Quote(channel), dataSource),
+              Seq(prefixRetReplacePattern, triePattern),
+              TaggedContinuation(ScalaBodyRef(lookupCallbackRef)),
+              false
+            ))
+    } yield ()
+  }
```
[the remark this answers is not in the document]
guardbotmk3: The order isn't required. I just think they read better in this order.
You'll have to be more specific about what you want me to comment.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,408 @@
...
+        insertRef         -> insert,
+        insertCallbackRef -> insertCallback)
+}
+
+object Registry {
+  val registryRoot = GPrivate(
+    ByteString.copyFrom(
+      Base16.decode("a4fd447dedfc960485983ee817632cf36d79f45fd1796019edfb4a84a81d1697")))
+
+  def testTableCreator(space: Runtime.RhoISpace, dispatch: Runtime.RhoDispatch) = {
```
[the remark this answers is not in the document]
guardbotmk3: It's not even used, I'll remove it.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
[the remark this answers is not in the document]
guardbotmk3: Read the comments I've added.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
guardbotmk3: I understand the last thing you wrote.
I think we can do it in that style.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
guardbotmk3: We will use a PureRSpace when we construct the registry in dispatch.scala

guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+import monix.eval.Task
+import scala.collection.immutable.Seq
+import scala.collection.{Seq => RootSeq}
+
+class Registry(private val space: ISpace[Channel,
+                                         BindPattern,
+                                         ListChannelWithRandom,
+                                         ListChannelWithRandom,
+                                         TaggedContinuation],
+               private val dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation]) {
```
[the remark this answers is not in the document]
guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+  def replace(data: Channel, replaceChan: Channel, dataRand: Blake2b512Random): Task[Unit] =
+    replaceChan match {
+      case Channel(q @ Quote(_)) =>
+        handleResult(
+          space
+            .produce(q, ListChannelWithRandom(Seq(data), dataRand, None), false))
+      case _ => Task.unit
+    }
+
+  def failAndReplace(data: Channel,
```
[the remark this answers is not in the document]
ArturGajowy: because that's the order it does things :) That's a minor nit/stylistic thing I don't insist on though :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -86,19 +109,28 @@ object Runtime {
...
-      3L -> SystemProcesses.stderrAck(space, dispatcher),
-      4L -> SystemProcesses.ed25519Verify(space, dispatcher),
-      5L -> SystemProcesses.sha256Hash(space, dispatcher),
-      6L -> SystemProcesses.keccak256Hash(space, dispatcher),
-      7L -> SystemProcesses.blake2b256Hash(space, dispatcher),
-      9L -> SystemProcesses.secp256k1Verify(space, dispatcher)
-    )
+    def dispatchTableCreator(space: RhoISpace, dispatcher: RhoDispatch): RhoDispatchMap = {
+      import BodyRefs._
+      val registry = new Registry(space, dispatcher)
```
goral09: `space` is not `PureRSpace` so it will throw as @ArturGajowy pointed out some time ago. This has some advantage like not having to deal with an effect (`F[_]`) when  doing `install`s in the `Registry`.

goral09: Although, the truth is that we use `ISpace` to install `SystemProcesses` as well.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,230 @@
...
+      data = Seq(
+        Expr(EMapBody(ParMap(Seq(
+          emptyByteArray  -> Par(exprs = Seq(Expr(ETupleBody(ETuple(Seq(GInt(0), emptyByteArray, GInt(7))))))),
+          dcByteArray     -> Par(exprs = Seq(Expr(ETupleBody(ETuple(Seq(GInt(0), branch1, GInt(8))))))),
+          eSevenByteArray -> Par(exprs = Seq(Expr(ETupleBody(ETuple(Seq(GInt(0), branch2, GInt(9)))))))
+        ))))),
+      persistent = false
+    )
+
+    val lookupPar: Par = Par.fromAscii("""
```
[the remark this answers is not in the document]
guardbotmk3: We literally can't parse the rholang to do this.
It uses private names that we can't put into the rholang.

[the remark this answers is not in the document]
guardbotmk3: This **doesn't** use persistent rspace, and it's not an integration test. It's testing only the registry functionality.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,216 @@
...
+class Registry(private val space: ISpace[Channel,
+                                         BindPattern,
+                                         ListChannelWithRandom,
+                                         ListChannelWithRandom,
+                                         TaggedContinuation],
+               private val dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation]) {
+  import Registry._
+  def commonPrefix(b1: ByteString, b2: ByteString): ByteString = {
+    val prefixOut = ByteString.newOutput()
+    def loop(it1: ByteIterator, it2: ByteIterator): ByteString = {
```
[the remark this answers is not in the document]
guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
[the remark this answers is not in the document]
guardbotmk3: Each of the patterns is simple, and taken together they make sense. The comment would literally be: `@x, @y, z`

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,2403 @@
...
+          ISpace[Channel,
+                 BindPattern,
+                 ListChannelWithRandom,
+                 ListChannelWithRandom,
+                 TaggedContinuation]) => R
+          ): R = {
+    withTestSpace { space =>
+      lazy val registry: Registry = new Registry(space, dispatcher)
+      lazy val dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation] =
+        RholangAndScalaDispatcher.create[Task, Task.Par](space, registry.testingDispatchTable, Map.empty)
```
goral09: If you create a `dispatcher` with `urnMap` 
```scala
lazy val urnMap: Map[String, Par] = Map(
  "rho:registry:lookup" -> GPrivate(ByteString.copyFrom(Array[Byte](10))),
  "rho:registry:insert" -> GPrivate(ByteString.copyFrom(Array[Byte](12))),
  "rho:registry:delete" -> GPrivate(ByteString.copyFrom(Array[Byte](14)))
)
lazy val dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation] =
  RholangAndScalaDispatcher.create[Task, Task.Par](space, registry.testingDispatchTable, urnMap)
```

then now instead of hardcoding ASCII we can parse rholang code:
```scala
def parse(rho: String): Par = {
  val source     = s"for (@x3, @x2, @x1, @x0 <- @0) { $rho }"
  val astTerm = Interpreter.buildNormalizedTerm(new StringReader(rho)).value
  val rhoTerm = astTerm.receives.head.body
  rhoTerm
}
…
val rholangString = """new r(`rho:registry:lookup`) in {
        #        r!("0897".hexToBytes(), "result0") |
        #        r!("0897dcbcb195efbad3b7ca343f50713e".hexToBytes(), "result1") |
        #        r!("0897e775426a859c8d3b25df4508cda9".hexToBytes(), "result2")
        #      }""".stripMargin('#')
val lookupRholang = parse(rholangString)
…
val result = withRegistryAndTestSpace { (reducer, space) =>
      implicit val env    = Env[Par]()
      Console.println(StoragePrinter.prettyPrint(space.store))
      val resultTask = for {
        _ <- reducer.eval(Par().addSends(rootSend, branchSend))
        _ <- reducer.eval(lookupRholang)
      } yield space.store.toMap

      Await.result(resultTask.runAsync, 3.seconds)
    }
```

goral09: Also, doesn't it show that a malicious validator can modify the code of rchain so that any data sent to `rho:registry:*` is also sent to the name he controls?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,591 @@
...
+                      _ <- replace(Quote(updatedMap), replaceChan, dataRand)
+                      _ <- replace(parentData, parentReplace, parentRand)
+                      _ <- succeed(ret, ps(2), callRand)
+                    } yield ()
+                  } else if (parMap.ps.size != 2) {
+                    localFail()
+                  } else {
+                    val it = (parMap.ps - parByteArray(head)).iterator
+                    for {
+                      _ <- Function.tupled(mergeWithParent(_, _))(it.next)
```
goral09: Why complicate it with an `Iterator` ? 
```scala
val it = parMap.ps - parByteArray(head)
  for {
    _ <- Function.tupled(mergeWithParent(_, _))(it.head)
```
works just the same

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,591 @@
...
+          val Some(Expr(ETupleBody(ETuple(ps, _, _)))) = value.singleExpr()
+          if (ps.length != 3)
+            localFail()
+          else {
+            // The second tuple field should be a bytearray in both cases.
+            val Some(Expr(GByteArray(edgeAdditional))) = ps(1).singleExpr()
+            ps(0).singleExpr() match {
+              case Some(Expr(GInt(0))) =>
+                if (tail == edgeAdditional) {
+                  if (parMap.ps.size > 2) {
```
goral09: Can you explain why there are 3 cases here and each has a different removal handling?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
goral09: No you don't pass the `PureRSPace` in the runtime. `space` is of type `RhoISpace` is just `ISpace`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -2400,4 +2406,387 @@ connective_used: false
         List()
       )))
   }
+
+  "Public lookup" should "decode and then call lookup" in {
+    /* lookup par is effectively:
+      new r(`rho:registry:lookup`) in {
+        r!(`rho:id:bnm61w3958nhr5u6wx9yx6c4js77hcxmftc9o1yo4y9yxdu7g8bnq3`, "result0") |
+        r!(`rho:id:bnmzm3i5h5hj8qyoh3ubmbu57zuqn56xrk175bw5sf6kook9bq8ny3`, "result1") |
```
goral09: Why expose another API? Why can't we do it like that:
```
new r(`rho:registry:lookup:id`) in {
   r!("bnm61w3958nhr5u6wx9yx6c4js77hcxmftc9o1yo4y9yxdu7g8bnq3".hexToBytes(), "result0") 
} 
```
?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -572,6 +628,73 @@ class Registry(private val space: ISpace[Channel,
...
+          if (uri.startsWith("rho:id:")) {
+            val tail = uri.substring(7)
+            if (tail.size != 54) {
+              localFail()
+            } else {
+              // Could fail
+              // 256 bits plus 14 bit crc-14
+              val bytes: Array[Byte] = ZBase32.decode(tail, 270)
+              val crc: Short         = ((bytes(32).toShort & 0xff) | ((bytes(33).toShort & 0xfc) << 6)).toShort
+              if (crc == CRC14.compute(bytes.view.slice(0, 32))) {
```
goral09: Can you explain why do we need CRC?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,591 @@
...
+                    _ <- replace(Quote(updatedMap), replaceChan, dataRand)
+                    _ <- replace(Quote(newMap), Quote(newName), callRand.splitByte(0))
+                    _ <- succeed(ret, valuePar, callRand.splitByte(1))
+                  } yield ()
+                }
+                ps(0).singleExpr() match {
+                  case Some(Expr(GInt(0))) =>
+                    // Replace key
+                    if (tail == edgeAdditional) {
+                      insert()
```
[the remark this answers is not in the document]
guardbotmk3: Yes, we have to consider duplicates, but an insert with an existing key should overwrite the previous contents.

We plan to use this for the public key registry

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -0,0 +1,591 @@
...
+          val Some(Expr(ETupleBody(ETuple(ps, _, _)))) = value.singleExpr()
+          if (ps.length != 3)
+            localFail()
+          else {
+            // The second tuple field should be a bytearray in both cases.
+            val Some(Expr(GByteArray(edgeAdditional))) = ps(1).singleExpr()
+            ps(0).singleExpr() match {
+              case Some(Expr(GInt(0))) =>
+                if (tail == edgeAdditional) {
+                  val updatedMap: Par = ParMap(SortedParMap(parMap.ps - parByteArray(head)))
```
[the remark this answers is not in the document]
guardbotmk3: Yes.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: When we add, we split.
splitting should result in a map with 2 keys.

When we merge, if our removal would shrink to one key, we have to merge.

Note that the third case is just a failure that should never occur.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
guardbotmk3: These are intended for transcription by hand, hence the base32 encoding.

We want to be able to tell a user if their transcription was invalid.
We can detect a mistake in transcription with probability (1 -  1/16000).
We use CRC14 because of the base32 encoding, and the choice of 256 bits. 256 bits is 52 base 32 characters, but the last character only stores 1 bit. We use the 4 leftovers and 2 more for 14. The CRC-14 is CRC-14-DARC

dckc: Add that design rationale as a comment in the code, perhaps?

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,641 @@
...
+        r!(`rho:id:bnmzm3i5h5hj8qyoh3ubmbu57zuqn56xrk175bw5sf6kook9bq8ny3`, "result1")
+      }"""
+    val lookupPar: Par = Interpreter.buildNormalizedTerm(new StringReader(lookupString)).value
+    val completePar                     = lookupPar.addSends(rootSend, branchSend)
+    implicit val rand: Blake2b512Random = baseRand.splitByte(4)
+    val newRand = rand.splitByte(2)
+    val randResult0                     = newRand.splitByte(0)
+    randResult0.next; randResult0.next
+    val randResult1 = newRand.splitByte(1)
+    randResult1.next; randResult1.next
```
goral09: This is sad that you need to do that dance with the PRNG in this test. `rho:registry:lookup` is high level interface.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,641 @@
...
+        List()
+      )))
+  }
+
+  "Random Registry" should "use the random generator and insert" in {
+    val registerString = """
+      new rr(`rho:registry:insertRandom`), rl(`rho:registry:lookup`), x, y in {
+        rr!(bundle+{*x}, *y) |
+        for(@{uri}, resultChan <- y) {
+          for(@{finalResult} <- resultChan) {
```
goral09: As discussed - can you try improving the UX so that the client of the API needs to listen only on one channel for the UUID under which the name was registered? Instead of having to listen on two as it's now.

guardbotmk3:
OK, I made the public API change that you requested.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -686,34 +665,69 @@ class Registry(private val space: PureRSpace[Task,
...
           }
         } catch {
           case _: MatchError               => localFail()
           case _: IllegalArgumentException => localFail()
         }
+      case _ => Task.unit
+    }
+
+  def publicRegisterInsertCallback(args: RootSeq[ListChannelWithRandom]): Task[Unit] = {
+    println("publicRegisterInsertCallback called.")
```
goral09: `println`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -608,32 +606,31 @@ class RegistrySpec extends FlatSpec with Matchers with RegistryTester {
...
       writeFlag = true,
       readFlag = false)
 
+    // Mimic the uri creation
+    val fullKey = new Array[Byte](34)
+    Array.copy(uriBytes, 0, fullKey, 0, 32)
+    val crc = Registry.CRC14.compute(fullKey.view.slice(0, 32))
+    fullKey(32) = (crc & 0xff).toByte
+    fullKey(33) = ((crc & 0xff00) >>> 6).toByte
+    val expectedUri = "rho:id:" + ZBase32.encodeToString(fullKey, 270)
```
goral09: Let's extract this code to a helper method `createUri(bytes: Array[Byte]): GUri` and use here and in the `Registry.`

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -0,0 +1,641 @@
...
+
+    result.get(resultChanList("result0")) should be(
+      Some(Row(
+        List(
+          Datum.create(Channel(Quote(GString("result0"))),
+                       ListChannelWithRandom(Seq(Quote(GUri(expectedUri))), randResult0),
+                       false)),
+        List()
+      )))
+    result.get(resultChanList("result1")) should be(
```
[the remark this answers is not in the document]
goral09: Yeah, that would remove quite a lot of lines. :+1:

guardbotmk3:
OK, extracted helper function to build URI's from bytes.

PR #1328, opened 2018-08-21
CORE-1038 Change TcpTransportLayer not to log the error messages when sending data

rabbitonweb:
## Overview
Change TcpTransportLayer not to log the error messages when sending data.
Correct error message will be return, it is up to the caller of the `roundTrip` method to decide what to do with the error

CORE-1038


rabbitonweb:
@sebtomba they are not being swallowed, if TransportLayer returns an error it is up to the client to do sth about it (and they do)

PR #1329, opened 2018-08-21
CORE-1030 Introduce two flags: maxMessageSize and threadPoolSize

rabbitonweb:
## Overview
Introduce two flags: maxMessageSize and threadPoolSize

I have troubles using `threadPoolSize`, will have to figure that out

CORE-1030

### DIFF
--- a/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
@@ -56,6 +56,8 @@ object Configuration {
   private val DefaultRequiredSigns              = 0
   private val DefaultApprovalProtocolDuration   = 5.minutes
   private val DefaultApprovalProtocolInterval   = 5.seconds
+  private val DefaultMaxMessageSize: Int        = 100 * 1024 * 1024
+  private val DefaultThreadPoolSize: Int        = 4000
 
   private val DefaultBootstrapServer: PeerNode = PeerNode
     .parse("rnode://de6eed5d00cf080fc587eeb412cb31a75fd10358@52.119.8.109:40400")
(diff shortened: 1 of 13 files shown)


node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -44,6 +44,8 @@ object Configuration {
   private val DefaultKeyFileName                = "node.key.pem"
   private val DefaultSecureRandomNonBlocking    = false
   private val DefaultMaxNumOfConnections        = 500
+  private val DefaultMaxMessageSize: Int        = 100 * 1024 * 1024
+  private val DefaultThreadPoolSize: Int        = 4000
```
ArturGajowy: That's a lot of threads! :) What's the motivation for a number this high?

Why not use `2 * java.lang.Runtime.getRuntime().availableProcessors()` by default? This way we can load all cores without causing too much context switching.

node/src/main/scala/coop/rchain/node/configuration/model.scala
```diff
@@ -16,7 +16,9 @@ case class Server(
     dataDir: Path,
     mapSize: Long,
     inMemoryStore: Boolean,
-    maxNumOfConnections: Int
+    maxNumOfConnections: Int,
+    maxMessageSize: Int,
+    threadPoolSize: Int
```
ArturGajowy: Shouldn't we mark the flags as not used/implemented until they are actually used?

rabbitonweb: they are being used, see the files changed

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
rabbitonweb: We are actually LIMITING number of threads , because normally scheduler will take all that your system got, killing it at some point (4k threas for mac)

goral09: I think what @ArturGajowy meant was to have
```scala
private val DefaultThreadPoolSize: Int        = 2 * java.lang.Runtime.getRuntime().availableProcessors()
```
instead of 4000.

rabbitonweb: Ok, fair enough? Why 2 * cores? And not `n`?

PR #1332, opened 2018-08-21
Maximum bipartite matching in listMatch (v2)

ArturGajowy:
## Overview

This is #1272 with review comments applied.

I've first added tests securing the corner cases we've discovered so far (3 for why MBM is required instead of Stable Marriage, 1 for why remainders have to be handled using placeholder patterns)

I've then made remainder Par-s' fields to be updated atomically (see `Make merger overwrite the whole Par field instead of adding to it`), so that the stale state from previous matches doesn't linger once switching to MBM. Notice this is tested with the old matching algo as well, before introducing MBM. The tests for that are also the tests showing why MBM is needed.

Then MBM is introduced in the next 2 commits.

Lastly, memoization for MBM's calls to `matchFunction` follows.

I've skipped adding the `MonadState` constraint for F and controlling state updates from inside MBM, because:
1. atomic updates of par's fields fix the issue that we wanted to address with that,
2. I'd have to also constrain the state in that `MonadState[F, S]` to be a `Monoid`, or otherwise assert its merge-ability,
3. That would require a substantial rework of the algo's code.

Hope this approach is acceptable. Looking forward to your comments :)

---

Reimplements matching of List[Pattern] with List[Term], making it deterministic and polynomial-time, according to the [Maximum Bipartite Matching algorithm](http://olympiad.cs.uct.ac.za/presentations/camp2_2017/bipartitematching-robin.pdf) (see: "alternate version" in the slides)

Supersedes #1201 due to a counterexample we've missed initially.

### JIRA issue:
https://rchain.atlassian.net/browse/RHOL-533

### Checklist

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
@@ -207,7 +207,7 @@ object SpatialMatcher extends SpatialMatcherInstances {
   def listMatchSingle[T](tlist: Seq[T], plist: Seq[T])(
       implicit lf: HasLocallyFree[T],
       sm: SpatialMatcher[T, T]): OptionalFreeMapWithCost[Unit] =
-    listMatchSingleNonDet(tlist, plist, (p: Par, _: T) => p, None, false)
+    listMatchSingleNonDet(tlist, plist, (p: Par, _: Seq[T]) => p, None, false)
       .mapK[OptionT[State[CostAccount, ?], ?]](
         new FunctionK[StreamT[State[CostAccount, ?], ?], OptionT[State[CostAccount, ?], ?]] {
           override def apply[A](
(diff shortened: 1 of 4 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -250,129 +234,95 @@ object SpatialMatcher extends SpatialMatcherInstances {
...
-            foldRemainder(rem, p)
-          else
-            NonDetFreeMapWithCost.emptyMap[Par]
+      sm: SpatialMatcher[T, T]): OptionalFreeMapWithCost[Unit] = {
+
+    sealed trait Pattern
+    case class Term(term: T)         extends Pattern
+    case class Remainder(level: Int) extends Pattern
+
+    val remainderPatterns: Seq[Pattern] = varLevel.fold(
```
goral09: Why are you creating as many remainder patterns as the `varLevel`?

goral09: I know.

guardbotmk3: He's using fold to pattern match on the option.

PR #1340, opened 2018-08-22
CORE-1051/52 -  Print the whole stacktrace &  Stop logging warn when disconnecting

rabbitonweb:
## Overview
1. Print the whole stacktrace when unhandled error happens
2. Stop logging warn when disconnecting a peer whose connection no longer exists

It is typicall that multiple threads will try to close same
connection. One will succeed, the otherone should just ignore if
connection was already closed (does not exist)

CORE-1051 & CORE-1052


rabbitonweb:
Closing because done from upstream ... 🤦‍♂️ Recreated at https://github.com/rchain/rchain/pull/1351

PR #1346, opened 2018-08-22
Rholang: Fix bug where pattern connectives were leaking.

guardbotmk3:
## Overview
ConnectiveUsed should not be set in a receive based on usage in patterns.

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -700,15 +700,12 @@ object ProcNormalizeMatcher {
                                                                  line,
                                                                  col))
                         })
-          bindCount = mergedFrees.countNoWildcards
-          binds     = receipts.map(receipt => receipt._1)
-          bindingsConnectiveUsed = binds
-            .flatMap(_.patterns)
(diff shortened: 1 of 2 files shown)


goral09:
As @ArturGajowy points out -- I would like to see a test which shows that `bind` influencing the `connectiveUsed` of the whole `Receive` is an error. Is that the case? Or is it only because of the performance (where we use bipartite matching instead of equality check). Because, from what it looks like:
https://github.com/rchain/rchain/blob/dev/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala#L641-L650

we call `spatialMatch` on `ReceiveBind`s anyway.

I guess what I'm asking is: how did you end up deciding that current implementation is incorrect.

guardbotmk3:
> I guess what I'm asking is: how did you end up deciding that current implementation is incorrect.
connectiveUsed has a meaning:
the meaning is: "is this term concrete, or only a pattern"
You shouldn't be able to run a pattern, only a concrete term.

We should fail to normalize if the resulting top-level connective has connectiveUsed set to true.

> I would like to see a test which shows that bind influencing the connectiveUsed of the whole Receive is an error.

The tests that I changed already did that.

What @ArturGajowy asked for was a test that had a connective not in the bind pattern.
However, I think that's out of scope for a fix that I noticed while working on something else.

PR #1349, opened 2018-08-23
make sure the base16 representation has an even number of digits when decoding

odeac:
## Overview
The previous implementation of Base16.decode asserts that the number of digits is even. Now we pad it with an 0 in such a situation.

https://rchain.atlassian.net/browse/CORE-981

### Notes

### DIFF
--- a/crypto/src/main/scala/coop/rchain/crypto/codec/Base16.scala
+++ b/crypto/src/main/scala/coop/rchain/crypto/codec/Base16.scala
@@ -4,10 +4,10 @@ object Base16 {
   def encode(input: Array[Byte]): String = bytes2hex(input, None)
 
   def decode(input: String): Array[Byte] = {
-    val paddedInput = {
+    val paddedInput = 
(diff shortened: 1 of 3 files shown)


crypto/src/main/scala/coop/rchain/crypto/codec/Base16.scala
```diff
@@ -3,13 +3,13 @@ package coop.rchain.crypto.codec
...
-      // TODO: Learn how to handle errors more functionally
-      throw new Error("Input must always be even in length")
-    }
+  def decode(input: String): Array[Byte] = {
+    val paddedInput = 
+      if (input.length % 2 == 0) input
+      else "0" + input
+
+    hex2bytes(paddedInput)
+  }
```
KentShikama: Do we have a round trip test for encode/decode?

odeac: now we have. See the last commit

PR #1361, opened 2018-08-24
Improve integration test

odeac:
## Overview
add command line arguments for setting the timeouts for the tests
change the scope of slow fixtures to "package" in order to run the setup it only once per package

### Notes

### DIFF
--- a/integration-testing/test/conftest.py
+++ b/integration-testing/test/conftest.py
@@ -43,12 +43,12 @@ def parse_config(request):
                   network_converge_timeout = converge_timeout if converge_timeout > 0 else 200 + peer_count * 2
                   )
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def config(request):
     return parse_config(request)
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="session")
 def docker():
     import docker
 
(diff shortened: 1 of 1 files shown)


integration-testing/test/conftest.py
```diff
@@ -11,6 +11,12 @@ def pytest_addoption(parser):
     parser.addoption(
         "--peer-count", action="store", default="2", help="number of peers in the network (excluding bootstrap node)"
     )
+    parser.addoption(
+        "--start-timeout", action="store", default="0", help="timeout in seconds for starting a node. Defaults to 30 + peer_count * 10"
```
birchmd: Maybe make the default equal to 30 and remove the if statement below?

integration-testing/test/conftest.py
```diff
@@ -11,6 +11,12 @@ def pytest_addoption(parser):
...
+    )
+    parser.addoption(
+        "--converge-timeout", action="store", default="0", help="timeout in seconds for network converge. Defaults to 200 + peer_count * 10"
```
birchmd: Same here

integration-testing/test/conftest.py
odeac: if there is no value provided the formula `start-timeout = 30 + peer-count * 10` is used. If a value is provided then we use that exact value

integration-testing/test/conftest.py
odeac: same here, with default formula `converge-timeout=200 + peer-count * 10`

PR #1363, opened 2018-08-24
Add a WalletCheck that holds a wallet until someone presents a 'ticket'...

pyrocto:
… a signed message including the public key that is the preimage of the hash.

## Overview
Ethereum accounts are the hashes of the public keys of the account holders.  We want to create wallets for account holders, but they need to be able to sign claims for funds.  This is something like the coat check pattern, but it holds a wallet.  Given a hash and a purse, it creates a claim contract at the hash that waits for a 'ticket', a signed message containing the public key and a status return channel.  Once it checks that everything's in order, it creates a BasicWallet containing the purse that uses the given public key, then creates a forwarder at the pubKey.

The use of forgeable names is completely insecure and should be replaced by use of the registry when it's ready.

### DIFF
--- a/casper/src/main/rholang/WalletCheck.rho
+++ b/casper/src/main/rholang/WalletCheck.rho
@@ -1,0 +1,72 @@
+// WalletCheck: holds funds until a public key with the given hash is presented,
+//     then creates a wallet with that public key and those funds.
+
+// create: makes a wallet check
+//
+// bytearray hash: the hash of the public key
+// Purse purse: the funds to put in the wallet
+contract @["WalletCheck", "create"](@hash, @purse) = {
+  new doneStore in {
+    doneStore!(false) |
+    
+    // claims the wallet by presenting the preimage of the hash
+    //
+    // bytearray pubkey
+    // name statusOut
+    // bytearray sig: signature of [pubKey, statusOut]
+    contract @hash(@[pubKey, statusOut], @sig) = {
+      for (@done <- doneStore) {
+        if (done) {
+          doneStore!(done) |
+          *statusOut!([false, "Already claimed wallet"])
+        } else {
+          // verify signature 
+          new verifiedOut in {
+            @"secp256k1Verify"!([pubKey, statusOut].toByteArray(), sig, pubKey, *verifiedOut) |
+            for (@verified <- verifiedOut) {
+              if (verified) {
+                new hashOut in {
+                  @"keccak256Hash"!(pubKey.toByteArray(), *hashOut) |
+                  for (@pkHash <- hashOut) {
+                    if (pkHash == hash) {
+                      new walletOut in {
+                        @"BasicWallet"!(purse, "secp256k1", pubKey, *walletOut) |
+                        for (@wallet <- walletOut) {
+                          doneStore!(true) |
+                          new pkbaOut in {
+                            pkbaOut!(pubKey.toByteArray()) |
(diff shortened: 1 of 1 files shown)


casper/src/main/rholang/WalletCheck.rho
```diff
@@ -0,0 +1,78 @@
...
+
+                  // Check that given pubKey is the preimage of hash
+                  @"keccak256Hash"!(pubKey.toByteArray(), *hashOut) |
+                  for (@pkHash <- hashOut) {
+                    if (pkHash == hash) {
+                      new walletOut in {
+
+                        // Create the wallet
+                        @"BasicWallet"!(purse, "secp256k1", pubKey, *walletOut) |
+                        for (@wallet <- walletOut) {
```
birchmd: Needs to be `for (@[wallet] <- walletOut)` with the current `BasicWallet` API

casper/src/main/rholang/WalletCheck.rho
```diff
@@ -0,0 +1,78 @@
...
+                          doneStore!(true) |
+                          
+                          // Install a forwarder to the wallet at a forgeable name.
```
birchmd: This is different than what I have done. I just put the wallet as a replicated send at a name related to the PK, so to use the wallet you would start with `for(@wallet <- rho:pubkey:...)` or something similar, and then you could call the wallet directly. I think is this a little cleaner than writing a forwarder for every method because it requires less boilerplate.

It is of course still valid in both approaches that we should use the registry when it be comes available.

casper/src/main/rholang/WalletCheck.rho
```diff
@@ -0,0 +1,78 @@
...
+                doneStore!(done) |
+                statusOut!([false, "Signature verification failed"])
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
birchmd: @guardbotmk3 Likes to see new lines at the end of a file.

casper/src/main/rholang/WalletCheck.rho
```diff
@@ -0,0 +1,78 @@
...
+    // bytearray pubkey
+    // name statusOut
+    // bytearray sig: signature of [pubKey, statusOut]
+    contract @hash(@[pubKey, statusOut], @sig) = {
+      
+      // Make sure the purse hasn't already been claimed
+      for (@done <- doneStore) {
+        if (done) {
+          doneStore!(done) |
+          *statusOut!([false, "Already claimed wallet"])
```
birchmd: No `*` here.

casper/src/main/rholang/WalletCheck.rho
pyrocto: Done.

casper/src/main/rholang/WalletCheck.rho
pyrocto: Done.

casper/src/main/rholang/WalletCheck.rho
pyrocto: Done.

casper/src/main/rholang/WalletCheck.rho
pyrocto: Should be an `@`.  Changed everywhere.

pyrocto:
Moving to https://github.com/rchain/rchain/pull/1366.

PR #1368, opened 2018-08-24
Use pre-wallet instead of wallet since we only know the hash

pyrocto:
## Overview
https://github.com/rchain/rchain/pull/1366 created a pre-wallet that holds the wallet until the user presents the public key.  Eventually this won't be necessary because we can get the public key out of the signature like Ethereum does.

Note that the algorithm should be "keccak256Hash" when invoking this.


casper/src/main/rholang/WalletCheck.rho
```diff
@@ -31,9 +31,9 @@ contract @["WalletCheck", "create"](@hash, @purse) = {
                 new hashOut in {
 
                   // Check that given pubKey is the preimage of hash
-                  @"keccak256Hash"!(pubKey.toByteArray(), *hashOut) |
+                  algorithm!(pubKey.toByteArray(), *hashOut) |
                   for (@pkHash <- hashOut) {
-                    if (pkHash == hash) {
+                    if (pkHash == hash.hexToBytes()) {
```
birchmd: You should update the type annotation in your comment if you are going to now treat `hash` like a string.

casper/src/main/rholang/WalletCheck.rho
```diff
@@ -31,9 +31,9 @@ contract @["WalletCheck", "create"](@hash, @purse) = {
                 new hashOut in {
 
                   // Check that given pubKey is the preimage of hash
-                  @"keccak256Hash"!(pubKey.toByteArray(), *hashOut) |
+                  algorithm!(pubKey.toByteArray(), *hashOut) |
```
birchmd: The `algorithm` field in the `Wallet` case class was originally the signature algorithm. Interpreting it as the hashing algorithm is probably fine, but if the point of these "pre-wallets" is to always be based on Ethereum addresses, then maybe we should just remove that degree of freedom.

pyrocto: Done.

birchmd:
@stay I don't think this PR is needed after my own changes made in https://github.com/rchain/rchain/pull/1397
If you could take a look at that I would appreciate it

pyrocto:
Subsumed by 1397.

PR #1369, opened 2018-08-25
Prevent simulatenous proposes

KentShikama:
## Overview

https://rchain.atlassian.net/browse/RHOL-653

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -36,7 +36,7 @@ trait Casper[F[_], A] {
   def addBlock(b: BlockMessage): F[BlockStatus]
   def contains(b: BlockMessage): F[Boolean]
   def deploy(d: DeployData): F[Either[Throwable, Unit]]
-  def estimator: F[A]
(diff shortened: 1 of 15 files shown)


casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -138,6 +139,8 @@ sealed abstract class MultiParentCasperInstances {
       private val lastFinalizedBlockContainer = Ref.unsafe[F, BlockMessage](genesis)
 
       private val processingBlocks = new AtomicSyncVar(Set.empty[BlockHash])
+      private val createBlockLock  = new SyncVar[Unit]()
```
birchmd: I've changed my mind about using `SyncVar` here. I think that something which knows whether or not the resource is available is better because we want to simply end the proposal in the case that another thread is already doing one (as opposed to blocking and waiting for it to finish. I think that [java.util.concurrent.locks.Lock](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html) suits this purpose because it has a `tryLock` method.

KentShikama:
@birchmd PTAL

PR #1373, opened 2018-08-27
Clean up log messages for Casper

KentShikama:
Relative to #1369 ; just see d0b81e1617656e90c04deaff7f2f67d81cd418dc

### DIFF
--- a/scripts/p2p-test-tool.py
+++ b/scripts/p2p-test-tool.py
@@ -308,10 +308,10 @@ def test_propose(container):
     for container in client.containers.list(all=True, filters={"name":f".{args.network}"}):
             #Check logs for warnings(WARN) or errors(ERROR) on CASPER    
             for line in container.logs().decode('utf-8').splitlines():
-                if "WARN" in line and "CASPER" in line and not "wallets" in line:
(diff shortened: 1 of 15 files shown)


scripts/p2p-test-tool.py
```diff
@@ -308,10 +308,10 @@ def test_propose(container):
     for container in client.containers.list(all=True, filters={"name":f".{args.network}"}):
             #Check logs for warnings(WARN) or errors(ERROR) on CASPER    
             for line in container.logs().decode('utf-8').splitlines():
-                if "WARN" in line and "CASPER" in line and not "wallets" in line:
+                if "WARN" in line and "coop.rchain.casper" in line and not "wallets" in line:
```
birchmd: This actually isn't good enough because sometimes the class path gets abbreviated, e.g.
```
20:04:42.362 [node-io-8] WARN  coop.rchain.casper.genesis.Genesis$ - No wallets file specified and no default file found. No wallets will exist at genesis.
```
vs 
```
20:14:48.816 [grpc-default-executor-4] INFO  c.rchain.casper.MultiParentCasper$ - Received Deploy #1535400888002 -- new x0 in { x0!(({"one" :...
```

PR #1378, opened 2018-08-28
Revert "Prevent simulatenous proposes"

MParlikar:
Reverts rchain/rchain#1369

The Integration tests have failed for this PR.  Confirmed with @birchmd as well that an issue exists with this PR.


birchmd:
I have a hypothesis about why this does not work. I think it could be because the lock is taken by one gRPC thread and then part way through the `createBlock` logic it switches threads, so that at the end when the lock is supposed to be returned the current thread does not have it. This is why the unit tests are successful, but the node is not. I'm going to try a different locking mechanism to see if this solves the issue.

MParlikar:
Closing in favor of 1379.

PR #1383, opened 2018-08-28
CORE-1075 Upgrade http4s to 0.19.0-M1 to adhere to cats-effect dependency hell

rabbitonweb:
## Overview
We live in cats dependency hell. Monix 3.0.0.RC1 runs on cats effect 0.10 that
maybe just maybe is compatible with cats effect 1.0.0-RC2. Http4s
0.18.x runs on 0.10 but bunch of other stuff in node runs on
1.0.0-RC2.
Cats ecosystem is very unstable at the moment and will likely not
change very fast soon. This is our reality at the moment.

CORE-1075

### DIFF
--- a/node/src/main/resources/logback.xml
+++ b/node/src/main/resources/logback.xml
@@ -21,6 +21,7 @@
   </appender>
 
   <logger name="coop.rchain.rspace" level="warn" />
+  <logger name="org.http4s.blaze.pipeline.stages" level="error" />
   <logger name="io.netty" level="warn" />
 
   <root level="debug">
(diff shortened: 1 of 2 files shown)


goral09:
Well, that's not  _entirely_ true. cats announced release of stable 1.0.0 version for 3rd of September and I know that monix is already working on the stability issues against cats 1.0.0 so we should fix all that at the beginning of September.

rabbitonweb:
@goral09 wanna bet? :)

rabbitonweb:
description updated

goral09:
@lukasz-golebiewski PTAL

PR #1393, opened 2018-08-28
Bug fixes for ByteString (de)serialization

ArturGajowy:
## Overview
Title says it all. ~Was super easy :joy:~ Turns out not *that* easy.

### JIRA issue:
RHOL-537

### Checklist

### DIFF
--- a/models/src/main/scala/coop/rchain/models/BitSetBytesMapper.scala
+++ b/models/src/main/scala/coop/rchain/models/BitSetBytesMapper.scala
@@ -15,7 +15,7 @@ object BitSetBytesMapper {
     )(bitSetToByteString)
 
   def bitSetToByteString(bitset: BitSet): ByteString = {
-    val longs: Array[Long] = bitset.toBitMask
+    val longs: Array[Long] = if (bitset.isEmpty) Array.empty else bitset.toBitMask
     val byteBuffer =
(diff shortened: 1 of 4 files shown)


models/src/main/scala/coop/rchain/models/BitSetBytesMapper.scala
```diff
@@ -15,7 +15,7 @@ object BitSetBytesMapper {
     )(bitSetToByteString)
 
   def bitSetToByteString(bitset: BitSet): ByteString = {
-    val longs: Array[Long] = bitset.toBitMask
+    val longs: Array[Long] = if (bitset.isEmpty) Array.empty else bitset.toBitMask
```
guardbotmk3: I wonder if there is hysteresis at other boundaries. Would it be better to convert the bitset, and then filter the empties from the back?

ArturGajowy:
Turns out it's been even more broken... I should've been more careful with the bugfix and the tests.

The bugs were:
1. The encoding always paded to 8 bytes, resulting in trailing '0' bytes
2. The decoding always dropped the bytes past last multiple of 8,
   because the capacity of `ByteBuffer(size = 7).asLongBuffer()` is 0.

models/src/main/scala/coop/rchain/models/BitSetBytesMapper.scala
ArturGajowy: Good catch! Thanks for that. This fixes the bug '1.' I've mentioned in the commit message.

PR #1397, opened 2018-08-29
Full support for Rev issuance based on RHOC balances

birchmd:
## Overview
This PR complete's @pyrocto 's work to introduce a "pre-wallet" as part of the Rev issuance process. It enables us to do the Rev issuance directly from the output produced by the script from https://rchain.atlassian.net/browse/RHOL-601 . It includes a unit test to show that funds can indeed be unlocked by providing the right public key (the one corresponding to the ethereum address) and then that funds can be transferred.

### Notes


casper/src/main/rholang/WalletCheck.rho
```diff
@@ -1,20 +1,23 @@
...
     // claims the wallet by presenting the preimage of the hash
     //
-    // bytearray pubkey
+    // string pubkey: Base16 encoded public key. It should be true that
+    //                "0x" + Base16.encode(keccak256.hash(Base16.decode(pubkey)).takeRight(20)) == ethAddr
     // name statusOut
-    // bytearray sig: signature of [pubKey, statusOut]
-    contract @hash(@[pubKey, statusOut], @sig) = {
+    // string sig: signature of [pubKey, statusOut]
+    contract @ethAddr(@[pubKey, statusOut], @sig) = {
```
goral09: Question about rholang style. Why you don't use tuples for small sets of values?

casper/src/main/scala/coop/rchain/casper/genesis/contracts/PreWallet.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+    * Produces Rholang code which adds a wallet to the blockchain based on the
+    * given Wallet case class.
+    * @param w the PreWallet object containing the information which will go
+    *          on the blockchain.
+    * @return  Rholang code to add the pre-wallet to the blockchain.
+    */
+  def rhoCode(w: PreWallet): String = s"""
+    |new purseCh, walletCh in {
```
goral09: You don't use `walletCh`.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
```diff
@@ -0,0 +1,106 @@
...
+        .head
+        .exprs
+        .head
+        .getGByteArray
+        .toByteArray)
+    val unlockSig = Secp256k1.sign(unlockSigData, secKey)
+    assert(Secp256k1.verify(unlockSigData, unlockSig, Base16.decode("04" + pubKey)))
+
+    val unlockDeploy = ProtoUtil.termDeployNow(mkTerm(s"""
+      |@"$ethAddress"!(["$pubKey", "$statusOut"], "${Base16.encode(unlockSig)}")
```
goral09: I thought that the problem was the people don't know their ETH public keys?

casper/src/main/rholang/WalletCheck.rho
birchmd: It's a good point, we really should be using tuples. It's more of a habit for me because tuples didn't exist when I started writing Rholang.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
goral09: OK, that explains it. I thought that you're already solving this problem.

PR #1402, opened 2018-08-29
Make Rholang integers 64-bit

ArturGajowy:
## Overview
Title says it all. Cc @birchmd @KentShikama :)

### JIRA issue:
RHOL-689

### Checklist

### DIFF
--- a/models/src/main/scala/coop/rchain/models/rholang/sort/ScoreTree.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sort/ScoreTree.scala
@@ -26,7 +26,7 @@ trait ScoreTree {
   case class Node[T](children: Seq[Tree[T]]) extends Tree[T]
 
   sealed trait TaggedAtom
-  case class IntAtom(i: Int)          extends TaggedAtom
+  case class IntAtom(i: Long)         extends TaggedAtom
   case class StringAtom(s: String)    extends TaggedAtom
   case class BytesAtom(b: ByteString) extends TaggedAtom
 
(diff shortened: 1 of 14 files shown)


models/src/main/scala/coop/rchain/models/rholang/sort/ScoreTree.scala
```diff
@@ -26,7 +26,7 @@ trait ScoreTree {
   case class Node[T](children: Seq[Tree[T]]) extends Tree[T]
 
   sealed trait TaggedAtom
-  case class IntAtom(i: Int)          extends TaggedAtom
+  case class IntAtom(i: Long)         extends TaggedAtom
```
ArturGajowy: Before anybody asks - I've heard we're heading to arbitrary-length integers in the long term, so it makes little sense to rename to `LongFoo`-s just to go back to `[Big]Int[eger]Foo`-s. I've only done the rename for `evalToInt -> evalToLong`, where it actually matters what concrete integer type is returned.

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -41,7 +41,7 @@ object GroundNormalizeMatcher {
   def normalizeMatch(g: AbsynGround): Expr =
     g match {
       case gb: GroundBool   => BoolNormalizeMatcher.normalizeMatch(gb.bool_)
-      case gi: GroundInt    => GInt(gi.integer_)
+      case gi: GroundInt    => GInt(gi.long_.toLong)
```
goral09: My only concern is this might throw, but as I understand we catch that and report as `UnrecognizedInterpreterError` ?

ArturGajowy: ummm... right... forgot about that :joy:

Yes, it will be caught and reported just as you've said.
I'll add a TODO for handling the error purely.

goral09:
@ArturGajowy compilation failure: https://travis-ci.org/rchain/rchain/jobs/422065296#L2032

PR #1410, opened 2018-08-30
RHOL-621: Add validation check for deploy count

itegulov:
## Overview
Adds a validation check that deployCount matches the number of deploys

https://rchain.atlassian.net/browse/RHOL-621


casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -361,6 +363,15 @@ object Validate {
     }
   }
 
+  def deployCount[F[_]: Applicative: Log](b: BlockMessage): F[Either[InvalidBlock, ValidBlock]] =
+    if (b.header.get.deployCount == b.body.get.deploys.length) {
```
birchmd: We validate these fields are populated first before calling `.get` on them right? We really should encode whether those field checks happened or not into the type...

KentShikama: Yes and I agree

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -361,6 +363,15 @@ object Validate {
...
+      Applicative[F].pure(Right(Valid))
+    } else {
+      for {
+        _ <- Log[F].warn(ignore(b, s"block deploy count does not match to the amount of deploys."))
+      } yield Left(InvalidUnslashableBlock)
```
KentShikama: I think we should make this one slashable.

KentShikama: A bit late but I think the block hash mismatch should be slashable too. Don't forget to modify Casper.scala to handle invalid block

casper/src/main/scala/coop/rchain/casper/Validate.scala
itegulov: Yes, I remember that we had a discussion on #rholang about this and the proposed solution was to write a macro which would automatically generate safe protobuf classes (which do not wrap body and header into Option for example).

I am wondering whether using `required` for header and body were considered... There are a few drawbacks to this approach, but this might make sense.

itegulov: Oh, I didn't know it was removed. NVM then.

PR #1418, opened 2018-08-30
CORE-1101 Attempt to catch generative tests failure

ys-pyrofex:
## Overview
Order-independent comparison in the validateIndexedStates method.

https://rchain.atlassian.net/browse/CORE-1101


lukasz-golebiewski:
@ys-pyrofex Have you managed to make tests fail?

PR #1420, opened 2018-08-31
COLRE-1087 Fix how connections are added and removed

rabbitonweb:
## Overview
Fix how connections are added and removed

CORE-1087

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
@@ -31,28 +31,30 @@ object Connect {
     implicit class ConnectionsOps(connections: Connections) {
 
       def addConn[F[_]: Monad: Log: Metrics](connection: Connection): F[Connections] =
-        connections
-          .contains(connection)
-          .fold(
-            connections.pure[F],
-            Log[F].info(s"Peers: ${connections.size + 1}.").as(connection :: connections) >>= (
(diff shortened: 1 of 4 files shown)


rabbitonweb:
This would have to look like this

```scala
 object Connections {
    def empty: Connections = List.empty[Connection]
    implicit class ConnectionsOps(connections: Connections) {

      def addConn[F[_]: Monad: Log: Metrics](connection: Connection): F[Connections] =
        addConn[F](List(connection))

      def removeConn[F[_]: Monad: Log: Metrics](connection: Connection): F[Connections] =
        removeConn[F](List(connection))

      private def modify[F[_]: Monad: Log: Metrics](
          ids: List[NodeIdentifier],
          calc: Connections => Connections): F[Connections] = {
        val newConnections = connections.partition(peer => ids.contains(peer.id)) match {
          case (_, rest) => calc(rest)
        }
        val size = newConnections.size.toLong
        Log[F].info(s"Peers: $size.") *> Metrics[F].setGauge("peers", size).as(newConnections)
      }

      def addConn[F[_]: Monad: Log: Metrics](toBeAdded: List[Connection]): F[Connections] =
        modify[F](toBeAdded.map(_.id), rest => rest ++ toBeAdded)

      def removeConn[F[_]: Monad: Log: Metrics](toBeRemoved: List[Connection]): F[Connections] =
        modify[F](toBeRemoved.map(_.id), rest => rest)

    }
  }
```

sucks imho :)

PR #1429, opened 2018-09-03
Remove untraced comm event

KentShikama:
## Overview

Paired with @goral09 and @odeac 

We do not need the untraced comm event because if there is a comm event that can happen but is not in the replay logs, then the resulting replay rspace's tuplespace/state hash will not match the hash in the block that contains the replay logs. We need to remove the untraced comm event because of the following case:

            for (@winner <- @"ch") {
              @"return"!(winner)
            } | @"ch"!(1) | @"ch"!(2)

The replay logs say that `for (@winner <- @"ch")` matches with `@"ch"!(1)`. Assume the replay rspace then gets called in the order of 1) `for (@winner <- @"ch")`, 2) `@"ch"!(2)`, and 3) `@"ch"!(1)` . Previously it would complain that `@"ch"!(2)` can match with `for (@winner <- @"ch")` and thus it would generate an untraced comm event. However, `for (@winner <- @"ch")` was just waiting for the `@"ch"!(1)` that was about to come.

Integration tests pass as below

```

collected 4 items                                                                                                                     

test/network/test_basics.py::test_metrics_api_socket PASSED                                                                     [ 25%]
test/network/test_basics.py::test_node_logs_for_errors PASSED                                                                   [ 50%]
test/network/test_basics.py::test_node_logs_for_RuntimeException PASSED                                                         [ 75%]
test/network/test_casper_propose_and_deploy.py::test_casper_propose_and_deploy PASSED                                           [100%]

===================================================== 4 passed in 196.43 seconds ======================================================
```

https://rchain.atlassian.net/browse/RHOL-694

### Notes

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
@@ -357,7 +357,7 @@ class InterpreterUtilTest
     tsHash should be(Some(computedTsHash))
   }
 
-  "validateBlockCheckpoint" should "pass linked list test" in {
+  it should "pass linked list test" in {
     val deploys = Vector(
       """
         |contract @"recursionTest"(@list) = {
@@ -405,7 +405,7 @@ class InterpreterUtilTest
     tsHash should be(Some(computedTsHash))
   }
 
-  "validateBlockCheckpoint" should "pass persistent produce test with causality" in {
+  it should "pass persistent produce test with causality" in {
     val deploys =
       Vector("""new x, y, delay in {
               contract delay(@n) = {
@@ -457,7 +457,7 @@ class InterpreterUtilTest
     tsHash should be(Some(computedTsHash))
   }
 
-  "validateBlockCheckpoint" should "pass tests involving primitives" in {
+  it should "pass tests involving primitives" in {
     val deploys =
       Vector(
         """
@@ -506,7 +506,7 @@ class InterpreterUtilTest
     tsHash should be(Some(computedTsHash))
   }
 
-  "validateBlockCheckpoint" should "pass tests involving pick" in {
+  it should "pass tests involving races" in {
     (0 to 10).foreach { _ =>
       val deploys =
         Vector(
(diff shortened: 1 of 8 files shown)


odeac:
Looking at your changes I don't understand why the crash happened when the list length contained 4,7,11... items. Can you explain that?

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
```diff
@@ -57,20 +57,6 @@ trait ReplayRSpaceTests extends ReplayRSpaceTestsBase[String, Pattern, String, S
       space.store.isEmpty shouldBe true
   }
 
-  "Creating a COMM Event that is not contained in the trace log" should "throw a ReplayException" in
```
pyrocto: This test should still pass.

lukasz-golebiewski: I agree. Why is it being removed?

birchmd:
I think what these changes show is that the logic for detecting an untraced comm event is not right yet. I think rather than removing the logic entirely we should change it so that the checks are deferred to the end and we only need to check
1) The final state is quiescent (if not then any comm events are flagged as untraced)
2) All comm events in the log have already happened.

casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
```diff
@@ -505,4 +505,47 @@ class InterpreterUtilTest
 
     tsHash should be(Some(computedTsHash))
   }
+
+  "validateBlockCheckpoint" should "pass tests involving pick" in {
```
lukasz-golebiewski: I ran this test against latest `dev` and it passes. Can you help me understand why the changes in ReplayRSpace are needed?

lukasz-golebiewski: OK, I see now that it fails randomly

KentShikama:
@birchmd checking at the end is the same thing as just checking if the state hashes match. If there were any untraced comm events, then they wouldn't match.

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
KentShikama: No it just doesn't reduce and thus at the end the state hashes will mismatch.

KentShikama: Think about the case when there is a produce and consume that could reduce but they were originally part of different comm events.

lukasz-golebiewski:
@KentShikama This means that you don't need ReplayRSpace at all, you could use normal RSpace on a different branch

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -161,35 +144,25 @@ class ReplayRSpace[C, P, A, R, K](store: IStore[C, P, A, K], branch: Branch)(
...
-                    store.getData(txn, Seq(c)).zipWithIndex.filter {
-                      case (Datum(_, _, source), _) => comm.produces.contains(source)
-                    }
-                  case None =>
-                    Random.shuffle(store.getData(txn, Seq(c)).zipWithIndex)
+                val as = store.getData(txn, Seq(c)).zipWithIndex.filter {
+                  case (Datum(_, _, source), _) => comm.produces.contains(source)
                 }
-                c -> { if (c == channel) (Datum(data, persist, produceRef), -1) +: as else as }
+                c -> { if (c == channel) Seq((Datum(data, persist, produceRef), -1)) else as }
```
KentShikama: To anyone reviewing: extra care on this change

KentShikama: I'm assuming that `as` is never going to contain any elements if `c == channel`

KentShikama:
@odeac We still don't really know why it only failed on 4, 7, 11, etc.

KentShikama:
@odeac I would say at this point it is just noise / the way memory is jiggered, allowing the produce/consume to enter the replay rspace in a way that raised the untraced comm event.

PR #1431, opened 2018-09-03
CORE-1113 Change behaviour of --thread-pool-size

rabbitonweb:
## Overview
The flag becomes an optional field (both in arguments and in
configuration file). If not provided the default pool size becomes
twice size the core numbers

CORE-1113

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -30,8 +30,10 @@ object Main {
 
     val exec: Task[Unit] =
       for {
-        conf     <- Configuration(args)
-        poolSize = conf.server.threadPoolSize
+        conf <- Configuration(args)
+        poolSize <- conf.server.threadPoolSize.fold(Task.delay {
+                     Runtime.getRuntime().availableProcessors() * 2
+                   })(_.pure[Task])
         //TODO create separate scheduler for casper
         scheduler = Scheduler.fixedPool("node-io", poolSize)
         _         <- Task.unit.asyncBoundary(scheduler)
--- a/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
@@ -57,7 +57,6 @@ object Configuration {
   private val DefaultApprovalProtocolDuration   = 5.minutes
   private val DefaultApprovalProtocolInterval   = 5.seconds
   private val DefaultMaxMessageSize: Int        = 100 * 1024 * 1024
-  private val DefaultThreadPoolSize: Int        = 4000
 
   private val DefaultBootstrapServer: PeerNode = PeerNode
     .parse("rnode://de6eed5d00cf080fc587eeb412cb31a75fd10358@52.119.8.109:40400")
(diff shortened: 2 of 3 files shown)


rabbitonweb:
Comments welcomed :)

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -30,8 +30,10 @@ object Main {
 
     val exec: Task[Unit] =
       for {
-        conf     <- Configuration(args)
-        poolSize = conf.server.threadPoolSize
+        conf <- Configuration(args)
+        poolSize <- conf.server.threadPoolSize.fold(Task.delay {
+                     Runtime.getRuntime().availableProcessors() * 2
+                   })(_.pure[Task])
         //TODO create separate scheduler for casper
```
rabbitonweb: @lukasz-golebiewski mentioned that availableProcessors * 2 is not good heuristic, comments welcomed

goral09: This is a good heuristic AFAIK. *2 if the task is I/O bound but we should test of course.

goral09: It's better than 4000 threads IMO.

lukasz-golebiewski: Is this thread pool used for CPU intensive work or IO? 
https://www.ibm.com/developerworks/library/j-jtp0730/index.html

goral09:
My impression was that we agreed that this is not a solution to our problem. I would even dare to say that this is even worse. Node operating on fixed thread pool with size `# of cores * 2` will either be very slow or not work at all. Rholang interpreter reduces each term in parallel so given 2 cores (4 threads) rholang program with 5 terms run in parallel will blow up the node. 

In my opinion a good start is to have multiple thread pools dedicated to each layer of the node:
1. fork join pool for non-blocking IO -- in practice this would be created with monix `Schedulers.io`. Node should shift to new thread pool when calling Casper and shift back once Casper is done.
2. For Casper layer I would go with cached thread pool since those are mostly CPU intensive tasks (in reality node can't propose two blocks in parallel so only one will be ever handled and Casper executes deploys within the block sequentially). I would also shift to RSpace thread pool when RSpace methods are called.
3. Rholang interpreter runs in parallel so at best it can eat up as many threads as there are terms but since the contention will be RSpace layer it should make use of cooperative yielding of green threads (ie run loop of monix or cats) and consume less threads.
4. Fixed size thread pool of size 1 for RSpace write path -- since we have only 1 global lock this should work (that is my reasoning at least)
5. Cached thread pool for RSpace read path -- since we can multiple read locks it makes sense to have more thread than one. 


Before writing it down I didn't realize how complicated this would become. It makes sense to try to simplify our concurrency model, maybe we should start from the things that are bottlenecks of the program and then build up from there since the whole node is as fast as its slowest part. Shifting (back) to a thread pool can make the code harder to read and I am sure it will be easy to forget about doing that and at some point we will have a bug because one of us forgets to shift to a specialized thread pool (or shift back after computation is finished). Having sad that I am not saying that my proposal is THE solution, I merely want to point out that the problem is more complex and solution for that is not as easy as changing a number of threads from some huge value to a small one.

KellyatPyrofex:
Closing per discussion during 9/14 standup. This work will be addressed in a different ticket and PR.

PR #1433, opened 2018-09-04
Include #1429 (Remove untraced comm event) into v0.6.3

KentShikama:


rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
```diff
@@ -663,57 +649,14 @@ trait ReplayRSpaceTests extends ReplayRSpaceTestsBase[String, Pattern, String, S
...
       replaySpace.consume(channels, patterns, k, persist = false)
-
-      mm.get(cr).map(_.size).value shouldBe 1
-
       replaySpace.consume(channels, patterns, k, persist = false)
-
-      mm.get(cr) shouldBe None
-  }
-
-  "producing" should "correctly remove things from replay data" in withTestSpaces {
```
lukasz-golebiewski: What's the reason for removing this test?

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -292,10 +250,23 @@ class ReplayRSpace[C, P, A, R, K](store: IStore[C, P, A, K], branch: Branch)(
       }
     }
 
-  def createCheckpoint(): Checkpoint = {
-    val root = store.createCheckpoint()
-    Checkpoint(root, Seq.empty)
-  }
+  private def replaysLessCommRef(replays: ReplayData,
+                                 commRef: COMM): MultisetMultiMap[IOEvent, COMM] =
+    commRef.produces.foldLeft(replays.removeBinding(commRef.consume, commRef)) {
```
goral09: Why aren't we doing this for `consumes` as well?

goral09 (changes requested):
If I understand the code correctly we check that there are no `COMM` events left in the replay log. Shouldn't we  also check that we didn't produce _extra_ `COMM` events that are not  in the replay log as well? They won't change the tuplespace hash so the `postStateHash` check would pass.

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
KentShikama: It is already covered by the test above.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
KentShikama: replays.removeBinding(commRef.consume, commRef)

KentShikama:
@goral09 

> They won't change the tuplespace hash so the postStateHash check would pass.

They will

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -84,14 +84,16 @@ object InterpreterUtil {
...
-                  none[StateHash]).leftCast[BlockException] -> knownStateHashes).pure[F]
               case UnknownFailure =>
                 Log[F].warn(s"Found unknown failure") *> (Right(none[StateHash])
                   .leftCast[BlockException] -> knownStateHashes).pure[F]
             }
-
+          case Left((None, status)) =>
+            status match {
+              case UnusedCommEvent(ex: ReplayException) =>
+                Log[F].warn(s"Found unused comm event ${ex.getMessage}") *> (Right(none[StateHash])
```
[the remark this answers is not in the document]
KentShikama: It isn't exhaustive indeed. But it shouldn't be possible to get anything else as of now.

We could just throw

                Log[F].warn(s"Found unknown failure") *> (Right(none[StateHash]).leftCast[BlockException] -> knownStateHashes).pure[F]

in the case it impossibly reaches that Left((None, state which is not UnusedCommEvent))

goral09 (approved):

PR #1436, opened 2018-09-04
Fix non-deterministic IgnorableEquivocation by locking BlockAPI.createBlock

KentShikama:


casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -57,29 +66,6 @@ class HashSetCasperTest extends FlatSpec with Matchers {
     node.tearDown()
   }
 
-  it should "not allow multiple threads to propose a block at the same time" in {
```
goral09: Where did this test go?

KentShikama: It now is redundant with the added test

goral09: If it's redundant, did you remove the lock as well?

KentShikama: As we discussed in person, I have removed the lower level lock indeed

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -57,18 +60,23 @@ object BlockAPI {
       DeployServiceResponse(success = false, "Error: Casper instance not available")
     )
 
-  def createBlock[F[_]: Monad: MultiParentCasperRef: Log]: F[DeployServiceResponse] =
+  def createBlock[F[_]: Sync: Monad: MultiParentCasperRef: Log]: F[DeployServiceResponse] =
```
birchmd: `Monad` not needed with `Sync` since `Sync` extends `Monad`

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -0,0 +1,102 @@
...
+import coop.rchain.p2p.EffectsTestInstances._
+import coop.rchain.casper.MultiParentCasperRef.MultiParentCasperRef
+
+import scala.concurrent.duration._
+import org.scalatest.{FlatSpec, Matchers}
+
+class CreateBlockAPITest extends FlatSpec with Matchers {
+  import HashSetCasperTest._
+  import CasperEffect.Effect
+  implicit val syncId: Sync[Id] = coop.rchain.catscontrib.effect.implicits.syncId
```
birchmd: What is this needed for? In the code we use `Effect` not `Id`.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -17,7 +18,7 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success}
 
 private[api] class DeployGrpcService[
-    F[_]: Monad: Capture: MultiParentCasperRef: Log: Futurable: SafetyOracle: BlockStore](
+    F[_]: Sync: Monad: Capture: MultiParentCasperRef: Log: Futurable: SafetyOracle: BlockStore](
```
birchmd: `Sync`/`Monad` redundancy

node/src/main/scala/coop/rchain/node/api/grpc.scala
```diff
@@ -46,7 +44,7 @@ object GrpcServer {
     }
 
   def acquireExternalServer[
-      F[_]: Capture: Monad: MultiParentCasperRef: Log: SafetyOracle: BlockStore: Futurable](
+      F[_]: Sync: Capture: Monad: MultiParentCasperRef: Log: SafetyOracle: BlockStore: Futurable](
```
birchmd: `Sync`/`Monad` redundancy. Also `Sync`/`Capture` redundancy.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -57,18 +60,23 @@ object BlockAPI {
...
-        for {
-          maybeBlock <- casper.createBlock
-          result <- maybeBlock match {
-                     case err: NoBlock =>
-                       DeployServiceResponse(success = false, s"Error while creating block: $err")
-                         .pure[F]
-                     case Created(block) => casper.addBlock(block).map(addResponse(_, block))
-                   }
-        } yield result,
+        Monad[F].ifM(Sync[F].delay { createBlockLock.tryLock() })(
```
birchmd: Maybe add a TODO item about using `Bracket`. It should be totally trivial to do because `Sync[F]` extends `Bracket[Throwable, F]`, so you already have all the right capabilities. But the issue right now is because of our conflicting `monix`/`cats`/`cats-effect` dependencies the `Task` instance for `Bracket` doesn't work correctly. 

Regardless though, we don't want to get into the situation we were in before where the lock was not being released under some circumstances.

node/src/main/scala/coop/rchain/node/api/grpc.scala
KentShikama: Sync/Capture is redundant?

KentShikama:
Integration tests pass

```
test/network/test_basics.py::test_metrics_api_socket PASSED                                   [ 25%]
test/network/test_basics.py::test_node_logs_for_errors PASSED                                 [ 50%]
test/network/test_basics.py::test_node_logs_for_RuntimeException PASSED                       [ 75%]
test/network/test_casper_propose_and_deploy.py::test_casper_propose_and_deploy PASSED         [100%]

==================================== 4 passed in 201.15 seconds =====================================
```

node/src/main/scala/coop/rchain/node/api/grpc.scala
birchmd: `Sync.delay` is the effectively the same as `Capture.capture`.

KentShikama: Sure but we can't remove capture until we remove atomic monad state

PR #1440, opened 2018-09-10
NO-TICKET: Rholang: Prevent from going into inf loop when printing error.

goral09:
## Overview
Title says it all.

Example stacktrace:
```
Unterminated string on line 14 begining at column 8
	at coop.rchain.rholang.interpreter.Interpreter$.$anonfun$buildAST$2(Interpreter.scala:62)
	at cats.syntax.EitherOps$.leftMap$extension(either.scala:142)
	at coop.rchain.rholang.interpreter.Interpreter$.buildAST(Interpreter.scala:59)
	at coop.rchain.rholang.interpreter.Interpreter$.buildNormalizedTerm(Interpreter.scala:40)
	at coop.rchain.rholang.interpreter.RholangCLI$.processFile(RholangCLI.scala:120)
	at coop.rchain.rholang.interpreter.RholangCLI$.$anonfun$main$1(RholangCLI.scala:55)
	at coop.rchain.rholang.interpreter.RholangCLI$.$anonfun$main$1$adapted(RholangCLI.scala:55)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at coop.rchain.rholang.interpreter.RholangCLI$.main(RholangCLI.scala:55)
	at coop.rchain.rholang.interpreter.RholangCLI.main(RholangCLI.scala)
```

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
@@ -96,7 +96,7 @@ object RholangCLI {
           case Right(par)                 => evaluatePar(runtime)(par)
           case Left(ie: InterpreterError) =>
             // we don't want to print stack trace for user errors
-            Console.err.print(ie.toString)
+            ie.printStackTrace(System.err)
           case Left(th) =>
             th.printStackTrace(Console.err)
         }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
@@ -76,9 +76,10 @@ object errors {
       s"Top level free variables are not allowed: $freeVars."
   }
   final case class SubstituteError(override val toString: String) extends InterpreterError
-  final case class UnrecognizedInterpreterError(throwable: Throwable)
-      extends RuntimeException(throwable)
-      with InterpreterError
+  final case class UnrecognizedInterpreterError(throwable: Throwable) extends InterpreterError {
+    override def getMessage: String = throwable.getMessage
+    override def toString: String   = throwable.getLocalizedMessage
+  }
   final case class SortMatchError(override val toString: String) extends InterpreterError
   final case class ReduceError(override val toString: String)    extends InterpreterError
   final case class MethodNotDefined(method: String, otherType: String) extends InterpreterError {
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
@@ -77,8 +77,7 @@ object errors {
   }
   final case class SubstituteError(override val toString: String) extends InterpreterError
   final case class UnrecognizedInterpreterError(throwable: Throwable) extends InterpreterError {
-    override def getMessage: String = throwable.getMessage
-    override def toString: String   = throwable.getLocalizedMessage
+    override def toString: String   = s"UnrecognizedInterpreterError(${throwable.getMessage})"
   }
   final case class SortMatchError(override val toString: String) extends InterpreterError
   final case class ReduceError(override val toString: String)    extends InterpreterError
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
@@ -1,0 +1,29 @@
+package coop.rchain.rholang.interpreter
+import java.io.StringReader
+
+import coop.rchain.rholang.interpreter.errors.UnrecognizedInterpreterError
+import org.scalatest.{FlatSpec, Matchers}
+
+class ErrorsSpec extends FlatSpec with Matchers {
+
+  "Printing stack trace from rholang interpreter" should "not go into infinite loop causing SOE" in {
+    val invalidRholang = "@\"network!(10)"
+    try {
+      Interpreter.buildNormalizedTerm(new StringReader(invalidRholang)).value
+    } catch {
+      case ex: UnrecognizedInterpreterError =>
+        try {
+          ex.printStackTrace(System.err)
+          assert(true)
+        } catch {
+          case _: StackOverflowError =>
+            // fail the test when SOE is thrown when printing the error stacktrace
+            assert(false)
+        }
+      case _ =>
+        // if error different than expected UnrecognizedInterpreterError was thrown also fail
+        assert(false)
+    }
+  }
+
+}


rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
```diff
@@ -76,9 +76,10 @@ object errors {
       s"Top level free variables are not allowed: $freeVars."
   }
   final case class SubstituteError(override val toString: String) extends InterpreterError
-  final case class UnrecognizedInterpreterError(throwable: Throwable)
-      extends RuntimeException(throwable)
-      with InterpreterError
+  final case class UnrecognizedInterpreterError(throwable: Throwable) extends InterpreterError {
+    override def getMessage: String = throwable.getMessage
+    override def toString: String   = throwable.getLocalizedMessage
+  }
```
ArturGajowy: This will discard the cause chain for the exception and cause debugging harder (not as hard as having the stackoverflow, addmittedly :)). I'd like us to find another way around this (that would retain the cause chain), or have a justification why its not possible.

ArturGajowy: Also, adding a test showing that throwable.cause is set properly would be nice (I should've done that when doing my change, I guess). But I won't insist and will do that myself if you prefer, after this is merged.

ArturGajowy: The stack trace you've pasted doesn't mention the `UnrecognizedInterpreterError` (which I believe it should) and doesn't illustrate that the whole cause chain would be included. Which I think would rely on the cause's toString method. It's not the case when cause is set properly.

`printStackTrace` relies on the cause chain only, and does some more logic than simply calling `toString` for the causes. The only way to have correct stack traces is to set the cause correctly.

You didn't provide a reason why changing that is not possible too.

Also, simply delegating all methods to the `cause` is a very misleading thing to do for a Throwable. Imagine e.g. how the debugger's output would look like.

Please change the `InterpreterError` to be an abstract sealed class, give it constructors with and without the `cause` (delegating to the Throwable's constructors appropriately), and use the one with `cause` for `UnrecognizedInterpreterError`.

goral09:
```
UnrecognizedInterpreterError(Unterminated string on line 14 begining at column 8)
	at coop.rchain.rholang.interpreter.Interpreter$.$anonfun$buildAST$2(Interpreter.scala:62)
	at cats.syntax.EitherOps$.leftMap$extension(either.scala:142)
	at coop.rchain.rholang.interpreter.Interpreter$.buildAST(Interpreter.scala:59)
	at coop.rchain.rholang.interpreter.Interpreter$.buildNormalizedTerm(Interpreter.scala:40)
	at coop.rchain.rholang.interpreter.RholangCLI$.processFile(RholangCLI.scala:120)
	at coop.rchain.rholang.interpreter.RholangCLI$.$anonfun$main$1(RholangCLI.scala:55)
	at coop.rchain.rholang.interpreter.RholangCLI$.$anonfun$main$1$adapted(RholangCLI.scala:55)
	at scala.collection.immutable.List.foreach(List.scala:389)
	at coop.rchain.rholang.interpreter.RholangCLI$.main(RholangCLI.scala:55)
	at coop.rchain.rholang.interpreter.RholangCLI.main(RholangCLI.scala)
```

is this better? It prints the _whole_ stacktrace for given error. Refactoring `InterpreterError` to become an `abstract class` is impossible because of different ways we construct subclasses of `InterpeterError` (from `Throwable`, plain string or contructor arguments ). This would require big engineering effort that is outside of the scope of this PR.

ArturGajowy: It's better.

> It prints the whole stacktrace for given error.

for any error, even. Problem being, the stacktraces are incomplete because the cause field is not being set. Have a look at the implementation of printStackTrace to see how simplified and potentially incompatbile with tools this approach is. For once, any causes of the direct cause will not be printed.

```scala
new RuntimeException("a", new NullPointerException("b")).toString
//res0: String = java.lang.RuntimeException: a
```

> Refactoring InterpreterError to become an abstract class is impossible because of different ways we construct subclasses of InterpeterError (from Throwable, plain string or contructor arguments ). This would require big engineering effort that is outside of the scope of this PR.

I don't get what's the problem here. All non-cause parameters should be interpolated into a string and passed upstream as the `message` parameter for Throwable. I now see that we misuse `toString` (we should be using `message`, or a custom field/method), but we can always make toString delegate to getMessage. Then we should stop using Throwable.toString in a follow-up PR.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.io.StringReader
+
+import coop.rchain.rholang.interpreter.errors.UnrecognizedInterpreterError
+import org.scalatest.{FlatSpec, Matchers}
+
+class ErrorsSpec extends FlatSpec with Matchers {
+
+  "Printing stack trace from rholang interpreter" should "not go into infinite loop causing SOE" in {
+    val invalidRholang = "@\"network!(10)"
+    try {
```
ArturGajowy: Please simplify this to:
```scala
    val e = intercept[UnrecognizedInterpreterError] {
      Interpreter.buildNormalizedTerm(new StringReader(invalidRholang)).value
    }
    noException should be thrownBy {
      e.printStackTrace(System.err)
    }
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
```diff
@@ -96,7 +96,7 @@ object RholangCLI {
           case Right(par)                 => evaluatePar(runtime)(par)
           case Left(ie: InterpreterError) =>
             // we don't want to print stack trace for user errors
```
lukasz-golebiewski: Is this comment still valid?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
```diff
@@ -96,7 +96,7 @@ object RholangCLI {
...
-            Console.err.print(ie.toString)
+            ie.printStackTrace(System.err)
```
lukasz-golebiewski: Why not `Console.err` for coherence?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
goral09: It is, my bad. Thanks for catching that.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+      Interpreter.buildNormalizedTerm(new StringReader(invalidRholang)).value
+    } catch {
+      case ex: UnrecognizedInterpreterError =>
+        try {
+          ex.printStackTrace(System.err)
+          assert(true)
```
lukasz-golebiewski: No need to do this

rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+        } catch {
+          case _: StackOverflowError =>
+            // fail the test when SOE is thrown when printing the error stacktrace
+            assert(false)
```
lukasz-golebiewski: `fail()` with a descriptive message. Comments tend to outdate very quickly

rholang/src/test/scala/coop/rchain/rholang/interpreter/ErrorsSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+        }
+      case _ =>
+        // if error different than expected UnrecognizedInterpreterError was thrown also fail
+        assert(false)
```
lukasz-golebiewski: Same here

rholang/src/main/scala/coop/rchain/rholang/interpreter/errors.scala
```diff
@@ -76,9 +76,9 @@ object errors {
       s"Top level free variables are not allowed: $freeVars."
   }
   final case class SubstituteError(override val toString: String) extends InterpreterError
-  final case class UnrecognizedInterpreterError(throwable: Throwable)
-      extends RuntimeException(throwable)
-      with InterpreterError
+  final case class UnrecognizedInterpreterError(throwable: Throwable) extends InterpreterError {
+    override def toString: String   = s"UnrecognizedInterpreterError(${throwable.getMessage})"
```
lukasz-golebiewski: Does this need to be a `def`?

goral09:
Closing in favor of #1450

PR #1462, opened 2018-09-11
Refactor HashSetCasperTestNode to allow any F

birchmd:
## Overview
This enables setting up a network of nodes which use `Task` and thus can act in parallel. It also removes the code duplication present in `helper.CasperEffect` vs `helper.HashSetCasperTestNode`.

NO-TICKET

### Notes


casper/src/test/scala/coop/rchain/casper/util/comm/TransportLayerTestImpl.scala
```diff
@@ -12,37 +13,55 @@ import scala.concurrent.duration.FiniteDuration
 import scala.collection.mutable
 import coop.rchain.comm.transport._
 
-class TransportLayerTestImpl[F[_]: Monad: Capture](
+class TransportLayerTestImpl[F[_]: Monad](
     identity: PeerNode,
-    val msgQueues: collection.Map[PeerNode, mutable.Queue[Protocol]])
+    val msgQueues: Map[PeerNode, Ref[F, mutable.Queue[Protocol]]])
```
goral09: If you don't plan to modify the `msgQueues` from the outside of `TransportlayerImpl` then you shouldn't need `Ref`. Am I missing something?

birchmd: The whole point of the reactor is to enable concurrent unit tests involving a network of nodes (in the same style as as existing unit tests involving race conditions on a single node). There is one set of queues for the whole network, so there could be concurrent modifications as one node sends a message at the same time as the recipient receives a different message.

goral09: I didn't notice that in the code. Thanks for explaining.

birchmd: To be fair, I haven't pushed an example of using it in that way yet. I'm trying to write a test that reproduces Rhol-722, but have not been successful yet.

PR #1463, opened 2018-09-11
RHOL-11: Change return type of RSpace match from Option to Either.

goral09:
## Overview
Refactors `rspace`'s `Match` typeclass from:
```scala
trait Match[P, A, R] {
  def get(p: P, a: A): Option[R]
}
```
to
```scala
trait Match[P, E, A, R] {
  def get(p: P, a: A): Either[E, Option[R]]
}
```
This change turned out to be necessary when wiring cost accounting with the RSpace lookups.  `RSpace` uses `Rholang`'s _spatial matcher_ ,as an implementation of `Match` typeclass, to find a match between receive patterns and targets. Every comparison done by the _spatial matcher_ is being charged for. We want to short circuit as soon as deploy runs out of phlogistons. Initial return type of `Match` - `Option[R]` - indicates only whether the match was found or not, there is no way to propagate the information about running out of phlogistons. New return type `Either[E, Option[R]]` still carries the information as previous implementation (in the form of `Right(Option[R])`) while `Left(E)` can represent any type of invalid state. At the moment it is just `OutOfPhlogistonError` and `Nothing` when we don't expect invalid states to occur.

Note:
I have also considered using different return type for `A` in `Option[A]` , one that could also represent invalid state, but this wouldn't be a good solution for various of reasons:
1. Internal logic of `RSpace` treats `None` as no match found and saves the data (or continuation) in the tuplespace.
2. `Option` with its two states is too constrained to represent three states:
-- match found
-- no match found
-- deploy runs out of phlo

All the logic of RSpace is retained.

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
@@ -10,7 +10,7 @@ import coop.rchain.rholang.interpreter.matcher._
 import OptionalFreeMapWithCost._
 import coop.rchain.models.rholang.implicits._
 import coop.rchain.rholang.interpreter.accounting.CostAccount
-import coop.rchain.rholang.interpreter.errors.OutOfPhloError
+import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 import coop.rchain.rspace.{Serialize, Match => StorageMatch}
 
 //noinspection ConvertExpressionToSAM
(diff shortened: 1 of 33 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
```diff
@@ -26,15 +27,16 @@ object implicits {
     }
 
   implicit val matchListQuote
-    : StorageMatch[BindPattern, ListChannelWithRandom, ListChannelWithRandom] =
-    new StorageMatch[BindPattern, ListChannelWithRandom, ListChannelWithRandom] {
+    : StorageMatch[BindPattern, OutOfPhloError.type, ListChannelWithRandom, ListChannelWithRandom] =
```
lukasz-golebiewski: Why not `InterpreterError`? Maybe we should introduce `RholangRuntimeError` or something similar?

goral09: Because it is the only error that matching can fail with.

rspace/src/test/scala/coop/rchain/rspace/TestImplicitHelpers.scala
```diff
@@ -0,0 +1,14 @@
...
+import cats.Id
+import org.scalatest.enablers.Definition
+
+//noinspection ConvertExpressionToSAM
+trait TestImplicitHelpers {
+  // Some helpers for usage only in the tests -- save us A LOT of explicit casting from Either to Option
+  // it is safe because left type of `Either` is `Nothing` -- we don't expect any invalid states from the matcher
+  implicit def eitherDefinitionScalatest[E, A]: Definition[Id[Either[E, Option[A]]]] =
+    new Definition[Id[Either[E, Option[A]]]] {
+      override def isDefined(thing: Id[Either[E, Option[A]]]): Boolean =
```
lukasz-golebiewski: Is this used anywhere?

goral09: Yes, everywhere there is `_sth_ isDefined` where type of `_sth_` is `Id[Either[Nothing, Option[_]]]` .

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
lukasz-golebiewski: Can anything aside from the matcher fail with an OutOfPhloError?

goral09:
@lukasz-golebiewski I did sth fishy with this PR (wrong rebase). I need to clean up, give me a moment.

PR #1464, opened 2018-09-12
[WIP] Enhance block distribution

itegulov:
## Overview
Enhances the casper's block distribution by keeping track of the distance from the creator of a block. The rate at which the number of peers whom the block is being sent decreases is configurable (the default is 2, i.e. the neighbors will send the block to the half of its peers, their neighbors will send the block to the quarter of its peers etc). There is also a minimal number of peers whom the block is being sent (the default is 1).

https://rchain.atlassian.net/browse/RHOL-567


models/src/main/protobuf/CasperMessage.proto
```diff
@@ -185,7 +185,8 @@ message BlockMessage {
   bytes                  sig            = 7; // signature generated by signing `hash(hash(justification) concat blockHash)`.
   string                 sigAlgorithm   = 8; // name of the algorithm used to sign
   string                 shardId        = 9; // identifier of the shard where the block was created
-  bytes extraBytes = 10;
+  int32                  creatorDist    = 10; // distance to the creator of the block
```
KentShikama: I think it will be confusing if we start placing things that aren't included in the blockHash in BlockMessage. We should probably move creatorDist and the sig out into a different message.

birchmd:
So looking at this PR, I think that this implements the proposed algorithm correctly, but I am concerned about it. We are still seeing problems with block distribution in the real test net today, despite the redundancy of block passing. I think that we should handle the problem of blocks apparently not being passed effectively around the network first and then handle the redundancy afterwards.

itegulov:
@birchmd I was afraid that would be the case. I will close the PR then for now and will create a new one when we will be ready to worry about redundancy.

PR #1468, opened 2018-09-13
Remove outdated Homebrew instructions

adaszko:


README.md
```diff
@@ -177,16 +177,7 @@ TBD
 
 ### Installing and running on macOS via Homebrew
```
odeac: probably you have to remove this as well?

Or add a TBD after it, just like for the other incomplete sections

adaszko: Good catch.  Corrected.

KellyatPyrofex:
@rabbitonweb, please confirm this change related to Mac users and Homebrew is correct.

README.md
rabbitonweb: Adam just remove the whole Homebrew section from docs altogether

adaszko:
@TomVasile Added a [Homebrew tap](https://github.com/rchain/homebrew-rchain/)

PR #1474, opened 2018-09-14
Switch to Java 10

adaszko:
https://rchain.atlassian.net/browse/OPS-232


ArturGajowy:
The build... passed!?

ArturGajowy:
I think I tried sth similar, but it failed on me. Maybe because I was testing hydra at that time?

adaszko:
> I think I tried sth similar, but it failed on me. Maybe because I was testing hydra at that time?

This PR includes a fix for a problem with certificates.  Perhaps that's what you've experienced.

PR #1480, opened 2018-09-14
Crypto: Blake2b512Random: Use 256-bit internal node size.

guardbotmk3:
## Overview
We're using 256-bit names, so it makes sense for the internal node size to be
256 bits. If someone can reliably create collisions in the first 256 bits of
output, they can create collisions on names.

### DIFF
--- a/crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Block.scala
+++ b/crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Block.scala
@@ -59,7 +59,9 @@ class Blake2b512Block {
                        outOffset: Int): Unit = {
     val tempChainValue: Array[Long] = new Array[Long](8)
     compress(block, inOffset, tempChainValue, true, true, false)
-    Pack.longToLittleEndian(tempChainValue, output, outOffset)
+    for (i <- 0 until 4) {
+      Pack.longToLittleEndian(tempChainValue(i), output, outOffset + i * 8)
+    }
   }
 
   override def equals(other: Any): Boolean =
@@ -227,8 +229,8 @@ object Blake2b512Block {
   val DIGEST_LENGTH_BYTES: Int = 64
   // Depth = 255, Fanout = ??, Keylength = 0, Digest length = 64 bytes
   val PARAM_VALUE_0: Long = 0xff000040L
-  // Inner length = 64 bytes
-  val PARAM_VALUE_2: Long = 0x4000L
+  // Inner length = 32 bytes
+  val PARAM_VALUE_2: Long = 0x2000L
 
   // This will give invalid results and is for testing only.
   def tweakT0(src: Blake2b512Block): Unit =
--- a/crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
+++ b/crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
@@ -142,6 +142,7 @@ object Blake2b512Random {
     apply(init, 0, init.length)
 
   def merge(children: Seq[Blake2b512Random]) = {
+    import coop.rchain.crypto.codec.Base16
     @tailrec
     def internalMerge(children: Vector[Blake2b512Random]): Blake2b512Random = {
       val squashedBuilder = Vector.newBuilder[Blake2b512Random]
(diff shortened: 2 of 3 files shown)


models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -494,4 +494,42 @@ object implicits {
           case _ => BitSet()
         }
     }
+
+  implicit val ParIsForgeable: IsForgeable[Par] =
```
ArturGajowy: I think those are unintended changes, and the CI build seems to agreee :)

guardbotmk3: I've fixed it so those aren't there.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
```diff
@@ -150,12 +151,12 @@ object Blake2b512Random {
...
         squashedBuilder += result
-        slice.grouped(2).foreach { pair =>
-          pair(0).digest.finalizeInternal(pair(0).lastBlock.array(), 0, chainBlock, 0)
-          if (pair.size > 1) {
-            pair(1).digest.finalizeInternal(pair(1).lastBlock.array(), 0, chainBlock, 64)
-          } else {
-            BLANK_HASH.asReadOnlyBuffer().get(chainBlock, 64, 64)
+        slice.grouped(4).foreach { quad =>
+          for (i <- 0 until quad.size) {
+            quad(i).digest.finalizeInternal(quad(i).lastBlock.array(), 0, chainBlock, i * 32)
```
ArturGajowy: So previously we were
- merging 2 * 128 byte lastBlock-s into one 128-byte chainBlock,
  by compressing each of the lastBlock-s into an 8-long = 64-byte array
  and appending each of the compressed results to the chainBlock

and now we are 
- merging 4 * 128 byte lastBlock-s into one 128-byte chainBlock,
  by compressing each of the lastBlock-s into an 8-long = 64-byte array
  and appending half of each of compressed results' bytes to the chainBlock

I guess this is OK, because the result of compression is a uniform hash? (is it?)

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Block.scala
```diff
@@ -59,7 +59,9 @@ class Blake2b512Block {
                        outOffset: Int): Unit = {
     val tempChainValue: Array[Long] = new Array[Long](8)
     compress(block, inOffset, tempChainValue, true, true, false)
-    Pack.longToLittleEndian(tempChainValue, output, outOffset)
+    for (i <- 0 until 4) {
+      Pack.longToLittleEndian(tempChainValue(i), output, outOffset + i * 8)
```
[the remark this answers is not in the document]
guardbotmk3: 1. Current is 256 *Bits* or 32 Bytes. Previous was 512 Bits, or 64 Bytes.
2. All 8 longs (64 Bytes) are computed by compress, and we pass out the first 4.
4. Because the names or only 256 bits anyway, so we don't need the security level provided by 512 bit internal nodes.
5. Because the output array may now be smaller. We have shrunk the internal node size.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Block.scala
```diff
@@ -227,8 +229,8 @@ object Blake2b512Block {
   val DIGEST_LENGTH_BYTES: Int = 64
   // Depth = 255, Fanout = ??, Keylength = 0, Digest length = 64 bytes
   val PARAM_VALUE_0: Long = 0xff000040L
-  // Inner length = 64 bytes
-  val PARAM_VALUE_2: Long = 0x4000L
+  // Inner length = 32 bytes
+  val PARAM_VALUE_2: Long = 0x2000L
```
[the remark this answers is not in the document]
guardbotmk3: In the parameters? Nothing else. Only the internal node size has changed.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
guardbotmk3: Previously we were merging 2 * 64 byte outputs into one 128 byte input.
Now we merge 3 * 32 byte outputs into one 128 byte input.

ArturGajowy:
@guardbotmk3 Thx, all my questions cleared now :)

PR #1482, opened 2018-09-16
Fix swallow exceptions in finally

tymm:
Currently when using `withResource` there is a chance that exceptions in the `try` block get swallowed by an exception thrown in `finally`.
See: https://medium.com/@dkomanov/scala-try-with-resources-735baad0fd7d

### DIFF
--- a/shared/src/main/scala/coop/rchain/shared/Resources.scala
+++ b/shared/src/main/scala/coop/rchain/shared/Resources.scala
@@ -12,8 +12,8 @@ object Resources {
     * @param a A given resource implementing [[AutoCloseable]]
     * @param f A function that takes this resource as its argument
     */
-  def withResource[A <: AutoCloseable, B](r: => A)(f: A => B): B = {
-    val resource: A = r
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/coop/rchain/shared/Resources.scala
```diff
@@ -11,10 +12,30 @@ object Resources {
     * @param a A given resource implementing [[AutoCloseable]]
     * @param f A function that takes this resource as its argument
     */
-  def withResource[A <: AutoCloseable, B](a: A)(f: A => B): B =
+  def withResource[A <: AutoCloseable, B](r: => A)(f: A => B): B = {
```
ArturGajowy: Would be great to have unit tests for that. Would you mind adding them in this PR? If you'd prefer not, please file a ticket for adding them.

tymm: Sorry, sure. Will add one.

PR #1481, opened 2018-09-15
Fix bug in test, add helper constructor, refactor PCost class.

goral09:
## Overview
This PR consists of bunch of changes I made during work on RHOL-11. 
1. Add helper constructor to `PureRSpace` - it's not necessary for all type annotations.
2. Fix `CompilerTests` - it didn't check for the (non)existence of errors in the execution of rholang program.
3. Change order of `idx` and `cost` fields in `PCost` to be consistent with `CostAccount`.
4. Move `tests` and `failure_tests` folders under `resource` directory (where they should be anyway) to allow for running the `CompilerTests` from intellij.

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
Add link to corresponding JIRA issue.

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [ ] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [ ] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1483, opened 2018-09-16
RHOL-11: Refactor CostAccountingAlg.

goral09:
## Overview
Relative to #1481 (start with that one)

This PR contains a subset of the changes I've made in the course or RHOL-11. Specifically this PR:
+ explicitly pass initial phlos when running spatial match  - this is preliminary work for deducting the phlos from the available and short-circuiting the matcher
+ simplifies API of the `CostAccountingAlg` - turns out it can be much simpler and still provide necessary functionality
+ adds a test for `CostAccountingAlg`
+ replaces `modifyCost(f: CostAccount => CostAccount)` method with `charge(c: Cost)`

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -9,15 +9,9 @@ import coop.rchain.models.Par
 import coop.rchain.models.rholang.implicits.VectorPar
 import coop.rchain.models.rholang.sort.Sortable
 import coop.rchain.rholang.interpreter.accounting.{CostAccount, CostAccountingAlg}
-import coop.rchain.rholang.interpreter.errors.{
-  InterpreterError,
-  SyntaxError,
-  TopLevelFreeVariablesNotAllowedError,
-  TopLevelWildcardsNotAllowedError,
-  UnrecognizedInterpreterError
-}
+import coop.rchain.rholang.interpreter.errors.{InterpreterError, SyntaxError, TopLevelFreeVariablesNotAllowedError, TopLevelWildcardsNotAllowedError, UnrecognizedInterpreterError}
 import coop.rchain.rholang.syntax.rholang_mercury.Absyn.Proc
-import coop.rchain.rholang.syntax.rholang_mercury.{parser, Yylex}
+import coop.rchain.rholang.syntax.rholang_mercury.{Yylex, parser}
 import monix.eval.{Coeval, Task}
 
 private class FailingTask[T](task: Task[Either[Throwable, T]]) {
@@ -99,7 +93,7 @@ object Interpreter {
       costAccounting <- CostAccountingAlg[Task](CostAccount.zero)
       _              <- runtime.reducer.inj(normalizedTerm)(rand, costAccounting)
       errors         <- Task.now(runtime.readAndClearErrorVector())
-      cost           <- costAccounting.getCost()
+      cost           <- costAccounting.get()
       _              <- Task.now(if (errors.nonEmpty) runtime.space.reset(checkpoint.root))
     } yield EvaluateResult(cost, errors)
   }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -90,7 +90,7 @@ object Interpreter {
     implicit val rand = Blake2b512Random(128)
     for {
       checkpoint     <- Task.now(runtime.space.createCheckpoint())
-      costAccounting <- CostAccountingAlg[Task](CostAccount.zero)
+      costAccounting <- CostAccountingAlg.of[Task](CostAccount.zero)
       _              <- runtime.reducer.inj(normalizedTerm)(rand, costAccounting)
       errors         <- Task.now(runtime.readAndClearErrorVector())
       cost           <- costAccounting.get()
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -9,9 +9,15 @@ import coop.rchain.models.Par
 import coop.rchain.models.rholang.implicits.VectorPar
 import coop.rchain.models.rholang.sort.Sortable
 import coop.rchain.rholang.interpreter.accounting.{CostAccount, CostAccountingAlg}
-import coop.rchain.rholang.interpreter.errors.{InterpreterError, SyntaxError, TopLevelFreeVariablesNotAllowedError, TopLevelWildcardsNotAllowedError, UnrecognizedInterpreterError}
+import coop.rchain.rholang.interpreter.errors.{
+  InterpreterError,
+  SyntaxError,
+  TopLevelFreeVariablesNotAllowedError,
+  TopLevelWildcardsNotAllowedError,
+  UnrecognizedInterpreterError
+}
 import coop.rchain.rholang.syntax.rholang_mercury.Absyn.Proc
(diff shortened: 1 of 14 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Registry.scala
```diff
@@ -95,36 +95,47 @@ class Registry(private val space: Runtime.RhoPureSpace,
       freeCount = 2))
 
   private val publicRegisterRandomRef: Long = Runtime.BodyRefs.REG_PUBLIC_REGISTER_RANDOM
-  private val publicRegisterInsertCallbackRef: Long = Runtime.BodyRefs.REG_PUBLIC_REGISTER_INSERT_CALLBACK
+  private val publicRegisterInsertCallbackRef: Long =
+    Runtime.BodyRefs.REG_PUBLIC_REGISTER_INSERT_CALLBACK
   private val publicRegisterRandomChannels = List(
     Channel(Quote(GPrivate(ByteString.copyFrom(Array[Byte](18))))))
+  // format: off
```
ArturGajowy: I take those fix the 'scalafmt has blown up' errors? :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingAlg.scala
```diff
@@ -1,44 +1,39 @@
...
+import cats.{FlatMap, Monad}
+import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 
 trait CostAccountingAlg[F[_]] {
   def charge(cost: Cost): F[Unit]
-  def modify(f: CostAccount => CostAccount): F[Unit]
-  def getCost(): F[CostAccount]
-  def setCost(cost: CostAccount): F[Unit]
+  def get(): F[CostAccount]
+  def set(cost: CostAccount): F[Unit]
```
ArturGajowy: This really should just be `def charge(cost: CostAccount) : F[Unit]`.
The only use case for set is initialization before running the computation.
The only use case for get is getting the result past the computation.
All other occurrences of get & set are equivalent to ~`.charge(costAlg.get)`

To satisfy the init + get result use cases, we should make so that the CostAccountingAlg instances do not hide the `Ref`-s internally, but use a ref that's provided by the code creating the instance. Notice this is not leaking internals, as the ref's state is fully accessible via the .get .set right now. Quite the contrary - instantiating based on a passed ref will allow **limiting** the places where the Ref's state can be accessed to just the end of the world and the implementation of `charge`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingAlgSpec.scala
```diff
@@ -0,0 +1,79 @@
...
+      _ <- alg.set(negativeAccount)
+      _ <- alg.charge(cost)
+      s <- alg.get()
+    } yield s
+
+    val res = Await.result(test.attempt.runAsync, 1.second)
+    assert(res === Left(OutOfPhlogistonsError))
+  }
+
+  override protected def withFixture(
```
ArturGajowy: I think this is unused

ArturGajowy: OMG no this is sorcery :D

ArturGajowy: While it's good to know there's such a facility, after limiting the accessibility of ref state we won't be able to use that. Unless there's a way to specify the outcome in test and verify it in withFixture? Anyway, that's just an impl detail for the tests.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -17,11 +17,12 @@ package object matcher {
   object OptionalFreeMapWithCost {
 
     class OptionalFreeMapWithCostOps[A](s: OptionalFreeMapWithCost[A]) {
-      def modifyCost(f: CostAccount => CostAccount): OptionalFreeMapWithCost[A] =
+      def charge(amount: Cost): OptionalFreeMapWithCost[A] =
```
ArturGajowy: I like this change very much! :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingAlg.scala
ArturGajowy: That's a part of what I meant, and the reason why we can make the types guard that :)

goral09: I don't understand why `Ref` would be better.

ArturGajowy: Because, currently in every part of spatial matcher or other place whrere I have CostAccountAlg, I can break the cost account by using set.

When we only have `CostAccountAlg.charge(Cost|CostAccount)` in the interface, this is no longer possible.

This is Interface Segragation Principle if you will: we separate the initialization capabilities (which can break the accounting!) from charging capabilities (which are the only that are needed where CostAccountAlg is used). The place where it's created and the costs are needed to be read is **another** usage site, with different use cases, which demands a different API.

Pardon the OOPsplanation ;)

goral09: Makes sense, I will see what I can do.

ArturGajowy: I've noticed there will be some churn around marrying CostCaccountAlg and SpatialMatcher (which both do cost accounting...). But not that much:
We'll simply expose a '.get' method on CostAccountAlg and will use 'charge' instead of 'set'. Which happens to be naturally supported by accumulating cost towards limit (vs subtracting from a limit) - we simply get the 'cost so far' from the matcher and do costAccountingAlg.charge(matcherCost)

goral09:
@ArturGajowy I've addressed your comments. PTAL

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -90,7 +90,7 @@ object Interpreter {
     implicit val rand = Blake2b512Random(128)
     for {
       checkpoint     <- Task.now(runtime.space.createCheckpoint())
-      costAccounting <- CostAccountingAlg[Task](CostAccount.zero)
+      costAccounting <- CostAccountingAlg.of[Task](CostAccount.zero)
```
ArturGajowy: :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingAlg.scala
```diff
@@ -1,44 +1,43 @@
...
-  def apply[F[_]: Sync](init: CostAccount): F[CostAccountingAlg[F]] = Ref[F].of(init).map { state =>
-    new CostAccountingAlg[F] {
-      override def charge(cost: Cost): F[Unit]                    = state.update(_.charge(cost))
-      override def modify(f: CostAccount => CostAccount): F[Unit] = state.update(f)
-      override def getCost: F[CostAccount]                        = state.get
-      override def setCost(cost: CostAccount): F[Unit]            = state.set(cost)
-    }
-  }
+  def of[F[_]: Sync](init: CostAccount): F[CostAccountingAlg[F]] =
+    Ref[F].of(init).map(ref => new CostAccountingAlgImpl[F](ref))
```
ArturGajowy: Turns out the API change was even easier than I thought! :tada: And the Ref stays inside thx to the (harmless, admittedly) `.get`. Feels almost like cheating :D Good work!

PR #1484, opened 2018-09-16
RHOL-11: Short circuit spatial matcher when deployment runs out of phlo.

goral09:
## Overview
Relative to #1483 (start with that one)
Short circuit spatial matcher when deployment runs out of phlo. Also, we now subtract phlos used from the initial available phlos.

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
@@ -3,7 +3,7 @@ package coop.rchain.rholang.interpreter
 import cats.arrow.FunctionK
 import cats.data.{OptionT, State, StateT}
 import coop.rchain.models.Par
-import coop.rchain.rholang.interpreter.accounting.CostAccount
+import coop.rchain.rholang.interpreter.accounting.{Cost, CostAccount}
 
 import scala.collection.immutable.Stream
 
@@ -17,11 +17,12 @@ package object matcher {
   object OptionalFreeMapWithCost {
 
     class OptionalFreeMapWithCostOps[A](s: OptionalFreeMapWithCost[A]) {
-      def modifyCost(f: CostAccount => CostAccount): OptionalFreeMapWithCost[A] =
+      def charge(amount: Cost): OptionalFreeMapWithCost[A] =
         StateT((m: FreeMap) => {
           OptionT(State((c: CostAccount) => {
             val (cost, result) = s.run(m).value.run(c).value
-            (f(cost), result)
+            val newCost        = cost + amount
+            (newCost, result)
           }))
         })
 
@@ -88,11 +89,12 @@ package object matcher {
 
   object NonDetFreeMapWithCost {
     class NonDetFreeMapWithCostOps[A](s: NonDetFreeMapWithCost[A]) {
-      def modifyCost(f: CostAccount => CostAccount): NonDetFreeMapWithCost[A] =
+      def charge(amount: Cost): NonDetFreeMapWithCost[A] =
         StateT((m: FreeMap) => {
           StreamT(State((c: CostAccount) => {
             val (cost, result) = s.run(m).value.run(c).value
-            (f(cost), result)
+            val newCost        = cost + amount
+            (newCost, result)
           }))
         })
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
@@ -1,9 +1,11 @@
 package coop.rchain.rholang.interpreter
 
 import cats.arrow.FunctionK
-import cats.data.{OptionT, State, StateT}
+import cats.data.{OptionT, StateT}
+import cats.implicits._
 import coop.rchain.models.Par
 import coop.rchain.rholang.interpreter.accounting.{Cost, CostAccount}
+import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 
 import scala.collection.immutable.Stream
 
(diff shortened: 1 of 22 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -1,47 +1,57 @@
...
 
 package object matcher {
 
   type FreeMap = Map[Int, Par]
 
-  type OptionalFreeMapWithCost[A] = StateT[OptionWithCost, FreeMap, A]
-  type OptionWithCost[A]          = OptionT[State[CostAccount, ?], A]
+  type OptionalFreeMapWithCost[A] =
+    StateT[OptionWithCost, FreeMap, A]
+  type OptionWithCost[A] = OptionT[StateT[Either[OutOfPhlogistonsError.type, ?], CostAccount, ?], A]
```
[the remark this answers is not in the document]
ArturGajowy: I need to figure out how to best pass the limit, but I guess adding a ReaderT[CostLimit, ...] on top of what we have would not be a crime given that this review is to reduce the transformer stack height by 1 from what you intended to do ;)

Another option to consider would be to put the limit into State[CostAccount, and maybe make sure the limit can't be constructed far from the world's edge.

Yet another option is to utilize the RWST transforemr:
 - cost limit is a reader
 - cost account is actually a writer
 - state could be empty

So we'd have
``` scala
  /**
   * Represents a stateful computation in a context `F[_]`, over state `S`, with an
   * initial environment `E`, an accumulated log `L` and a result `A`.
   */
  type ReaderWriterStateT[F[_], E, L, S, A] = IndexedReaderWriterStateT[F, E, L, S, S, A]
```
E = CostLimit
L = CostAccount
S = Unit
F = OptionT[State[FreeMap  or  StreamT[State[FreeMap

So, still just 3 levels of transformer stack, but with all the features we need.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
```diff
@@ -43,18 +43,19 @@ object TuplespaceAlg {
           case Right(Some((continuation, dataList))) =>
             val rspaceMatchCost =
               dataList
-                .map(_.cost.map(CostAccount.fromProto(_)).getOrElse(CostAccount.zero))
+                .map(_.cost.map(CostAccount.fromProto(_)).getOrElse(CostAccount(0)))
                 .toList
+                .map(ca => ca.copy(cost = Cost(Integer.MAX_VALUE) - ca.cost))
```
ArturGajowy: The 'fishy' comment should go here ^

> This looks veeeeery fishy and suggests to me we should actually be adding cost and have an explicit limit to compare against. What does it even mean to subtract cost from an arbitrary value? thinking

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
```diff
@@ -78,19 +79,23 @@ object TuplespaceAlg {
               case Right(Some((continuation, dataList))) =>
                 val rspaceMatchCost =
                   dataList
-                    .map(_.cost.map(CostAccount.fromProto(_)).getOrElse(CostAccount.zero))
+                    .map(
+                      _.cost
+                        .map(CostAccount.fromProto(_))
+                        .getOrElse(CostAccount(0)))
                     .toList
+                    .map(ca => ca.copy(cost = Cost(Integer.MAX_VALUE) - ca.cost))
```
ArturGajowy: And here (fishy!)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
goral09: An arbitrary value is the phlos that a user is willing to pay when deploying the code.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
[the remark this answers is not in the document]
goral09: I think this was resolved during the call.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/TuplespaceAlg.scala
goral09: Resolved on a call.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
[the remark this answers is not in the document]
ArturGajowy: Not quite:)

1. we didn't discuss if the additional OptionT (besides the Either/EitherT) is needed. To me, there's no harm in allowing access to the final cost after the computation short-circuited. I'd even wager it'd be needed or useful at least. In contrast, I see much more harm in adding yet another layer of monad transformers.

2. We also didn't discuss whether we need the Either at all (see my 2nd comment). Now I've noticed that we can't simply replace OptionT with EitherT, because we need to do the same for StreamT - so we'd either (:D) need a StreamErrT, or Either under Option as you did. Given the fact that `Either[OutOfPhlo.type, A]` is isomorphic to `Option[A]` (b/c OutOfPhlo.type only has one value), I'm getting more convinced that we shouldn't change the effect stack in the above respects at all - only its behavior in the charge method.

~Also, in my view, the Either should be an EitherT in place of the previously used OptionT.~

3. We discussed whether charging should go up/down with the cost and honestly I'm not convinced charging down is easier/better - you'd need to show me the places where charging down would actually be easier. Notice both `CostAccountingAlg` and `*FreeMapWithCost` lend easily to having the `limit` parameter passed in a reader-like manner (the first by passing it as a parameter similarly to how we discussed for the Ref, the second - by easily migrating to RWST). So far I haven't seen a single line in use-site code that would benefit from charging down, and I've seen plenty that would not be changed if we charged up. You'd need to show me the places where charging down makes life easier.

4. Lastly, whether we decide to charge up or down, it'd be good service to all of us if we employed the RWST and acknowledged that CostAccount is a Monoid, and charge actually is Writer, not State. It'd be a nice two birds with one stone if we settled on charging up.

PR #1489, opened 2018-09-17
CORE-1150 Replace KademliaRPC implementation with dedicate solution

rabbitonweb:
## Overview

`----- important -----`

The *main* change from the user perspective is the change of the rnode address. 

BEFORE: `rnode://cb74ba04085574e9f0102cc13d39f0c72219c5bb@bootstrap.rchain.coop:40400`

AFTER:
`rnode://cb74ba04085574e9f0102cc13d39f0c72219c5bb@bootstrap.rchain.coop?protocol=40400&discovery=40404`

The change is due to the fact that this change brings a *separate* communication abstraction just for Kademlia. This requires a separate port (known here has `--kademlia-port`)

`------------------------`

After this change KademliaRPC will not be using TransportLayer as an
abstraction anymore. Instead it will get its own dedicated soltion
that is based on blocking grpc implementation.

Motivation for those changes are following:

1. TransportLayer exposes a lot of abstraction (send, roundtrip,
broadcast) which only a fraction (roundTrip) is used by KademliaRPC

2. Due to extensivens of features that are required by layers that are
using TransportLayer, it became a very complex solution. We anticipate
it will become more complex in upcoming months (queues, streaming
content etc) - for a simple need of KademliaRPC this becomes
overengineered

3. KademliaRPC does not require TLS

4. KademliaRPC potentially could go back to use UDP, as it was
originally intended by the authors of Kademlia paper

Becasue KademliaRPC needs will use its own stack for communication, seperate port is required. Configurable both in config file and via flags by `kademlia-port`. Default value is 40404.

CORE-1150


KellyatPyrofex:
I don't have concerns about the cosmetics of the bootstrap address as long as it does look so different to end users that it creates confusion. 

Please do make a ticket to update the RNode README to bring in the information in the PR overview. This will help to re-educate end-users. 

Also, as a reminder from the conversation in standup, please post to Discord when you merge this so others who may be using dev can update their bootstrap addresses.

rabbitonweb:
@woky as stated during standup, during discussion on discord: the whole point was to raise the awareness that this looks weird and maybe we should change the format

rabbitonweb:
In other words: I'm more then happy to change the format, that was my original question: do we need to stick with "URIsh" encoding or can we go with completely different format

comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
```diff
@@ -0,0 +1,109 @@
...
+    extends KademliaRPC[Task] {
+
+  private implicit val logSource: LogSource = LogSource(this.getClass)
+
+  def ping(peer: PeerNode): Task[Boolean] =
+    for {
+      _       <- Metrics[Task].incrementCounter("protocol-ping-sends")
+      channel <- clientChannel(peer)
+      pongErr <- Task
+                  .fromFuture {
```
sebtomba: Is there a reason why you don't use grpc-monix?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -59,7 +60,7 @@ object Configuration {
   private val DefaultThreadPoolSize: Int        = 4000
 
   private val DefaultBootstrapServer: PeerNode = PeerNode
-    .parse("rnode://de6eed5d00cf080fc587eeb412cb31a75fd10358@52.119.8.109:40400")
+    .fromAddress("rnode://de6eed5d00cf080fc587eeb412cb31a75fd10358@52.119.8.109:40400,40404")
```
sebtomba: Don't forget to change the URI (and in the unit tests, f.e. TOML)

rabbitonweb:
I've changed format to `rnode://cb74ba04085574e9f0102cc13d39f0c72219c5bb@bootstrap.rchain.coop?protocol=40400&discovery=40404`

rabbitonweb:
```
================================================================
====================END OF SHOW LOGS=======================
===========================================================
=================TEST SUMMARY RESULTS======================
PASS: peer1.rchain.coop: Peers count correct in node logs.
PASS: peer0.rchain.coop: Peers count correct in node logs.
PASS: peer1.rchain.coop: Rholang evaluation of files performed correctly.
PASS: peer0.rchain.coop: Rholang evaluation of files performed correctly.
PASS: peer0.rchain.coop: REPL loader success!
PASS: peer1.rchain.coop: Proposal of blocks for deployed contracts worked.
PASS: peer0.rchain.coop: Proposal of blocks for deployed contracts worked.
PASS: bootstrap.rchain.coop: Proposal of blocks for deployed contracts worked.
PASS: peer1.rchain.coop: No errors defined by "ERROR" in logs.
PASS: peer0.rchain.coop: No errors defined by "ERROR" in logs.
PASS: peer1.rchain.coop: No text of "RuntimeException" in logs.
PASS: peer0.rchain.coop: No text of "RuntimeException" in logs.
PASS ALL: All tests successfully passed
===========================================================
===========================================================
```

comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
rabbitonweb: next PR

PR #1493, opened 2018-09-17
Topological sort of DAG; uncommon ancestor parents post-state computation

birchmd:
## Overview
This PR introduces the topological sort we will need for serializing blocks to disk in a predictable order. It makes use of the sort to implement the uncommon ancestors algorithm proposed by @guardbotmk3 . This algorithm is used in computing the post-state of multiple parent blocks.

https://rchain.atlassian.net/browse/RHOL-612

### Notes


casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,39 @@ object DagOperations {
...
+    def isCommon(set: BitSet): Boolean                 = (0 until blocks.length).forall(set.contains)
+
+    val initMap = blocks.zipWithIndex.map { case (b, i) => b -> BitSet(i) }.toMap
+    val q       = new mutable.PriorityQueue[BlockMetadata]()
+    q.enqueue(blocks: _*)
+
+    @tailrec
+    def loop(currMap: Map[BlockMetadata, BitSet],
+             visited: HashSet[BlockMetadata]): Map[BlockMetadata, BitSet] =
+      if (q.isEmpty) currMap
```
goral09: If you pass the `q` as an argument this will be even cleaner.

birchmd: Scala doesn't have an immutable PriorityQueue :(

casper/src/main/scala/coop/rchain/casper/BlockDag.scala
```diff
@@ -6,10 +6,26 @@ import coop.rchain.casper.protocol.BlockMessage
 
 import scala.collection.immutable.{HashMap, HashSet}
 
+/*
+ * toposort - Topological sort of the DAG. The index in the outer Vector
+ *            is the block number (length of the longest path back to
+ *            genesis) and the index of the inner Vector breaks ties for
+ *            blocks with the same number based when the node observed the block.
+ *            `topoSort.flatten` gives the sorted blocks.
```
KentShikama: "based on when the node observed"

Does this not make it non-deterministic for validators?

casper/src/main/scala/coop/rchain/casper/BlockDag.scala
```diff
@@ -6,10 +6,26 @@ import coop.rchain.casper.protocol.BlockMessage
...
+ *
+ * sortOffset - This allows us to sort from some block number going forward.
+ *              The idea is that the block height may get too large to use as the index
+ *              of a Vector, or that after a checkpoint on the chain we won't need
+ *              to remember the sorting before that block. The offset gives the block
+ *              number for the first block in the currently maintained sorting.
+ */
```
KentShikama: Is this the latest finalized block?

casper/src/main/scala/coop/rchain/casper/util/TopologicalSortUtil.scala
```diff
@@ -0,0 +1,26 @@
...
+    if (number == sort.length) {
+      //this is a new block height
+      sort :+ Vector(hash)
+    } else if (number < sort.length) {
+      //this is another block at a known height
+      val curr = sort(number)
+      sort.updated(number, curr :+ hash)
+    } else {
+      //impossible because block numbers must be sequential
+      throw new Exception("Attempted to add block with invalid block number to state.")
```
odeac: My understanding is that this exception can only happen in case of an algorithm error. 

If this is correct then we should assert  that `number <= sort.length` as a function precondition such that we make it clear that it's due to a bug in the algorithm.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -152,36 +148,31 @@ object InterpreterUtil {
...
-                        else ProtoUtil.unsafeGetParents[F](block)
-                      }
-                      .filter(_ != gca) //do not include gca deploys
-                      .toList
-
-        deploys = ancestors
-          .flatMap(ProtoUtil.deploys(_).reverse.flatMap(ProcessedDeployUtil.toInternal))
-          .toIndexedSeq
-          .reverse
+      //made in all of the branches of the DAG being
```
odeac: This function is pretty big and, as good boy-scout, I was thinking how to clean it up a little bit.

My suggestion is to replace the if/else-if/else with a match like this:

```parentTupleSpaces match {
  case Seq() => ...no parent case
  case Seq((_, parentStateHash)) => ...single parent case
  case Seq((initParent, initStateHash), _) => ...
}```

...which is more readable and probably clean enough for now.

File: casper/src/main/scala/coop/rchain/casper/BlockDag.scala (hunk 2 above)
Comment (birchmd) [reply casper/src/main/scala/coop/rchain/casper/BlockDag.scala]: Yes, each node could have a different sorting. But they would only differ for blocks at the same height (i.e. blocks with no causal relation), which does not matter. Moreover, this is the only way to ensure that the sorting is stable under append, which I think was a requirement for easy update of the flattened version.

File: casper/src/main/scala/coop/rchain/casper/BlockDag.scala (hunk 3 above)
Comment (birchmd) [reply casper/src/main/scala/coop/rchain/casper/BlockDag.scala]: No, it's the last checkpoint. Remember from our meeting with @nashef that we are assuming the DAG-related information can be completely stored in-memory between checkpoints.

By the way, `BlockMetadata` contains enough information for the safety oracle I think, so we should update the ticket on refactoring it to make use of this in-memory representation of the DAG.  Then it won't need to hit the database at all and will be super fast again.

File: casper/src/main/scala/coop/rchain/casper/util/TopologicalSortUtil.scala [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+
+import coop.rchain.casper.Estimator.BlockHash
+import coop.rchain.casper.protocol.BlockMessage
+import coop.rchain.casper.util.ProtoUtil.blockNumber
+
+object TopologicalSortUtil {
+  type BlockSort = Vector[Vector[BlockHash]]
+  def update(sort: BlockSort, offset: Long, block: BlockMessage): BlockSort = {
+    val hash   = block.blockHash
+    val number = (blockNumber(block) - offset).toInt
```
itegulov: You should probably check this conversion on errors (e.g. if it fits to `Int` and it actually greater than 0.

casper/src/main/scala/coop/rchain/casper/BlockMetadata.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import com.google.protobuf.ByteString
+
+import coop.rchain.casper.Estimator.BlockHash
+import coop.rchain.casper.protocol.{BlockMessage, Justification}
+
+final case class BlockMetadata(blockHash: BlockHash,
+                               parents: List[BlockHash],
+                               sender: ByteString,
+                               justifications: List[Justification])
```
itegulov: I assume we will use this class as "BlockMessage without Body" you mentioned previously. Don't you think we need other fields from `BlockMessage` and `Header` here?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -152,36 +148,31 @@ object InterpreterUtil {
...
+      val initIndex = indexedParents.indexOf(dag.dataLookup(initParent.blockHash))
+      //filter out blocks that already included by starting from the chosen initParent
+      val blocksToApply = uncommonAncestors
+        .filterNot { case (_, set) => set.contains(initIndex) }
+        .keys
+        .toVector
+        .sorted //ensure blocks to apply is topologically sorted to maintain any causal dependencies
 
+      for {
+        blocks  <- blocksToApply.traverse(b => ProtoUtil.unsafeGetBlock[F](b.blockHash))
```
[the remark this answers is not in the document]
itegulov: It just assumes that the block definitely exists in the storage. Which I think is reasonable here.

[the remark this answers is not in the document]
itegulov: It just assumes that the block definitely exists in the storage. Which I think is reasonable here.

casper/src/main/scala/coop/rchain/casper/BlockMetadata.scala
birchmd: Indeed, that is what I was going for here. I think the fields there are sufficient. But if we realize that we need more then it is easy to add them to this case class.

birchmd:
@odeac @itegulov Thanks for the comments! I think I have addressed them all, PTAL

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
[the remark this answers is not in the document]
birchmd: @odeac Would you prefer
```scala
for {
  maybeBlocks <- blocksToApply.traverse(b => BlockStore[F].get(b.blockHash))
  _           = assert(maybeBlocks.forall(_.isDefined))
  blocks      = maybeBlocks.flatten
  deploys     = blocks.flatMap(_.getBody.deploys.flatMap(ProcessedDeployUtil.toInternal))
} yield ...
```
?

casper/src/main/scala/coop/rchain/casper/BlockDag.scala
birchmd: We could use something like validator ID, but then we have to do some work when deciding where to insert a new block in the sorting (where as right now it is a very minimal operation). I don't see why it matters what order they are in for the sorting though. If there was indeed a race between two blocks at the same height then they conflict and so could never be merged together anyways.

KentShikama: As of now; that's why I said eventual implementation.

birchmd: In the eventual implementation the merge preference is determined by the block's score, so the sort still won't matter. As far as I can tell, the sorting is an internal detail which allows us to compute the uncommon ancestors and serialize the DAG to disk, but different nodes don't need to agree on the precise ordering because any topological sort will do for those purposes.

guardbotmk3: @birchmd Exactly. Any topological sort will do for the algorithm. You'll get the same results with any.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+    def parents(b: BlockMetadata): List[BlockMetadata] = b.parents.map(lookup)
+    def isCommon(set: BitSet): Boolean                 = (0 until blocks.length).forall(set.contains)
+
+    val initMap = blocks.zipWithIndex.map { case (b, i) => b -> BitSet(i) }.toMap
+    val q       = new mutable.PriorityQueue[BlockMetadata]()
+    q.enqueue(blocks: _*)
+
+    @tailrec
+    def loop(currMap: Map[BlockMetadata, BitSet],
+             visited: HashSet[BlockMetadata]): Map[BlockMetadata, BitSet] =
```
guardbotmk3: You don't need the visited set, Instead you want a set that mirrors the contents of the queue so that nothing is in the queue twice.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+      if (q.isEmpty) currMap
+      else {
+        val currBlock = q.dequeue()
+        val currSet   = currMap.getOrElse(currBlock, BitSet.empty)
```
guardbotmk3: The orElse should never occur. It's probably best to at least add a comment to that effect.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+        if (visited(currBlock) || isCommon(currSet)) {
```
guardbotmk3: If it's common, you can remove it from the map. (You won't see it again because you're going in reverse topological order)

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+          //already visited or is common ancestor
+          loop(currMap, visited)
+        } else {
+          val (newMap, newVisted) = parents(currBlock).foldLeft(currMap -> visited) {
```
guardbotmk3: Don't use `->` for arbitrary tuples.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+            case ((map, vstd), p) =>
+              if (!vstd(p)) q.enqueue(p)
+              updatedWith(map, p)(currSet)(_ | currSet) -> (vstd + p)
```
guardbotmk3: Same here. `->` is for associations, and this isn't one.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+          }
+
+          loop(newMap, newVisted)
+        }
+      }
+
+    loop(initMap, HashSet.empty[BlockMetadata]).filter { case (_, set) => !isCommon(set) }
```
guardbotmk3: If you remove common ancestors from the set as you find them, you won't need this filter.

casper/src/test/scala/coop/rchain/casper/util/DagOperationsTest.scala
```diff
@@ -66,4 +68,50 @@ class DagOperationsTest
...
+
+    implicit val ordering = BlockDag.deriveOrdering(chain.dag)
+    DagOperations.uncommonAncestors(Vector(b6, b7), chain.dataLookup) shouldBe Map(
+      b6 -> BitSet(0),
+      b2 -> BitSet(0),
+      b7 -> BitSet(1),
+      b5 -> BitSet(1)
+    )
+  }
+
```
guardbotmk3: Please test additional sets. {b6, b2} and {b2, b4, b5} at a minimum.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,53 @@ object DagOperations {
...
+    *               traversal)
+    * @param topoSort topological sort of the DAG, ensures ancestor computation is
+    *                 done correctly
+    * @return A map from uncommon ancestor blocks to BitSets, where a block B is
+    *         and ancestor of starting block with index i if B's BitSet contains i.
+    */
+  def uncommonAncestors(blocks: IndexedSeq[BlockMetadata], lookup: BlockMetadata.Lookup)(
+      implicit topoSort: Ordering[BlockMetadata]): Map[BlockMetadata, BitSet] = {
+    def parents(b: BlockMetadata): List[BlockMetadata] = b.parents.map(lookup)
+    def isCommon(set: BitSet): Boolean                 = (0 until blocks.length).forall(set.contains)
```
guardbotmk3: I would prefer that you built the common set, and then did a single comparison.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
guardbotmk3: Also you should only need to test visited on insert, not on removal.

casper/src/main/scala/coop/rchain/casper/BlockDag.scala
goral09: But if we have `in-memory` representation of the DAG then the memory footprint rises again right?

birchmd: Yes. But holding only the meta-data (i.e. hashes, parent hash pointer and justification hash pointers) it totally doable because it is very little data per block. The main block with it's full proto-representation (including deployed code, logs, etc) will only be on disk and only fetched when needed (which will hopefully be very rarely compared to the metadata usage)

birchmd:
@guardbotmk3 Thanks for pointing out the problems with the `visited` set. I have fixed those and taken care of your other comments as well. I needed to make a slight change to the algorithm in order to propagate the common ancestor information down through the DAG, otherwise a block that should be common could miss one of the ancestors because a particular branch got stopped early (e.g. `b3` in the test with `{b6, b7}`). As a result I needed to keep in the filter at the end of the loop removing the last common anscestors that got included (which should be the parents of the GCA from the previous algorithm).

With that in mind PTAL

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,62 @@ object DagOperations {
...
+    def isCommon(set: BitSet): Boolean                 = set == commonSet
+
+    val initMap = blocks.zipWithIndex.map { case (b, i) => b -> BitSet(i) }.toMap
+    val q       = new mutable.PriorityQueue[BlockMetadata]()
+    q.enqueue(blocks: _*)
+
+    @tailrec
+    def loop(currMap: Map[BlockMetadata, BitSet],
+             enqueued: HashSet[BlockMetadata]): Map[BlockMetadata, BitSet] =
+      if (q.isEmpty) currMap
```
guardbotmk3: I realized that the termination condition has to be slightly different.
You have to keep enqueuing parents even if they are common.
The stopping condition is "The queue is empty" or "Everything in the queue is common."

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,52 @@ object DagOperations {
...
+    def isCommon(set: BitSet): Boolean                 = set == commonSet
+
+    val initMap = blocks.zipWithIndex.map { case (b, i) => b -> BitSet(i) }.toMap
+    val q       = new mutable.PriorityQueue[BlockMetadata]()
+    q.enqueue(blocks: _*)
+
+    @tailrec
+    def loop(currMap: Map[BlockMetadata, BitSet],
+             enqueued: HashSet[BlockMetadata]): Map[BlockMetadata, BitSet] =
+      if (q.forall(b => isCommon(currMap(b)))) currMap
```
guardbotmk3: I would rather you kept a set of things in the queue that weren't common.
q.forall in a loop looks like a performance bottleneck.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -32,6 +33,52 @@ object DagOperations {
...
+      else {
+        val currBlock = q.dequeue()
+        //Note: The orElse case should never occur because we traverse in
+        //      reverse topological order (i.e. down parent links)
+        val currSet = currMap.getOrElse(currBlock, BitSet.empty)
+        val (newMap, newEnqueued) = parents(currBlock).foldLeft((currMap, enqueued)) {
```
guardbotmk3: Shouldn't that be enqueued - currBlock? since it's no longer enqueued?

birchmd: I subtract `currBlock` from the enqueued set on lines 75 and 76. So the current version is not wrong, but yes, it would be cleaner if I did it here instead of two different places later. So I will change as you suggest.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
birchmd: `forall` will terminate early if it finds a `false`, so it might not be as bad as you expect, but you are right that it is probably faster to track what is uncommon and still queued and wait for that set to be empty.

PR #1500, opened 2018-09-19
RHOL-592: Catch lexer errors

tymm:
## Overview
No more `UnrecognizedInterpreterError` for errors thrown by the lexer.

https://rchain.atlassian.net/browse/RHOL-592

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/LexerTest.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/LexerTest.scala
@@ -12,7 +12,12 @@ class LexerTest extends FlatSpec with Matchers {
     Interpreter.buildNormalizedTerm(new StringReader(input)).runAttempt()
 
   "Lexer" should "return LexerError for unterminated string at EOF" in {
-    val attempt = attemptMkTerm("""{{ @"ack!(0)""")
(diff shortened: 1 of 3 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/LexerTest.scala
```diff
@@ -0,0 +1,24 @@
...
+
+  "Lexer" should "return LexerError for unterminated string at EOF" in {
+    val attempt = attemptMkTerm("""{{ @"ack!(0)""")
+    attempt.left.value shouldBe a[LexerError]
+  }
+
+  it should "return LexerError for illegal character" in {
+    val attempt =
+      attemptMkTerm("""("x is ${value}" % {"value" : x})""")
+    attempt.left.value shouldBe a[LexerError]
```
ArturGajowy: I've noticed there are 6 places where an exception is thrown in Yylex, one of them being for "things that cannot be" - so not of interest for this change (we'll need the stacktrace should that happen, so it should be left as an UnrecognizedError).

Of the other 5 though, I think we have test cases for 2 of them (3 once you include the `Unterminated string on line ` case). This leaves us with 2 more test cases that I deem optional - but it'd be decent to have them. Up to you to decide :)

PR #1503, opened 2018-09-19
Rholang: Make the registry available at fixed URNs

guardbotmk3:
This makes the simple regsitry at arbitrary names available in the default
Runtime and not just for testing.


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -97,6 +97,22 @@ object Runtime {
     val REG_PUBLIC_REGISTER_INSERT_CALLBACK: Long = 19L
   }
 
+  def byteName(b: Byte): Par = GPrivate(ByteString.copyFrom(Array[Byte](b)))
+
+  object FixedChannels {
+    val STDOUT: Par            = byteName(0)
+    val STDOUT_ACK: Par        = byteName(1)
+    val STDERR: Par            = byteName(2)
```
ArturGajowy: I very much like this refactoring, but would even love it if it were a separate commit :) Just FYI, the commit and the PR is small enough for this to not matter really.

guardbotmk3:
I went to write a test, but then I stopped. The most obvious test would be to make sure that the correct name is returned for the correct urn.

That amounts to writing the table down twice, once in the code, and again as a test.

That's not really a test.

PR #1504, opened 2018-09-19
Registry setup refactor

goral09:
## Overview
This is preliminary work for [RHOL-804](https://rchain.atlassian.net/browse/RHOL-804). It look like all the tasks are so connected together that it's really hard to create small PRs. 

This PR:
1. Refactors the `Registry` to remove circular dependency.
2. Refactors the `RholangAndScalaDispatcher.create` method.
3. Refactors the `PureRSpace` construction.
4. Introduces concept of `ChargingRSpace`.

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
[RHOL-804](https://rchain.atlassian.net/browse/RHOL-804) 

### Complete this checklist before you submit the PR
- [ ] This PR contains no more than 200 lines of code, excluding test code.
- [ ] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [ ] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [ ] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [ ] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1505, opened 2018-09-19
RHOL-11: Introduce ChargingReducer

goral09:
## Overview
Relative to #1504 (review it first).
This PR introduced `ChargingReducer` that wraps `DebruijInterpreter` and `CostAccounting`.
This is necessary for proper sharing of the cost state between different parts of the interpreter (specifically dispatcher and reducer).

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -11,9 +11,9 @@ import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Channel.ChannelInstance.Quote
 import coop.rchain.models.Expr.ExprInstance.GString
 import coop.rchain.models._
-import coop.rchain.rholang.interpreter.accounting.{CostAccount, CostAccountingAlg}
+import coop.rchain.rholang.interpreter.accounting.{Cost, CostAccount}
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
-import coop.rchain.rholang.interpreter.{ErrorLog, Reduce, Runtime}
+import coop.rchain.rholang.interpreter.{ChargingReducer, ErrorLog, Runtime}
 import coop.rchain.rspace.internal.{Datum, WaitingContinuation}
 import coop.rchain.rspace.trace.Produce
 import coop.rchain.rspace.{Blake2b256Hash, ReplayException}
@@ -22,7 +22,8 @@ import monix.execution.Scheduler
 
 import scala.annotation.tailrec
 import scala.collection.immutable
-import scala.concurrent.SyncVar
+import scala.concurrent.duration._
+import scala.concurrent.{Await, SyncVar}
 import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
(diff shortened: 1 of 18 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -112,15 +100,22 @@ object RholangAndScalaDispatcher {
...
     lazy val tuplespaceAlg = TuplespaceAlg.rspaceTuplespace(pureSpace, dispatcher)
     lazy val dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation] =
-      new RholangAndScalaDispatcher(reducer, dispatchTable)
-    lazy val reducer: Reduce[Task] =
+      new RholangAndScalaDispatcher(chargingReducer, dispatchTable)
+    implicit lazy val costAccounting: CostAccountingAlg[Task] =
+      CostAccountingAlg.unsafe[Task](CostAccount(0))
+    implicit lazy val reducer: Reduce[Task] =
       new Reduce.DebruijnInterpreter[Task, Task.Par](tuplespaceAlg, urnMap)
+    lazy val chargingReducer    = new ChargingReducer[Task]() {}
```
ArturGajowy: It seems ChargingReducer should not be abstract then

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -143,11 +144,11 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
       terms match {
         case deploy +: rem =>
           runtime.space.reset(hash)
-          val availablePhlos             = CostAccount(Integer.MAX_VALUE)
-          implicit val costAccountingAlg = CostAccountingAlg.unsafe[Task](availablePhlos) //FIXME this needs to come from the deploy params
-          val (phlosLeft, errors)        = injAttempt(deploy, runtime.reducer, runtime.errorLog)
-          val cost                       = phlosLeft.copy(cost = availablePhlos.cost.value - phlosLeft.cost)
-          val newCheckpoint              = runtime.space.createCheckpoint()
+          val availablePhlos = Cost(Integer.MAX_VALUE)
+          Await.ready(runtime.reducer.setAvailablePhlos(availablePhlos).runAsync, 1.second) // FIXME: This needs to come from the deploy params
```
ArturGajowy: I am very wary of the `Await` in here. I'm also not yet sure why we can't avoid mutating the reducer's state. Please provide a high level overview of what exactly you're trying to achieve and why a cleaner approach is not possible.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -58,10 +58,10 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
 
   "evalExpr" should "handle simple addition" in {
     implicit val errorLog = new ErrorLog()
-    implicit val costAccounting =
-      CostAccountingAlg.unsafe[Task](CostAccount(Integer.MAX_VALUE))
+
     val result = withTestSpace { space =>
-      val reducer      = RholangOnlyDispatcher.create[Task, Task.Par](space)._2
+      val reducer = RholangOnlyDispatcher.create[Task, Task.Par](space)._2
+      Await.ready(reducer.setAvailablePhlos(Cost(Integer.MAX_VALUE)).runAsync, 1.second)
```
ArturGajowy: This commit leaves no doubt as to whether we should elliminate the duplication of setup in this test first. Please do so, as proposed in review for the other PR.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
goral09: Because when we leave RSpace with a matching data in `Dispatcher` we need to continue burning the same phlos. If we have a `costAccountingAlg` dependency at method call level this won't work. We already share same `Reduce` instance in `Dispatch` so it makes sense to move the state inside to `Reduce`. Since `costAccountingAlg` is not required at method call anymore, we need a way to set new `phloLimit` which is known only when we know deploy data.

goral09:
Closing in favor of #1513

PR #1513, opened 2018-09-20
Share same cost accounting state between dispatcher an reducer

goral09:
## Overview
When we leave RSpace with a matching data in dispatcher we need to continue burning the same phlos. If we have a cost accounting algebra dependency at method call level this won't work. We already share same Reduce instance in Dispatch so it makes sense to move the state inside to Reduce. Since cost algebra is not available at method call anymore, we need a way to set new phlo limit, which is known only when we know deploy params.

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -11,9 +11,9 @@ import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Channel.ChannelInstance.Quote
 import coop.rchain.models.Expr.ExprInstance.GString
 import coop.rchain.models._
-import coop.rchain.rholang.interpreter.accounting.{CostAccount, CostAccountingAlg}
+import coop.rchain.rholang.interpreter.accounting.{Cost, CostAccount}
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
-import coop.rchain.rholang.interpreter.{ErrorLog, Reduce, Runtime}
+import coop.rchain.rholang.interpreter.{ChargingReducer, ErrorLog, Runtime}
 import coop.rchain.rspace.internal.{Datum, WaitingContinuation}
 import coop.rchain.rspace.trace.Produce
 import coop.rchain.rspace.{Blake2b256Hash, ReplayException}
@@ -23,6 +23,7 @@ import monix.execution.Scheduler
 import scala.annotation.tailrec
 import scala.collection.immutable
 import scala.concurrent.SyncVar
+import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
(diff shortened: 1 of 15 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -151,11 +152,11 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
       terms match {
         case deploy +: rem =>
           runtime.space.reset(hash)
-          val availablePhlos             = CostAccount(Integer.MAX_VALUE)
-          implicit val costAccountingAlg = CostAccountingAlg.unsafe[Task](availablePhlos) //FIXME this needs to come from the deploy params
-          val (phlosLeft, errors)        = injAttempt(deploy, runtime.reducer, runtime.errorLog)
-          val cost                       = phlosLeft.copy(cost = availablePhlos.cost.value - phlosLeft.cost)
-          val newCheckpoint              = runtime.space.createCheckpoint()
+          val availablePhlos = Cost(Integer.MAX_VALUE) // FIXME: This needs to come from the deploy params
+          runtime.reducer.setAvailablePhlos(availablePhlos).runSyncUnsafe(1.second)
```
ArturGajowy: This is very unfortunate, and I really can't get over it.

I see that the `set` is needed, because we only want to create the ChargingReducer and the undrlying CostAccountingAlg once.

But please remind me, what's the reason for us not being able to thread the CostAccountingAlg through implicit method parameters, and create it here?

I appreciate the fact that we can't initialize the CostAccountingAlg until we have the phloLimit. But then, **this is the exact place it should be created at**.

I can see now yet another reason why @rabbitonweb insists on passing capabilities via methods' type parameters instead of fixing them once in trait/class header.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -125,14 +105,16 @@ object RholangAndScalaDispatcher {
...
+  ): (Dispatch[M, ListChannelWithRandom, TaggedContinuation], ChargingReducer[M], Registry[M]) = {
     val pureSpace          = PureRSpace[M].of(tuplespace)
     lazy val tuplespaceAlg = TuplespaceAlg.rspaceTuplespace(pureSpace, dispatcher)
     lazy val dispatcher: Dispatch[M, ListChannelWithRandom, TaggedContinuation] =
-      new RholangAndScalaDispatcher(reducer, dispatchTable)
-    lazy val reducer: Reduce[M] =
+      new RholangAndScalaDispatcher(chargingReducer, dispatchTable)
+    implicit lazy val reducer: Reduce[M] =
       new Reduce.DebruijnInterpreter[M, F](tuplespaceAlg, urnMap)
+    implicit lazy val costAlg      = CostAccountingAlg.unsafe[M](CostAccount(0))
```
ArturGajowy: This is as bad as saying: "I'll set the field to null and initialize it using a setter later". It's also admitting we're creating this too soon, in a place we don't have all the data needed.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -55,6 +55,30 @@ trait Reduce[M[_]] {
   )(implicit env: Env[Par], costAccountingAlg: CostAccountingAlg[M]): M[Par]
 }
 
+// TODO:
+// In a perfect world we would an algebra of an interpreter encoded in the terms of
+// rholang's elimination and have the `ChargingReducer` forward the calls to the plain algebra
+// and charge accordingly. This would eliminate the "charge" calls from the algebra of a language.
+class ChargingReducer[M[_]](implicit R: Reduce[M], C: CostAccountingAlg[M]) {
```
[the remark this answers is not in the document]
ArturGajowy: Had another look and it actually removes some duplication in tests (that shouldn't be there in the first place) and 2 instantiations of CostAccountingAlg in the dispatchers - which also shouldn't have been there. All of these could've been avoided without introducing a ChargingReducer, simply by deduplicating in the first case and threading CostAccountingAlg by methods' implicit parameters in the second one.

PR #1514, opened 2018-09-20
Implement message queues in TCP transport layer

sebtomba:
## Overview
Implement message queues in TCP transport layer with Monix `Observable` and parallel dispatching of messages. The messaging pattern is motivated by actor systems. There is a `tell` operation for one-way messages without a reply. However, the client waits for the message being delivered and enqueued to a message queue on the peer node. The second operation is `ask` for request-response messaging. This operation is also served over a message queue with a callback. The `tell` and `ask` messages queues are getting zipped into one message queue and then dispatched in parallel. The default parallelism is the number of CPU cores.
 
https://rchain.atlassian.net/browse/CORE-1148


goral09:
Will review tomorrow, but I have a question already, @sebtomba did you make a performance tests? Because from the description it sounds like awful lot of awaiting/synchronization. Is this gonna happen for every message sent to any node?

PR #1519, opened 2018-09-21
CORE-1134 introduced F with Sync.delay implementation to RSpace.

dzajkowski:
## Overview
Another step in the effort to deliver green threads in RSpace.

https://rchain.atlassian.net/projects/CORE/issues/CORE-1134

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
@@ -3,6 +3,7 @@ package coop.rchain.rholang.interpreter
 import java.nio.file.{Files, Path}
 
 import cats.Id
+import cats.effect.Sync
 import cats.mtl.FunctorTell
 import com.google.protobuf.ByteString
 import coop.rchain.models.Channel.ChannelInstance.{ChanVar, Quote}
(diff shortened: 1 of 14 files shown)


ArturGajowy:
I had a skim and one thing that came to my mind is that, since `delay` is being used extensively, there is a risk some effects would be lost. What I mean by that is:

```scala
def foo: F[Unit] = Sync[F].delay {
  //...
  bar(x, y, z)
  //...
}

def bar(x: Any*): F[Unit] = Sync.delay {
//...
}
```

After changing both foo and bar to return via F by using `delay`, we can be loosing effects of now-suspended methods if they were used as statements, or if their F-return value wasn't joined properly.

A very careful review should be conducted having this in mind.

lukasz-golebiewski:
This is a good catch @ArturGajowy! I think it should be enough to replace `ISpace[Id` with `ISpace[F` in `PureRSpace` and the related errors would appear in rholang tests

ArturGajowy:
> related errors would appear in rholang tests

I wouldn't put too much faith in tests in general :)

One way to really make sure we don't lose anything (that has just came to me) is changing the method names of all effect-converted methods, say by adding M at the end - but without 'rename' or 'search and replace'. This way we get each and every usage of those methods in red by the compiler and we can make sure case-by-case that effects are joined properly. In fact I did that when porting the PrettyPrinter.

Another additional security measure would be going through "expression as statement" warnings from Scapegoat, as that's one of the possible causes for loosing effects. That is, if I understand this inspections purpose from its name correctly.

PR #1546, opened 2018-09-24
Re-work LMDB transaction handling

lukasz-golebiewski:
### JIRA issue:
https://rchain.atlassian.net/browse/CORE-1192

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
@@ -11,7 +11,7 @@ import coop.rchain.shared.ByteVectorOps._
 import coop.rchain.shared.PathOps._
 import scodec.bits.BitVector
 import kamon._
-import org.lmdbjava.Txn.NotReadyException
+import scala.util.control.NonFatal
 
 trait LMDBOps extends CloseOps {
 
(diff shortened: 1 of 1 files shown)


nashef:
Posting my comments from discord, as I don't think this Pull Request is enough.

@lukasz.golebiewski It seems odd to me that the catch clause would convert a MapFullException into a NotReadyException. The logic must be broken. It's not just a question of when to print the stack trace.

lukasz.golebiewskiToday at 2:32 PM
It doesn't convert, the original exception is re-thrown
It looks like the exception is being swallowed in Casper tests, so the least I can do is print the stack trace of the original cause which gives more info than mere "txn not in ready state"

leafToday at 2:53 PM
Oh, I see... so the inner exception was NotReadyException and its stack trace was printed. Then, the outer exception was re-thrown, and the Casper tests swallowed that one silently?
I think the bug is that it's wrong to call abort() on an LMDB transaction that has been commit()ted.
If the commit() succeeded, then abort() can't possibly do anything. If the commit() failed, the transaction is already dead. You don't need to abort(). The C documentation says this:

```
int mdb_txn_commit    (    MDB_txn *     txn    )    
Commit all the operations of a transaction into the database.

The transaction handle is freed. It and its cursors must not be used again after this call, except with mdb_cursor_renew().

Note
Earlier documentation incorrectly said all cursors would be freed. Only write-transactions free cursors.

Parameters

[in]    txn    A transaction handle returned by mdb_txn_begin()

Returns

A non-zero error value on failure and 0 on success. Some possible errors are:

EINVAL - an invalid parameter was specified.
ENOSPC - no more disk space.
EIO - a low-level I/O error occurred while writing.
ENOMEM - out of memory.
```

Nota bene: It and its cursors must not be used again after this call, except with mdb_cursor_renew().

Also, I think you should explicitly catch and handle the four exceptions that checkRc() could throw here. That way the user will have a clue what happened.

dzajkowski:
@nashef agreed, the code can be done neater, but it will not change how it behaves. the current implementation does not skip exceptions, does not throw unnecessary ones. it throws the root cause and prints the failure in abort.
IOW this does not solve the problem that was raised on discord - failures in tests.

nashef:
LGTM

I don't know Scala well enough to know if you need to explicitly catch all exception types inside of a `catch` statement. Lukasz says not, so I'm approving this.

rabbitonweb:
I can confirm that it will catch all exceptions

rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
```diff
@@ -45,37 +45,8 @@ trait LMDBOps extends CloseOps {
...
-          // due to the way LMDBjava tries to handle txn commit
-          // IF the DB runs out of space AND this occurs while trying to commit a txn (2)
-          // the internal java state (STATE in Txn) will be set to DONE (1) before the Txn is committed
-          // (the exception happens in the attempt to commit)
-          // the abort action will interpret this as an invalid state (3)
-          // A fix is to ignore this exception as the original one is a valid cause
-          // and it still will cause txn.close in the finally block
-        }
+      case NonFatal(ex) =>
+        ex.printStackTrace
```
rabbitonweb: Scala exception mechanism does not allow you to specify which exceptions are thrown by method. Thus you end up in situation like the one above where:
1. you re-throw exception for somebody else to deal with it,
2. but the type signature does not allow to specify that you are doing it (throwing exceptions)
3. thus the caller of your function might not handle the error, giving you bad times "it's rspace fault!"
4. so you end up printing out stack traces

A better approach is to return the error, inform the caller about possibilities of errors in type signature. That will be
1. return `Either[Throwable, R`] instead of `R`
2. or return F[R] (tagles final to the rescue) with `ApplicativeError` instance

Regardless wich approach you chose, the type signature will change, thus the usage of the `withTxn` will change - users of this method will have to handle the possible error. This would be a change that would cut across the whole project, but its step in the right direction.

Other then that changes below LGTM, but I would consider my comment above with future changes

PR #1552, opened 2018-09-25
Refactor tests

odeac:
## Overview
Simplify integration tests

https://rchain.atlassian.net/browse/RHOL-770

### Notes


integration-testing/readme.md
```diff
@@ -54,8 +54,10 @@ Summary: A Python library for the Docker Engine API.
 
 ``` 
 
-# Running the tests
 
+# Test organization
+The tests are organized in a treee structure. The
```
KentShikama: tree

KentShikama: actually incomplete?

integration-testing/src/fixtures/common.py
```diff
@@ -57,33 +68,20 @@ def docker():
...
     with os.fdopen(fd, "w") as f:
         for pair in validator_keys:
             bond = random.randint(1, 100)
             f.write(f"{pair.public_key} {bond}\n")
 
-    yield bonds_file
+    yield file
 
-    os.unlink(bonds_file)
+    os.unlink(file)
```
adaszko: I think the file won't be deleted if an exception gets thrown within the body of `with` statement that uses this context manager.  You may need a `try...finally` block around `yield`.

integration-testing/src/fixtures/common.py
```diff
@@ -98,24 +96,27 @@ def validators_data(config):
...
-@pytest.fixture(scope="package")
-def started_bootstrap(config, bootstrap):
-    wait_for( string_contains( node_logs(bootstrap),
-                               "coop.rchain.node.NodeRuntime - Listening for traffic on rnode"),
-              config.node_startup_timeout,
-              "Bootstrap node didn't start correctly")
-    yield bootstrap
\ No newline at end of file
+    with bonds_file(validator_keys) as f:
+        yield (f, validator_keys[0], validator_keys[1:])
```
adaszko: Can this be made into a namedtuple/dataclass to have clarity what the tuple elements are?

integration-testing/src/fixtures/network.py
```diff
@@ -0,0 +1,41 @@
...
+                     docker,
+                     bootstrap_node,
+                     validators_data,
+                     [bootstrap_node.name]) as n:
+
+            wait_for_started_network(config.node_startup_timeout, n)
+
+            wait_for_converged_network(config.network_converge_timeout, n, 1)
+
+            yield n
```
adaszko: `n` could be a bit more descriptive.

integration-testing/src/tools/fixture.py
```diff
@@ -19,7 +19,7 @@ def make_wrapper(fn, fixtures):
     namespace = {"fn":fn, "fixtures":fixtures}
 
     wrapper_code = f"""
-def wrapper(request, {parameter_list}):
+def {fn.__name__}(request, {parameter_list}):
```
adaszko: Is this runtime code generation?  This is usually a bad idea.  Why do we need this?

integration-testing/src/tools/network.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+@contextmanager
+def start_network(config, docker, bootstrap, validators_data, allowed_peers=None):
+    logging.debug(f"Docker network = {bootstrap.network}")
+
+    bonds_file, _, peer_keys = validators_data
+    peers = create_peer_nodes(docker, bootstrap, bootstrap.network, bonds_file, peer_keys, config.rnode_timeout, allowed_peers)
+
+    yield RChain(network = bootstrap.network, bootstrap = bootstrap, peers = peers)
```
adaszko: There are no spaced around `=` in PEP8.

integration-testing/src/tools/network.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for peer in peers:
+        peer.cleanup()
+
+
+def wait_for_started_network(node_startup_timeout, network):
+    for peer in network.peers:
+        wait_for(node_started(peer), node_startup_timeout, f"Peer {peer.name} did not start correctly.")
```
adaszko: I'm strongly against using callback in this way.  It is much more readable and debuggable to have a explicit loop.  Higher order functions in Python are more problematic than in a statically-typed language.

integration-testing/src/tools/network.py
```diff
@@ -0,0 +1,40 @@
...
+
+def wait_for_converged_network(timeout, network, peer_connections):
+    wait_for( has_peers( network.bootstrap, len(network.peers)),
```
adaszko: Ditto, see the callbacks argument

integration-testing/src/tools/network.py
```diff
@@ -0,0 +1,40 @@
...
+              timeout,
+              "The network did NOT converge. Check container logs for issues. One or more containers might have failed to start or connect.")
+
+    for node in network.peers:
+        wait_for( has_peers( node, peer_connections),
```
adaszko: Ditto, see the callbacks argument

integration-testing/src/tools/rnode.py
```diff
@@ -168,3 +171,24 @@ def create_peer(i, key_pair):
...
+
+
+@contextmanager
+def create_bootstrap(docker, docker_network, timeout, validators_data):
+    bonds_file, bootstrap_keys, _ = validators_data
+    node = create_bootstrap_node(docker, docker_network, bonds_file, bootstrap_keys, timeout)
+
+    yield node
+
+    node.cleanup()
```
adaszko: Again, is this immune to raising exceptions within the `with` body?

integration-testing/src/tools/rnode.py
```diff
@@ -168,3 +171,24 @@ def create_peer(i, key_pair):
...
+
+@contextmanager
+def start_bootstrap(docker_client, node_start_timeout, node_cmd_timeout, validators_data):
+    with docker_network(docker_client) as network:
+        with create_bootstrap(docker_client, network, node_cmd_timeout, validators_data) as node:
+
+            wait_for( node_started(node),
```
adaszko: Ditto, see the callbacks argument

integration-testing/src/tools/fixture.py
odeac: I needed to parametrize the tests by fixtures and I havent' found a different way to say that a testX runs with all the combination of fixtures. 

For example I want a test to run with a network fixture, a contract generation fixture and a deployment fixture

The network fixture ca be complete, star or line shaped
The contract deployment fixture can be big, small or many comms.
The deployment fixture can be fast deployment, slow or bursts.

So I want to run my test, with the same logic in all the 27 combinations (i.e. the cartezian product of the above). 

With this decorator I would say something like this

```python
@parametrized(
    network=[complete, star, line], 
    contract_factory=[small, big, many_comms], 
    deployer=[fast, slow, bursts])
def test_fancy_stuff(network, contract_factory, deployer):
  #...
```
I wanted to use this to run test_casper_propose_and_deploy whebn the network is either  complete or star shaped network and the blocks could be valid or invalid. So that's 4 tests with an identical body. 

In the end my understanding is that we won't test the invalid blocks in the integration tests but the tool is already implemented and working well afaik so I think it's worth keeping it.

integration-testing/src/tools/network.py
odeac: wait_for is a generic function. It waits for any condition. 

If you want a specific function for a certain condition you can implement it on top of wait_for and a couple of basic predicates

integration-testing/src/tools/network.py
odeac: I don't understand the problem. Without a basic tool like wait_for and its predicates you will end up with tons of functions wait_for_node, wait_for_peers, wait_for_lunch...

Maybe I should document wait_for and its mechanisms better?

integration-testing/src/tools/network.py
odeac: see my previous comments

integration-testing/src/tools/rnode.py
odeac: Yes. A context manager will call the code after the `yield` if an exception is thrown within the `with` statement where it is used

integration-testing/src/fixtures/common.py
odeac: The `@contextmanager` does that

integration-testing/src/fixtures/common.py
odeac: good point

integration-testing/src/fixtures/network.py
odeac: will do

integration-testing/src/tools/rnode.py
odeac: Houston, we have a problem! :)

integration-testing/readme.md
odeac: Will fix

integration-testing/src/fixtures/common.py
adaszko: That's not what the example indicates: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager

There's a explicit try...finally there.

integration-testing/src/tools/network.py
adaszko: > Without a basic tool like wait_for and its predicates you will end up with tons of functions wait_for_node, wait_for_peers, wait_for_lunch...

This is exactly what I'm proposing.  The savings in repeated code nowhere near offset the detriment to readability and simplicity.

integration-testing/src/fixtures/common.py
odeac: oops! you are right. 

it seems I was using the contextmanager in a wrong way

adaszko: Just check yourself:

```
from contextlib import contextmanager

@contextmanager
def foo():
  print('prologue')
  yield 123
  print('epilogue')


with foo():
  raise ValueError("bar")
```

```
(dev)% python3 manager.py
prologue
Traceback (most recent call last):
  File "manager.py", line 11, in <module>
    raise ValueError("bar")
ValueError: bar
zsh: exit 1     python3 manager.py
```

integration-testing/src/tools/network.py
adaszko: What you're saying is crystal clear.  The problem is with debugging such a code.  A test will fail eventually.  When it fails, the stack trace should ideally tell the expectation that failed.  With `wait_for()` the stack trace will be deeper, names more cryptic and the reason for failing more obscure.  Trust me, let's not take this path.

integration-testing/src/fixtures/common.py
odeac: I tried that and you are right. Thanks

integration-testing/src/tools/network.py
odeac: I was focusing on readability from the final user perspective, where the final user is the Scala developer, with minimal Python knowledge, who wants to write an intregration test. With these tools the code that waits for the bootstrap node to start is this:

```python
def start_bootstrap(docker_client, node_start_timeout, node_cmd_timeout, validators_data):
    with docker_network(docker_client) as network:
        with create_bootstrap(docker_client, network, node_cmd_timeout, validators_data) as node:

            wait_for( node_started(node),
                      node_start_timeout,
                      "Bootstrap node didn't start correctly")

            yield node
```

...and a test failure because the bootstrap node failed to start would look like this:

```
docker_client = <docker.client.DockerClient object at 0x10445dd30>, node_start_timeout = 50, node_cmd_timeout = 10
validators_data = ValidatorsData(bonds_file='/tmp/rchain-bonds-file-zdyeeawm.txt', bootstrap_keys=KeyPair(private_key='2574c8e500a0de1d3...9455377bf8a805d54dd3933612aeb3d970a5', public_key='8b9f35f7cb951b4c89d450e418c921e6faf484f93c94a04ec00e310b624339d0')])

    @contextmanager
    def start_bootstrap(docker_client, node_start_timeout, node_cmd_timeout, validators_data):
        with docker_network(docker_client) as network:
            with create_bootstrap(docker_client, network, node_cmd_timeout, validators_data) as node:

                wait_for( node_started(node),
                          node_start_timeout,
>                         "Bootstrap node didn't start correctly")
E               Failed: Bootstrap node didn't start correctly

src/tools/rnode.py:193: Failed
```

I really think this is as simple and readable as you can probably get.

ArturGajowy: Both variants are equally readable to me.

OTOH I think - if not for other concerns - all of us would prefer a composable, no-duplication apporach like the higher-order functions one. 

All that being said, my feeling is that after seeing a stacktrace for a failed `wait_for` we'll all be on the same page.

@odeac can you provide an example stacktrace when this fails due to 1) timeout after many checks, 2) failed requirement (not timeout)?

@adaszko can you provide yours as well, so that we're sure the issues you've mentioned are present in the example?

odeac: This shows how a failure looks with such a tool: https://gist.github.com/odeac/171667c9bee66a317dd8c22a678572af

For me it's a no brainer that in this case we should prefer a solution with focus on readability .

adaszko: It would be nice if a message such as `Condition not fulfilled yet (same as above). Time left: 20s. Sleeping 3s...` said *what* condition it is waiting to be satisfied.  The word 'Condition' by itself isn't helpful to the user.

nashef: Anyway, I vote with @odeac here. I don't see why you wouldn't use `wait_for` since it's a standard part of the coroutines library. If it becomes a problem to debug, @adaszko will get to say, "I told you so!" But, I'm dubious since the output @odeac provided seems pretty readable and the code is definitely nice and readable.

odeac: > It would be nice if a message such as `Condition not fulfilled yet (same as above). Time left: 20s. Sleeping 3s...` said _what_ condition it is waiting to be satisfied. The word 'Condition' by itself isn't helpful to the user.

I choose to print "same as above" for two reasons:
1. to reduce clutter in the logs
2. to make it obvious that the status hasn't changed since the previous check. Otherwise, if the difference in the messages is subtle, the difference could be easy to miss

PR #1556, opened 2018-09-25
Disable repl test in p2p-test-tool.py

woky:
It doesn't check any failures.

### DIFF
--- a/scripts/p2p-test-tool.py
+++ b/scripts/p2p-test-tool.py
@@ -113,7 +113,7 @@ parser.add_argument("-T", "--tests-to-run",
                     dest='tests_to_run',
                     type=str,
                     nargs='+',
-                    default=['count', 'eval', 'repl', 'propose', 'errors', 'RuntimeException'],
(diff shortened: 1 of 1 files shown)


woky:
> I'm not sure... it seems o do _some_ checks - that the repl execution didn't fail:
> [rchain/scripts/p2p-test-tool.py](https://github.com/rchain/rchain/blob/08c6c51c8406460a9375ef0e3c2d8077d78371ce/scripts/p2p-test-tool.py#L756)
> 
> Line 756 in [08c6c51](/rchain/rchain/commit/08c6c51c8406460a9375ef0e3c2d8077d78371ce)
> 
>  except Exception as e:

It uses expect to communicate with REPL and the exception thrown from expect will most probably not be related to error in REPL. There's not even timeout argument passed to `run_command`, so hang won't be reported neither. It'll most probably be some IO error, e.g. closed pipe or something like that, but that's really narrow coverage for quite a big amount of work this test does.

I'll merge this for now, as it fills our test with too much junk. Feel free to improve and enable.

PR #1560, opened 2018-09-25
Test cost accounting in the reducer

goral09:
## Overview
Tests correctness of cost accounting in the rholang interpreter.

### Notes


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -81,7 +81,7 @@ class ChargingReducer[M[_]](implicit R: Reduce[M], C: CostAccountingAlg[M]) {
 
 object Reduce {
 
-  private def substituteAndCharge[A: Chargeable, M[_]: Substitute[?[_], A]: Sync](
+  def substituteAndCharge[A: Chargeable, M[_]: Substitute[?[_], A]: Sync](
```
odeac: I don't agree to make these public in order to be tested. 

What we need to test here for cost accounting is `costAccountingAlg.charge(Cost(Chargeable[A].cost(term)))` not the whole thing

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -119,7 +119,7 @@ object Reduce {
         )
       )
 
-  private def spatialMatchAndCharge[M[_]: Sync](target: Par, pattern: Par)(
+  def spatialMatchAndCharge[M[_]: Sync](target: Par, pattern: Par)(
```
odeac: here we should extract the cost-accounting stuff in a function and test that separately. I'm thinking of lines 128-142

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -0,0 +1,151 @@
...
+
+    val program = Interpreter.buildNormalizedTerm(new StringReader(raw)).value
+
+    implicit val rand   = Blake2b512Random(Array.empty[Byte])
+    implicit val errLog = new ErrorLog()
+    lazy val pureRSpace = ChargingRSpaceTest.createRhoISpace()
+    lazy val (_, reducer, _) =
+      RholangAndScalaDispatcher.create[Task, Task.Par](pureRSpace, Map.empty, Map.empty)
+
+    val plainSendCost = CHANNEL_EVAL_COST + Cost(Chargeable[Channel].cost(channel)) + plainSendStorageCost + SEND_EVAL_COST
```
odeac: for this test we should use a simpler model because the focus is not on the aggregation of the cost but on the parallel execution stuff

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -0,0 +1,151 @@
...
+      nestedSendStorageCost + SEND_EVAL_COST
+    val initPhlos = plainSendCost + nestedSendCost - SEND_EVAL_COST
+    reducer.setAvailablePhlos(initPhlos).runSyncUnsafe(1.second)
+
+    val test   = reducer.inj(program)
+    val result = test.attempt.runSyncUnsafe(5.seconds)
+    val store  = pureRSpace.store.toMap
+    val x      = store.get(List(channel))
+
+    val left = reducer.getAvailablePhlos().runSyncUnsafe(1.second)
```
odeac: not used

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -0,0 +1,151 @@
...
+    implicit val errorLog = new ErrorLog()
+    implicit val rand     = Blake2b512Random(128)
+    implicit val costAlg  = CostAccountingAlg.unsafe[Task](CostAccount(1000))
+    val reducer           = new DebruijnInterpreter[Task, Task.Par](tuplespaceAlg, Map.empty)
+    val send              = Send(Channel(Quote(GString("x"))), Seq(Par()))
+    val test              = reducer.inj(send).attempt.runSyncUnsafe(1.second)
+    assert(test === Left(OutOfPhlogistonsError))
+  }
+
+  it should "stop interpreter threads as soon as deploy runs out of phlo" in {
```
odeac: We don't have the concept of "first" and "second"contract when they run in parallel so our test should not try to make that difference.

Instead what we should do is that we should execute two things in parallel and assert that the resulting cost is an error and the consumed phlos are either the cost of the first one or the cost of the second one but not both

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: We want to test that substitution works for failure and success. Not that we charge proper cost.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: This function was created to encapsulate the cost of the spatial match and charging for it.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: still, these are implementation details. The API of the reducer should not be affected by our testing needs. If that is the case then our architecture is not ok

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
goral09: I can't see how to test that because we don't control how fast the threads reduce. This test is set up to have the _slower_ branch reduce later and have the phlos used before it hits the tuplespace.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: I've been there and had this discussion like many times with Artur. This is not refactoring PR.

goral09: And they aren't implementation details. This is exactly the logic of this method. We have different methods for pure substitution which we test elsewhere.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
goral09: Although I think this test can be generalized so it doesn't care which expression reduced first and just check the final tuplespace content and cost state.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
goral09: The focus of this test is the aggregation of the cost in the face of parallel execution.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
goral09: :+1:

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
odeac: Aggregation of the cost in face of parallel execution meanning that you are testing the aggregation of the costs when the | operator is present. This means that you can get rid of all the other unimportant details.

For testing cost when | is present you need two operations with different costs and expect that you charged either one cost or the other but not both. You can do that with a Rholang code like this:

```rholang
@"c" ! "x" | @"c" ! "xy"
```

That's the bare minimum rholang code needed to test the aggregation of costs for | operator. Everything else is noise.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
odeac: if we have the code `p1 | p2` where p1 has cost c1 and p2 has cost c2 then we have two different valid outcomes for the resulting cost: c1 or c2. We don't care which because both are valid.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: The reducer is doing also cost accounting. To separate that you would need at least a sprint.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
goral09: How do you know how much phlos you need to trigger the desired behavior?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: aftet the discussions we had I understand what you mean that this is not part of the reducer interface but we should then move these functions to the Substitute object. They will be public there and would be tested separately from the reducer. This way everybody is happy

odeac: basically what I'm saying is that this logic:

```scala
 def substituteAndCharge(...)=
   Substitute[M, A]
      .substitute(term)(depth, env)
      .attempt
      .flatMap(
        _.fold(
          th => // On error charge for the initial term
            (th, Cost(Chargeable[A].cost(term)))
          substTerm =>
            (substTerm, Cost(Chargeable[A].cost(substTerm))) 
        )
      )
```

should go to substitute it also becomes simpler as it's not aware of the M

goral09: but it's changing the original logic because 
```scala
Substitute[M, A]
      .substitute(term)(depth, env)
```
failed and already did `F.raiseError(…)` and now you take that error and return it as a tuple. 

Also, this implementation will make the usage of it unnecessarily complex and we loose the expressiveness that comes from monadic chain.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
odeac: aren't the costs the same now? I think you should use strings with different sizes such that the left and right side of | have different costs

goral09: They are the same but it doesn't matter. The effect is that only one of the produces makes it to the tuplespace meaning that the other one was interrupted.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: You are right. I wasn't careful with the types. The charging code is dependent on M. I just copied the piece of code that I thought would be the reusable part. 

See the code below which should be ok:

```scala

  def charge[A : Chargeable, M[_] : Sync](term : A,
                                          action : A => M[A],
                                          costAccountingAlg: CostAccountingAlg[M]) : M[A] =
    action(term)
      .attempt
      .map(
        _.fold(
          th => (Left(th), Cost(Chargeable[A].cost(term))),
          substTerm => (Right(substTerm), Cost(Chargeable[A].cost(substTerm)))))
      .flatMap({case (result, cost) => costAccountingAlg.charge(cost) *> Sync[M].pure(result)})
      .rethrow

  def substituteAndCharge[A: Chargeable, M[_]: Substitute[?[_], A]: Sync](
      term: A,
      depth: Int,
      env: Env[Par],
      costAccountingAlg: CostAccountingAlg[M]
  ): M[A] =
      charge(term,
            Substitute[M, A].substitute(_)(depth, env),
            costAccountingAlg)
```

In this code the function `charge` is the reusable code that can be extracted and unit-tested outside the reducer and then the reducer only adds the choice of choosing the substitute policy on top of it. 

Probably the function `spatialMatchAndCharge` could be written in terms of this `charge` as well.

goral09: `spatialMatchAndCharge` is bit different, because the notion of failure is totally something else there.
Also, I don't like that you change the error handling logic:
```scala
 .map(
        _.fold(
          th => (Left(th), Cost(Chargeable[A].cost(term))),
          substTerm => (Right(substTerm), Cost(Chargeable[A].cost(substTerm)))))
      .flatMap({case (result, cost) => costAccountingAlg.charge(cost) *> Sync[M].pure(result)})
      .rethrow
```
The original implementation is simpler IMO. 

I don't like  the idea of the rewrite because we have duplicated 2 lines of code. I don't think that this method is tested incorrectly.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1960,4 +1961,33 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+          Receive(
+            Seq(
+              ReceiveBind(
+                Seq(ChanVar(FreeVar(0)), ChanVar(FreeVar(1)), ChanVar(FreeVar(2))),
+                Quote(GString("channel"))
+              )
+            ),
+            Par(),
+            false,
+            3,
```
odeac: What does 3 represent?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: I'm ok with this

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -20,6 +21,45 @@ trait Substitute[M[_], A] {
 }
 
 object Substitute {
+
+  def substituteAndCharge[A: Chargeable, M[_]: Substitute[?[_], A]: Sync](
```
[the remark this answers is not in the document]
odeac: your proposal is good. I'm taking this back

rholang/src/test/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpaceTest.scala
```diff
@@ -216,11 +218,11 @@ object ChargingRSpaceTest {
           BindPattern,
           errors.OutOfPhlogistonsError.type,
           ListChannelWithRandom,
-          ListChannelWithRandom
+          ListChannelWithRandomAndPhlos
         ]
-    ): Id[Option[(TaggedContinuation, immutable.Seq[ListChannelWithRandom])]] = ???
-    override def createCheckpoint(): Id[Checkpoint]                           = ???
-    override def reset(root: Blake2b256Hash): Id[Unit]                        = ???
+    ): Id[Option[(TaggedContinuation, immutable.Seq[ListChannelWithRandomAndPhlos])]] = ???
```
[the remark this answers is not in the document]
goral09: Yes, we don't invoke them in the test.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1960,4 +1961,33 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
     )
   }
+
+  "Running out of phlogistons" should "stop the evaluation" in {
+    implicit val errorLog = new ErrorLog()
+
+    val test = withTestSpace(errorLog) {
+      case TestFixture(space, reducer) =>
+        implicit val env = Env.makeEnv[Par]()
+        reducer.setAvailablePhlos(Cost(5)).runSyncUnsafe(1.second)
```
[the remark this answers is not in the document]
goral09: I don't know how much the program costs but I know that 5 PHLOs is not enough. I've extracted it to the `noEnoughPhlos` variable.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
goral09: Number of free variables in the expression.

odeac: ok. nevermind

PR #1563, opened 2018-09-25
Prevent sending or receiving on forgeable names.

guardbotmk3:
## Overview
A forgeable name can always be intercepted. Because of this we don't want users to use them thinking they are secure. We therefore treat a send or receive on a forgeable channel as if they never occurred.

### Jira ticket
https://rchain.atlassian.net/browse/RHOL-701


models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -502,4 +502,131 @@ object implicits {
           case _ => BitSet()
         }
     }
+
+  implicit val ChannelIsForgeable: IsForgeable[Channel] =
+    new IsForgeable[Channel] {
+      def isForgeable(chan: Channel) =
+        chan.channelInstance match {
+          case Quote(p) => ParIsForgeable(p)
+          case _        => true
```
goral09:
```scala
case ChanVar(Var(BoundVar(…))) => 
```
if it binds to unforgeable name that this should be `false` right?

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -502,4 +502,131 @@ object implicits {
...
+    }
+
+  implicit val ConnectiveIsForgeable: IsForgeable[Connective] =
+    new IsForgeable[Connective] {
+      def isForgeable(conn: Connective) =
+        conn.connectiveInstance match {
+          case ConnAndBody(body)        => body.ps.forall(ParIsForgeable.apply)
+          case ConnOrBody(body)         => body.ps.forall(ParIsForgeable.apply)
+          case ConnNotBody(body)        => ParIsForgeable(body)
+          case VarRefBody(_)            => true
```
goral09: OK, so I see that you took the approach that any variable is forgeable, even if it might reference unforgeable one?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -316,24 +316,33 @@ object Reduce {
...
-        substData <- data.traverse(
-                      p =>
-                        substituteAndCharge[Par, M](p, 0, env, costAccountingAlg)
-                          .map(par => Channel(Quote(par)))
-                    )
-        _ <- produce(unbundled, substData, send.persistent, rand)
-        _ <- costAccountingAlg.charge(SEND_EVAL_COST)
+        _ <- {
+          if (ParIsForgeable.isForgeable(subChan.value))
+            Applicative[M].pure(())
```
ArturGajowy: Please add a test showing that sending on a forgeable channel is a NOOP. So far we only have tests that check various cases when sending to a channel isn't a noop.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
ArturGajowy: what worries me is that there isn't a single `false` literal in the added code :)

Also I'd expect a case for a GPrivate.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -502,4 +502,131 @@ object implicits {
...
+    new IsForgeable[Par] {
+      def isForgeable(par: Par) =
+        par.ids.isEmpty &&
+          par.sends.forall(SendIsForgeable.apply) &&
+          par.receives.forall(ReceiveIsForgeable.apply) &&
+          par.news.forall(NewIsForgeable.apply) &&
+          par.exprs.forall(ExprIsForgeable.apply) &&
+          par.matches.forall(MatchIsForgeable.apply) &&
+          par.bundles.forall(BundleIsForgeable.apply) &&
+          par.connectives.forall(ConnectiveIsForgeable.apply)
```
ArturGajowy: I see no way of telling whether all relevant properties of every AST variant are included here.

Pattern matching kind of helps with that, because one can at least see if all the (relevant) things on the left have been used on the right.

Ideally, we'd derive those typeclasses using a scheme of 'a product is fogreable if all its components are', with the sole exception of GPrivate - which isn't forgrable.

I'm not saying we should do neither of those now.

models/src/main/scala/coop/rchain/models/IsForgeable.scala
```diff
@@ -0,0 +1,6 @@
+package coop.rchain.models
+
+trait IsForgeable[T] {
```
ArturGajowy: Please add a unit test exploring a couple of cases:

- one for a positive result (is fograble)
- one for a negative result (is fograble)
- a negative and a positive result involving nesting.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
goral09: Because it is `IsForgeable` - it checks whether the `Par` is forgeable. Being not forgeable is just the negation.

goral09: There's no `GPrivate` explicitly but the check is here:
> par.ids.isEmpty

ArturGajowy: I'd say this case should be done using forAll, and the GPrivate case should be explicit. Ohterwise we're making a strong assumption that a GPrivate can only appear in par.ids. Which could change in the future, and I wouldn't like to be the miser debugging this.

guardbotmk3: I disagree. We explicitly separated GPrivates to make this easy. I can add the instance for completeness if it bugs you, but it's unused code.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
guardbotmk3: We do the check post substitution. So any variables left are forgeable ones.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
goral09: > Ideally, we'd derive those typeclasses using a scheme of 'a product is fogreable if all its components are', with the sole exception of GPrivate - which is unforgrable.

This idea sounds cool but I am not sure whether it's doable at the moment. Maybe once you finish with your work and we have better understanding of type class derivation and overwriting protos we can do that.

models/src/main/scala/coop/rchain/models/IsForgeable.scala
guardbotmk3: Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
guardbotmk3: Done.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
guardbotmk3: On the plus side, the polarity matters, so the default will be that things are forgeable.
So if we add a new construct, it will be obviously broken, rather than un-obviously insecure.

guardbotmk3: I didn't do it on purpose, but I might have re-arranged it after this discussion if it went the other way.

goral09:
Once this PR is merged won't this mean that all util contracts that @birchmd is using, like `@("ListOps", "fold")`, `@("Either", _x_)` will stop working?

dzajkowski:
Closing as its not mergeable (easily).
Will keep it for reference.

PR #1568, opened 2018-09-26
Cap storage content when printed.

goral09:
To cap storage content set env variable `PRETTY_PRINTER_OUTPUT_TRIM` to true/1/on/yes.

## Overview

### Notes

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,9 @@ stages:
 test:integration-test-for-p2p-network:
   stage: test 
   tags:
-    - preloaded-ubuntu1804 
+    - preloaded-ubuntu1804
+  before_script:
+    - export PRETTY_PRINTER_OUTPUT_TRIM_AFTER=150
   script:
     - ./scripts/install_bnfc.sh
     - sudo sbt -Dsbt.log.noformat=true clean rholang/bnfc:generate casper/test:compile node/docker:publishLocal
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ matrix:
   - language: nix
     env: SUBPROJECT=rosette
   - language: scala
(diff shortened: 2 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -5,7 +5,9 @@ stages:
 test:integration-test-for-p2p-network:
   stage: test 
   tags:
-    - preloaded-ubuntu1804 
+    - preloaded-ubuntu1804
+  before_script:
+    - export CAP_STORAGE_CONTENT=true
```
ArturGajowy: it's not about just storage content, but PrettyPrinter output in general.

Also, would much prefer it to be a number that - when non-negative - would be the length cap value.

So `PRETTY_PRINTER_OUTPUT_TRIM_AFTER=200`

(100 chars is a bit too little)

goral09: I agree that I can change the name (because we actually cap the PrettyPrinter itself) but I am against adding another variable. I think 100 chars is enough, most of the time when you will cap you are not interested in the content at all. Just make sure that your code made to the storage and that's enough.

ArturGajowy: In MatchTest.scala I hava a truncate() function that uses a bigger limit - so it's not enough. Useful match explanations would get truncated.

If you're not interested at all, just disable output whatsoever. But we are interested. And need a limit of at least 120.

Having the property a numeric one instead of a boolean would help SREs if they ever need more detail (just change the thing in .travis.yml instead of not even thinking about finding the actual limit in code). It's not much work. If you disagree, let's have another voice.

goral09: > In MatchTest.scala I hava a truncate() function that uses a bigger limit - so it's not enough. Useful match explanations would get truncated.

OK, that convinced me.

PR #1570, opened 2018-09-26
Switch GrpcKademliaRPC to grpc-monix

DoSOfRedRiver:
https://rchain.atlassian.net/browse/CORE-1233


rabbitonweb:
@sebtomba are you planning to comment on this?
@DoSOfRedRiver i think this is ready to merge, just go with it :)

PR #1585, opened 2018-09-28
Add 'dump block' functionality

DoSOfRedRiver:
## Overview
Draft of the `dump` program for node, contains the implementation of "should store the current state of the block" part.

https://rchain.atlassian.net/browse/CORE-1234


casper/src/main/scala/coop/rchain/casper/util/comm/DumpProgram.scala
```diff
@@ -0,0 +1,49 @@
...
+import scala.language.higherKinds
+
+class DumpProgram[F[_]: Sync: DeployService: DateTime] {
+  val formatter      = DateTimeFormatter.ofPattern("YYYY-mm-dd_HH-mm-ss")
+  val charset        = Charset.forName("UTF-8")
+  val standardPrefix = "node_dump"
+
+  private def filename: F[String] =
+    for {
+      dateTime  <- DateTime[F].dateTime
```
goral09: Do we really need all that code to do
```scala
dateTime <- Sync[F].delay(LocalDateTime.now())
```
?

In my opinion this is an overkill.

DoSOfRedRiver: Not sure about that btw, open for discussion.
My impression was that it might be helpful to have such functionality, but it doesn't make sense if we will never use it again ofc.

casper/src/main/scala/coop/rchain/casper/util/comm/DumpProgram.scala
```diff
@@ -0,0 +1,49 @@
...
+  private def fileResource: F[Resource[F, BufferedWriter]] =
+    for {
+      name <- filename
+      path = Paths.get(name)
+      res = Resource.fromAutoCloseable(
+        Sync[F].delay(Files.newBufferedWriter(path, charset))
+      )
+    } yield res
+
+  private def dumpBlocks(write: F[String => Unit]): F[Unit] =
```
goral09: I have never really seen good usage of `ap` / `Applicative` in the wild (not counting the `|@|` operator) :)

casper/src/main/scala/coop/rchain/casper/util/comm/DumpProgram.scala
```diff
@@ -0,0 +1,49 @@
...
+    write.ap(DeployService[F].showBlocks())
+
+  private def dumpNode(writer: BufferedWriter): F[Unit] = {
+    val write = Sync[F].delay[String => Unit](writer.write)
+
+    for {
+      _ <- dumpBlocks(write)
+    } yield ()
```
goral09: Isn't this unnecessary? `dumpBlock` return `F[Unit]` already so you can get rid of whole `for-comprehension`

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -373,5 +373,10 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
   }
   addSubcommand(propose)
 
+  val dump = new Subcommand("dump") {
+    descr("Dumps the current state of the Node to the file")
```
KentShikama: Technically it just dumps all the blocks in the current main chain of the fork choice, so I would just say "Dumps the current state of the blockDAG to the file" (or chain).

casper/src/main/scala/coop/rchain/casper/util/comm/DumpProgram.scala
DoSOfRedRiver: You right, but I will add more functionality later and `for-comprehension` will eventually appear.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
DoSOfRedRiver: We are planned to dump more info like heapdump, metrics, the configuration used, so I think the "state" word reflects this correctly.

casper/src/main/scala/coop/rchain/casper/util/comm/DumpProgram.scala
KentShikama: Sounds good

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
KentShikama: Sounds good

pyrocto:
@KentShikama You have to explicitly mark the PR as approved.  LGTM doesn't work.

PR #1586, opened 2018-09-28
Enable registry for usage on-chain

birchmd:
## Overview
This PR introduces a system process for populating the registry root with an empty map and calls that process in the genesis block. We test that the registry works properly in `HashSetCasperTest`.

https://rchain.atlassian.net/browse/RHOL-892

### Notes


rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -77,6 +77,28 @@ object SystemProcesses {
...
 
+  def initRegistry(
+      space: RhoISpace,
+      dispatcher: Dispatch[Task, ListChannelWithRandom, TaggedContinuation]
+  ): Seq[ListChannelWithRandom] => Task[Unit] = {
+    case Seq(ListChannelWithRandom(_, rand, cost)) =>
+      Task.now {
+        val rootPar: Par         = Registry.registryRoot
+        val rootChannel: Channel = Channel(Quote(rootPar))
+        val rootData             = space.getData(rootChannel)
```
guardbotmk3: This won't work. Because of the way we do locking, someone may be able to race a lookup and insert a second root. It may be better to have the genesis code specifically do a produce.

goral09:
I don't understand something - is registry not usable without this?

birchmd:
@goral09 The registry needs to have an empty map already on the channel in order to work.

goral09:
Right, otherwise there's nothing to recurse into or add to. Makes sense @birchmd .

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -224,15 +227,37 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
...
+    val emptyChPt = active.space.createCheckpoint()
+    val registryInit: Task[Array[Byte]] = for {
+      _          <- active.reducer.setAvailablePhlos(max)
+      _          <- active.reducer.inj(produceTerm)
+      checkpoint = active.space.createCheckpoint()
+      hash       = checkpoint.root.bytes.toArray
+      _          <- active.reducer.setAvailablePhlos(zero)
+      _          <- active.replayReducer.setAvailablePhlos(max)
+      _          = active.replaySpace.rig(emptyChPt.root, checkpoint.log)
+      _          <- active.replayReducer.inj(produceTerm)
```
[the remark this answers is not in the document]
birchmd: Calling `rig` only tells the replay rspace what it should expect. Nothing actually happens until to give it the terms though. This is where the "unused comm event" error comes from, if you create a checkpoint (indicating that you are done replaying) but the replay still expected there to be more terms then it will throw that error. `rig` also sets things up so that comm events happen in the same way when the terms are actually added.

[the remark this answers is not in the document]
goral09: Thank you very much sir.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -233,9 +233,10 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
 object RuntimeManager {
   type StateHash = ByteString
 
-  def fromRuntime(active: Runtime): RuntimeManager = {
+  def fromRuntime(active: Runtime)(implicit scheduler: Scheduler): RuntimeManager = {
     active.space.clear()
     active.replaySpace.clear()
+    active.injectEmptyRegistryRoot[Task].unsafeRunSync
```
guardbotmk3: I feel like this should be in Genesis.scala
You don't want to do this every time you create a Runtime, only when you create the Genesis block.

birchmd: I disagree. I can't think of any time that we would want a `RuntimeManager`, but not want to be able to use the registry. This is just part of the setup process now; we clear everything, initialize the registry and then snapshot that state so that we can go back to an "empty" tuplespace whenever we need to.

goral09: I had the same concern as @guardbotmk3 . It doesn't feel right that initialization of the Registry should be in `RuntimeManager`.

PR #1587, opened 2018-09-28
Rholang: Remove Channel and Quote from the proto

guardbotmk3:
They aren't necessary, and they complicate working with the protos.

Fixed everywhere they are used, and all tests pass after the removal. After this
commit, you will need to run sbt rholangProtoBuild/clean casper/clean at a
minimum. This a breaking change for the term encoding, which is why I'm trying
to get it in before the release.


goral09:
I like that `Quote` was an explicit way of saying - "here, this is a name now". It felt that we are closer tho rho calculus.

goral09:
Hmm, I don't see the removal of `Quote` from the `RhoTypes.proto` @guardbotmk3 .

models/src/main/protobuf/RhoTypes.proto
```diff
@@ -36,20 +36,13 @@ message ParWithRandom {
     bytes randomState = 2 [(scalapb.field).type = "coop.rchain.crypto.hash.Blake2b512Random"];
 }
 
-message Channel {
-    oneof channel_instance {
-        Par quote = 1;
-        Var chanVar = 2;
-    }
-}
-
```
guardbotmk3: @goral09 It's removed right here.

guardbotmk3:
> I like that Quote was an explicit way of saying - "here, this is a name now". It felt that we are closer tho rho calculus.

I like it for user input
But once we've validated the user input, it's not necessary anymore. When writing scala, we needed to pull things apart twice, rather than once. Look at how much it shortens the registry.

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -70,7 +69,7 @@ class CollectMatcherSpec extends FlatSpec with Matchers {
     result.par should be(
       inputs.par.prepend(
         EList(
-          List[Par](EVar(BoundVar(1)), EEvalBody(ChanVar(BoundVar(0))), GInt(7)),
+          List[Par](EVar(BoundVar(1)), EVar(BoundVar(0)), GInt(7)),
```
goral09: I am really glad we are getting rid of this. I didn't like that `BoundVar`s could be existing both in a `Channel` and as a `Var`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
```diff
@@ -192,8 +191,8 @@ class ProcPrinterSpec extends FlatSpec with Matchers {
       )
     val target =
       """new x0 in {
-        |  for( x1 <- x0 ) {
-        |    *x1
+        |  for( @{x1} <- @{x0} ) {
```
goral09: So it seems that previously it was printing them wrong?

PR #1589, opened 2018-09-28
Remove Replay performed by proposing node

AbnerZheng:
## Overview
Remove Replay performed by proposing node

https://rchain.atlassian.net/browse/RHOL-651

### Notes

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
@@ -46,7 +46,8 @@ class HashSetCasperTest extends FlatSpec with Matchers {
     val nodes = HashSetCasperTestNode.network(validatorKeys.take(2), genesis)
     val deployData = ProtoUtil.sourceDeploy(
       """new s(`rho:io:stdout`) in { s!("Hello, World!") }""",
-      System.currentTimeMillis()
+      System.currentTimeMillis(),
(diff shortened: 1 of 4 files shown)


AbnerZheng:
@birchmd PTAL

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -42,6 +42,31 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+
+    nodes(0).casperEff.addBlock(signedBlock)
+    nodes(1).receive()
+
+    val received = nodes(1).casperEff.contains(signedBlock)
+    received should be(true)
+    nodes(0).logEff.infos
+      .count(it => it.contains("skip replay the deploys sent from itself")) should be(
+      1
+    )
```
AbnerZheng: Before this commit, the "Hello, World!" will be printed three times in your console. Once for node0 deploy the data, and twice for both node1 and node0 replay the RSpace. And now, it will be only be printed twice. But the output is by the interpret. I have no idea to catch them. So I use the info log to prove.

birchmd:
I also don't like adding the log message saying that something did not happen because it is not supposed to happen. I think log messages should always be relevant to the users. It could be that there is not a good way to check in a unit test that replay is not happening for a node's own blocks. And that's fine because we can see it working properly in performance tests.

PR #1590, opened 2018-10-01
Add --depth to show blocks client endpoint and depth to listen-at-name GRPC endpoint

KentShikama:
We also make the show blocks endpoint list all blocks (not just main chain ones) and move the old show-blocks to showMainChain, which is now a GRPC API only.

Example usage below

```
$ ./bin/rnode show-blocks --depth 2

------------- block 1 ---------------
blockHash: "7eddba159490a12bf7872e98da7a83c3f978626834bfefb9ea39755a5a3fff5c"
blockSize: "3071"
blockNumber: 1
version: 0
deployCount: 1
tupleSpaceHash: "99147a32691e6a5556fc899a86a3bd54a1c46bba3c0b93fb3f2ce6ebaa96771f"
timestamp: 1538543714452
faultTolerance: -1.0
mainParentHash: "78b63643a4f5966b3ea6899e673cc11df2057c06b1202c432a2070016650d3b8"
parentsHashList: "78b63643a4f5966b3ea6899e673cc11df2057c06b1202c432a2070016650d3b8"
sender: "0d0a8788cbad40b45f812167f18e2e44f6e3444bc248adf96d5cafacefd62442"

-----------------------------------------------------


------------- block 0 ---------------
blockHash: "78b63643a4f5966b3ea6899e673cc11df2057c06b1202c432a2070016650d3b8"
blockSize: "47110"
blockNumber: 0
version: 0
deployCount: 8
tupleSpaceHash: "9a7881b5a8738da6d62727dd7492f76750ea8570d3f676125abcc683bb4a3997"
timestamp: 1
faultTolerance: -0.8666667
mainParentHash: ""
sender: ""

-----------------------------------------------------


count: 2

```

By default the --depth is 1

```
$ ./bin/rnode show-blocks

------------- block 1 ---------------
blockHash: "7eddba159490a12bf7872e98da7a83c3f978626834bfefb9ea39755a5a3fff5c"
blockSize: "3071"
blockNumber: 1
version: 0
deployCount: 1
tupleSpaceHash: "99147a32691e6a5556fc899a86a3bd54a1c46bba3c0b93fb3f2ce6ebaa96771f"
timestamp: 1538543714452
faultTolerance: -1.0
mainParentHash: "78b63643a4f5966b3ea6899e673cc11df2057c06b1202c432a2070016650d3b8"
parentsHashList: "78b63643a4f5966b3ea6899e673cc11df2057c06b1202c432a2070016650d3b8"
sender: "0d0a8788cbad40b45f812167f18e2e44f6e3444bc248adf96d5cafacefd62442"

-----------------------------------------------------


count: 1
```

Part of https://rchain.atlassian.net/browse/CORE-1261 and https://rchain.atlassian.net/browse/RHOL-759


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -218,25 +221,66 @@ object BlockAPI {
...
+        startHeight = dag.topoSort.length - depth
+        indexedBlockInfosAtHeightList <- dag.topoSort.foldM(
+                                          (startHeight, List.empty[BlockInfosAtHeight])
+                                        ) {
+                                          case (
+                                              (height, blockInfosAtHeightAcc),
+                                              blockHashesAtHeight
+                                              ) =>
+                                            for {
+                                              blocksAtHeight <- blockHashesAtHeight.toList.traverse(
```
birchmd: You don't need to convert to `List`, `traverse` is also defined on `Vector`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -218,25 +221,66 @@ object BlockAPI {
...
 
-  def getBlocksResponse[F[_]: Monad: MultiParentCasperRef: Log: SafetyOracle: BlockStore]
-    : F[BlocksResponse] = {
+  def showBlocks[F[_]: Monad: MultiParentCasperRef: Log: SafetyOracle: BlockStore](
+      depth: Long
+  ): F[List[BlockInfosAtHeight]] = {
+    def casperResponse(implicit casper: MultiParentCasper[F]) =
+      for {
+        dag         <- MultiParentCasper[F].blockDag
+        startHeight = dag.topoSort.length - depth
```
birchmd: Something here is wrong. I think the goal of the PR is to only show blocks that are at most `depth` back from the current forkchoice, but this is not what is being done in this function. The height of the blocks at `dag.topoSort.head` is `dag.sortOffset`, so `startHeight` should be `dag.topoSort.length + dag.sortOffset - 1 - depth`. But more than that, the fold goes over the entire sort, so `depth` is actually not doing what it is supposed to. You should fold over `dag.topoSort.takeRight(depth)`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -83,11 +83,12 @@ object BlockAPI {
     )
 
   def getListeningNameDataResponse[F[_]: Sync: MultiParentCasperRef: Log: SafetyOracle: BlockStore](
+      depth: Long,
```
birchmd: `depth` should be an `Int` if it is the number of blocks back that you are going as opposed to the height that you with to stop at.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -83,11 +83,12 @@ object BlockAPI {
...
       listeningName: Channel
   ): F[ListeningNameDataResponse] = {
     def casperResponse(implicit casper: MultiParentCasper[F], channelCodec: Codec[Channel]) =
       for {
-        mainChain           <- getMainChainFromTip[F]
+        mainChain           <- getMainChainFromTip[F](depth)
```
birchmd: Why do we only look at the main chain here anyways? Names could be effected anywhere in the DAG, so we should traverse the whole thing, not just a single chain.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -114,14 +115,15 @@ object BlockAPI {
...
 
   def getListeningNameContinuationResponse[F[_]: Sync: MultiParentCasperRef: Log: SafetyOracle: BlockStore](
-      listeningNames: Channels
+      depth: Long,
+      listeningNames: Seq[Channel]
   ): F[ListeningNameContinuationResponse] = {
     def casperResponse(implicit casper: MultiParentCasper[F], channelCodec: Codec[Channel]) =
       for {
-        mainChain           <- getMainChainFromTip[F]
+        mainChain           <- getMainChainFromTip[F](depth)
```
birchmd: Same here.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -22,9 +22,10 @@ service DeployService {
   rpc addBlock(BlockMessage) returns (DeployServiceResponse) {}
   rpc createBlock(google.protobuf.Empty) returns (DeployServiceResponse) {}
   rpc showBlock(BlockQuery) returns (BlockQueryResponse) {}
-  rpc showBlocks(google.protobuf.Empty) returns (stream BlockInfoWithoutTuplespace) {}
-  rpc listenForDataAtName(Channel) returns (ListeningNameDataResponse) {}
-  rpc listenForContinuationAtName(Channels) returns (ListeningNameContinuationResponse) {}
+  rpc showMainChain(BlocksQuery) returns (stream BlockInfoWithoutTuplespace) {}
```
birchmd: I don't think this is a useful API to keep around. I know this is what `showBlocks` used to do, but that was wrong, we always should have been showing the whole dag.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -110,6 +121,12 @@ message BlocksResponse {
   int64 length = 3;
 }
 
+message BlockInfosAtHeight {
```
birchmd: I don't understand the point of grouping blocks at the same height when returning the result from `showBlocks`. I know it is grouped like that in the topological sort, but you could just do `dag.topoSort.takeRight(depth).flatten` and then return the `BlockInfoWithoutTuplespace` sequence directly.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -387,11 +389,20 @@ object Configuration {
...
-      case _                        => Help
+      case Some(options.showMainChain) =>
+        import options.showMainChain._
+        ShowMainChain(depth.getOrElse(1L))
+      case Some(options.run) => Run
+      case Some(options.dataAtName) =>
+        import options.dataAtName._
+        DataAtName(
+          depth.getOrElse(default = Long.MaxValue),
+          options.dataAtName.name()
```
DoSOfRedRiver: `options.dataAtName.name()`, but you already have `import options.dataAtName._`.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -387,11 +389,20 @@ object Configuration {
...
+        )
+      case Some(options.contAtName) =>
+        import options.contAtName._
+        ContAtName(depth.getOrElse(default = Long.MaxValue), options.contAtName.name())
```
DoSOfRedRiver: See above.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -218,25 +221,66 @@ object BlockAPI {
...
+        dag         <- MultiParentCasper[F].blockDag
+        startHeight = dag.topoSort.length - depth
+        indexedBlockInfosAtHeightList <- dag.topoSort.foldM(
+                                          (startHeight, List.empty[BlockInfosAtHeight])
+                                        ) {
+                                          case (
+                                              (height, blockInfosAtHeightAcc),
+                                              blockHashesAtHeight
+                                              ) =>
+                                            for {
```
DoSOfRedRiver: I would prefer to extract such large expressions into functions to increase readability a little bit.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: The main chain block's state hash will allow you to reset the tuplespace to everything you care about, and in any case, the other blocks will only get finalized once the main chain blocks are finalized.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: So we can't print out a depth of 3 million?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: Nice catch; I'll make sure to include a test for this too.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
birchmd: Sure you can, `Int.MaxValue = 2,147,483,647`, which is over 2 billion. The point is that this API should not support more than that. If you need more than that then you should just copy the state directly (like literally copying the files used by casper to store blocks); it would be much faster that way.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
birchmd: I see. So the idea is that a block with multiple parents implicitly also contains the state changes that occurred in its non-main parent?

KentShikama: Indeed

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: Oh lol. I think for the past 2 years I've been thinking 2.1 million is the max for ints. Whoops.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: Done

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
KentShikama: Done

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
KentShikama: Done

models/src/main/protobuf/CasperMessage.proto
KentShikama: I thought it would be clearer if it is grouped and thus ordered by height. Otherwise it will be ordered sort of by height but actually randomly.

KentShikama:
@birchmd @DoSOfRedRiver PTAL. I've updated the description of this PR to reflect the changes as discussed with @birchmd yesterday.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -218,25 +221,59 @@ object BlockAPI {
...
+    MultiParentCasperRef.withCasper[F, List[BlockInfoWithoutTuplespace]](
+      casperResponse(_),
+      List.empty[BlockInfoWithoutTuplespace]
+    )
+  }
+
+  private def getFlattenedBlockInfosUntilDepth[F[_]: Monad: MultiParentCasper: Log: SafetyOracle: BlockStore](
+      depth: Int,
+      dag: BlockDag,
+      startHeight: Long
```
birchmd: Unused function argument

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -3,7 +3,7 @@ package coop.rchain.casper.util.comm
 import cats.{Id, Monad, MonadError}
 import cats.effect.Sync
 import cats.implicits._
-
+import coop.rchain.casper.protocol
```
birchmd: New unused import?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -73,9 +73,18 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
...
-  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] = {
+  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] =
+    Try(tryComputeBonds(hash)) match {
+      case Success(value) => value
+      case Failure(ex) =>
+        ex.printStackTrace()
+        throw ex
+    }
+
+  private def tryComputeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] = {
```
birchmd: Maybe `unsafeComputeBonds` would be a better name?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -73,9 +73,18 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
     result
   }
 
-  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] = {
+  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] =
+    Try(tryComputeBonds(hash)) match {
+      case Success(value) => value
+      case Failure(ex) =>
+        ex.printStackTrace()
```
birchmd: I definitely don't love this, but maybe it is the best we can do for now.

casper/src/test/scala/coop/rchain/casper/api/BlocksResponseAPITest.scala
```diff
@@ -104,11 +104,39 @@ class BlocksResponseAPITest
...
+  "showBlocks" should "return all blocks" in {
+    val blocksResponse =
+      BlockAPI.showBlocks[Id](Int.MaxValue)(
+        syncId,
+        casperRef,
+        logEff,
+        turanOracleEffect,
+        blockStore
+      )
+    blocksResponse.length should be(8) // TODO: Switch to 4 when we implement block height correctly
```
birchmd: What is this TODO message about?

KentShikama: This is what we discussed in discord chat a few days ago. We just increment blockNumber by 1 instead of max(parents) + 1 in tests.

PR #1591, opened 2018-10-01
CORE-1244 stream blob

rabbitonweb:
## Overview
* Add streamBlob method
    
    This method is not yet implemented. It will eventually alow sending
    Blob messages in a stream manner. Blobs will be put on a queue where
    they will be consumed by senders.

* Added BlockMessage to ServerMessage
    
    This way our queue that handles messages on the server side can be
    reused for Blob messages

* Extend receive method in TransportLayer with handleBlob
    
    After this change when starting up stransport layer, caller has to
    provide handler lambda `handleBlob` that will know how to handle
    incoming blobs.
    In `node.scala` this work is provided by new BlobHandler (not yet
    implemented).

* Provide temporary request-response implementation of streamBlob
    
    In future this will store blobs on a queue that will be send in stream
    manner. For now - in order to have a working implementation - we are
    providing simplest sequential implementation

* Remove latch from TransortLayerSpec

*   Change Dispatcher in test to be more parametric
    
    Dispatcher is used to mock dispatch method in
    TransportLayerSpec. Since we are extending TrasnportLayer.receive
    method with new handleBlob - which will have to be testes as well - it
    seems a good idea to reuse existing Dispatcher logic in tests. In
    order to do so, Dispatcher must become more polymorphic

* Extract protocolDispatcher in Result, simplify the Result
    
    Int TransportLayerSpec `Result` type will have to differencate between `protocolDispatcher` (for
    `Protocol` messages) and `blobDispatcher`. Two methods were removed
    from `Result` (they were only pushing forward the calls to
    dispatcher). Alternatively `protocolDispatcher` was exposed externally
    via `Result`.

* Remove proccessed timestampe from TransportLayerSpec - no longer used

* Provide tests for streamBlob
    
    ```
    [info]   when streamBlob
    [info]   - should send a blob and receive by (single) remote side (72 milliseconds)
    [info]   - should send a blob and receive by (multiple) remote side (54 milliseconds)
    ```

    

CORE-1244

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/BlobHandler.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/BlobHandler.scala
@@ -1,0 +1,8 @@
+package coop.rchain.casper.util.comm
+
+import coop.rchain.comm.protocol.routing._
+import cats._, cats.data._, cats.implicits._
+
+object BlobHandler {
+  def handleBlob[F[_]: Applicative]: Blob => F[Unit] = (blob: Blob) => ().pure[F] // FIX-ME
+}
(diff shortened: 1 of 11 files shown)


rabbitonweb:
@sebtomba no, `Blob` (or `Packet` as it turns out) will stay, chunks will be added as internal implementation detail. You provide a `Blob` (or `Packet`) and `stream` will know how to chunk it

PR #1595, opened 2018-10-01
Assign thread pools to modules

sebtomba:
## Overview
Assign specific thread pools to node modules:

### Scheduler `node-io`
number of cores threads to 2000 (default)
- Main thread pool

### Scheduler `grpc-io`
0 to 2000 threads
- Transport Layer server message queues
- Transport Layer client 
- Kademlia gRPC client/server
- API gRPC (metrics, REPL, deploy)

### Scheduler `loop`
4 threads
- Kademlia lookup
- Connect (handshake)

### Scheduler `tl-dispatcher`
number of cores threads
- Transport Layer dispatcher

This scheme prevents (to some point) starvation of modules. On the other hand, it improves debugging since we are logging thread names.
There is still a lot of room for improvement in Casper and RSpace.

https://rchain.atlassian.net/browse/CORE-1230


goral09:
My math tells me that this PR give  `comm` layer alone gets from 71 threads to max(`max`, 4000) and creates 10 schedulers. This is A LOT.

sebtomba:
@goral09 I can reduce the number of threads in `comm`. Probably we don't need as much. However, what is the problem to have many `Schedulers`? That's the whole point of this PR.

goral09:
@sebtomba Thread pools are heavy, creating a thread pool only for one thread seems like an overkill. How about splitting the `comm` not into domain parts but the thread pool types they need?

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -189,6 +184,7 @@ class NodeRuntime(conf: Configuration, host: String)(implicit scheduler: Schedul
                      val prometheusService     = NewPrometheusReporter.service(prometheusReporter)
                      implicit val contextShift = IO.contextShift(scheduler)
                      BlazeBuilder[IO]
+                       .withExecutionContext(Scheduler.singleThread("http-server"))
```
lukasz-golebiewski: Does this mean that the http server is going to have only a single thread at it's disposal?

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -200,7 +202,7 @@ class TcpTransportLayer(host: String, port: Int, cert: String, key: String, maxM
                        new RuntimeException("TransportLayer server is already started")
                      )
                    case _ =>
-                     val parallelism = Runtime.getRuntime.availableProcessors()
+                     val parallelism = Runtime.getRuntime.availableProcessors() + 1
```
lukasz-golebiewski: Why +1?

sebtomba: @goral09 suggested to do `max(number cores, 2)` because it's best practice. I thought that +1 would achieve the same thing.

node/src/main/scala/coop/rchain/node/node.scala
sebtomba: yes, but I think that I'll remove this due to what @goral09 said

sebtomba:
@goral09 Hmm, I can use one thread pool for all gRPC IO: `tl-grpc`, `tl-grpc-client`, `kademlia-grpc` and `api-grpc`. On the other hand I would create one dedicated thread pool instead of all the thread pools with 1 thread. What do you think?

sebtomba:
And reduce the default for the main thread `node-io` pool from 4000 to 1000

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
goral09: @sebtomba If you have 4 cores, the original version will set `2` and yours `5`.

goral09:
@sebtomba if your i/o is properly non-blocking I don't see a reason why having one fork-join cached thread pool won't work.

goral09:
As already mentioned I would prefer to see executors divided into 3 groups:

1. CPU-bound
2. Blocking IO
3. Non-blocking IO polling

[Here](https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c) is a short write-up that might help.

sebtomba:
@goral09 Do we have blocking IO?

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
tymm: The following article gives a good explanation on why people often use the `+ 1`:
https://medium.com/@sergeykuptsov/how-it-works-in-java-thread-pool-2a15dac75620
Basically to still have maximum parallelism in the case one thread takes a break.

tymm:
FWIW according to https://medium.com/@sergeykuptsov/how-it-works-in-java-thread-pool-2a15dac75620 one can find the optimal thread pool size by looking at the time spent in I/O.

goral09:
@sebtomba sure, RSpace ;)

sebtomba:
@goral09 On a side-project I was able to prove that `Reply` (response `Task`) is non-blocking. On a thread pool with **one** thread, the server added 1000 such `Reply`-Tasks to the message queue.

sebtomba:
@tymm We are using `ForkJoin` pools which automatically increase and shrink the number of threads. So if a thread is not used for (default) 60 seconds it gets removed.

tymm:
@sebtomba But creating a `ForkJoinPool` with too many threads would still hurt performance (at least for the first 60 seconds), right?

goral09:
I still think that you are creating too many threads. `comm` layer is not the only module that needs processor time and you are creating over 4000 threads. Have you tested that it is the optimal number?

goral09:
Good test would be to set up a network consisting of few nodes and running a test that will `deploy` and `propose` from each of the node in the network, for few hours. Nothing fancy, just to force the node to do some work and pass blocks around. Then observe the thread count, memory usage, GC, leaks etc. Have you done something like that?

sebtomba:
@goral09 I'm **not** creating thread pools with 4000 threads! The `ForkJoinPool`s are started with a very small number of threads (the gRPC pool even with 0). The value 2000 is the maximum value for the thread pool.

goral09:
@sebtomba I know, but it can grow to as many as 4000. Do you have any logic that will backpressure the creation of threads somewhere higher in your application layer?

sebtomba:
@goral09 This PR is just the first part of work that needs to be done. I'm not worried that the gRPC thread pool will grow to 2000 threads with our current implementation of the Transport Layer. Also comm runs now on a fixed pool of 4 threads. Received messages are getting dispatched by a second fixed pool (nuber of cores). However, the `node-io` pool is the problem and can maybe grow without control. As I said before, we need to investigate Casper, `Runtime` and `RuntimeManager`. And then we can take a look at Rholang parallel processing.

goral09:
@sebtomba if you're sure you won't need that many threads then decrease the thread pool maximum size. Did you do a test as described [here](https://github.com/rchain/rchain/pull/1595#issuecomment-426634401) ?

sebtomba:
@goral09 I'll run our test harness on this PR

goral09:
How does the test look like? Does it create a lot of blocks?

lukasz-golebiewski:
Something doesn't look right here: http://stress-docker.pyr8.io:8080/rchain/rchain-perf-harness/186/7

sebtomba:
Resolved issue in stress-docker. We need a fixed size `node-io` thread pool. Further investigation is needed why.

PR #1597, opened 2018-10-01
Faucet implementation (no registry usage)

birchmd:
## Overview
Introduces the faucet on the public name. Adds configuration for min/max bonding thresholds as well as the existence of the faucet (so that it can be enabled for testnet and disabled for mainnet).

https://rchain.atlassian.net/browse/RHOL-750
https://rchain.atlassian.net/browse/RHOL-751

### Notes


casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -6,6 +6,14 @@ import coop.rchain.crypto.codec.Base16
 //TODO: include other fields relevent to PoS (e.g. rewards channel)
 case class ProofOfStakeValidator(id: Array[Byte], stake: Long)
 
+case class ProofOfStakeParams(
+    minimumBond: Long,
+    maximumBond: Long,
+    validators: Seq[ProofOfStakeValidator]
+) {
+  require(minimumBond <= maximumBond)
```
goral09: If we do `require` now, is `require(validators.nonEmpty)` also correct?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -59,6 +59,9 @@ object Configuration {
   private val DefaultApprovalProtocolInterval   = 5.seconds
   private val DefaultMaxMessageSize: Int        = 100 * 1024 * 1024
   private val DefaultThreadPoolSize: Int        = 4000
+  private val DefaultMinimumBond: Long          = 1L
+  private val DefaultMaximumBond: Long          = Long.MaxValue
+  private val DefaultHasFaucet: Boolean         = false
```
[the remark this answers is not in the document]
birchmd: Well they are only important when creating the genesis block, so most of the time people won't care about their values. Probably what I should do is replace the `createGenesis` Boolean with an `Option[GenesisConfig]` which includes all the parameters only applicable to creating the genesis block and not have defaults for any of them. I don't think I want to do that refactor in this PR, so I'll just make a ticket for it. https://rchain.atlassian.net/browse/RHOL-911

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
birchmd: Yeah, good call.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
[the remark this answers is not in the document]
goral09: Sounds good :+1:

birchmd:
Thanks for the suggestions @goral09  @KentShikama . PR is updated. PTAL

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Faucet.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import coop.rchain.rholang.mint.BasicWalletFaucet
+
+/**
+  * The purpose of a "Faucet" is to give a place where users
+  * can obtain REV for testing their contracts for free on the
+  * testnet. Since this is only applicable to testnet, the
+  * `noopFaucet` will be used in main net, where there is no
+  * way to obtain REV for free.
+  */
```
goral09: Will the faucet disappear once we start charging from the real wallets?

casper/src/test/scala/coop/rchain/casper/genesis/GenesisTest.scala
```diff
@@ -62,169 +59,178 @@ class GenesisTest extends FlatSpec with Matchers with BeforeAndAfterEach with Bl
...
+      numValidators: Int = numValidators,
+      maybeWalletsPath: Option[String] = None,
+      minimumBond: Long = 1L,
+      maximumBond: Long = Long.MaxValue,
+      faucet: Boolean = false,
+      shardId: String = rchainShardId,
+      deployTimestamp: Option[Long] = Some(System.currentTimeMillis())
+  )(
+      implicit runtimeManager: RuntimeManager,
+      genesisPath: Path,
```
goral09: It feels like the `genesisPath` and `runtimeManager` shouldn't be implicits. It looks like never make use of that fact.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -145,10 +146,13 @@ object Genesis {
       bonds       <- getBonds[F](bondsFile, numValidators, genesisPath)
       timestamp   <- deployTimestamp.fold(Time[F].currentMillis)(_.pure[F])
       initial     = withoutContracts(bonds = bonds, timestamp = 1L, version = 0L, shardId = shardId)
+      validators  = bonds.map(bond => ProofOfStakeValidator(bond._1, bond._2)).toSeq
+      faucetCode  = if (faucet) Faucet.basicWalletFaucet(_) else Faucet.noopFaucet
```
KentShikama: I feel like this is the critical time in which http://degoes.net/articles/destroy-all-ifs applies. We need to somehow make it such that we can't launch mainnet until this is toggled correctly.

casper/src/main/rholang/BasicWalletFaucet.rho
```diff
@@ -0,0 +1,17 @@
...
+      new walletPurseCh in {
+        @(mint, "makePurse")!(amount, *walletPurseCh) |
+        for(@walletPurse <- walletPurseCh) {
+          @"BasicWallet"!(walletPurse, algorithm, publicKey, *return)
+        }
+      }
+    } |
+    return!(*this)
+  }
+}
```
KentShikama: I wonder if allowing for unlimited REV to go around in testnet is a good idea. I know other blockchain testnets do not allow this but maybe that is just a side-effect of PoW.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Faucet.scala
KentShikama: If you look above there is a boolean flag that is switched on for testnet/testing and off for when it is mainnet.

casper/src/main/scala/coop/rchain/casper/util/comm/BlockApproverProtocol.scala
```diff
@@ -110,11 +119,15 @@ object BlockApproverProtocol {
...
-      genesisBlessedDeploys   = genesisBlessedContracts.flatMap(_.raw)
+      validators = blockBonds.toSeq.map(b => ProofOfStakeValidator(b._1.toByteArray, b._2))
+      posParams  = ProofOfStakeParams(minimumBond, maximumBond, validators)
+      faucetCode = if (faucet) Faucet.basicWalletFaucet(_) else Faucet.noopFaucet
+      genesisBlessedContracts = Genesis
+        .defaultBlessedTerms(timestamp, posParams, wallets, faucetCode)
+        .toSet
+      blockDeploys          = body.deploys.flatMap(ProcessedDeployUtil.toInternal)
+      genesisBlessedTerms   = genesisBlessedContracts.flatMap(_.term)
+      genesisBlessedDeploys = genesisBlessedContracts.flatMap(_.raw)
```
KentShikama: This feels like a lot of duplicated code. Given how it has the critical boolean flag, it doesn't help to make it a critical boolean flag in two places.

casper/src/main/rholang/BasicWalletFaucet.rho
goral09: I don't like it either but I hope that this idea has been attacked and considered by "big heads" from different perspectives.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Faucet.scala
birchmd: @goral09 Part of the reason for having the faucet is because we will be charging from wallets soon. We still want people to be able to try out the system even once we start charging, so they will be able to by getting a wallet from the faucet. But as @KentShikama says, of course we won't have this in main net where the REV actually represents something of value.

casper/src/test/scala/coop/rchain/casper/genesis/GenesisTest.scala
birchmd: Yeah, that's fair. I wanted them to be implicit, but then couldn't figure out the multi-variable version of `implicit blockStore => ...` that gets used in the `BlockStore` test fixture.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Faucet.scala
goral09: What I meant by asking this question was if on the main net we will replace faucets with the real wallets.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
birchmd: But the Boolean is not the core API, the `faucetCode` function is. The Boolean is pushed all the way to the edge of the program where we are reading things from the config file. The config necessarily has to be a simplification of the full API because we don't expect people to write code in the config file.
Also, in the config the default is `false` so we should just not have to specify anything to get the right behaviour for main net.

casper/src/main/rholang/BasicWalletFaucet.rho
birchmd: What would be the argument against unlimited REV? We want people to be able to try out the platform, so there needs to be some way of getting REV. If we went with a faucet that had a limited supply then the number of people able to use test net also becomes limited. Or if there was unlimited REV but their was some barrier to obtaining it (e.g sending an application to the coop) then I think people would wonder why this barrier was put in place.

casper/src/main/scala/coop/rchain/casper/util/comm/BlockApproverProtocol.scala
birchmd: But the flag just gets read from the config in both cases. It's not really in two places, it's just referenced in two places. And again this is because this code touches the boundary of the program and so needs to get information from the config.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Faucet.scala
birchmd: Ah, yes, absolutely. And we even already have the infrastructure for that; it's the `walletsFile` that defines all the REV which will be available in main net.

PR #1604, opened 2018-10-03
RHOL-882 Added the ability to concat two byte arrays with  GByteArray…

cboscolo:
… ++ GByteArray

## Overview
Provides support to concat byte arrays.

https://rchain.atlassian.net/browse/RHOL-882

### Notes
N/A

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -707,6 +707,12 @@ object Reduce {
                                  STRING_APPEND_COST * (lhs.length + rhs.length)
                                )
                          } yield Expr(GString(lhs + rhs))
+                       case (GByteArray(lhs), GByteArray(rhs)) =>
+                         for {
+                           _ <- costAccountingAlg.charge(
+                                 STRING_APPEND_COST * (lhs.size() + rhs.size())
+                               )
+                         } yield Expr(GByteArray(lhs.concat(rhs)))
                        case (EListBody(lhs), EListBody(rhs)) =>
                          for {
                            _ <- costAccountingAlg.charge(PREPEND_COST * lhs.ps.length)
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -707,6 +707,12 @@ object Reduce {
                                  STRING_APPEND_COST * (lhs.length + rhs.length)
                                )
                          } yield Expr(GString(lhs + rhs))
+                       case (GByteArray(lhs), GByteArray(rhs)) =>
+                         for {
+                           _ <- costAccountingAlg.charge(
+                                 STRING_APPEND_COST * (lhs.size() + rhs.size())
```
goral09: I think that the cost for concatenating to byte arrays should what the cost for concatenating lists is, rather than strings. Please change to:
```scala
costAccountingAlg.charge(PREPEND_COST * lhs.size())
```
and we will adjust if this is not the case. In general, it is more consistent as byte array and list are both collections and `String` is not.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -710,7 +710,7 @@ object Reduce {
                        case (GByteArray(lhs), GByteArray(rhs)) =>
                          for {
                            _ <- costAccountingAlg.charge(
-                                 STRING_APPEND_COST * (lhs.size() + rhs.size())
+                                 PREPEND_COST * (lhs.size() + rhs.size())
```
goral09: Instead of `PREPEND_COST * (lhs.size() + rhs.size())`
just
`PREPEND_COST * lhs.size()`

cboscolo: I will make this change, but it seems wrong to me that the cost of concating two ByteArrays would not match the cost of concating two Strings.

goral09: In FP languages to concatenating two collections needs to traverse the first one and glue the 2nd one (which is `O(1)`).

PR #1609, opened 2018-10-03
RHOL-688, adds toUtf8Bytes to Rholang to turn UTF-8 strings into byte…

cboscolo:
… arrays

## Overview
This adds `toUtf8Bytes` to Rholang to turn UTF-8 strings into `GByteArray`. This is useful if you want to pass a string directly into one of the hash functions in Rholang.

https://rchain.atlassian.net/browse/RHOL-688

- [ ] You assigned one person to review this PR (How to do this?)

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -913,6 +913,26 @@ object Reduce {
         }
     }
 
+    private[this] val toUtf8Bytes: Method = new Method() {
+
+      override def apply(
+          p: Par,
+          args: Seq[Par]
+      )(implicit env: Env[Par], costAccountingAlg: CostAccountingAlg[M]): M[Par] =
+        if (args.nonEmpty) {
+          s.raiseError(ReduceError("Error: toUtf8Bytes does not take arguments"))
+        } else {
+          p.singleExpr() match {
+            case Some(Expr(GString(encoded))) =>
+              for {
+                _ <- costAccountingAlg.charge(hexToByteCost(encoded))
+              } yield Expr(GByteArray(ByteString.copyFrom(encoded.getBytes("UTF-8"))))
+            case _ =>
+              s.raiseError(ReduceError("Error: toUtf8Bytes can be called only on single strings."))
+          }
+        }
+    }
+
     private[this] def method(methodName: String, expectedArgsLength: Int, args: Seq[Par])(
         thunk: => M[Par]
     ): M[Par] =
@@ -1313,6 +1333,7 @@ object Reduce {
         "nth"         -> nth,
         "toByteArray" -> toByteArray,
         "hexToBytes"  -> hexToBytes,
+        "toUtf8Bytes" -> toUtf8Bytes,
         "union"       -> union,
         "diff"        -> diff,
         "add"         -> add,
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -873,7 +873,7 @@ object Reduce {
           args: Seq[Par]
       )(implicit env: Env[Par], costAccountingAlg: CostAccountingAlg[M]): M[Par] =
         if (args.nonEmpty) {
-          s.raiseError(ReduceError("Error: toByteArray does not take arguments"))
+          s.raiseError(MethodArgumentNumberMismatch("toByteArray", 0, args.length))
         } else {
           for {
             exprEvaled <- evalExpr(p)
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -913,6 +913,26 @@ object Reduce {
...
     }
 
+    private[this] val toUtf8Bytes: Method = new Method() {
+
+      override def apply(
+          p: Par,
+          args: Seq[Par]
+      )(implicit env: Env[Par], costAccountingAlg: CostAccountingAlg[M]): M[Par] =
+        if (args.nonEmpty) {
+          s.raiseError(ReduceError("Error: toUtf8Bytes does not take arguments"))
```
goral09: We have some predefined constructors for method errors. For wrong number of arguments you should use `MethodArgumentNumberMismatch`

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -913,6 +913,26 @@ object Reduce {
...
+        } else {
+          p.singleExpr() match {
+            case Some(Expr(GString(encoded))) =>
+              for {
+                _ <- costAccountingAlg.charge(hexToByteCost(encoded))
+              } yield Expr(GByteArray(ByteString.copyFrom(encoded.getBytes("UTF-8"))))
+            case _ =>
+              s.raiseError(ReduceError("Error: toUtf8Bytes can be called only on single strings."))
```
goral09: For these kind of message use `MethodNotDefined`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1268,6 +1268,89 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+        } yield space.store.toMap
+        Await.result(inspectTask.runAsync, 3.seconds)
+    }
+    result should be(HashMap.empty)
+    errorLog.readAndClearErrorVector should be(
+      Vector(ReduceError("Error: toUtf8Bytes can be called only on single strings."))
+    )
+  }
+
+  "eval of `toUtf8Bytes`" should "transform string to UTF-8 byte array (not the rholang term)" in {
```
goral09: You should change the order of your test cases, ie `"eval of …" should "…"` should go first and the two previous tests should go after. 

When you run the tests the `it` in
```scala
it should "…" { … }
```
will get replaced with the previous content of `"__prefix__" should "test name"` `__prefix__`. What YOU want is to have the output looking like:
```
eval of `toUtf8Bytes` should transform …
eval of `toUtf8Bytes` should return an error …
```
but your current implementation will look like:
```
eval of hexToBytes should return an error when `toUtf8Bytes` is called with arguments
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -913,6 +913,26 @@ object Reduce {
...
+
+      override def apply(
+          p: Par,
+          args: Seq[Par]
+      )(implicit env: Env[Par], costAccountingAlg: CostAccountingAlg[M]): M[Par] =
+        if (args.nonEmpty) {
+          s.raiseError(ReduceError("Error: toUtf8Bytes does not take arguments"))
+        } else {
+          p.singleExpr() match {
+            case Some(Expr(GString(encoded))) =>
```
goral09: This is really a nitpick but `encoded` is not a good name because it actually _not_ encoded :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
cboscolo: I patterned my code after the `hexToBytes` right above `toUtf8Bytes`.  Should I fix that to also use `MethodArgumentNumberMismatch`?

goral09: Yes please, probably the code above wasn't patched after those error classes were introduced. We need to change that.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
cboscolo: I am NOT a scala dev, so, I appologized for needing to ask basic questions, refactoring to use `MethodNotDefined`, can I do the following:
```
          p.singleExpr() match {
            case Some(Expr(GString(encoded))) =>
              for {
                _ <- costAccountingAlg.charge(hexToByteCost(encoded))
              } yield Expr(GByteArray(ByteString.copyFrom(encoded.getBytes("UTF-8"))))
            case Some(Expr(other)) =>
              s.raiseError(MethodNotDefined("toUtf8Bytes", other.typ))
          }
```

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
cboscolo: Ooops... major oversight...

PR #1610, opened 2018-10-03
Add cost accounting for methods we missed

goral09:
## Overview
Some of the methods weren't charged for, this PR adds them and adjust cost of few others.

### Notes

### DIFF
--- a/models/src/main/scala/coop/rchain/models/ParSet.scala
+++ b/models/src/main/scala/coop/rchain/models/ParSet.scala
@@ -36,13 +36,23 @@ object ParSet {
 
   def apply(
       ps: Seq[Par],
-      connectiveUsed: Boolean = false,
-      remainder: Option[Var] = None
+      remainder: Option[Var]
   ): ParSet = {
     val shs = SortedParHashSet(ps)
-    ParSet(shs, connectiveUsed, Coeval.delay(updateLocallyFree(shs)).memoize, remainder)
+    ParSet(
+      shs,
+      connectiveUsed(ps) || remainder.isDefined,
+      Coeval.delay(updateLocallyFree(shs)).memoize,
+      remainder
+    )
   }
 
-  def updateLocallyFree(ps: SortedParHashSet): BitSet =
+  def apply(ps: Seq[Par]): ParSet =
+    apply(ps, None)
+
+  private def connectiveUsed(seq: Seq[Par]): Boolean =
+    seq.exists(_.connectiveUsed)
+
+  private def updateLocallyFree(ps: SortedParHashSet): BitSet =
     ps.sortedPars.foldLeft(BitSet())((acc, p) => acc | p.locallyFree)
 }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -639,6 +639,8 @@ object Reduce {
                 Applicative[M].pure[(String, String)](keyString -> valueString)
               case (GString(keyString), GInt(valueInt)) =>
                 Applicative[M].pure[(String, String)](keyString -> valueInt.toString)
+              case (GString(keyString), GBool(valueBool)) =>
+                Applicative[M].pure[(String, String)](keyString -> valueBool.toString)
               // TODO: Add cases for other ground terms as well? Maybe it would be better
               // to implement cats.Show for all ground terms.
               case (_: GString, value) =>
@@ -688,7 +690,7 @@ object Reduce {
                                       .map(
                                         keyValuePairs => GString(interpolate(lhs, keyValuePairs))
                                       )
-                           _ <- costAccountingAlg.charge(LOOKUP_COST * lhs.length)
+                           _ <- costAccountingAlg.charge(LOOKUP_COST * (lhs.length + rhs.size))
                          } yield result
                        case (_: GString, other) =>
                          s.raiseError(OperatorExpectedError("%%", "Map", other.typ))
@@ -711,12 +713,12 @@ object Reduce {
                        case (GByteArray(lhs), GByteArray(rhs)) =>
                          for {
                            _ <- costAccountingAlg.charge(
-                                 PREPEND_COST * lhs.size()
+                                 PREPEND_COST * (lhs.size() + rhs.size())
                                )
                          } yield Expr(GByteArray(lhs.concat(rhs)))
                        case (EListBody(lhs), EListBody(rhs)) =>
                          for {
-                           _ <- costAccountingAlg.charge(PREPEND_COST * lhs.ps.length)
+                           _ <- costAccountingAlg.charge(PREPEND_COST * (lhs.ps.length + rhs.ps.length))
                          } yield
                            Expr(
                              EListBody(
@@ -953,7 +955,7 @@ object Reduce {
               ESetBody(base @ ParSet(basePs, _, _, _)),
               ESetBody(other @ ParSet(otherPs, _, _, _))
               ) =>
-            costAccountingAlg.charge(ADD_COST * basePs.size) *> Applicative[M].pure[Expr](
+            costAccountingAlg.charge(ADD_COST * otherPs.size) *> Applicative[M].pure[Expr](
               ESetBody(
                 ParSet(
                   basePs.union(otherPs.sortedPars.toSet),
@@ -967,7 +969,7 @@ object Reduce {
               EMapBody(base @ ParMap(baseMap, _, _, _)),
               EMapBody(other @ ParMap(otherMap, _, _, _))
               ) =>
-            costAccountingAlg.charge(ADD_COST * baseMap.size) *>
+            costAccountingAlg.charge(ADD_COST * otherMap.size) *>
               Applicative[M].pure[Expr](
                 EMapBody(
                   ParMap(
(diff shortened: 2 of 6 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -687,7 +689,7 @@ object Reduce {
                                       .map(
                                         keyValuePairs => GString(interpolate(lhs, keyValuePairs))
                                       )
-                           _ <- costAccountingAlg.charge(LOOKUP_COST * lhs.length)
+                           _ <- costAccountingAlg.charge(LOOKUP_COST * (lhs.length * rhs.size))
```
KentShikama: This type of change feels like it should have a test. Also the .length vs .size difference seems to stand out - there is no different right?

KentShikama: Also why is this multiplicative versus the prepending costs additive?

goral09: Oh, it should be additive. It was late, thanks for catching that.

goral09: Because there's no `length` on `Map`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
```diff
@@ -47,10 +47,14 @@ trait Costs {
   // + allocates byte array of the same size as `serializedSize`
   // + then it copies all elements of the Par
   def toByteArrayCost[T <: GeneratedMessage](a: T): Cost = Cost(a.serializedSize)
+  //TODO: adjust the cost of size method
+  def sizeMethodCost(size: Int): Cost = Cost(size % 100)
```
ArturGajowy: Why `% 100`? If we need a temporary implementation - why not just size?

goral09: Cost equal to size of the collection is definitely too much. It is temporary but I felt like that big disproportion between simple `size` method and other things is just too much.

ArturGajowy: I'd feel more comfortable if somebody else chimed in here. To me, this should be just `size`, and even `size  / 100` doesn't seem safe to me.

It might be safe, but under the assumption that other parts of the cost model account for this cost. In such a case, I'd rather not be doing any reasoning about the cost model, because it can't be reasoned about locally.

To me, each part that incurs a computational / other cost should account for the cost itself. This also means that the least common denominator of all costs is the cost of the cheapest operation, and if e.g. reading from rspace shold be O(size) and size should be O(size), but the latter with much lower constant - then the rspace read should have a bigger constant.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
```diff
@@ -53,6 +53,10 @@ trait Costs {
   def toByteArrayCost[T <: GeneratedMessage](a: T): Cost = Cost(a.serializedSize)
   //TODO: adjust the cost of size method
   def sizeMethodCost(size: Int): Cost = Cost(size)
+  def sliceCost(from: Int, to: Int): Cost = {
+    val diff = to - from
+    if (diff <= 0) Cost(0) else Cost(diff)
+  }
```
ArturGajowy: I think the cost of slicing should be `from + (to - from) == to`, becuase we probably first need to pass `from` elements to then read the `to - from` of the ones we're interested in.

Also, we should probably not charge 100 for [1,2,3].slice(2, 100) (the cost should be parameterized - and capped by - collection size).

Did we do any range checking for costs so far? If not, I'd add it here and file a JIRA for adding it in other places.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -955,7 +955,7 @@ object Reduce {
               ESetBody(base @ ParSet(basePs, _, _, _)),
               ESetBody(other @ ParSet(otherPs, _, _, _))
               ) =>
-            costAccountingAlg.charge(ADD_COST * otherPs.size) *> Applicative[M].pure[Expr](
+            costAccountingAlg.charge(unionCost(otherPs.size)) *> Applicative[M].pure[Expr](
```
ArturGajowy: :+1:

rholang/src/test/scala/coop/rchain/rholang/interpreter/SortSpec.scala
```diff
@@ -14,7 +14,7 @@ class SortSpec extends FlatSpec with Matchers {
   "GroundSortMatcher" should "discern sets with and without remainder" in {
     assertOrder[Expr](
       ParSet(Seq.empty),
-      ParSet(Seq.empty, remainder = Some(FreeVar(0)))
+      ParSet(Seq.empty, connectiveUsed = false, remainder = Some(FreeVar(0)))
```
ArturGajowy: why this change? It doesn't help with the tests readability.

models/src/main/scala/coop/rchain/models/ParSet.scala
```diff
@@ -36,13 +36,22 @@ object ParSet {
 
   def apply(
       ps: Seq[Par],
-      connectiveUsed: Boolean = false,
-      remainder: Option[Var] = None
+      connectiveUsed: Boolean,
+      remainder: Option[Var]
```
ArturGajowy: May I suggest making `remainder` the 2nd argument on the occasion? :) Up to you, since it's an additional refactoring.

rholang/src/test/scala/coop/rchain/rholang/interpreter/SortSpec.scala
ArturGajowy: I think you can do `ParSet(Seq.empty, remainder = Some(FreeVar(0)))`

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
goral09: > Also, we should probably not charge 100 for [1,2,3].slice(2, 100) (the cost should be parameterized - and capped by - collection size).

good call.

> Did we do any range checking for costs so far? If not, I'd add it here and file a JIRA for adding it in other places.

I am writing tests at the very moment.

models/src/main/scala/coop/rchain/models/ParSet.scala
goral09: It is 2nd argument.

models/src/main/scala/coop/rchain/models/ParSet.scala
```diff
@@ -36,13 +36,23 @@ object ParSet {
...
+      remainder
+    )
   }
 
-  def updateLocallyFree(ps: SortedParHashSet): BitSet =
+  def apply(ps: Seq[Par]): ParSet =
+    apply(ps, None)
+
+  private def connectiveUsed(seq: Seq[Par]): Boolean =
+    seq.exists(_.connectiveUsed)
```
ArturGajowy: We actually need the field, or we'll rely on our children's caches - which is a terrible thing for a parent to do

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
```diff
@@ -447,18 +446,15 @@ class VarMatcherSpec extends FlatSpec with Matchers with TimeLimits with TripleE
...
+      ParSet(Seq(GInt(2), GInt(5), EVar(FreeVar(0)), EVar(Wildcard(WildcardMsg()))))
     assertSpatialMatch(target, nonMatchingPattern, None)
   }
 
   it should "work with wildcard remainders" in {
     val targetElements = Seq[Par](GInt(1), GInt(2), GInt(3), GInt(4), GInt(5))
     val target: Expr   = ParSet(targetElements)
     val pattern: Expr =
-      ParSet(Seq(GInt(1), GInt(4)), connectiveUsed = true, remainder = Wildcard(WildcardMsg()))
+      ParSet(Seq(GInt(1), GInt(4)), remainder = Wildcard(WildcardMsg()))
```
ArturGajowy: ~I think this will no longer have `connectiveUsed` in your implementeation.~

It will. But we should add a test if no test of ParSet fails when you don't include remainder.

We should add a generative test of the form:
```scala
forAll { (a: Expr, b: Expr, r: Var) =>
   ParSet(Seq(a, b), remainer = r).connectiveUsed == Seq(a, b, r).conains(_.connectiveUsed) //assumin Var has connectiveUsed == true...
}
```
or sth like that

models/src/main/scala/coop/rchain/models/ParSet.scala
ArturGajowy: Also, we should probably make the `connectiveUsed` computation lazy... At this point I think we should maybe allow passing it from outside as it used to be?

goral09: > We actually need the field, or we'll rely on our children's caches - which is a terrible thing for a parent to do

what caches? I think this version is safer than it was before.

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
goral09: There is a test for that: https://github.com/rchain/rchain/blob/72c6dff860ee6147443feac60b5fddf8367d058f/rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala#L434

models/src/main/scala/coop/rchain/models/ParSet.scala
ArturGajowy: It is and I'd very much welcome the change. `connectiveUsed` exists on the tree so that it doesn't have to be re-evaluated each time we ask about it. Ohterwise we'd just use the HasLocallyFree typeclass.

PR #1614, opened 2018-10-04
Run all tests on both Java 8 and Java 10

adaszko:
* [JIRA ticket](https://rchain.atlassian.net/browse/CORE-1246)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,7 +26,7 @@ test:integration-test-for-p2p-network-java10:
   script:
     - java -version
     - ./scripts/install_bnfc.sh
-    - sudo sbt -Dsbt.log.noformat=true clean rholang/bnfc:generate casper/test:compile node/docker:publishLocal
+    - sudo DOCKER_BASE_IMAGE=openjdk:11-jre-slim sbt -Dsbt.log.noformat=true clean rholang/bnfc:generate casper/test:compile node/docker:publishLocal
     - sudo python3.6 -m pip install argparse docker pexpect requests
     - sudo ./scripts/p2p-test-tool.py -b -t -p 2 -m 2048m
   only:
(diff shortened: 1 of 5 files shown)


.gitlab-ci.yml
```diff
@@ -78,3 +134,34 @@ deploy:deploy-rnode-artifacts:
...
+    - artifacts_dir=$(mktemp -d /tmp/artifacts_out.XXXXXXXX)
+    - cp node/target/rnode_*_all.deb ${artifacts_dir}/
+    - cp node/target/rpm/RPMS/noarch/rnode-*.noarch.rpm ${artifacts_dir}/
+    - cp rosette/build.out/rosette-*.deb ${artifacts_dir}/
+    - cp node/target/universal/*.tgz ${artifacts_dir}/
+    # deploy artifacts to dispatcher
+    - mkdir -p ~/.ssh
+    - echo "${SSH_PRIVATE_KEY}" | tr -d '\r' > ~/.ssh/id_rsa
+    - chmod 600 ~/.ssh/id_rsa
+    - rsync -avrz --delete -e  "ssh -p 22 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" ${artifacts_dir}/* inbound@dispatch.pyr8.io:/home/inbound/cd/artifacts-java10/${CI_COMMIT_REF_NAME}/
```
adaszko: TODO Make sure this dir exists

build.sbt
```diff
@@ -187,7 +187,7 @@ lazy val node = (project in file("node"))
     /* Dockerization */
     dockerUsername := Some(organization.value),
     dockerUpdateLatest := true,
-    dockerBaseImage := "openjdk:8u171-jre-slim-stretch",
+    dockerBaseImage := "openjdk:11-jre-slim",
```
adaszko: This may be controversial.  The tests don't pass without it.

TomVasile:
> * Caveats: I'm not sure how Travis will handle uploading build artifacts at the `deploy` stage when there are 2 JDK versions in use.  Does it upload tarballs built with both JDK versions at https://github.com/rchain/rchain/releases/ ?

I would think the files would need to be named differently otherwise they would overwrite.

.gitlab-ci.yml
adaszko: Done

build.sbt
adaszko: This entails *always* using openjdk11 in `sbt pushLocal`

adaszko:
> I would think the files would need to be named differently otherwise they would overwrite.

@TomVasile That's my understanding too.  I've added renaming of those files.

.gitlab-ci.yml
ArturGajowy: we should deduplicate that into an `.sh` script and use it here and above. Similar with other duplication in tasks above. We aren't dropping java8 support anytime soon, it seems to me.

adaszko: Makes sense.  I though we were going to ditch 8.

ArturGajowy:
> Why Java 10 is called openjdk-11 in Ubuntu 18.04: lists.ubuntu.com/archives/ubuntu-release/2018-February/004275.html

Please put this into the .yml file itself, or at least in the relevant commit's message. Otherwise it's going to be lost, like tears in the rain...

adaszko:
Superseded by https://github.com/rchain/rchain/pull/1702

PR #1615, opened 2018-10-04
refactor reducer

odeac:
## Overview
This refactoring is part of my learning process in the Reducer code, as I learn  how things work I refactor them such that I can read them better. 

RHOL-897

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -159,7 +159,7 @@ object Reduce {
       ).scanLeft(0)(_ + _)
 
       def handle[A](
-          eval: (A => (Env[Par], Blake2b512Random, CostAccountingAlg[M]) => M[Unit]),
+          evalImpl: ((A, Blake2b512Random) => M[Unit]),
           start: Int
       )(input: (A, Int)): M[Unit] = {
         val (term, idx) = input
@@ -172,19 +172,36 @@ object Reduce {
           else
             rand.splitByte((start + idx).toByte)
 
-        eval(term)(env, newRand, costAccountingAlg).handleErrorWith {
-          case e: OutOfPhlogistonsError.type =>
-            s.raiseError(e)
-          case e =>
-            fTell.tell(e) *> s.unit
-        }
+        evalImpl(term, newRand)
       }
+
+      def mkTermHandler[A](
+          impl: (A => (Env[Par], Blake2b512Random, CostAccountingAlg[M]) => M[Unit])
+      )(term: A, rand: Blake2b512Random): M[Unit] =
+        impl(term)(env, rand, costAccountingAlg)
+          .handleErrorWith {
+            case e: OutOfPhlogistonsError.type =>
+              s.raiseError(e)
+            case e =>
+              fTell.tell(e) *> s.unit
+          }
+
+      def exprHandler(expr: Expr, rand: Blake2b512Random): M[Unit] =
+        expr.exprInstance match {
+          case EVarBody(EVar(v)) =>
+            (for {
+              varref <- eval(v)
+              _      <- eval(varref)(env, rand, costAccountingAlg)
+            } yield ()).handleError(fTell.tell)
+          case e: EMethodBody =>
+            (for {
+              p <- evalExprToPar(Expr(e))
+              _ <- eval(p)(env, rand, costAccountingAlg)
+            } yield ()).handleError(fTell.tell)
+          case _ => s.unit
+        }
+
       List(
-        Parallel.parTraverse(par.sends.zipWithIndex.toList)(handle(evalExplicit, starts(0))),
-        Parallel.parTraverse(par.receives.zipWithIndex.toList)(handle(evalExplicit, starts(1))),
-        Parallel.parTraverse(par.news.zipWithIndex.toList)(handle(evalExplicit, starts(2))),
-        Parallel.parTraverse(par.matches.zipWithIndex.toList)(handle(evalExplicit, starts(3))),
-        Parallel.parTraverse(par.bundles.zipWithIndex.toList)(handle(evalExplicit, starts(4))),
         Parallel.parTraverse(filteredExprs.zipWithIndex.toList)(texpr => {
           val (expr, idx) = texpr
           val newRand =
@@ -195,20 +212,13 @@ object Reduce {
             else
               rand.splitByte((starts(5) + idx).toByte)
 
-          expr.exprInstance match {
-            case EVarBody(EVar(v)) =>
-              (for {
-                varref <- eval(v)
-                _      <- eval(varref)(env, newRand, costAccountingAlg)
-              } yield ()).handleError(fTell.tell)
-            case e: EMethodBody =>
-              (for {
-                p <- evalExprToPar(Expr(e))
-                _ <- eval(p)(env, newRand, costAccountingAlg)
-              } yield ()).handleError(fTell.tell)
-            case _ => s.unit
-          }
+          exprHandler(expr, newRand)
         })
+        Parallel.parTraverse(par.sends.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(0))),
+        Parallel.parTraverse(par.receives.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(1))),
+        Parallel.parTraverse(par.news.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(2))),
+        Parallel.parTraverse(par.matches.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(3))),
+        Parallel.parTraverse(par.bundles.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(4))),
       ).parSequence.as(Unit)
     }
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -158,21 +158,13 @@ object Reduce {
         filteredExprs.size
       ).scanLeft(0)(_ + _)
 
-      def handle[A](
-          evalImpl: ((A, Blake2b512Random) => M[Unit])
-      )(input: (A, Int)): M[Unit] = {
-        val (term, idx) = input
-
-        val newRand =
-          if (starts(6) == 1)
-            rand
-          else if (starts(6) > 256)
-            rand.splitShort((starts(idx) + idx).toShort)
-          else
-            rand.splitByte((starts(idx) + idx).toByte)
-
-        evalImpl(term, newRand)
-      }
+      def split(rand: Blake2b512Random, idx: Int): Blake2b512Random =
+        if (starts(6) == 1)
+          rand
+        else if (starts(6) > 256)
+          rand.splitShort((starts(idx) + idx).toShort)
+        else
+          rand.splitByte((starts(idx) + idx).toByte)
 
       def mkTermHandler[A](
           impl: (A => (Env[Par], Blake2b512Random, CostAccountingAlg[M]) => M[Unit])
@@ -200,8 +192,10 @@ object Reduce {
           case _ => s.unit
         }
 
-      def mkJob[A](terms : Seq[A], handler : (A, Blake2b512Random) => M[Unit]) : M[List[Unit]] =
-        Parallel.parTraverse(terms.zipWithIndex.toList)(handle(handler))
+      def mkJob[A](terms: Seq[A], handler: (A, Blake2b512Random) => M[Unit]): M[List[Unit]] =
+        Parallel.parTraverse(terms.zipWithIndex.toList) {
+          case (term, idx) => handler(term, split(rand, idx))
+        }
 
       List(
         mkJob(par.sends, mkTermHandler[Send](evalExplicit)),
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -126,7 +126,7 @@ object Reduce {
       tuplespaceAlg.consume(binds, ParWithRandom(body, rand), persistent)
 
     private trait EvalJob {
-      def run(totalSize: Int, termSize: Int)(
+      def run(randSplitter: (Blake2b512Random, Int) => Blake2b512Random)(
           env: Env[Par],
           rand: Blake2b512Random,
           costAccountingAlg: CostAccountingAlg[M]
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -195,20 +212,13 @@ object Reduce {
...
-            case e: EMethodBody =>
-              (for {
-                p <- evalExprToPar(Expr(e))
-                _ <- eval(p)(env, newRand, costAccountingAlg)
-              } yield ()).handleError(fTell.tell)
-            case _ => s.unit
-          }
+          exprHandler(expr, newRand)
         })
+        Parallel.parTraverse(par.sends.zipWithIndex.toList)(handle(mkTermHandler(evalExplicit), starts(0))),
```
ArturGajowy: Please retain the original order of traverses this and the lines 217-221 should be above line 215

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -175,19 +175,6 @@ object Reduce {
         evalImpl(term, newRand)
       }
 
-      def handle2(start : Int)(texpr : (Expr, Int)) : M[Unit] = {
```
ArturGajowy: Ideally, all your commit messages would start with a capital letter, and there would be no commits undoing / refactoring code introduced in the same PR (refactors would be applied in-place, in the commit introducing the refactored code). If you can do a rebase dropping this and the previous commit, I'll appreciate one :) Otherwise that's just a general remark for the future (mostly the commit message part).

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -159,18 +159,17 @@ object Reduce {
...
 
         val newRand =
           if (starts(6) == 1)
             rand
           else if (starts(6) > 256)
-            rand.splitShort((start + idx).toShort)
+            rand.splitShort((starts(idx) + idx).toShort)
           else
-            rand.splitByte((start + idx).toByte)
+            rand.splitByte((starts(idx) + idx).toByte)
```
ArturGajowy: I'm not sure why this change is correct. Can you explain?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -200,8 +192,10 @@ object Reduce {
           case _ => s.unit
         }
 
-      def mkJob[A](terms : Seq[A], handler : (A, Blake2b512Random) => M[Unit]) : M[List[Unit]] =
-        Parallel.parTraverse(terms.zipWithIndex.toList)(handle(handler))
+      def mkJob[A](terms: Seq[A], handler: (A, Blake2b512Random) => M[Unit]): M[List[Unit]] =
+        Parallel.parTraverse(terms.zipWithIndex.toList) {
+          case (term, idx) => handler(term, split(rand, idx))
```
ArturGajowy: can we make it a total funciton using `Function.tupled`?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -192,18 +192,22 @@ object Reduce {
           case _ => s.unit
         }
 
-      def mkJob[A](terms: Seq[A], handler: (A, Blake2b512Random) => M[Unit]): M[List[Unit]] =
+      def mkJob[A](
+          terms: Seq[A],
+          handler: (A, Blake2b512Random) => M[Unit],
+          startIdx: Int
```
ArturGajowy: How about making `startIdx` the first parameter, for some nice alignment and faster-spotted copy-paste errors?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -158,13 +158,13 @@ object Reduce {
...
-      def split(rand: Blake2b512Random, idx: Int): Blake2b512Random =
+      def split(rand: Blake2b512Random, startIdx: Int, idx: Int): Blake2b512Random =
         if (starts(6) == 1)
           rand
         else if (starts(6) > 256)
-          rand.splitShort((starts(idx) + idx).toShort)
+          rand.splitShort((starts(startIdx) + idx).toShort)
         else
-          rand.splitByte((starts(idx) + idx).toByte)
+          rand.splitByte((starts(startIdx) + idx).toByte)
```
ArturGajowy: turns out that change wasn't correct :D Again, Ideally you'd just fixup! + rebase

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -192,23 +217,17 @@ object Reduce {
...
-        mkJob(par.bundles, mkTermHandler[Bundle](evalExplicit), 4),
-        mkJob(filteredExprs, exprHandler, 5)
-      ).parSequence.as(Unit)
+        EvalJob(par.sends, mkTermHandler[Send](evalExplicit)),
+        EvalJob(par.receives, mkTermHandler[Receive](evalExplicit)),
+        EvalJob(par.news, mkTermHandler[New](evalExplicit)),
+        EvalJob(par.matches, mkTermHandler[Match](evalExplicit)),
+        EvalJob(par.bundles, mkTermHandler[Bundle](evalExplicit)),
+        EvalJob(filteredExprs, exprHandler)
+      ).zipWithIndex
```
ArturGajowy: Honeslty, I don't see any benefit from introducing the EvalJob. If you wanted to get rid of the hardcoded ints, I think you could've done:

```scala
        mkJob(par.sends, mkTermHandler[Send](evalExplicit), _),
        mkJob(par.receives, mkTermHandler[Receive](evalExplicit), _),
        mkJob(par.news, mkTermHandler[New](evalExplicit), _),
        mkJob(par.matches, mkTermHandler[Match](evalExplicit), _),
        mkJob(par.bundles, mkTermHandler[Bundle](evalExplicit), _),
        mkJob(filteredExprs, exprHandler, _)
      )
.zipWithIndex
.map(i, f -> f(i))
.parSequence.as(Unit)

```

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -217,14 +211,18 @@ object Reduce {
...
         EvalJob(par.sends, mkTermHandler[Send](evalExplicit)),
         EvalJob(par.receives, mkTermHandler[Receive](evalExplicit)),
         EvalJob(par.news, mkTermHandler[New](evalExplicit)),
         EvalJob(par.matches, mkTermHandler[Match](evalExplicit)),
         EvalJob(par.bundles, mkTermHandler[Bundle](evalExplicit)),
         EvalJob(filteredExprs, exprHandler)
-      ).zipWithIndex
+      )
+
+      val starts = jobs.map(_.size).scanLeft(0)(_ + _).toVector
```
ArturGajowy: You've just convinced me the EvalJob has merit, due to it's size property :joy:

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: I don't understand what you mean here

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: Here the intention was to make small steps such that each of them are obviously correct. This is why I introduced this function first and then it was obvious that only the exprHandler was different compared to the other one so I parameterized the difference. I think these small steps are a very good way to help the reviewer understand the refactoring process.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: It's not correct. I committed this without running the tests and I only realized it after I did the next step. Then I didn't undo because the fix was easy. See the commit "fix the tests"

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: This is function was abandoned during the refactoring but thanks for telling me about Function.tupled.  I didn't know about its existence.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: for some reason I got some rebase conflicts so it was easier to just add a new commit

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: outdated

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: Good, because I was preparing the response for your previous comment :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
goral09: This doesn't really matter because we traverse it in parallel.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -125,6 +125,89 @@ object Reduce {
...
+          rand.splitShort((starts(startIdx) + idx).toShort)
+        else
+          rand.splitByte((starts(startIdx) + idx).toByte)
+
+      private def mkJob[A](
+          input: Seq[A],
+          handler: A => (Env[Par], Blake2b512Random, CostAccountingAlg[M]) => M[Unit]
+      ): EvalJob =
+        new EvalJob() {
+          override def run(starts: Vector[Int], startIdx: Int)(
```
goral09: Previous implementation, when starting the evaluation of certain kind of terms didn't need to know about all start indexes (`starts`). Single `run` call on specific type of terms should only know about it's own index. It means that splitting needs to be done outside of `run`. Your `split` methods splits at `starts(startIdx) + idx` so it can be improved even more.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -185,39 +224,13 @@ object Reduce {
...
-        }
-
       val jobs = List(
-        EvalJob(par.sends, mkTermHandler[Send](evalExplicit)),
-        EvalJob(par.receives, mkTermHandler[Receive](evalExplicit)),
-        EvalJob(par.news, mkTermHandler[New](evalExplicit)),
-        EvalJob(par.matches, mkTermHandler[Match](evalExplicit)),
-        EvalJob(par.bundles, mkTermHandler[Bundle](evalExplicit)),
-        EvalJob(filteredExprs, exprHandler)
+        EvalJob[Send](par.sends, evalExplicit),
```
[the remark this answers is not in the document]
odeac: Tell that to the Scala compiler. Without putting the type explicitly, it doesn't compile

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: If you compare to the original implementation, the "rand splitting code" was referring start indices. Indeed, it might be a good idea to split the rngs upfront and zip them with the jobs

odeac:
@goral09 By this refactoring a few things are achieved:

1. There are (almost) no magic numbers left in the code. Those are error prone and make it difficult to understand the intention of the code
2. The possibility of having subtle bugs because the following are not in sync: position in par, start index, position in the list of parTraverse's. Such connections make the code extremely brittle. 
3. It becomes obvious how the evaluation of expressions is different than the evaluation of the other terms. 
4. The rand splitting code is in one single function reused twice. This shows that we only have one rand splitting mechanism. This is important if we want to understand how the 
5. There is no more code duplication and the fact that the code is the same size it's just because those verbose signatures

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -244,7 +239,12 @@ object Reduce {
       val starts = jobs.map(_.size).scanLeft(0)(_ + _).toVector
 
       jobs.zipWithIndex
-        .map { case (job, idx) => job.run(starts, idx)(env, rand, costAccountingAlg) }
+        .map {
+          case (job, idx) => {
+            val splitter = split(starts.last, starts(idx))(_, _)
```
goral09: If you already know `rand` here, why not make `splitter` of type `Int => Blake2b512Random` ?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: Yes, and another advantage of making those small steps is you don't even have to show things that don't make it to the final shape, like these two commits: one introducing handle2 and the other one removing it. How you avoid showing such steps - and achieve the *ultimate* ease for the reviewer (less material to process) - is by doing a rebase.

Unless you mean that showing the introduction and retraction itself would help me understand sth. All I understood from those two is you changed your mind in the process.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: Yeah, as I mentioned - that commit could've easily be a `fixup! [Original commit message]`, squashed automatically upon rebase (provided you have `git config --global rebase.autosquash true`)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: good point

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -125,6 +125,39 @@ object Reduce {
...
+      ): Blake2b512Random =
+        if (starts(6) == 1)
+          rand
+        else if (starts(6) > 256)
+          rand.splitShort((starts(startIdx) + idx).toShort)
+        else
+          rand.splitByte((starts(startIdx) + idx).toByte)
+
+      def apply[A](input: Seq[A], handler: (A, Blake2b512Random) => M[Unit]): EvalJob =
+        new EvalJob() {
```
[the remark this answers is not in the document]
odeac: ok

PR #1617, opened 2018-10-04
Rholang: Initialize the registry for RholangCLI and the REPL

guardbotmk3:
## Overview
This PR makes the registry available to the REPL and RholangCLI. It also introduces a mechanism that should allow casper to initialize the registry at genesis.

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
@@ -6,6 +6,7 @@ import cats.Id
 import cats.mtl.FunctorTell
 import cats.effect.Sync
 import com.google.protobuf.ByteString
+import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Expr.ExprInstance.GString
 import coop.rchain.models.TaggedContinuation.TaggedCont.ScalaBodyRef
 import coop.rchain.models.Var.VarInstance.FreeVar
@@ -13,6 +14,7 @@ import coop.rchain.models._
 import coop.rchain.models.rholang.implicits._
 import coop.rchain.rholang.interpreter.Runtime._
 import coop.rchain.rholang.interpreter.accounting.Cost
+import coop.rchain.rholang.interpreter.errors.SetupError
 import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 import coop.rchain.rholang.interpreter.storage.implicits._
 import coop.rchain.rspace.IReplaySpace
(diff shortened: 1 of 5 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -41,6 +43,38 @@ class Runtime private (
     replaySpace.close()
     context.close()
   }
+  def injectEmptyRegistryRoot[F[_]]()(implicit F: Sync[F]): F[Unit] = {
```
birchmd: Empty parentheses are only needed to indicate a method with no arguments has side effects. I think you can drop the parentheses and wrap the `produce` calls in `Sync[F].delay` in order to leave this as a pure method.
```scala
for {
  spaceResult <- F.delay(space.produce(...))
  replayResult <- F.delay(replaySpace.produce(...))
  _ <- spaceResult match { ... }
} yield ()
```

goral09: Isn't this going to duplicate the work done in #1586 @birchmd ?

birchmd: I'll let your PR go though first and then update mine to call `injectEmptyRegistryRoot`

guardbotmk3: I've wrapped the side effects in F.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -233,24 +276,25 @@ object Runtime {
...
+        ED25519_VERIFY                      -> SystemProcesses.ed25519Verify(space, dispatcher),
+        SHA256_HASH                         -> SystemProcesses.sha256Hash(space, dispatcher),
+        KECCAK256_HASH                      -> SystemProcesses.keccak256Hash(space, dispatcher),
+        BLAKE2B256_HASH                     -> SystemProcesses.blake2b256Hash(space, dispatcher),
+        SECP256K1_VERIFY                    -> SystemProcesses.secp256k1Verify(space, dispatcher),
+        REG_LOOKUP                          -> (registry.lookup(_)),
+        REG_LOOKUP_CALLBACK                 -> (registry.lookupCallback(_)),
+        REG_INSERT                          -> (registry.insert(_)),
+        REG_INSERT_CALLBACK                 -> (registry.insertCallback(_)),
+        REG_PUBLIC_REGISTER_INSERT_CALLBACK -> (registry.publicRegisterInsertCallback(_)),
```
goral09: Can you remind me what `publicRegisterInsertCallback` is doing?

PR #1622, opened 2018-10-05
Make starting REV amount in HashSetCasperTest constant

KentShikama:
This resolves non-deterministic failures due to low funds

Resolves https://rchain.atlassian.net/browse/RHOL-929


birchmd:
Why did this lead to an invalid block instead of just failing to bond?

goral09:
@birchmd insufficent funds in the deploy will mark block as errored right?

birchmd:
No, this isn't about running out of phlo (we don't charge wallets yet). It would be a totally valid block which just did not have the intended effect.

But I see the problem now. I have the new validator try to propose a block and that would be invalid because they have zero weight.

goral09:
> No, this isn't about running out of phlo (we don't charge wallets yet). It would be a totally valid block which just did not have the intended effect.
> 

we don't charge wallets but if the node proposes the block and it runs out of phlos you still mark the block as errored. It's probably irrelevant to the issue but I want to make this clear because I think that we might talk past each other.

PR #1623, opened 2018-10-05
Make CompiledRholangSource be compiled at runtime

KentShikama:
This removes a lot of unnecessary issues related to building; Kayvan had an issue with this too. We can revert once the codebase becomes more stable.


ArturGajowy:
Merging this as this fixes dev build as far as I can tell (it helped on my machine)

PR #1626, opened 2018-10-05
RHOL-862: Prohibit interception of standard output and error

tymm:
## Overview
Prohibits interception of `stdout`, `stdoutAck`, `stderr` and `stderrAck` by making them write-only bundles.
Stole test functions from `InterpreterSpec` and created a `RuntimeSpec`.

https://rchain.atlassian.net/browse/RHOL-862

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
@@ -14,35 +14,37 @@ class RuntimeSpec extends FlatSpec with Matchers {
 
   val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
 
-  val reduceError = "ReduceError: Trying to read from non-readable channel."
+  val channelReadOnlyError = "ReduceError: Trying to read from non-readable channel."
 
   "rho:io:stdout" should "not get intercepted" in {
-    val intercept =
-      """new s(`rho:io:stdout`) in { for(x <- s) { Nil } }"""
-
-    failure(intercept).getMessage.stripLineEnd should endWith(reduceError)
+    checkError("""new s(`rho:io:stdout`) in { for(x <- s) { Nil } }""", channelReadOnlyError)
   }
 
   "rho:io:stdoutAck" should "not get intercepted" in {
-    val intercept =
-      """new s(`rho:io:stdoutAck`) in { for(x <- s) { Nil } }"""
-
-    failure(intercept).getMessage.stripLineEnd should endWith(reduceError)
+    checkError("""new s(`rho:io:stdoutAck`) in { for(x <- s) { Nil } }""", channelReadOnlyError)
   }
 
   "rho:io:stderr" should "not get intercepted" in {
-    val intercept =
-      """new s(`rho:io:stderr`) in { for(x <- s) { Nil } }"""
-
-    failure(intercept).getMessage.stripLineEnd should endWith(reduceError)
+    checkError("""new s(`rho:io:stderr`) in { for(x <- s) { Nil } }""", channelReadOnlyError)
   }
 
   "rho:io:stderrAck" should "not get intercepted" in {
-    val intercept =
-      """new s(`rho:io:stderrAck`) in { for(x <- s) { Nil } }"""
-
-    failure(intercept).getMessage.stripLineEnd should endWith(reduceError)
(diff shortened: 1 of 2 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+      "ReduceError: Trying to read from non-readable channel."
+    )
+  }
+
+  "rho:io:stderrAck" should "not get intercepted" in {
+    val intercept =
+      """new ackCh, stderrAck(`rho:io:stderrAck`) in { stderrAck!("foo") | for (@x <- stderrAck) { stderrAck!(x, *ackCh) } }"""
+
+    failure(intercept).getMessage.stripLineEnd should endWith(
+      "ReduceError: Trying to read from non-readable channel."
```
goral09: Refactor that message to a variable, it repeats 4 times in the test.

tymm:
@goral09 I think the same about having an AST as the input to the tests. However for that it would be nice to have `Interpreter.normalizeTerm` to be public. Can I make it public?

goral09:
@tymm you would actually need a Reducer. I was thinking about having already a normalized term.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
```diff
@@ -0,0 +1,380 @@
...
+  val tmpPrefix   = "rspace-store-"
+  val maxDuration = 5.seconds
+
+  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
+
+  val nonReadableErr = ReduceError("Trying to read from non-readable channel.")
+
+  "rho:io:stdout" should "not get intercepted" in {
+    // new s(`rho:io:stdout`) in { for (x <- s) { Nil } }
+    val interceptNormalized = Par(
```
ArturGajowy: I am very much against using the AST directly, unless it affects the tests potency in securing the change. From the conversation and the test itself I don't see why using AST here would be necessary.

You mention:
> I think the same about having an AST as the input to the tests. However for that it would be nice to have Interpreter.normalizeTerm to be public. Can I make it public?

There is `Interpreter.buildNormlizedTerm`, can we use that?

@goral09 mentions a reducer would be needed, but:
1) I'm not sure why, if the error is produced during normalizaton
2) Even if not, let's just use the reducer. You can have a look at `IntepreterSpec` if you need an e2e-ish approach.

goral09: The error is thrown during _reduce_ phase so it makes sense to test that and not lexing, parsing, normalizing and only then the logic.

goral09: The error is thrown during _reduce_ phase so it makes sense to test that and not lexing, parsing, normalizing and only then the logic.

ArturGajowy: >  it makes sense to test that and not lexing, parsing, normalizing and only then the logic

It's not *testing* the lexing, parsing and normalizing. It's just using them in the test's setup. Any error in them is just a test setup error, not a test failure in the strict sense (not an assertion error). Notice the red vs yellow failure icon in IDEA, for example.

The test input data is hardly parseable to the bare eye, and comments are often outdated even at the time they're being commited.

I stand by my claim that we should `Interpreter.buildNormalizedTerm` in setup, and then run the reducer.

tymm:
I changed it back to end-to-end test cases.

@ArturGajowy, @goral09, @guardbotmk3 should we make `rho:registry:lookup` and `rho:registry:insertArbitrary` write-only as well?

ArturGajowy:
@tymm Thanks, this reads much better.

> should we make rho:registry:lookup and rho:registry:insertArbitrary write-only as well

I think yes, please do.

One final request from me: please extract a `checkError(rho, error)` method error
so that each test can be expressed as a single method call, e.g.

```scala
checkError("""new s(`rho:io:stdout`) in { for(x <- s) { Nil } }""", channelReadOnlyError)
```

Please also rename `reduceError` to `channelReadOnlyError` or similar, so that the name does not require looking up the definition.

I'm approving the PR in advance :)

PR #1627, opened 2018-10-05
CORE-597 Series of TrieUpdates in a single Txn

ys-pyrofex:
## Overview
Performance optimization

https://rchain.atlassian.net/browse/CORE-597

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/IStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/IStore.scala
@@ -109,7 +109,7 @@ trait IStore[C, P, A, K] {
   private[rspace] def getTrieUpdateCount: Long =
     _trieUpdateCount.get()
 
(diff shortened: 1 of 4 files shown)


dzajkowski:
@ys-pyrofex So as far as I understand the discussion that happened on #rspace this change does not accomplish much. Is there any point/gain in merging it?

ys-pyrofex:
I think it's good idea to merge this commit: (1) while it doesn't reduces amount of operations, it reduces count of opened and closed transactions from replay log size to only 1 and (2) we may finish with the total amount of operations in tree merging algo will be similar or even bigger then with current algo and we will decide to not include it.

lukasz-golebiewski:
But on the other hand it's better to have more fine grained transactions than less bigger ones. I don't think this makes sense. Can you show some benchmark result that this indeed makes things better?

ys-pyrofex:
Replaced during further work on createCheckpoint algorithm.

PR #1628, opened 2018-10-06
Refund for clearing the tuplespace.

goral09:
## Overview
Fixes refunds so they are deterministic and correct.

https://rchain.atlassian.net/browse/RHOL-888

### Notes


rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
```diff
@@ -88,5 +88,8 @@ trait Costs {
     def storageCost: Cost = Cost(a.map(a => a.serializedSize).sum)
   }
 
-  final val MAX_VALUE = PhloLimit(Long.MaxValue)
+  // Even though we use Long as phlo limit we pass in less than maximum value
+  // Because when we pass Long.MaxValue and we refund the deploy for clearing the tuplespace
+  // phlo limit overflows causing the deploy to throw OutOfPhlogistonError.
```
ArturGajowy: Seems like a missing `toIntExact` at the place where it overflows. We should fix that.

ArturGajowy: Also, this just begs for bugs. I'd either make phloLimit an Interger, or change casper code to use Longs.

goral09: This doesn't solve the problem. If casper used  integers it would be the same. Not to mention that phlo limit has to be Long. In normal cases this won't ever be the problem because the deployer would need to have equivalent of `Long.MaxValue` REVs in the wallet and get refunded.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -56,6 +54,7 @@ object SystemProcesses {
             ListParWithRandom(Seq(Par.defaultInstance), rand),
             false
           )(MATCH_UNLIMITED_PHLOS)
+          .map(unpackOption(_))
```
lukasz-golebiewski: You can omit `(_)`

goral09: it didn't work. At least Intellij was complaining.

rspace/src/main/scala/coop/rchain/rspace/ISpace.scala
```diff
@@ -5,7 +5,13 @@ import coop.rchain.rspace.internal._
 
 import scala.collection.immutable.Seq
 
-case class Result[R](value: R, persistent: Boolean)
+final case class Result[R](value: R, persistent: Boolean)
```
lukasz-golebiewski: Looks like this isn't used anywhere anymore. In that case why don't you remove this and rename `ContResult` to `Result`?

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
ArturGajowy: I don't understand the problem then from how you described it in the comment.

When do we pass Long.MaxValue?
Why does a refund cause an overflow (I thought we only refund what has been spent, not more?)

BTW 'clearing the tuplespace' suggests clearing everything in the tuplespace. Is that what you tried to convey?

goral09: I've reworded the comment. Is this better?

rholang/src/test/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpaceTest.scala
```diff
@@ -60,16 +61,14 @@ class ChargingRSpaceTest extends fixture.FlatSpec with TripleEqualsSupport with
...
     val test = for {
+      _                 <- costAlg.set(CostAccount(0, minimumPhlos))
       _                 <- chargingRSpace.produce(channels.head, data, false)
       phlosAfterProduce <- costAlg.get()
       _                 = phlosAfterProduce.cost shouldBe (minimumPhlos - produceStorageCost)
       res               <- chargingRSpace.consume(channels, patterns, cont, false)
       phlosLeft         <- costAlg.get()
-      // we expect Cost(15), because we will be refunded for storing the consume
-      _ = phlosLeft.cost shouldBe (consumeStorageCost)
+      _                 = phlosLeft.cost shouldBe (consumeStorageCost + produceStorageCost)
```
ArturGajowy: why did the assertion change?

ArturGajowy: I see - we previously didn't refund the produceStorageCost

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
goral09: Yes, the failure was in the test but tests delegate setting initial phlo limit to this value.

goral09: > Either the refund is done by production code, in which case an overflow should mean a legitimate test failure or a bad setup

How will you differentiate between `Cost(-1)`, `Cost(-100)` or `Cost(Long.MinValue)` ? All are valid values but one is less probable than others.

PR #1629, opened 2018-10-06
RHOL-558

pyrocto:
## Overview
Modifies the casper rholang files to not use forgeable names.

https://rchain.atlassian.net/browse/RHOL-558

### Notes
The existing registry APIs do not support what's needed for this PR without major restructuring.  This PR proposes a stopgap solution until the package API for the registry is implemented.  It is incomplete; the StopGap contract needs to be installed in a way similar to the registry.  I'd appreciate advice on how to do that.


pyrocto:
I recommend reviewing without whitespace differences:
https://github.com/rchain/rchain/pull/1629/files?w=1

casper/src/main/rholang/StopGap.rho
```diff
@@ -0,0 +1,26 @@
+new builtinsStore, activeStore in {
+  builtinsStore!({}) |
+  activeStore!(true) |
+  contract @(*StopGap, "set")(@title, @value, ack) = {
```
goral09: First, where does the `StopGap` name come frome? I don't see it created/passed anywhere.

----------
I don't quite get how this will work. 
```
new StopGap(`rho:stopGap`) in { … } 
```
will return unforgeable name that  will reference `StopGap` contract. We can then use contracts that are leaving under this name to register new common contracts (like `("ListOps", "fold")` ) and use them if needed. The `activeStore` flag indicates wheter the contract map living under `StopGap` name can be mutated (i.e. if new contracts can be added) and anyone who has access to the name. It  means that anyone can "close" the map and anyone can  replace the contract leaving under the name. What am I missing here?

pyrocto: Ugh, I missed the notification from this due to a bad filter.  I wrote you on discord, but it looks like you missed that.

Yeah, StopGap.rho isn't quite right.  In fact, I just realized that the `(*name, "method")` idiom can't be made secure if you expose `name`.  I'll have to redo `StopGap.rho` and change `MakeMint.int2NN` to use a different pattern.

The intent was that code in the genesis block could install these contracts in a way that they could be looked up by name, but nothing running after the genesis block could install contracts this way.  The idea was that code in StopGap.rho would be moved into scala to provide the "injection" syntax.  I also wanted to make it so that each piece of code could run independently, since that's how it currently works.

The alternative you suggested is to install each process under its own public key.  However, the particular name that is generated in the `new` will depend on the timestamp of the deployment, so I can't just write rholang code.  I would need to figure out how the code is deployed in the genesis block, then write scala code to get the timestamp for the deployment, store all the private keys somewhere to generate the signatures, and generate the names that would be used, then inject all that stuff into the genesis block somehow.  This is much more complicated than being able to use the package API we discussed.  Is there a ticket for writing up that part of the registry?

casper/src/main/rholang/MakeMint.rho
```diff
@@ -29,7 +29,7 @@ new int2NN, MakeMint, NonNegativeNumber, sgset(`rho:stopgap:set`),
         contract internalMakePurse(@balance, return) = {
           //balance must be a name that NonNegativeNumber contract is listening on
           new thisPurse in {
-            contract thisPurse(=*decr, @amount, success) = { @(balance, "sub")!(amount, *success) } |
+            contract thisPurse(@=*decr, @amount, success) = { @(balance, "sub")!(amount, *success) } |
```
goral09: Are you taking advantage of the fact that `contract` is syntactic sugar for non-linear `for`s and just put a pattern here?

guardbotmk3:
I don't see anywhere where you've described how the stopgap isn't going to continue on.

casper/src/main/rholang/MakeMint.rho
pyrocto: Yes.  The security of the purse relies on a purse of a different currency being unable to construct this first argument.  Which means I have to add a bunch of bundle+ wrappers.  BRB!

casper/src/main/rholang/BasicWallet.rho
```diff
@@ -1,7 +1,7 @@
 //scalapackage coop.rchain.rholang.wallet
 
 //assumes standard purse API from MakeMint
-contract @"BasicWallet"(@purse, @algorithm, @pk, return) = {
+contract BasicWallet(@purse, @algorithm, @pk, return) = {
```
goral09: Don't you have to declare `new BasicWallet` first?

pyrocto:
Obsolete due to changes made by @birchmd .

PR #1631, opened 2018-10-07
Use main thread pool node-runner as CPU-bound worker

sebtomba:
## Overview
This is a follow-up PR for https://github.com/rchain/rchain/pull/1595
The previous PR didn't pass the integration tests on Gitlab. This PR takes a further step in optimizing the thread pools.

https://rchain.atlassian.net/browse/CORE-1230

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
@@ -211,7 +211,7 @@ class TcpTransportLayer(host: String, port: Int, cert: String, key: String, maxM
     Task.delay {
       new TcpServerObservable(port, serverSslContext, maxMessageSize)
         .mapParallelUnordered(parallelism)(dispatchInternal)
-        .subscribe()(Scheduler.computation(parallelism, "tl-dispatcher"))
+        .subscribe()(Scheduler.fixedPool("tl-dispatcher", parallelism))
     }
   }
 
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -27,14 +27,18 @@ object Main {
 
   def main(args: Array[String]): Unit = {
 
+    implicit val scheduler: Scheduler = Scheduler.computation(
+      Math.max(java.lang.Runtime.getRuntime.availableProcessors(), 2),
+      "node-runner"
+    )
+
     val exec: Task[Unit] =
       for {
-        conf      <- Configuration(args)
-        scheduler = Scheduler.fixedPool("node-io", conf.server.threadPoolSize)
-        _         <- Task.defer(mainProgram(conf)(scheduler)).executeOn(scheduler)
+        conf <- Configuration(args)
+        _    <- Task.defer(mainProgram(conf))
       } yield ()
 
-    exec.unsafeRunSync(Scheduler.singleThread("main-io"))
+    exec.unsafeRunSync
   }
 
   private def mainProgram(conf: Configuration)(implicit scheduler: Scheduler): Task[Unit] = {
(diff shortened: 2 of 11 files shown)


comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -211,7 +211,7 @@ class TcpTransportLayer(host: String, port: Int, cert: String, key: String, maxM
     Task.delay {
       new TcpServerObservable(port, serverSslContext, maxMessageSize)
         .mapParallelUnordered(parallelism)(dispatchInternal)
-        .subscribe()(Scheduler.computation(parallelism, "tl-dispatcher"))
+        .subscribe()(Scheduler.fixedPool("tl-dispatcher", parallelism))
```
goral09: Why did you choose `fixedPool` ? Isn't the TCP layer non-blocking? If so then why you didn't choose `Scheduler.io` ? Why not _cached_ thread pool?

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -27,14 +27,18 @@ object Main {
 
   def main(args: Array[String]): Unit = {
 
+    implicit val scheduler: Scheduler = Scheduler.computation(
```
goral09: Why `computation`? It uses `ForkJoinPool` internally, is `mainProgram` of this _fork-join_ characteristic?

node/src/main/scala/coop/rchain/node/api/grpc.scala
```diff
@@ -35,23 +40,28 @@ object GrpcServer {
     Task.delay {
       NettyServerBuilder
         .forPort(port)
-        .executor(scheduler)
+        .executor(grpcExecutor)
```
goral09: What is this `executor` used for? You set manually the `grpcExecutor` betlow for `ReplGrpcMonix` and `DiagnosticsGprcMonix` manually so for which calls is, this `executor` call, setting the scheduler?

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -251,7 +251,7 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
       opt[Int](descr = "Maximum size of message that can be sent via transport layer")
 
     val threadPoolSize =
-      opt[Int](descr = "Maximum number of threads used by rnode")
+      opt[Int](descr = "Maximum number of threads used by rnode", hidden = true)
```
goral09: If `hidden = true` hides this flag from the help output why do we need this option in the first place?

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
sebtomba: First of all, this is the thread pool for consuming the message queue and dispatching the messages to Casper and Connect. There shouldn't be any IO here and I expect the work to be CPU-bound. On the other hand, a `Scheduler.compution` thread pool is backed by a `ForkJoinPool` which potentially creates and removes threads. We need a fixed size thread pool with `Math.max(Runtime.getRuntime.availableProcessors(), 2)` thread. Maybe we can at some point in future use the global `node-runner` thread pool.

node/src/main/scala/coop/rchain/node/api/grpc.scala
sebtomba: This is `grpc-java` specific. The executor is used internally for incoming requests. If we don't set here an `executor` then `grpc-java` will create its own `ForkJoinPool`. We want to use the same thread pool for all gRPC IO in `ReplGrpcMonix`, `DiagnosticsGprcMonix`, `DeployGrpcService` and `TcpTransportLayer`.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
sebtomba: I want to remove this flag in the future. But ATM I'm just hiding it to not break some existing scripts. This flag is like _depricated_.

PR #1632, opened 2018-10-08
Stop traversing DAG after passing block number of LFB

KentShikama:
This is an optimization to speed up buildScoreMap in the estimator. If you're adding the weight of a validator down a chain and you reach a block that has a block height that is less than the block height of the last finalized block, if you continue adding the weight of the validator down the chain, you will never add more score to any block that is the child of the last finalized block. Hence, at this point you can stop building the score map for this validator and move onto the next.


casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -69,17 +69,25 @@ object Estimator {
...
 
   // TODO: Fix to stop at genesis/LFB
-  def buildScoresMap[F[_]: Monad: BlockStore](blockDag: BlockDag): F[Map[BlockHash, Long]] = {
-    def hashParents(hash: BlockHash): F[List[BlockHash]] =
-      blockDag.dataLookup(hash).parents.pure[F]
+  def buildScoresMap[F[_]: Monad: BlockStore](
+      blockDag: BlockDag,
+      genesis: BlockMessage
+  ): F[Map[BlockHash, Long]] = {
+    def hashParents(hash: BlockHash, genesisBlockNumber: Long): F[List[BlockHash]] = {
```
birchmd: Looks like this doesn't need to return an `F` any more because everything is pure.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -25,7 +25,7 @@ object Estimator {
     */
   def tips[F[_]: Monad: BlockStore](
       blockDag: BlockDag,
-      genesisBlockHash: BlockHash
+      genesis: BlockMessage
```
birchmd: Perhaps `lastFinalizedBlock` would be a better name for this argument?

KentShikama:
@birchmd PTAL

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -25,7 +25,7 @@ object Estimator {
     */
   def tips[F[_]: Monad: BlockStore](
       blockDag: BlockDag,
-      genesisBlockHash: BlockHash
+      lastFinalizedBlock: BlockMessage
```
birchmd: I don't think you need to pass in the whole `BlockMessage` here. I think passing in the `BlockHash` as before is sufficient and you can look up the block number (which is the only other field you use as far as I can tell) from `blockDag.dataLookup`.

PR #1633, opened 2018-10-08
Add Drone CI

woky:
This adds Drone CI configuration and scripts. There are small tweaks to p2p-test-tool to allow concurrent executions. After this gets merged, PRs will be tested until unit tests and merges will be tested until the end.

CI can be reached https://drone.pyr8.io/ here. Members of rchain GitHub organization should have access.

There's a problem with some unit test failing. Pipeline is set to continue even if some unit tests fail for now. Please see e.g. https://drone.pyr8.io/woky/rchain-testci/83/5 for failures.

EDIT: There're no test failures. See comments in RHOL-898.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,102 @@
+branches:
+  - dev
+  - master
+  - release-*
+  - testing*
+
+# Common parts
+
+# FYI: There's no merging of arrays in YAML. At least not in the Go library
+# Drone uses to parse YAML. So if you mix in &sbtenv and want to add
+# additional volumes, as in package step, you need to specify all volumes.
+
+__buildenv: &buildenv
+  image: rchain/buildenv:latest-java8
+
+__sbtenv: &sbtenv
+  <<: *buildenv
+  environment:
+    - _JAVA_OPTIONS=-Xms2G -Xmx4G -XX:MaxMetaspaceSize=1G
+  volumes:
+    - /var/cache/rchain-build/.sbt:/root/.sbt
+    - /var/cache/rchain-build/.ivy2:/root/.ivy2
+
+# Build definition
+
+clone:
+  git-clone:
+    image: plugins/git:next
+
+pipeline:
+
+  # pr
+
+  sbt-update:
+    <<: *sbtenv
+    commands:
+      - sbt update
+
+  compile:
+    <<: *sbtenv
+    commands:
+      - sbt rholang/bnfc:generate compile test:compile it:compile
+
+  run-unit-tests:
+    <<: *sbtenv
+    commands:
+      - sbt test it:test
+
+  run-code-analysis:
+    <<: *sbtenv
+    commands:
+      - sbt scalafmtCheck coverage
+      - ./scripts/ci/codecov-upload
+
+  # push
+
+  package:
+    <<: *sbtenv
+    commands:
+      - sbt node/docker:publishLocal node/rpm:packageBin node/debian:packageBin node/universal:packageZipTarball
+    volumes:
+      - /var/cache/rchain-build/.sbt:/root/.sbt
+      - /var/cache/rchain-build/.ivy2:/root/.ivy2
+      - /var/run/docker.sock:/var/run/docker.sock
+    when:
+      event: push
+
+  run-integration-tests:
+    <<: *buildenv
+    commands:
+      - ./scripts/ci/run-p2p-test
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - /tmp:/tmp
+    when:
+      event: push
+
+  publish:
+    <<: *buildenv
+    commands:
+      - ./scripts/ci/publish-artifacts
+    secrets:
+      - docker_username
+      - docker_password
+      - ssh_private_key
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+    when:
+      event: push
+
+  notify:
+    <<: *buildenv
+    commands:
+      - ./scripts/ci/send-email
+    secrets:
+      - email_host
+      - email_username
+      - email_password
+      - email_recipients
+    when:
+      event: push
+      status: [ changed, failure ]
--- a/scripts/ci/publish-artifacts
+++ b/scripts/ci/publish-artifacts
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -eo pipefail
+
+if [[ -n $DRONE_TAG ]]; then
+	REF=$DRONE_TAG
+	TAG=1
+elif [[ -n $DRONE_BRANCH ]]; then
+	REF=$DRONE_BRANCH
+	TAG=0
+else
+	# Like, really. It'll overwrite your SSH key.
+	echo "This script doesn't work outside of Drone" >&2
+	exit 1
+fi
+
+# Upload packages to repo.pyr8.io
+
+(
+umask 077
+mkdir -p ~/.ssh
+echo "$SSH_PRIVATE_KEY" | base64 -d | tr -d '\r' >~/.ssh/id_rsa
+)
+
+# We don't check host keys because artifacts are public.
+rsync -avrz --delete \
+	-e "ssh -p 22 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null" \
+	node/target/rnode_*_all.deb \
+	node/target/rpm/RPMS/noarch/rnode-*.noarch.rpm \
+	node/target/universal/*.tgz \
+	inbound@dispatch.pyr8.io:/home/inbound/cd/artifacts/$REF/
+
+# Upload Docker image
+
+builtin echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin
+docker tag coop.rchain/rnode:latest rchain/rnode:$REF
+docker push rchain/rnode:$REF
+if [[ $TAG == 1 ]]; then
+	docker tag coop.rchain/rnode:latest rchain/rnode:latest
+	docker push rchain/rnode:latest
+fi
--- a/scripts/ci/send-email
+++ b/scripts/ci/send-email
@@ -1,0 +1,30 @@
+#!/usr/bin/python3
+
+import os
+from smtplib import SMTP
+from email.mime.text import MIMEText
+
+repo = os.environ['DRONE_REPO_NAME']
+if 'DRONE_TAG' in os.environ:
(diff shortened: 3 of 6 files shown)


.drone.yml
```diff
@@ -0,0 +1,96 @@
+branches:
+  - dev
+  - master
+  - release-*
+  - testing
```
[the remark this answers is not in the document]
ArturGajowy: We can include a restriction for the branch builds, but honestly I wouldn't bother until we find that the branch build for every push is a problem. The build speed is very promising, so wouldn't worry for now.

.drone.yml
```diff
@@ -0,0 +1,96 @@
...
+  package:
+    <<: *sbtenv
+    commands:
+      - sbt node/docker:publishLocal node/rpm:packageBin node/debian:packageBin node/universal:packageZipTarball
+    volumes:
+      - /var/cache/rchain-build/.sbt:/root/.sbt
+      - /var/cache/rchain-build/.ivy2:/root/.ivy2
+      - /var/run/docker.sock:/var/run/docker.sock
+    when:
+      event: push
```
ArturGajowy: We should be doing this for the PR build as well, otherwise we don't know we've broken packaging till we're in dev.

.drone.yml
```diff
@@ -0,0 +1,96 @@
...
+
+  run-integration-tests:
+    <<: *buildenv
+    commands:
+      - ./scripts/ci/run-p2p-test
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - /tmp:/tmp
+    when:
+      event: push
```
ArturGajowy: How long do those take on drone?

I'd be in favor of including them in the PR build, just like the overwhelming majority back in July:
https://discordapp.com/channels/391435113415573504/391435113415573507/466034933249605643

.drone.yml
```diff
@@ -0,0 +1,96 @@
...
+    <<: *buildenv
+    commands:
+      - ./scripts/ci/send-email
+    secrets:
+      - email_host
+      - email_username
+      - email_password
+      - email_recipients
+    when:
+      event: push
```
ArturGajowy: I believe notificaitons should apply to all builds (please delete this line)

.drone.yml
```diff
@@ -0,0 +1,96 @@
+branches:
+  - dev
+  - master
+  - release-*
+  - testing
+
+__buildenv: &buildenv
```
ArturGajowy: I was aware of yaml anchors, but didn't remember the syntax, especially for dereferencing them - and was quite confused by the `__`, the `&`, and `<<: *`. Having a comment here would let me understand that faster.

Please add a comment saying `# Common parts` here, and `# Build definition` above `clone` :)

scripts/ci/publish-artifacts
```diff
@@ -0,0 +1,40 @@
...
+	exit 1
+fi
+
+# Upload packages to repo.pyr8.io
+
+(
+umask 077
+mkdir -p ~/.ssh
+echo "$SSH_PRIVATE_KEY" | base64 -d | tr -d '\r' >~/.ssh/id_rsa
+)
```
adaszko: Why is this executed in a subshell?

scripts/ci/publish-artifacts
```diff
@@ -0,0 +1,40 @@
...
+
+# Upload Docker image
+
+builtin echo $DOCKER_PASSWORD | docker login -u $DOCKER_USERNAME --password-stdin
+docker tag coop.rchain/rnode:latest rchain/rnode:$REF
+docker push rchain/rnode:$REF
+if [[ $TAG == 1 ]]; then
+	docker tag coop.rchain/rnode:latest rchain/rnode:latest
+	docker push rchain/rnode:latest
+fi
```
adaszko: It would be nice to break this file down into bash functions.  Having one, large scope gets harder to comprehend with every addition to the file.

scripts/ci/send-email
```diff
@@ -0,0 +1,30 @@
...
+''')
+msg['From'] = 'Lumbergh CI <noreply-ci@pyr8.io>'
+msg['To'] = os.environ['EMAIL_RECIPIENTS']
+msg['Subject'] = f'''Build {bn} of {repo}:{ref} FAILED'''
+
+with SMTP(os.environ['EMAIL_HOST'], 587) as smtp:
+    smtp.starttls()
+    smtp.login(os.environ['EMAIL_USERNAME'], os.environ['EMAIL_PASSWORD'])
+    smtp.send_message(msg)
+    smtp.quit()
```
adaszko: Again, why not put this code into functions

scripts/p2p-test-tool.py
```diff
@@ -130,7 +131,8 @@
 client = docker.from_env()
 RNODE_CMD = '/opt/docker/bin/rnode'
 # bonds_file = f'/tmp/bonds.{args.network}' alternate when dynamic bonds.txt creation/manpiulation file works
-bonds_file = os.path.dirname(os.path.realpath(__file__)) + '/demo-bonds.txt'
+src_bonds_file = os.path.dirname(os.path.realpath(__file__)) + '/demo-bonds.txt'
```
adaszko: `os.path.join()` is a better `+` for paths

scripts/ci/publish-artifacts
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+set -eo pipefail
+
+if [[ -n $DRONE_TAG ]]; then
```
adaszko: 👍 for using `[[` instead of `[`

.drone.yml
woky: Done.

scripts/ci/publish-artifacts
adaszko: `chmod` would be more explicit here then

scripts/ci/send-email
adaszko: >  It's just a tiny little single-purpose loveless script. :-)
With several global variables and the burden to reorganize it into functions is shifted to the next person adding something the file.  It is a reasonable assumption to assume it will get changed eventually.

scripts/p2p-test-tool.py
adaszko: Running on Windows is no the only reason to use it.  It eliminates duplicated slashes which become a problem if the path needs to be processed further.  It also indicates to the reader that a string is indeed a path.

scripts/ci/publish-artifacts
woky: I'd have to do it twice.

scripts/ci/publish-artifacts
woky: I find it more readable than if there were functions. Feel free to checkout your changes and push into my branch.

scripts/ci/send-email
woky: I find it more readable than if there were functions. Feel free to checkout your changes and push into my branch.

scripts/p2p-test-tool.py
woky: `dirname()` returns path without trailing `/` (I think). Feel free to checkout, make changes and push into my branch.

scripts/ci/publish-artifacts
adaszko: Small price to pay for improved readability.  Code is read multiple times by multiple people, most often written once by one person.

scripts/ci/publish-artifacts
adaszko: Is this is going to be the way to resolve PR disputes?  If so, I think we need to think of something better.

scripts/ci/publish-artifacts
jeremybusk: While I prefer being explicit in code I understand both sides of the issue. It is just as easy to look up what is going on. You can always add a comment to the side to aid understnading. i.e. <code> # some comment. In general, if it functionality of the code works and there is no definite issue with code I'll commit and then you can evolve the code in next PR. These issues seem to be hard to resolve mutually on small scripts. As it grows it becomes evident what needs to be done.

.drone.yml
ArturGajowy: I see, the docker sock thing is very unfortunate.

I think though since the build definition is taken from the branch being built, the drone machine can be compromised by **any** PR, regardless of what we put in .drone.yml. I can put whatever I want in the .drone.yml in my PR.

Unless we're using gated/protected builds:
https://discourse.drone.io/t/gated-builds-example-usage-help/702/4
Then every change in .drone.yml needs approval before being run.

Gated builds are underdocumented, but we have to enable those (or disable PR builds whatsoever ;/)

We should make a JIRA issue for that.

---

I'm reluctant to agree that checking packaging logic is time wasted for PR builds, but admittedly this fails much less often than integration tests do.

> if someone breaks it, merge-push build will catch it, just as it catches other regressions in integration test.

That's way too late. See the poll result.

I think we could guard the parts using docker-sock by checking the PR/branch author is in our ORG (we could call github or use a secret with all allowed usernames). Since we have to `guard` the .drone.yml anyway, putting the check into the .yml should make it safe.

---

Also, on the docker.sock / security issue:
Is the drone docker host in a DMZ subnet?
Is it running in a VM?
I think - given the threats - both answers should be true.

.drone.yml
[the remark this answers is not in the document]
ArturGajowy: >  I don't think it's plausible to make everyone aware that giving push access to a collaborator equals giving them ability to get root access to part of our infrastructure. 

As mentioned in me previous comment, it's even worse: we're giving ability to get root access on the drone to anyone issuing PRs.

Unless we enable gated builds :)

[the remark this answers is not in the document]
ArturGajowy: > Build failures result in email notifications and I don't think anyone following our mailing list cares about someone's code failing when the said person uses branches in main repo just to stage their work in preparation for PR

What you're saying is: notifications are broken (sent to everyone), so we don't allow building other branches.

While I agree building other branches might not be necessary, it should be noted we **could** build all the branches after we fix notifications to only spam the ones involved in the changes causing the failure. Push access is restricted to collaborators anyway, as you said. This is mostly a note to self :)

ArturGajowy:
Can't wait for a faster build with drone! :fireworks:

.drone.yml
ArturGajowy: BTW: how was this docker.sock issue solved for gitlab? I'm sure we're building docker images there - what prevents a PR/branch author from getting root in there?

woky:
@ArturGajowy I agree with your comments. Let's address it later.

PR #1634, opened 2018-10-08
Fix non-deterministic test failure caused by lack of reAttempt

KentShikama:
There are two problems this PR fixes:

1) We previously removed block hashes that were dependency free that
weren't yet in the block buffer. Thus when the block B actually came in
any block that depended on B might have not been triggered to be added.

2) We forgot that when we added the "processing" BlockStatus that now this
makes it such that some blocks may fail on reAttempt even though they
are dependencyFree. So while it is a bit ugly, we whitelist instead of
blacklist which blocks get successfully added during a reAttempt call.

https://rchain.atlassian.net/browse/RHOL-944

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -554,21 +554,75 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
       dependencyFreeBlocks = blockBuffer
         .filter(block => dependencyFree.contains(block.blockHash))
         .toList
-      attempts <- dependencyFreeBlocks.traverse(b => attemptAdd(b))
+      attempts <- dependencyFreeBlocks.traverse { b =>
+                   for {
+                     status <- attemptAdd(b)
+                   } yield (b, status)
+                 }
       _ <- if (attempts.isEmpty) {
             ().pure[F]
           } else {
-            Capture[F].capture {
-              dependencyFreeBlocks.map {
-                blockBuffer -= _
-              }
-            } *>
-              blockBufferDependencyDagState.set(dependencyFree.foldLeft(blockBufferDependencyDag) {
-                case (acc, hash) =>
-                  DoublyLinkedDagOperations.remove(acc, hash)
-              }) *> reAttemptBuffer
+            for {
+              _ <- removeAdded(blockBufferDependencyDag, attempts)
+              _ <- reAttemptBuffer
+            } yield ()
           }
     } yield ()
+
+  private def removeAdded(
+      blockBufferDependencyDag: DoublyLinkedDag[BlockHash],
+      attempts: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    for {
+      successfulAdds <- attempts
+                         .filter {
+                           case (_, status) => successfulAddStatus(status)
+                         }
+                         .pure[F]
+      _ <- unsafeRemoveFromBlockBuffer(successfulAdds)
+      _ <- removeFromBlockBufferDependencyDag(blockBufferDependencyDag, successfulAdds)
+    } yield ()
+
+  private def unsafeRemoveFromBlockBuffer(
+      successfulAdds: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    Sync[F].delay {
+      successfulAdds.map {
+        blockBuffer -= _._1
+      }
+      ()
+    }
+
+  private def removeFromBlockBufferDependencyDag(
+      blockBufferDependencyDag: DoublyLinkedDag[BlockHash],
+      successfulAdds: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    blockBufferDependencyDagState.set(
+      successfulAdds.foldLeft(blockBufferDependencyDag) {
+        case (acc, successfulAdd) =>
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -554,21 +554,75 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
...
+      status == InvalidFollows ||
+      status == InvalidSequenceNumber ||
+      status == InvalidShardId ||
+      status == JustificationRegression ||
+      status == NeglectedInvalidBlock ||
+      status == NeglectedEquivocation ||
+      status == InvalidTransaction ||
+      status == InvalidBondsCache ||
+      status == InvalidBlockHash ||
+      status == InvalidDeployCount
```
goral09: So everything except `Processing` and `BlockException` is a _successful add status_.  This is shorter:
```scala
status match {
  case Processing | BlockException => false
  case _ => true
}
```

goral09: OK, only now I read:
>  So while it is a bit ugly, we whitelist instead of blacklist which blocks get successfully added during a reAttempt call.

This is a good call, because we are we will be obviously broken, rather than un-obviously insecure.

So how about adding a field to `BlockStatus`:
```scala
sealed trait BlockStatus {
  val addedSuccessfuly: Boolean
}

case object Processing                   extends BlockStatus {
  override  val addedSuccessfuly: Boolean = false
}
case class BlockException(ex: Throwable) extends BlockStatus {
  override  val addedSuccessfuly: Boolean = false
}

sealed trait ValidBlock   extends BlockStatus {
  override  val addedSuccessfuly: Boolean = true
}
sealed trait InvalidBlock extends BlockStatus {
  override  val addedSuccessfuly: Boolean = true
}
```
and now the this method becomes:
```scala
  private def successfulAddStatus(status: BlockStatus): Boolean = status.addedSuccessfuly
```

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -554,21 +554,75 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
...
+      blockBufferDependencyDag: DoublyLinkedDag[BlockHash],
+      attempts: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    for {
+      successfulAdds <- attempts
+                         .filter {
+                           case (_, status) => successfulAddStatus(status)
+                         }
+                         .pure[F]
+      _ <- unsafeRemoveFromBlockBuffer(successfulAdds)
```
goral09: Why is this called `unsafe*` ? You are removing only blocks for which `addStatus` is _successful_ right?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -554,21 +554,75 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
...
+
+  private def unsafeRemoveFromBlockBuffer(
+      successfulAdds: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    Sync[F].delay {
+      successfulAdds.map {
+        blockBuffer -= _._1
+      }
+      ()
+    }
```
goral09: Personally I don't like using `()` , there is better looking alternative. `Sync[F].delay(…).as(Unit)` but it's your call.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -554,21 +554,75 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
...
+      _ <- unsafeRemoveFromBlockBuffer(successfulAdds)
+      _ <- removeFromBlockBufferDependencyDag(blockBufferDependencyDag, successfulAdds)
+    } yield ()
+
+  private def unsafeRemoveFromBlockBuffer(
+      successfulAdds: List[(BlockMessage, BlockStatus)]
+  ): F[Unit] =
+    Sync[F].delay {
+      successfulAdds.map {
+        blockBuffer -= _._1
```
goral09: This global and mutable `blockBuffer` might bite us back at some point. Please be very careful with access and modifications.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
KentShikama: Because we have something mutable going on inside

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
KentShikama: Yeah we've been meaning to get rid of it for a while

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
KentShikama: Yeah I like that

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
KentShikama: Sounds good

PR #1635, opened 2018-10-08
Add find block with deploy GRPC endpoint

KentShikama:
This endpoint takes a user/timestamp and maybe returns a block with such
a deploy.

https://rchain.atlassian.net/browse/RHOL-786


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -271,6 +271,56 @@ object BlockAPI {
...
+      case blockHash +: rem =>
+        for {
+          block <- ProtoUtil.unsafeGetBlock[F](blockHash)
+          result <- if (ProtoUtil.containsDeploy(block, user, timestamp)) {
+                     Option[BlockMessage](block).pure[F]
+                   } else {
+                     findBlockWithDeploy[F](rem, user, timestamp)
+                   }
+        } yield result
+      case _ => none[BlockMessage].pure[F]
```
goral09: I prefer to explicitly mark it as `Nil`. It is easy to read and obvious that you handle empty list case. `case _ => ` can match anything.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -56,5 +55,8 @@ private[api] object DeployGrpcService {
         Observable
           .fromTask(defer(BlockAPI.showMainChain[F](request.depth)))
           .flatMap(Observable.fromIterable)
+
+      override def findBlockWithDeploy(request: FindDeployInBlockQuery): Task[BlockQueryResponse] =
+        defer(BlockAPI.findBlockWithDeploy[F](request.user, request.timestamp))
```
goral09: @sebtomba Is this OK for casper to run its jobs on this `worker` scheduler? This is not a _short_ task (like dispatching) but CPU-bound.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -1,7 +1,7 @@
 package coop.rchain.node.api
 
 import cats.effect.Sync
-
+import com.google.protobuf.ByteString
```
goral09: Unused import?

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -215,6 +217,13 @@ object ProtoUtil {
       ProtoUtil.unsafeGetBlock[F](parentHash)
     }
 
+  // TODO: Ensure deploy.get and raw.get are safe
```
goral09: What about this **TODO** is this for the future Kent to solve?

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,35 +431,27 @@ object ProtoUtil {
...
+      DeployData()
+        .withUser(ByteString.EMPTY)
+        .withTimestamp(now)
+        .withTerm(term)
+        .withPhloLimit(accounting.MAX_VALUE)
+
+  def basicDeploy[F[_]: Monad: Time](id: Int): F[Deploy] =
+    for {
+      d    <- basicDeployData[F](id)
+      term = InterpreterUtil.mkTerm(d.term).right.get
```
goral09: I think this can fail and `.right.get` will throw.
```scala
      term <- Sync[F].fromEither(InterpreterUtil.mkTerm(d.term))
```
is cleaner and safer.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -35,6 +37,8 @@ class HashSetCasperTest extends FlatSpec with Matchers {
 
   import HashSetCasperTest._
 
+  implicit val timeEff = new LogicalTime[Id]
```
goral09: Do you need this global `timeEff` since you're creating one in each test?

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -146,4 +175,45 @@ class BlockQueryResponseAPITest extends FlatSpec with Matchers with BlockStoreFi
       s"Error: Failure to find block with hash ${badTestHashQuery}"
     )
   }
+
+  "findBlockWithDeploy" should "return error when no block matching query exists" in withStore {
```
goral09: I would group `findBlockWithDeploy` tests together, one after the other, rather than interleaving them with `showBlocks` tests.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -146,4 +175,45 @@ class BlockQueryResponseAPITest extends FlatSpec with Matchers with BlockStoreFi
...
+      val timestamp = 0L
+      val blockQueryResponse = BlockAPI.findBlockWithDeploy[Id](user, timestamp)(
+        syncId,
+        casperRef,
+        logEff,
+        turanOracleEffect,
+        blockStore
+      )
+      blockQueryResponse.status should be(
+        s"Error: Failure to find block containing deploy signed by  with timestamp ${timestamp.toString}"
```
goral09: Please put `${PrettyPrinter.buildString(user)}` after _signed by_ . It looks like you're missing it but it's just the pub key is empty.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -94,6 +94,35 @@ class BlockQueryResponseAPITest extends FlatSpec with Matchers with BlockStoreFi
     blockInfo.shardId should be(shardId)
   }
 
+  "findBlockWithDeploy" should "return successful block info response" in withStore {
```
goral09: It's not clear for me by just looking at this test why it succeeds. There's no explicit process that sets up the underlying block store and/or Casper to actually have the `(ByteString.EMPTY, 1L)` deploy in it. Please make this explicit otherwise in the future someone will be looking at this code pulling his hair trying to figure out why this test passes.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
KentShikama: Well I intentionally put as close as possible to the actual rendering in the be(...) part. I can add `${PrettyPrinter.buildString(user)}` if you think that is clearer.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -271,6 +271,56 @@ object BlockAPI {
     )
   }
 
+  private def findBlockWithDeploy[F[_]: Monad: Log: BlockStore](
```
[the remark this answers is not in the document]
goral09: You're right. I missed that.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -271,6 +271,56 @@ object BlockAPI {
...
+      blockHashes: Vector[BlockHash],
+      user: ByteString,
```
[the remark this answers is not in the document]
goral09: Can we rename it to pubKey then?

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
goral09: It is, thanks.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
goral09: By explicit I meant creating a block and then trying to look it up.

KentShikama: Yeah that is sort of what I did above.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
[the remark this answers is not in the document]
odeac: Or even better, use the appropriate wrappers: https://github.com/rchain/rchain/blob/dev/crypto/src/main/scala/coop/rchain/crypto/keys.scala

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
odeac: I understand Kent's pain that the optional fields in the proto classes are horrible to deal with but I totally agree with Mateusz and we should properly handle the missing field now, not in the future.

We should either getOrElse like in the code below, if that's a good default, or pass the responsibility of choosing such a default to the caller by changing the return type of the function to Optional[Boolean]

```scala
  def containsDeploy(b: BlockMessage, user: ByteString, timestamp: Long): Boolean = {
    deploys(b)
      .toStream
      .traverse(getDeployData)
      .map(_.exists(deployData => deployData.user == user && deployData.timestamp == timestamp))
      .getOrElse(false)
    
  def getDeployData(d : ProcessedDeploy) =
    for {
      deploy <- d.deploy
      deployData <- deploy.raw
    } yield  deployData
  }
```

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,35 +431,27 @@ object ProtoUtil {
...
+  def basicDeployData[F[_]: Monad: Time](id: Int): F[DeployData] =
+    for {
+      now  <- Time[F].currentMillis
+      term = s"@${id}!($id)"
+    } yield
+      DeployData()
+        .withUser(ByteString.EMPTY)
+        .withTimestamp(now)
+        .withTerm(term)
+        .withPhloLimit(accounting.MAX_VALUE)
```
odeac: Here I would use `map` instead of `for-expression` and move the `s"@${id}!($id)"` directly in the call to `withTerm`

The code becomes:
```scala
Time[F].currentMillis.map(now => ...)
```

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
KentShikama: Yup thanks for the simplification @odeac

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
KentShikama: Done

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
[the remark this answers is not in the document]
KentShikama: This naming is consistent with the deploy data proto. @birchmd is it ok if I change it there too?

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
KentShikama: Done

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -271,6 +271,56 @@ object BlockAPI {
...
+                                     blockInfo = Some(blockInfo)
+                                   )
+                               case None =>
+                                 BlockQueryResponse(
+                                   status =
+                                     s"Error: Failure to find block containing deploy signed by ${PrettyPrinter
+                                       .buildString(user)} with timestamp ${timestamp.toString}"
+                                 ).pure[F]
+                             }
+      } yield blockQueryResponse
```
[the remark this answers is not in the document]
KentShikama: Handling the errors last does indeed make it clearer, thanks

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,35 +431,27 @@ object ProtoUtil {
...
-    val deploy = basicDeploy(id)
-    ProcessedDeploy(
-      deploy = Some(deploy),
-      errored = false
-    )
-  }
+  def basicProcessedDeploy[F[_]: Monad: Time](id: Int): F[ProcessedDeploy] =
+    for {
+      deploy <- basicDeploy[F](id)
+    } yield ProcessedDeploy(deploy = Some(deploy))
```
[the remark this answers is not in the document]
KentShikama: > `map` has stronger restrictions than `flatMap`

Good to know

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
KentShikama: Done

KentShikama:
@goral09 @odeac PTAL

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,35 +436,24 @@ object ProtoUtil {
...
-      .withPhloLimit(accounting.MAX_VALUE)
-  }
-
-  def basicDeploy(id: Int): Deploy = {
-    val d    = basicDeployData(id)
-    val term = InterpreterUtil.mkTerm(d.term).right.get
-    Deploy(
-      term = Some(term),
-      raw = Some(d)
+  def basicDeployData[F[_]: Monad: Time](id: Int): F[DeployData] =
```
birchmd: Hmm, I thought we were working to phase out usages of `shared.Time` in favour of `cats.effect.Clock`/`cats.effect.Timer`

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -271,6 +271,44 @@ object BlockAPI {
...
+    )
+  }
+
+  private def findBlockWithDeploy[F[_]: Monad: Log: BlockStore](
+      blockHashes: Vector[BlockHash],
+      user: ByteString,
+      timestamp: Long
+  ): F[Option[BlockMessage]] =
+    blockHashes.toStream
+      .traverse(ProtoUtil.unsafeGetBlock[F](_))
```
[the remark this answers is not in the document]
KentShikama: ^ @odeac Do you know?

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
KentShikama: I'll let someone else handle all of them at once.

KentShikama:
@birchmd PTAL

And @odeac please see https://github.com/rchain/rchain/pull/1635#discussion_r225611494

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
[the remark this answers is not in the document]
birchmd: Regardless of what the answer is, `findM` would probably be more efficient I think.

KentShikama:
`findM` returns a blockhash so it doesn't quite work

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
goral09: @birchmd I think that with recent changes made by @sebtomba the `shared.Time` is now based on  the cats.

PR #1645, opened 2018-10-09
CI cleanup

woky:
- ci: Remove GitLab and Travis configs
- ci: Cleanup leftover files


adaszko:
> I wouldn't be too quick about that. For sure this should happen past 0.7 release. Ideally we'd use drone for building and otherwise doing the release, so that we have the whole process tested.

+1

PR #1647, opened 2018-10-10
Handle any exceptions that bubble up to the GRPC server

KentShikama:
This allows us to get useful stack traces on the server side.
We can clean up any Try/Catch call within the Casper codebase now.

https://rchain.atlassian.net/browse/RHOL-917

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -73,15 +73,7 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: 
     result
   }
 
-  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] =
-    Try(unsafeComputeBonds(hash)) match {
-      case Success(value) => value
-      case Failure(ex) =>
-        ex.printStackTrace()
-        throw ex
-    }
-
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -73,15 +73,7 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
     result
   }
 
-  def computeBonds(hash: StateHash)(implicit scheduler: Scheduler): Seq[Bond] =
```
goral09: So, previous implementation was catching the exception and printing it (on the server side) but now you allow it to bubble all the way up to `DeployGrpcService` ?

goral09:
I think that @rabbitonweb and @sebtomba are better suited for this review. I am not familiar with thread models and the exception handling in the node.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -24,7 +25,10 @@ private[api] object DeployGrpcService {
     new CasperMessageGrpcMonix.DeployService {
 
       private def defer[A](task: F[A]): Task[A] =
-        Task.defer(task.toTask).executeOn(worker)
+        Task.defer(task.toTask).executeOn(worker).attempt.flatMap {
```
DoSOfRedRiver: Is that a case for `adaptError`?

KentShikama: I'll look into this later. Will merge for now.

PR #1649, opened 2018-10-10
Integrate new transport layer API into Casper

KentShikama:
We now use the new stream endpoint to send blocks.

https://rchain.atlassian.net/browse/RHOL-772


casper/src/test/scala/coop/rchain/casper/util/comm/TransportLayerTestImpl.scala
```diff
@@ -34,15 +35,16 @@ class TransportLayerTestImpl[F[_]: Monad](
...
       handleStreamed: Blob => F[Unit]
   ): F[Unit] =
     TransportLayerTestImpl.handleQueue(dispatch, msgQueues(identity))
 
-  // FIX-ME
-  def stream(peers: Seq[PeerNode], blob: Blob): F[Unit] = ???
+  def stream(peers: Seq[PeerNode], blob: Blob): F[Unit] =
+    peers.toList
+      .traverse(peer => send(peer, protocol(blob.sender).withPacket(blob.packet)))
+      .map(_.toSeq)
```
birchmd: What is the point of this `map` since the return type is `F[Unit]`?

Or better yet, why not 
```scala
def stream(peers: Seq[PeerNode], blob: Blob): F[Unit] =
  broadcast(peers, protocol(blob.sender).withPacket(blob.packet)).void
```

rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
```diff
@@ -44,10 +44,6 @@ trait LMDBOps extends CloseOps {
       val ret: R = f(txn)
       txn.commit()
       ret
-    } catch {
```
birchmd: This deletion seems unrelated to the purpose of the PR.

comm/src/test/scala/coop/rchain/p2p/EffectsTestInstances.scala
```diff
@@ -82,18 +81,12 @@ object EffectsTestInstances {
...
-          map.updated(peer, msgs)
-      }
-      broadcasted = newMap
+      requests = requests ++ peers.map(peer => Request(peer, msg))
       Seq()
     }
 
-    def stream(peers: Seq[PeerNode], blob: Blob): F[Unit] = ???
+    def stream(peers: Seq[PeerNode], blob: Blob): F[Unit] =
+      broadcast(peers, ProtocolHelper.protocol(blob.sender).withPacket(blob.packet)).as(())
```
birchmd: I think `.void` is clearer than `.as(())`

rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
KentShikama: It was an addition to https://github.com/rchain/rchain/pull/1647 that I decided not to commit. I think it is no longer necessary and but it definitely should go in its own PR. Thanks for the catch.

KentShikama:
@birchmd PTAL

PR #1670, opened 2018-10-12
Refactor certificate utils

DoSOfRedRiver:
This PR is related to two tickets:
https://rchain.atlassian.net/browse/CORE-694
https://rchain.atlassian.net/browse/CORE-672

i.e. this is a refactoring of functionality related to certificates utils


comm/src/main/scala/coop/rchain/comm/transport/package.scala
```diff
@@ -0,0 +1,51 @@
...
+case class Tls(
+    certificate: Path,
+    key: Path,
+    customCertificateLocation: Boolean,
+    customKeyLocation: Boolean,
+    secureRandomNonBlocking: Boolean
+)
+
+package object transport {
+  def generateSertificateIfAbsent[A: Iso[?, Tls]](a: A) = {
```
rabbitonweb: This is effectful, please close over `F[_]`

comm/src/main/scala/coop/rchain/comm/transport/package.scala
```diff
@@ -0,0 +1,51 @@
...
+    val tls = Iso[A, Tls].to(a)
+
+    // Generate certificate if not provided as option or in the data dir
+    if (!tls.customCertificateLocation
+        && !tls.certificate.toFile.exists()) {
+      println(s"No certificate found at path ${tls.certificate}")
```
rabbitonweb: Please refactor to use `Log` typeclass

crypto/src/test/scala/coop/rchain/crypto/util/CertificateHelperTest.scala
```diff
@@ -0,0 +1,3 @@
+package coop.rchain.crypto.util class CertificateHelperTest {
```
rabbitonweb: Missing tests

comm/src/main/scala/coop/rchain/comm/transport/package.scala
DoSOfRedRiver: It wasn't previously. Where can I place the call of this function? `NodeRuntime.nodeProgram`?

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -70,36 +69,7 @@ class NodeRuntime(conf: Configuration, host: String, scheduler: Scheduler) {
...
-      val keyPair = CertificateHelper.generateKeyPair(conf.tls.secureRandomNonBlocking)
-      withResource(new java.io.PrintWriter(conf.tls.certificate.toFile)) { pw =>
-        pw.write(CertificatePrinter.print(CertificateHelper.generate(keyPair)))
-      }
-      withResource(new java.io.PrintWriter(conf.tls.key.toFile)) { pw =>
-        pw.write(CertificatePrinter.printPrivateKey(keyPair.getPrivate))
-      }
-    }
-  }
+  transport.generateSertificateIfAbsent(conf.tls)
```
goral09: Am I reading this correctly that `generateSertificateIfAbsent` here will become generateSertificateIfAbsent[Tls]` which means it will require `Iso[Tls, Tls]` ?

DoSOfRedRiver: Nope. It will require `Iso[coop.rchain.node.configuration.Tls, coop.rchain.comm.transport.Tls]`.

comm/src/main/scala/coop/rchain/comm/transport/GenerateCertificateIfAbsent.scala
```diff
@@ -0,0 +1,86 @@
...
+      generateCertificate(tls)
+    } else ().pure[F]
+  }
+
+  def generateCertificate(tls: Tls): F[Unit] =
+    for {
+      _ <- info(s"No certificate found at path ${tls.certificate}")
+      _ <- info("Generating a X.509 certificate for the node")
+      _ <- // If there is a private key, use it for the certificate
+      if (tls.key.toFile.exists()) readKeyPair(tls)
```
rabbitonweb: `File.exists()` should be lifted to `F`, then `ifM` can be used on it

comm/src/main/scala/coop/rchain/comm/transport/GenerateCertificateIfAbsent.scala
```diff
@@ -0,0 +1,86 @@
...
+import cats.syntax.applicativeError._
+import cats.syntax.flatMap._
+import cats.syntax.functor._
+
+import coop.rchain.crypto.util.{CertificateHelper => certHelp, CertificatePrinter => certPrint}
+import coop.rchain.shared.{Iso, Log}
+
+import scala.language.higherKinds
+
+class GenerateCertificateIfAbsent[F[_]: Sync](implicit log: Log[F]) {
```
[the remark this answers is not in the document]
DoSOfRedRiver: It because I want to import `log` functions and use them like `_ <- info("blabla")`

node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -38,101 +49,93 @@ import org.http4s.server.{Server => Http4sServer}
...
+    !conf.tls.key.toFile.exists(),
+    s"Secret key file ${conf.tls.certificate} not found"
+  )
+}
 
-    if (publicKeyHash.isEmpty) {
-      println("Certificate must contain a secp256r1 EC Public Key")
-      System.exit(1)
-    }
+class NodeRuntime(
```
sebtomba: I believe that we need to make this class now private.

shared/src/main/scala/coop/rchain/shared/Log.scala
```diff
@@ -11,6 +17,21 @@ object LogSource {
   def apply(c: Class[_]): LogSource = new LogSource {
     val clazz: Class[_] = c
   }
+
+  implicit def matLogSource: LogSource = macro LogSourceMacros.mkLogSource
```
rabbitonweb: can you provide context what this macro is doing? maybe it's worth documenting in comments?

rabbitonweb: I mean, I assume it adds automatically `LogSource` but I don't see usage so I'm not 100% certain

DoSOfRedRiver: It is used in `GenerateCertificateIfAbsent` but I think we could eventually replace all manual definitions.

shared/src/main/scala/coop/rchain/shared/Iso.scala
```diff
@@ -0,0 +1,22 @@
+package coop.rchain.shared
+import shapeless.LabelledGeneric
+
+trait Iso[A, B] {
+  def to(a: A): B
+  def from(b: B): A
+}
+
+object Iso {
```
rabbitonweb: I understand what `isomorphism` is doing in this abstract context, but I don't understand why we are using it? Is that in order to abstract from the context from "what" a TLS is created?

comm/src/main/scala/coop/rchain/comm/transport/GenerateCertificateIfAbsent.scala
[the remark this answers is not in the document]
rabbitonweb: I guess this is not a problem, just a bit different from other parts of code. I guess it's ok...

shared/src/main/scala/coop/rchain/shared/Log.scala
rabbitonweb: awesome

shared/src/main/scala/coop/rchain/shared/Iso.scala
rabbitonweb: Ok, I see, dependencies between modules. Why the shapeless implementaiton though?

DoSOfRedRiver: I can't move `Tls` class to `comm` because it is part of the configuration, and I can't use it due to cyclic dependencies in sbt, so I just created the same class which is converted implicitly with `Iso` (to avoid that ugly manual converting).

PR #1671, opened 2018-10-12
Fix missing timestamps for some log message

DoSOfRedRiver:
https://rchain.atlassian.net/browse/CORE-791

### DIFF
--- a/node/src/main/scala/coop/rchain/node/node.scala
+++ b/node/src/main/scala/coop/rchain/node/node.scala
@@ -5,6 +5,7 @@ import cats.data._
 import cats.effect._
 import cats.implicits._
 
+import com.typesafe.scalalogging.Logger
 import coop.rchain.blockstorage.{BlockStore, LMDBBlockStore}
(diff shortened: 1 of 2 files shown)


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -47,14 +48,18 @@ class NodeRuntime(conf: Configuration, host: String, scheduler: Scheduler) {
 
   private implicit val logSource: LogSource = LogSource(this.getClass)
 
+  val logger = Logger(this.getClass)
```
rabbitonweb: `node.scala` returns a program that runs under the `Effect`, we should *NOT* introduce a logger here, we should reuse `Log[Effect]`. Yes, I know part of this code is imperative, but we should not contribute to its "imperativeness" even more. This was a fairly easy task (replace println with logs), pushing a bit of technical debt here seems logical

DoSOfRedRiver:
Implemented as a part of https://github.com/rchain/rchain/pull/1670

PR #1674, opened 2018-10-12
CORE-1316 Add compression to stream api

rabbitonweb:
## Overview
Based on previous @tymm work.
All messages that go through `TransportLayer.stream` are compressed by default, and decompressed when received on the other side

CORE-1316


rabbitonweb:
@tymm s`stream ` is intended to be used *only* for bigger objects - like blocks

rabbitonweb:
What do you mean by not creating a thread pool?

tymm:
In `ByteStringOps.scala` I am creating a new fixed thread pool (with two threads) every time we compress or decompress (see line 60 and 109).
That is not really optimal due to the overhead of creating new threads.
The node should probably provide a suitable thread pool and pass it to `compressLZ4` and `decompressLZ4`. Maybe @sebtomba can help with that since he did the work on thread pool management in the node.

Can we really assume that blocks are bigger than 1MB most of the time?

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -0,0 +1,116 @@
...
+    val pos = new PipedOutputStream()
+    val pis = new PipedInputStream(pos)
+    val is  = bs.newInput()
+
+    // Writes compressed data into `pos`
+    val producer = new Runnable {
+      override def run(): Unit = {
+        var lz4os: FramedLZ4CompressorOutputStream = null
+        try {
+          lz4os = new FramedLZ4CompressorOutputStream(pos, Parameters.DEFAULT)
```
odeac: why do you use a var and not initialize lz4os with `new FramedLZ4CompressorOutputStream(...)` in the first place?

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -0,0 +1,116 @@
...
+          val data  = new Array[Byte](8192)
+          var count = is.read(data)
+
+          while (count != -1) {
+            lz4os.write(data, 0, count)
+            count = is.read(data)
+          }
```
odeac: this can be extracted into a `transferAll(InputStream, OutputStream)` function and used twice

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -0,0 +1,116 @@
...
+    val pis = new PipedInputStream(pos)
+    val is  = bs.newInput()
+
+    // Writes compressed data into `pos`
+    val producer = new Runnable {
+      override def run(): Unit = {
+        var lz4os: FramedLZ4CompressorOutputStream = null
+        try {
+          lz4os = new FramedLZ4CompressorOutputStream(pos, Parameters.DEFAULT)
+          val data  = new Array[Byte](8192)
```
odeac: why 8192?

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -0,0 +1,116 @@
...
+    // Needs a thread pool with at least two threads
+    val es = Executors.newFixedThreadPool(2)
+    es.submit(producer)
+    val bytestring = es.submit(consumer)
+    es.shutdown()
+
+    bytestring.get()
+  }
+
+  private def decompressLZ4(bs: ByteString): Option[ByteString] = {
```
odeac: this is mostly a copy+paste of the `compressLZ4` function. Can you avoid the duplication?

rabbitonweb:
@tymm I have not even noticed the executor. Why would we need to do it like that? Is is it due to one consumer and one producer working in parallel?

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
rabbitonweb: not really though, there are differences worth not keeping them DRY

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
rabbitonweb: This is the most likely java'like copy-paste. There is an try-catch part, which could be exchanged for `Try`

tymm: I wish I could have just copy-pasted this from somewhere ;)

I'm using a var because I want to close `lz4os` in the finally block in the case that `new FramedLZ4CompressorOutputStream(...)` explodes.

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
tymm: See for example this: https://stackoverflow.com/a/237495

By the way `java.io.BufferedReader` also uses 8192 as the default buffer size.

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
odeac: if the c-tor explodes there is nothing to close

tymm: Not necessarily.
For example `FramedLZ4CompressorOutputStream.close()` closes the output stream that gets passed to the constructor.
If the constructor explodes after the output stream is set to `this.out`, there is something to close.
After thinking about this it seems like line 44 and 92 are not needed.

tymm:
@rabbitonweb as discussed on Discord we just need to make sure that there are two threads available per compression (or decompression). We should have a thread pool where `compress` and `decompress` can run on.
I just thought it was better to leave that part for you guys because I didn't want to mess with the different thread pools in the node.

shared/src/main/scala/coop/rchain/shared/Compression.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  def compress(content: Array[Byte]): Array[Byte] = {
+    val maxCompressedLength = compressor.maxCompressedLength(content.length)
+    val compressed          = new Array[Byte](maxCompressedLength)
+    compressor.compress(content, 0, content.length, compressed, 0, maxCompressedLength)
+    compressed
+  }
+  def decompress(compressed: Array[Byte], decompressedLength: Int): Array[Byte] = {
+    val restored = new Array[Byte](decompressedLength)
+    decompressor.decompress(compressed, 0, restored, 0, decompressedLength);
```
tymm: `;` at the end of the line.

rabbitonweb:
thx @tymm for catching this! see my next commit

rabbitonweb:
@tymm I've fixed it and added test

```scala
    it("should compress effectivly when data is compressable (repeatable patterns)") {
      val word       = Array.fill(1024)((Random.nextInt(24) + 33).toByte)
      val ar         = Array.concat(List.fill(1024)(word): _*)
      val compressed = Compression.compress(ar)
      val ratio      = compressed.length.toDouble / ar.length.toDouble
      ar.length should be > compressed.length
      ratio shouldBe (0.005 +- 0.0001)
    }
```

shared/src/main/scala/coop/rchain/shared/Compression.scala
```diff
@@ -0,0 +1,28 @@
...
+  val compressor: LZ4Compressor         = factory.highCompressor(17)
+  val decompressor: LZ4FastDecompressor = factory.fastDecompressor()
+
+  def compress(content: Array[Byte]): Array[Byte] = {
+    val maxCompressedLength = compressor.maxCompressedLength(content.length)
+    val compressed          = new Array[Byte](maxCompressedLength)
+    val length              = compressor.compress(content, 0, content.length, compressed, 0, maxCompressedLength)
+    compressed.take(length)
+  }
+  def decompress(compressed: Array[Byte], decompressedLength: Int): Array[Byte] = {
```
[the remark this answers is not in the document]
rabbitonweb: This was not throwing exceptions, but I will add property based test just in case.

shared/src/test/scala/coop/rchain/shared/CompressionSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+    it("should decompress to uncompressed data") {
+      forAll(byteArrays) { ar: Array[Byte] =>
+        val compressed = Compression.compress(ar)
+        val backAgain  = Compression.decompress(compressed, ar.length)
+        backAgain.size shouldBe (ar.size)
+        backAgain shouldBe ar
+      }
+    }
+
+    it("should compress effectivly when data is compressable (repeatable patterns)") {
```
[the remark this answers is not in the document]
rabbitonweb: 👍

shared/src/main/scala/coop/rchain/shared/Compression.scala
rabbitonweb: java copy-paste :D

shared/src/main/scala/coop/rchain/shared/Compression.scala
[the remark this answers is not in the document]
tymm:
```
Exception in thread "main" net.jpountz.lz4.LZ4Exception: Error decoding offset 1936282 of input buffer
	at net.jpountz.lz4.LZ4JNIFastDecompressor.decompress(LZ4JNIFastDecompressor.java:39)
```
That's the exception I saw.

PR #1676, opened 2018-10-12
Swap out p2p-test-tool.py for integration tests

adaszko:
* https://rchain.atlassian.net/browse/CORE-1273


woky:
Keep in mind that multiple integration tests can run at the same time from multiple PR merges. It should be enough to use dedicated Docker network for each build. Example in current run script:

```
NET=t$DRONE_BUILD_NUMBER.rchain.coop
PEERCMD="run --bootstrap rnode://cb74ba04085574e9f0102cc13d39f0c72219c5bb@bootstrap.${NET}?protocol=40400&discovery=40404"

python3 -u ./scripts/p2p-test-tool.py -bt -p2 -m2048m -n$NET --peer-command "$PEERCMD"
```

`$NET` is passed to Docker run.

adaszko:
> Finally :)

@KentShikama Not so fast :)  This is at the experimentation stage still.  Notice the `do-no-merge` label.

adaszko:
> Keep in mind that multiple integration tests can run at the same time from multiple PR merges. It should be enough to use dedicated Docker network for each build. Example in current run script:

Currently, [the network name is generated randomly](https://github.com/rchain/rchain/blob/34afbc6551adec58814e8455e795e88bc2a1168a/integration-testing/rnode_testing/docker.py#L21), so the problem is averted.

PR #1684, opened 2018-10-15
Integration tests prep

adaszko:
* https://rchain.atlassian.net/browse/CORE-1273

Preparations needed before switching over from p2p tests to integration tests completely.

### DIFF
--- a/integration-testing/Pipfile
+++ b/integration-testing/Pipfile
@@ -9,6 +9,3 @@ pytest = "*"
 delayed-assert = "*"
 
 [dev-packages]
-
-[requires]
-python_version = "3.7"
--- a/scripts/ci/run-integration-tests.sh
+++ b/scripts/ci/run-integration-tests.sh
@@ -1,0 +1,6 @@
+#/bin/bash -ue
+
+cd /tmp/rchain/integration-testing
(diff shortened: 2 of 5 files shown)


scripts/ci/run-integration-tests.sh
```diff
@@ -0,0 +1,6 @@
+#/bin/bash -ue
+
+cd /tmp/rchain/integration-testing
+python3 -m pip install pipenv
```
odeac: this should go in integration-testing/install_dependencies.sh and here you should call that script. This way people can run integration-tests outside CI

adaszko: install_dependencies.sh is no more

odeac: then what do I need to do in order to run the tests locally?

integration-testing/Pipfile
```diff
@@ -9,6 +9,3 @@ pytest = "*"
 delayed-assert = "*"
 
 [dev-packages]
-
-[requires]
-python_version = "3.7"
```
woky: Don't you _require_ it for `f'Hello: {name_var}'`?

PR #1690, opened 2018-10-16
Remove known state hashes

itegulov:
## Overview
As per discussion with @KentShikama, assuming our state trie hasn't been corrupted, we aren't allowed to add a block until its parent has been processed. Thus, the known state hashes work only as a sanity check right now and need to be removed for the successful migration to the new BlockDagStorage API.

https://rchain.atlassian.net/browse/RHOL-540

### Notes


casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -548,24 +548,15 @@ object Validate {
...
-                                                               )
-                             (maybeStateHash, updatedknownStateHashes) = validateBlockCheckpointResult
-                           } yield (updatedknownStateHashes, maybeStateHash)
-                         }
+      maybeStateHash <- InterpreterUtil
+                         .validateBlockCheckpoint[F](
+                           block,
+                           dag,
+                           runtimeManager
+                         )
```
KentShikama: Note we allow for different blocks to be added in parallel. Are you sure removing the synchronization at this level is safe? If so, we at least should add a test for this.

itegulov: As far as I can tell, the only mutable parameter here (which could in theory break something after removing the synchronization) is `runtimeManager`. However, `runtimeManager` is implemented through `SyncVar`, so I do not think it really affects safeness of the code. I can add a test, sure. What kind of a test do you want to see? One that tries to `Validate.transaction` a bunch of blocks simultaneously?

KentShikama: That is a good point. The test shouldn't be necessary then. If you're bored enough, I would rather just see integration tests that will most likely force a bunch of blocks to be validated simultaneously.

PR #1698, opened 2018-10-17
Use InMemBlockStore in production

KentShikama:
The current LMDBBlockStore consumes the same amount of memory anyways (as it is memory-mapped).
Given how we are soon going to replace LMDBBlockStore, might as well just
eliminate any problems LMDB gives us the easy way.


node/src/main/scala/coop/rchain/node/node.scala
```diff
@@ -380,11 +383,12 @@ class NodeRuntime(conf: Configuration, host: String, scheduler: Scheduler) {
                         kademliaRPC
                       )
                       .toEffect
-    blockStore = LMDBBlockStore.create[Effect](conf.blockstorage)(
+    blockMap <- Ref.of[Effect, Map[BlockHash, BlockMessage]](Map.empty[BlockHash, BlockMessage])
+    blockStore = InMemBlockStore.create[Effect](
```
birchmd: I think a comment saying this change is temporary until the work by @itegulov is finished would be good.

itegulov:
What problems does LMDB give us? We are planning to use LMDB for indexing in the new block storage any way, so I am not sure if it makes sense to give up on the current problems.

birchmd:
Yes, I definitely want this to only be temporary. And we will have to write some stress tests for the new blockstore when it gets implemented so that we can be confident it works before it gets put into any release.

KentShikama:
https://rchain.atlassian.net/browse/RHOL-986

PR #1705, opened 2018-10-17
[WIP] Adapt casper to the new DAG API

itegulov:
## Overview
Title says it all.

Depends on https://github.com/rchain/rchain/pull/1690

https://rchain.atlassian.net/browse/RHOL-945

### Notes


casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -21,103 +19,103 @@ object Estimator {
...
-      blockDag: BlockDag,
+      blockDag: BlockDagRepresentation[F],
       lastFinalizedBlockHash: BlockHash,
       latestMessagesHashes: Map[Validator, BlockHash]
   ): F[IndexedSeq[BlockMessage]] = {
-    @tailrec
     def sortChildren(
-        blocks: IndexedSeq[BlockHash],
-        childMap: Map[BlockHash, Set[BlockHash]],
+        blocks: List[BlockHash],
```
birchmd: I don't like that this has become a `List`. The complexity for sorting a `List` is much higher than sorting an `IndexedSeq` because you don't have constant-time random element access. Since the whole point of this function is to do sorting I think `IndexedSeq` will be significantly more performant, which is why it used `IndexedSeq` originally.

shared/src/main/scala/coop/rchain/shared/StreamT.scala
```diff
@@ -29,6 +29,19 @@ sealed abstract class StreamT[F[_], +A] { self =>
...
   }
 
+  def filterF[AA >: A](p: A => F[Boolean])(implicit monad: Monad[F]): F[StreamT[F, AA]] =
+    self match {
+      case SCons(curr, lazyTail) =>
+        p(curr).flatMap { cond =>
+          if (cond)
+            monad.pure[StreamT[F, AA]](StreamT.cons(curr, lazyTail.map(_.flatMap(_.filterF(p)))))
+          else
+            lazyTail.value.flatMap(_.filterF(p))
```
birchmd: I don't think `StreamT` should eagerly try to find the first element after the filter is applied. It think it would be better to `map` over the tail instead of forcing its evaluation with `.value`.

shared/src/main/scala/coop/rchain/shared/StreamT.scala
```diff
@@ -29,6 +29,19 @@ sealed abstract class StreamT[F[_], +A] { self =>
...
+        }
+      case SLazy(lazyTail) => lazyTail.value.flatMap(_.filterF(p))
```
birchmd: Same here.

itegulov:
@birchmd You are right that we assume that we can hold DAG in memory in between checkpoints. However, as we discussed a couple of weeks ago, there could still be some rare cases where the older parts of the DAG are being accessed. In usual cases `BlockDagRepresentation[F[_]]` does not actually require any effectful actions, but in case we access an older part of the DAG, it will try to load it from disk (in case if it is not already loaded) into a weak reference and that is where its effectful nature is required. Note that `BlockDagRepresentation[F[_]]` is still immutable in terms of the view on DAG it represents: it hold a snapshot of the active part of the DAG (i.e. part after the last checkpoint) and optional weak references to older checkpoints. It can mutate in terms of loaded checkpoints, but not in terms of view on DAG.

I was also considering exposing a effectless methods to work with the DAG (e.g. `childrenShallow(block: BlockHash): Option[Set[BlockHash]]`), which would only try to look up data in the relevant parts of the DAG, but I decided against it since it can actually be harmful to use them. What do you think?

casper/src/main/scala/coop/rchain/casper/Estimator.scala
itegulov: The reason why I decided to change it to `List` is heavy usage of `traverse`, `flatTraverse` in `Estimator` in this PR which can't be easily invoked on `IndexedSeq`. I have looked through the code and failed to understand why `IndexedSeq` is needed here instead of a `List`. @KentShikama also did not know the reason  when I contacted him about this. I think you are wrong about sorting complexity here: both `List` and `IndexedSeq` use default `SeqLike.sorted` implementation which just collects the collection into an array, sorts it by using `java.util.Arrays.sort` and builds the collection back.

shared/src/main/scala/coop/rchain/shared/StreamT.scala
```diff
@@ -29,6 +29,20 @@ sealed abstract class StreamT[F[_], +A] { self =>
     case _: SNil[F]      => StreamT.empty[F, A]
   }
 
+  def filterF[AA >: A](p: A => F[Boolean])(implicit monad: Monad[F]): F[StreamT[F, AA]] =
+    self match {
+      case SCons(curr, lazyTail) =>
+        p(curr).flatMap { cond =>
```
birchmd: `p(curr).map` would now suffice because you call `pure` on every line where you return something in the `flatMap` body.

casper/src/main/scala/coop/rchain/casper/CreateBlockStatus.scala
```diff
@@ -1,13 +1,19 @@
 package coop.rchain.casper
 
+import cats.Monad
+import cats.implicits._
 import coop.rchain.casper.protocol.BlockMessage
 
 sealed trait CreateBlockStatus {
   def map(f: BlockMessage => BlockMessage): CreateBlockStatus = this
+  def mapF[F[_]: Monad](f: BlockMessage => F[BlockMessage]): F[CreateBlockStatus] =
```
birchmd: Again, `Monad` is a stronger constraint than you actually need.

itegulov: Yep, sorry

itegulov:
Shouldn't be merged until at least @KentShikama is happy with merging https://github.com/rchain/rchain/pull/1690. Might make sense to merge this only after DAG storage (https://github.com/rchain/rchain/pull/1716) is finished completely.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -476,22 +476,32 @@ object Validate {
...
+    dag.latestMessage(b.sender).map(_.get.justifications)
+    for {
+      maybeLatestMessagesFromSenderView <- dag
+                                            .latestMessage(b.sender)
+                                            .map(
+                                              _.map(
+                                                bm =>
+                                                  ProtoUtil.toLatestMessageHashes(bm.justifications)
+                                              )
+                                            )
```
KentShikama: Can we extract out the nested map?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -476,22 +476,32 @@ object Validate {
...
-          latestMessagesOfBlock,
-          latestMessagesFromSenderView,
-          genesis
-        )
-      case None =>
-        // We cannot have a justification regression if we don't have a previous latest message from sender
-        Applicative[F].pure(Right(Valid))
-    }
+    val latestMessagesOfBlock = ProtoUtil.toLatestMessageHashes(b.justifications)
+    dag.latestMessage(b.sender).map(_.get.justifications)
```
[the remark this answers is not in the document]
itegulov: Seems like an accidental junk. Thanks for catching this.

KellyatPyrofex:
@itegulov and @KentShikama, I'm working to make sense of open PRs. It looks like this one depends on https://github.com/rchain/rchain/pull/1716 (open) and 
https://github.com/rchain/rchain/pull/1690 (merged). 

What are next steps to complete PR1716 and this PR?

itegulov:
@KellyatPyrofex #1716 needs a little bit more work (and a new approval since Michael Birch's is no longer valid). Once it is merged I will just make sure that all tests are still passing and will merge this one. Once this is done, DAG/latest messages storage will officially replace the existing in-memory storage.

KentShikama:
Seems like you still got some conflicts but looks good in general

PR #1719, opened 2018-10-22
CORE-597 Implement Batch TrieUpdate Strategy

ys-pyrofex:
## Overview
When creating a checkpoint we're doing a long series of write operations related to every insert/delete. We want to optimize this.

https://rchain.atlassian.net/browse/CORE-597

### Notes
After investigating and partially implementing "reverse breadth-first" approach I found that to make it work we need to load entire trie of current branch to ram (which may be really huge). Instead I decided to implement all operations on lazy TrieCache which later pushes to the store only updated nodes. Now 99% of the time we're spend in read transactions, and only when all data prepared we use a single write transaction to save nodes. This approach greatly reduces time in write transacton and amount of data actually stored in the LMDB, however overall performance improvement isn't big because of huge serialization costs.


dzajkowski:
"After investigating and partially implementing "reverse breadth-first" approach I found that to make it work we need to load entire trie of current branch to ram (which may be really huge)" 
can you share the results of this investigation? I think it would be useful 
a) if others reviewed it 
b) if it was persisted for future reference.

"This approach greatly reduces time in write transacton and amount of data actually stored in the LMDB, however overall performance improvement isn't big because of huge serialization costs." 
can you also share the method you used to arrive at this conclusion? I strongly feel that it should be repeatable (e.g. a benchmark) and documented so we can refer to it.

I think that the need for such transparency has been stressed lately and we should adhere. A good spot IMO would be a confluence page.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/EvalBenchStateBase.scala
```diff
@@ -26,6 +29,11 @@ trait EvalBenchStateBase {
   def doSetup(): Unit = {
     deleteOldStorage(dbDir)
 
+    implicit val scheduler: Scheduler = monix.execution.Scheduler.Implicits.global
+
+    runtime.reducer.setAvailablePhlos(Cost(Integer.MAX_VALUE)).runSyncUnsafe(1.second)
+    runtime.replayReducer.setAvailablePhlos(Cost(Integer.MAX_VALUE)).runSyncUnsafe(1.second)
```
dzajkowski: we have in code `TaskContrib.unsafeRunSync` maybe it would make sense to use it?

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/ReplayWideCheckpoint.scala
```diff
@@ -0,0 +1,59 @@
...
+  @BenchmarkMode(Array(Mode.SingleShotTime))
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  @Fork(value = 1)
+  @Threads(1)
+  @Warmup(iterations = 0)
+  @Measurement(iterations = 1)
+  def wideReduce(bh: Blackhole, state: ReplayWideCheckpointState): Unit = {
+    println("begin wideReduce()")
+    val checkpoint = state.runtime.space.createCheckpoint()
+    //state.runtime.replaySpace.rig(checkpoint.root, checkpoint.log)
```
dzajkowski: in this needed?

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/ReplayWideCheckpoint.scala
```diff
@@ -0,0 +1,59 @@
...
+class ReplayWideCheckpoint {
+  @Benchmark
+  @BenchmarkMode(Array(Mode.SingleShotTime))
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  @Fork(value = 1)
+  @Threads(1)
+  @Warmup(iterations = 0)
+  @Measurement(iterations = 1)
+  def wideReduce(bh: Blackhole, state: ReplayWideCheckpointState): Unit = {
+    println("begin wideReduce()")
```
dzajkowski: not convinced that `println` are the best benchmarking tools.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/ReplayWideCheckpoint.scala
```diff
@@ -0,0 +1,59 @@
...
+    val checkpoint = state.runtime.space.createCheckpoint()
```
dzajkowski: maybe use the `Blackhole` to force the jvm not to swallow the result?

dzajkowski: have you verified how much data is being processed in this checkpoint? I'm under the strong impression (as in I checked) that wide will result in exactly one element (after collapse).
please choose a different scenario to test checkpointing.

rspace/src/main/scala/coop/rchain/rspace/history/Trie.scala
```diff
@@ -39,8 +62,29 @@ object Trie {
       }((codecByteVector :: codecNonEmptyPointer).as[Skip])
 
   def hash[K, V](trie: Trie[K, V])(implicit codecK: Codec[K], codecV: Codec[V]): Blake2b256Hash =
-    codecTrie[K, V]
-      .encode(trie)
-      .map((vector: BitVector) => Blake2b256Hash.create(vector.toByteArray))
-      .get
+    if(TrieCache.useCache) {
+      trie.asInstanceOf[HashedEntity].initialize {
```
lukasz-golebiewski: Do you intend to leave this `asInstanceOf` here?

lukasz-golebiewski:
In my opinion you should write a micro benchmark for the `createCheckpoint()` method. 
Aside from that, measuring performance using `printlns` and `System.currentMilllis` is error prone. Could you acquaint yourself with the following document? https://www.oracle.com/technetwork/articles/java/architect-benchmarking-2266277.html

dzajkowski:
rant:
"Implement Batch TrieUpdate Strategy" was the original task. 
I'm really surprised that the scope change was not discussed with the team.

ys-pyrofex:
I actually noted about progress and changes in approach on 2 stand-ups: on Monday, Oct 15 I informed that I completely unhappy with merge algorithm performance (it was even slower), and on Wednesday I told that after analyzing profiler results I'm going to try absolutely another approach to the createCheckpoint performance problem. Of course, next time if similar case occur, I will request a sidebar to discuss possible changes in approach.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/EvalBenchStateBase.scala
ys-pyrofex: It exactly suits my needs, updated also WideBenchBaseState, since it also used `.runSyncUnsafe(1.second)` approach.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/ReplayWideCheckpoint.scala
ys-pyrofex: Done, thanks.

ys-pyrofex: Lol. I actually tried to find a "big and slow contract", and wide looked "big enough" :) complete fault. I replaced this test with `CreateCheckpointBench`, which creates around 3000 trie updates (actual time depends from the counter in loop-with-wks.rho).

rspace/src/main/scala/coop/rchain/rspace/history/Trie.scala
ys-pyrofex: No,  `asInstanceOf` is already removed, and `if(TrieCache.useCache)` should be removed before final merging.

ys-pyrofex:
I've updated the CreateCheckpointBench, so it now runs both old and new versions. Here are test results from my machine (launched with `rspaceBench/jmh:run CreateCheckpoint -i 2 -wi 1 -f 1 -t 1`):
```
[info] # Run complete. Total time: 00:06:14
[info] Benchmark                                   Mode  Cnt      Score   Error  Units
[info] CreateCheckpointBench.createCheckpoint_new    ss   2      7367,102          ms/op
[info] CreateCheckpointBench.createCheckpoint_old    ss   2   47026,480          ms/op
[success] Total time: 389 s, completed 24.10.2018 19:01:43
```

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
```diff
@@ -0,0 +1,181 @@
...
+
+private[rspace] case class StoredItem[TItem](value: TItem) extends CachedItem[TItem] {
+  override def asOption: Option[TItem] = Some(value)
+}
+
+private[rspace] case class AbsentItem[TItem]() extends CachedItem[TItem] {
+  override def asOption: Option[TItem] = None
+}
+
+private[rspace] class TrieCache[T, K, V](
```
dzajkowski: don't you feel that this class is complex enough to merit some unit testing?

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
```diff
@@ -0,0 +1,181 @@
...
+    trieStore: ITrieStore[T, K, V],
+    val trieBranch: Branch,
+    trieStoreOwner: Boolean = false
```
dzajkowski: what's the purpose of this flag? (I did not see it used other then the default)

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
```diff
@@ -0,0 +1,181 @@
+package coop.rchain.rspace.history
+import coop.rchain.rspace.Blake2b256Hash
+
+import scala.collection.immutable.{Map => ImmutableMap}
+import scala.collection.immutable.Seq
+import scala.collection.mutable.Map
+
+private[rspace] trait CachedItem[TItem] {
```
dzajkowski: `sealed`?

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
```diff
@@ -0,0 +1,181 @@
...
+
+private[rspace] case class AbsentItem[TItem]() extends CachedItem[TItem] {
+  override def asOption: Option[TItem] = None
+}
+
+private[rspace] class TrieCache[T, K, V](
+    trieStore: ITrieStore[T, K, V],
+    val trieBranch: Branch,
+    trieStoreOwner: Boolean = false
+) extends ITrieStore[T, K, V] {
```
dzajkowski: have you had the chance to measure the memory impact on rnode?

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
```diff
@@ -0,0 +1,181 @@
...
+
+import scala.collection.immutable.{Map => ImmutableMap}
+import scala.collection.immutable.Seq
+import scala.collection.mutable.Map
+
+private[rspace] trait CachedItem[TItem] {
+  def asOption: Option[TItem]
+}
+
+private[rspace] case class LoadedItem[TItem](value: TItem) extends CachedItem[TItem] {
```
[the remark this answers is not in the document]
ys-pyrofex: LoadedItem is unmodified, and never send back to store. Only StoredItems are saved to the underlying store in `applyCache`.

rspace/src/main/scala/coop/rchain/rspace/IStore.scala
```diff
@@ -109,22 +112,40 @@ trait IStore[C, P, A, K] {
...
     _trieUpdateCount.get()
 
-  protected def processTrieUpdate(update: TrieUpdate[C, P, A, K]): Unit
-
   private[rspace] def clearTrieUpdates(): Unit = {
     _trieUpdates.update(const(Seq.empty))
     _trieUpdateCount.set(0L)
   }
 
+  protected def processTrieUpdate(cacheStore: TrieStoreType, update: TrieUpdate[C, P, A, K]): Unit
```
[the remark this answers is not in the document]
ys-pyrofex: I'm afraid of confusion as `*Type` looks more like a type, not a store. Replaced `cacheStore` with `store`.

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
ys-pyrofex: It is already covered vert well by existing unit-tests - there just few uncovered methods left, like `getEmptyRoot`, `clear` and `toMap` which aren't used at all, and implemented just for consistency and possibility to reuse the `TrieCache` in another similar scenarios.

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
ys-pyrofex: If it's enabled closing the cache closes also the underlying store. This might be convenient if we decide to create TrieCache-specific unit-tests or storages.

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
ys-pyrofex: Done.

rspace/src/main/scala/coop/rchain/rspace/LMDBOps.scala
```diff
@@ -119,5 +120,17 @@ trait LMDBOps extends CloseOps {
       if (!dbi.delete(txn, codecK.encode(key).get.bytes.toDirectByteBuffer)) {
         throw new Exception(s"could not delete: $key")
       }
+
+    def putBytes(txn: Txn[ByteBuffer], key: Blake2b256Hash, data: Array[Byte]): Unit =
+      if (!dbi.put(txn, key.bytes.toDirectByteBuffer, data.toDirectByteBuffer)) {
+        throw new Exception(s"could not persist array of ${data.length} bytes")
+      }
+
+    def putBytes[K](txn: Txn[ByteBuffer], key: K, data: Array[Byte])(
```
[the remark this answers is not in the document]
ys-pyrofex: You're right. Seems left from intermediate steps. Thanks.

rspace/src/main/scala/coop/rchain/rspace/history/InMemoryTrieStore.scala
```diff
@@ -140,6 +140,42 @@ class InMemoryTrieStore[K, V]
       hash: Blake2b256Hash
   ): Unit =
     txn.writeState(state => (state.changeEmptyRoot(hash), ()))
+
+  override private[rspace] def applyCache(
```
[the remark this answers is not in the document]
ys-pyrofex: We can disable it for the in-mem trie, however for the test loop with 15000 iterations I'm getting very close results:
```
[info] Benchmark                                   Mode  Cnt      Score   Error  Units
[info] CreateCheckpointBench.createCheckpoint_new    ss       25747,094          ms/op
[info] CreateCheckpointBench.createCheckpoint_old    ss       25932,673          ms/op
```

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
ys-pyrofex: To minimize memory pressure I'm using WeakReferences, so gc will be able to collect all additional instances if needed. On our loop with 3000 iterations:

With TrieCache peak memory usage: 0.48gb
Old version peak memory usage: 0.46gb

Profiler snapshots comparison shows that additional memory related to new classes takes less than 10 megabytes, however most of alive objects aren`t related to rspace, so it's hard to measure memory impact with precision of kilobytes.

rspace/src/main/scala/coop/rchain/rspace/IStore.scala
```diff
@@ -109,22 +112,40 @@ trait IStore[C, P, A, K] {
...
+        trieStore.applyCache(txn, trieCache, rootHash)
+      }
+      rootHash
+    } else {
+      collapse(trieUpdates).foreach(processTrieUpdate(trieStore, _))
+      trieStore.withTxn(trieStore.createTxnWrite()) { txn =>
+        val rootHash = trieStore
+          .persistAndGetRoot(txn, trieBranch)
+          .getOrElse(throw new Exception("Could not get root hash"))
+        rootHash
```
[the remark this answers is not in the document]
ys-pyrofex: Fixed.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/CreateCheckpointBench.scala
```diff
@@ -0,0 +1,56 @@
...
+
+  val useCache = false
+
+  override def doSetup(): Unit = {
+    TrieCache.useCache = this.useCache
+    super.doSetup()
+    val runTask = createTest(term, runtime.reducer)
+    assert(runTask.unsafeRunSync.isEmpty)
+  }
+}
```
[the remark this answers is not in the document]
ys-pyrofex: Seems a bug in scalafmt - it fails with exception in regex (they're analyzing code with regexes?!), fixed by formatting file-by-file.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/CreateCheckpointBench.scala
```diff
@@ -0,0 +1,56 @@
+package coop.rchain.rspace.bench.wide
```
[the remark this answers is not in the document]
ys-pyrofex: Agree, it's not so wide anymore.. Moved it away.

rspace-bench/src/test/scala/coop/rchain/rspace/bench/wide/CreateCheckpointBench.scala
```diff
@@ -0,0 +1,56 @@
...
+}
+
+@State(Scope.Benchmark)
+class CreateCheckpointBenchStateVsCache extends CreateCheckpointBenchState {
+  override val useCache = true
+}
+
+@State(Scope.Benchmark)
+class CreateCheckpointBenchState extends WideBenchBaseState {
+  override val rhoSetupScriptPath: String = "/rholang/loop-with-wks.rho"
```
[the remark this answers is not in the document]
ys-pyrofex: Left from the day when the bench used wide contract. Removed.

rspace/src/main/scala/coop/rchain/rspace/history/TrieCache.scala
lukasz-golebiewski: Is it possible to add a test that would enforce the removal of weak references and check whether the store still behaves correctly?

ys-pyrofex: Sounds interesting, will try implement it.

lukasz-golebiewski:
Let's get back to this when time allows

PR #1720, opened 2018-10-22
CORE-597 Batch TrieUpdate Strategy

ys-pyrofex:
## Overview
When creating a checkpoint we're doing a long series of write operations related to every insert/delete. We want optimize this.

https://github.com/rchain/rchain/pull/1719

After investigating and partially implementing "reverse breadth-first" approach I found that to make it work we need to load entire trie of current branch to ram (which may be really huge). Instead I decided to implement all operations on lazy TrieCache which later pushes to the store only updated nodes. Now 99% of the time we're spend in read transactions, and only when all data prepared we use a single write transaction to save nodes. This approach greatly reduces time in write transaciton and amount of data actually stored in the LMDB, however overall performance improvement isn't big because of huge serialization costs.


ys-pyrofex:
Duplicated due to git bugs.

PR #1725, opened 2018-10-22
Expose block timestamp to Rholang

KentShikama:
On the Rholang side, this is a similar to what was done with GET_DEPLOY_PARAMS .

https://rchain.atlassian.net/browse/RHOL-810

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -27,10 +27,10 @@ import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
 class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: SyncVar[Runtime]) {
-  def setTimestamp(now: Long): Unit = {
+  def setTimestamp(now: Long)(implicit scheduler: Scheduler): Unit = {
     val runtime        = runtimeContainer.take()
-    val timestamp: Par = Par(exprs = Seq(Expr(Expr.ExprInstance.GString(String.valueOf(now)))))
-    runtime.blockTime.setParams(timestamp)
+    val timestamp: Par = Par(exprs = Seq(Expr(Expr.ExprInstance.GInt(now))))
+    runtime.blockTime.setParams(timestamp).unsafeRunSync
     runtimeContainer.put(runtime)
   }
 
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -27,13 +27,6 @@ import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
 class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: SyncVar[Runtime]) {
-  def setTimestamp(now: Long)(implicit scheduler: Scheduler): Unit = {
-    val runtime        = runtimeContainer.take()
-    val timestamp: Par = Par(exprs = Seq(Expr(Expr.ExprInstance.GInt(now))))
-    runtime.blockTime.setParams(timestamp).unsafeRunSync
-    runtimeContainer.put(runtime)
-  }
-
   def captureResults(start: StateHash, term: Par, name: String = "__SCALA__")(
       implicit scheduler: Scheduler
   ): Seq[Par] = {
@@ -54,23 +47,39 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: 
     result.flatMap(_.a.pars)
   }
 
-  def replayComputeState(hash: StateHash, terms: Seq[InternalProcessedDeploy])(
+  def replayComputeState(
+      hash: StateHash,
+      terms: Seq[InternalProcessedDeploy],
+      time: Option[Long] = None
+  )(
       implicit scheduler: Scheduler
   ): Either[(Option[Deploy], Failed), StateHash] = {
     val runtime = runtimeContainer.take()
-    val result  = replayEval(terms, runtime, hash)
+    unsafeSetTimestamp(time, runtime)
+    val result = replayEval(terms, runtime, hash)
     runtimeContainer.put(runtime)
     result
   }
 
-  def computeState(hash: StateHash, terms: Seq[Deploy])(
+  def computeState(hash: StateHash, terms: Seq[Deploy], time: Option[Long] = None)(
       implicit scheduler: Scheduler
   ): (StateHash, Seq[InternalProcessedDeploy]) = {
     val runtime = runtimeContainer.take()
-    val result  = newEval(terms, runtime, hash)
+    unsafeSetTimestamp(time, runtime)
+    val result = newEval(terms, runtime, hash)
     runtimeContainer.put(runtime)
     result
   }
+
+  private def unsafeSetTimestamp(time: Option[Long], runtime: Runtime)(
+      implicit scheduler: Scheduler
+  ): Unit =
+    time match {
+      case Some(t) =>
+        val timestamp: Par = Par(exprs = Seq(Expr(Expr.ExprInstance.GInt(t))))
+        runtime.blockTime.setParams(timestamp).unsafeRunSync
+      case None => ()
+    }
 
   def storageRepr(hash: StateHash): Option[String] =
     getResetRuntimeOpt(hash).map { resetRuntime =>
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -27,6 +27,12 @@ import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
 class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: SyncVar[Runtime]) {
+  def setTimestamp(now: Long): Unit = {
+    val runtime        = runtimeContainer.take()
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -27,6 +27,12 @@ import scala.util.{Failure, Success, Try}
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
 class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer: SyncVar[Runtime]) {
+  def setTimestamp(now: Long)(implicit scheduler: Scheduler): Unit = {
+    val runtime        = runtimeContainer.take()
```
birchmd: I see you ended up making the `setTimestamp ` a separate method. I still think it would be better to include timestamp as a parameter of `computeState` and `replayComputeState`.

casper/src/test/scala/coop/rchain/casper/RholangBuildTest.scala
```diff
@@ -23,11 +23,15 @@ class RholangBuildTest extends FlatSpec with Matchers {
...
-      """new double, dprimes, rl(`rho:registry:lookup`), ListOpsCh in {
+      """new double, dprimes, rl(`rho:registry:lookup`), ListOpsCh, time(`rho:block:timestamp`), timeRtn, stdout(`rho:io:stdout`) in {
         |  contract double(@x, ret) = { ret!(2 * x) } |
         |  rl!(`rho:id:dputnspi15oxxnyymjrpu7rkaok3bjkiwq84z7cqcrx4ktqfpyapn4`, *ListOpsCh) |
         |  for(@(_, ListOps) <- ListOpsCh) {
         |    @ListOps!("map", [2, 3, 5, 7], *double, *dprimes)
+        |  } |
+        |  time!(*timeRtn) |
+        |  for (@timestamp <- timeRtn) {
+        |   stdout!(timestamp)
```
birchmd: This is insufficient as a test of this feature in my opinion. I would prefer to see something added to `HashSetCasperTest ` that actually checks the output is correct.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -38,7 +41,7 @@ object SystemProcesses {
         dispatcher: Dispatch[Task, ListParWithRandomAndPhlos, TaggedContinuation]
     ): Task[Unit] =
       res.fold(err => Task.raiseError(OutOfPhlogistonsError), _.fold(Task.unit) {
-        case (cont, channels) => _dispatch(dispatcher)(cont, channels)
+        case (cont, channels) => Task.pure(_dispatch(dispatcher)(cont, channels).unsafeRunSync)
```
birchmd: This seems bad. You are making a Task that runs a Task unsafely... Why not just leave it the way it was?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
KentShikama: No I'm still WIP. I pushed this code specifically to show you the unsafeRunSync I want to remove

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
KentShikama: Yeah so this is what I wanted to show you. The test doesn't work unless I do this...

KentShikama: Specifically it doesn't print the timestamp to stdout

KentShikama: But as you say, it is ugly. I'm not sure how to fix.

birchmd: Hmm, interesting. I would expect that `Task.pure(t.unsafeRunSync) == t` would be a law about how `Task` behaves. I can look into this more closely to see if I can figure out what is going on.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -164,6 +165,23 @@ object Runtime {
       )
   }
 
+  class BlockTime[F[_]](val timestamp: Ref[F, Par]) {
```
ArturGajowy: Correct me if I'm wrong, but I think both `blockTime` and `shortLeashParams` must not change during a single deploy. I see 3 options for modelling this:

1. Each deploy gets its own Runtime (with parts of it possibly reused between deploys),
2. Runtime stays the same between deploys, so do system processes, but we have a, `DeployContext` (name tentative; all the things that change per deploy but not during it) passed to each `reducer.eval`,
3. We use Ref-s inside the `Runtime` and pray that the `set-s` are done only in the right places.

I'm really worried that we're using the weakest option - and not only for time and shortLeash, but for deploy phlo limit too (which is additionally conflated with deploy cost). What we should be doing probably is 2.

casper/src/test/scala/coop/rchain/casper/RholangBuildTest.scala
```diff
@@ -23,11 +23,15 @@ class RholangBuildTest extends FlatSpec with Matchers {
...
-      """new double, dprimes, rl(`rho:registry:lookup`), ListOpsCh in {
+      """new double, dprimes, rl(`rho:registry:lookup`), ListOpsCh, time(`rho:block:timestamp`), timeRtn, stdout(`rho:io:stdout`) in {
         |  contract double(@x, ret) = { ret!(2 * x) } |
         |  rl!(`rho:id:dputnspi15oxxnyymjrpu7rkaok3bjkiwq84z7cqcrx4ktqfpyapn4`, *ListOpsCh) |
         |  for(@(_, ListOps) <- ListOpsCh) {
         |    @ListOps!("map", [2, 3, 5, 7], *double, *dprimes)
+        |  } |
+        |  time!(*timeRtn) |
+        |  for (@timestamp <- timeRtn) {
+        |   @"timestampOutput"!(timestamp)
```
birchmd: Since we have been working on moving away from forgable names, I think this should be an unforgable name instead of a string.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
birchmd: I like the idea of a `DeployContext`, although I think the current scheme is safer than perhaps you are giving credit to. Because a deploy is a single `Par` we process it with a single call to `inj`, so in order to change the value of the parameters in the middle of a deploy you would need to actively try to do that by simultaneously running two threads: one for the `inj` call and one for the `set` call.

Regardless, I still think having an immutable `DeployContext` that gets threaded through everything on the Rholang side, and injected once on the Casper side is a better model. Perhaps we should create a ticket for this refactoring as well.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -221,16 +221,15 @@ object SystemProcesses {
...
-      }
+        _ <- space
+              .produce(
+                ack,
+                ListParWithRandom(Seq(codeHash, phloRate, userId, timestamp), rand),
+                false
+              )(MATCH_UNLIMITED_PHLOS)
+              .map(unpackOption(_))
+              .foldResult(dispatcher)
+      } yield ()
```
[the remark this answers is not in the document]
birchmd: Yes. You can see that `getDeployParams` used to have `space.produce` being `yield`ed, but that was wrong because `space.produce` returned a `Task[Unit]`. This is why Kent has refactored to pull it into the `for`-comprehension. This is exactly why we should turn on `-Ywarn-value-discard`, then Scala won't silently throw away values and this kind of thing will get caught by the compiler. We should probably make a ticket to turn that on and fix all the problems it finds.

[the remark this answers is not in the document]
birchmd: Created https://rchain.atlassian.net/browse/RHOL-1001

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
birchmd: Created https://rchain.atlassian.net/browse/RHOL-1003

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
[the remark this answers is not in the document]
guardbotmk3: That's because space.produce returned an Option when I wrote it.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -54,24 +47,40 @@ class RuntimeManager private (val emptyStateHash: ByteString, runtimeContainer:
...
+  def replayComputeState(
+      hash: StateHash,
+      terms: Seq[InternalProcessedDeploy],
+      time: Option[Long] = None
+  )(
       implicit scheduler: Scheduler
   ): Either[(Option[Deploy], Failed), StateHash] = {
     val runtime = runtimeContainer.take()
-    val result  = replayEval(terms, runtime, hash)
+    unsafeSetTimestamp(time, runtime)
```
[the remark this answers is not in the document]
KentShikama: Technically most of the work has been taken with https://rchain.atlassian.net/browse/CORE-1267 . I've put CORE-1267 as blocking RHOL-1002.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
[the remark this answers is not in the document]
KentShikama: @birchmd I had created https://rchain.atlassian.net/browse/RHOL-998 but I like your ticket description better

PR #1729, opened 2018-10-23
RHOL-936: Fix equality of SortedParMap

tymm:
## Overview
This is a fix to an issue discovered by a generative test in `RhoTypesTest.scala`.
The issue is that `SortedParMap` in certain situations is not `equal` to another `SortedParMap` with the same entries.
This is the case because `SortedParMap` is a `Map` and therefore uses `GenMapLike.equal` to decide on equality. `GenMapLike.equal` uses the iterator defined in `SortedParMap` that works on `SortedParMap.sortedMap` (because that is how we defined the iterator).
This however is problematic because we compare `sortedMap` with `ps` (the unsorted variant of `sortedMap`). `sortedMap` and `ps` can differ.

Also I would like to make `sortedMap` a `Map` but I'm unsure if there is a reason why it is actually a `List`.

https://rchain.atlassian.net/browse/RHOL-936

### DIFF
--- a/models/src/main/scala/coop/rchain/models/ParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/ParMap.scala
@@ -30,7 +30,7 @@ object ParMap {
       locallyFree: Coeval[BitSet],
       remainder: Option[Var]
   ) =
-    new ParMap(SortedParMap(seq), connectiveUsed, locallyFree.memoize, remainder)
+    new ParMap(new SortedParMap(seq.toMap), connectiveUsed, locallyFree.memoize, remainder)
 
   def apply(
       seq: Seq[(Par, Par)],
--- a/models/src/main/scala/coop/rchain/models/ParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/ParMap.scala
@@ -30,7 +30,7 @@ object ParMap {
       locallyFree: Coeval[BitSet],
       remainder: Option[Var]
   ) =
-    new ParMap(new SortedParMap(seq.toMap), connectiveUsed, locallyFree.memoize, remainder)
+    new ParMap(SortedParMap(seq), connectiveUsed, locallyFree.memoize, remainder)
 
   def apply(
       seq: Seq[(Par, Par)],
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -7,23 +7,17 @@ case class SortedParMap(private val ps: Map[Par, Par])
     extends Map[Par, Par]
     with MapLike[Par, Par, SortedParMap] {
 
-  lazy val sortedMap: List[(Par, Par)] = ps.sort
+  lazy val sortedMap: Map[Par, Par] = ps.sort.toMap
 
   override def empty: SortedParMap = SortedParMap.empty
 
-  override def get(key: Par): Option[Par] = sortedMap.toMap.get(key)
+  override def get(key: Par): Option[Par] = sortedMap.get(key)
 
   override def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
 
-  override def +[V1 >: Par](kv: (Par, V1)): Map[Par, V1] = sortedMap.toMap + kv
+  override def +[V1 >: Par](kv: (Par, V1)): Map[Par, V1] = sortedMap + kv
 
-  override def -(key: Par): SortedParMap = SortedParMap(sortedMap.toMap - key)
-
-  override def equals(that: Any): Boolean =
-    that match {
-      case _: SortedParMap => this.sortedMap == that.asInstanceOf[SortedParMap].sortedMap
-      case _               => false
-    }
+  override def -(key: Par): SortedParMap = SortedParMap(sortedMap - key)
 }
 
 object SortedParMap {
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -11,14 +11,19 @@ case class SortedParMap(private val ps: Map[Par, Par])
 
   override def empty: SortedParMap = SortedParMap.empty
 
-  override def get(key: Par): Option[Par] = ps.get(key)
+  override def get(key: Par): Option[Par] = sortedMap.toMap.get(key)
 
   override def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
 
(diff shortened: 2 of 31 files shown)


models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -11,14 +11,19 @@ case class SortedParMap(private val ps: Map[Par, Par])
 
   override def empty: SortedParMap = SortedParMap.empty
 
-  override def get(key: Par): Option[Par] = ps.get(key)
+  override def get(key: Par): Option[Par] = sortedMap.toMap.get(key)
```
ArturGajowy: Let's just make `sortedMap: SortedMap[Par, Par]` instead of incurring the `toMap` overhead on every `get`.

ArturGajowy: Ideally, the commit message would also mention that there was a possibility of accessing an un-sorted par via `get`/`apply` due to get relying on the unsorted `ps`. Which is actually the reason why the default equals implementation would say 'false'. Won't push on the commit message change though.

ArturGajowy: Please though add a test securing the behavior of the `get` method as well. So the test should say 'when I lookup a value of the SortedMap using get, the returned value should be sorted'

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -11,14 +11,19 @@ case class SortedParMap(private val ps: Map[Par, Par])
...
   override def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
 
   override def +[V1 >: Par](kv: (Par, V1)): Map[Par, V1] = sortedMap.toMap + kv
 
-  override def -(key: Par): SortedParMap = SortedParMap(ps - key)
+  override def -(key: Par): SortedParMap = SortedParMap(sortedMap.toMap - key)
 
+  override def equals(that: Any): Boolean =
+    that match {
+      case _: SortedParMap => this.sortedMap == that.asInstanceOf[SortedParMap].sortedMap
```
ArturGajowy: IIRC this should not be needed now, as `get` now returns sorted terms. Equally so after we make it a `SortedMap`. Please drop this method.

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
```diff
@@ -61,4 +64,24 @@ class SortedParMapSpec extends FlatSpec with Matchers {
...
+        Par(
+          List(),
+          List(),
+          List(),
+          List(),
+          List(),
+          List(GPrivate(ByteString.copyFrom(Array[Byte](0))), GPrivate(ByteString.EMPTY)),
+          List(),
+          List(),
+          AlwaysEqual(BitSet()),
```
ArturGajowy: Please replace this with the output of `Pretty.pretty(ps)`, which will output a less noisy version with named arguments, and will only include non-default values for each field.

ArturGajowy:
I've just noticed that the same error is repeated in SortedParHashSet - please either fix it in this PR or open a JIRA for it

ArturGajowy:
Also, neither of the SortedPar* classes should be case classes. The only thing we use the `case` for them is the `apply()` method in companion, and they really aren't defined by their field's contents in a straightforward way...

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
tymm: Did that by using `assertEqual`.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done. Should have seen that myself...

tymm:
I'll create a JIRA ticket for the issue in `SortedParHashSet`.

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
ArturGajowy: That is great! :) But what I had in mind is replacing this code - so that it's easier to read, maintain and modify in the future :)

models/src/main/scala/coop/rchain/models/ParMap.scala
```diff
@@ -30,7 +30,7 @@ object ParMap {
       locallyFree: Coeval[BitSet],
       remainder: Option[Var]
   ) =
-    new ParMap(SortedParMap(seq), connectiveUsed, locallyFree.memoize, remainder)
+    new ParMap(new SortedParMap(seq.toMap), connectiveUsed, locallyFree.memoize, remainder)
```
ArturGajowy: Oh, I thought you'd add the `apply()` method(s). It's OK, but please fix the Pretty instance for ParMap so that it still outputs compiling code (add `new `).

I'm sorry that I didn't leave a test that would tell you that earlier - it'd require compiling Pretty's output (I'd like to do it sometime in the future, simply had no time)

tymm: Sorry, added `apply` methods now.

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
tymm: Ouch, totally misread it.

goral09:
I don't understand this change. If the equality was broken then it would be broken for both values of the comparison (in `sortedMap1 == sortedMap2` both sorted maps should use the "broken" equality). You say:
> This however is problematic because we compare `sortedMap` with `ps` (the unsorted variant of sortedMap). `sortedMap` and `ps` can differ.

When/where are we comparing `sortedMap` and `ps` ? If that's the case then the we shouldn't be doing that. 

> This is the case because `SortedParMap` is a Map and therefore uses `GenMapLike.equal` to decide on equality. `GenMapLike.equal` uses the iterator defined in `SortedParMap` that works on `SortedParMap.sortedMap` (because that is how we defined the `iterator`).

I don't understand this part either. Even if underlying implementation of `equals` uses the iterator that iterators on the sorted pars this shouldn't affect the test because sorting MUSTN'T modify the content of the `value` stored under the `key` the order should not matter for equality. If the sorting modifies the underlying map then the sorting is broken.

I don't like the `MapLike` interface for the `SortedParMap` anymore because it creates broken constraints - `+` method returns `Map[Par, Par]` and `-` returns `SortedParMap`. Removing that dependency should lead to safer API. 

Again, I still don't understand why the changes are making the tests pass.

PR #1730, opened 2018-10-24
RHOL-1000: Make SortedParHashSet cleaner and more robust

tymm:
## Overview
This is inspired by a bug discovered in RHOL-936.

https://rchain.atlassian.net/browse/RHOL-1000

### DIFF
--- a/models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
@@ -40,7 +40,7 @@ final class SortedParHashSet(ps: HashSet[Par]) extends Iterable[Par] {
 object SortedParHashSet {
   def apply(seq: Seq[Par]): SortedParHashSet = new SortedParHashSet(HashSet[Par](seq: _*))
 
-  def apply(set: Set[Par]): SortedParHashSet = new SortedParHashSet(HashSet[Par](set.toSeq: _*))
+  def apply(set: Set[Par]): SortedParHashSet = SortedParHashSet(set.toSeq)
 
   def empty: SortedParHashSet = SortedParHashSet(HashSet.empty[Par])
 }
--- a/models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
@@ -1,41 +1,32 @@
 package coop.rchain.models
 
-import scala.collection.generic.CanBuildFrom
-import scala.collection.immutable.HashSet
-import scala.collection.{mutable, SetLike}
 import coop.rchain.models.rholang.sorter.ordering._
+
+import scala.collection.SetLike
+import scala.collection.immutable.HashSet
 
 //Enforce ordering and uniqueness.
 // - uniqueness is handled by using HashSet.
 // - ordering comes from sorting the elements prior to serializing.
-case class SortedParHashSet(private val ps: HashSet[Par])
+class SortedParHashSet(private val ps: HashSet[Par])
     extends Set[Par]
     with SetLike[Par, SortedParHashSet] {
 
   lazy val sortedPars: List[Par] = ps.toList.sort
 
-  override def equals(that: Any): Boolean = that match {
-    case shs: SortedParHashSet => shs.sortedPars == this.sortedPars
-    case _                     => false
-  }
+  override def empty: SortedParHashSet = new SortedParHashSet(HashSet.empty[Par])
 
-  override def hashCode(): Int = sortedPars.hashCode()
+  override def contains(elem: Par): Boolean = sortedPars.contains(elem)
 
-  override def empty: SortedParHashSet = SortedParHashSet.empty
+  override def +(elem: Par): SortedParHashSet = new SortedParHashSet(ps + elem)
 
-  override def contains(elem: Par): Boolean = ps.contains(elem)
-
-  override def +(elem: Par): SortedParHashSet = SortedParHashSet(ps + elem)
-
-  override def -(elem: Par): SortedParHashSet = SortedParHashSet(ps - elem)
+  override def -(elem: Par): SortedParHashSet = new SortedParHashSet(ps - elem)
 
   override def iterator: Iterator[Par] = sortedPars.toIterator
-
 }
 
 object SortedParHashSet {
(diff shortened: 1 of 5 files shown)


models/src/test/scala/coop/rchain/models/SortedParHashSetSpec.scala
```diff
@@ -82,4 +79,20 @@ class SortedParHashSetSpec extends FlatSpec with Matchers {
...
+            ByteString.copyFrom(Array[Byte](
+              0
+            ))
+          ),
+          GPrivate()
+        ),
+        connectiveUsed = true
+      )
+    )
+    assertEqual(SortedParHashSet(elements), SortedParHashSet(elements))
```
goral09: Would this test still be passing without the changes you made?

tymm: Yes, it would still be passing because `SortedParHashSet` was overriding `equals` while `SortedParSet` didn't. The problem is that if someone would remove the overridden `equals` we would get the same hard-to-find bug as with `SortedParSet`.

models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
```diff
@@ -1,41 +1,32 @@
...
 
-  override def empty: SortedParHashSet = SortedParHashSet.empty
+  override def +(elem: Par): SortedParHashSet = new SortedParHashSet(ps + elem)
 
-  override def contains(elem: Par): Boolean = ps.contains(elem)
-
-  override def +(elem: Par): SortedParHashSet = SortedParHashSet(ps + elem)
-
-  override def -(elem: Par): SortedParHashSet = SortedParHashSet(ps - elem)
+  override def -(elem: Par): SortedParHashSet = new SortedParHashSet(ps - elem)
```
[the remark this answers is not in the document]
ArturGajowy: There's another bug lurking in here. Both current and the `sortedPs`-based implementations have  the problem, that they can only remove an element with the same sorted-ness (and exact same unsorted shape if we're talking about the current impl).

We should be using `sortedPs` and sort the input `elem: Par` before attempting to remove it. This will work whether `elem` was originally sorted or not.

This ofc is a problem for the Map too... But let's fix that in a follow-up PR after this and #1729 gets merged.

models/src/test/scala/coop/rchain/models/SortedParHashSetSpec.scala
```diff
@@ -82,4 +79,20 @@ class SortedParHashSetSpec extends FlatSpec with Matchers {
     shs.sortedPars should contain theSameElementsAs (expected)
   }
 
+  it should "be equal when it is equal" in {
```
ArturGajowy: Please add a generative test in the spirit of what I've pasted for Map:

```scala
  it should "preserve sortedness of all elements and the whole map for all its operations" in {
    import coop.rchain.models.testImplicits._

    forAll { (pars1: Seq[Par], pars2: Seq[Par]) =>
      val map1 = SortedParMap(pars1.zip(pars2))
      val map2 = SortedParMap(pars2.zip(pars1))
      checkSorted(map1)
      checkSorted(map2)
      Seq(map1.headOption, map2.headOption).flatten.foreach(kv => {
        checkSorted(map1 + kv)
        checkSorted(map2 - kv._1)
        checkSorted(map2 + kv)
        checkSorted(map1 - kv._1)
      })
    }
  }

  private def checkSorted(sortedParMap: Map[Par, Par]): Unit = {
    checkSorted(sortedParMap.keys)
    sortedParMap.values.foreach(p => assert(isSorted(p)))
    sortedParMap.foreach(kv => {
      assert(isSorted(kv._1))
      assert(isSorted(kv._2))
      assert(isSorted(sortedParMap(kv._1)))
    })
  }

  private def checkSorted(iterable: Iterable[Par]) = {
    import coop.rchain.models.rholang.sorter.ordering._
    iterable.foreach(p => assert(isSorted(p)))
    assert(iterable.toList == iterable.toList.sort)
  }

  def isSorted[A: Sortable](a: A): Boolean = {
    a == Sortable[A].sortMatch[Coeval](a).value().term
  }
```

ArturGajowy:
@tymm do we have any progress on that?

tymm:
@ArturGajowy I'm currently trying to make your generative test pass for `SortedParMap` and then I'm back to this.

models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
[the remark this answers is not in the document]
tymm: How would the lazy init blow up with an NPE?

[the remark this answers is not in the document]
ArturGajowy: I think I've mixed val initialization order problems and implicit initialization problems with `lazy val` (which seems to be brought up more often as a solution than a problem). Scala is "funny" language ;/... Please disregard the mentioned comment :)

models/src/main/scala/coop/rchain/models/SortedParHashSet.scala
```diff
@@ -1,41 +1,46 @@
...
+
+  private def sort(par: Par): Par = Sortable[Par].sortMatch[Coeval](par).map(_.term).value()
 }
 
 object SortedParHashSet {
-  def apply(seq: Seq[Par]): SortedParHashSet = SortedParHashSet(HashSet[Par](seq: _*))
-  def apply(): SortedParHashSet              = SortedParHashSet(HashSet.empty[Par])
+  def apply(seq: Seq[Par]): SortedParHashSet = new SortedParHashSet(HashSet[Par](seq: _*))
+
+  def apply(set: Set[Par]): SortedParHashSet = new SortedParHashSet(HashSet[Par](set.toSeq: _*))
```
ArturGajowy: That's a total nit and feel free to omit that, but:

ideally, since we have the `apply` method variants on the companion, we'd only call the `new SortedParHashSet(...)` constructor once in the whole codebase. Each of the apply methods would either delegate to some other apply method variant, or would otherwise be the single method that uses `new`. I'm bringing this up only because I've seen bugs lurking in different possible initialization paths in my career, and the pattern I'm suggesting prevents that by delegating everything that's possible to the other factory methods.

PR #1731, opened 2018-10-24
Stacksafe (de)seralization of protobuffs

ArturGajowy:
## Overview
Title says it all :)

The `serializedSize` and `writeTo` methods were implemented as standalone interpreters over the generated proto classes (see the ProtoM class).

The `mergeFrom` method (used for reading) was implemented by modifying the scalapb code generator, because the interpreter approach was not viable (I tried hard, believe me).

See 8c8096d for what needed to be changed in the scalapb generator for `mergeFrom`.

### JIRA issue:
RHOL-854

### Checklist

### DIFF
--- a/models/src/main/scala/coop/rchain/models/Memo.scala
+++ b/models/src/main/scala/coop/rchain/models/Memo.scala
@@ -16,7 +16,7 @@ class Memo[A](f: => Coeval[A]) {
             case e: Eager[A] => e
             case _ =>
               thunk.map { r =>
-                thunk = null
+                thunk = null //allow GC-ing the thunk
                 result = Coeval.now(r)
                 r
               }
--- a/models/src/main/scala/coop/rchain/models/Memo.scala
+++ b/models/src/main/scala/coop/rchain/models/Memo.scala
@@ -1,0 +1,26 @@
+package coop.rchain.models
+import monix.eval.Coeval
+import monix.eval.Coeval.Eager
+
+class Memo[A](f: => Coeval[A]) {
+
+  private[this] var thunk             = f
+  private[this] var result: Coeval[A] = _
+
+  def get: Coeval[A] =
+    result match {
+      case e: Eager[A] => e
+      case _ =>
+        Coeval.defer {
+          result match {
+            case e: Eager[A] => e
+            case _ =>
+              thunk.map { r =>
+                thunk = null
+                result = Coeval.now(r)
+                r
+              }
+          }
+        }
+    }
+}
--- a/models/src/main/scala/coop/rchain/models/Pretty.scala
+++ b/models/src/main/scala/coop/rchain/models/Pretty.scala
@@ -1,7 +1,8 @@
 package coop.rchain.models
+import java.io.{PrintWriter, StringWriter}
+
 import com.google.protobuf.ByteString
 import monix.eval.Coeval
-import monix.eval.Coeval.Now
 
 import scala.annotation.switch
 import scala.collection.immutable.{BitSet, HashSet}
@@ -38,7 +39,11 @@ trait PrettyInstances extends PrettyDerivation {
 
   // obviously won't print compiling code,
   // but seeing the stacktrace is more important in this case
-  implicit val PrettyThrowable = fromToString[Throwable]
+  implicit val PrettyThrowable: Pretty[Throwable] = (value: Throwable, indentLevel: Int) => {
+    val stackTrace = new StringWriter()
+    value.printStackTrace(new PrintWriter(stackTrace))
+    "\n" + stackTrace.toString
+  }
 
   implicit val PrettyByteString: Pretty[ByteString] = (value: ByteString, indentLevel: Int) =>
     "ByteString.copyFrom(Array[Byte]" + parenthesised(value.toByteArray, indentLevel) + ")"
--- a/models/src/main/scala/coop/rchain/models/ProtoM.scala
+++ b/models/src/main/scala/coop/rchain/models/ProtoM.scala
@@ -1,15 +1,139 @@
 package coop.rchain.models
-import cats.{Applicative, Monad}
 import cats.effect.Sync
 import cats.implicits._
+import cats.{Applicative, Monad}
 import com.google.protobuf.Descriptors.FieldDescriptor
 import com.google.protobuf.{ByteString, CodedOutputStream, Descriptors, MessageLite}
 import scalapb.GeneratedMessage
 import scalapb.compiler.{DescriptorPimps, GeneratorParams, Types}
+import scalapb.{GeneratedMessage, WireType}
 
 object ProtoM extends DescriptorPimps {
 
   def params: GeneratorParams = ??? //required by DescriptorPimps, but we don't use it transitively
+
+  def toByteArray[M[_]: Sync](message: GeneratedMessage): M[Array[Byte]] =
+    for {
+      size  <- ProtoM.serializedSize[M](message)
+      array = new Array[Byte](size)
+      out   = CodedOutputStream.newInstance(array)
+      _     <- ProtoM.writeTo[M](out, message)
+      _     <- Sync[M].catchNonFatal { out.checkNoSpaceLeft() }
+    } yield array
+
+  def writeTo[M[_]: Sync](
+      out: CodedOutputStream,
+      message: GeneratedMessage
+  ): M[Unit] = {
+    val companion       = message.companion
+    val descriptor      = companion.javaDescriptor
+    val defaultInstance = companion.defaultInstance.asInstanceOf[GeneratedMessage]
+    for {
+      _ <- raiseUnsupportedIf(descriptor.preservesUnknownFields, "Unknown fields are not supported")
+      _ <- descriptor.fields
+            .sortBy(_.getNumber)
+            .toList
+            .traverse[M, Unit](f => {
+              val fieldValue = message.getField(f)
+              val default    = defaultInstance.getField(f)
+              if (fieldValue != default)
+                writeField(out, fieldValue, f)
+              else Monad[M].pure(())
+            })
+    } yield ()
+  }
+
+  private def writeField[M[_]: Sync](
+      out: CodedOutputStream,
+      value: Any,
+      field: FieldDescriptor
+  ): M[Unit] =
+    if (field.isRepeated) {
+      writeRepeatedField(out, value, field)
+    } else if (field.isRequired || field.isSingular || field.isOptional) {
+      writeSingleField(out, value, field)
+    } else {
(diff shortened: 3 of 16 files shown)


ArturGajowy:
See https://github.com/rchain/rchain/pull/1731/commits/8c8096d5fd473684f486d06923c6afd6a4e68ab4 for what needed to be changed in the scalapb generator for mergeFrom

models/src/test/scala/coop/rchain/models/RhoTypesTest.scala
```diff
@@ -45,23 +47,50 @@ class RhoTypesTest extends FlatSpec with PropertyChecks with Matchers {
...
+    assert(ProtoM.toByteArray(a).value sameElements a.toByteArray)
+
+  def stacksafeReadFromSameAsReference[A <: StacksafeMessage[A]](a: A)(
+      implicit companion: GeneratedMessageCompanion[A]
+  ): Assertion = {
+    // We don't want to rely on the (sometimes overridden) equals,
+    // so instead we compare the output of reference serializer
+    // on the messages parsed by the stacksafe deserializer.
+    // In short we check `referenceSerialize === referenceSerialize . stacksafeDeserialize . referenceSerialize`.
+    // We also check check `stacksafeDeserialize(referenceSerialize(a)).equals(a)` just in case :)
```
tymm: *check check*

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
```diff
@@ -71,11 +90,17 @@ class StackSafetySpec extends FlatSpec with Matchers {
...
 
+  //FIXME java.lang.OutOfMemoryError: GC overhead limit exceeded
+  //it should "handle a huge nested set" in {
+  //  checkAll(hugeNested("Set(", "42", ")"))
+  //}
+
+  //FIXME 11.5 s
   it should "handle a nested new" in {
-    checkNormalize(hugeNested("new x in { ", "1", "}"))
+    checkAll(hugeNested("new x in { ", "1", "}"))
```
tymm: This took 23 seconds on my machine.
Just wondering if it shouldn't actually fail because of `val maxDuration = 10.seconds`?

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
```diff
@@ -71,11 +90,17 @@ class StackSafetySpec extends FlatSpec with Matchers {
   }
 
   it should "handle a huge nested list" in {
-    checkNormalize(hugeNested("[", "", "]"))
+    checkAll(hugeNested("[", "", "]"))
   }
 
+  //FIXME java.lang.OutOfMemoryError: GC overhead limit exceeded
```
tymm: Do you think this could be related to the way you calculate `depth`?

models/src/main/scala/coop/rchain/models/Memo.scala
```diff
@@ -0,0 +1,26 @@
...
+  def get: Coeval[A] =
+    result match {
+      case e: Eager[A] => e
+      case _ =>
+        Coeval.defer {
+          result match {
+            case e: Eager[A] => e
+            case _ =>
+              thunk.map { r =>
+                thunk = null
```
goral09: Why are you setting `thunk = null` ? I don't see you're making use of this fact anywhere.

models/src/main/scala/coop/rchain/models/Pretty.scala
```diff
@@ -38,7 +39,11 @@ trait PrettyInstances extends PrettyDerivation {
 
   // obviously won't print compiling code,
   // but seeing the stacktrace is more important in this case
-  implicit val PrettyThrowable = fromToString[Throwable]
+  implicit val PrettyThrowable: Pretty[Throwable] = (value: Throwable, indentLevel: Int) => {
```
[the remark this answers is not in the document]
ArturGajowy: Why infamous? Yes, that's it. What's wrong with it?

models/src/main/scala/coop/rchain/models/Memo.scala
ArturGajowy: GC. I'll add a comment :)

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
ArturGajowy: `checkAll` consists of: `buildNormalizedTerm`, pretty-printing, and doing a full reduce from rholang sources (which includes buildNormalizedTerm again). Only the last part is under the timeout.

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
ArturGajowy: It's been there way before I had depth calculation in place. I suppose this is due to each level of the set normalizing everything that's being put inside, resulting in quadratic waste.

models/src/main/scala/coop/rchain/models/ProtoM.scala
```diff
@@ -0,0 +1,113 @@
...
+      repeatedSize(value, field)
+    } else if (field.isRequired || field.isSingular || field.isOptional) {
+      singleFieldSize(value, field)
+    } else {
+      Sync[M].raiseError(new RuntimeException("This cannot be!"))
+    }
+
+  private def repeatedSize[M[_]: Sync](value: Any, field: Descriptors.FieldDescriptor): M[Int] =
+    for {
+      _ <- raiseUnsupportedIf[M](field.isPacked, "Packed fields are unsupported")
```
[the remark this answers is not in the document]
ArturGajowy: yup!

> Only repeated fields of primitive numeric types (types which use the varint, 32-bit, or 64-bit wire types) can be declared "packed".

(https://developers.google.com/protocol-buffers/docs/encoding#packed)

I didn't check on that until now I think - I simply followed the generator code for this part and translated it step-by-step, including any 'unsupported' errors whenever a given unneeded feature was used.

models/src/main/scala/coop/rchain/models/serialization/implicits.scala
```diff
@@ -8,24 +8,27 @@ import scodec.bits.ByteVector
...
-  implicit val serializeEMap: Serialize[EMap]       = mkProtobufInstance(EMap)
+  implicit val serializePar: Serialize[Par]         = mkProtobufInstance
+  implicit val serializeVar: Serialize[Var]         = mkProtobufInstance
+  implicit val serializeSend: Serialize[Send]       = mkProtobufInstance
+  implicit val serializeReceive: Serialize[Receive] = mkProtobufInstance
+  implicit val serializeNew: Serialize[New]         = mkProtobufInstance
+  implicit val serializeExpr: Serialize[Expr]       = mkProtobufInstance
+  implicit val serializeMatch: Serialize[Match]     = mkProtobufInstance
+  implicit val serializeESet: Serialize[ESet]       = mkProtobufInstance
+  implicit val serializeEMap: Serialize[EMap]       = mkProtobufInstance
```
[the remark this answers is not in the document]
ArturGajowy: You're right that we could make `mkProtobufInstance` an `implicit def`. Having the `implicit val`-s instantiated by hand has the advantage of caching the instances between usages (and thus reducing compile time), but given their small overhead (all the 'derivations' are constant-time) I think let's go with the implicit def.

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
ArturGajowy: I've just seen 26 s on that, so it's not just your machine - it might be there's been some overhead introduced by the stacksafety. We won't fight it for now though.

rholang/src/test/scala/coop/rchain/rholang/StackSafetySpec.scala
```diff
@@ -155,13 +161,20 @@ class StackSafetySpec extends FlatSpec with Matchers {
 
   private def checkReduce(rho: String): Unit =
     isolateStackOverflow {
+      //FIXME make this pass:
+      //val reduceRho = s"@0!($rho) | for (@x <- @0) { @1 ! (x) | @2 ! (x)  } | for (@y <- @1; @z <- @2) { @0!(Set(y, z)) }"
+      //val reduceRho = s"""@0!( { @"serializeMe"!($rho) } )"""
+      //val reduceRho = s"""for (_ <- @0) { Nil } | @0!( { @"serializeMe"!($rho) } )"""
+      //val reduceRho = s"@0!($rho)"
```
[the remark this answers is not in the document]
ArturGajowy: for various reasons, some of them being non-safe equals and hashcode being used underneath. I'm going to address as many of those as I can in my next PR.

models/src/main/scala/coop/rchain/models/ProtoM.scala
```diff
@@ -1,16 +1,140 @@
...
 
   def params: GeneratorParams = ??? //required by DescriptorPimps, but we don't use it transitively
 
+  def toByteArray[M[_]: Sync](message: GeneratedMessage): M[Array[Byte]] =
+    for {
+      size  <- ProtoM.serializedSize[M](message)
+      array = new Array[Byte](size)
+      out   = CodedOutputStream.newInstance(array)
+      _     <- ProtoM.writeTo[M](out, message)
+      _     <- Sync[M].catchNonFatal { out.checkNoSpaceLeft() }
```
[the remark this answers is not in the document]
ArturGajowy: Funny thing, it hasn't a Close method, nor does it extend OutputStream. I'll `checkNoSpaceLeft` thoguh, as the original impl does.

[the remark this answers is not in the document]
ArturGajowy: oh wait, I'm doing it already :D

models/src/test/scala/coop/rchain/models/MemoSpec.scala
```diff
@@ -0,0 +1,22 @@
...
+    val random = new Memo[Int](Coeval.delay {
+      timesExecuted += 1
+      9
+    })
+    assert(timesExecuted == 0)
+    assert(random.get.value() == 9)
+    assert(timesExecuted == 1)
+    assert(random.get.value() == 9)
+    assert(timesExecuted == 1)
+  }
```
goral09: How does this test compare with plain `Coeval.memoize` ?

ArturGajowy: Coeval.memoize is not stacksafe. I'd rather not dive into testing that - we've got other tests that will catch it.

goral09:
:tada: :champagne:

PR #1736, opened 2018-10-25
Implement property tests for CostAccounting

odeac:
## Overview
see title

https://rchain.atlassian.net/browse/RHOL-949

### Notes

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
@@ -31,6 +31,8 @@ class CostAccountingPropertyTest extends FlatSpec with PropertyChecks with Match
     ProcGen.topLevelGen(5).map(PrettyPrinted[Proc](_, PrettyPrinter.print))
   )
 
+  implicit val taskExecutionDuration = 5.seconds
+
   def cost(proc: Proc): Cost = Cost(Interpreter.buildPar(proc).apply)
 
   behavior of "Cost accounting in Reducer"
@@ -45,30 +47,27 @@ class CostAccountingPropertyTest extends FlatSpec with PropertyChecks with Match
 
   it should "charge the same if the evaluation order is reversed" in {
     forAll { (p: PrettyPrinted[Proc], q: PrettyPrinted[Proc]) =>
-      val task =
-        for {
-          cost1 <- costOfExecution(p.value, q.value)
-          cost2 <- costOfExecution(q.value, p.value)
-        } yield (cost1 == cost2)
-
-      task.runSyncUnsafe(5.seconds)
+      haveEqualResults(costOfExecution(p.value, q.value), costOfExecution(q.value, p.value))
     }
   }
 
   it should "sequential execution and parallel execution have the same cost" in {
     forAll { (p: PrettyPrinted[Proc], q: PrettyPrinted[Proc]) =>
-      val task =
-        for {
-          cost1 <- costOfExecution(new PPar(p.value, q.value))
-          cost2 <- costOfExecution(q.value, p.value)
-        } yield (cost1 == cost2)
-
-      task.runSyncUnsafe(5.seconds)
+      haveEqualResults(
+        costOfExecution(new PPar(p.value, q.value)),
+        costOfExecution(q.value, p.value)
+      )
     }
   }
 }
 
 object CostAccountingPropertyTest {
+  def haveEqualResults[A](tasks: Task[A]*)(implicit duration: Duration): Boolean =
+    tasks.toList
+      .sequence[Task, A]
+      .map { _.sliding(2).forall { case List(r1, r2) => r1 == r2 } }
+      .runSyncUnsafe(duration)
+
   def createRhoISpace(): RhoISpace = {
     implicit val syncF: Sync[Id] = coop.rchain.catscontrib.effect.implicits.syncId
     import coop.rchain.rholang.interpreter.storage.implicits._
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
@@ -1,14 +1,31 @@
 package coop.rchain.rholang.interpreter.accounting
+import java.nio.file.Files
+
+import cats.Id
+import cats.effect.Sync
+import coop.rchain.crypto.hash.Blake2b512Random
+import cats.implicits._
+import coop.rchain.models._
+import coop.rchain.rholang.interpreter.Runtime.{RhoContext, RhoISpace}
+import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 import coop.rchain.rholang.{PrettyPrinted, ProcGen}
-import coop.rchain.rholang.interpreter.Interpreter
-import coop.rchain.rholang.syntax.rholang_mercury.Absyn.Proc
+import coop.rchain.rholang.interpreter._
+import coop.rchain.rholang.syntax.rholang_mercury.Absyn.{PPar, Proc}
 import coop.rchain.rholang.syntax.rholang_mercury.PrettyPrinter
+import coop.rchain.rspace.history.Branch
+import coop.rchain.rspace.{Context, RSpace}
+import monix.eval.Task
 import org.scalacheck.Arbitrary
 import org.scalacheck.Test.Parameters
 import org.scalatest.{FlatSpec, Matchers}
 import org.scalatest.prop.PropertyChecks
 
+import scala.concurrent.duration._
+import monix.execution.Scheduler.Implicits.global
+
 class CostAccountingPropertyTest extends FlatSpec with PropertyChecks with Matchers {
+  import CostAccountingPropertyTest._
+
   implicit val params: Parameters = Parameters.defaultVerbose.withMinSuccessfulTests(1000)
   implicit val procArbitrary = Arbitrary(
     ProcGen.topLevelGen(5).map(PrettyPrinted[Proc](_, PrettyPrinter.print))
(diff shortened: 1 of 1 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
+      ListParWithRandomAndPhlos,
+      TaggedContinuation
+    ](context, Branch("test"))
+    space
+  }
+
+  def execute(reducer: ChargingReducer[Task], p: Proc)(
+      implicit rand: Blake2b512Random
+  ): Task[Long] = {
+    val program = Interpreter.buildPar(p).apply
```
goral09: I was thinking - `implicitly[Gen[Proc]].map(proc => Interpreter.buildPar(proc).getOrElse(throw new ...))` ? This way we have generators of `Par`s for tests.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
 
+import scala.concurrent.duration._
+import monix.execution.Scheduler.Implicits.global
+
 class CostAccountingPropertyTest extends FlatSpec with PropertyChecks with Matchers {
+  import CostAccountingPropertyTest._
+
   implicit val params: Parameters = Parameters.defaultVerbose.withMinSuccessfulTests(1000)
   implicit val procArbitrary = Arbitrary(
     ProcGen.topLevelGen(5).map(PrettyPrinted[Proc](_, PrettyPrinter.print))
```
ArturGajowy: What happens for bigger levels? Our AST is nested enough that in 5 hops we won't even be able to reach some of the productions...

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
+      )
+    }
   }
 }
+
+object CostAccountingPropertyTest {
+  def haveEqualResults[A](tasks: Task[A]*)(implicit duration: Duration): Boolean =
+    tasks.toList
+      .sequence[Task, A]
+      .map { _.sliding(2).forall { case List(r1, r2) => r1 == r2 } }
```
ArturGajowy: a foldLeft (or right) would be clearer here for me

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
+  ): Task[Long] = {
+    val program = Interpreter.buildPar(p).apply
+
+    val initPhlos = Cost(accounting.MAX_VALUE)
+
+    for {
+      _         <- reducer.setAvailablePhlos(initPhlos)
+      _         <- reducer.inj(program)
+      phlosLeft <- reducer.getAvailablePhlos()
+    } yield (initPhlos - phlosLeft.cost).value
```
ArturGajowy: Why can't we reuse Interpreter.evaluate in here? I see two possible reasons, both questionable:
1. it does checkpoints - just extract the checkpointing away
2. it checks for evaluation error in the `errorVector` - which we should probably do too in this test? I mean: we want the tested scenarios to error at the same times, and there should probably be no errors in the first place.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
+    val space: RhoISpace = RSpace.create[
+      Id,
+      Par,
+      BindPattern,
+      OutOfPhlogistonsError.type,
+      ListParWithRandom,
+      ListParWithRandomAndPhlos,
+      TaggedContinuation
+    ](context, Branch("test"))
+    space
```
ArturGajowy: This seems like an n-th copy sth needed in a lot of tests. If it's the case, please try to deduplicate with the source copy at least. And file a ticket for deduplicating this across all the tests.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -1,27 +1,115 @@
...
+  implicit val taskExecutionDuration = 5.seconds
+
   def cost(proc: Proc): Cost = Cost(Interpreter.buildPar(proc).apply)
 
   behavior of "Cost accounting in Reducer"
 
-  it should "cost of substitution is never negative" in {
+  it should "have positive cost of substitution" in {
     forAll { p: PrettyPrinted[Proc] =>
       cost(p.value).value should be >= 0L
```
ArturGajowy: I think we should have some test saying that overall cost is indeed `> 0` for every PProc possible (even the Nil one shouldn't be for free)

We should also think of other properties that we could test about the cost - some more advanced and closer to our particular domain.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: true!

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: I agree that we need more tests. I'll talk to Mateusz when he is back. Also if you or anybody has ideas about tests that might be relevant here, please shoot.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: indeed.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: The current implementation has a unlimited branching factor so for bigger heights the chances are the AST grows way too big and sometimes getting an OutOfMemory error.

I'm thinking of adding a "maxBranchFactor" parameter to the generator and that would allow us to better control the shape of the AST.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: yes, we can certainly build Par generators on top of the Proc generator

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: actually, when I tried to implement it I realised that the foldLeft approach is more complex and less readable

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: I used the example from CostAccountingReducerTest but, indeed, it would be better to use Interpreter.evaluate directly

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: I created this https://rchain.atlassian.net/browse/RHOL-1006

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
goral09: We should have a JIRA for that and I think I have created one that should capture the work: [RHOL-979](https://rchain.atlassian.net/browse/RHOL-979).

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
goral09: `CostAccountingReducerTest` needs to control the init phlos limit so it doesn't reuse the `interpreter.evaluate`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
goral09: There is not much more properties that we can test. From the high level view we have `p` and `q` processes and respective `cost(p)` and `cost(q)` . You can  test:

- `cost(p | q) == cost(p) + cost(q)`
- `cost(p) + cost(q) == cost(q) + cost(p)`
- `cost(p | Nil) == cost(p)`

I can't think of any more tests.

goral09:
All these tests will fail if you generate contracts that interact with each other. You should exclude names generated in `p` from the names generated in `q`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
ArturGajowy: Then evaluate should have an overload taking the limit as parameter :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
ArturGajowy: I think one more property that should hold is:

`cost(p(q1, q2, .., qn)) > cost(q1) + ... + cost(qn)`

where `p(q1, q2, .., qn)` means `q...` are all the children of `p`.

I realize though testing this might be a separete jira.

@goral09 please let us know if you think the property makes sense :)

goral09: And what this `parent`-`child` relation implies? I don't understand it.

ArturGajowy: E.g. in `EPlus(GInt(1), GInt(2))` the GInt-s are children of EPlus. Or in 
`for (x <- z)  P` the children are `z` and `P`.

goral09: It would test that cost of `p` is non-zero. Which is not true for in all the cases and in those which it is we already test it elsewhere.

ArturGajowy: Allowing zero cost for any `p` is dangerous, becuase then I can stuff a lot of those into a container that doesn't add its own cost, like [p, p, p, p]... and troll the chain.

ArturGajowy: `we already test it elsewhere` - not with generative tests though?

goral09: > Allowing zero cost for any `p` is dangerous, becuase then I can stuff a lot of those into a container that doesn't add its own cost, like [p, p, p, p]... and troll the chain.

The cost of calling contracts defined in `SystemProcesses` is zero: `std[Out|Err](Ack)` and crypto methods. Although I think we should discuss again the rationale behind allowing for crypto methods be free. I will raise that issue.

goral09: > `we already test it elsewhere` - not with generative tests though?

No, but since some of the methods are free those tests will fail.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: let's do this as a future refactoring step https://rchain.atlassian.net/browse/RHOL-1018

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
odeac: I also created this JIRA https://rchain.atlassian.net/browse/RHOL-1019

goral09:
Do not ingore my comment https://github.com/rchain/rchain/pull/1736#issuecomment-433606279

PR #1680, opened 2018-10-13
Estimate cost of deploying short-leash payment code.

goral09:
## Overview
Adds a test estimating the cost of deploying payment code.

Confluence page about [short leash deployment API](https://rchain.atlassian.net/wiki/spaces/RHOL/pages/564331259/Rholang+Short-leash+deployment+API)

As of 14.10.2018 it's:
> Cost of deploying payment contract is at minimum: 15667

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
n/a

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] If this PR adds a new feature, this PR includes tests related to this feature.
- [x] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [ ] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [ ] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1738, opened 2018-10-25
Add payment API to PoS contract

birchmd:
## Overview
Adds a method to the proof-of-stake contract that deploy payment code can use. An example of how it will be used is given in the new unit test. This PR also makes `RuntimeManager.captureResults` take a `Deploy` directly, instead of simply a `Par`, so that we can control the user/timestamp and thus the unforgable names generated.

https://rchain.atlassian.net/browse/RHOL-782

### Notes

### DIFF
--- a/casper/src/main/rholang/MakePoS.rho
+++ b/casper/src/main/rholang/MakePoS.rho
@@ -17,7 +17,10 @@ in {
   rl!(`rho:id:j6trahbxycumerwpr5qype7j43b7eqh8auebwsa9nn68if47gswh73`, *EitherCh) |
   for(@(_, ListOps) <- ListOpsCh; @(_, Either) <- EitherCh) {
     contract MakePoS(purse, @minimumBond, @maximumBond, @initBonds, return) = {
-      new this, bondsCh, joiningFeeCh, clonePurse, depositPurse, updateBonds, feeDistribution in {
+      new 
+        this, bondsCh, joiningFeeCh, clonePurse, depositPurse, 
+        updateBonds, feeDistribution, paymentStore
+      in {
         bondsCh!(initBonds) |
         joiningFeeCh!!(minimumBond) | //joining fee equal to minimum bond amount
       
@@ -178,6 +181,48 @@ in {
                   }
                 }
                 ("Left", message) => { return!((false, message)) }
+              }
+            }
+          }
+        } |
+      
+        //TODO: Fill in this contract (RHOL-462)
+        contract this(@"distribute", @purse) = {
+          Nil
+        } |
+      
+        paymentStore!((Nil, Nil, Nil, Nil)) |
+      
+        //Make it easy to query the last payment so that validators
+        //can check how much phlo there is for the deploy.
+        contract this(@"lastPayment", return) = {
+          for(@payment <- paymentStore) {
+            paymentStore!(payment) |
+            return!(payment)
+          }
+        } |
+      
+        //This is the contract that deploy payment code ultimately needs
+        //to call in order to pay for the deploy.
+        contract this(@"pay", @purse, return) = {
+          new payPurseCh, p(`rho:deploy:params`), paramsCh in {
+            clonePurse!(purse, *payPurseCh) |
+            p!(*paramsCh) |
+            for(
+              @cloneResult <- payPurseCh;
+              @codeHash, @phloRate, @userId, @timestamp <- paramsCh
+            ) {
+              match cloneResult {
+                ("Right", (payPurse, payAmount)) => {
+                  this!("distribute", payPurse) | //distribute rev paid to validators
+                  //update paymentStore
+                  for(_ <- paymentStore){ paymentStore!((codeHash, userId, timestamp, phloRate * payAmount)) } |
+                  return!(phloRate * payAmount) //return amount of phlo purchased to user (in case they care)
+                }
+                ("Left", message) => {
+                  for(_ <- paymentStore){ paymentStore!((codeHash, userId, timestamp, 0)) } |
+                  return!(0)
+                }
               }
             }
           }
--- a/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
@@ -551,7 +551,10 @@ class HashSetCasperTest extends FlatSpec with Matchers {
       )
       .withUser(user)
     val sigData = node.runtimeManager
-      .captureResults(genesis.getBody.getPostState.tuplespace, ProtoUtil.deployDataToDeploy(sigDeployData))
+      .captureResults(
+        genesis.getBody.getPostState.tuplespace,
+        ProtoUtil.deployDataToDeploy(sigDeployData)
+      )
       .head
       .exprs
       .head
@@ -586,7 +589,10 @@ class HashSetCasperTest extends FlatSpec with Matchers {
     ) *> node.casperEff.createBlock
     val blockStatus = node.casperEff.addBlock(block)
     val queryResult = node.runtimeManager
-      .captureResults(block.getBody.getPostState.tuplespace, ProtoUtil.deployDataToDeploy(paymentQuery))
+      .captureResults(
+        block.getBody.getPostState.tuplespace,
+        ProtoUtil.deployDataToDeploy(paymentQuery)
+      )
 
     val (codeHashPar, _, userIdPar, timestampPar) =
       ProtoUtil.getRholangDeployParams(paymentDeployData)
@@ -599,6 +605,8 @@ class HashSetCasperTest extends FlatSpec with Matchers {
       timestampPar,
       phloPurchasedPar
     )
+
+    node.tearDown()
   }
 
   it should "reject addBlock when there exist deploy by the same (user, millisecond timestamp) in the chain" in {
--- a/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
@@ -573,8 +573,8 @@ class HashSetCasperTest extends FlatSpec with Matchers {
          |  for(faucet <- @"faucet"; pos <- @"proofOfStake"){
          |    faucet!($amount, "ed25519", "$pkStr", *walletCh) |
          |    for(@[wallet] <- walletCh) {
-         |      for(@purse <- paymentForward){ pos!("pay", purse, Nil) } |
-         |      @wallet!("transfer", $amount, 0, "$sig", *paymentForward, Nil)
+         |      @wallet!("transfer", $amount, 0, "$sig", *paymentForward, Nil) |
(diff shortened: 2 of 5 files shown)


casper/src/main/rholang/MakePoS.rho
```diff
@@ -183,6 +186,48 @@ in {
...
+        contract this(@"lastPayment", return) = {
+          for(@payment <- paymentStore) {
+            paymentStore!(payment) |
+            return!(payment)
+          }
+        } |
+      
+        //This is the contract that deploy payment code ultimately needs
+        //to call in order to pay for the deploy.
+        contract this(@"pay", @purse, return) = {
```
goral09: Who is going to validate that the payment is being done for valid deploy? I thought that the purpose of deploy parameters was to disallow substitution of deployed code with validator's one. You're not making any use of the deploy parameters other than putting them into the `paymentStore`.

Shouldn't wallet/purse be checking that it's requested to make a transfer for the code deployed from the owner of it?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -523,6 +536,79 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+
+    val Created(block) = node.casperEff.deploy(
+      paymentDeployData
+    ) *> node.casperEff.createBlock
+    val blockStatus = node.casperEff.addBlock(block)
+    val queryResult = node.runtimeManager
+      .captureResults(
+        block.getBody.getPostState.tuplespace,
+        ProtoUtil.deployDataToDeploy(paymentQuery)
+      )
```
goral09: This is a pattern I've seen multiple times - _deploy and capture_. Can you factor it out? I've already tried doing that [here](https://github.com/rchain/rchain/pull/1680/files#diff-0d32ac2de7f4f712df4777ff7a04d512R171), and the example usage is [here](https://github.com/rchain/rchain/pull/1680/files#diff-0d32ac2de7f4f712df4777ff7a04d512R63).

casper/src/main/rholang/MakePoS.rho
```diff
@@ -183,6 +186,48 @@ in {
...
+                  return!(phloRate * payAmount) //return amount of phlo purchased to user (in case they care)
+                }
+                ("Left", message) => {
+                  for(_ <- paymentStore){ paymentStore!((codeHash, userId, timestamp, 0)) } |
+                  return!(0)
+                }
+              }
+            }
+          }
+        } |
```
goral09: If you didn't store the deploy params on the `paymentStore` channel this contract could be reduced to:
```
clonePurse!(purse, *payPurseCh) | 
for(@cloneResult <- payPurseCh) { 
  match cloneResult {
    ("Right", (payPurse, payAmount)) => {
      this!("distribute", payPurse) | …
```
am I right?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -523,6 +536,79 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+      .getGByteArray
+    val sig   = Base16.encode(Ed25519.sign(sigData.toByteArray, sk))
+    val pkStr = Base16.encode(pk)
+    val paymentCode =
+      s"""new paymentForward, walletCh in {
+         |  for(faucet <- @"faucet"; pos <- @"proofOfStake"){
+         |    faucet!($amount, "ed25519", "$pkStr", *walletCh) |
+         |    for(@[wallet] <- walletCh) {
+         |      for(@purse <- paymentForward){ pos!("pay", purse, Nil) } |
+         |      @wallet!("transfer", $amount, 0, "$sig", *paymentForward, Nil)
```
goral09: How does the verification here work? the `paymentForward` channel is the same as the one signed at the beginning because we use the same timestamp and user right?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -523,6 +536,79 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+    blockStatus shouldBe Valid
+    queryResult.head.exprs.head.getETupleBody.ps shouldBe Seq(
+      codeHashPar,
+      userIdPar,
+      timestampPar,
+      phloPurchasedPar
+    )
+
+    node.tearDown()
+  }
```
goral09: I am not sure if I follow the intent for this test but it seems to me that the only thing you test is that the `pay` contract puts data on the `paymentStore` channel. It doesn't test that purse can be used only to pay for the deploys coming from the owner. That being said - we should have a test showing that it's not possible to use purse for others people deployments.

casper/src/main/rholang/MakePoS.rho
```diff
@@ -183,6 +186,48 @@ in {
...
+        contract this(@"pay", @purse, return) = {
+          new payPurseCh, p(`rho:deploy:params`), paramsCh in {
+            clonePurse!(purse, *payPurseCh) |
+            p!(*paramsCh) |
+            for(
+              @cloneResult <- payPurseCh;
+              @codeHash, @phloRate, @userId, @timestamp <- paramsCh
+            ) {
+              match cloneResult {
+                ("Right", (payPurse, payAmount)) => {
```
goral09: Do I understand it correctly - we use all REV-s stored in the purse time the phloRate as the phlo limit?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -523,6 +536,79 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+  it should "allow paying for deploys" in {
+    val node      = HashSetCasperTestNode.standalone(genesis, validatorKeys.head)
+    val (sk, pk)  = Ed25519.newKeyPair
+    val user      = ByteString.copyFrom(pk)
+    val timestamp = System.currentTimeMillis()
+    val phloPrice = 1L
+    val amount    = 847L
+    val sigDeployData = ProtoUtil
+      .sourceDeploy(
+        s"""new retCh in { @"blake2b256Hash"!([0, $amount, *retCh].toByteArray(), "__SCALA__") }""",
```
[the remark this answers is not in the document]
goral09: OK, I get it now.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -523,6 +536,79 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+      .withUser(user)
+    val sigData = node.runtimeManager
+      .captureResults(
+        genesis.getBody.getPostState.tuplespace,
+        ProtoUtil.deployDataToDeploy(sigDeployData)
+      )
+      .head
+      .exprs
+      .head
+      .getGByteArray
```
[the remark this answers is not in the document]
goral09: OK, again - I see it now. You rely on that the name generated will be the one you use as a destination channel for transfer.

casper/src/main/rholang/MakePoS.rho
```diff
@@ -183,6 +186,48 @@ in {
...
       
+        //TODO: Fill in this contract (RHOL-462)
+        contract this(@"distribute", @purse) = {
+          Nil
+        } |
+      
+        paymentStore!((Nil, Nil, Nil, Nil)) |
+      
+        //Make it easy to query the last payment so that validators
+        //can check how much phlo there is for the deploy.
```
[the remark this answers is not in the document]
birchmd: The last element of the tuple is the amount of phlo purchased.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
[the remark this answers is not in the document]
birchmd: I don't have strong opinions about the order. I can swap the two lines

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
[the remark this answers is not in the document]
birchmd: Yes, I am relying on being able to consistently generate the same name. Which does not preclude doing it in Scala because the logic for choosing the name is in the interpreter code somewhere, but lacking a complete understanding of how that works I chose to do it this way.

casper/src/main/rholang/MakePoS.rho
birchmd: The purpose of this API is to simply give money to the validators. My model for how the whole process would go is that the user's "payment code" that gets run on the "short leash" handles the verification of things like the code hash and signature. Then as a final step it calls this API to actually pay the validators.

casper/src/main/rholang/MakePoS.rho
birchmd: Yes. But I think storing the last payment is useful. The new "short leash deploy" workflow I am imagining is validators run the "payment code", then inspect what was paid via the `lastPayment` method and proceed to run the actual deploy with the appropriate amount of phlo. It could be  that the user's payment code did not call the pay API, in which case `lastPayment` will not have changed; so I include the extra information in order to identify the correct deploy has paid.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
birchmd: Yes

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
birchmd: As I said above, the purpose of this `pay` API is simply to pay the validators, hence that is all this test looks for. We will certainly have broader tests of the whole "short leash" pipeline by the time your work is completed.

birchmd:
@goral09 PTAL

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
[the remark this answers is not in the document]
goral09: Currently available alternative is to write the rholang AST in Scala. That's what I did but it doesn't scale well and  is error prone.

casper/src/main/rholang/MakePoS.rho
goral09: OK, apparently I misunderstood how the short leash will work.

casper/src/main/rholang/MakePoS.rho
goral09: Approach I took in my [PR](https://github.com/rchain/rchain/pull/1757/files#diff-8172c898e2d53ee886672b92b2e248c9R182) is that the deployment of payment code returns phlos used.

casper/src/main/rholang/MakePoS.rho
goral09: Thanks for this explanation. I was not familiar with the properties of the purses. This explain a lot.

PR #1742, opened 2018-10-25
Add 'validator-private-key-path' option

DoSOfRedRiver:
https://rchain.atlassian.net/browse/CORE-1025

This PR adds `validator-private-key-path` option which is used to specify a path to the file which contains validators private key.
It is an alternative to `validator-private-key` option which is unsafe to use.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
@@ -266,10 +266,16 @@ object Configuration {
     // Validators
     val numValidators =
       get(_.run.numValidators, _.validators.flatMap(_.count), DefaultNumValidators)
-    val bondsFile           = getOpt(_.run.bondsFile, _.validators.flatMap(_.bondsFile))
-    val knownValidators     = getOpt(_.run.knownValidators, _.validators.flatMap(_.known))
-    val validatorPublicKey  = getOpt(_.run.validatorPublicKey, _.validators.flatMap(_.publicKey))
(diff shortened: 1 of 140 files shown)


node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -266,10 +266,15 @@ object Configuration {
...
       get(_.run.numValidators, _.validators.flatMap(_.count), DefaultNumValidators)
-    val bondsFile           = getOpt(_.run.bondsFile, _.validators.flatMap(_.bondsFile))
-    val knownValidators     = getOpt(_.run.knownValidators, _.validators.flatMap(_.known))
-    val validatorPublicKey  = getOpt(_.run.validatorPublicKey, _.validators.flatMap(_.publicKey))
-    val validatorPrivateKey = getOpt(_.run.validatorPrivateKey, _.validators.flatMap(_.privateKey))
+    val bondsFile          = getOpt(_.run.bondsFile, _.validators.flatMap(_.bondsFile))
+    val knownValidators    = getOpt(_.run.knownValidators, _.validators.flatMap(_.known))
+    val validatorPublicKey = getOpt(_.run.validatorPublicKey, _.validators.flatMap(_.publicKey))
+
+    val maybePrivateKeyPath = options.run.validatorPrivateKeyPath.toOption.map(Right(_))
```
sebtomba: Why can't the private key path be configured in `rnode.toml`?

DoSOfRedRiver: Hmm, you right. Will do.

KellyatPyrofex:
@DoSOfRedRiver, I restarted the CI test and it failed again. Please investigate.

PR #1748, opened 2018-10-26
OPS-347 Evaluation of a file in RholangCLI & same file in Node eval & node repl

zsluedem:
## Overview
Reference: https://rchain.atlassian.net/browse/CORE-1100

When evaluating any file in node eval mode or repl mode, the time to evaluate a file should be the same - within a 5% threshold as the time taken to evaluate the file in the RholangCLI.

https://rchain.atlassian.net/browse/OPS-347

- [X] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [X] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes

### DIFF
--- a/integration-testing/test/test_eval.py
+++ b/integration-testing/test/test_eval.py
@@ -1,5 +1,15 @@
 import os
-import pytest
+import time
+import logging
+
+
+def time_func(func, *args, **kwargs):
+    """
+    measure time taken by the func
+    """
+    timestamp = time.time()
+    func(*args, **kwargs)
+    return time.time() - timestamp
 
 
 def test_eval(bootstrap_node):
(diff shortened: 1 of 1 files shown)


integration-testing/test/test_eval.py
```diff
@@ -1,9 +1,36 @@
...
+        reformat_rho_script = bootstrap_node.shell_out("sed" ,"-e", 's/\/\/.*//', full_path)
+        reformat_rho_script = reformat_rho_script.replace("\n", "")
+
+        repl_time = time_func(bootstrap_node.repl, reformat_rho_script)
+        logging.info(f"It takes {repl_time} to run {full_path} in repl mode")
+
+        eval_time = time_func(bootstrap_node.eval, full_path)
+        logging.info(f"It takes {eval_time} to run {full_path} in eval mode")
+
+        # performance gap between repl and eval should be +- 5% , details see OPS-347
```
adaszko: In general, pointers to tickets should go in the PR description, not comments.  The PR template hints that.

adaszko:
Unfortunately, I think the entire [ticket](https://rchain.atlassian.net/browse/OPS-347) is flawed.  Basing whether a test passes or fails on a timing condition is bound to fail eventually, producing a flaky test.  It is subject to system load, which is beyond the control of tests.   I think the test should not check the timing but e.g. whether evaluation via repl and a file produces the same results.

MParlikar:
The purpose of this test is to make sure that minimal overhead exists when invoking the interpreter via the node versus the Rholang CLI.  The check should look for a measurable difference versus an absolute number.  If we don't look at times for passing /failing tests, how does one test for performance?

goral09:
@adaszko there are other tests that check the correctness of the result. The ticket is not flawed, we just need to agree that, for example, the time shouldn't be more than twice as long as in the CLI.

adaszko:
> @adaszko there are other tests that check the correctness of the result. The ticket is not flawed, we just need to agree that, for example, the time shouldn't be more than twice as long as in the CLI.

Twice as long is as arbitrary as 5%.  It will just take longer for it to fail.  And it will fail sooner or later.

Your work is appreciated @zsluedem, but I'm in favor of scrapping the entire ticket and the PR.

zsluedem:
@adaszko @MParlikar @goral09 We have to reach to a consensus here. 
The testcase here makes the test difficult to pass. And it depends on the situations. If you set a critical condition (like 5%), the test may pass or not depends on the condition of the machine.

I suggest we produce a test report for that(or just log) and provide it to anyone who focuses on the performance of the runtime between eval and repl.

leithaus:
Adam is correct. This is not the best way to measure the desired behavior. It is desirable to have the timing in repl mode commensurate with timing in cli, but this approach will generate a test that fails non-deterministically based on system load, and other entropic factors.

zsluedem:
As we discussed above, close this ticket now.

PR #1749, opened 2018-10-26
Remove "faucet" and "proofOfStake" forgable names

birchmd:
## Overview
In this PR we replace the forgable names which used to hold the instances for the faucet and proof-of-stake system contracts and replace them with a contract dedicated to storing such instances. System contracts are now accessed by first looking up the `SystemInstancesRegistry` from the registry, then calling `lookup` for the relevant contract.

This PR also changes the behaviour of `WalletCheck` to no longer put the claimed wallet at a forgable name, instead it keeps them internally and they can be lookup up using the public key.

Only one major usage of forgable names remain. This is the name of the forwarders created by `BondingUtil` in order to facilitate bonding. Changing these to unforgable names is a little tricky and warrants some discussion.

https://rchain.atlassian.net/browse/RHOL-920

### Notes

### DIFF
--- a/casper/src/main/rholang/SystemInstancesRegistry.rho
+++ b/casper/src/main/rholang/SystemInstancesRegistry.rho
@@ -1,0 +1,50 @@
+//scalapackage coop.rchain.rholang.registry
+
+//The purpose of this contract is to allow users to
+//look up the various instances of system contracts that
+//were created in the genesis block.
+
+//Registry info:
+//  sk: 7f2a3526664f04096067654705776950e2e475a1479d3569d8d7423e27bad603
+//  pk: 08a0f8fccace949453dcb6a885e9f50dd96a58ed51c49fc2e346aa4d42ffb7c1
+//  user == pk
+//  timestamp: 1540563894858
+//Resulting unforgable name: Unforgeable(0xeaf4d685675c199b2a2cceef521dacaaa711cd70c7d082c20d154507e5c58407) 
+//  ==> signature data == 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20eaf4d685675c199b2a2cceef521dacaaa711cd70c7d082c20d154507e5c584071001
+//  ==> signature == fbf76226c16f1327058fba896fe4d43f69b3707981ae7b8498de6fcbc36c4bbe66885904dde6a8a4f5cdf8e74f035f8e505b895a381833d3ecfb717748bc8b0d
+//URI derived from pk == `rho:id:wdwc36f4ixa6xacck3ddepmgueum7zueuczgthcqp6771kdu8jogm8`
+new
+  SystemInstancesRegistry, rs(`rho:registry:insertSigned:ed25519`), uriOut,
+  faucetStore, posStore, revStore
+in {
+  contract SystemInstancesRegistry(@"lookup", @systemContract, return) = {
+    match systemContract {
+      "faucet"     => { for(@value <- faucetStore)    { return!(value) | faucetStore!(value) } }
+      "pos"        => { for(@value <- posStore)       { return!(value) | posStore!(value) } }
+      "rev"        => { for(@value <- revStore)       { return!(value) | revStore!(value) } }
+      _            => { return!(Nil) }
+    }
+  } |
+
+  //one time use only call for registering a faucet instance
+  for(@"register", @"faucet", @value <- SystemInstancesRegistry) {
+    faucetStore!(value)
+  } |
+  
+  //one time use only call for registering a proof-of-stake instance
+  for(@"register", @"pos", @value <- SystemInstancesRegistry) {
+    posStore!(value)
+  } |
+  
+  //one time use only call for registering a rev-mint instance
+  for(@"register", @"rev", @value <- SystemInstancesRegistry) {
+    revStore!(value)
+  } |
+  
+  rs!(
+    "08a0f8fccace949453dcb6a885e9f50dd96a58ed51c49fc2e346aa4d42ffb7c1".hexToBytes(), 
+    (9223372036854775807, bundle+{*SystemInstancesRegistry}), 
+    "fbf76226c16f1327058fba896fe4d43f69b3707981ae7b8498de6fcbc36c4bbe66885904dde6a8a4f5cdf8e74f035f8e505b895a381833d3ecfb717748bc8b0d".hexToBytes(), 
+    *uriOut
+  )
+}
--- a/casper/src/main/rholang/WalletCheck.rho
+++ b/casper/src/main/rholang/WalletCheck.rho
@@ -46,7 +46,7 @@ in {
           for (@done <- doneStore) {
             if (done) {
               doneStore!(done) |
-              @statusOut!([false, "Already claimed wallet"])
+              @statusOut!((false, "Already claimed wallet"))
             } else {
 
               // Verify signature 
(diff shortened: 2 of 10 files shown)


casper/src/main/rholang/SystemInstancesRegistry.rho
```diff
@@ -0,0 +1,50 @@
...
+//  ==> signature data == 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20eaf4d685675c199b2a2cceef521dacaaa711cd70c7d082c20d154507e5c584071001
+//  ==> signature == fbf76226c16f1327058fba896fe4d43f69b3707981ae7b8498de6fcbc36c4bbe66885904dde6a8a4f5cdf8e74f035f8e505b895a381833d3ecfb717748bc8b0d
+//URI derived from pk == `rho:id:wdwc36f4ixa6xacck3ddepmgueum7zueuczgthcqp6771kdu8jogm8`
+new
+  SystemInstancesRegistry, rs(`rho:registry:insertSigned:ed25519`), uriOut,
+  faucetStore, posStore, revStore
+in {
+  contract SystemInstancesRegistry(@"lookup", @systemContract, return) = {
+    match systemContract {
+      "faucet"     => { for(@value <- faucetStore)    { return!(value) | faucetStore!(value) } }
```
goral09: We could use some syntactic sugar for this pattern.

casper/src/main/rholang/SystemInstancesRegistry.rho
```diff
@@ -0,0 +1,50 @@
...
+    revStore!(value)
+  } |
+  
+  rs!(
+    "08a0f8fccace949453dcb6a885e9f50dd96a58ed51c49fc2e346aa4d42ffb7c1".hexToBytes(), 
+    (9223372036854775807, bundle+{*SystemInstancesRegistry}), 
+    "fbf76226c16f1327058fba896fe4d43f69b3707981ae7b8498de6fcbc36c4bbe66885904dde6a8a4f5cdf8e74f035f8e505b895a381833d3ecfb717748bc8b0d".hexToBytes(), 
+    *uriOut
+  )
+}
```
goral09: I don't know, personally I don't like this indirection - I would prefer to follow the pattern you have already set up for each of the contracts (`rev`, `pos` and `faucet`) individually but you have much more experience so it's up to you.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/Rev.scala
```diff
@@ -22,30 +22,36 @@ class Rev[A](
...
     |
     |      ${faucetCode("revMint")} |
     |
     |      //PoS purse and contract creation
     |      @revMint!("makePurse", $initialTotalBond, *posPurseCh) |
     |      for(@posPurse <- posPurseCh) {
-    |        @MakePoS!(posPurse, $minimumBond, $maximumBond, $initialBondsCode, "proofOfStake")
+    |        @MakePoS!(posPurse, $minimumBond, $maximumBond, $initialBondsCode, posCh) |
+    |        for(@pos <- posCh) {
+    |          SystemInstancesRegistry!("register", "pos", pos)
```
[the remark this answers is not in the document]
goral09: I would polarize the `pos` again -- for the sake of clarity and consistency. polarities are merged so `bundle+ { bundle+ { p } }` should become `bundle+ { p }`.

casper/src/main/rholang/WalletCheck.rho
```diff
@@ -112,10 +114,20 @@ in {
...
+          Nil           => { statusOut!((false, "Unknown Ethereum address")) }
           claimContract => { @claimContract!([pubKey, *statusOut], sig) }
         }
       }
+    } |
+    
+    contract WalletCheck(@"access", @pubKey, return) = {
+      for(@wallets <- walletStore) {
+        walletStore!(wallets) |
+        match wallets.get(pubKey) {
```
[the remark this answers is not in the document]
birchmd: Hmm, that is a good point. Although perhaps this is not as bad as you might think. The map only stores the unforgable names of the wallets, not the wallets themselves. Also, it is not all wallets ever, only those that used to be RHOC at Ethereum addresses before the REV issuance at genesis. So this map will contain ~8k key-value pairs, where each key is a 64-character string and 32-byte array (the unforgable name reference). This comes out to a map of ~800kB, which is not insignificant, but also not huge.

casper/src/main/rholang/SystemInstancesRegistry.rho
birchmd: One thing I don't like about the `rho:id:...` pattern is how human unfriendly it is. People will constantly need to look at some table which gives the ID for the different contracts they care about. I think if I were to go back, I would make everything be done in this way. Then there would only be 1 ID to ever have to look up and human-readable names could be used from that point on.

That said, the current status with these set up in this way and the others being inserted into the registry directly, does give a nice conceptual distinction, which could otherwise be confusing to some users. You use `rho:id:...` to look up the generic `MakePoS` contract, but you use this method to get the specific proof-of-stake instance which is used in the shard.

casper/src/main/rholang/WalletCheck.rho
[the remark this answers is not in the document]
goral09: How will the future wallets (post genesis launch) be registered ?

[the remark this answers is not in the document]
birchmd: That will be up to the user. In my view the easiest thing would be to use `insertSigned` with the same public key that you use with the wallet.

casper/src/main/rholang/SystemInstancesRegistry.rho
goral09: why is `MakePos` different contract than `rev`, `pos` or `faucet`?

birchmd: That question is exactly the conceptual distinction I am talking about :)

`MakePoS` is a contract for constructing proof-of-stake contracts, while `pos` is the specific proof-of-stake contract we are actually using. Similarly, `MakeMint` creates mint instances, while `rev` is (an attenuated forwarder of) the particular mint which represents the REV token. And same thing for `BasicWalletFaucet` vs `faucet`.

casper/src/main/scala/coop/rchain/casper/util/BondingUtil.scala
```diff
@@ -115,10 +121,15 @@ object BondingUtil {
...
       transferSig     = Secp256k1.sign(transferSigData, secKey)
-    } yield s"""
-             |for(wallet <- @"$pubKey") {
-             |  wallet!("transfer", $amount, $nonce, "${Base16.encode(transferSig)}", "$destination", "$transferStatusOut")
-             |}""".stripMargin
+    } yield s"""new rl(`rho:registry:lookup`), WalletCheckCh, result in {
+               |  rl!(`rho:id:oqez475nmxx9ktciscbhps18wnmnwtm6egziohc3rkdzekkmsrpuyt`, *WalletCheckCh) |
+               |  for(@(_, WalletCheck) <- WalletCheckCh) {
+               |    @WalletCheck!("access", "$pubKey", *result) |
+               |    for(@(true, wallet) <- result) {
```
KentShikama: Does this `true` match anything? It is amazing that it just works.

birchmd: `WalletCheck` returns a pair where the first element is true/false (success/failure) and the second element is an error message or a wallet. I only want to continue on the success case. It would probably be good to give some indication to the user in the failure case, but there is not really a good way of doing that right now.

PR #1750, opened 2018-10-27
Bump version for blocks

KentShikama:
Also add validation for block version

## Overview

https://rchain.atlassian.net/browse/RHOL-1013

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -157,6 +157,22 @@ object Validate {
       true.pure[F]
     }
 
+  def version[F[_]: Applicative: Log](b: BlockMessage, version: Long): F[Boolean] = {
+    val blockVersion = b.header.get.version
+    if (blockVersion == version) {
+      true.pure[F]
+    } else {
+      for {
+        _ <- Log[F].warn(
+              ignore(
+                b,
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -157,6 +157,22 @@ object Validate {
...
+      true.pure[F]
+    } else {
+      for {
+        _ <- Log[F].warn(
+              ignore(
+                b,
+                s"received block version ${blockVersion.toString} is the expected version ${version.toString}."
+              )
+            )
+      } yield false
```
goral09: Instead of using `for` comprehension I think that `map` would be cleaner.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -157,6 +157,20 @@ object Validate {
...
 
+  def version[F[_]: Applicative: Log](b: BlockMessage, version: Long): F[Boolean] = {
+    val blockVersion = b.header.get.version
+    if (blockVersion == version) {
+      true.pure[F]
+    } else {
+      Log[F].warn(
+        ignore(
+          b,
+          s"received block version ${blockVersion.toString} is the expected version ${version.toString}."
```
itegulov: Why do you need to call `.toString` here explicitly?

PR #1751, opened 2018-10-27
Merge dev into release-rnode-v0.7

KentShikama:


MParlikar:
I don't think we want to do this.  This work should go into Node 0.8 when the time is right

PR #1755, opened 2018-10-28
Bump version for blocks

KentShikama:
Same as #1750

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -157,6 +157,22 @@ object Validate {
       true.pure[F]
     }
 
+  def version[F[_]: Applicative: Log](b: BlockMessage, version: Long): F[Boolean] = {
+    val blockVersion = b.header.get.version
+    if (blockVersion == version) {
+      true.pure[F]
+    } else {
+      for {
+        _ <- Log[F].warn(
+              ignore(
+                b,
+                s"received block version ${blockVersion.toString} is the expected version ${version.toString}."
+              )
+            )
+      } yield false
+    }
+  }
+
   /*
    * TODO: Double check ordering of validity checks
    * TODO: Add check for missing fields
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -140,7 +140,7 @@ object Genesis {
       wallets     <- getWallets[F](walletsFile, maybeWalletsPath)
       bonds       <- getBonds[F](bondsFile, numValidators, genesisPath)
       timestamp   <- deployTimestamp.fold(Time[F].currentMillis)(_.pure[F])
-      initial     = withoutContracts(bonds = bonds, timestamp = 1L, version = 0L, shardId = shardId)
+      initial     = withoutContracts(bonds = bonds, timestamp = 1L, version = 1L, shardId = shardId)
```
odeac: is the `version=1L` always the same as `private val version = 1L` in `MultiParentCasperImpl.scala`? 

If that's the case then we should have a constant in order to make sure that we use the same value everywhere.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -157,6 +157,20 @@ object Validate {
       true.pure[F]
     }
 
+  def version[F[_]: Applicative: Log](b: BlockMessage, version: Long): F[Boolean] = {
+    val blockVersion = b.header.get.version
```
odeac: I don't really like the unsafe `b.header.get` here. I understand that the blocks should always have a header but still it feels bad. 

I think it's safe to say that the version validation fails if the block header is missing.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: I'll address this in the on dev branch

KentShikama: And here is the ticket for it so that I don't forget https://rchain.atlassian.net/browse/RHOL-1015

PR #1759, opened 2018-10-29
Fix http4s shutdown exception

DoSOfRedRiver:
https://rchain.atlassian.net/browse/CORE-1290

This PR fixes the exception thrown by http4s during the shutdown procedure.
Also refactored some http4s code to use `Task` instead `IO`.

Note that some dependencies were changed.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -34,12 +34,10 @@ import coop.rchain.p2p.effects._
 import coop.rchain.rholang.interpreter.Runtime
 import coop.rchain.shared._
 
-import io.grpc.Server
 import kamon._
 import kamon.zipkin.ZipkinReporter
 import monix.eval.Task
 import monix.execution.Scheduler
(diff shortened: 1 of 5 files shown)


node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -134,7 +134,7 @@ class NodeRuntime private[node] (
       _   <- transport.shutdown(msg)
       _   <- log.info("Shutting down HTTP server....")
       _   <- Task.delay(Kamon.stopAllReporters())
-      _   <- LiftIO[Task].liftIO(servers.httpServer.shutdown).attempt
+      _   <- servers.httpServer.cancel
```
sebtomba: I'm not sure if this is a proper way to shut down an http4s HTTP server

DoSOfRedRiver: In 0.19.0 they removed shutdown methods in favor to cats effects Resource and fs2 streams.
This code was basically taken from documentation to 0.19.0
https://http4s.org/v0.19/service/

PR #1769, opened 2018-10-30
sbt: Make tests run parallel

woky:
Michael reported slower test run times and suspected that setting
fork:=true might have caused it. Lukasz linked to relevant sections of
https://www.scala-sbt.org/1.x/docs/Testing.html

  By default, sbt runs all tasks in parallel and within the same JVM as
  sbt itself. Because each test is mapped to a task, tests are also run
  in parallel by default.

  The setting:

  Test / fork := true specifies that all tests will be executed in a
  single external JVM. See Forking for configuring standard options for
  forking. By default, tests executed in a forked JVM are executed
  sequentially.

This commit
  - limits forking to (integration) test executions and
  - makes tests run in parallel in forked JVMs (testForkedParallel).

This keeps test threads isolated to separate JVMs and makes the run time
roughly same as with fork:=false.


KellyatPyrofex:
@woky, are you able to resolve conflicts and get another review? Michael's review is no longer valid.

PR #1776, opened 2018-11-01
Remove temp directories used for rspace in tests

odeac:
## Overview
The temp dirs accumulate and, especially when we run property-based tests, they become a serious problem. We need to remove them in a clean way.

https://rchain.atlassian.net/browse/RHOL-979

### Notes


rholang/src/test/scala/coop/rchain/rholang/ResourceTools.scala
```diff
@@ -0,0 +1,46 @@
...
+      release: (R, Option[Throwable]) => F[Unit]
+  )(job: R => F[T])(implicit F: MonadError[F, Throwable]): F[T] =
+    for {
+      resource <- acquire
+      attempt  <- job(resource).attempt
+      result <- attempt.fold(
+                 ex => release(resource, Some(ex)) *> F.raiseError(ex),
+                 r => release(resource, None) *> F.pure(r)
+               )
+    } yield result
```
goral09: Aren't you implementing the cats' `Resource` typeclass by hand?

https://typelevel.org/cats-effect/datatypes/resource.html#with-javaio-using-io

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
```diff
@@ -1,45 +1,28 @@
...
-            )
-      })
-
-    withResource[F, A, Path](
-      F.pure(Files.createTempDirectory(prefix)),
-      deleteTempDir
-    )(job)
-  }
+  def mkTempDir[F[_]: Applicative](prefix: String): Resource[F, Path] =
+    Resource.makeCase(Applicative[F].pure(Files.createTempDirectory(prefix)))(
```
goral09: `Files.createTempDirectory(prefix)` is effectful. You should `delay` it with `Sync.delay`.

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
```diff
@@ -1,45 +1,28 @@
...
+      (path, exitCase) =>
+        Applicative[F].pure(exitCase match {
```
goral09: Same here, effectful.

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
```diff
@@ -1,45 +1,28 @@
...
+          case Error(ex) =>
+            logger
+              .error(
+                s"Exception thrown while using the tempDir '$path'. Temporary dir NOT deleted.",
```
goral09: Are you sure? Even if exception is thrown during the usage we should delete the directory.

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
```diff
@@ -25,4 +30,29 @@ object ResourceTools {
...
+
+  def mkRhoISpace[F[_]: Sync, A](
+      prefix: String = "",
+      branch: String = "test",
+      mapSize: Long = 1024L * 1024L * 4
+  ): Resource[F, RhoISpace[F]] = {
+    import coop.rchain.rholang.interpreter.storage.implicits._
+
+    def mkRspace(dbDir: Path): F[RhoISpace[F]] = {
+      val context: RhoContext = Context.create(dbDir, mapSize)
```
goral09: I think that `RhoContext` has a `close` method which should be called in the `release` part of the `Resource`.

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
```diff
@@ -85,4 +85,39 @@ package object implicits {
...
+    override def handleErrorWith[A](fa: Try[A])(f: Throwable => Try[A]): Try[A] =
+      cats.implicits.catsStdInstancesForTry.handleErrorWith(fa)(f)
+
+    override def pure[A](x: A): Try[A] = cats.implicits.catsStdInstancesForTry.pure(x)
+
+    override def flatMap[A, B](fa: Try[A])(f: A => Try[B]): Try[B] =
+      cats.implicits.catsStdInstancesForTry.flatMap(fa)(f)
+
+    override def tailRecM[A, B](a: A)(f: A => Try[Either[A, B]]): Try[B] =
+      cats.implicits.catsStdInstancesForTry.tailRecM(a)(f)
```
goral09: Do I understand that you can't import the `catsStdInstancesForTry` implicitly into a scope because it triggers the compilation errors? If so, then please asssign the `cats.implicits.catsStdInstancesForTry` to some local variable and call it, instead of `cats.implicits.catsStdInstancesForTry` every time.

odeac: exactly. I had a lot of headaches with implicit conflicts when I imported all the implicits. By calling them explicitely I managed to get over the compilation errors. I would say it's not a big problem because the users of this class won't have these problems

rholang/src/test/scala/coop/rchain/rholang/ResourceTools.scala
odeac: Yes, I am. Initially I wasn't aware that cats.Resource does exactly what I needed. In the end I refactored everything to use cats.Resource. See last commits

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
odeac: This is to be discussed. Let's get @ArturGajowy in this discussion as well. He suggested this approach and I think it's a good idea not to delete temp files when the test fails because there may be debugging information in those temp files which can be helpful.

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
odeac: your suggestion to use a local name it's already there. See commit 2b74112

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
goral09: How would you use the residual directory for debugging? I am just asking b/c I have never found a need for that and we don't have a tooling.

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
goral09: Why did you resolve this discussion @odeac ?

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
odeac: Fixed

rholang/src/main/scala/coop/rchain/rholang/ResourceTools.scala
odeac: That's something we should figure out how to do.

ArturGajowy: I'd rather have people do rm /tmp/* from time to time than say 'oh, this blew up on me. Do I debug this? No, there's not trace after that... Rerun, tests green, my job is done here'.

PR #1779, opened 2018-11-05
CORE-1383

rabbitonweb:
## Overview
After this change, whenever you call stream with given Blob, that
message is pushed on a queue and the code finishes the
execution. There is a separate consumer of this queue that will then
pop elements from it and send it (in stream fashion) to the
destination.

This way calling `stream` API executes immediately (in matter of
seconds) even for big number of peers involved.

Next PRs should bring to it:
- back pressure (as much as it is possible)
- retries

Bring TcpTransportlayerspec back to live.

CORE-1383

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
@@ -12,20 +12,25 @@ import monix.reactive.subjects.ConcurrentSubject
 
 case class ToStream(peerNode: PeerNode, blob: Blob)
 
-class StreamObservable(bufferSize: Int)(implicit scheduler: Scheduler)
+class StreamObservable(bufferSize: Int)(implicit log: Log[Task], scheduler: Scheduler)
     extends Observable[ToStream] {
 
-  val subject = ConcurrentSubject.publishToOne[ToStream](DropNew(bufferSize))
+  val subject = buffer.LimitedBufferObservable.dropNew[ToStream](bufferSize)
 
-  def stream(peers: List[PeerNode], blob: Blob): Task[Unit] =
-    peers
-      .traverse(
-        peer =>
-          Task.fromFuture {
-            subject.onNext(ToStream(peer, blob))
-          }
-      )
-      .as(())
+  def stream(peers: List[PeerNode], blob: Blob): Task[Unit] = {
+    def push(peer: PeerNode): Task[Boolean] =
+      Task.delay(subject.pushNext(ToStream(peer, blob)))
+
+    def retry(failed: List[PeerNode]): Task[Unit] =
+      log.debug(s"Retrying for $failed") *> stream(failed, blob)
+
+    for {
+      results     <- peers.traverse(push _)
+      paired      = peers.zip(results)
+      (_, failed) = paired.partition(_._2)
+      _           <- if (!failed.isEmpty) retry(failed.map(_._1)) else Task.unit
+    } yield ()
+  }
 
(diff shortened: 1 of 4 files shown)


comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -0,0 +1,36 @@
...
+    peers
+      .traverse(
+        peer =>
+          Task.fromFuture {
+            subject.onNext(ToStream(peer, blob))
+          }
+      )
+      .as(())
+
+  def unsafeSubscribeFn(subscriber: Subscriber[ToStream]): Cancelable = {
```
odeac: The point of the names starting with unsafe is to trigger an alarm signal in the developer writing that code and make him think twice whether he should use that function in that context or not. 

We have a LOT of calls to unsafe* functions and it became a habit to deal with unsafe stuff. Can we use a SafeSubscriber here?

rabbitonweb: @odeac this is the monix API I has to stick with

rabbitonweb: what's a safesubscriber? :)

odeac: I was looking at this: https://monix.io/docs/2x/reactive/observers.html#safe-subscriber  I'm not sure how it actually works but I'm worried about our familiarity with the unsafe* stuff in our code

odeac: after the call I understand the problems and why unsafeSubscribeFn is unsafe and needed

comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -0,0 +1,36 @@
...
+import monix.reactive.observers.Subscriber
+import monix.reactive.OverflowStrategy._
+import monix.reactive.subjects.ConcurrentSubject
+
+case class ToStream(peerNode: PeerNode, blob: Blob)
+
+class StreamObservable(bufferSize: Int)(implicit scheduler: Scheduler)
+    extends Observable[ToStream] {
+
+  val subject = ConcurrentSubject.publishToOne[ToStream](DropNew(bufferSize))
```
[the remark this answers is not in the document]
rabbitonweb: makes sense!

comm/src/test/scala/coop/rchain/comm/transport/TransportLayerRuntime.scala
```diff
@@ -66,13 +66,14 @@ abstract class TransportLayerRuntime[F[_]: Monad, E <: Environment] {
...
             remote   = e2.peer
+            _        <- localTl.receive(null, null)
             _ <- remoteTl.receive(
                   protocolDispatcher.dispatch(remote),
                   streamDispatcher.dispatch(remote)
                 )
             r <- execute(localTl, local, remote)
             // arbitrary sleep value, so environment has time to handle requests
-            _ <- time.sleep(200 millisecond)
+            _ <- time.sleep(1000 millisecond)
```
odeac: This sleep looks bad. I know it was here before but it still looks bad. Is this really needed? Why 1s?

sebtomba: @odeac This is a unit test environment. I'll take a look at this in a different task because of nondeterminism.

PR #1809, opened 2018-11-06
Catch Closed Channel Exception when sending a response

sebtomba:
## Overview
Catch Closed Channel Exception when sending a response and log it as a warning

https://rchain.atlassian.net/browse/CORE-1061

### DIFF
--- a/shared/src/main/scala/coop/rchain/grpcmonix/GrpcMonix.scala
+++ b/shared/src/main/scala/coop/rchain/grpcmonix/GrpcMonix.scala
@@ -1,10 +1,14 @@
 package coop.rchain.grpcmonix
 
+import scala.concurrent.Future
+
+import coop.rchain.shared.{Log, LogSource}
+
 import com.google.common.util.concurrent.ListenableFuture
 import io.grpc.stub.StreamObserver
 import monix.eval.{Callback, Task}
-import monix.execution.Ack.{Continue, Stop}
 import monix.execution.{Ack, Scheduler}
+import monix.execution.Ack.{Continue, Stop}
 import monix.reactive.Observable
 import monix.reactive.Observable.Operator
 import monix.reactive.observers.Subscriber
@@ -12,9 +16,10 @@ import monix.reactive.subjects.PublishSubject
 import org.reactivestreams.{Subscriber => SubscriberR}
 import scalapb.grpc.Grpc
 
-import scala.concurrent.Future
-
 object GrpcMonix {
+
+  private val logger                        = Log.logId
+  private implicit val logSource: LogSource = LogSource(this.getClass)
 
   type GrpcOperator[I, O] = StreamObserver[O] => StreamObserver[I]
   type Transformer[I, O]  = Observable[I] => Observable[O]
@@ -64,10 +69,13 @@ object GrpcMonix {
   def grpcObserverToMonixCallback[T](observer: StreamObserver[T]): Callback[T] =
     new Callback[T] {
       override def onError(t: Throwable): Unit = observer.onError(t)
-      override def onSuccess(value: T): Unit = {
-        observer.onNext(value)
-        observer.onCompleted()
-      }
+      override def onSuccess(value: T): Unit =
+        try {
+          observer.onNext(value)
+          observer.onCompleted()
+        } catch {
+          case t: Throwable => logger.warn(s"Failed to send response: ${t.getMessage}")
+        }
     }
 
   def liftByGrpcOperator[I, O](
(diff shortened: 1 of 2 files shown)


shared/src/main/scala/coop/rchain/grpcmonix/GrpcMonix.scala
```diff
@@ -64,10 +69,13 @@ object GrpcMonix {
...
-      override def onSuccess(value: T): Unit = {
-        observer.onNext(value)
-        observer.onCompleted()
-      }
+      override def onSuccess(value: T): Unit =
+        try {
+          observer.onNext(value)
+          observer.onCompleted()
+        } catch {
+          case t: Throwable => logger.warn(s"Failed to send response: ${t.getMessage}")
```
Primetalk: WARN: 
1. Serious exceptions might be hidden. Consider using `NonFatal`.
2. Logging level might be insufficient. `Throwable` might better be logged with `error`.
3. Stacktrace is omitted. Consider adding `, t`.

sebtomba: Yes! Thank you.

Primetalk: Also I would personally rethrow the exception to signal the failure. Or send a message to a supervisor.

sebtomba: No, that's the whole point of this PR. The exception would not get catched anywhere and eventually handled by the default `UncaughtExceptionReporter` which just prints the exception and the stack trace to the console. We may want at some point to provide a custom `UncaughtExceptionReporter` to all our `Scheduler`s and log it there. I'll create a ticket for that.

sebtomba: Maybe I can provide more context. When a peer node A sends a request to a node B and then closes the channel (f.e. because of a timeout) before node B can send a response then this isn't considered an error (on peer B side). On the other hand, we want to log the exception as a warning to the log file (not the console) for debugging purposes.

Primetalk: I see what you mean. I thought that even if `A` recovered from the error on it's own, `B` might still want to report the failure.
Also it might worth handling exactly this specific case of timeout in `A` and escalating other exceptions.

Primetalk: What is going to happen in the other method - `onError`? Shouldn't we have identical handling there?

sebtomba: I was considering to apply the same handling here and catch only `NonFatal` exceptions. However, the Monix `Obeserver` contract states that we shouldn't throw any exceptions inside `onNext`

Primetalk: I mean not `onNext` but `onError`. There we also notify `observer` about something. In this case it's an error.

sebtomba: The timeout is handled on peer node `A` side. Here we are handling an exception on the responding node `B` and there is not much we can do from here. We don't change the semantics of the node-to-node communication in any way. Node `A` may retry with a larger timeout. BTW, any network should be always considered non-deterministic and not reliable. It's not possible (in a synchronous request-response messaging pattern) to guarantee that **all** responses will reach the requesting peer. If a requesting node is gone, it's gone. However, it's possible to build robust and resilient systems in non-deterministic environments.

Primetalk: I see what you mean. This looks like an "implicit exception propagation". We have exception on side B (of type `Exception1`) and we get exception on side A (of type `TimeoutException`). So the communication is already unhealthy. On side A we take ordinary measures about exceptions (retry, escalate, etc.) and on side B we can simply ignore the actual problem or log it for investigation.
It's now more clear to me and sounds reasonable. Thank you for the explanation.

What about exception propagation via the call stack on side B? Do we lose anything here?

Also if we anyway have exception handled at some point. Why don't we change the handling there and log all exceptions to log rather than console? Like in `uncaughtExceptionHandler`?

sebtomba: Just for clarification. This is **not** an implicit exception propagation. Node `A` expects a response in time `t`. It's an implementation detail how we achieve this. It may be a `TimeoutException` or something else. Anyways, after time `t` side A gives up and closes the connection to node `B`. Obviously, node `B` can't send a response if the connection has been already closed by side `A`. The gRPC subsystem throws an exception in this case. Since it's a 3rd party subsystem (grpc-java) we can't change it. In Java, it's normal to signal faults with exceptions. We are catching the exception and logging it as a warning for debugging purposes. That's all.

PR #1823, opened 2018-11-12
RHOL-1028: Clean up SortedParMap

tymm:
## Overview

- In a previous PR I changed `sortedMap` to `Map[Par,Par]` which caused a bug since `Map` does not preserve insertion order. I added a test case where the non-preservation of `Map` would show. 
- I added a generative test from @ArturGajowy that checks if sortedness is preserved.

I guess I should still extend from `IterableLike` but I thought I first put this out and see what @ArturGajowy says.

https://rchain.atlassian.net/browse/RHOL-1028

### DIFF
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -1,29 +1,50 @@
 package coop.rchain.models
 
-import scala.collection.MapLike
 import coop.rchain.models.rholang.sorter.ordering._
 
-class SortedParMap private (private val ps: Map[Par, Par])
-    extends Map[Par, Par]
-    with MapLike[Par, Par, SortedParMap] {
+import scala.collection.GenTraversableOnce
+import scala.collection.immutable.ListMap
 
-  lazy val sortedMap: Map[Par, Par] = ps.sort.toMap
+final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)] {
 
-  override def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
+  lazy val sortedMap: ListMap[Par, Par] = ListMap(ps.sort: _*)
 
-  override def get(key: Par): Option[Par] = sortedMap.get(key)
+  def +(kv: (Par, Par)): SortedParMap = SortedParMap(sortedMap + kv)
 
-  override def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
+  def ++(kvs: GenTraversableOnce[(Par, Par)]): SortedParMap = SortedParMap(sortedMap ++ kvs)
 
-  override def +[V1 >: Par](kv: (Par, V1)): Map[Par, V1] = sortedMap + kv
+  def -(key: Par): SortedParMap = SortedParMap(sortedMap - key)
 
-  override def -(key: Par): SortedParMap = SortedParMap(sortedMap - key)
+  def --(keys: GenTraversableOnce[Par]): SortedParMap = SortedParMap(sortedMap -- keys)
+
+  def apply(par: Par): Par = sortedMap(par)
+
+  def contains(par: Par): Boolean = sortedMap.contains(par)
+
+  def empty: SortedParMap = SortedParMap(ListMap.empty[Par, Par])
+
+  def get(key: Par): Option[Par] = sortedMap.get(key)
+
+  def getOrElse(key: Par, default: Par): Par = sortedMap.getOrElse(key, default)
+
+  def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
+
+  def keys: Iterable[Par] = sortedMap.keys
+
+  def values: Iterable[Par] = sortedMap.values
+
+  override def equals(that: Any): Boolean = that match {
+    case spm: SortedParMap => spm.sortedMap == this.sortedMap
+    case _                 => false
+  }
 }
 
 object SortedParMap {
   def apply(map: Map[Par, Par]): SortedParMap = new SortedParMap(map)
 
   def apply(seq: Seq[(Par, Par)]): SortedParMap = SortedParMap(seq.toMap)
+
+  def apply(sortedParMap: SortedParMap): SortedParMap = SortedParMap(sortedParMap.sortedMap)
 
   def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
 }
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -17,7 +17,10 @@ final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)
 
   def -(key: Par): SortedParMap = SortedParMap(sortedMap - key.sort)
 
-  def --(keys: GenTraversableOnce[Par]): SortedParMap = SortedParMap(sortedMap -- keys.sort)
+  def --(keys: GenTraversableOnce[Par]): SortedParMap =
+    SortedParMap(keys.foldLeft(sortedMap) { (map, kv) =>
+      map - kv.sort
+    })
 
   def apply(par: Par): Par = sortedMap(par.sort)
 
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -48,7 +48,5 @@ object SortedParMap {
 
   def apply(seq: Seq[(Par, Par)]): SortedParMap = SortedParMap(seq.toMap)
 
-  def apply(sortedParMap: SortedParMap): SortedParMap = SortedParMap(sortedParMap.sortedMap)
-
   def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
 }
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -7,7 +7,9 @@ import scala.collection.immutable.ListMap
 
 final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)] {
 
-  lazy val sortedMap: ListMap[Par, Par] = ListMap(ps.sort: _*)
+  // TODO: Merge `sortedMap` and `sortedMapLookup` into one VectorMap once available
+  lazy val sortedMap: ListMap[Par, Par]   = ListMap(ps.sort: _*)
+  lazy val sortedMapLookup: Map[Par, Par] = ps.sort.toMap
 
   def +(kv: (Par, Par)): SortedParMap = SortedParMap(sortedMap + kv)
 
@@ -17,15 +19,15 @@ final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)
 
   def --(keys: GenTraversableOnce[Par]): SortedParMap = SortedParMap(sortedMap -- keys)
 
-  def apply(par: Par): Par = sortedMap(par)
+  def apply(par: Par): Par = sortedMapLookup(par)
 
-  def contains(par: Par): Boolean = sortedMap.contains(par)
+  def contains(par: Par): Boolean = sortedMapLookup.contains(par)
 
-  def empty: SortedParMap = SortedParMap(ListMap.empty[Par, Par])
+  def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
 
-  def get(key: Par): Option[Par] = sortedMap.get(key)
+  def get(key: Par): Option[Par] = sortedMapLookup.get(key)
 
-  def getOrElse(key: Par, default: Par): Par = sortedMap.getOrElse(key, default)
+  def getOrElse(key: Par, default: Par): Par = sortedMapLookup.getOrElse(key, default)
 
   def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
 
@@ -37,6 +39,8 @@ final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)
     case spm: SortedParMap => spm.sortedMap == this.sortedMap
     case _                 => false
   }
+
+  override def hashCode(): Int = sortedMap.hashCode()
 }
 
 object SortedParMap {
--- a/models/src/main/scala/coop/rchain/models/SortedParMap.scala
+++ b/models/src/main/scala/coop/rchain/models/SortedParMap.scala
@@ -3,13 +3,13 @@ package coop.rchain.models
 import coop.rchain.models.rholang.sorter.ordering._
 
 import scala.collection.GenTraversableOnce
-import scala.collection.immutable.ListMap
+import scala.collection.immutable.HashMap
 
 final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)] {
 
-  // TODO: Merge `sortedMap` and `sortedMapLookup` into one VectorMap once available
-  lazy val sortedMap: ListMap[Par, Par]   = ListMap(ps.sort: _*)
-  lazy val sortedMapLookup: Map[Par, Par] = ps.sort.toMap
+  // TODO: Merge `sortedList` and `sortedMap` into one VectorMap once available
+  lazy val sortedList: List[(Par, Par)] = ps.sort
+  lazy val sortedMap: HashMap[Par, Par] = HashMap(sortedList: _*)
 
   def +(kv: (Par, Par)): SortedParMap = SortedParMap(sortedMap + kv)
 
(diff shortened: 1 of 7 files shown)


models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,50 @@
...
+    case _                 => false
+  }
 }
 
 object SortedParMap {
   def apply(map: Map[Par, Par]): SortedParMap = new SortedParMap(map)
 
   def apply(seq: Seq[(Par, Par)]): SortedParMap = SortedParMap(seq.toMap)
 
+  def apply(sortedParMap: SortedParMap): SortedParMap = SortedParMap(sortedParMap.sortedMap)
```
ArturGajowy: Why is that needed? Why wouldnt `def apply(sortedParMap: SortedParMap): SortedParMap = sortedMap` suffice - and in that case - why even call the method?

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,50 @@
...
 import coop.rchain.models.rholang.sorter.ordering._
 
-class SortedParMap private (private val ps: Map[Par, Par])
-    extends Map[Par, Par]
-    with MapLike[Par, Par, SortedParMap] {
+import scala.collection.GenTraversableOnce
+import scala.collection.immutable.ListMap
 
-  lazy val sortedMap: Map[Par, Par] = ps.sort.toMap
+final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)] {
```
ArturGajowy: Why do you say we should implement IterableLike instead of Iterable?

ArturGajowy:
I've restarted the build, but it failed again with (amongst others) a timeout in `StackSafetySpec`. Can you bump the timeout to 20 s?

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: There is some code, e.g. `Reduce.scala:1294` that depends on this. Should I refactor the code that depends on that?

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+  lazy val sortedMapLookup: Map[Par, Par] = ps.sort.toMap
 
-  override def get(key: Par): Option[Par] = sortedMap.get(key)
+  def +(kv: (Par, Par)): SortedParMap = SortedParMap(sortedMap + kv)
 
-  override def iterator: Iterator[(Par, Par)] = sortedMap.toIterator
+  def ++(kvs: GenTraversableOnce[(Par, Par)]): SortedParMap = SortedParMap(sortedMap ++ kvs)
 
-  override def +[V1 >: Par](kv: (Par, V1)): Map[Par, V1] = sortedMap + kv
+  def -(key: Par): SortedParMap = SortedParMap(sortedMap - key)
```
ArturGajowy: the key itself may be not sorted. Take e.g. 

//ks == Sortable[Par].sort(ku), ks != ku
val m = SortedMap(ku -> 1)
val m2 = m - ku
//m2 == m, while it shouldn't
//this is because:
//m ~= Map(ks -> 1)

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
 
-  override def -(key: Par): SortedParMap = SortedParMap(sortedMap - key)
+  def --(keys: GenTraversableOnce[Par]): SortedParMap = SortedParMap(sortedMap -- keys)
```
ArturGajowy: ditto

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+
+  def apply(par: Par): Par = sortedMapLookup(par)
```
ArturGajowy: same thing about sortedness of the input par

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+
+  def contains(par: Par): Boolean = sortedMapLookup.contains(par)
```
ArturGajowy: ditto

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+
+  def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
+
+  def get(key: Par): Option[Par] = sortedMapLookup.get(key)
```
ArturGajowy: ditto

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+
+  def getOrElse(key: Par, default: Par): Par = sortedMapLookup.getOrElse(key, default)
```
ArturGajowy: ditto

models/src/main/scala/coop/rchain/models/SortedParMap.scala
ArturGajowy: How about this?

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
```diff
@@ -80,4 +83,73 @@ class SortedParMapSpec extends FlatSpec with Matchers {
...
+
+  it should "preserve sortedness of all elements and the whole map for all its operations" in {
+    import coop.rchain.models.testImplicits._
+
+    forAll { (pars1: Seq[Par], pars2: Seq[Par]) =>
+      val map1 = SortedParMap(pars1.zip(pars2))
+      val map2 = SortedParMap(pars2.zip(pars1))
+      checkSorted(map1)
+      checkSorted(map2)
+      Seq(map1.headOption, map2.headOption).flatten.foreach(kv => {
```
ArturGajowy: Try to make this test find all the 'inputs have to be sorted' bugs. I think making this say
`Seq(parts1.headOption, parts2.headOption)` should suffice for the methods already called (please call the remaining ones as needed). If not, please find a different way to test that.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
```diff
@@ -1,29 +1,54 @@
...
+import scala.collection.GenTraversableOnce
+import scala.collection.immutable.ListMap
 
-  lazy val sortedMap: Map[Par, Par] = ps.sort.toMap
+final class SortedParMap private (ps: Map[Par, Par]) extends Iterable[(Par, Par)] {
 
-  override def empty: SortedParMap = SortedParMap(Map.empty[Par, Par])
+  // TODO: Merge `sortedMap` and `sortedMapLookup` into one VectorMap once available
+  lazy val sortedMap: ListMap[Par, Par]   = ListMap(ps.sort: _*)
+  lazy val sortedMapLookup: Map[Par, Par] = ps.sort.toMap
```
ArturGajowy: This means sorting twice. There's no need for two maps - just a:
 - List[Par] (for the key order) or List[(Par, Par)] (for key and entry order) 
and 
 - a HashMap[Par, Par] for the lookup
should suffice.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
ArturGajowy: Yes, please do. It's redundant.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
goral09: It should be enough to only sort the keys and not values when adding and removing elements from the Map and sort everything else when checking equality and before serialization.

ArturGajowy: Nopes, you'd also have to values when accessing elements. Also, since the types don't allow us to discern sorted from non-sroted ones, I want to keep the invariant of 'everything inside is sorted'.

goral09: Then sort when `EMap` are reduced/interpreted/evaled.

goral09: I am not sure if your approach is correct. You make internals leak to the user.
```scala
sortedMap.add(key, value).get(key) != value
```
because of sortedness. It is your call though.

ArturGajowy: Thanks for that remark! Indeed, that's undesirable.

Again though, the types are failing us here, because we've failed to use them in the first place. It is not an implementation detail of the SortedMap. Quite the contrary, it's the `raison d'etre` for it ;) SortedMap is used to represent a sorted map term. Such term has a unique canonical shape and this shape should be used here :)

Previously, the SortedMap had broken equals because `get` returned unsorted elements. But `SortedMap(1-> [2, 3])` should equal `SortedMap(1 -> [3, 2])`. Otherwise the latter map does not represent a sorted term :)

Let me know if that convinces you to the changed behavior :)

re: your proposal about postponing the sorting - I'd rather not reason about subsequent potential usages of the SortedMap and instead make all of them behave consistently. I'm not good at such reasoning. The types are great at that - but we haven't invited them :hits_blunt:

goral09: > But `SortedMap(1-> [2, 3])` should equal `SortedMap(1 -> [3, 2])`

Well, this example is not correct because `[2,3]` is not equal to `[3,2]` but `SortedMap(1 -> p | q)` is equal to `SortedMap(1 -> q | p)`. 
It also depends which algorithm is used for equality check. For example, if you use Rholang's `==` then `p == q != q == p`. 

Anyway, I think that users should not observe the sortedness property of a Map. The sorting was added for the serialization sake and has now bubbled up up to the user.

IMO there should be an internal, sorted representation for purposes of `contains` and `equals` but `get` should return unsorted value.

ArturGajowy: > if you use Rholang's == then `p == q != q == p`

WDYM? Rholang's `==` is not symmetric?

Also, I'd totally expect `{ p | q } == { q | p }` in rholang.

> sorting was added for the serialization sake 

AFAIK it's also a crucial concept behind all equality and pattern matching (pattern matching is just equality whenever the pattern has no free variables or wildcards).

Does that change your view of things?

tymm: I agree with @goral09 that `SortedMap(1-> [2, 3])` should not be equal to `SortedMap(1-> [3, 2])` since a List should keep its order.
AFAIK this is also how it is implemented right now.

ArturGajowy: I also agree with him on that :) I really meant the example he's given.

`SortedMap(1 -> q | p) == SortedMap(1 -> p | q)` should be true in current implementation, and in general.

ArturGajowy: > IMO there should be an internal, sorted representation for purposes of contains and equals but get should return unsorted value.

If this was true, the following would be false:

```
Map(1 -> 3 | 2).get(1) == 3 | 2
```

That is, if I'm able to create an unsorted (sub)term in the first place.
The thing is, at times the terms are being made by Scala code other than the normalizer (reductions, substitutions, `Method` calls) and there's no easy way to guarantee their outputs are sorted.

Since all the terms that can be written are 'born' sorted, there's no error in only returning sorted terms from the map.

Since the dynamically created terms (e.g. method call results) can't be guaranteed to be sorted, sorting them on entry might be costly, but is safer.

goral09: That's what I said. Rholang's `==` (equality sign) doesn't support these properties (like spatial matcher does). It just uses Java equality and I think this is something users should be aware of.

ArturGajowy: Since there's no way to have unsorted terms written in soruce code, and all rholang's internals should be returning sorted terms, there's nothing for the users to be aware of. 

*Throws chair.*

goral09: > Since there's no way to have unsorted terms written in soruce code,

That's not true. IIRC sorting happens only after substitution. It means that if you write down concrete, unsorted term then it stays unsorted.

ArturGajowy: Nopes, buildNormalizedTerm also sorts (see buildPar) :) I think we can finally agree :)

goral09: Doesn't it sort only the outermost elements? Not the inner ones? Like elements of the collection for example.

ArturGajowy: Nopes, sorting is recursive.

goral09: Then why do we have to sort at all when interacting with Map if everything is already sorted?

ArturGajowy:
As I said, we don't have such guarantees when getting dynamically-generated
terms (subtitutions, Method results). I'd like the map to sort again until
we find a way to encode the sortedness in types or otherwise guarantee
sortedness of dynamically generated terms. Unless we prove that this is a
huge performance drag (i.e. have an e2e benchmark for that).

models/src/main/scala/coop/rchain/models/SortedParMap.scala
ArturGajowy: As I said, we don't have such guarantees when getting dynamically-generated
terms (subtitutions, Method results). I'd like the map to sort again until
we find a way to encode the sortedness in types or otherwise guarantee
sortedness of dynamically generated terms. Unless we prove that this is a
huge performance drag (i.e. have an e2e benchmark for that).

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/test/scala/coop/rchain/models/SortedParMapSpec.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Done.

models/src/main/scala/coop/rchain/models/SortedParMap.scala
tymm: Not instead but in addition to. However seems like it is not needed for our purposes.

models/src/main/scala/coop/rchain/models/rholang/sorter/ordering.scala
```diff
@@ -6,12 +6,24 @@ import coop.rchain.models.rholang.sorter.ScoredTerm._
...
+  implicit class ParOps(par: Par) {
+    implicit val sync = implicitly[Sync[Coeval]]
+
+    def sort: Par = Sortable[Par].sortMatch[Coeval](par).map(_.term).value()
+  }
+
+  implicit class GenTraversableOnceOps(ps: GenTraversableOnce[Par]) {
+    implicit val sync = implicitly[Sync[Coeval]]
+
+    def sort: GenTraversableOnce[Par] = ps.toList.sort
```
ArturGajowy: Please remove those implicit conversions - I'd like Sortable to be the entry point for sorting for all new usages, as it's explicit about monads being used underneath. At some point we're going to have to join all the computations into a single one, in a single monad - otherwise performance will continue to suffer and there's no stacksafety

ArturGajowy: parOps.sort should be just sort(Par) - a private method in the SortedMap.

GenTraverasableOps is not needed at all - it's enough if we just:
`kvs.foldLeft(sortedMap) { map, kv -> sortedMap - kv }` in the `--` mehtod

PR #1827, opened 2018-11-13
Test fixes

ArturGajowy:
## Overview
Title says it all :)

Notice the sbt-repeat plugin (see commit log)

### JIRA issue:
n/a

### Checklist


ArturGajowy:
I'm disabling parallel tests because most of the issues with tests I've seen can't be reproduced by simply running that single test. This can mean that our tests simply aren't suited for being run in parallel. It can also mean there are legit errors in some cases, but due to them being non-deterministic, and there being lots of failures (both legit and setup-related), we shouldn't be fighting them all at once (unless we want all hands on board with that).

As suggested by @dzajkowski, we probably would benefit from a separate, optional CI check - that would allow us to discover, ticket, and schedule fixing those issues in a more sustainable pace until we make it all stable and can make the parallel tests check required.

PR #1828, opened 2018-11-13
Backpressure for concurrent message queues

sebtomba:
## Overview
Monix doesn't support concurrent (buffered) `Subjects` with a backpressure overflow strategy. This PR employs a semaphore with a maximum parallelism of 1 to implement a `Sequencer`. This `Sequencer` allows threads to push messages concurrently to a subject. However, the push operation does not complete until the subject accepts the message. Until then the pushing thread gets blocked, which is sort of backpressure.
**Caveat:** The `Sequencer` should be operated by a strictly limited number of threads (or fibers) to achieve a real backpressure. Otherwise, the system will continue to produce new messages on new threads (or fibers) without slowing down in case of backpressure.

https://rchain.atlassian.net/browse/CORE-1415

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/buffer/Sequencer.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/buffer/Sequencer.scala
@@ -1,0 +1,62 @@
+package coop.rchain.comm.transport.buffer
+
+import scala.concurrent.Future
+
+import cats.implicits._
+
+import monix.catnap.Semaphore
+import monix.eval.Task
+import monix.execution._
+import monix.reactive._
+import monix.reactive.observers.{BufferedSubscriber, Subscriber}
+import monix.reactive.subjects._
+import monix.reactive.OverflowStrategy._
+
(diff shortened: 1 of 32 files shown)


sebtomba:
@rabbitonweb CI passed

comm/src/main/scala/coop/rchain/comm/transport/buffer/Sequencer.scala
```diff
@@ -0,0 +1,62 @@
...
+import monix.catnap.Semaphore
+import monix.eval.Task
+import monix.execution._
+import monix.reactive._
+import monix.reactive.observers.{BufferedSubscriber, Subscriber}
+import monix.reactive.subjects._
+import monix.reactive.OverflowStrategy._
+
+trait Sequencer[I, +O] extends Observable[O] {
+  def pushNext(elem: I): Task[Ack]
```
rabbitonweb: @sebtomba why this is returning Ack not Boolean as it was with buffers?

PR #1841, opened 2018-11-15
Add back self-validation for blocks temporarily

KentShikama:
This will make it easier to debug validation issues


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -118,8 +118,9 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
                 else if (!validSig) InvalidUnslashableBlock.pure[F]
                 else if (!validSender) InvalidUnslashableBlock.pure[F]
                 else if (!validVersion) InvalidUnslashableBlock.pure[F]
-                else if (validatorId.exists(id => ByteString.copyFrom(id.publicKey) == b.sender))
-                  addEffects(Valid, b).map(_ => Valid)
+                // TODO: Re-enable once node is more stable
+                //else if (validatorId.exists(id => ByteString.copyFrom(id.publicKey) == b.sender))
+                //  addEffects(Valid, b).map(_ => Valid)
```
ArturGajowy: oh, I actually thought it's the other way around. This seems like to remove checks instead of adding them?

Anyway, please file a JIRA issue for that TODO - otherwise it's going to live forever :)

KentShikama: No `addEffects(Valid, b).map(_ => Valid)` means you add the block to the dag with a status of "Valid" without bothering to actually check whether it is valid. This was an optimization as you in theory you don't need to validate the blocks you create yourself - as you can trust yourself to not be malicious.

KentShikama: Any block that would have gone to this case is now pushed through the `attemptAdd` below, which actually does a bunch of validation.

KentShikama: Done https://rchain.atlassian.net/browse/RHOL-1064

itegulov: I remember having some difficulties debugging because of this, so I agree with temporarily disabling this optimization.

KentShikama:
@ArturGajowy @itegulov I forgot to run the casper/test locally when I pushed this (assuming there would be nothing - but of course there was). Basically what happened was that any test that used `addBlock` did nothing thanks to this optimization and so when we added the validation back, there was an issue with blockTime that we had not caught originally.

KentShikama:
There are 3 other failing tests excluding the blockTime one but those are more like "checking for the wrong log message" rather than a bug discovery. I'll handle those soon.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
ArturGajowy: > This was an optimization as you in theory you don't need to validate the blocks you create yourself - as you can trust yourself to not be malicious.

Having the self-validation is an important self-check then that should only be disabled for development / bug tracking :) In case of a bug, I'd rather see the node self-validate and fail the deployment than push the faulty block and get slashed :)

KentShikama: @ArturGajowy I know we had a push to increase performance by not self-validating but I don't think the minor performance gain will ever be worth the risk pushing a faulty block out into the network. So yeah, I'll make this change permanent.

PR #1718, opened 2018-10-21
Ops 308 fix integration tests after the changes in show-blocks needed.

zsluedem:
## Overview
fix integration tests after the changes in show-blocks needed.

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please 
https://rchain.atlassian.net/browse/OPS-308

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [ ] If this PR adds a new feature, this PR includes tests related to this feature.
- [ ] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [x] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [x] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1842, opened 2018-11-15
[OPS-308] Fix integration tests after the changes in show-blocks

KentShikama:
This is a follow up from #1718


KentShikama:
Paused until issues that have come out with https://github.com/rchain/rchain/pull/1841 have been resolved.

PR #1873, opened 2018-11-16
Unforgeable preview 752

dckc:
## Overview

The previewPrivateNames rpc allows gRPC clients to know the id of to an unforgeable name in advance of deployment -- for example, to compute a signature over it.

https://rchain.atlassian.net/browse/RHOL-752

- [X] If this PR adds a new feature, this PR includes tests related to this feature.

- [X] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [X] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.


casper/src/test/scala/coop/rchain/casper/api/PreviewPrivateNameTest.scala
```diff
@@ -0,0 +1,58 @@
...
+    // When we deploy `new x ...` code from a javascript gRPC client,
+    // we get this private name id in the log:
+    // 16:41:08.995 [node-runner-15] INFO  c.r.casper.MultiParentCasperImpl - Received Deploy #1542308065454 -- new x0, x1 in {
+    //   @{x1}!(...
+    // [Unforgeable(0xb5630d1bfb836635126ee7f2770873937933679e38146b1ddfbfcc14d7d8a787), bundle+ {   Unforgeable(0x00) }]
+    // 2018-11-15T18:54:25.454Z
+    previewId(myNodePk, 1542308065454L) should equal(
+      "b5630d1bfb836635126ee7f2770873937933679e38146b1ddfbfcc14d7d8a787"
+    )
+  }
```
KentShikama: Does this work for predicting `x2` in

`new x0, x1 in { x0!(5) | for (_ <- x0) { new x2 in { ... } } }`?

dckc: I doubt it.

I suppose that should be documented better.

Meanwhile, I'll test to see if I get lucky...

dckc: Nope. The preview gRPC gives:

```
[ '64b2d2e4dcbb7a0d691431ea0aa270f88f3cfa19bd4740cfced007e578bc98a3',
  'de727884e43d52a7b8b9c532f4e6634212243fd2567fa4c6e1709a48051deef9',
  '57ec7b8bd6491bf0dbf31b0de2c4e850dc5a07941ba50b8ab5362c412ff30749' ]
```

but when I put `stdout!([*x0, *x1, *x2])` in for `...` in the above, I get:
```
[Unforgeable(0x64b2d2e4dcbb7a0d691431ea0aa270f88f3cfa19bd4740cfced007e578bc98a3),
 Unforgeable(0xde727884e43d52a7b8b9c532f4e6634212243fd2567fa4c6e1709a48051deef9),
 Unforgeable(0x0ab721c6da23a6d2ad1d62cd1fea661cc860f6e3d5cc7aa042d442109b5d109f)]
```

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -477,4 +478,19 @@ object BlockAPI {
...
+      timestamp: Long,
+      nameQty: Int
+  ): F[PrivateNamePreviewResponse] = {
+    val seed    = DeployData().withUser(user).withTimestamp(timestamp)
+    val rand    = Blake2b512Random(DeployData.toByteArray(seed))
+    val safeQty = nameQty min 1024
+    val ids = for {
+      _  <- 0 until safeQty
+      id = ByteString.copyFrom(rand.next())
+    } yield id
```
KentShikama: Can we use forEach or map here? I think we generally don't want to use List as a monad.

casper/src/test/scala/coop/rchain/casper/api/PreviewPrivateNameTest.scala
goral09: It won't work because the seed is being split for `send`s and then merged for `consume`s.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
dckc: @goral09 is there some benefit to writing the code that way? It's not self-explanatory, to me.

casper/src/test/scala/coop/rchain/casper/api/PreviewPrivateNameTest.scala
dckc: right... figuring out where the splits and merges happen is basically equivalent to writing a rholang interpreter

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
goral09: @dckc there is a rule to use least powerful abstraction. `Functor` (map) is less powerful than `Monad` (flatMap). The other thing is that I treated `rand.next()` an effectful call (which I am not sure it is now). Let's assume for now that it is. Effectful calls should be "suspended" (hence the usage of `Sync[F].delay(…)`). If you run a method `A => F[B]` over the list of `A`s you end up with `List[F[B]]` but what you want is `F[List[B]]` and this is where `traverse` goes into play. Its signature is `List[A] => (A => F[B]) => F[List[B]]`.

dckc: OK, I see now.

`rand.next()` definitely mutates rand; that's an effect, right? otherwise, `rand.next()` would have to return the same thing every time, right? On the other hand, the state in `rand` doesn't escape this `previewPrivateNames` method, so I don't think sychronization is necessary.

PR #1884, opened 2018-11-19
CORE-1450 Exist with system 1 when encountered errors (deploy runtime)

rabbitonweb:
## Overview
When program attempts and error is caught, we should exit with system
code > 0.

CORe-1450

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
@@ -107,7 +107,10 @@ object DeployRuntime {
     for {
       result <- program.attempt
       _ <- result match {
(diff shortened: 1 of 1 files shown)


rabbitonweb:
This has to hold, because apparently not returning 1 when response returns an error

rabbitonweb:
@adaszko @dzajkowski 

It seems that the change in this PR works correctly. See below

```
➜  rchain git:(core-1450-exist-1) ✗ cd node/target/universal/stage 
➜  stage git:(core-1450-exist-1) ✗ ./bin/rnode propose
Error: Connection refused: localhost/0:0:0:0:0:0:0:1:40401
➜  stage git:(core-1450-exist-1) ✗ echo $?                       
1
```

rabbitonweb:
I'm merging as it is because it seems good enough to enter `dev`. If we will need more, I will create a next PR.

PR #1885, opened 2018-11-19
Disconnect from peers and clear connection when node's dynamic IP changes

sebtomba:
## Overview
In case the dynamic IP address changes send a `Disconnect` message to all peers, clear RP connections and `TransportLayer` connections cache.

https://rchain.atlassian.net/browse/CORE-1418


sebtomba:
Reverted some changes and merged the dynamic IP loop with the clear connections/find new peers loop. Changed sleep interval to 1 minute.

comm/src/test/scala/coop/rchain/comm/connect/ClearConnectionsSpec.scala
```diff
@@ -97,45 +97,68 @@ class ClearConnectionsSpec
...
         val cleared = Connect.clearConnections[Id]
-        // tehn
-        cleared shouldBe (1)
+        // then
+        cleared shouldBe 1
+      }
+    }
+
+    describe("if dynamic local host changes") {
+      it("should disconnect from all peers and clear connections") {
```
rabbitonweb: This should be in its own test `ResestConnectionsSpec` as "reset" has nothing to do with "dynami host change". We *know* we call it when host will change dynamically, but that is not implied in comm layer nor in tests. Description does not match the "//given". I would just move it to its own spec and defined that this is feature that resets the connections.

PR #1889, opened 2018-11-20
Fix multiple bugs in test_heterogenous_validators

adaszko:


integration-testing/rnode_testing/rnode.py
```diff
@@ -297,6 +296,7 @@ def create_peer(
     memory="1024m",
     cpuset_cpus="0",
 ):
+    assert '_' not in name, 'Underscore is not allowed in host name'
```
lukasz-golebiewski: Could this be moved to a utility method? DRY

adaszko: Will do that in subsequent PRs.  I want to have the most pressing bugs fixed first.

PR #1892, opened 2018-11-20
CORE-1417 - Client queue must be able to store Blobs to disk

rabbitonweb:
## Overview
In order to not get OOME, all messages that are accepted by client queue (see CORE-1383 ) need to be stored on file system. 
In this PR we are storing every single message to disk. We can potentially store messages which are bigger then some significant size (e.g 1MB). This way we know that memory will not explode, the queue will stay not bigger thne queue_size x max_message_size (for queue of 1024 elements, and max message size 1mb the memory footprint will be ~1GB)

CORE-1417

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/errors.scala
+++ b/comm/src/main/scala/coop/rchain/comm/errors.scala
@@ -1,5 +1,6 @@
 package coop.rchain.comm
 
+import java.nio.file._
 import coop.rchain.comm.protocol.routing._
 import cats._, cats.data._, cats.implicits._
 import coop.rchain.catscontrib._, Catscontrib._, ski._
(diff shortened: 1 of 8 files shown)


comm/src/main/scala/coop/rchain/comm/errors.scala
```diff
@@ -1,35 +1,38 @@
 package coop.rchain.comm
 
+import java.nio.file._
 import coop.rchain.comm.protocol.routing._
 import cats._, cats.data._, cats.implicits._
 import coop.rchain.catscontrib._, Catscontrib._, ski._
 
 // TODO we need lower level errors and general error, for now all in one place
 // TODO cleanup unused errors (UDP trash)
 sealed trait CommError
```
dzajkowski: `sealed abstract class` is said to be better

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -309,7 +308,8 @@ class NodeRuntime private[node] (
       port,
       conf.tls.certificate,
       conf.tls.key,
-      conf.server.maxMessageSize
+      conf.server.maxMessageSize,
+      conf.server.dataDir.resolve("comm-temporary/")
```
dzajkowski: maybe it's a good opportunity to introduce a naming scheme?
e.g.
`tmp/comm`?

comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -9,17 +11,24 @@ import monix.reactive.Observable
...
 
   def stream(peers: List[PeerNode], blob: Blob): Task[Unit] = {
     def push(peer: PeerNode): Task[Boolean] =
-      Task.delay(subject.pushNext(ToStream(peer, blob)))
+      blob.packet.store[Task](folder) >>= {
+        case Right(file) => Task.delay(subject.pushNext(ToStream(peer, file, blob.sender)))
+        case Left(UnableToStorePacket(p, er)) =>
+          log.error(s"Could not serialize packet $p. Error message: $er") *> true.pure[Task]
+        case Left(er) =>
+          log.error(s"Could not serialize packet ${blob.packet}. Error: $er") *> true.pure[Task]
```
dzajkowski: `true` confuses me. could you explain?

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -205,12 +205,22 @@ class TcpTransportLayer(port: Int, cert: String, key: String, maxMessageSize: In
...
+          }.attempt
+            .flatMap {
+              case Left(error) => log.error(s"Error while streaming packet, error: $error")
+              case Right(_)    => Task.unit
+            }
+            .flatMap(kp(Task.delay {
+              if (path.toFile.exists)
+                path.toFile.delete
+            }))
+        case Left(error) => log.error(s"Error while streaming packet, error: $error")
```
dzajkowski: so a failed restore will never be cleared?

comm/src/main/scala/coop/rchain/comm/errors.scala
rabbitonweb: how come?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
rabbitonweb: 👍

comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
rabbitonweb: yeah, I thought `true` might be confusing. 
`push` returns `Boolean`, if it returns `false` then the given packet will be 'retried'. However if we failed to store packet on disk once, we will probably fail again, thus eventually the whole function will loop forever
There is a ticket to introduce sensible retries for stream API (put the message back on the queue, retry few times and discard if impossible - this will be addressed). For now it's safer to log error and drop the message.

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
rabbitonweb: you sir are correct, i will issue a ticket to have this functionality tested as well

comm/src/main/scala/coop/rchain/comm/errors.scala
ArturGajowy: It's needed if you want proper stacktraces. Try printing a stacktrace for any of the cause-containing errors and you'll see they're broken and are missing debug info.

We've been through that in rholang.

See InterpreterError and subclasses

```scala
  sealed abstract class InterpreterError(message: String) extends Throwable(message) {

    def this(message: String, cause: Throwable) {
      this(message)
      initCause(cause)
    }
  }
```

ArturGajowy: OTOH you're not extending exception and all stacktraces are printed (if at all) from the underlying exceptions, so I'd stay with a trait here. Unless there's another reason, @dzajkowski?

dzajkowski: I don't want to block the pr, will link to resource.

PR #1901, opened 2018-11-22
Revert to use coarser grained locking for addBlock

KentShikama:
In theory, addBlock can be called with multiple threads
if they are processing different blocks but there seems
to be some evidence that this locking algorithm isn't working.

See https://rchain.atlassian.net/projects/CORE/issues/CORE-1454

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -26,6 +26,7 @@ import coop.rchain.catscontrib.TaskContrib._
 
 import scala.collection.immutable.HashSet
 import scala.collection.mutable
+import scala.concurrent.SyncVar
 
 class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer: Log: Time: ErrorHandler: SafetyOracle: BlockStore: RPConfAsk](
     runtimeManager: RuntimeManager,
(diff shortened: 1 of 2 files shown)


KentShikama:
Hold the review please; I realized I need to check that we don't add the same block twice.

KentShikama:
Ok ready again

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -68,43 +69,27 @@ class MultiParentCasperImpl[F[_]: Sync: Capture: ConnectionsCell: TransportLayer
...
-                   Log[F]
-                     .warn(
-                       s"Block ${PrettyPrinter.buildString(b.blockHash)} encountered an exception during processing: ${ex.getMessage}"
-                     )
-                     .map(_ => BlockStatus.exception(ex))
+                   .map(_ => BlockStatus.processing)
+               } else {
+                 internalAddBlock(b)
                }
+      _ <- Sync[F].delay(processingBlock.put(()))
```
lukasz-golebiewski: Are you sure this line will always get executed (even when an exception between lines 79 and 90 is thrown?). Otherwise a deadlock might occur

KentShikama: If internalAddBlock throws an exception, might as well take down the node

PR #1903, opened 2018-11-22
Stacksafe hashcode

ArturGajowy:
## Overview
Title says it all :)

### JIRA issue:
RHOL-992 Make AST equals stacksafe and RHOL-865 Make AST hashCode stacksafe

### Checklist

### DIFF
--- a/models/src/main/scala/coop/rchain/models/EqualsM.scala
+++ b/models/src/main/scala/coop/rchain/models/EqualsM.scala
@@ -1,0 +1,159 @@
+package coop.rchain.models
+
+import cats.effect.Sync
+import cats.implicits._
+import com.google.protobuf.ByteString
+import coop.rchain.casper.protocol._
+import coop.rchain.catscontrib.Catscontrib._
+import coop.rchain.crypto.hash.Blake2b512Random
+import monix.eval.Coeval
+
+import scala.Function.tupled
+import scala.collection.immutable.BitSet
+
+/**
+  * A typeclass for computing Scala's default `equals` for case classes in a stacksafe manner.
+  *
+  * An instance of this class must be provided for every case where `equals` is overridden.
+  * The generative tests should catch that, provided the offending case class is used transitively.
+  *
+  * @tparam A
+  */
+trait EqualM[A] {
+
+  def equal[F[_]: Sync](self: A, other: A): F[Boolean]
+
+  def equals[F[_]: Sync](a: A, b: Any): F[Boolean] =
+    if (a.getClass.isInstance(b)) {
+      equal[F](a, a.getClass.cast(b))
+    } else false.pure[F]
+
+}
+
+object EqualM extends EqualMDerivation {
+
+  def apply[A](implicit ev: EqualM[A]): EqualM[A] = ev
+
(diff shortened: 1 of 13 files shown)


models/src/test/scala/coop/rchain/models/HashMSpec.scala
```diff
@@ -0,0 +1,96 @@
...
+      case x: ParSet          => Objects.hash(x.ps, x.remainder, Boolean.box(x.connectiveUsed))
+      case x: ParMap          => Objects.hash(x.ps, x.remainder, Boolean.box(x.connectiveUsed))
+      case caseClass: Product => ScalaRunTime._hashCode(caseClass)
+      case _                  => a.hashCode()
+    }
+    // With this check we know that:
+    //   HashM.hash[Id] == _.hashCode == reference
+    // This makes this test valid both before and after we override the generated hashCode
+    // (which is long done when you're reading this).
+    reference should be(a.hashCode())
```
tymm: Could you explain why this is a useful test? I don't get it.

models/src/main/scala/coop/rchain/models/HashM.scala
```diff
@@ -0,0 +1,207 @@
...
+      for {
+        componentHashes <- hashFs.toList.sequence
+      } yield productHash(ctx.typeName.short, componentHashes)
+    }
+
+  }
+
+  //copied and adapted from scala.util.hashing.MurmurHash3,
+  //which is used in Scala's case class hash code implementation
+  private def productHash(prefix: String, elements: Seq[Int]): Int = {
```
tymm: Probably overkill so feel free to ignore but maybe it's worth it to write a test that proofs that this is not generating too many collisions.

ArturGajowy: The primary goal is to have it consistent with the default hashCode we had previously, even if it had excessive collisions (which I doubt). Also, since this is copied from scala std lib, I'd prefer to have at least a hint of experience that the collisions are excessive before testing / investigating / defining that :)

models/src/main/scala/coop/rchain/models/EqualsM.scala
```diff
@@ -0,0 +1,159 @@
...
+  implicit val ReceiveBindEqual      = gen[ReceiveBind]
+  implicit val NewEqual              = gen[New]
+  implicit val MatchEqual            = gen[Match]
+
+  implicit val ConnectiveEqual = gen[Connective]
+
+  implicit val ESetEqual = gen[ESet]
+  implicit val EMapEqual = gen[EMap]
+
+  implicit val SortedParHashSetEqual: EqualM[SortedParHashSet] = by(_.sortedPars)
```
tymm: Why not do `gen[SortedParHashSet]`? Now we have the logic of how to do `equal` for `SortedParHashSet` in two places. Same goes for `SortedParMapEqual`.

ArturGajowy: `gen` is only doing what we should be doing in the first place - implementing equals/hashCode using all the fields. We have a custom `equals` for those types though, and thus the corresponding typeclass instances need to be custom too :)

PR #1912, opened 2018-11-25
CORE-1438: Open some JDK modules in startup script

woky:
Modify startup script to check if the JVM version is >= 9 and in that
case to open java.base module to unnamed modules, to get rid of illegal
access warnings like this:

  WARNING: An illegal reflective access operation has occurred
  WARNING: Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/usr/share/rnode/lib/com.google.protobuf.protobuf-java-3.5.1.jar) to field java.nio.Buffer.address
  WARNING: Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil
  WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
  WARNING: All illegal access operations will be denied in a future release

I tested validator node run with and deploy and propose of dup.rho on
dev bootstrap without getting any WARNING.

Required read: https://rchain.atlassian.net/wiki/spaces/CORE/pages/629243905/Java+9+Modularity+WARNING+An+illegal+reflective+access

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -199,6 +199,34 @@ lazy val node = (project in file("node"))
         val oldStrategy = (assemblyMergeStrategy in assembly).value
         oldStrategy(x)
     },
+    /*
+     * This monstrosity exists because
+     * a) we want to get rid of annoying JVM >= 9 warnings,
+     * b) we must support Java 8 for RedHat (see below) and
+     * c) sbt-native-packager puts bashScriptExtraDefines before it
+     *    initializes all useful variables (like $java_version).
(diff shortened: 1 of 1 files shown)


woky:
While this is a bit ugly, it's least intrusive way to get rid of JVM warnings. Once this is merged I'm going to submit RFE with link to this commit to sbt-native-packager to make another setting similar to _bashScriptExtraDefines_ that'll be put right before Java launch to at least get rid of that Bash hack.

PR #1929, opened 2018-11-28
OPS-375 Collect type signatures for mypy

zsluedem:
## Overview
Integration tests: Collect type signatures for mypy

https://rchain.atlassian.net/browse/OPS-375

### Notes


zsluedem:
@adaszko Needs a reviewer here. : )

integration-testing/rnode_testing/casper_propose_and_deploy.py
```diff
@@ -6,8 +6,13 @@
...
 from rnode_testing.random import random_string
 
+import typing
+if typing.TYPE_CHECKING:
+    from rnode_testing.rnode import Node
+    from conftest import TestConfig
+    from rnode_testing.network import RChain
 
-def deploy_block(node, expected_string, contract_name):
+def deploy_block(node: "Node", expected_string: str, contract_name: str) -> str:
```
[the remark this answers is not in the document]
zsluedem: You may want to take a look at [this issue](https://github.com/python/typing/issues/105)

I have encountered an import loop error using the annotation before. The easy way to get rid of it is to import `TYPE_CHECKING` and use `"Node"`.

integration-testing/test/conftest.py
```diff
@@ -75,7 +82,7 @@ def make_test_config(request):
 
 
 @contextlib.contextmanager
-def temporary_bonds_file(validator_keys):
+def temporary_bonds_file(validator_keys: List[KeyPair]) -> Iterator[str]:
```
adaszko: The return value is surprising to me.  Isn't there a specialized type for files?

integration-testing/test/conftest.py
```diff
@@ -88,7 +95,7 @@ def temporary_bonds_file(validator_keys):
 
 
 @contextlib.contextmanager
-def validators_data(config):
+def validators_data(config: TestConfig) -> Iterator[ValidatorsData]:
```
adaszko: The return type here should be https://docs.python.org/3/library/typing.html#typing.ContextManager

integration-testing/test/conftest.py
```diff
@@ -99,7 +106,7 @@ def validators_data(config):
 
 
 @pytest.yield_fixture(scope='session')
-def docker_client_session():
+def docker_client_session() -> Iterator["DockerClient"]:
```
adaszko: Iterator is too vague here, we would need a specific type for fixtures

integration-testing/test/conftest.py
adaszko: Would https://docs.python.org/3/library/typing.html#typing.IO work here?

integration-testing/rnode_testing/docker.py
```diff
@@ -2,9 +2,11 @@
 from contextlib import contextmanager
 import rnode_testing.random
 
+from docker.client import DockerClient
+from typing import Generator
 
 @contextmanager
-def docker_network(docker_client):
+def docker_network(docker_client: DockerClient) -> Generator[str, None, None]:
```
adaszko: `Generator` is too generic.  There must be some more specialized type that doesn't indicate a series of values is produces but just one.

integration-testing/test/conftest.py
zsluedem: Yeah. It is wrong. It should be `Iterator[IO]`

zsluedem: Nonono, the one I wrote is right. IT IS `Iterator[str]`.
The [file](https://github.com/rchain/rchain/pull/1929/files/fa8e119d27d18bd6de33818135af782356edc560#diff-a541af834ea45ee80dba232655f4558eR92) here is a str. Not an IO object.

integration-testing/test/conftest.py
zsluedem: No. If I change it to `ContextManager`, mypy would throw an error, `The return type of a generator function should be "Generator" or one of its supertypes`.

It is a generator func because you use `yield`, so the return type should be `Generator`. The `yield` object should be type inside the `[]`

integration-testing/test/conftest.py
zsluedem: See [the link ](https://github.com/python/mypy/issues/4709)

Maybe we can wrap these Iterator into one type. I give it a try.

integration-testing/rnode_testing/docker.py
zsluedem: Another return type for this one is `Iterator[str]`

zsluedem: For a function which would `yield`, the return type is either `Iterator[yield_object_type]` or `Generator[yield_object_type, None, None]`

More information for the generator should see this https://docs.python.org/3/library/typing.html#typing.Generator

zsluedem:
@adaszko you can run the mypy test to give yourself a sense of annotation. The command should be `pipenv run pytest --mypy -m mypy`.

integration-testing/rnode_testing/util.py
```diff
@@ -16,12 +19,14 @@ def log_box(log_function, title="", char="*", length=150):
...
 
 
-def make_tempfile(prefix, content):
+def make_tempfile(prefix: str, content: str) -> str:
     fd, path = tempfile.mkstemp(dir="/tmp", prefix=prefix)
 
     with os.fdopen(fd, 'w') as tmp:
         tmp.write(content)
 
+    return path
```
zsluedem: The mypy test did find some error. This function used to return nothing which the [rnode code](https://github.com/rchain/rchain/pull/1929/files#diff-154a13f8a2df4a0e8a88d1db476928aeR293) should expect a str path to return.

adaszko: The question is why did the tests pass.  The return value must have been unused.

integration-testing/rnode_testing/rnode.py
```diff
@@ -29,12 +38,12 @@ class InterruptedException(Exception):
 
 
 class NonZeroExitCodeError(Exception):
-    def __init__(self, command, exit_code, output):
+    def __init__(self, command: Tuple[Union[int, str], ...], exit_code: int, output: str):
```
adaszko: I think it should say:
```suggestion
    def __init__(self, command: Tuple[str, ...], exit_code: int, output: str):
```
I can help correcting places where `int` is passed.

integration-testing/rnode_testing/rnode.py
```diff
@@ -44,22 +53,22 @@ def __repr__(self):
 
 
 class TimeoutError(Exception):
-    def __init__(self, command, timeout):
+    def __init__(self, command: Iterator, timeout: int) -> None:
```
adaszko: Why `Iterator` and not `Tuple[...]`?

integration-testing/rnode_testing/rnode.py
```diff
@@ -176,71 +186,71 @@ def execution():
             process.join()
             raise TimeoutError(cmd, self.timeout)
 
-    def shell_out(self, *cmd, stderr=True):
+    def shell_out(self, *cmd: Union[str, int], stderr=True) -> str:
```
adaszko: Again, we should correct the code so that the type declaration can read: 
```suggestion
    def shell_out(self, *cmd: Union[str], stderr=True) -> str:
```

integration-testing/rnode_testing/rnode.py
```diff
@@ -176,71 +186,71 @@ def execution():
...
         exit_code, output = self.exec_run(cmd, stderr=stderr)
         if exit_code != 0:
             raise NonZeroExitCodeError(command=cmd, exit_code=exit_code, output=output)
         return output
 
-    def call_rnode(self, *node_args, stderr=True):
+    def call_rnode(self, *node_args: Union[str, int], stderr: bool = True) -> str:
```
adaszko: Similarly to my comments above:
```suggestion
    def call_rnode(self, *node_args: Union[str], stderr: bool = True) -> str:
```

integration-testing/rnode_testing/rnode.py
```diff
@@ -252,27 +262,27 @@ def run(self):
...
-):
+    docker_client: "DockerClient",
+    name: str,
+    network: str,
+    bonds_file: str,
+    container_command: str,
+    container_command_options: Dict,
+    rnode_timeout: int,
+    extra_volumes: List[str],
+    allowed_peers: Optional[List[str]],
```
adaszko: 👍

integration-testing/rnode_testing/rnode.py
```diff
@@ -470,7 +480,7 @@ def create_peer_nodes(
 
 
 @contextmanager
-def bootstrap_node(docker, docker_network, timeout, validators_data, *, container_name=None, cli_options=None, mount_dir=None):
+def bootstrap_node(docker: "DockerClient", docker_network: str, timeout: int, validators_data: "ValidatorsData", *, container_name: str = None, cli_options=None, mount_dir: str = None) -> Generator[Node, None, None]:
```
adaszko: Why `Generator[Node, None, None]` and not `Generator[Node]`?

zsluedem: The `Generator` expects 3 type arguments. If you put only one argument, mypy would report as an error.

integration-testing/rnode_testing/rnode.py
zsluedem: same above

integration-testing/rnode_testing/rnode.py
zsluedem: It should be `Union[Tuple[str, ...], str]`

integration-testing/rnode_testing/rnode.py
```diff
@@ -104,59 +118,61 @@ def __init__(self, container, deploy_dir, docker_client, timeout, network):
...
 
-    def get_rnode_address(self):
+    def get_rnode_address(self) -> str:
         log_content = self.logs()
         m = re.search("Listening for traffic on (rnode://.+@{name}\\?protocol=\\d+&discovery=\\d+)\\.$".format(name=self.container.name),
                       log_content,
                       re.MULTILINE | re.DOTALL)
-        address = m[1]
+        if m is None:
+            raise NoMatchException()
```
adaszko: It is a great idea to have exception raised here.  However, it would be nice to give it a name which indicates clearly what has happened, something like:
```suggestion
            raise RNodeAddressNotFoundError()
```

integration-testing/rnode_testing/rnode.py
```diff
@@ -30,13 +39,17 @@ class InterruptedException(Exception):
     pass
 
 
+class NoMatchException(Exception):
```
adaszko:
```suggestion
class RNodeAddressNotFoundError(Exception):
```

zsluedem:
@adaszko After some tries, I think we'd better use `Generator` for the yield function which would be easier to understand.

integration-testing/rnode_testing/network.py
```diff
@@ -1,13 +1,17 @@
...
 
+from typing import List, TYPE_CHECKING, Callable, Any, Optional, Generator
+
+if TYPE_CHECKING:
+    from conftest import TestConfig, ValidatorsData
+    from docker.client import DockerClient
+    from rnode_testing.rnode import Node
 
-def make_wrapper(fn, fixtures):
+def make_wrapper(fn: Callable, fixtures: Any):
```
dckc: what are the args and return type of the Callable? Does it not matter?

integration-testing/rnode_testing/wait.py
```diff
@@ -54,13 +60,14 @@ def wait_for(condition, timeout, error_message):
 # Warning: The __doc__ has to be explicitly assigned as seen below if it's a formatted string, otherwise it will be None.
 
 
-def node_logs(node):
+def node_logs(node: "Node") -> Callable:
```
dckc: again, does it not matter what the args and return type are of the Callable?

zsluedem:
@dckc @adaszko Give it a look~! I hope it is the last commit for this pr~! : )

zsluedem:
@adaszko add the env `MYPYPATH={project_path}/test` when you run the mypy test which is `MYPYPATH={project_path}/test pipenv run pytest --mypy -m mypy`

zsluedem:
@adaszko I believe all the `type signatures` should be collected right which is all the ticket want. I hope this pr can be merged as soon as possible. I don't want to deal with any conficts merge again which is really painful. If you want to get rid of the import error , we should create another ticket for that.

PR #1930, opened 2018-11-29
Store channels' hashes separately in Consume events

lukasz-golebiewski:
## Overview
This is needed to fix the conflicts resolution algorithm in ProtoUtils

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
https://rchain.atlassian.net/browse/CORE-1480

### Complete this checklist before you submit the PR
- [x] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] If this PR adds a new feature, this PR includes tests related to this feature.
- [x] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [ ] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [ ] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #1934, opened 2018-11-30
Modify ProtoUtil.conflicts to use replay logs

KentShikama:
https://rchain.atlassian.net/browse/RHOL-1084

~~This will require a small modification to include
https://github.com/rchain/rchain/pull/1930 that
involves un-ignoring one test~~


KentShikama:
Incorporated https://github.com/rchain/rchain/pull/1930

@lukasz-golebiewski @dzajkowski PTAL

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -258,43 +260,54 @@ object ProtoUtil {
...
-   * TODO: Update the logic of this function to make use of the trace logs and
-   * say that two blocks don't conflict if they act on disjoint sets of channels
+   * Block b1 conflicts with b2 if any of b1's ancestors contains a replay log entry that
+   * touches a channel that any of b2's ancestors' (that are not common with b1's ancestors)
+   * replay log entries touch.
    */
   def conflicts[F[_]: Monad: BlockStore](
       b1: BlockMessage,
       b2: BlockMessage,
       genesis: BlockMessage,
```
lukasz-golebiewski: `genesis` isn't used anymore

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -258,43 +260,54 @@ object ProtoUtil {
...
       dag: BlockDag
-  ): F[Boolean] =
+  ): F[Boolean] = {
+    val b1MetaData       = dag.dataLookup.get(b1.blockHash)
+    val b2MetaData       = dag.dataLookup.get(b2.blockHash)
+    val blockMetaDataSeq = Vector(b1MetaData, b2MetaData).flatten
+    if (blockMetaDataSeq.size == 2) {
```
lukasz-golebiewski: I would be for raising an error when one (or both) of the blocks isn't present in the DAG. Otherwise the program might fail elsewhere with a completely unrelated reason

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -258,43 +260,54 @@ object ProtoUtil {
...
+      maybeAncestors <- blockAncestorsMeta.traverse(
+                         blockAncestorMeta => BlockStore[F].get(blockAncestorMeta.blockHash)
+                       )
+      ancestors      = maybeAncestors.flatten
+      ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.log))
+      ancestorChannels = ancestorEvents.flatMap {
+        case Event(Produce(produce: ProduceEvent)) =>
+          Set(produce.channelsHash)
+        case Event(Consume(consume: ConsumeEvent)) =>
+          consume.channelsHashes.toSet
```
lukasz-golebiewski: After seeing what you are doing here I guess it would make sense to change the type of the `channelsHashes` container to `Set` in `Consume`. Could be done in a separate PR

KentShikama:
@lukasz-golebiewski No the point is that those DAG shapes are invalid. It shouldn't be allowed to merge from two blocks that touch the same name.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
KentShikama: We need them ordered in the proto though because these Events all get hashed and the block hash needs to be deterministic.

KentShikama:
https://github.com/rchain/rchain/pull/1934/files#diff-5a4ef740dcf915d23a895682cd049c9eR362 essentially creates the same test as https://github.com/rchain/rchain/pull/1936/files#diff-a5cbe4dc99e7d8278bd8b7abca31a075R314 at the casper level (it uses a different contract) and shows that it will only take one parent

KentShikama:
@lukasz-golebiewski PTAL

PR #1946, opened 2018-12-03
Fix issue with reactive streams observer

sebtomba:
## Overview
We observed the following error when using `show-blocks`:
`onSubscription never happened, see rule 2.13 in the Reactive Streams spec`


casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -58,23 +58,27 @@ class GrpcDeployService(host: String, port: Int, maxMessageSize: Int)
...
       val showLength =
         s"""
-           |count: ${response.length}
+           |count: ${bs.length}
            |""".stripMargin
 
-      Right(showResponses + "\n" + showLength)
+      Right(bs.mkString("\n") + "\n" + showLength)
     }
+  }
```
dzajkowski: is this additional named variable needed? `.toListL.map` would also work.

PR #1956, opened 2018-12-04
CORE-1510 Renamed FineGrained RSpace to just RSpace.

dzajkowski:
## Overview
Cleanup. Naming back to normal.

CORE-1510


rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
```diff
@@ -51,7 +81,7 @@ abstract class RSpaceOps[F[_], C, P, E, A, R, K](val store: IStore[C, P, A, K],
       throw new IllegalArgumentException(msg)
     }
     logger.debug(s"""|install: searching for data matching <patterns: $patterns>
-                     |at <channels: $channels>""".stripMargin.replace('\n', ' '))
+                       |at <channels: $channels>""".stripMargin.replace('\n', ' '))
```
lukasz-golebiewski: Misalignment (also in a few other places)

PR #1964, opened 2018-12-06
Log if there is a doppelganger on the network

KentShikama:
This might be the cause of https://rchain.atlassian.net/browse/RHOL-1085

Relative to #1963

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
@@ -32,6 +32,7 @@ import monix.execution.Scheduler
 import monix.execution.Scheduler.Implicits.global
 import org.scalatest.{Assertion, FlatSpec, Matchers}
 import coop.rchain.casper.scalatestcontrib._
+import coop.rchain.catscontrib.ski.kp2
 import coop.rchain.metrics.Metrics
 
 import scala.collection.immutable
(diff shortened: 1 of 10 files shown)


casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -165,7 +171,7 @@ class HashSetCasperTest extends FlatSpec with Matchers {
       _                 <- casperEff.deploy(registerDeploy)
       createBlockResult <- casperEff.createBlock
       Created(block)    = createBlockResult
-      blockStatus       <- casperEff.addBlock(block)
+      blockStatus       <- casperEff.addBlock(block, kp2(().pure[Effect]))
```
ArturGajowy: please replace all repetitions of this expression with an usage of a

`val ignoreDoppelgangers = kp2(().pure[Effect]) //FIXME replace with proper handler`

placed in some public object

ArturGajowy: Also, this handler is being passed so often I'm starting to think it should be a capability required for creating a `MultiParentCasper` / for calling its methods

ArturGajowy: I mean a typeclass over the effect type - like anything after `F` in this line:

```scala
  def sendBlockRequest[F[_]: Monad: ConnectionsCell: TransportLayer: Log: Time: ErrorHandler: RPConfAsk](...)
```

KentShikama: But we need different handlers for the same instance of `MultiParentCasper`so I don't think it makes sense?

ArturGajowy: Didn't notice that, you're right. Still, please extract the `ignoreDoppelganger` constant so that the name at least tells us what's happening.

KentShikama:
@ArturGajowy PTAL

PR #1966, opened 2018-12-06
dropped blockBuffer contains check

dzajkowski:
this change results in a node being able to catch up to the current state of the chain.

note that the contains check seems to be excluded by: 
```
      case MissingBlocks =>
        Capture[F].capture { blockBuffer += block } *> fetchMissingDependencies(block)
```


KentShikama:
As mentioned in chat, I think the status quo is fine. The block will be added once all of its dependencies come in. Feel free to reopen as I could be mistaken.

PR #1968, opened 2018-12-07
ByteArray methods: nth, length

dckc:
## Overview

handle `ByteArray` methods in order to implement [rlp](https://github.com/ethereum/wiki/wiki/RLP)

I can make on if it's really important.

### Notes

I copied cost accounting from string methods.

See also worked example: https://github.com/dckc/RChain-API/blob/payment/src/ethSig.rho
with https://github.com/dckc/RChain-API/blob/payment/src/rlp.rho

### DIFF
--- a/rholang/examples/tut-bytearray-methods.rho
+++ b/rholang/examples/tut-bytearray-methods.rho
@@ -2,7 +2,7 @@ new x, y, stdout(`rho:io:stdout`) in {
 
   stdout!({
     "nth for byte array": "deadbeef".hexToBytes().nth(3),
-    "expected": "expected": 239 // 0xef
+    "expected": 239 // 0xef
   })
   |
   stdout!({
--- a/rholang/examples/tut-bytearray-methods.rho
+++ b/rholang/examples/tut-bytearray-methods.rho
@@ -1,0 +1,12 @@
+new x, y, stdout(`rho:io:stdout`) in {
+
+  stdout!({
+    "nth for byte array": "deadbeef".hexToBytes().nth(3),
+    "expected": "ef".hexToBytes()
+  })
+  |
+  stdout!({
+    "length for byte array": "deadbeef".hexToBytes().length(),
+    "expected": 4
+  })
+}
--- a/rholang/examples/tut-bytearray-methods.rho
+++ b/rholang/examples/tut-bytearray-methods.rho
@@ -2,7 +2,7 @@ new x, y, stdout(`rho:io:stdout`) in {
 
   stdout!({
     "nth for byte array": "deadbeef".hexToBytes().nth(3),
-    "expected": "ef".hexToBytes()
+    "expected": "expected": 239 // 0xef
   })
   |
   stdout!({
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -952,6 +952,13 @@ object Reduce {
                          s.fromEither(localNth(ps, nth))
                        case ETupleBody(ETuple(ps, _, _)) =>
                          s.fromEither(localNth(ps, nth))
+                       case GByteArray(bs) =>
+                         s.fromEither(if (0 <= nth && nth < bs.size) {
+                           val p: Par = Expr(GInt(bs.byteAt(nth).toLong))
+                           Right(p)
+                         } else {
+                           Left(ReduceError("Error: index out of bound: " + nth))
+                         })
                        case _ =>
                          s.raiseError(
                            ReduceError(
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1374,6 +1374,8 @@ object Reduce {
         baseExpr.exprInstance match {
           case GString(string) =>
             Applicative[M].pure[Expr](GInt(string.length.toLong))
+          case GByteArray(bytes) =>
+            Applicative[M].pure[Expr](GInt(bytes.size.toLong))
           case EListBody(EList(ps, _, _, _)) =>
             Applicative[M].pure[Expr](GInt(ps.length.toLong))
           case other =>
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -954,7 +954,8 @@ object Reduce {
                          s.fromEither(localNth(ps, nth))
                        case GByteArray(bs) =>
                          s.fromEither(if (0 <= nth && nth < bs.size) {
-                           val p: Par = Expr(GInt(bs.byteAt(nth).toLong))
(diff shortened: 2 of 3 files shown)


dckc:
The bigger goal is to be able to use signatures made by metamask where rlp plays the role of protobuf / toByteArray, especially in BasicWallet signatures.

I wrote https://github.com/rchain-community/RSign , but it's just a prototype. It would take a ton of work to make it usable by a wide audience.

No, there's no Jira ticket.

And in any case, surely nth should work on ByteArray as well as it works on String, no?

dckc:
ugh... bytes are signed. That's not what I want...

dckc:
turns out ethereum signatures aren't DER encoded the way our Secp256k1 library expects, so I implemented %% on ByteStrings

The details are in https://github.com/dckc/RChain-API/blob/payment/src/ethSig.rho .
The `tut-bytearray-methods.rho` test in this PR shows the `%%` functionality but gets some of the DER details wrong.

There's quite a bit of duplication in the interpolate code. I tried to abstract over String and ByteString but couldn't figure out where to put the `M[_]`s.

dckc:
How exactly should we encode each of the values? I started thinking in that direction but didn't see an obvious way to do it.

Interpolation is an extremely good fit for the particular use case of DER encoding of signatures.

ArturGajowy:
We should be doing sth. along the lines of:
`stdout!("deadbeef".hexToBytes() ++ 42.toHex().hexToBytes() ++ "c4f3b4b3".hexToBytes())`

Granted, there isn't a `.toHex()` method on integers so far, but I'd feel much more comfortable having this added to the language than array interpolation.

In principle, we don't even have to add the `.toHex()` method: we can write a contract that will convert an int to a hex string, using division, modulo division and array / map lookup. I won't argue for this being the way to go here - let's add  a `.toHex()` method on ints.

dckc:
> we can write a contract that will convert an int to a hex string, using division, modulo division and array / map lookup

yes; I did that first... though I didn't think of using a map; I used match:
https://github.com/dckc/RChain-API/blob/d09b0795717b7104acfe85ce0b423207bbf5aea2/src/ethSig.rho#L50-L74

`42.toHex().hexToBytes()` seems pretty awkward too. I think I'll skip it for now and go back to using a contract.

dckc:
I don't understand the drone failure; https://build.rchain-dev.tk/rchain/rchain/243 is 404 for me.

Clues?

KentShikama:
```
[info]   - should charge proportionally to the logarithm of length of target byte array *** FAILED *** (27 milliseconds)
[info]     TestFailedException was thrown during property evaluation. (RholangMethodsCostsSpec.scala:627)
[info]       Message: 0.0 was not equal to 2.0
[info]       Location: (RholangMethodsCostsSpec.scala:107)
[info]       Occurred at table row 0 (zero based, not counting headings), which had values (
[info]         left = GByteArray(<ByteString@63765383 size=1>),
[info]         right = GByteArray(<ByteString@3f073f7b size=1>)
[info]       )
```

https://drone.rchain-dev.tk/rchain/rchain/243/5

ArturGajowy:
The `ByteArray ++ - fix accounting` is far from fixing it. You had it right that the concatenation cost is `O(sum of lengths)`.

The commit log contains changes that cancel each other. This should not be the case.
Please rebase on `upstream/dev` and make sure no traces of interpolation are in the commit log, and that the test fixes are provided in the commits that introduce the change that requires each test fix.

> It looks like the drone CI check passed, once I made the 58eb657 fix. I don't understand what bors is waiting for.

Bors is waiting for you telling it to do a thing :) You can see a descritpion of any recent PR for a cheat-sheet. TL;DR once you've got a review :heavy_check_mark:, you comment `bors r+` and bors tests and merges.

dckc:
I didn't decide how to set the cost for ++; I just un-did the breakage I introduced.  Note the comments in Cost.scala:

```
  // GByteArray uses ByteString internally which in turn are implemented using
  // data structure called Rope for which append operation is O(logN)
```

++ was already supported on ByteArray; the only change I made relevant to ++ was to document it in tut-bytearray-methods.rho

I guess I don't mind learning how to edit the history, but would you please clarify CONTRIBUTING.md? It seems to say I should not:

> If you're picking up someone else's pull request, then whatever you do must preserve the commit history. We're not going to squash commits for now.

KentShikama:
In my opinion, try to preserve history but if you can't for whatever reason, don't worry about it.

ArturGajowy:
> // GByteArray uses ByteString internally which in turn are implemented using
>  // data structure called Rope for which append operation is O(logN)

I believe N here is the sum of lengths of concatenated strings. Or if you will, the size of the tree that might need rebalancing after it was modified. Anyway, I'm pretty sure the cost should depend on both of the concatenated strings. Hence my request for the jira-s.

> If you're picking up someone else's pull request, then whatever you do must preserve the commit history. We're not going to squash commits for now.

This only refers to other people's work one is continuing on.

dckc:
There. History is cleaned up.

Note there is nothing about ++ in this PR. I'll do the .rho example for that in a separate PR.

I don't have git signing set up where I am now. I hope I can fix that later.

dckc:
I can't find the drone build failure log. https://build.rchain-dev.tk/rchain/rchain/556 is 404

KentShikama:
```
- should sort so that unequal terms have unequal scores and the other way around *** FAILED *** (2 seconds, 937 milliseconds)
[info]   TestFailedException was thrown during property evaluation.
[info]     Message: Node(List(Leaf(ScoreAtom(IntAtom(7))))) equaled Node(List(Leaf(ScoreAtom(IntAtom(7)))))
[info]     Location: (SortTest.scala:60)
[info]     Occurred when passed generated values (
[info]       arg0 = List(Expr(ETupleBody(ETuple(Vector(),AlwaysEqual(BitSet()),true))), Expr(ETupleBody(ETuple(Vector(),AlwaysEqual(BitSet()),false)))) // 3 shrinks
```
https://drone.rchain-dev.tk/rchain/rchain/556/4

Looks transient @ArturGajowy @tymm ?

dckc:
How did you know to stick /4 on the end of the url, @KentShikama ?

KentShikama:
I didn't. Try without /4 and you'll see why.

dckc:
ah! **drone**.rchain-dev.tk vs build.rchain-dev.tk

tymm:
@KentShikama yes, that's transient. Will fix soon.

ArturGajowy:
@dckc this looks like a legit integration test failure (was reproduced twice):

```
Using tests random number generator seed: 1545407183
FAILED                                                                   [ 72%]
test/test_repl.py::test_repl 
```

See https://drone.rchain-dev.tk/rchain/rchain/610/6 and https://drone.rchain-dev.tk/rchain/rchain/614/6

dckc:
That is _strange_. I'm struggling to reproduce it. This works fine:

```
~/projects/rchain/rholang/examples$ for rho in *.rho; do echo $rho; rnode deploy --from=0x1 --phlo-limit=1000000 --phlo-price=1 --nonce=0 $rho; done
```

dckc:
My `rnode deploy` command didn't fail because I had the wrong version checked out.

Deploying does parse the code. The integration test failure was in deploy.

PR #1972, opened 2018-12-07
Implement genesis insertion in BlockDagStorage

KentShikama:
This is a hot fix to unblock everyone. This probably needs some refactoring/tests but the general idea of the issue should be clear.

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockDagStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockDagStorage.scala
@@ -8,7 +8,6 @@ import coop.rchain.blockstorage.BlockStore.BlockHash
 import coop.rchain.blockstorage.util.BlockMessageUtil.parentHashes
 import coop.rchain.blockstorage.util.TopologicalSortUtil
 import coop.rchain.casper.protocol.{BlockMessage, Bond}
-import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.shared.Log
 
 import scala.collection.immutable.HashSet


block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockDagStorage.scala
```diff
@@ -64,6 +64,31 @@ final class InMemBlockDagStorage[F[_]: Concurrent: Sync: Log: BlockStore](
...
+  override def insertGenesis(genesis: BlockMessage): F[Unit] =
+    genericInsert(genesis, genesisLatestMessagesStrategy)
+  private def genesisLatestMessagesStrategy(
+      genesis: BlockMessage,
+      latestMessages: Map[Validator, BlockHash]
+  ): Map[Validator, BlockHash] = {
+    val genesisBonds = (for {
+      bd <- genesis.body
+      ps <- bd.state
+    } yield ps.bonds).getOrElse(List.empty[Bond])
```
KentShikama: For some reason I couldn't import ProtoUtil.bonds

itegulov: You can't because block-storage doesn't depend on casper code

itegulov: You can use `BlockMessageUtil.bonds` though which will hopefully be removed once I get back to hiding optional fields from proto classes

KentShikama:
No we need to fill the latest messages with every bonded validator pointing to genesis at the start. Otherwise some validators will have empty in the latest messages map at the start which breaks validation. For every block after that you only update the latest messages for the sender of the block.

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockDagStorage.scala
KentShikama: Yeah I'll leave it to you to do the refactoring, tests, and implementation for the non in memory version. If you don't see any other issues let's merge this for now to unblock the failing integration tests.

itegulov:
I was basing my answer on the previous implementation which updated latests messages with every bonded validator for *every* block: https://github.com/rchain/rchain/blob/511785772878dcec7eedda7a81c4829616cb7738/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L497-L508

itegulov:
This way newly bonded validators can join later. Or am I missing something here?

KentShikama:
I'm on mobile so I can't take a deep look now but the lines you quoted only are relevant for validator rotation.

KentShikama:
> This way newly bonded validators can join later. Or am I missing something here?

Yeah we do need this ability. I'm a bit confused now what you're getting at but I'll take a deeper look later.

itegulov:
I am also on mobile so I am not sure if I have quoted the right block. We can merge it as is if you think this is very urgent.

KentShikama:
The integration tests now pass: https://drone.rchain-dev.tk/rchain/rchain/143/8 . @itegulov Lets discuss your earlier comment through chat or a different PR.

PR #1973, opened 2018-12-09
CORE-1332 Non-interactive terminal mode.

dzajkowski:
## Overview
non interactive mode.
```
(echo 5 | ./main repl)
5
```

CORE-1332

### Notes


node/src/main/scala/coop/rchain/node/TerminalMode.scala
```diff
@@ -0,0 +1,16 @@
+package coop.rchain.node
+
+trait TerminalMode[F[_]] {
+  def interactive: Option[Unit]
```
ArturGajowy: `Option[Unit]` is isomorphic to `Boolean`, so we should be using `Boolean` here (use the simplest type possible)

shared/src/main/scala/coop/rchain/shared/StringOps.scala
```diff
@@ -4,10 +4,16 @@ import scala.util.Try
 
 object StringOps {
   implicit class StringColors(str: String) {
-    def green: String     = s"\u001B[32m" + str + "\u001B[0m"
-    def red: String       = s"\u001B[31m" + str + "\u001B[0m"
-    def blue: String      = s"\u001B[34m" + str + "\u001B[0m"
-    def isNumber: Boolean = str.matches("[+-]?\\d+.?\\d+")
+    private def conditionalColor(coloring: String)(implicit tty: Boolean): String =
```
ArturGajowy: Boolean is not specific enough to be used as an implicit parameter. We either need a custom value type here, or some type tag, or maybe the `TerminalMode` capability itself.

node/src/main/scala/coop/rchain/node/repl.scala
```diff
@@ -43,10 +48,15 @@ class ReplRuntime() {
...
+    TerminalMode[F].interactive match {
+      case Some(_) => ConsoleIO[F].println(logo.red) >>= kp(repl)
+      case None => repl
+    }
   }
 
-  def evalProgram[F[_]: Monad: ReplClient: ConsoleIO](fileNames: List[String]): F[Unit] = {
+  def evalProgram[F[_]: Monad: ReplClient: ConsoleIO: TerminalMode](fileNames: List[String]): F[Unit] = {
+
+    implicit val isTty: Boolean = TerminalMode[F].interactive.exists(Function.const(true))
```
ArturGajowy: `interactive.nonEmpty`, but it shouldn't be Option in the first place

node/src/main/scala/coop/rchain/node/repl.scala
```diff
@@ -43,10 +48,15 @@ class ReplRuntime() {
       case false => ConsoleIO[F].close.as(false)
     }
 
-    ConsoleIO[F].println(logo) >>= kp(repl)
+    TerminalMode[F].interactive match {
+      case Some(_) => ConsoleIO[F].println(logo.red) >>= kp(repl)
```
ArturGajowy: `ConsoleIO[F].println(logo.red) >> repl`

node/src/main/scala/coop/rchain/node/TerminalMode.scala
rabbitonweb: I agree with `Boolean`, not see the point of `Option[Unit]` here

node/src/main/scala/coop/rchain/node/TerminalMode.scala
```diff
@@ -0,0 +1,16 @@
+package coop.rchain.node
+
+trait TerminalMode[F[_]] {
```
rabbitonweb: Why do we need extra typeclass? I think this makes perfect sense to put it as part of `Console`

node/src/main/scala/coop/rchain/node/TerminalMode.scala
dzajkowski: since it's isomorphic - why not use something that has a  `.map`?

ArturGajowy: Because it's misleading. Also: why not just use an `if` instead, or convert the boolean to an option in the call site?

node/src/main/scala/coop/rchain/node/TerminalMode.scala
dzajkowski: I don't mind adding it to `Console` but it seems to be an orthogonal topic.

ArturGajowy:
@dzajkowski when ready, please try merging using bors, by writing `bors r+`

ArturGajowy:
actually, please do a merge/rebase with dev first, because bors config and ci config are taken from the PR branch :)

adaszko:
❤️

PR #1989, opened 2018-12-11
CORE-1541 Change the default value of max message size to be 0.25 MB

rabbitonweb:
## Overview
This way all messages (including the ones that are streamed) can not
be bigger then 0.25 MB. User can still set the value to be something
different, but that value has to be in range [0.25MB, 10MB]

### JIRA ticket:
CORE-1541


rabbitonweb:
Just realized that grpc api is also using this flag, they should not. Will make changes accordingly.

PR #2002, opened 2018-12-13
Refactor compute parents post state

KentShikama:
Fix the fact we call indexOf on an Option[Int] instead of
the actual Int.

Group ordering with the calls that require implicit ordering.

Clean up some comments.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
@@ -173,53 +173,81 @@ object InterpreterUtil {
     val parentTuplespaces = parents.flatMap(p => ProtoUtil.tuplespace(p).map(p -> _))
 
     parentTuplespaces match {
-      //no parents to base off of, so use default
+      // For genesis, use empty trie's root hash
       case Seq() =>
         Right(runtimeManager.emptyStateHash).leftCast[Throwable].pure[F]
 
-      //For a single parent we look up its checkpoint
       case Seq((_, parentStateHash)) =>
         Right(parentStateHash).leftCast[Throwable].pure[F]
 
-      //In the case of multiple parents we need
-      //to apply all of the deploys that have been
-      //made in all of the branches of the DAG being
-      //merged. This is done by computing uncommon ancestors
-      //and applying the deploys in those blocks.
       case (initParent, initStateHash) +: _ =>
-        dag.deriveOrdering(0L).flatMap { implicit ordering: Ordering[BlockMetadata] => // TODO: Replace with an actual starting number
-          for {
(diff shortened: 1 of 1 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -173,54 +173,68 @@ object InterpreterUtil {
...
+      initStateHash: StateHash
+  )(implicit scheduler: Scheduler): F[Either[Throwable, StateHash]] =
+    for {
+      parentsMetadata <- parents.toList.traverse(b => dag.lookup(b.blockHash).map(_.get))
+      ordering        <- dag.deriveOrdering(0L) // TODO: Replace with an actual starting number
+      blockHashesToApply <- {
+        implicit val o: Ordering[BlockMetadata] = ordering
+        for {
+          uncommonAncestors          <- DagOperations.uncommonAncestors[F](parentsMetadata.toVector, dag)
+          ancestorsOfInitParentIndex = 0
```
lukasz-golebiewski: This doesn't change between iterations so can be moved to a constant at the beginning of the method

KentShikama: There is no iterations going on here.

ArturGajowy:
hmmm... that's unfortunate. I guess the way to stop bors would be `bors r-`. I'm also not sure if bors didn't check the review statuses again, or was just happy with there being at least one :heavy_check_mark:. Should we revert that commit, or just have a follow-up PR fixing @rabbitonweb's comment?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -173,54 +173,68 @@ object InterpreterUtil {
...
+            .toVector
+            .sorted // Ensure blocks to apply is topologically sorted to maintain any causal dependencies
+        } yield result
+      }
+      blocksToApply <- blockHashesToApply.traverse(b => ProtoUtil.unsafeGetBlock[F](b.blockHash))
+      deploys       = blocksToApply.flatMap(_.getBody.deploys.flatMap(ProcessedDeployUtil.toInternal))
+    } yield
+      runtimeManager
+        .replayComputeState(initStateHash, deploys, time)
+        .runSyncUnsafe(Duration.Inf) match {
```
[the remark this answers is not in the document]
KentShikama: @odeac RHOL-1050 will fix this. We should really let him work on merging it.

KentShikama:
In any case don't revert this PR, we need the changes in here. The runSyncUnsafe is nothing new.

KellyatPyrofex:
This PR supports work in [RHOL-1119](https://rchain.atlassian.net/browse/RHOL-1119)

PR #2017, opened 2018-12-17
CORE-[1499;1503] features.md for P2P and Node

rabbitonweb:
## Overview
Defining acceptance criteria for features in `features.md`

### JIRA ticket:
CORE-[1499;1503]

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -85,6 +85,22 @@ database = "rnode"
 * after a period of time new node connects (via protocol handshake) with `nodeB` and `nodeC`
 
 ### As a Node operator, I want to know how many peers I am connected to
+#### number of protocol peers
+##### test: test/test_p2p.py::test_number_of_protocol_peers
+##### steps:
+
+* create a p2p network with 3 nodes `nodaA`, `nodeB` and `nodeC`
+* access `nodeA` http endpoint under `/info` should print `connected_peers = 2`
+* access `nodeA` http endpoint under `/peers` to list `nodeB` and `nodeC` in JSON format
+
+#### number of protocol peers
+##### test: test/test_p2p.py::test_number_of_discovery_peers
+##### steps:
+
+* create a p2p network with 3 nodes `nodaA`, `nodeB` and `nodeC`
+* access `nodeA` http endpoint under `/info` should print `discovered_peers = 2`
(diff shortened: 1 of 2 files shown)


docs/features.md
```diff
@@ -9,12 +9,53 @@
...
+* node should succesfully start and connect to `standalone` via protocol handshake
+
+#### connecting to non-existing node
+##### test: test/test_p2p.py::test_connecting_to_non_existing_node
+##### steps:
+
+* start new node with `--bootstrap` pointing to some non-existing address
+* node should exit
+* exit code should be 1
+* node should log that bootstrap could not been found
```
ArturGajowy: I think we shouldn't be putting corner cases like that in the feature.md file, not to mention testing them using integration tests.

docs/features.md
```diff
@@ -9,12 +9,53 @@
...
+* join p2p network as described in "As a Node operator, I want to be able to bootstrap to the network by connecting to any known node" bootstaping from `nodeA`
+* after a period of time new node discovers `nodeB` and `nodeC`
+* after a period of time new node connects (via protocol handshake) with `nodeB` and `nodeC`
+
 ### As a Node operator, I want to know how many peers I am connected to
+#### number of protocol peers
+##### test: test/test_p2p.py::test_number_of_protocol_peers
+##### steps:
+
+* create a p2p network with 3 nodes `nodaA`, `nodeB` and `nodeC`
```
ArturGajowy: shouldn't this say `as described in "%the previous test%"`?

docs/features.md
rabbitonweb: why not @ArturGajowy , wouldn't it be good to know how node behaves with weird behaviour?

docs/features.md
rabbitonweb: Hmm good point, let me think how could I link that with a `source of truth` - but you are right

rabbitonweb:
bors +r

docs/features.md
ArturGajowy: I'd say, not for the price of an integration test, most of the time. And this particular scenario I'd rather accept as 'being obvious' than wait 30s or more every build for it to be tested - even if we don't want such talk in general. If at all, this should be unit-tested.

rabbitonweb: obvious as it may seem, currently node does not behave like describe
i don't mind for test running 3h as long as they cover functionality

PR #2020, opened 2018-12-18
Task-compatible effect abstraction over SystemProcesses 

JosDenmark:
## Overview
The previous version of `SystemProcesses` was just an object with a factory method for each system process. This PR refactors `SystemProcesses` into a type-class `SystemProcesses[F[_]]`. By imposing a context bound `Async[F]`, we get an instance of `SystemProcesses[Task]`, making the new implementation compatible with `Task`- based implementations that depend on it. See `Runtime`.

It's not as fine-grained as @lukasz-golebiewski's approach, but it's a more permanent solution and its consistent with coding standards. With `SystemProcesses` abstracted, we can start on `Runtime` and then do `Interpreter`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RHOL-1050

### Notes
For a fully algebraic description of `SystemProcesses`, we would want to express `Dispatch` and `RhoISpace` as context bounds on the `apply` function, instead of having them as explicit parameters.

### Checklist:


JosDenmark:
bors try

dzajkowski:
could you format the code (scalafmt) and push it? would make it easier to read.

JosDenmark:
@dzajkowski Done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -1,284 +1,293 @@
...
+  def blockTime(timestamp: BlockTime[F]): (Seq[ListParWithRandomAndPhlos], Int) => F[Unit]
+}
+
 object SystemProcesses {
 
-  private val prettyPrinter = PrettyPrinter()
+  def apply[F[_]](
+      dispatcher: Dispatch[F, ListParWithRandomAndPhlos, TaggedContinuation],
+      space: RhoISpace[F]
+  )(implicit F: Async[F]): SystemProcesses[F] =
```
lukasz-golebiewski: Isn't `Sync` enough to bring `delay` into scope?

lukasz-golebiewski:
This looks good and is exactly where we're headed! You managed to get rid of `Task` from `SystemProcesses` completely and so you didn't need to summon the capabilities of lifting it to an abstract context. This isn't always possible (in reasonable number of LOC) - I'm glad it was here

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
JosDenmark: You're right. I thought that I would need `Async` later to pass to `Dispatch`, but I don't. I just need it to be a monad for `Parallel[F,G]`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -213,17 +211,18 @@ object Runtime {
         shortLeashParams: ShortLeashParams[Task],
         blockTime: BlockTime[Task]
     ): RhoDispatchMap = {
+      val systemProcesses = SystemProcesses[Task](dispatcher, space)
```
goral09: I think there's still place for improvement. Since space and dispatcher are used only in SystemProcesses then you can pass its instance to this method and remove space and dispatcher from it completely.

goral09: Once you do that I think that new possibilities to improve the method arise. I think that it will be possible to get rid of `Task` entirely and use `F`

JosDenmark:
@lukasz-golebiewski This one was straight-forward because it only depended on `Dispatch`. You're alright if I continue the campaign? I think your approach is useful around the edges of the interpreter, while I'm working on this. Also, I noticed that your conversion type-class looks like a natural transformation, so `FunctionK` should provide the same capabilities.

PR #2022, opened 2018-12-18
RHOL-1094 RuntimeManager depends on F instead of Task.

dzajkowski:
## Overview
Continued work on abstracting core components to F

### JIRA ticket:
RHOL-1094

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -28,7 +28,6 @@ import monix.execution.Scheduler
 import scodec.Codec
 
 import scala.collection.immutable
-import coop.rchain.catscontrib.TaskContrib._
 import coop.rchain.casper.util.EventConverter
 import coop.rchain.casper._
 import coop.rchain.casper.util.rholang.RuntimeManager
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -2,7 +2,7 @@ package coop.rchain.casper.api
 
 import cats.Monad
 import cats.effect.concurrent.Semaphore
-import cats.effect.{Bracket, Concurrent, Sync}
+import cats.effect.Concurrent
 import cats.effect.Sync
 import cats._, cats.data._, cats.implicits._
 import cats.mtl._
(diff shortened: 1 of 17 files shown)


dzajkowski:
bors try

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -174,7 +166,7 @@ object BlockAPI {
       val stateHash =
         ProtoUtil.tuplespace(block).get
       val data =
-        runtimeManager.getData(stateHash, sortedListeningName)
+        runtimeManager.getData(stateHash, sortedListeningName).unsafeRunSync
```
lukasz-golebiewski: Instead of doing `unsafeRunSync` here, add a type constraint of `ToAbstractContext` to the function's signature, lift the task to `F` and move this line into the for comprehension below

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -7,79 +7,77 @@ import com.google.protobuf.ByteString
...
-import scala.concurrent.SyncVar
-import scala.concurrent.duration._
-import scala.util.{Failure, Success, Try}
+
+class RuntimeManager private (emptyStateHash: ByteString, runtimeContainer: MVar[Task, Runtime])
+    extends AbstractRuntimeManager[Task](emptyStateHash, runtimeContainer)
 
 //runtime is a SyncVar for thread-safety, as all checkpoints share the same "hot store"
-class RuntimeManager private (
+class AbstractRuntimeManager[F[_]: Concurrent: Sync: ToAbstractContext] protected (
```
ArturGajowy: Concurrent implies Sync (which is retarded, but nvm)

ArturGajowy: I mean, Concurrent implying Sync is retarded, not using both of them :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -148,70 +145,70 @@ class RuntimeManager private (
...
-                newCheckpoint.log,
-                DeployStatus.fromErrors(errors)
-              )
-              cont <- if (errors.isEmpty)
-                       doEval(rem, newCheckpoint.root, acc :+ deployResult)
-                     else doEval(rem, hash, acc :+ deployResult)
-            } yield cont
+        acc: Seq[InternalProcessedDeploy]
+    ): F[(StateHash, Seq[InternalProcessedDeploy])] =
+      terms match {
```
ArturGajowy: the `Defer` might've been securing us from stack overflows. We have two opstions here:

1. investigating if that could be the case (e.g. showing this isn't called recursively)
2. assuming it was not guarding against soe, as there'd be a failing test otherwise (:P)
3. putting a `Defer` constraint on the F, and using it here.

Pick whichever you prefer :)

casper/src/test/scala/coop/rchain/casper/genesis/DeployPaymentCostTest.scala
```diff
@@ -287,10 +288,12 @@ object DeployPaymentCostTest {
...
-    val transferStatus = node.runtimeManager.getData(
-      tuplespaceHash,
-      statusChannel
-    )
+    val transferStatus = node.runtimeManager
+      .getData(
+        tuplespaceHash,
+        statusChannel
+      )
+      .unsafeRunSync
```
ArturGajowy: We've had a pinky promise to have a negative balance of `unsafeRunSync` usages in the diff for further pull requests, but at this point I'm not sure if we should require that in tests.

I guess what could be done is to extract the repeated usage in each test to a priv. The ultimate prudence would be to slowly start converting tests into monadic computations. This way they'd show the *actual* way how the tested code is to be used...

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -7,79 +7,77 @@ import com.google.protobuf.ByteString
...
 import coop.rchain.catscontrib.TaskContrib._
 
-import scala.annotation.tailrec
 import scala.collection.immutable
-import scala.concurrent.SyncVar
-import scala.concurrent.duration._
-import scala.util.{Failure, Success, Try}
+
+class RuntimeManager private (emptyStateHash: ByteString, runtimeContainer: MVar[Task, Runtime])
+    extends AbstractRuntimeManager[Task](emptyStateHash, runtimeContainer)
```
ArturGajowy: I see you've chosen the path of fixing the F-type per class. I'd prefer to have each method parameterized with just the needed effects, but won't push on that.

dzajkowski: It's less code to go this way (IMO).

casper/src/test/scala/coop/rchain/casper/genesis/DeployPaymentCostTest.scala
dzajkowski: agreed. Ill figure out how to drop em.

dzajkowski:
bors try

PR #2023, opened 2018-12-18
Validate costs in replayEval

KentShikama:
Note this does not pass unit tests but eventually should when cost accounting becomes deterministic.


KentShikama:
bors try

KentShikama:
Removed all reviewers as it will be a while before this gets merged.

KentShikama:
I guess this would also need to handle the CostMismatch error

PR #2044, opened 2018-12-24
CORE-1564 - Document usage of ski package, refactor the code to be consistent

rabbitonweb:
## Overview

refactor means: All usages of const should be replaced with kp.

### JIRA ticket:
CORE-1564

### DIFF
--- a/shared/src/main/scala/coop/rchain/catscontrib/ski/package.scala
+++ b/shared/src/main/scala/coop/rchain/catscontrib/ski/package.scala
@@ -1,9 +1,25 @@
 package coop.rchain.catscontrib
 
+/**
+  * Name ski comes from SKI combinator calculus (https://en.wikipedia.org/wiki/SKI_combinator_calculus)
+  * It represents three functions, from which we are using two: constant function (kp) and identity function (id)
+  *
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/coop/rchain/catscontrib/ski/package.scala
```diff
@@ -1,9 +1,25 @@
...
+  */
 package object ski {
+
+  /**
+    * Returns a function that takes number of arguments and always returns the last argument as a result
+    */
   def kp[A, B](x: => B): A => B                               = _ => x
   def kp2[A, B, C](x: => C): (A, B) => C                      = (_, _) => x
   def kp3[A, B, C, D](x: => D): (A, B, C) => D                = (_, _, _) => x
   def kp6[A, B, C, D, E, F, G](x: G): (A, B, C, D, E, F) => G = (_, _, _, _, _, _) => x
```
ArturGajowy: Why can't we have just 
```scala
   def kp[T, U](x : => T)(y : U) : T
```
which is just a rename of Scala's `Function.const`, with the `x` passed by-name?

PR #2049, opened 2018-12-27
CORE-1566 new graph

rabbitonweb:
## Overview
As requested by @KentShikama trying to provide a new graph visualization layout that has following properties:

1. layouts all blocks of each validator in a single group / cluster
2. layouts all blocks that happened at given "tick" in the same line

It seemed that turning on clusters in graphviz and setting `rank` to same for both vertical and horizontal alignment would do the trick. Well, it did not. Graphiz has powerful ranking mechanism that I quite don't yet fully comprehend, but alining both vertically (per validator) and horizontally (per tick) was quite a challange. Some modifications were needed in the Graphz module.

In this PR I did following:
1. Refactor, init should just initialized graph, not accumulator
2. Add invisible style to edges
3. Change graph names to always be prefixed
4. Add style and contraint to edges definition
5. Extend GraphzGenerator to visualize DAG as cluster of validators
6. Modify BlockAPI to extract visualizer

image
 
--

### JIRA ticket:
https://rchain.atlassian.net/browse/CORE-1566

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
@@ -17,14 +17,12 @@ case class Acc[G[_]](timeseries: List[Long] = List.empty, graph: G[Graphz[G]])
 
 object GraphzGenerator {
 
-  private def init[G[_]: Monad: GraphSerializer](name: String): Acc[G] =
-    Acc[G](
-      graph = Graphz[G](
-        name,
-        DiGraph,
-        rankdir = Some(BT),
-        node = Map("width" -> "0", "height" -> "0", "margin" -> "0.03", "fontsize" -> "8")
-      )
+  private def initGraph[G[_]: Monad: GraphSerializer](name: String): G[Graphz[G]] =
+    Graphz[G](
+      name,
+      DiGraph,
+      rankdir = Some(BT),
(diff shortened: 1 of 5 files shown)


rabbitonweb:
@dzajkowski @KentShikama this PR is ready for review

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -135,7 +136,7 @@ object Graphz {
       case (Graph, _)   => s"graph"
       case (DiGraph, _) => s"digraph"
     }
-    if (name == "") s"$prefix {" else s"$prefix $name {"
+    if (name == "") s"$prefix {" else s"""$prefix "$name" {"""
```
KentShikama: Is there no isEmpty?

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -36,7 +111,7 @@ object GraphzGenerator {
       if (blockHash == lastFinalizedBlockHash) Some(Filled) else None
 
     for {
-      acc <- topoSort.foldM(init[G]("dag")) {
+      acc <- topoSort.foldM(Acc[G](graph = initGraph[G]("dag"))) {
```
KentShikama: I feel like we can do better than Acc and Acc2

KentShikama:
Also it seems like you have a build failure

PR #2014, opened 2018-12-17
Run integration tests in parallel on CI

adaszko:
https://rchain.atlassian.net/browse/CORE-1371

PR #2053, opened 2018-12-28
Rchain 2211 Make integration tests run in parallel

zsluedem:
## Overview
Make integration tests run in parallel

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2211

### Notes

### DIFF
--- a/integration-tests/test/conftest.py
+++ b/integration-tests/test/conftest.py
@@ -100,7 +100,7 @@ def temporary_wallets_file(random_generator: Random, validator_keys: List[KeyPai
         os.unlink(file)
 
 
-@pytest.yield_fixture(scope='session')
+@pytest.yield_fixture(scope='module')
 def docker_client() -> Generator[DockerClient, None, None]:
     docker_client = docker_py.from_env()
     try:
--- a/scripts/ci/run-integration-tests.sh
+++ b/scripts/ci/run-integration-tests.sh
@@ -21,7 +21,7 @@ main () {
 
     ./mypy.sh
     ./pylint.sh
-    ./run_tests.sh --log-cli-level=ERROR --mount-dir="$TEMP_RESOURCES_DIR"
+    ./run_tests.sh --log-cli-level=ERROR --mount-dir="$TEMP_RESOURCES_DIR" -n 2 --dist=loadscope
 }
 
 
--- a/integration-tests/Pipfile
+++ b/integration-tests/Pipfile
@@ -10,6 +10,7 @@ pytest-json = "*"
 pytest-pylint = "*"
 typing-extensions = "*"
 dataclasses = "*"
+pytest-xdist = "*"
 mypy = "*"
 
 [dev-packages]
--- a/integration-tests/Pipfile.lock
+++ b/integration-tests/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "5cbe7b930bdefb5ef443dfd373b9342ffb4aeff24b578e251f4eddbd24f643af"
+            "sha256": "d9b416276e7614df564ba31ead30cf626b54e23ddc427a28cc3479909002b8a5"
         },
         "pipfile-spec": 6,
         "requires": {},
@@ -14,6 +14,13 @@
         ]
     },
     "default": {
+        "apipkg": {
+            "hashes": [
+                "sha256:37228cda29411948b422fae072f57e31d3396d2ee1c9783775980ee9c9990af6",
+                "sha256:58587dd4dc3daefad0487f6d9ae32b4542b185e1c36db6993290e7c41ca2b47c"
+            ],
+            "version": "==1.5"
+        },
         "astroid": {
             "hashes": [
                 "sha256:35b032003d6a863f5dcd7ec11abd5cd5893428beaa31ab164982403bcb311f22",
@@ -72,6 +79,13 @@
             ],
             "version": "==0.4.0"
         },
+        "execnet": {
+            "hashes": [
+                "sha256:a7a84d5fa07a089186a329528f127c9d73b9de57f1a1131b82bb5320ee651f6a",
+                "sha256:fc155a6b553c66c838d1a22dba1dc9f5f505c43285a878c6f74a79c024750b83"
+            ],
+            "version": "==1.5.0"
+        },
         "idna": {
             "hashes": [
                 "sha256:c357b3f628cf53ae2c4c05627ecc484553142ca23264e593d327bcde5e9c3407",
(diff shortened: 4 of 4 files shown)


zsluedem:
bors try

zsluedem:
@adaszko Please delegate the permission for me to run the test and put the `do not merge` label on this pr. I want to do some test.  :)

adaszko:
bors delegate+

ArturGajowy:
Why are we even working on this, when we've seen the
```
Potential problems: Conflicts in Docker resources like networks and bootstrap node. Testing will be needed to see where problems lay.

(see jira ticket)
```
to be actual problems? We should measure the resource requirements for i-tests before even trying to parallelise them. I've mentioned that in https://github.com/rchain/rchain/pull/2014#issuecomment-447864128.

Not to mention the fact, that even if we parallelise the integration tests, the unit tests take the same time ATM and the total build time won't change because of that.

I'd rather start with parallelizng / speeding up the unit tests, or doing the measurments.

I think our efforts could be better directed.

adaszko:
> Potential problems: Conflicts in Docker resources like networks and bootstrap node. Testing will be needed to see where problems lay.

This work is about resolving those problems.

> Not to mention the fact, that even if we parallelise the integration tests, the unit tests take the same time ATM and the total build time won't change because of that.

There are lots of integration tests coming as a result of implementing features.md scenarios.

> I'd rather start with parallelizng / speeding up the unit tests, or doing the measurments.

Those two aren't mutually exclusive.

> We should measure the resource requirements for i-tests before even trying to parallelise them

Parallelizing should show where the bottlenecks are and we'll be able to cap the resources there.

We can continue this discussion in a private conversation.  Comments here should be about the code itself.

ArturGajowy:
> Comments here should be about the code itself.

Disagree :) If there are errors on the conceptual level, they're relevant to the PR as well.

>> Potential problems: Conflicts in Docker resources like networks and bootstrap node. Testing will be needed to see where problems lay.

> This work is about resolving those problems.

Apologies, I've read `resources` and stopped reading. This will blow up on more trivial resources, like CPU and RAM.

> There are lots of integration tests coming as a result of implementing features.md scenarios.

Maybe we should then think of a more efficient test harness than 'python over shell over java' - before we start writing those. Parallelizing the tests can only take us this far.

>> I'd rather start with parallelizng / speeding up the unit tests, or doing the measurments.
> Those two aren't mutually exclusive.

My point is about prioritisation and minimising the potential for wasted work.

> Parallelizing should show where the bottlenecks are and we'll be able to cap the resources there.

I don't see how making the tests use more CPU and RAM and IOPS - all at once, by running many at a time - will tell us how much resources a single test needs, or - more importantly - what are the max resource requirements of any of our tests.

I can see though how this helps us measure the 'builds to run in parallel' resource, which we both know to be approaching 1 once we parallelize the integration tests.

zsluedem:
bor try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

integration-tests/test/conftest.py
```diff
@@ -125,7 +125,7 @@ def testing_context(command_line_options: CommandLineOptions, random_generator:
 testing_context.__test__ = False # type: ignore
 
 
-@pytest.yield_fixture(scope='module')
```
adaszko: I think this change will slow down the execution of tests in the sequential case.

zsluedem: You are right! It also would increase the memory which the test would use.

zsluedem:
Ok. The newest push above cause the `error` and it looks not like anyone I had in my local computer. I wonder that someone who had the access to the testing machine would check and see if there are any undeleted containers staying there.

I have met several errors in my local machine.
1. docker client delete network error.(something like the error of the above pr had)
2.timeout error
3.metrics return nothing error
I tried to reproduce the errors above and found that it is mostly because the machine is out of the memory. 

When I tried to parallel running the test, the memory usage would go up very high and it cause the machine to run very slowly which would cause the 2.timeout error and the 3.metrics not functioning normally(I guess). 

But I can not know why the first one docker deleting network error would happen. It is kind of strange that the test start 2 bonded nodes and the third one wouldn't start yet but the test stops with error. I guess it might use all of the memory and the system or dockerd tried to kill the test or the containers by itself.

If I set lower parallel nums of the test(like 2, default would be the cores of the CPU) in my local machine, the test can be successfully passed  with several tries.

 @woky told me that the integration test machine has 16 CPUs and 60GB of RAM. I think if we should limit the cores we use to parallel the integration test maybe even limit the memory use of each container.

zsluedem:
bors try

zsluedem:
bors try

integration-tests/test/conftest.py
adaszko: This is very hard to accept.  I think majority of the times (esp. on the dev's machines), the tests will be run sequentially.  I was hoping for something that doesn't cause regression in performance in the sequential case.  I'm aware this hunk has been removed but the point still stands for the other similar hunks.

zsluedem: I was just doing some tests.
It can be done that way. `xdist` has a param `dist`. If it is set to be `loadscope `, the tests would be groupped by module.

zsluedem:
bors try

scripts/ci/run-integration-tests.sh
```diff
@@ -21,7 +21,7 @@ main () {
 
     ./mypy.sh
     ./pylint.sh
-    ./run_tests.sh --log-cli-level=ERROR --mount-dir="$TEMP_RESOURCES_DIR"
+    ./run_tests.sh --log-cli-level=ERROR --mount-dir="$TEMP_RESOURCES_DIR" -n 4 --dist=loadscope
```
adaszko: From https://pypi.org/project/pytest-xdist/ :

> --dist=loadscope: tests will be grouped by module for test functions and by class for test methods, then each group will be sent to an available worker, guaranteeing that all tests in a group run in the same process. This can be useful if you have expensive module-level or class-level fixtures. Currently the groupings can’t be customized, with grouping by class takes priority over grouping by module. This feature was added in version 1.19.

Wasn't aware of this setting  👍

ArturGajowy:
I feel as if you had just subtweeted me @adaszko :stuck_out_tongue: 

We've seen multiple builds running same integration tests pass at once recently, so I'd give it a try at least.

Let's do this:

bors try

and then restart that build 3 or 4 times while it's on top of the queue. This should allow us to see how stable this is.

zsluedem:
When the memery is out , it would take much longer time to do the integration test which doesn't gain speed up actually(even worse). The situation of that 5 tests met the situation I had in my local machine. We might set a lower `n` to give a try.

zsluedem:
bors try

ArturGajowy:
bors try

ArturGajowy:
I've just started builds 1065-1069. 3 of them started right away, there's one other unrelated build halfway through.

zsluedem:
It looks like the machine doesn't have enough memery for paralleling the tests.

zsluedem:
@adaszko @ArturGajowy Should we give up paralleling the integration test? It looks like that the machine doesn't have enought memery for that.

PR #2054, opened 2018-12-28
CORE-1376 Introduce a circut breaker mechanism when handling a stream

rabbitonweb:
## Overview
When handling a stream (receiving incoming data), handler might decide
to break a circuit (base on how much data was read so far). This way
if node receives a block that is simply to big, it will ignore
it (will not handle it).

Even though the mechanism is in place, currently circuit breaker
accepts any incomming stream (it never breaks). What should be
conditions of it breaking the circuit are a matter of discussion
planned in RCHAIN-2792

### JIRA ticket:
CORE-1376


comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -51,30 +56,51 @@ object StreamHandler {
...
+        stmd.fos.flush()
+        val readSoFar = stmd.readSoFar + array.length
+        if (circuitBreaker(readSoFar))
+          Right(stmd.copy(circuitBroken = true))
+        else
+          Left(stmd.copy(readSoFar = readSoFar))
+    }
+
+    EitherT(collectStream.attempt.map {
+      case Right(stmd) if stmd.circuitBroken => Left(new RuntimeException("Circuit was broken"))
```
sebtomba: Please remove the file when the circuit is broken. Also remove the file in case of `Left`.

comm/src/test/scala/coop/rchain/comm/transport/StreamHandlerSpec.scala
```diff
@@ -76,13 +79,32 @@ class StreamHandlerSpec extends FunSpec with Matchers with BeforeAndAfterEach {
       // then
       msg.path.getParent shouldBe nonExistingWithPersmission
     }
+
+    it("should stop receiving a stream if will not fit in memory") {
+      // given
+      val alwaysBreak: CircuitBreaker = kp(true)
+      val stream                      = createStream()
```
lukasz-golebiewski: Perhaps it would make sense to test for a handful of generated streams

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
rabbitonweb: 👍

comm/src/test/scala/coop/rchain/comm/transport/StreamHandlerSpec.scala
rabbitonweb: 👍

rabbitonweb: I've changed that not with property test, but with a bit more complex unit test

PR #2055, opened 2018-12-30
Let syntax 2641 (WIP: IOU tests)

dckc:
## Overview
Implement syntactic sugar for let declarations in rholang

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2641 

### Notes

There's one noteable KLUDGE:

```
        // KLUDGE: assume these are distinct from names used elsewhere
        val idents = 0 until p.listbinding_.length map {
          case ix => s"let_%{p.line_num}_${p.col_num}_${ix}"
        }
```

but for my own work, writing `new ch in { ch!(123) | for(x <- ch) { ... use x ...} }` over and over is so mind-numbing that I'm going to use this patch to save my sanity.

- [ ] has a reviewer assigned,
- [x] has [all commits signed]

### DIFF
--- a/rholang/src/main/bnfc/rholang_mercury.cf
+++ b/rholang/src/main/bnfc/rholang_mercury.cf
@@ -64,6 +64,7 @@ PInput.          Proc2  ::= "for" "(" Receipt ")" "{" Proc "}" ;
 PChoice.         Proc2  ::= "select" "{" [Branch] "}" ;
 PMatch.          Proc2  ::= "match" Proc4 "{" [Case] "}" ;
 PBundle.         Proc2  ::= Bundle "{" Proc "}" ;
+PLet.            Proc2  ::= "let" [Binding] "in" "{" Proc3 "}";
 PIf.             Proc1  ::= "if" "(" Proc ")" Proc2 ;
 -- Use precedence to force braces around an interior if.
 PIfElse.      Proc1 ::= "if" "(" Proc ")" Proc2 "else" Proc1 ;
@@ -119,6 +120,9 @@ separator nonempty Branch "" ;
 -- Match Cases
 CaseImpl. Case ::= Proc13 "=>" Proc3 ;
 separator nonempty Case "" ;
+
+BindingProcToPat. Binding ::= [Name] NameRemainder "<-" Proc3 ;
+separator nonempty Binding "," ;
 
 -- Name Declarations.
 -- Eventually will have IOPairs.
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -1069,6 +1069,42 @@ object ProcNormalizeMatcher {
         normalizeIfElse(p.proc_1, p.proc_2, p.proc_3, input.copy(par = VectorPar()))
           .map(n => n.copy(par = n.par ++ input.par))
 
+      case p: PLet =>
+        // KLUDGE: assume these are distinct from names used elsewhere
+        val idents = 0 until p.listbinding_.length map {
+          case ix => s"let_%{p.line_num}_${p.col_num}_${ix}"
+        }
+        val newNames = {
+          val names = new ListNameDecl()
+          idents foreach { case i => names.add(new NameDeclSimpl(i)) }
+          names
+        }
+        val sends = {
+          var ps: Proc = new PNil()
+          p.listbinding_.zip(idents) foreach {
+            case (b: BindingProcToPat, ident) =>
+              val rhs = new ListProc()
+              rhs.add(b.proc_)
+              val send = new PSend(new NameVar(ident), new SendSingle(), rhs)
+              ps = if (ps == new PNil()) {
+                send
+              } else {
+                new PPar(ps, send)
+              }
+            case (b, _) => throw new RuntimeException("unknown binding:" + b.toString())
+          }
+          ps
+        }
+        val forBindings = new ListLinearBind()
+        p.listbinding_.zip(idents) foreach {
+          case (b: BindingProcToPat, ident) =>
+            forBindings.add(new LinearBindImpl(b.listname_, b.nameremainder_, new NameVar(ident)))
+          case (b, _) => throw new RuntimeException("unknown binding:" + b.toString())
+        }
+        val forProc      = new PInput(new ReceiptLinear(new LinearSimple(forBindings)), p.proc_)
+        val letExpansion = new PNew(newNames, new PPar(sends, forProc))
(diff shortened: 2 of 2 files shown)


ArturGajowy:
> Is it possible to directly build the ADT instead of calling normalize?

I'm sure it is, but why would you want that? My default would be to translate the syntax sugar on lexical level and then let the normalizer do its job.

goral09:
> > Is it possible to directly build the ADT instead of calling normalize?
> 
> I'm sure it is, but why would you want that? My default would be to translate the syntax sugar on lexical level and then let the normalizer do its job.

I don't think it is possible to translate `let` to some other (already existing) construct. It is much more than just syntactic sugar. `let x = @0!(1)` cannot be expressed using `new`, send or receive. I think it's only possible to do in the `reduce` phase where `x` becomes `@0!(1)` in the environment. You just have to error when variable name used in binding conflicts with ones defined in `new`... Unless you will do lexical scoping.

ArturGajowy:
@goral09 I'm not sure what's the semantics you're after for `let x = @0!(1)`, but according to the proposal this would translate to:

```
let x = @0!(1) in { P }

||| desugars to
vvv

new let1 in {
  let1!(@0!(1)) |
  for (x <- let1) {
    P
  }
}
```

which should probably be optimized to just inlining the x, but that's a different question.

The semantics for the sugar are the semantics of the desugaring. ;)

goral09:
This is awfully inefficient - both from the CPU/storage perspective as well as from the economics perspective. 

My understanding was that in `P` `x` binds to send but it's not required to store/retrieve it from RSpace but put it in the env directly.

ArturGajowy:
That's a good point - we probably shouldn't be implementing such sugar until we get escape analysis for names and inline such non-escaping names.

dckc:
Is escape analysis scheduled? https://rchain.atlassian.net/browse/RCHAIN-2541 is "local GC" which seems like the same thing. It doesn't seem to be scheduled.

I'm less excited about this sort of `let` now that I have it working... it only works for expressions. The mind-numbing stuff I really want to get rid of looks more like:

```
          new lenCh in {
            oneByte!(bias + payload.length(), *lenCh) |
            for(@len <- lenCh) {
```
_taken from: https://github.com/dckc/RChain-API/blob/ethWallet/src/rlp.rho_

I want to write:

```
  let len <- oneByte!(bias + payload.length()) in {
```

@leithaus  told me translating function calls to process calculi is well-trodden ground, but when I looked it up, the expansion of **[[M N : B]]** is parameterized by **u**, a name, which suggests `lenCh` doesn't go away.


[Sequentiality and the π-Calculus](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.32.9361)
by Martin Berger , Kohei Honda , Nobuko Yoshida 2000

JoshOrndorff:
@dckc I understand what you want and why you want in in the case of
```rholang
  let len <- oneByte!(bias + payload.length()) in {
```
But how would it generalize? What if `oneByte` sent its result on one of three different channels depending on some runtime condition? Like this for example:
```rholang
someCrazyListOperation!(myList, *oddLengthAnswerCh, *evenLengthAnswerCh) |
for (@ans <- oddLengthAnswerCh) { ...do something... } |
for (@ans <- evenLengthAnswerCh) { ...do something else... }
```
Or in another context, we might might get messages sent on either, neither, or both
```rholang
jury!(caseDetails, *aliceGuiltyCh, *bobGuiltyCh)
```

IMO the expression-only let you've coded here is a good useful start, although Kyle's performance concerns sound serious. I don't know enough about the rholang interpreter to know exactly what the concerns are.

JoshOrndorff:
> I don't expect it to [generalize]. But the case of a single return channel is interesting enough...
I agree it's still useful enough to warrant writing for this specific case, but we still need to specify semantics for what happens in those other cases. Or perhaps you meant that it should not parse if the contract being called fits the pattern of having a single return channel as the last argument? I don't think that works since the contract may not even appear in the tuplespace until after the send is made.

I think what we're discussing now is a different animal and should have a separate ticket. What do you think about this syntax?
```rholang
oneByte!(bias + payload.length(), let lenCh)
```

Then it actually does generalize in an interesting way
``` rholang
jury!(caseDetails, let aliceGuilty, let bobGuilty) {
  // Use the values from the let
  // Possible semantic meanings
  // * It's a join
  // * It runs twice if they're both guilty?
}
```

This may even obviate [my request](https://joshyorndorff.com/rholang-wishlist) for `new @x in {`

cc @Isaac-DeFrain

> does a produce to send 4 to r-space and and a consume to get it back out, which is a little silly.
Makes sense. Thanks.

> So... <- vs. = and , vs. ; – how to decide?
> (copied from [atlassian comment](https://rchain.atlassian.net/browse/RCHAIN-2641?focusedCommentId=13190&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13190))
I'm fine with either.

dckc:
> ... the case of a single return channel is interesting enough ...

It's also what Turner's pict does. In section 2.7 returning results. Definition 2.6 says

`let x1, ..., x1 = f(a1, ..., am) in P`

expands to

`new r in { f!(a1, ..., am, r) | for (r <- x1, ..., xn) { P } }` where `r` is not free in `P`, `f`, `a1`, ..., `am`.

and "We omit the `in` and `let` keywords in nested `let` expressions" so you get `let xs = f(...) ys = g(...) in P` rather than `let xs = f(...) in let y = g(...) in P`.

When I was thinking about this in tooth-brushing mode, I supposed I could detect the case where the RHS of a let was a send: `let x = f!(...)`. But now that I'm looking up the details, I see Turner introduces new `f(...)` syntax without the `!`. Hmm...

dckc:
I just discovered the following match idiom that subsumes this let syntax:

```
    match (
      "%FROM",
      "%TO",
      100
    ) {
      (from, to, amount) => { ... }
```
 -- [3.transfer_funds.rho](https://github.com/rchain/rchain/blob/dev/rholang/examples/vault_demo/3.transfer_funds.rho)

JoshOrndorff:
I still think it's worth having let syntax. The explicit match leads to deeper nesting and may mislead readers of the code as to what is happening.

I agree match is superior to send and receive in the meantime, and if let is just sugar, it should wrap this match idea. I guess that would also solve the concern about accessing the tuplespace for let bindings.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

dckc:
de-sugaring `let` to this `match` idiom is a reasonable idea. But I'm not inclined to re-open this PR to take it on.

I'm less interested in `let` syntax unless / until it comes with the pict-style `let x = f(a)` function call syntax.

PR #2057, opened 2018-12-31
RCHAIN-2790 justifications added

rabbitonweb:
## Overview

1. Fill lastFinalizedblock in cluster DAG visualization
2. Put all ancestors under the clusters
3. Add justifications into cluster dag
4. Add arrow type to graphz

image

### JIRA ticket:
RCHAIN-2790


ArturGajowy:
The arrows are superficial (even the ones we have), because they always point down (to the earlier blocks)

goral09:
I know that, you know that but not everybody knows that.

rabbitonweb:
I've removed arrows because they were obfuscating the output too much

rabbitonweb:
I'm considering adding colors though. Each block gets its colors and same colors get justifications that this block has.

ArturGajowy:
That'd be a lot of colors :rainbow: :unicorn:

Instead, I'd give each column a letter A, B, C
Each block in each column, a number (for the picture above): 0 (Genesis), B1, A2, C3, ...
Then, instead of justification edges, I'd have each block annotated with its justifications:

E.g. for the block B4 (second block in column B) I'd have the justifications "0 B1 C3" (if I'm reading the diagram correctly...)

For reverse references, we could have the referring blocks listed above the block. So incoming justification edges would be listed above the block, and outgoing ones would be listed below (because the justifying blocks are under the justified ones)

ArturGajowy:
Then, I'd maybe add colors on top of that :)

KentShikama:
Is it not possible to make them look like this?

KentShikama:
I would rather have no justifications than curvy justification lines.

KentShikama:
Having the state hash inside the block is actually pretty helpful. In the future we could add more information such as deploy count, block height, etc.

rabbitonweb:
I can not make those lines straight in graphviz, I've tried. I will make the justifications an option flag when you run the `vdag` - makes sense?

rabbitonweb:
@ArturGajowy I'm making you a reviewer since you are already engage.
@KentShikama @ArturGajowy - I'll make justifications lines an flag option, so you will run it like `rnode vdag --show-justification-lines`, by default it will be disabled

In next PR I will extend the graph so that each node can display some useful info about the block, I need to yet to know what that info is (will ask queistions in ticket)

casper/src/main/scala/coop/rchain/casper/util/comm/DeployService.scala
```diff
@@ -58,8 +58,8 @@ class GrpcDeployService(host: String, port: Int, maxMessageSize: Int)
       else Left(new RuntimeException(response.status))
     }
 
-  def visualizeBlocks(q: BlocksQuery): Task[Either[Throwable, String]] =
-    stub.visualizeBlocks(q).map { response =>
+  def visualizeDag(q: VisualizeDagQuery): Task[Either[Throwable, String]] =
```
ArturGajowy: How easily can I DOS the node by calling `visualizeDag(depth = -1)` repeatedly? :smiling_imp:

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -13,17 +13,143 @@ import cats.mtl._
...
+  }
+
+  def dagAsCluster[
+      F[_]: Monad: Sync: MultiParentCasperRef: Log: SafetyOracle: BlockStore,
+      G[_]: Monad: GraphSerializer
+  ](
+      topoSort: Vector[Vector[BlockHash]],
+      lastFinalizedBlockHash: String,
+      config: GraphConfig
+  ): F[G[Graphz[G]]] =
```
ArturGajowy: As it is now, Graphz[G] contains both the data structure and one of the possible interpreters for it (which requires the GraphSerializer and the  G-effect).

I think it'd be better to separate the Graphz[G] into:
 - pure non-parametrized data structure
 - an G-parameterized inerpreter for it that is using the graph serializer

Then, this method could be `dagAsGraphz[F: ...](...) : F[Graphz]`, and its output could go to `serializeGraphs[F: GraphSerializer] : F[String]`. Or if we really want it, we could have it `serializeGraphs[G: GraphSerializer] : G[Unit]` and then run the G (under F?) to extract the result.

ArturGajowy: I've seen your test/diagnostic util you've deleted from the other commit, and I thought: how do I even test that? My idea would be:

1. Have a data structure for the **abstract** graph (not the graphiviz-styling-oriented one Graphz is)
2. Generate an `Arbitrary` `Graph`, call it g
3. Interpret g into Graph-viz annotated structure using test code. Call it `expected`.
4. Interpret g into a BlockDag using test code. Call it `input`
5. (the 'when' part of the test:) Interpret `input` into Graph-viz using prod code, call it `actual`
6. Compare `expected` and `actual`.

Maybe we could even s/Graph-viz structure/String for a more precise and bigger-scoped test. But it'd be harder to maintain.

Just an idea, given that I've not seen actual tests for this :)

rabbitonweb: 1. regarding structure vs serializer

This is exactly design that I was aiming for, for two reasons:

a) this represents how graphviz is constucted. Using the DOT language you can add edges, node, configurations in randomly - arbitrarly. There is very little restrictions on how youy provide your graph. You can first define node, then edges. Or first edges, then nodes and then subrgraphs (or any other mixture for all that matter). 
a) secondly the idea originally was to allow bulding arbitrary HUGE graphs that will not blow up the memory.

Bottom line is that, somebody might fold over its own internal structure (potentially even on the fly, reading data from the stream/iterator) and building graph along the way. Your serializer might accumulate data (as the current `StringSerialzier` is doing) but it could also be a serialzier with very little memory footprint (one that for example drops everything to a file).

Initial graph generator (the one that is removed in this PR) was pushing that concept. You were foldLefting the structure and the graph was serializaed alon the way - it was still using the `StringSerialzier` but it could be replaced with `FileSerializer` whenever needed. 

The new generator (dagCluster) can not into account that "low memory footprint" feature into account because of how limited the graphviz is in generating "cluster-like" visualizations. Nevertheless it would be a bad idea to remove the feature (once its already here), because it is not being used. 

TLDR: The desing is as you see it to have ability to create a lazy graphs (all effects are closed over `G[_]` that has a small memory footprint (useage of serializer). In theory you could process a huge, tremendous graph, reading it from stream, and serialize it to a file on a very low memory consumption.

2. Regarding tests

a) testing graphz module

Now, you are mistaken that there are NO tests for the graphz module. Please see  `GraphzSpec` . Those tests are checking if the graphz module is propely rendering graphs. And it is covering quite a portion of its features.

b) testing visualize-dag functionality

What we are missing though is testing not the graphz module, but testing the usegae of graphz to visualize DAG.

I have an open ticket to add tests to it (RCHAIN-2801). Features requests are starting to grow over this `visualize-dag` functionality. It started as something that might or might not be used. Since it is being used, and there are features requests I will add tests to it. It will be based on my (VisalizeDagTest) that I've been using locally and accidently added to this PR.

c) property based testing

I am huge fun of property based testing. But that kind of testing only works if you have clearly defined properties. I would have to think if there are any that immediately are obvious. If you have an idea let me know.

PR #2059, opened 2019-01-02
Delete tmp comm folder on startup

sebtomba:
## Overview
Cleanup tmp folder for comm on startup

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2794

### DIFF
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -348,12 +348,14 @@ class NodeRuntime private[node] (
     multiParentCasperRef <- MultiParentCasperRef.of[Effect]
     lab                  <- LastApprovedBlock.of[Task].toEffect
     labEff               = LastApprovedBlock.eitherTLastApprovedBlock[CommError, Task](Monad[Task], lab)
+    commTmpFolder        = conf.server.dataDir.resolve("tmp").resolve("comm")
+    _                    <- Task.delay(commTmpFolder.toFile.delete()).attempt.toEffect
     transport = effects.tcpTransportLayer(
       port,
       conf.tls.certificate,
       conf.tls.key,
       conf.server.maxMessageSize,
-      conf.server.dataDir.resolve("tmp").resolve("comm")
+      commTmpFolder
     )(grpcScheduler, log, metrics, tcpConnections)
     kademliaRPC = effects.kademliaRPC(kademliaPort, defaultTimeout)(
       grpcScheduler,
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -39,6 +39,7 @@ import coop.rchain.node.diagnostics._
 import coop.rchain.p2p.effects._
 import coop.rchain.rholang.interpreter.Runtime
 import coop.rchain.shared._
+import coop.rchain.shared.PathOps._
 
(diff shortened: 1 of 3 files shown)


node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -348,12 +348,14 @@ class NodeRuntime private[node] (
     multiParentCasperRef <- MultiParentCasperRef.of[Effect]
     lab                  <- LastApprovedBlock.of[Task].toEffect
     labEff               = LastApprovedBlock.eitherTLastApprovedBlock[CommError, Task](Monad[Task], lab)
+    commTmpFolder        = conf.server.dataDir.resolve("tmp").resolve("comm")
+    _                    <- Task.delay(commTmpFolder.toFile.delete()).attempt.toEffect
```
rabbitonweb: logs? if you can't remove it, folder will grow. add `warn`?

sebtomba: Ok, then we need some more sophisticated logic that checks if the folder exists. If the folder exists then try to delete. Log in case of an exception. Log in case of result `false`. We have already this logic in Tcp TL. Maybe I can move it to `shared`.

ArturGajowy: Mabye add a log message saying 'deleting comm tmp folder $folderPath'?

shared/src/main/scala/coop/rchain/shared/PathOps.scala
```diff
@@ -38,4 +42,49 @@ object PathOps {
...
+      import java.io.File
+      import java.util.Comparator
+
+      import scala.collection.JavaConverters._
+
+      import cats.implicits._
+
+      def deleteFile(file: File)(implicit logSource: LogSource): F[Unit] =
+        for {
+          deleted <- Sync[F].delay(file.delete)
```
ArturGajowy: what if file.delete throws?

shared/src/main/scala/coop/rchain/shared/PathOps.scala
```diff
@@ -38,4 +42,49 @@ object PathOps {
...
+          _ <- if (deleted) Log[F].info(s"Deleted file ${file.getAbsolutePath}")
+              else Log[F].warn(s"Can't delete file ${file.getAbsolutePath}")
+        } yield ()
+
+      def getFiles: F[List[Path]] =
```
ArturGajowy: I don't like the behavior change being made in one commit with the move - it's hard to see what's actually being changed. Please keep that in mind for future PRs.

shared/src/main/scala/coop/rchain/shared/PathOps.scala
```diff
@@ -38,4 +42,49 @@ object PathOps {
...
+            .walk(path)
+            .sorted(Comparator.reverseOrder())
+            .iterator()
+            .asScala
+            .toList
+        )
+
+      def delete0(): F[Unit] =
+        for {
+          files <- getFiles
```
ArturGajowy: what if `path` is a single file?

sebtomba: this will be fine. I tested it in a side project

shared/src/main/scala/coop/rchain/shared/PathOps.scala
sebtomba: this is an internal function used only in `delete0()`. And `delete0()` is called once `delete0().attempt`

PR #2060, opened 2019-01-02
RCHAIN-993 unit test showing state not catching up with a fix.

dzajkowski:
## Overview
Introduces a unit test that shows that a change in how the blocks are added to the 2-way graph fixes a catch up problem.

essentially the fix boils down to:
if a block is a parent of 2 block then this fact needs to make its way to `blockBufferDependencyDag`. it is not sufficient to hold only one of the dependencies (out of potentially many children).
The clash happens when a parent block is also a justification for a younger block (c2 and f2).

### JIRA ticket:
https://rchain.atlassian.net/projects/RCHAIN/issues/RCHAIN-993

### Notes
none really. it's a step in identifying a bug that's been in the project for over 4 months.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -489,23 +489,26 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
         .toSet
       allDependencies = (missingParents union missingJustifications).toList
       missingDependencies <- allDependencies.filterA(
-                              blockHash =>
-                                dag
-                                  .lookup(blockHash)
-                                  .map(_.isEmpty && !blockBuffer.exists(_.blockHash == blockHash))
-                            )
+        blockHash =>
+          dag
+            .lookup(blockHash)
+            .map(_.isEmpty))
+      missingUnseenDependencies = missingDependencies.filter(blockHash => !blockBuffer.exists(_.blockHash == blockHash))
       _ <- missingDependencies.traverse(hash => handleMissingDependency(hash, b))
+      _ <- missingUnseenDependencies.traverse(hash => requestMissingDependency(hash))
     } yield ()
 
-  private def handleMissingDependency(hash: BlockHash, parentBlock: BlockMessage): F[Unit] =
+  private def handleMissingDependency(hash: BlockHash, childBlock: BlockMessage): F[Unit] =
     for {
       _ <- blockBufferDependencyDagState.modify(
-            blockBufferDependencyDag =>
-              DoublyLinkedDagOperations
-                .add[BlockHash](blockBufferDependencyDag, hash, parentBlock.blockHash)
-          )
-      _ <- CommUtil.sendBlockRequest[F](BlockRequest(Base16.encode(hash.toByteArray), hash))
+        blockBufferDependencyDag =>
+          DoublyLinkedDagOperations
+            .add[BlockHash](blockBufferDependencyDag, hash, childBlock.blockHash)
+      )
     } yield ()
+
+  private def requestMissingDependency(hash: BlockHash) =
+    CommUtil.sendBlockRequest[F](BlockRequest(Base16.encode(hash.toByteArray), hash))
 
   private def handleInvalidBlockEffect(status: BlockStatus, block: BlockMessage): F[Unit] =
     for {
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -46,7 +46,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
 
   private val blockBuffer: mutable.HashSet[BlockMessage] =
     new mutable.HashSet[BlockMessage]()
-  private val blockBufferDependencyDagState =
+  private val dependencyDagState =
     new AtomicMonadState[F, DoublyLinkedDag[BlockHash]](AtomicAny(BlockDependencyDag.empty))
 
   private[casper] val deployHist: mutable.HashSet[Deploy] = new mutable.HashSet[Deploy]()
@@ -107,7 +107,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
       _ <- attempt match {
             case MissingBlocks => ().pure[F]
             case _ =>
-              Capture[F].capture { blockBuffer -= b } *> blockBufferDependencyDagState.modify(
+              Capture[F].capture { blockBuffer -= b } *> dependencyDagState.modify(
                 blockBufferDependencyDag =>
                   DoublyLinkedDagOperations.remove(blockBufferDependencyDag, b.blockHash)
               )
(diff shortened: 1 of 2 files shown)


casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -1057,6 +1057,101 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+        _ <- nodes(1).receive()
+        _ <- nodes(2).transportLayerEff.clear(nodes(2).local) //nodes(2) misses this block
+      } yield ()
+    }
+
+    def propagate(nodes: Seq[HashSetCasperTestNode[Effect]]) = {
+      for {
+        _ <- nodes(0).receive()
+        _ <- nodes(1).receive()
+        _ <- nodes(2).receive() //nodes(2) gets the 5th block and starts requesting the dag
```
KentShikama: This tests uses a mocked version of the transport layer. When you call `receive` on a node, it takes things off its mocked network queue and then processes them. If you want to simulate the process of node A asking for a block from node B and having node B send it back, you have to do something like

```
// node A sends block request
nodeB.receive() // node B receives block request and sends the block
nodeA.receive() // node A receives block
```

dzajkowski: that's why it's done 20 times - to let the network do its thing.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -1057,6 +1057,101 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+
+      _ <- stepSingle(nodes)
+      _ <- stepSingle(nodes)
+
+      _ <- stepSplit(nodes)
+      _ <- stepSplit(nodes)
+
+      br <- deploy(nodes(0), deployDatasFs(0).apply())
+
+      _ <- propagate(nodes)
```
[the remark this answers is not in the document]
ArturGajowy: :100:

dzajkowski:
bors try

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
KentShikama: I see; so in theory you only need to do like 5~6 times but you just do it 20 times to be sure?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -1057,6 +1057,78 @@ class HashSetCasperTest extends FlatSpec with Matchers {
...
+      } yield ()
+    }
+
+    def stepSingle(nodes: Seq[HashSetCasperTestNode[Effect]]) = {
+      for {
+        _ <- deploy(nodes(0), deployDatasFs(0).apply())
+
+        _ <- nodes(0).receive()
+        _ <- nodes(1).receive()
+        _ <- nodes(2).transportLayerEff.clear(nodes(2).local) //nodes(2) misses this block
```
KentShikama: If nodes(2) misses the block and no node proposes a new block, how does nodes(2) know there is a block they need to fetch?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
KentShikama: How does nodes(2) get the 5th block?

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
KentShikama: Nvm ignore the above

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
KentShikama: > How does nodes(2) get the 5th block?

Nvm ignore

KentShikama:
Nice catch. Thanks for cleaning up after me; I always had a single parent block model in my head to simplify things but it backfired here.

KentShikama:
But I'll have to think a bit more. I don't quite understand why this change makes the tests pass.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -1057,6 +1057,78 @@ class HashSetCasperTest extends FlatSpec with Matchers {
     } yield result
   }
 
+  it should "ask peers for blocks it is missing and add them" in effectTest {
```
KentShikama: Is it possible to draw out a diagram of what this test looks like?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -489,24 +489,27 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
...
     for {
       _ <- blockBufferDependencyDagState.modify(
-            blockBufferDependencyDag =>
-              DoublyLinkedDagOperations
-                .add[BlockHash](blockBufferDependencyDag, hash, parentBlock.blockHash)
-          )
-      _ <- CommUtil.sendBlockRequest[F](BlockRequest(Base16.encode(hash.toByteArray), hash))
+        blockBufferDependencyDag =>
+          DoublyLinkedDagOperations
+            .add[BlockHash](blockBufferDependencyDag, hash, childBlock.blockHash)
```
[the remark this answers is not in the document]
dzajkowski: because it can be a missing parent of a block that's already been added to blockBuffer.

IOW two blocks can have one parent.

when the "missing blocks" status is being resolved for a block - this block will be added to the `blockBufferDependencyDag` (with its parents) and to the block buffer.

the `blockBufferDependencyDag` is used in reattempt to get the `dependencyFree` blocks.

now if a block is already in `blockBuffer` but it's a parent of another block (which was requested but not yet processed) then it will never be added to `blockBufferDependencyDag` (because it's in the blockBuffer).
This results in the processing getting "stuck" - the blockBuffer holds a block that should be processed (so it's not allowing new blocks via the network) and the internal "catch up" mechanism isn't aware that the second block has a parent that is already available and should be added (hence blocking the catch up process).

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
dzajkowski: in theory yes BUT:
I've observed a funny "self healing" mechanism - if the network layer holds enough requested blocks it will make the catch up mechanism look like it works (it processes stale requested blocks). This way I'm making sure all the blocks are processed.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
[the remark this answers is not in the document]
KentShikama: Oh whoops; I see my mistake... essentially blockBufferDependencyDag represents blocks that aren't in the block dag. And so blocks in the block buffer don't count.

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
KentShikama: Ignore this unless you've already started. I quickly drew it out.

KentShikama: Well thanks anyways

PR #2061, opened 2019-01-03
Replay deploys from a block with the corresponding timestamp

KentShikama:
Previously we were injecting the current time to replay
deploys for parents blocks.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-1122

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
@@ -47,8 +47,7 @@ object InterpreterUtil {
       possiblePreStateHash <- computeParentsPostState[F](
                                parents,
                                dag,
-                               runtimeManager,
-                               timestamp
+                               runtimeManager
                              )
       _ <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(b)}.")
       result <- processPossiblePreStateHash[F](
@@ -155,7 +154,7 @@ object InterpreterUtil {
       implicit scheduler: Scheduler
   ): F[Either[Throwable, (StateHash, StateHash, Seq[InternalProcessedDeploy])]] =
     for {
-      possiblePreStateHash <- computeParentsPostState[F](parents, dag, runtimeManager, time)
+      possiblePreStateHash <- computeParentsPostState[F](parents, dag, runtimeManager)
     } yield
       possiblePreStateHash match {
         case Right(preStateHash) =>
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -199,29 +196,41 @@ object InterpreterUtil {
...
       dag: BlockDagRepresentation[F],
       runtimeManager: RuntimeManager,
-      time: Option[Long],
       initStateHash: StateHash
   ): F[Either[Throwable, StateHash]] =
     for {
       blockHashesToApply <- findMultiParentsBlockHashesForReplay(parents, dag)
       blocksToApply      <- blockHashesToApply.traverse(b => ProtoUtil.unsafeGetBlock[F](b.blockHash))
-      deploys            = blocksToApply.flatMap(_.getBody.deploys.flatMap(ProcessedDeployUtil.toInternal))
-      replayResult <- ToAbstractContext[F].fromTask(
```
dzajkowski: what will happen when the `blocksToApply` collection holds more than 1 block?

KentShikama: It will run `replayComputeState` as many times as there are `blocksToApply`.

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
dzajkowski: I believe that using the replay mechanic this way will result in further hits to performance. Essentially this makes the algorithm n^2 and forces the tuplespace to perform very expensive operations multiple times.
Is this the only way to solve passing time to replay?

KentShikama: What is the expensive operation? `createCheckpoint`? `reset` in theory should be free.

KentShikama: In the future we could have an optimization that compresses it back together if there is no deploy that uses the blockTime

lukasz-golebiewski: `reset` isn't free, because it has to re-populate previously cleared hot store with data

KentShikama: Why is the hot store previously cleared? In 99% of cases, you're reseting to the previous checkpoint value and so if you just left the hot store alone, things would smoothly just continue.

KentShikama:
bors try

KentShikama:
Yeah I agree; I think we are finally at the point where we can do that though

PR #2064, opened 2019-01-03
Edited test plan

zscole:
Made changes to test plan. 

1. I updated the system specifications and adjusted the validator/static node count to reflect these. 

2. There is no practical way to enforce a deterministic propagation time, so I removed this row as a variable. 

3. I set 1Gb as the default bandwidth per test series (apart from the test series in which bandwidth is a variable). The idea is to only observe the effects of particular variables on performance, so if the original 5Mb value is applied by default, this could have adverse effects on overall performance since 5Mb is considerably low. 

4. I increased the range of latency between test cases within Series 5. Cross continental latency is often <150ms.

5. I increased the range of packet loss between test cases within Series 6. since average packet loss can sometimes be much greater than what was provided. Admittedly, the 1% packet loss in Test Case C is very high and may result in complete system failure, but it's good to observe these edge cases for future planning. 

Again, this is a boilerplate test plan and each test series is subject to change since the results on one can have an effect on others.

### DIFF
--- a/docs/whiteblock-test-plan.md
+++ b/docs/whiteblock-test-plan.md
@@ -30,6 +30,13 @@ The proposed testing initiatives will be conducted within using the Whiteblock S
 | CPU Max MHz | 4000.0000                                      |
 | RAM         | 256GB DDR4                                     |
 
+## Contracts Used
+The following contracts will be deployed within the various test cases outlined in this document:
+
+- [dupe.rho](https://github.com/rchain/rchain/blob/dev/rholang/examples/dupe.rho)
+- [shortslow.rho](https://github.com/rchain/rchain/blob/dev/rholang/examples/shortslow.rho)
+- [shortfast.rho](https://github.com/rchain/rchain/blob/dev/rholang/examples/shortfast.rho)
+
 ## Performance Metrics
 
 Time measurements are expressed in terms of the time passed on the node
@@ -64,7 +71,7 @@ which define the variable to be tested.
 | Validators       | 10          | 20          | 30          |
 | Static Nodes     | 6           | 6           | 6           |
 | Contract         | dupe.rho    | dupe.rho    | dupe.rho    |
-| Bandwidth        | 1GB         | 1Gb         | 1Gb         |
+| Bandwidth        | 1Gb         | 1Gb         | 1Gb         |
 | Network Latency  | 0ms         | 0ms         | 0ms         |
 | Packet Loss      | 0%          | 0%          | 0%          |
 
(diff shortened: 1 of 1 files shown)


adaszko:
It looks good to me.  Let's if devs will have something to add.

docs/whiteblock-test-plan.md
```diff
@@ -66,49 +57,48 @@ series focuses on observing and documenting the effects of certain conditions
...
-### Series 1: Number of Validators
+### Series 1: Validators
 
 | Variable         | Test Case A | Test Case B | Test Case C |
 |------------------|------------:|------------:|------------:|
-| Validators       | 300         | 600         | 1000        |
-| Static Nodes     | 600         | 600         | 600         |
+| Validators       | 10          | 20          | 30          |
+| Static Nodes     | 6           | 6           | 6           |
 | Contract         | dupe.rho    | dupe.rho    | dupe.rho    |
```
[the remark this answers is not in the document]
zscole: updated

docs/whiteblock-test-plan.md
```diff
@@ -66,49 +57,48 @@ series focuses on observing and documenting the effects of certain conditions
...
-
 ### Series 3: Contract
 
 | Variable        | Test Case A | Test Case B   | Test Case C   |
 |-----------------|------------:|--------------:|--------------:|
-| Validators      | 1000        | 1000          | 1000          |
-| Static Nodes    | 1000        | 1000          | 1000          |
+| Validators      | 18          | 18            | 18            |
+| Static Nodes    | 18          | 18            | 18            |
 | Contract        | dupe.rho    | shortfast.rho | shortslow.rho |
```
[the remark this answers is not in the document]
zscole: updated

KellyatPyrofex:
bors +

docs/whiteblock-test-plan.md
```diff
@@ -66,49 +63,48 @@ series focuses on observing and documenting the effects of certain conditions
...
+| Packet Loss      | 0%          | 0%          | 0%          |
 
 ### Series 2: Number of Static Nodes
 
 | Variable        | Test Case A | Test Case B | Test Case C |
 |-----------------|------------:|------------:|------------:|
-| Validators      | 600         | 600         | 600         |
-| Static Nodes    | 300         | 600         | 1000        |
+| Validators      | 6           | 5           | 5           |
+| Static Nodes    | 10          | 20          | 30          |
```
lukasz-golebiewski: What's a static node? Can't find anything on that in the wiki

zscole: A passive node that isn't acting as a validator. Could be a node that's sending or receiving transactions, deploying contracts, etc.

lukasz-golebiewski: This is a bit confusing. My first thought was that this is a new name for a read only node. See: https://rchain.atlassian.net/wiki/spaces/CORE/pages/495747126/Read+Only+Nodes
I think the words static/passive should be reserved for those kinds of nodes.

Perhaps the nodes you described should be called "standard" to avoid confusion?

zscole: The description in that link is a bit vague. Can Read Only Nodes still deploy smart contracts and conduct transactions?

zscole: Ok, then I think the term static node is more appropriate and general since they can be filling any other role that isn't specific to validators.

PR #2065, opened 2019-01-04
Express matcher effects in terms of MTL typeclasses

ArturGajowy:
## Overview

This is not to say that the whole rholang uses the effects via MTL. This is to say that what's left should be trivial to fix :)

### JIRA ticket:

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
@@ -4,6 +4,7 @@ import cats.MonadError
 import cats.arrow.FunctionK
 import cats.data.{OptionT, StateT}
 import cats.implicits._
+import cats.mtl.implicits._
 import cats.mtl.MonadState
 import coop.rchain.models.Par
 import coop.rchain.rholang.interpreter.accounting.Cost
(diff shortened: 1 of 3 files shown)


ArturGajowy:
bors try

ArturGajowy:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -24,37 +28,38 @@ package object matcher {
   type NonDetFreeMapWithCost[A] = StateT[StreamWithCost, FreeMap, A]
   type StreamWithCost[A]        = StreamT[ErroredOrCostA, A]
 
+  type _freeMap[F[_]] = MonadState[F, FreeMap]
```
lukasz-golebiewski: Why the `_` prefixes? Are you trying to express something which should be expressed with an access modifier?

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -24,37 +28,38 @@ package object matcher {
...
+  type _cost[F[_]]    = MonadState[F, Cost]
+  type _error[F[_]]   = MonadError[F, OutOfPhlogistonsError.type]
+  type _short[F[_]]   = MonadError[F, Unit] //arises from and corresponds to the OptionT/StreamT in the stack
+
+  def _freeMap[F[_]](implicit ev: _freeMap[F]) = ev
```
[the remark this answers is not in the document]
ArturGajowy: For the reader of those declarations, probably. It's though very convenient on the call site, and in fact brings the convenience you gain with a custom typeclass to the type alias without:
 - a) having to copy the typeclass and have back-and-forth conversions between the custom typeclass and the cats equivalent
 - b) extending the cats typeclass and getting into even more trouble

TL;DR: Just like you do `Sync[F].delay { ... }` you'd (probably) like to do `_freeMap[F].modify(...)` without worrying that `freeMap` is a an alias (vs a type with a companion object) and thus has no implicit summoner method.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
ArturGajowy: This convention is inspired by the [eff library](http://atnos-org.github.io/eff/org.atnos.site.Tutorial.html#write-an-interpreter-for-your-program) and I use it because it elegantly avoids the name conflict between the `Cost` typeclass and the `_cost` effect. I could go with suffixes, but it'd:
 1. be more verbose
 2. expose implementation detail that's likely to change (_cost is likely to be writer in a while, and _error to be ApplicativeError/ApplicativeThrow vs MonadError)

In my mind, this convention means: this is an effect-type alias. Will be used similarly to capabilities, but for more generic and probably low-level/implementation stuff.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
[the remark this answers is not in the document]
ArturGajowy: I can add a
```
// Implicit summoner methods, just like `Monad.apply` on `Monad`'s companion object.
```
comment if it helps :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
lukasz-golebiewski: Makes sense. Shouldn't it read `_freeMapState` then?

ArturGajowy: I'd like to avoid the suffix for the same reason `_error` is not `_monadError` :)

rabbitonweb: it's a bit unusual for the type to be lowercased, i was confused when I saw the usage

rabbitonweb: you'r saying this is convention?

ArturGajowy: See my reply to Łukasz. It's up to us to shape our conventions. I've seen it in the wild though, in the `eff` lib's docs. I guess I'll add my explanation to a comment near those type aliases :)

PR #2068, opened 2019-01-04
Stream a packet to all peers using only one single file

sebtomba:
## Overview
Stream a packet to all peers using only one single file

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2793


rabbitonweb:
The more I read PR's that operate on files, the more I think we need `FileSystem` algebra (in tagless final encoding). PR looks good so far.

comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -11,38 +11,41 @@ import monix.reactive.Observable
...
-          log.error(s"Could not serialize packet ${blob.packet}. Error: $er") *> true.pure[Task]
+          log.error(s"Could not serialize packet ${blob.packet}. Error: $er") *> None.pure[Task]
       }
 
-    def retry(failed: List[PeerNode]): Task[Unit] =
+    def push(file: Path): Task[Boolean] =
+      Task.delay(subject.pushNext(StreamToPeers(peers, file, blob.sender)))
+
+    def propose(file: Path): Task[Unit] =
+      push(file) >>= (_.fold(Task.unit, retry(file)))
```
rabbitonweb: I know this is unrelated to this PR, but isn't that a potential infinite loop? we need to stop at some point?

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
```diff
@@ -208,34 +208,29 @@ class TcpTransportLayer(
   def broadcast(peers: Seq[PeerNode], msg: Protocol): Task[Seq[CommErr[Unit]]] =
     innerBroadcast(peers, msg)
 
-  def handleToStream(toStream: ToStream): Task[Unit] = {
+  private def streamToPeer(peer: PeerNode, path: Path, sender: PeerNode): Task[Unit] = {
```
rabbitonweb: name is misleading, `handle` convention was always good, you knew you are now "handling" a message, you are "on the server side"

comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
sebtomba: Yes, I had the same concern. On the other hand we are pushing an item to a queue. Eventually the queue will accept the item. Until then Casper must be blocked and not produce new items. We have limited reties (ATM 3x) on the other end of the message queue. So `propose`/`retry` should work as expected and not run forever.

comm/src/main/scala/coop/rchain/comm/transport/TcpTransportLayer.scala
sebtomba: I don't agree. We are not handling a message (`ToStream') here anymore. It's a function with parameters. We are on the client side and want to stream a file to some peer node (server side). Maybe I can change the parameter name `path` to `file`?

PR #2076, opened 2019-01-07
New Block Storage Implementation

itegulov:
## Overview
First part of the new block storage implementation. Only provides basic implementation of `BlockStore` API through a file storage with LMDB index and checkpoints loading (but not looking up in them as this will likely be a part of the new `BlockStore` API). Also introduces a couple of other small changes such as removing `asMap` method from `BlockStore` API and enclosing BlockStoreTest under Task.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-652

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -3,6 +3,7 @@ package coop.rchain.blockstorage
 import java.io.RandomAccessFile
 import java.nio.ByteBuffer
 import java.nio.file.{Files, Path}
+import java.util.stream.Collectors
 
 import cats.Monad
 import cats.effect.{Concurrent, ExitCase, Sync}
@@ -10,20 +11,26 @@ import cats.implicits._
 import cats.effect.concurrent.{Ref, Semaphore}
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.BlockStore.BlockHash
+import coop.rchain.blockstorage.FileLMDBIndexBlockStore.Checkpoint
+import coop.rchain.blockstorage.errors.{CheckpointsAreNotConsecutive, CheckpointsDoNotStartFromZero}
 import coop.rchain.casper.protocol.BlockMessage
 import coop.rchain.shared.Resources.withResource
 import coop.rchain.blockstorage.util.byteOps._
+import coop.rchain.shared.Log
 import org.lmdbjava.DbiFlags.MDB_CREATE
 import org.lmdbjava.Txn.NotReadyException
 import org.lmdbjava._
 
 import scala.collection.JavaConverters._
+import scala.ref.WeakReference
+import scala.util.matching.Regex
 
 class FileLMDBIndexBlockStore[F[_]: Monad: Sync] private (
     lock: Semaphore[F],
     env: Env[ByteBuffer],
     index: Dbi[ByteBuffer],
-    blockMessageRandomAccessFileRef: Ref[F, RandomAccessFile]
+    blockMessageRandomAccessFileRef: Ref[F, RandomAccessFile],
+    checkpointsRef: Ref[F, List[Checkpoint]]
 ) extends BlockStore[F] {
   implicit class RichBlockHash(byteVector: BlockHash) {
 
@@ -124,6 +131,9 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync] private (
       } yield ()
     )
 
+  override def checkpoint(): F[Unit] =
+    ().pure[F]
+
   override def clear(): F[Unit] =
     lock.withPermit(
       for {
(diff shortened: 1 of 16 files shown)


itegulov:
bors try

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
```diff
@@ -33,78 +43,96 @@ trait BlockStoreTest
...
-        store.asMap().size shouldEqual items.size
-        store.clear()
+        for {
+          _ <- items.traverse_(store.put(_))
+          _ <- items.traverse[Task, Assertion] {
+                case (k, v) =>
+                  store.get(k).map(_ shouldBe Some(v))
+              }
+          result <- store.find(_ => true).map(_.size shouldEqual items.size)
+          _      <- store.clear()
```
lukasz-golebiewski: I don't think you need to clear the store if you're creating a new store for every generated `blockStoreElements`

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,272 @@
...
+    } {
+      case (txn, ExitCase.Completed) => Sync[F].delay(txn.close())
+      case (txn, _) =>
+        Sync[F].delay {
+          try {
+            txn.abort()
+          } catch {
+            case ex: NotReadyException =>
+              ex.printStackTrace()
+              TxnOps.manuallyAbortTxn(txn)
```
lukasz-golebiewski: This isn't needed anymore (vide LMDBOps.scala)

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,272 @@
...
+    lock: Semaphore[F],
+    env: Env[ByteBuffer],
+    index: Dbi[ByteBuffer],
+    blockMessageRandomAccessFileRef: Ref[F, RandomAccessFile],
+    checkpointsRef: Ref[F, List[Checkpoint]]
+) extends BlockStore[F] {
+  implicit class RichBlockHash(byteVector: BlockHash) {
+
+    def toDirectByteBuffer: ByteBuffer = {
+      val buffer: ByteBuffer = ByteBuffer.allocateDirect(byteVector.size)
```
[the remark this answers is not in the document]
itegulov: They are different as `coop.rchain.shared.RichByteVector` wraps `ByteVector` while this one wraps `ByteString`. I have moved it into a separate file `coop.rchain.shared.RichByteString` though.

itegulov:
bors try

itegulov:
@lukasz-golebiewski PTAL

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,255 @@
...
+    for {
+      checkpointDirectories <- Sync[F].delay {
+                                checkpointsDirPath.toFile.mkdir()
+                                Files.list(checkpointsDirPath).filter(p => Files.isDirectory(p))
+                              }
+      checkpointDirectoriesList = checkpointDirectories
+        .collect(Collectors.toList[Path])
+        .asScala
+        .toList
+      checkpoints <- checkpointDirectoriesList.flatTraverse { dirPath =>
```
rabbitonweb: you should use just `traverse` (instead of `flatTraverse`) here are all the elements that your function is returning are single elements (which you artificially wrap into `List`

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,255 @@
...
+                              start.toLong,
+                              end.toLong,
+                              dirPath,
+                              dirPath.resolve("storage"),
+                              dirPath.resolve("index"),
+                              None
+                            )
+                          ).pure[F]
+                        case other =>
+                          Log[F].warn(s"Ignoring directory '$other': not a valid checkpoint name") *>
```
rabbitonweb: shouldn't that just stop rnode working all toghether? log error and close node?

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,255 @@
...
+                    }
+      sortedCheckpoints = checkpoints.sortBy(_.start)
+      result <- if (sortedCheckpoints.headOption.forall(_.start == 0)) {
+                 if (sortedCheckpoints.isEmpty ||
+                     sortedCheckpoints.zip(sortedCheckpoints.tail).forall {
+                       case (current, next) => current.end == next.start
+                     }) {
+                   sortedCheckpoints.pure[F]
+                 } else {
+                   Sync[F].raiseError(
```
rabbitonweb: We should not use `raiseError`, this is equivalent to throwing an exception. You can not potentially recover from it (unless you use `attempt` on the caller side). But this essentially bring us to imperative programming at this point - because requires using `attempt` on the caller side each time I'm using a function that returns `F[_]:Sync`. 

If you have a logical error that should not stop the system you should do one of the folowing:
1. return an error as a value (e.g `Either[Error, Value]`)
2. use `ApplicativeError` that is based on some ADT of errors (not exceptions that will be thrown by Task)

In other words: `Sync[F].raiseError` should **only** be used when the application should just simply crash. No recover, nothing. Just plain, old dead process. 
If that is what you were planning (stopping the system), potential risk that needs to be taken into consideration is resource cleanup - using for example `bracket` when calling `FileLMDBIndexBlockStore.create` . This is one more time a situation were we are not taking leverage of using FP paradigm - this is getting us closer to imperative hell. Why? Well, nothing is stopping the call to just use `FileLMDBIndexBlockStore.create` without taking care of resources.

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,255 @@
...
+import scala.util.matching.Regex
+
+class FileLMDBIndexBlockStore[F[_]: Monad: Sync] private (
+    lock: Semaphore[F],
+    env: Env[ByteBuffer],
+    index: Dbi[ByteBuffer],
+    blockMessageRandomAccessFileRef: Ref[F, RandomAccessFile],
+    checkpointsRef: Ref[F, List[Checkpoint]]
+) extends BlockStore[F] {
+  private[this] def withTxn[R](txnThunk: => Txn[ByteBuffer])(f: Txn[ByteBuffer] => R): F[R] =
```
rabbitonweb: this is copy-paste from `LMDBBlockStore`, copied, not reused

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -0,0 +1,255 @@
...
+      case (txn, _) => Sync[F].delay(txn.close())
+    }
+
+  private[this] def withWriteTxn(f: Txn[ByteBuffer] => Unit): F[Unit] =
+    withTxn(env.txnWrite())(f)
+
+  private[this] def withReadTxn[R](f: Txn[ByteBuffer] => R): F[R] =
+    withTxn(env.txnRead())(f)
+
+  private def readBlockMessage(offset: Long): F[BlockMessage] =
```
rabbitonweb: every call to `RandomAccessFile` is just closed over `Sync[F]` which (as mentioned before) will happly rethorw exception (when `Task` is used as instance). 

why this is bad can simply be seen in usage of `readBlockMessage`. This private method is for example used in `get` method. If while calling `get` the `readBlockMessage` explosdes, so will the `get` method. Your `clear` function (that you've added in this PR) will never be executed.

Instread `readBlockMessage should return errors as values from which you could easily recover (compoiler will guide you). By closing it over `delay` you allow swallowing of every exception that might happen

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
itegulov: One of the branches returns an empty `List`. The semantics here is that some of the files inside of the checkpoints directory could be named differently from what is expected and should be ignored. If the resulting list of checkpoint files is still satisfactory (i.e. starts from 0 and is consecutive) then we can proceed forward. This also answers you question about stopping rnode. I don't think it is a fatal error as having a *consistent* list of checkpoints is a necessary condition, not having a pure checkpoints directory.

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
itegulov: Answered above.

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
itegulov: I have replaced most of the creation methods' return types with `Either[StorageError, BlockStore[F]]`  and got rid of `Sync[F].raiseError` calls.

Regarding resource cleanup: I can make `create` methods to return a `Resource[F, BlockStore[F]]`. Which would basically force the caller to cleanup the block store resources. WDYT?

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
itegulov: The plan is to replace `LMDBBlockStore` with this one so `LMDBBlockStore` will have to go away eventually (once I merge all PRs regarding new block storage).

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
itegulov: I have improved error handling in `readBlockMessage` (and other IO methods as well) so it returns `StorageIOErrT[F, BlockMessage]` now. I have achieved this through calling `attempt` on every `Sync[F].delay { ioStuff }` call and wrapping IOExceptions into a blockstore ADT of errors. Not sure if this is the best way, feel free to comment.

Also, I have left the ADT of errors a subclass of exception because some of the older code in `blockstorage` module still tries to throw it through `Sync[F].raiseError`. I won't address it in this PR to keep it on point but will get back to it in another PR.

itegulov:
@rabbitonweb Thanks for your review!

I have addressed the remarks in your review. PTAL.

Here are the answers for you general questions:
1. This is just the first part of the block storage implementation (I have a few more to come). It indeed passes the unit tests but I was mainly struggling with existing integration tests and new integration tests (testing the node persistency along with the block dag storage enabled). As you can see this PR does not replace the in-mem block storage used in the actual rnode. This block store implementation is currently used only in tests.
2. A separate algebra for filesystem operations would definitely benefit the `blockstorage` module. I would argue that it falls out of scope for this particular PR though.

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
rabbitonweb: 👍

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
rabbitonweb: Ok, in that case consider cleaning it up. Use `Log.error("...", ex)` instead of `ex.printStackTrace`

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
rabbitonweb: 👍 
I'm more then happy to work with you on that with the next PR after this one

PR #2088, opened 2019-01-09
RCHAIN-2790 Make justifications visible with flag, sort validators in the graph

rabbitonweb:
## Overview
Make justifications visible with flag, sort validators in the graph

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2790

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
@@ -11,6 +11,7 @@ import cats.effect.Sync
 import cats._, cats.data._, cats.implicits._
 import cats.mtl._
 import cats.mtl.implicits._
+import coop.rchain.catscontrib.Catscontrib._
 import coop.rchain.catscontrib.ski._
 
 case class ValidatorBlock(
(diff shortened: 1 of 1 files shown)


casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -80,40 +83,27 @@ object GraphzGenerator {
...
-                        arrowHead = Some(NoneArrow)
-                      )
-                  )
-                parentsEdges *> justificationsEdges
-              }
-
-            }
+        _ <- config.showJustificationLines.fold(
+              drawJustificationDottedLines[G](g, validators),
+              ().pure[G]
```
lukasz-golebiewski: Is this line necessary?

rabbitonweb: yes
by default we will *NOT* draw justifictactions, only if `showJustificatioLines` flag is `true`
so here i'm folding over that flag, if `true` i draw, if `false` `(().pure[G])` which is a NOP

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -158,6 +148,47 @@ object GraphzGenerator {
       node = Map("width" -> "0", "height" -> "0", "margin" -> "0.03", "fontsize" -> "8")
     )
 
+  private def drawParentDependencies[G[_]: Applicative](
+      g: Graphz[G],
+      validators: List[(String, ValidatorsBlocks)]
```
lukasz-golebiewski: Why does this method take `List[(String, ValidatorsBlocks)]` if enough would be `List[ValidatorsBlocks]`?

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -158,6 +148,47 @@ object GraphzGenerator {
...
+        case ValidatorBlock(blockHash, parentsHashes, _) => {
+          parentsHashes
+            .traverse(p => g.edge(blockHash, p, constraint = Some(false)))
+        }
+      }
+      .as(())
+
+  private def drawJustificationDottedLines[G[_]: Applicative](
+      g: Graphz[G],
+      validators: Map[String, ValidatorsBlocks]
```
lukasz-golebiewski: Similar story here, you're using only the values so no need to pass the whole map with keys

PR #2096, opened 2019-01-10
Replace turan oracle with clique oracle

KentShikama:
This should help us debug any issues with the safety oracle
as now we aren't using a lower bound on the maximum clique size.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-2835

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -168,6 +168,13 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
           )
     } yield ()
 
+  /*
+   * On the first pass, block B is finalized if B's main parent block is finalized
+   * and the safety oracle says B's normalized fault tolerance is above the threshold.
+   * On the second pass, block B is finalized if any of B's children blocks are finalized.
+   *
+   * TODO: Implement the second pass in BlockAPI
+   */
   private def isGreaterThanFaultToleranceThreshold(
       dag: BlockDagRepresentation[F],
       blockHash: BlockHash
--- a/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
+++ b/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
@@ -48,7 +48,7 @@ object SafetyOracle extends SafetyOracleInstances {
 }
 
 sealed abstract class SafetyOracleInstances {
-  def turanOracle[F[_]: Monad: Log]: SafetyOracle[F] =
+  def cliqueOracle[F[_]: Monad: Log]: SafetyOracle[F] =
     new SafetyOracle[F] {
       def normalizedFaultTolerance(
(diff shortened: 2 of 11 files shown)


KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -53,108 +51,82 @@ sealed abstract class SafetyOracleInstances {
     new SafetyOracle[F] {
```
lukasz-golebiewski: `def turanOracle` is misleading now. 
If I understand correctly you want to fall back to the clique oracle implementation to ease debugging. 
Are you intending to go back to the Turan oracle in the near future? If so, why remove it?
Wouldn't it be beneficial to have both implementations available so we can freely switch between them?

KentShikama: No we won't go back to the turan oracle as far as I can see. We used to have it because we thought we might have all validators with equal weights but otherwise it is an inferior oracle.

KentShikama:
bors try

KentShikama:
@odeac PTAL

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -168,6 +168,13 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
           )
     } yield ()
 
+  /*
+   * On the first pass, block B is finalized if B's main parent block is finalized
+   * and the safety oracle says B's normalized fault tolerance is above the threshold.
+   * On the second pass, block B is finalized if any of B's children blocks are finalized.
```
lukasz-golebiewski: Is the threshold irrelevant in this case?

KentShikama: Yes

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -281,23 +256,30 @@ sealed abstract class SafetyOracleInstances {
                 false.pure[F]
               )
           }
+
+        def findMaximumClique(
```
lukasz-golebiewski: This should move to the `Clique`object and be separately tested. 
It no longer returns the max clique, but rather it's weight

KentShikama:
bors try

KentShikama:
@lukasz-golebiewski PTAL

casper/src/main/scala/coop/rchain/casper/util/Clique.scala
```diff
@@ -1,6 +1,17 @@
 package coop.rchain.casper.util
 
 object Clique {
+  def findMaximumCliqueByWeight[A](edges: List[(A, A)], weights: Map[A, Long]): Long = {
+    val maxWeight = weights.values.max
+    Clique
```
lukasz-golebiewski: I think the `Clique.` qualifier isn't needed in this context

PR #2098, opened 2019-01-10
voting, locker example contracts RCHAIN-2403 (WIP)

dckc:
## Overview
 - a port of the solidity voting contract, using ocaps
 - a locker contract for using digital signatures for accessing the ocaps from off-chain

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-2403

### Notes

This PR is to facilitate discussion / review; it has conflicts; I'm not sure why, but **do-not-merge** yet.

  - locker might be beyond the scope of RCHAIN-2403
  - I don't know why ByteArray commits are showing up here; #1968 is merged.

- [?] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),


dckc:
Note also https://github.com/dckc/RChain-API/blob/loader/test/testVoting.js 725da71  along with refactored voting.rho and locker.rho with the examples split out as chairRole.rho and voterRole.rho with the glue in the js code. WIP. some TODOs remain.

dckc:
regarding ByteArray stuff... I think what happened is: I re-based the byteArray stuff for #1968, but this branch uses the byteArray stuff before I rebased.

IOU a history fix-up here.

ArturGajowy:
I think you might want to specify the '--from' commit when doing the rebase (the last commit you **do not** want to be re-applied on top of upstream/dev). 

I think `git rebase --from 58eb657 upstream/dev` should do the trick

rholang/examples/locker.rho
```diff
@@ -0,0 +1,301 @@
...
+  new uriCh in {
+    insertArbitrary!(bundle+{*Locker}, *uriCh) | for(@uri <- uriCh) {
+      stdout!({"Locker registry uri": uri})
+    }
+  }
+  |
+
+  // WARNING: if sig is the wrong size, we get a java.lang.RuntimeException,
+  // and our nonceCh is stuck.
+  contract verify(
```
odeac: I don't understand the need for the nonce here. 

The nonce would be a token used to avoid reusing the same signature twice. In our case the signed data includes the return channel, which is an unforgeable name. The `Locker` will send its response to this ret channel so nobody else will have access to that data.

So in this case we don't really need to worry about replay attacks.

odeac: After the discussions in #cryptography discord channel I'm convinced that we do need the nonce. Still, I don't like the current implementation because the management of the nonces will be pretty complicated.

I have the following alternative proposal for the Locker contract (probably not syntactic correct):

```
contract Locker(
    @{label /\ String},
    @{pubKey /\ ByteArray},
    @seedCode, seedEntry,
    return) = {
    new self, seedCh in {
      return!(*self) |
      
      seedCode |
      seedEntry!(*seedCh) | for (seed <- seedCh) {
        contract self(@"getSeedProxy", @{sig /\ ByteArray}, return) = { 
          verify!(sig, pubKey, ("getSeedProxy", *return), *eject, *okCh) |
          for(_ <- okCh) {
            new proxy in {
              for (msg <- proxy) { seed ! (msg)} |
              return ! (proxy)
            }
          }
        }
      }
    }
  }
```

And now the Locker would be used like this:
```
Locker!("getSeedProxy", signature, return) | 
for (proxy <- return) { 
  // use the proxy
}```

dckc: @JoshOrndorff in this morning's discussion you gave some feedback on this code... were you saying that this replay prevention technique is not OK? It seems OK to me.

Modulo the kludge where  I split msg into 3 parts, which should be left to a separate layer, I think the code I wrote recently pretty much matches this sketch.
https://github.com/rchain/rchain/blob/fe3a8cd65f6da64810a6f1be3e144d1c7bb7efd2/casper/src/main/rholang/MakeRev.rho#L208-L269

Thinking again about splitting msg or not... the `seed ! (msg)` pattern only works for 1-ary seed contracts. So it wouldn't work with stuff like `purse!("getBalance", *ch)`, right? let alone `purse!("deposit", src, amount, return)`.

JoshOrndorff: My realization during your office hours was that it meant keeping track of all unforgeable names ever used somehow. That is guaranteed at the blockchain level and is not the problem of the contract. This will work fine once the assumption that the same unforgeable name cannot be generated twice holds.

> only works for unary contracts
I know that's a big bummer. It's why I've added "make all contracts unary by explicit tupling" to the librho style guide.

> Wouldn't work with stuff like `purse!("getBalance", *ch)`, right?
That's right. It's easier to make design changes to the purse now than it will be in the future.

FWIW, I've convinced myself that the best pattern is contracts called on compound names with args passed in via explicit tupling.

dckc:
This is overtaken by the 0.9 vault design.

PR #2101, opened 2019-01-11
RCHAIN-2838 - bool algebra on F

rabbitonweb:
## Overview

1. Introduce BooleanF
BooleanF represents boolean algebra operations on Boolean values that
are closed over F

2. Make usage of BooleanF in `computeAgreementGraphEdges`

### JIRA ticket:
RCHAIN-2838


ArturGajowy:
We already have operators for that in FlatMapBooleanOps:

```
final class FlatMapBooleanOps[F[_]: FlatMap](self: F[Boolean]) {
  def <&&>(other: => F[Boolean]): F[Boolean] = FlatMap[F].ifM(self)(other, self)
  def <||>(other: => F[Boolean]): F[Boolean] = FlatMap[F].ifM(self)(self, other)
}
```

Usage can bee seen in EqualsM:72

```
      Sync[F].delay(self.length == other.length) <&&>
        pairs.forallM(tupled(EqualM[A].equal[F]))
```

Won't that suffice?

rabbitonweb:
@ArturGajowy I will leave the BooleanF and remove FlatMapBooleanOps as BooleanF is restriced onl to Applicative

ArturGajowy:
@rabbitonweb requiring just `Applicative` won't work in a strict language :)

ArturGajowy:
Also, why the `^` in the operator's names? I find `<&&>` cleaner and more standard convention of saying "we're working inside the F context" than `&&^`.

Also you're adding an unused `BooleanOps` object that has api leading to:

```
import BooleanOps._
&&^(bool1, bool2)
//which might make sense now that I've written it
```

But is unused and could just as well be:

```
<&&>(bool1, bool2)
```

That is, if we're even able to call such methods without prefixing it with `BooleanOps.`.

rabbitonweb:
@ArturGajowy  `&&^` is inspired from haskell naming convention

all you have to do is to import `Catscontrib._` and then you can do `f1 &&6 f2` - that's it

rabbitonweb:
adding unused "BooleanOps" ? i've used it like yesterday and did not really add it? it was already there. i dont think i follow

shared/src/main/scala/coop/rchain/catscontrib/BooleanF.scala
```diff
@@ -0,0 +1,27 @@
...
+  def &&^(other: F[Boolean]): F[Boolean] =
+    BooleanF.&&^[F](self, other)
+  def ||^(other: F[Boolean]): F[Boolean] =
+    BooleanF.&&^[F](self, other)
+  def ~^(): F[Boolean] = BooleanF.~^[F](self)
+}
+
+object BooleanF {
+  def &&^[F[_]: Applicative](m1: F[Boolean], m2: F[Boolean]): F[Boolean] =
+    Applicative[F].map2(m1, m2)(_ && _)
```
ArturGajowy: I've meant this object - it's unused AFAIR

rabbitonweb:
Commit added: Replace ToFlatMapBooleanOps with BooleanF

image


 I'm biased towards BooleanF because:

1. constrained with Applicative only
2. following haskell naming conventions

shared/src/main/scala/coop/rchain/catscontrib/BooleanF.scala
rabbitonweb: I've used it in this PR :)

rabbitonweb: oh... you mean that there is both object and implict class that uses the object? that is style i've seen in both cats and scalaz and i've been using that ever since. you give API with methods that can be used in both was: as a function as a method on an object

ArturGajowy: You've used it in this very file only. It's effectively-unused and has API I'd rather not use anywhere else than this file.

`&&^(foo, &&^(bar, baz))` ?!

ArturGajowy:
I'm fine with your replacing of my operators with yours (:stuck_out_tongue:), because their name choices are of reasonably similar quality :smiling_imp:

I though can't stand that you're making my previously lazy condition in EqualsM now evaluate all the fields, leading to traversing the whole object tree when a short-circuit should've been made :)

Either switch to FlatMap or revert the last commit.

shared/src/main/scala/coop/rchain/catscontrib/BooleanF.scala
ArturGajowy: I'm OK if it's private. I'm ok if it's public too, it's not that bad - unless somebody nests calls like me above :)

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -269,17 +270,10 @@ sealed abstract class SafetyOracleInstances {
...
-                    false.pure[F]
-                  ),
-                  false.pure[F]
-                ),
-                false.pure[F]
-              )
+              seesAgreement(first, second) &&^ seesAgreement(second, first) &&^ neverEventuallySeeDisagreement(
+                first,
+                second
+              ) &&^ neverEventuallySeeDisagreement(second, first)
```
ArturGajowy: please also consider that the effects of all the conjuncts are going to be executed, which is a change to what's been the previous (lazy) semantics

PR #2105, opened 2019-01-11
Use random port in transport layer unit tests (let the JVM pick a port)

sebtomba:
## Overview
Use random port in transport layer unit tests (let the JVM pick a port)

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-1789

### DIFF
--- a/comm/src/test/scala/coop/rchain/comm/transport/TransportLayerRuntime.scala
+++ b/comm/src/test/scala/coop/rchain/comm/transport/TransportLayerRuntime.scala
@@ -1,10 +1,9 @@
 package coop.rchain.comm.transport
 
-import java.util.concurrent.atomic.AtomicInteger
+import java.net.ServerSocket
 
 import scala.collection.mutable
 import scala.concurrent.duration._
-import scala.util.Random
 
 import cats._
 import cats.effect.Timer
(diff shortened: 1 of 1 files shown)


comm/src/test/scala/coop/rchain/comm/transport/TransportLayerRuntime.scala
```diff
@@ -27,20 +27,24 @@ abstract class TransportLayerRuntime[F[_]: Monad: Timer, E <: Environment] {
 
   def extract[A](fa: F[A]): A
 
-  private val nextPort = new AtomicInteger((Random.nextInt(100) + 400 + 1) * 100)
+  private def getFreePort: Int =
+    Resources.withResource(new ServerSocket(0)) { s =>
+      s.setReuseAddress(true)
```
ArturGajowy: If I understand the docs correctly, this option only affects *this particular socket*. So the real user of that port won't be able to benefit from that setting. Am I missunderstanding sth? https://stackoverflow.com/a/23124217/52142

sebtomba: TBH, I don't really know if this flag has an impact on potential future use of the already closed socket. Maybe the flag makes the socket quicker available by the OS? I added this because all solutions to the *free port check* I saw on the internet are setting this flag. However, I don't think that it's doing any harm.

PR #2116, opened 2019-01-17
Modify MultiParentCasperRef to return default closed over effect

rabbitonweb:
## Overview
Whole approach of `MultiParentCasperRef` is **at least**
bizzare... Transitions are being made in `PacketHandler`, which sets a
value of a `MultiParentCasperRef`. This does not look as a right
design and should be addressed. In particular, node should be
constantly aware in which state it's in and react accordingly.

For now, in this change, we are giving a visibility **which* method
was used, to trigger `casper not available` output.

We must rememgber that `casper not available` is **not** an
error. This is normal that node might be in transition while receving
different message and command from api. `casper not available` is just
kind remainder to the caller "i'm not ready, yet".

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2855

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -39,7 +39,7 @@ object BlockAPI {
              }
       } yield re
 
-    val errorMessage = "Could not deploy, casper instance was not available."
+    val errorMessage = "Could not deploy, casper instance was not available yet."
 
     MultiParentCasperRef
       .withCasper[F, DeployServiceResponse](
@@ -53,7 +53,7 @@ object BlockAPI {
   def createBlock[F[_]: Sync: Concurrent: MultiParentCasperRef: Log](
       blockApiLock: Semaphore[F]
   ): F[DeployServiceResponse] = {
-    val errorMessage = "Could not create block, casper instance was not available."
+    val errorMessage = "Could not create block, casper instance was not available yet."
     MultiParentCasperRef.withCasper[F, DeployServiceResponse](
       casper => {
         Sync[F].bracket(blockApiLock.tryAcquire) {
@@ -90,7 +90,7 @@ object BlockAPI {
       listeningName: Par
   ): F[ListeningNameDataResponse] = {
 
-    val errorMessage = "Could not get listening name data, casper instance was not available."
+    val errorMessage = "Could not get listening name data, casper instance was not available yet."
 
     def casperResponse(implicit casper: MultiParentCasper[F]) =
       for {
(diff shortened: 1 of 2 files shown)


rabbitonweb:
This could be polished a bit before merging - especially fetching casper could be extracted to some external method - but that can always be in different PR. I just had to go to bed today :)

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -47,16 +47,21 @@ object BlockAPI {
              }
       } yield re
 
+    val errorMessage = "Could not deploy, casper instance was not available."
```
dzajkowski: "was not available yet." ?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -112,14 +123,18 @@ object BlockAPI {
...
+        .as(ListeningNameDataResponse(status = s"Error: $errorMessage"))
     )
   }
 
   def getListeningNameContinuationResponse[F[_]: Concurrent: MultiParentCasperRef: Log: SafetyOracle: BlockStore: ToAbstractContext](
       depth: Int,
       listeningNames: Seq[Par]
   ): F[ListeningNameContinuationResponse] = {
+    val errorMessage =
+      "Could not get listening names confinuation, casper instance was not available."
```
adaszko: confinuation

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -312,6 +340,10 @@ object BlockAPI {
       user: ByteString,
       timestamp: Long
   ): F[BlockQueryResponse] = {
+
+    val errorMessage =
+      "Could not find bloc with deploy, casper instance was not available."
```
adaszko:
```suggestion
      "Could not find block with deploy, casper instance was not available."
```

adaszko:
> We must rememgber that casper not available is not an
error. This is normal that node might be in transition while receving
different message and command from api. casper not available is just
kind remainder to the caller "i'm not ready, yet".

Please assume this is isn't obvious to the user.  It being displayed a warning is sufficiently suggestive that isn't an error but the ideal message would suggest what the user can do to rectify the problem.  Please ignore this comment if that belongs to other layers of the code.

goral09:
> Whole approach of `MultiParentCasperRef` is **at least**
> bizzare... Transitions are being made in `PacketHandler`, which sets a
> value of a `MultiParentCasperRef`. 
I don't totally agree with the _bizzare_ statement. `CasperPacketHandler` is the only input to the system that knows what to do with "Casper" packets and how to react. 

> In particular, node should be constantly aware in which state it's in and react accordingly.

This is EXACTLY what is happening. It is similar to akka actor's `become`. It's a state machine. 

That saying, I will be very happy to see cleaner design of that.

PR #2124, opened 2019-01-18
RCHAIN-2837 linting

rabbitonweb:
### Remove roscala and roscalaMacros as scala projects

Roscala is currently **not** being used.
Source code stays in repository but we remove it from configuration,
so it is not compiled, when the project is compiled.

### Introduce Wart.Remover

There is a lot of rules that were excluded currently. Some of them are
false-positives, but majority of them would just require a lot more
changes in this commit. They will be eventually removed from the
exclusion list.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2837

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -194,10 +194,6 @@ The [comm](comm) subproject contains code for network related operations for RCh
 
 The [rholang](rholang) subproject contains compiler related code for the Rholang language.
 
-### Rosette
-
-The [rosette](rosette) subproject contains code for a low level virtual machine for RChain.
-
 ### Rspace
 
 The [rspace](rspace) subproject contains code related to the key-value storage of the RChain blockchain.
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -194,10 +194,6 @@ The [comm](comm) subproject contains code for network related operations for RCh
 
 The [rholang](rholang) subproject contains compiler related code for the Rholang language.
 
-### Roscala
-
-The [roscala](roscala) subproject contains a Scala translation of the Rosette VM.
-
 ### Rosette
 
 The [rosette](rosette) subproject contains code for a low level virtual machine for RChain.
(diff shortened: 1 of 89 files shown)


DEVELOPER.md
```diff
@@ -194,10 +194,6 @@ The [comm](comm) subproject contains code for network related operations for RCh
 
 The [rholang](rholang) subproject contains compiler related code for the Rholang language.
 
-### Roscala
-
-The [roscala](roscala) subproject contains a Scala translation of the Rosette VM.
-
 ### Rosette
```
tymm: You can also remove the *Rosette* part.

shared/src/main/scala/coop/rchain/catscontrib/BooleanOps.scala
```diff
@@ -8,7 +10,7 @@ final class BooleanOps(self: Boolean) {
 
   final class ConditionalEither[A](a: => A) {
     def or[B](b: => B): Either[B, A] =
-      if (self) Right(a) else Left(b)
+      if (self) a.asRight[B] else b.asLeft[A]
```
ArturGajowy: What's wrong with `if (self) Right(a) else Left(b)`?

DEVELOPER.md
rabbitonweb: ok

shared/src/main/scala/coop/rchain/catscontrib/BooleanOps.scala
rabbitonweb: https://www.scala-lang.org/api/2.12.6/scala/util/Right.html

Right(a) does not even give you `Either[Nothing, A]`, it actually gives you `Right[Nothing, A]`
while `Left(b)` gives you `Left[B, Nothing]`

rabbitonweb:
I've decided to remove: `roscala`, `rosette` and `regex` in separate PR.

shared/src/main/scala/coop/rchain/catscontrib/BooleanOps.scala
ArturGajowy: I see. So it's better type inference (`Either[B, A]` in this case) vs some ops-wrapping overhead which is negligible most of the times. I guess we should strive for the better ergonomics (inference!) and override as needed for performance (if at all) :+1:

rabbitonweb:
Artur it is rebase to remove those commits already. You probably referring to commit that removes entries from build.sbt, correct?

PR #2128, opened 2019-01-18
Rchain 2791

rabbitonweb:
This is duplicate of https://github.com/rchain/rchain/pull/2114 because https://github.com/rchain/rchain/pull/2114 was messing drone somehow


woky:
bors try

PR #2140, opened 2019-01-24
purse security: mutual authentication (RCHAIN-2559)

dckc:
## Overview
The essential purse security property is: **if you can deposit into a purse you hold from an alleged purse, then the alleged purse was authentic**; i.e. it was actually a purse from the same mint and it conserved balances. The failing test case in 59e8a9b shows that's not the case for our current implementation and the fix (8f369b4) restores this property.

cc @odeac @cboscolo @JoshOrndorff 

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2559

### Notes

 - TODO: document the essential purse security property. Where?
   - The test case shows why the previous design didn't work, but perhaps it's worth further documentation? as well as some others I considered. My style would be to add those to MakeMint.rho, but comments in implementation code doesn't seem to be project style.


casper/src/main/rholang/MakeMint.rho
```diff
@@ -30,7 +56,17 @@ in {
         contract internalMakePurse(balance, return) = {
           //balance must be a name that NonNegativeNumber contract is listening on
           new thisPurse in {
-            contract thisPurse(@=*decr, @amount, success) = { balance!("sub", amount, *success) } |
+            new letCh in { // let thisDecr = ...
+              letCh!(bundle0{*thisPurse|*decr}) | for (thisDecr <- letCh) {
+                contract thisPurse(@=*thisDecr, @amount, return) = {
+                  new success in {
+                    balance!("sub", amount, *success) | for (@subOk <- success) {
+                      return!(subOk, *decr)
```
JoshOrndorff: Is returning `*decr` like signing the message?

dckc: yes; it's proving that you were created by the right mint.

Note that we don't hand it out to just anybody; only to those who had the mint to construct `bundle0{*thisPurse|*decr}`.

and just using `decr` there wouldn't work, because an attacker could call "deposit' with a fake purse and we'd hand `decr` to the fake purse. So what we hand out is specific to `src`.

dckc:
oops; this fix prevents all deposits from happening. The one-line fix is in 1a8e4a4 but belongs here.

casper/src/main/rholang/MakeMint.rho
```diff
@@ -18,6 +18,32 @@ new
 in {
   rl!(`rho:id:nd74ztexkao5awjhj95e3octkza7tydwiy7euthnyrt5ihgi9rj495`, *NonNegativeNumberCh) |
   for(@(_, NonNegativeNumber) <- NonNegativeNumberCh) {
+    new roguePurse1, trace(`rho:io:stderr`) in {
```
odeac: please write a Rholang unittest that fails before this change and succeeds after.

If you rebase on top of https://github.com/rchain/rchain/pull/2100 you will be able to use RhoSpec for it and your unittesting experience will be considerably better

odeac:
The tests are executed as part of casper unit tests.

```
$ sbt shell
...
sbt:rchain> project casper
...
sbt:casper> testOnly *MakeMintSpec
```

There is no documentation but you can find usage examples in casper/src/test/rholang/SuccessfulResultCollectorTest.rho

dckc:
Got it. Thanks.

dckc:
I think this is ready to go now, though it depends on RhoSpec (#2100). I'm not sure how this situation should be handled.

casper/src/main/rholang/MakeMint.rho
dckc: yes, like a hash.

odeac:
This looks good. I'll push for RhoSpec to be merged

dckc:
I see RhoSpec was merged... oh... I guess it was rebased, so I should do likewise.

odeac:
Yes. It was merged with minor changes. Please rebase

dckc:
argh! git is impossible to use...

dckc:
help?! somebody re-open this PR?

dckc:
I think 2fd8b7c70 is right.

dckc:
@odeac are the changes OK now?

(thanks, @KentShikama )

PR #2141, opened 2019-01-25
RCHAIN-2837 xfatal, var, throw

rabbitonweb:
## Overview

1. compilation warnings as errors for most modules (excluding casper)
2. Remove Wart.Var from exclusion list
3. Remove Wart.Throw
4. Some TODOs introduced - removing some var and throw warts was non-trivial and should be a matters of separate PR (thus `//TODO`)

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2837

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,6 +119,12 @@ lazy val graphz = (project in file("graphz"))
   .settings(commonSettings: _*)
   .settings(
     version := "0.1",
+    scalacOptions ++= Seq(
+      "-Xfatal-warnings",
+      "-unchecked",
+      "-deprecation",
+      "-feature"
+    ),
     libraryDependencies ++= commonDependencies ++ Seq(
       catsCore,
       catsEffect,
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,12 @@ lazy val shared = (project in file("shared"))
   .settings(commonSettings: _*)
   .settings(
     version := "0.1",
+    scalacOptions ++= Seq(
+      "-Xfatal-warnings",
+      "-unchecked",
+      "-deprecation",
+      "-feature"
+    ),
     libraryDependencies ++= commonDependencies ++ Seq(
       catsCore,
       catsEffect,
--- a/build.sbt
+++ b/build.sbt
@@ -24,9 +24,9 @@ lazy val projectSettings = Seq(
   wartremoverExcluded += sourceManaged.value,
   wartremoverErrors in (Compile, compile) ++= Warts.allBut(
     // those we want
-    Wart.DefaultArguments,
+    Wart.DefaultArguments, Wart.ImplicitParameter, Wart.ImplicitConversion,
     // those don't want
-    Wart.ImplicitParameter, Wart.Recursion, Wart.ImplicitConversion,
+    Wart.Recursion,
     Wart.LeakingSealed, Wart.Overloading, Wart.Nothing, Wart.NonUnitStatements,
     Wart.Equals, Wart.PublicInference, Wart.TraversableOps, Wart.ArrayEquals,
     Wart.Throw, Wart.While, Wart.Any, Wart.Product, Wart.Serializable, Wart.OptionPartial,
--- a/build.sbt
+++ b/build.sbt
@@ -23,9 +23,12 @@ lazy val projectSettings = Seq(
   ),
   wartremoverExcluded += sourceManaged.value,
   wartremoverErrors in (Compile, compile) ++= Warts.allBut(
-    Wart.ImplicitParameter, Wart.Recursion, Wart.DefaultArguments, Wart.ImplicitConversion,
+    // those we want
+    Wart.DefaultArguments,
+    // those don't want
+    Wart.ImplicitParameter, Wart.Recursion, Wart.ImplicitConversion,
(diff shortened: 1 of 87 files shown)


build.sbt
```diff
@@ -95,6 +98,12 @@ lazy val shared = (project in file("shared"))
   .settings(commonSettings: _*)
   .settings(
     version := "0.1",
+    scalacOptions ++= Seq(
```
tymm: Since you have this five times, might be worth creating a `val` for that.

regex/src/main/scala/coop/rchain/regex/Multiplier.scala
```diff
@@ -82,6 +82,7 @@ private[regex] object Multiplier {
   * also permit a max of 0 (iff min is 0 too). This allows the multiplier
   * "zero" to exist, which actually are quite useful in their own special way.
   */
+@SuppressWarnings(Array("org.wartremover.warts.Throw"))
```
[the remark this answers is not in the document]
rabbitonweb: in this particular case, because `rtegex` will be remove next week :)

in other places, sometimes there is a class (very imp3rative and ugly) and alternative would be like 5 or 6 warning. putting this all a class is an indicator "this need to be redone". i've always tried to add comment, maybe i've missed some cases?

build.sbt
rabbitonweb: I honestly want to at the end of the week push that to commonSettings - when all modules have it 
but for now, maybe val makes sense...

comm/src/main/scala/coop/rchain/comm/transport/buffer/ConcurrentQueue.scala
```diff
@@ -21,7 +21,7 @@ abstract class ConcurrentQueue[A] {
   def drain(buffer: mutable.Buffer[A], limit: Int): Unit
 }
 
-@SuppressWarnings(Array("org.wartremover.warts.Return"))
+@SuppressWarnings(Array("org.wartremover.warts.Return", "org.wartremover.warts.Var"))
```
ArturGajowy: the `var` is only used once, in `drain` - move the suppression there?

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -107,6 +107,7 @@ object CertificateHelper {
     info.set(X509CertInfo.ALGORITHM_ID, new CertificateAlgorithmId(algorithmId))
 
     // Sign the cert to identify the algorithm that's used.
+    @SuppressWarnings(Array("org.wartremover.warts.Var"))
     var cert = new X509CertImpl(info)
```
ArturGajowy: This one can be easily fixed by declaring `cert2` in line 117.

models/src/main/scala/coop/rchain/models/HashM.scala
```diff
@@ -155,6 +155,7 @@ object HashM extends HashMDerivation {
   implicit val PrivateNamePreviewQueryHash = gen[PrivateNamePreviewQuery]
 }
 
+@SuppressWarnings(Array("org.wartremover.warts.Var"))
```
ArturGajowy: Please only suppress that in `productHash`

node/src/main/scala/coop/rchain/node/diagnostics/NewPrometheusReporter.scala
```diff
@@ -11,6 +11,7 @@ import kamon.metric._
 /**
   * Based on kamon-prometheus but without the embedded server
   */
+@SuppressWarnings(Array("org.wartremover.warts.Var"))
```
ArturGajowy: can't we just flag the offending declaration instead of the whole class?

rspace/src/main/scala/coop/rchain/rspace/IStore.scala
```diff
@@ -104,6 +106,7 @@ trait IStore[C, P, A, K] {
 
   protected def processTrieUpdate(update: TrieUpdate[C, P, A, K]): Unit
 
+  @SuppressWarnings(Array("org.wartremover.warts.Var"))
   private[rspace] def getAndClearTrieUpdates(): Seq[TrieUpdate[C, P, A, K]] =
```
ArturGajowy: I don't see a `var` in this or the two previous methods. I think those suppressions are unneeded :)

rabbitonweb: This one is false-positive and unfortunate this is why supress is here

comm/src/main/scala/coop/rchain/comm/transport/buffer/ConcurrentQueue.scala
rabbitonweb: `ConcurrentQueue` is copy/paste from monix - in thi case I thought it is a good idea to keep all supresswarning upfront showing that sooner we move away from monix queues the better ;)

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
rabbitonweb: 👍

models/src/main/scala/coop/rchain/models/HashM.scala
rabbitonweb: 👍

PR #2145, opened 2019-01-28
Rchain 1117: make purse deposit fail on overflow

odeac:
## Overview
Make purse deposit fail on overflow

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-1117

### Notes


casper/src/main/rholang/MakeMint.rho
```diff
@@ -47,18 +49,27 @@ in {
...
+              new ch in {
+                balance!("add", amount, *ch) |
+                for(@addSuccess <- ch) {
+                  if(addSuccess) {
+                    @src!(*decr, amount, *ch) |
+                    for(@decrSuccess <- ch) {
+                      if (decrSuccess) {
+                        successCh!(true)
+                      } else {
+                        balance!("sub", amount, *successCh)
```
KentShikama: So in case you fail, you subtract the amount that you added? Why not just never add the amount in the first place?

casper/src/test/rholang/FailingResultCollectorTest.rho
```diff
@@ -1,8 +1,16 @@
...
-new assert(`rho:test:assert`), assertEquals(`rho:test:assertEquals`)  in {
+new assert(`rho:test:assertAck`),
+    testCompleted(`rho:test:testSuiteCompleted`),
+    ack
+in {
+  assert!("false should fail", false, "clue assertTrue", *ack) |
 
-  assert!("false should fail", false) |
+  for (_ <- ack) {
+    assert!("0 != 1 should fail", (0, "==", 1), "clue assertEquals", *ack) |
```
KentShikama:
```suggestion
    assert!("0 == 1 should fail", (0, "==", 1), "clue assertEquals", *ack) |
```

casper/src/main/rholang/MakeMint.rho
```diff
@@ -47,18 +49,27 @@ in {
...
-            contract thisPurse(@"deposit", @amount, @src, success) = {
-              new result in {
-                @src!(*decr, amount, *result) | //call src decr function.
-                for(@decrSuccess <- result) {
-                  if (decrSuccess) {
-                    balance!("add", amount, *success) // add transferred amount to this purse
+
+            contract thisPurse(@"deposit", @amount, @src, successCh) = {
+              new ch in {
+                balance!("add", amount, *ch) |
```
dckc: NonNegativeNumber already checks for overflow. I don't see why this change is needed.

dckc: oops; now I see; never mind...

casper/src/main/rholang/MakeMint.rho
```diff
@@ -47,18 +49,27 @@ in {
...
+                for(@addSuccess <- ch) {
+                  if(addSuccess) {
+                    @src!(*decr, amount, *ch) |
```
dckc: All possible failure modes have to be checked **before** the call to "decr".

See https://github.com/Agoric/PlaygroundVat/blob/master/examples/contract/makeMint.js#L42

casper/src/main/rholang/MakeMint.rho
```diff
@@ -47,18 +49,27 @@ in {
...
               }
             } |
-            contract thisPurse(@"deposit", @amount, @src, success) = {
-              new result in {
-                @src!(*decr, amount, *result) | //call src decr function.
-                for(@decrSuccess <- result) {
-                  if (decrSuccess) {
-                    balance!("add", amount, *success) // add transferred amount to this purse
+
+            contract thisPurse(@"deposit", @amount, @src, successCh) = {
```
dckc: Is changing `success` to `successCh` worthwhile? Is there a style guide we're trying to follow? It made the diff hard for me to read.

odeac: I'm not sure if we have a style guide but we should. I wanted to make it clear that the parameter is the return channel

casper/src/main/rholang/MakeMint.rho
odeac: what other failure modes are you thinking about?

casper/src/main/rholang/MakeMint.rho
odeac: The failure comes from the balance, which is a NonNegativeNumber, if I was to check before add I would duplicate the code in NonNegativeNumber

casper/src/test/rholang/FailingResultCollectorTest.rho
odeac: this was rebased on top of PR 2100 because I needed the tests. Please ignore the rest of the changes except MakeMint.rho and MakeMintTest.rho

casper/src/main/rholang/MakeMint.rho
dckc: Not other failure modes but the order of checking.

I'm saying something similar to what Kent is saying: check before changing any balances. Don't change and then undo.

dckc: Also... we need to make sure there is no observable state in between increasing the destination balance and decreasing the source balance. Somehow we need to do something like:

```
  for (@destBal <- destBalCh; @srcBal <- srcBalCh) {
    // check that srcBal >= amount and that destBal + amount doesn't overflow
    if (okToCommit) {
      destBalCh!(destBal + amount) | srcBalCh!(srcBal - amount)
    } else {
      destBal!(destBal) | srcBal!(srcBal)
    }
  }
```

In fact, when we get peek, we shouldn't even consume the balances from their channels until the commit point.

odeac: >> we need to make sure there is no observable state in between 
This is a very good point!

casper/src/test/rholang/EitherTest.rho
```diff
@@ -1,82 +1,81 @@
...
-  rl(`rho:registry:lookup`), EitherCh, TestSetCh,
+new
+  rl(`rho:registry:lookup`), EitherCh, RhoSpecCh,
+  stdlog(`rho:io:stdlog`),
+  double, divide, divide10By, divide8By,
+  /* TestSetCh,
   double, divide, divide100by, divide50by, divide10by,
-  divide8by, doubleMap, divide10byFlatMap, composeTest,
-  test1, test2, test3
+  divide8by, doubleMap, divide10byFlatMap, composeTest, */
```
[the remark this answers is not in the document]
odeac: this is due to the rebase on top of PR 2100. Will fix these after 2100 is merged

casper/src/main/rholang/MakeMint.rho
dckc: probably should be its own JIRA ticket

odeac: Still processing...

odeac: I've created a new JIRA ticket for this: https://rchain.atlassian.net/browse/RCHAIN-2926

PR #2147, opened 2019-01-28
 Abstract over effect type in SpatialMatcher

ArturGajowy:
## Overview

Title says it all.

### JIRA ticket:

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
@@ -1,0 +1,25 @@
+package coop.rchain.rholang.interpreter
+
+/**
+  * This is a lawless typeclass to abstract over types providing a `takeFirst()` method, as defined in
+  * NonDetFreeMapWithCost's ops.
+  *
+  * @tparam F
+  */
+trait Splittable[F[_]] {
+
+  def takeFirst[A](fa: F[A]): F[A]
+
+}
+
+object Splittable {
+
+  def apply[F[_]](implicit ev: Splittable[F]) = ev
+
+  implicit class SplittableOps[F[_]: Splittable, A](self: F[A]) {
+
+    def takeFirst(): F[A] =
+      Splittable[F].takeFirst[A](self)
+  }
+
+}
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
@@ -4,7 +4,7 @@ import cats.effect.Sync
 import cats.implicits._
 import cats.mtl.MonadState
 import cats.mtl.implicits._
-import cats.{Alternative, Monad, MonoidK, Eval => _}
+import cats.{Monad, MonoidK, Eval => _}
 import coop.rchain.catscontrib._
 import coop.rchain.models.Connective.ConnectiveInstance
 import coop.rchain.models.Connective.ConnectiveInstance._
@@ -194,21 +194,21 @@ object SpatialMatcher extends SpatialMatcherInstances {
           case Term(p) =>
             if (!lf.connectiveUsed(p)) {
               //match using `==` if pattern is a concrete term
-              charge[F](COMPARISON_COST) *> Alternative[F].guard(t == p)
+              charge[F](COMPARISON_COST) *> Alternative_[F].guard(t == p)
             } else {
               spatialMatch(t, p)
             }
           case Remainder(_) =>
             //Remainders can't match non-concrete terms, because they can't be captured.
             //They match everything that's concrete though.
-            charge[F](COMPARISON_COST) *> Alternative[F].guard(lf.locallyFree(t, 0).isEmpty)
+            charge[F](COMPARISON_COST) *> Alternative_[F].guard(lf.locallyFree(t, 0).isEmpty)
         }
         attemptOpt[F, FreeMap](isolateState[F, FreeMap](matchEffect))
       }
     val maximumBipartiteMatch = MaximumBipartiteMatch(memoizeInHashMap(matchFunction))
 
     for {
-      matches                <- Alternative[F].unite(maximumBipartiteMatch.findMatches(allPatterns, targets))
+      matches                <- Alternative_[F].unite(maximumBipartiteMatch.findMatches(allPatterns, targets))
       freeMaps               = matches.map(_._3)
       updatedFreeMap         <- aggregateUpdates(freeMaps)
       _                      <- _freeMap[F].set(updatedFreeMap)
@@ -248,7 +248,7 @@ object SpatialMatcher extends SpatialMatcherInstances {
   private def aggregateUpdates(freeMaps: Seq[FreeMap]): F[FreeMap] =
     for {
       currentFreeMap <- _freeMap[F].get
-      _ <- Alternative[F].guard {
+      _ <- Alternative_[F].guard {
             //The correctness of isolating MBM from changing FreeMap relies
             //on our ability to aggregate the var assignments from subsequent matches.
             //This means all the variables populated by MBM must not duplicate each other.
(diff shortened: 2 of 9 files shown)


ArturGajowy:
bors try

ArturGajowy:
bors try

shared/src/main/scala/coop/rchain/catscontrib/MonadError_.scala
```diff
@@ -0,0 +1,28 @@
+package coop.rchain.catscontrib
+import cats.MonadError
+
+trait MonadError_[F[_], E] extends ApplicativeError_[F, E] {
```
lukasz-golebiewski: It would make sense to test laws of this type class

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala
```diff
@@ -47,8 +47,12 @@ package object matcher {
       _           <- error.ensure(cost.get)(OutOfPhlogistonsError)(_.value >= 0)
     } yield ()
 
-  private[matcher] def attemptOpt[F[_], A](f: F[A])(implicit short: _short[F]): F[Option[A]] =
-    short.attempt(f).map(_.fold(_ => None, Some(_)))
+  private[matcher] def attemptOpt[F[_], A](
+      f: F[A]
+  )(implicit short: _short[F]): F[Option[A]] = {
+    import short.instance
+    f.attempt.map(_.fold(_ => None, Some(_)))
```
lukasz-golebiewski: What exactly is the improvement here?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.rholang.interpreter
+
+/**
+  * This is a lawless typeclass to abstract over types providing a `takeFirst()` method, as defined in
+  * NonDetFreeMapWithCost's ops.
+  *
+  * @tparam F
+  */
+trait Splittable[F[_]] {
```
lukasz-golebiewski: Adding this in complete separation without tests or usage makes it hard to capture the intent. This commit should be squashed with those containing usage and tests

shared/src/main/scala/coop/rchain/catscontrib/Alternative_.scala
```diff
@@ -0,0 +1,39 @@
+package coop.rchain.catscontrib
+
+import cats.{Alternative, Applicative, Bifoldable, Foldable, Monad, MonoidK}
+
+trait Alternative_[F[_]] extends MonoidK[F] {
```
lukasz-golebiewski: Is this type class also lawless?

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -30,22 +29,29 @@ import scala.collection.mutable
 // In a few places we use StateT[Stream, FreeMap, A] for backtracking. In order
 // to help cut down on backtracking, wherever one of several possible matches
 // will do, we just take one.
-trait SpatialMatcher[T, P] {
-  def spatialMatch(target: T, pattern: P): NonDetFreeMapWithCost[Unit]
+trait SpatialMatcher[T, P, F[_]] {
```
lukasz-golebiewski: `F[_]` should move to the first position as it is done in all other places. For consistence

shared/src/main/scala/coop/rchain/catscontrib/Alternative_.scala
ArturGajowy: All the `Foo_` typeclasses are merely wrappers around the corresponding `Foo` typeclass from cats - just elliminating the inheritance that leads to ambiguous implicits in presence of (most often) another Applicative instance.

I could (1) test laws for `Alternative_` by providing a test-only Applicative instance that bases on another Applicative instance (mind-blown.gif). I could (2) test the `Foo_` classes by providing a stub `Foo` and asserting results as expected. Or we could (3) agree that testing those is an overkill because they merely delegate all methods to the underlying cats `instance`.

Since (1) is an overkill, let's choose between (2) and (3). My feeling is (3) should be sufficient, but I'm fine doing (2) if you insist :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
ArturGajowy: There aren't tests. The usage is already there in SpatialMatcher - yet not via a typeclass (`allMatches.takeFirst()`, `allMathches` is of type `NonDetFreeMatchWithCost`). After changing from `type F = NonDetFreeMatchWithCost` to `F: Splittable` the code still compiles.

Splittable actually looks similar to a dual of MonoidK - so ComonoidK? It's not in cats though and when I asked, Luka Jacobovitz said it'd probably hard to have proper laws for that.

My proposal is we leave this as is. The only place I could logically squash that too is the already-huge last commit. WDYT?

shared/src/main/scala/coop/rchain/catscontrib/MonadError_.scala
ArturGajowy: See my comment down below - I'll apply our resolution from that discussion here as well

rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
lukasz-golebiewski: Fair enough. It should definitely be squashed together with a589606b96153484616762c27903b11f082b0a42. Let's leave the "already huge commit" as is

shared/src/main/scala/coop/rchain/catscontrib/Alternative_.scala
lukasz-golebiewski: I guess this would be just "nice to have". (1) sounds alright

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -346,108 +365,109 @@ trait SpatialMatcherInstances {
...
-          charge[F](cost) *> MonoidK[F].empty[Unit]
-        }
-      } else {
+  implicit def parSpatialMatcherInstance[F[_]: Splittable: Alternative: Monad: _error: _cost: _freeMap: _short]
+    : SpatialMatcher[Par, Par, F] = fromFunction[Par, Par, F] { (target, pattern) =>
+    if (!pattern.connectiveUsed) {
+      val cost = equalityCheckCost(pattern, target)
+      if (pattern == target)
+        charge[F](cost) *> ().pure[F]
+      else {
```
tymm: Curly-braces should be omitted here :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Splittable.scala
JosDenmark: `ComonoidK` would have two functions `cocombine[A](fa: F[A]): (F[A],F[A])` and `coempty[A](fa: F[A]): Unit`. If it's really desired, I don't see why it would be any harder to have laws than for `MonoidK`.

JosDenmark: But that would really only be helpful if you want to return the head of `NonDetFreeMatchWithCost` along with the dropped tail.

ArturGajowy: Ignoring the tail is not an issue. I'm though not sure if we can get away with having `cocombine` (or `split`?) return `(h, t)` for a list of `h :: t`). My hunch is that the split is a bit arbitrary and it is unclear what (if any) laws we could have for interactions with the Monoid part.

Anyway, I'm sure both our codebase, and the cats project, would enjoy a `ComonoidK` typeclass with a nice set of laws and at least a few imaginable useful instances :) I'll be happy to review such a PR and/or help make it happen if we can first sort out the laws :)

ArturGajowy: FWIW here's the discussion in cats gitter, with each reply being less comprehensible for me than the previous one :sweat_smile: Maybe it'll spark some ideas :) https://gitter.im/typelevel/cats?at=5c37d0609289cc7aa753882d

PR #2148, opened 2019-01-28
RCHAIN-2891: Make CostAccounting[M] a global implicit in DebruijnInterpreter[M,F]

JosDenmark:
# Overview
This is needed to give a more MTL-friendly structure to Reduce[M].

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2891

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -36,22 +36,21 @@ trait Reduce[M[_]] {
   def eval(par: Par)(
       implicit env: Env[Par],
       rand: Blake2b512Random,
-      sequenceNumber: Int,
-      costAccountingAlg: CostAccounting[M]
+      sequenceNumber: Int
   ): M[Unit]
 
   def inj(
       par: Par
-  )(implicit rand: Blake2b512Random, costAccountingAlg: CostAccounting[M]): M[Unit]
+  )(implicit rand: Blake2b512Random): M[Unit]
 
   /**
     * Evaluate any top level expressions in @param Par .
     */
-  def evalExpr(par: Par)(implicit env: Env[Par], costAccountingAlg: CostAccounting[M]): M[Par]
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -61,6 +60,7 @@ object Reduce {
       private val urnMap: Map[String, Par]
   )(
       implicit
+      costAccountingAlg: CostAccounting[M],
```
goral09: Just my 2 cents: this was NOT DONE this way on purpose. Interpreter should be decoupled from the `CostAccounting` logic. It should be layered on top of it using `ChargingReducer`.

JosDenmark:
@ArturGajowy Yep. I can do that in the next PR.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosDenmark: I figured, but the current implementation doesn't do that. It just adds a wrapper (ChargingReducer) around Reduce that delegates literally all of the calls to to Reduce, which handles the charging anyway. Reduce is parametric over M anyway, so it makes sense to pull state into the evaluation monad. That's far more parametric than an additional type-class.

goral09: It wasn't possible to do it back then but it should be now. `ChargingReducer` would be delegating core logic to its normal counterpart but handle the cost logic by itself. This requires charging counterparts for Substitute and every function within Reducer though.

ArturGajowy: This change is very close to code, easily-reversible, and it actually simplifies things. It might even not be problematic if/when we move cost accounting to a charging reducer. Anyway, yhx for the reminder about the intent behind it :)

JosDenmark:
bors try

woky:
bors try

woky:
bors try

ArturGajowy:
go home @bors[bot], you're drunk

PR #2153, opened 2019-01-30
RCHAIN-1052: Fix pretty printing referenced variables

tymm:
## Overview
Pretty printing `VarRef`s was not working.
Also added the possibility to separate match cases with semicolons which comes in handy for the added test case.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-1052

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
@@ -221,7 +221,7 @@ final case class PrettyPrinter(
           (pure("") /: m.cases.zipWithIndex) {
             case (string, (matchCase, i)) =>
               string |+| pure(indentStr * (indent + 1)) |+| buildMatchCase(matchCase, indent + 1) |+| pure {
-                if (i != m.cases.length - 1) " ;\n"
+                if (i != m.cases.length - 1) "\n"
                 else ""
               }
           } |+| pure("\n" + (indentStr * indent) + "}")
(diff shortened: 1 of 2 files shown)


rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -119,6 +119,7 @@ separator nonempty Branch "" ;
 -- Match Cases
 CaseImpl. Case ::= Proc13 "=>" Proc3 ;
 separator nonempty Case "" ;
+separator nonempty Case ";" ;
```
KentShikama: So a semi-colon is optional? Shouldn't we either make it mandatory or not?

tymm: Yes, this would make it optional. Making it mandatory would break a lot of code I guess. What are the objections to making it optional?

ArturGajowy: This allows the following:

```
match x0 {
  =x0 => Nil
  =x1 => Nil
  =x2 => Nil;
  =x3 => Nil;
```

which is a bit messy. I'd rather make the pretty-printer not output a `;`, or make it just `separator nonempty Case ";" ;` and fix the rholang we have so far.

Yeah, let's just fix the pretty-printer.

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -234,7 +234,7 @@ final case class PrettyPrinter(
             pure("{") |+| value.ps.map(buildStringM).toList.intercalate(pure(" \\/ ")) |+| pure("}")
           case ConnNotBody(value) => pure("~{") |+| buildStringM(value) |+| pure("}")
           case VarRefBody(value) =>
-            pure("=") |+| buildStringM(Var(FreeVar(value.index)))
+            pure("=") |+| pure(s"$freeId${freeShift - value.index - 1}")
```
ArturGajowy: please merge the pure-s :) 
```scala
pure(s"=${freeId}${freeShift - value.index - 1}")
```

ArturGajowy:
@tymm there seems to be a quite consistent error in the build - please do a successful `bors try ` before merging (so we have at least two successes when we include the following bors r+ build) and ping me if you need help diagnosing/fixing this :)

tymm:
bors try

tymm:
bors try

PR #2155, opened 2019-01-30
RCHAIN-2837 Remove NonUnitStatements from wart exclusion list

rabbitonweb:
## Overview
Remove NonUnitStatements from wart exclusion list. 

### From wartremover documentation:

Scala allows statements to return any type. Statements should only return Unit (this ensures that they’re really intended to be statements).
```
// Won't compile: Statements must return Unit
10
false
```
When interfacing with APIs that violate this principle, one can define a function in the project

```
@specialized def discard[A](evaluateForSideEffectOnly: A): Unit = {
  val _: A = evaluateForSideEffectOnly
  () //Return unit to prevent warning due to discarding value
}

discard{ badMutateFun(foo) } // Use like this
```

### JIRA ticket:
continued work on RCHAIN-2837

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
@@ -284,18 +284,19 @@ object Genesis {
     val bonds        = pubKeys.zipWithIndex.toMap.mapValues(_.toLong + 1L)
     val genBondsFile = genesisPath.resolve(s"bonds.txt").toFile
 
-    val skFiles = Capture[F].capture {
-      genesisPath.toFile.mkdir()
-      keys.foreach { //create files showing the secret key for each public key
-        case (sec, pub) =>
-          val sk      = Base16.encode(sec)
-          val pk      = Base16.encode(pub)
-          val skFile  = genesisPath.resolve(s"$pk.sk").toFile
-          val printer = new PrintWriter(skFile)
-          printer.println(sk)
-          printer.close()
-      }
-    }
+    val skFiles =
+      Capture[F].capture(genesisPath.toFile.mkdir()) >>
+        Capture[F].capture {
+          keys.foreach { //create files showing the secret key for each public key
+            case (sec, pub) =>
+              val sk      = Base16.encode(sec)
+              val pk      = Base16.encode(pub)
+              val skFile  = genesisPath.resolve(s"$pk.sk").toFile
+              val printer = new PrintWriter(skFile)
+              printer.println(sk)
+              printer.close()
+          }
+        }
 
     //create bonds file for editing/future use
     for {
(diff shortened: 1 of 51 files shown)


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -108,13 +108,14 @@ object Main {
       case Left(CouldNotConnectToBootstrap) =>
         log.error("Node could not connect to bootstrap node.")
       case Left(InitializationError(msg)) =>
-        log.error(msg)
-        Task.delay(System.exit(-1))
+        log.error(msg) *>
+          Task.delay(System.exit(-1))
```
KentShikama: So previously does the System.exit(-1) not run?

kayvank: same question as Kent. as I understand, there is no grantee in ordering with applicatives, so is int possible to exit before logging the error?

rabbitonweb: no, it would be `log.error(msg)` that was ignored
it returned a value (`Task[Unit]`) and that value was discarded

rabbitonweb:
@ArturGajowy what `discard` file? Can't find anything in docs beside

```To exclude a file or directory from all checks, use wartremoverExcluded in your build.sbt file:```

I prefer seeing `@SuppressWarning` because that shows you were the issue is. It suppose to hurt your eyes! That's the whole point. I can agree that maybe false-positives should go to the `wartremoverExcluded` - how about that?

**Comment by ArturGajowy** [comment]
`@ArturGajowy what discard file? Can't find anything in docs beside`

`discard` function. You're the one that copied the docs excerpt mentioning it to this PR's description :) I hope you'll agree that it shows where that particular wart takes place even better than the suppress warnings. If you feel `discard` is not screaming enough, we can invent a different name, like `suppressNonUnitStatement` or `runForSideEffects`

**Comment by ArturGajowy** [comment]
As agreed offline, there'll be another pr for using the discard function instead of the suppressions

**Comment by rabbitonweb** [comment]
@ArturGajowy agreed!

**Comment by ArturGajowy** [comment]
bors try

File: casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala [hunk 2]
```diff
@@ -284,18 +284,19 @@ object Genesis {
...
-          val sk      = Base16.encode(sec)
-          val pk      = Base16.encode(pub)
-          val skFile  = genesisPath.resolve(s"$pk.sk").toFile
-          val printer = new PrintWriter(skFile)
-          printer.println(sk)
-          printer.close()
-      }
-    }
+    val skFiles =
+      Capture[F].capture(genesisPath.toFile.mkdir()) *>
```
ArturGajowy: kaboom! :stuck_out_tongue:

node/src/main/scala/coop/rchain/node/Main.scala
ArturGajowy: `s/*>/>>`

ArturGajowy:
I've spotted two more usages of `*>` introduced in this PR, both should be `>>`. Could this be the reason for the failing integration test?

PR #2160, opened 2019-02-01
Introduce pure file system wrappers

itegulov:
## Overview
Introduces some pure abstractions over filesystem operations used in blockstorage module.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-652

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -282,7 +282,7 @@ object FileLMDBIndexBlockStore {
       checkpointsDirPath: Path
   ): StorageErrT[F, List[Checkpoint]] =
     for {
-      checkpointFilesList <- toStorageIOErrT(listFiles(checkpointsDirPath))
+      checkpointFilesList <- toStorageIOErrT(listRegularFiles(checkpointsDirPath))
       checkpoints <- EitherT.liftF[F, StorageError, List[Checkpoint]](
                       checkpointFilesList.flatTraverse { filePath =>
                         filePath.getFileName.toString match {
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -91,10 +91,10 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
   private def readBlockMessage(indexEntry: IndexEntry): StorageIOErrT[F, BlockMessage] = {
     def readBlockMessageFromFile(storageFile: RandomAccessIO[F]): StorageIOErrT[F, BlockMessage] =
       for {
-        _                      <- storageFile.seek(indexEntry.offset).toStorageIOErrT
-        blockMessageSize       <- storageFile.readInt.toStorageIOErrT
+        _                      <- EitherT(storageFile.seek(indexEntry.offset)).toStorageIOErrT
+        blockMessageSize       <- EitherT(storageFile.readInt).toStorageIOErrT
         blockMessagesByteArray = Array.ofDim[Byte](blockMessageSize)
-        _                      <- storageFile.readFully(blockMessagesByteArray).toStorageIOErrT
+        _                      <- EitherT(storageFile.readFully(blockMessagesByteArray)).toStorageIOErrT
         blockMessage           = BlockMessage.parseFrom(blockMessagesByteArray)
       } yield blockMessage
 
@@ -116,7 +116,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
                                       } { storageFile =>
                                         readBlockMessageFromFile(storageFile)
                                       } { storageFile =>
-                                        storageFile.close()
+                                        EitherT(storageFile.close()).toStorageIOErrT
                                       }
                                     case None =>
                                       EitherT
@@ -180,12 +180,12 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
                              getBlockMessageRandomAccessFile
                            )
         currentIndex              <- EitherT.liftF(getCurrentIndex)
-        endOfFileOffset           <- randomAccessFile.length.toStorageIOErrT
-        _                         <- randomAccessFile.seek(endOfFileOffset).toStorageIOErrT
+        endOfFileOffset           <- EitherT(randomAccessFile.length).toStorageIOErrT
+        _                         <- EitherT(randomAccessFile.seek(endOfFileOffset)).toStorageIOErrT
         (blockHash, blockMessage) = f
         blockMessageByteArray     = blockMessage.toByteArray
-        _                         <- randomAccessFile.writeInt(blockMessageByteArray.length).toStorageIOErrT
-        _                         <- randomAccessFile.write(blockMessageByteArray).toStorageIOErrT
+        _                         <- EitherT(randomAccessFile.writeInt(blockMessageByteArray.length)).toStorageIOErrT
+        _                         <- EitherT(randomAccessFile.write(blockMessageByteArray)).toStorageIOErrT
         _ <- EitherT.liftF[F, StorageIOError, Unit](withWriteTxn { txn =>
               index.put(
                 txn,
@@ -204,7 +204,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
         blockMessageRandomAccessFile <- EitherT.liftF[F, StorageIOError, RandomAccessIO[F]](
                                          getBlockMessageRandomAccessFile
                                        )
-        _                               <- blockMessageRandomAccessFile.close().toStorageIOErrT
+        _                               <- EitherT(blockMessageRandomAccessFile.close()).toStorageIOErrT
         _                               <- moveFile(storagePath, checkpointPath, StandardCopyOption.ATOMIC_MOVE).toStorageIOErrT
         newBlockMessageRandomAccessFile <- RandomAccessIO.open[F](storagePath).toStorageIOErrT
         _ <- EitherT.liftF[F, StorageIOError, Unit](
@@ -226,7 +226,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
         _ <- withWriteTxn { txn =>
               index.drop(txn)
             }
-        result <- blockMessageRandomAccessFile.setLength(0).toStorageIOErrT.value
+        result <- EitherT(blockMessageRandomAccessFile.setLength(0)).toStorageIOErrT.value
       } yield result
     )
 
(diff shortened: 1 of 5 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/util/io/RandomAccessIO.scala
```diff
@@ -0,0 +1,71 @@
...
+import cats.data.EitherT
+import cats.effect.Sync
+import cats.implicits._
+import IOError.IOErrT
+
+import scala.language.higherKinds
+
+final case class RandomAccessIO[F[_]: Sync](file: RandomAccessFile) {
+  def close(): IOErrT[F, Unit] =
+    EitherT(Sync[F].delay { file.close() }.attempt).leftMap[IOError] {
```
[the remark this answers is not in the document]
dzajkowski: `EitherT(Sync[F].delay { action }.attempt).leftMap[IOError]` is repeated like a lot. maybe it would make sense to figure out either what @goral09 suggested or at least extract it to a helper?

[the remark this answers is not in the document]
sebtomba: I would prefer a different approach without `Either` and `EitherT`. The main reason for this mapping is because `IOException` is not specific enough. We are mapping to a more specific error. `Sync` extends `ApplicativeError` for `Throwable`. Why not use `Sync[F].raiseError` here? However, `IOError` would have to extend `Throwable` to satisfy the error type. But the return type could then be `F[Unit]`. The layer above this functions could then take a happy path approach, call `attempt` over the whole program and then map the left side to `IOError` (with a partial function).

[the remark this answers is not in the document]
goral09: What I don't like with the 
```scala
def openFile[F[_]: MonadError[?[_], Error]](file: Path): F[Unit] = ???
```
the approach is that type system doesn't enforce (or at least points at) how the error should be handled because it [error] is now hidden inside `MonadError` and caller of this method has to extract it explicitly. With opposed to:
```scala
def openFile[F[_]](file: Path): F[Either[Error, Unit]] = ???
```
where an error is explicit and the caller has to map over the returned values explicitly. He [caller] can then make decision to ignore the error (by simple mapping `F[Either[E, Unit]]` to `F[Unit]` or handle it and then map. 

I am curious to hear your opinion on that.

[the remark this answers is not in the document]
itegulov: @sebtomba I undestand your point that there are still 2 error channels but @rabbitonweb mentioned in our [previous conversation](https://github.com/rchain/rchain/pull/2076#discussion_r247084165) that `Sync[F].raiseError` should only be used for fatal errors and agree with him on that. Since I believe this kind of IO errors could be handled by, for example, block storage in some meaningful way I decided to return explicit errors here instead of throwing them with `Sync[F]`.

Your and @goral09's point on returning `F[Either[IOErr, Unit]]` is completely reasonable though.

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/RandomAccessIO.scala
```diff
@@ -0,0 +1,55 @@
...
+import cats.effect.Sync
+import cats.implicits._
+import IOError.{IOErr, IOErrT}
+
+import scala.language.higherKinds
+
+final case class RandomAccessIO[F[_]: Sync](file: RandomAccessFile) {
+  private def handleIo[A](io: => A, handleIoException: IOException => IOError): F[IOErr[A]] =
+    Sync[F].delay { io }.attempt.map[IOErr[A]] {
+      case Left(e: IOException) => Left(handleIoException(e))
```
ArturGajowy: :clap: :+1:

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/RandomAccessIO.scala
```diff
@@ -0,0 +1,55 @@
...
+import java.nio.file.Path
+
+import cats.data.EitherT
+import cats.effect.Sync
+import cats.implicits._
+import IOError.{IOErr, IOErrT}
+
+import scala.language.higherKinds
+
+final case class RandomAccessIO[F[_]: Sync](file: RandomAccessFile) {
```
ArturGajowy: maybe it'd make sense to encapsulate the `file`, so that receivers of `RandomAccessIO` don't have access to the impure interface underneath

itegulov: Good point

goral09:
@ArturGajowy @sebtomba haven't you two come to an agreement that it can be done better?

ArturGajowy:
@goral09 WDYM by `that`?

goral09:
`FunctorRaise` with `Applicative.pure(()).map(=> f) `

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -0,0 +1,74 @@
...
+  def makeDirectory[F[_]: Sync](dirPath: Path): F[IOErr[Unit]] =
+    handleIo(dirPath.toFile.mkdir(), UnexpectedIOError.apply)
+
+  def listInDirectory[F[_]: Sync](dirPath: Path): F[IOErr[List[Path]]] =
+    (for {
+      _ <- EitherT(makeDirectory(dirPath))
+      files <- EitherT(handleIo(Files.list(dirPath), {
+                case e: NotDirectoryException => FileIsNotDirectory(e)
+                case e                        => UnexpectedIOError(e)
+              }))
```
ArturGajowy: Same as below, we shouldn't be using any EitherT in here and should instead propagate errors upstream.

This method (`listInDirectory`) should fail for non-directory paths. Otherwise we're making the users delusional about them interacting with a directory while they're not :D

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -87,25 +89,13 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
...
+        blockMessageSize       <- toStorageIOErrT(storageFile.readInt)
         blockMessagesByteArray = Array.ofDim[Byte](blockMessageSize)
-        _ <- EitherT(Sync[F].delay { storageFile.readFully(blockMessagesByteArray) }.attempt)
-              .leftMap[StorageIOError] {
-                case e: IOException => ByteArrayReadFailed(e)
-                case e              => UnexpectedIOStorageError(e)
-              }
-        blockMessage = BlockMessage.parseFrom(blockMessagesByteArray)
+        _                      <- toStorageIOErrT(storageFile.readFully(blockMessagesByteArray))
+        blockMessage           = BlockMessage.parseFrom(blockMessagesByteArray)
```
ArturGajowy: This is a **very** neat refactoring. I love how IDEA shows how boilerplate and noise is disappearing around the actual method calls that remain unchanged :heart:

This made me less cautious though and I didn't check that the boilerplate was faithfully moved to the pure wrapper. It doesn't matter though, as we rarely (never?) actually handle those errors other than by terminating execution. So even if it ends up being an UnexpectedError - it's fine.

Good job! :tada:

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -0,0 +1,74 @@
...
+        .asScala
+        .toList
+    } yield filesList).value
+
+  def listFiles[F[_]: Sync](dirPath: Path): F[IOErr[List[Path]]] = {
+    type IOErrTF[A] = IOErrT[F, A]
+    (for {
+      inDirectoryList <- EitherT(listInDirectory(dirPath))
+      directoryList   <- inDirectoryList.filterA[IOErrTF](f => EitherT(isRegularFile(f)))
+    } yield directoryList).value
```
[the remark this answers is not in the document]
itegulov: Why do you think that this will silently return no files in presence of an error? Maybe I am missing something but isn't the whole `EitherT` purpose to propagate erros without having to deal with boilerplate of stacking `Either` with `F`?

See a small example of how I am assuming EitherT behaves here: https://scastie.scala-lang.org/aLsjdebDRUigbSdcUzekrw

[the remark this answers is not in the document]
ArturGajowy: You're right, I must've been drunk when writing that :man_facepalming:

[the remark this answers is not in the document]
ArturGajowy: I guess I wanted to get rid of the eithert wrapper and came up with reasons that are obviously false. I've got an idea how to do that, but I'll share it later (probably in the other PR) :)

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/errors.scala
```diff
@@ -0,0 +1,75 @@
...
+final case class FileSecurityViolation(exception: SecurityException)        extends IOError
+final case class FileIsNotDirectory(exception: NotDirectoryException)       extends IOError
+final case class UnsupportedFileOperation(e: UnsupportedOperationException) extends IOError
+final case class FileAlreadyExists(e: FileAlreadyExistsException)           extends IOError
+final case class DirectoryNotEmpty(e: DirectoryNotEmptyException)           extends IOError
+final case class AtomicMoveNotSupported(e: AtomicMoveNotSupportedException) extends IOError
+final case class UnexpectedIOError(throwable: Throwable)                    extends IOError
+
+object IOError {
+  type IOErr[A]        = Either[IOError, A]
```
odeac: I find this name misleading. You have code like ```someFunction(a) : F[IOErr[String]]``` and you would read this as "somefunction returns an error wrapped in F"

Maybe `IO` or `IOAction` would be a better name

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import cats.data.EitherT
+import cats.implicits._
+import cats.effect.Sync
+import coop.rchain.blockstorage.util.io.IOError.{IOErr, IOErrT}
+
+import scala.collection.JavaConverters._
+
+package object io {
+  private[io] def handleIo[F[_]: Sync, A](
```
odeac: I don't think this "handleIo" is a good name.

when I read code like ```handleIo(Files.move(...), {case e: FileAlreadyExistsException => ...})``` I get the feeling that the exception is being handled.

In fact you are only wrapping&delaying the action so this is why I find the name misleading.

I would rather have something like:
```scala
object IOAction {
  def apply[A](a : => A) = {...}
}

def moveFile(...)  : F[IOAction[Path]] = IOAction(Files.move(from, to, options: _*))
```

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -0,0 +1,76 @@
...
+      io: => A,
+      handleIoException: IOException => IOError
+  ): F[IOErr[A]] =
+    Sync[F].delay {
```
rabbitonweb: this change was from `Sync.delay.attempt` to `Sync.delay(try-catch)` - why is that?

ArturGajowy: Because of this comment: https://github.com/rchain/rchain/pull/2160#discussion_r254684151

I hope you'll agree that it's still pure externally?

rabbitonweb:
I think `Fix RandomAccessIO opening compilation errors` commit should be squashed to `Change remaining pure io functions' types to `IOErr` instead of `IOErrT`` since it  was fixing compilation issue

ArturGajowy:
bors r+

Let's get this in.

PR #2164, opened 2019-02-03
Fix parent hash empty case in parents validation

KentShikama:
Note the only time that genesis validation happens is
at the beginning of the chain. Hence, this would
never really have appeared as a bug: it just makes
no logical sense to label a block with an empty parent
hash as a last finalized block that isn't genesis -
as that would have a parent.

I've also added some logging to debug the invalid parents
warning we are getting on testnet.

### JIRA ticket:

Part of https://rchain.atlassian.net/browse/RCHAIN-2909

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -199,7 +199,7 @@ object Validate {
                         _ => Validate.justificationFollows[F](block, genesis, dag)
                       )
       parentsStatus <- followsStatus.joinRight.traverse(
-                        _ => Validate.parents[F](block, lastFinalizedBlockHash, dag)
+                        _ => Validate.parents[F](block, genesis, lastFinalizedBlockHash, dag)
                       )
       sequenceNumberStatus <- parentsStatus.joinRight.traverse(
                                _ => Validate.sequenceNumber[F](block, dag)
@@ -423,12 +423,13 @@ object Validate {
     */
   def parents[F[_]: Monad: Log: BlockStore](
       b: BlockMessage,
+      genesis: BlockMessage,
       lastFinalizedBlockHash: BlockHash,
       dag: BlockDagRepresentation[F]
   ): F[Either[InvalidBlock, ValidBlock]] = {
     val maybeParentHashes = ProtoUtil.parentHashes(b)
     val parentHashes = maybeParentHashes match {
-      case hashes if hashes.isEmpty => Seq(lastFinalizedBlockHash)
+      case hashes if hashes.isEmpty => Seq(genesis.blockHash)
       case hashes                   => hashes
     }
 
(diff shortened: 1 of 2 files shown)


dzajkowski:
bors try

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -392,22 +392,24 @@ class ValidateTest
...
-                       _ <- Validate.parents[Task](b9, b0.blockHash, dag)
+                       _ <- Validate.parents[Task](b7, b0, b0.blockHash, dag)
+                       _ <- Validate.parents[Task](b8, b0, b0.blockHash, dag)
+                       _ <- Validate.parents[Task](b9, b0, b0.blockHash, dag)
 
                        _ = log.warns.size should be(3)
                        result = log.warns.forall(
-                         _.contains("block parents did not match estimate based on justification")
+                         _.matches(
+                           "Ignoring block .* because block parents .* did not match estimate .* based on justification .*"
```
lukasz-golebiewski: One of the messages could be checked whether it contains correct hashes

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -442,7 +443,12 @@ object Validate {
                else
                  for {
                    _ <- Log[F].warn(
-                         ignore(b, "block parents did not match estimate based on justification.")
+                         ignore(
+                           b,
+                           s"block parents ${parentHashes.map(parentHash => PrettyPrinter.buildString(parentHash)).mkString(",")} did not match estimate ${computedParentHashes
```
lukasz-golebiewski: This message is hard to read when looking at the code. Perhaps it would make sense to extract the variable parts of it

ArturGajowy:
bors try

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
KentShikama: I don't think the hashes are deterministic.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -423,12 +423,13 @@ object Validate {
     */
   def parents[F[_]: Monad: Log: BlockStore](
       b: BlockMessage,
+      genesis: BlockMessage,
```
dzajkowski: wouldn't it make sense to start passing `genesisHash` instead of the whole genesis block?

casper/src/main/scala/coop/rchain/casper/Validate.scala
dzajkowski: maybe a helper in PrettyPrinter?
```
  def buildString(hashesh: Seq[ByteString]) =
    s"${hashesh.map(hash => PrettyPrinter.buildString(hash)).mkString(",")}"
```

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
lukasz-golebiewski: It's not a problem if you use references

PR #2169, opened 2019-02-05
RCHAIN-2892: Fix EMethod.connectiveUsed being ignored while sorting

tymm:
## Overview
Title says it all.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2892

### DIFF
--- a/models/src/main/scala/coop/rchain/models/rholang/sorter/ExprSortMatcher.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sorter/ExprSortMatcher.scala
@@ -253,13 +253,15 @@ private[sorter] object ExprSortMatcher extends Sortable[Expr] {
           )
       case EMethodBody(em) =>
         for {
-          args         <- em.arguments.toList.traverse(Sortable[Par].sortMatch[F])
-          sortedTarget <- Sortable.sortMatch(em.target)
+          args                <- em.arguments.toList.traverse(Sortable[Par].sortMatch[F])
+          sortedTarget        <- Sortable.sortMatch(em.target)
+          connectiveUsedScore = if (em.connectiveUsed) 1 else 0
         } yield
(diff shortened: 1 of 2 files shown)


tymm:
bors try

tymm:
@KentShikama an example or a test case? I'm not quite sure how either would help here? :)

KentShikama:
An example. It is for my understanding because as of now my belief is that connectives don't make sense with methods.

tymm:
Oh, I see. An at least nonsensical example would be `{1 : "a"}.union({2 : Bool})` but not even sure if this should be allowed.

So yeah, I see your point. At least right now I think we don't use connectives with methods.
However I could imagine something like for example a `filter` or `find` method where connectives would make sense.

KentShikama:
I'm confused does `{1 : "a"}.union({2 : Bool})` contain connectives?  My understanding of a connective was the use of the symbols `\/` or `/\`.

tymm:
`Bool`, `Uri` and such are also handled as connectives.
https://github.com/rchain/rchain/blob/cb555b6b0b5d1e07982adf177d8c57b57bf4f47b/models/src/main/protobuf/RhoTypes.proto#L368

The more I think about it, the more I think the example I posted should not be allowed.
@ArturGajowy do you agree?

tymm:
bors try

ArturGajowy:
> The more I think about it, the more I think the example I posted should not be allowed.
`{1 : "a"}.union({2 : Bool})`

I'm not sure. I can imagine having a language where types are first-class citizens (values) as well, in which case it makes sense to put them into a map.

tymm:
bors try

tymm:
bors try

tymm:
bors try-

tymm:
bors try

PR #2185, opened 2019-02-13
RCHAIN-2941 Introduce lazy serializedSizeM

rabbitonweb:
### Issue (what was):

Memo class is implements an memoization concept. It has a thunk that will be “nulled” (for GC to collect it).
The problem is, that memory is freed only when calculcation is executed (get method is called and evaluated). Which is not always the case (in memory block storage, genesis block in memory). All proto generated objects have that memo value added.
This can eat up 10% of all heap!

### Fix (what is):

We init the value when it's needed (and also evaluated thus freeing memory)

Fixes issues described in assign https://rchain.atlassian.net/browse/RCHAIN-2876?focusedCommentId=13463&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13463 

### JIRA ticket:
RCHAIN-2941

### DIFF
--- a/project/StacksafeScalapbGenerator.scala
+++ b/project/StacksafeScalapbGenerator.scala
@@ -132,8 +132,6 @@ class StacksafeProtobufGenerator(params: GeneratorParams) extends ProtobufGenera
     super
       .generateSerializedSize(message)(withEqualsAndHashCode)
       .newline
-      .add("import monix.eval.Coeval")
-      .newline
       .add("@transient var _serializedSizeM: coop.rchain.models.Memo[Int] = null")
       .newline
       .add("def serializedSizeM: coop.rchain.models.Memo[Int] = if(_serializedSizeM == null) {")
--- a/project/StacksafeScalapbGenerator.scala
+++ b/project/StacksafeScalapbGenerator.scala
@@ -132,10 +132,18 @@ class StacksafeProtobufGenerator(params: GeneratorParams) extends ProtobufGenera
     super
       .generateSerializedSize(message)(withEqualsAndHashCode)
       .newline
-      .add(
-        "@transient val serializedSizeM = new coop.rchain.models.Memo(coop.rchain.models.ProtoM.serializedSize(this))"
-      )
+      .add("import monix.eval.Coeval")
       .newline
+      .add("@transient var _serializedSizeM: coop.rchain.models.Memo[Int] = null")
+      .newline
+      .add("def serializedSizeM: coop.rchain.models.Memo[Int] = if(_serializedSizeM == null) {")
+      .newline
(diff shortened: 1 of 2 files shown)


project/StacksafeScalapbGenerator.scala
```diff
@@ -132,10 +132,16 @@ class StacksafeProtobufGenerator(params: GeneratorParams) extends ProtobufGenera
     super
       .generateSerializedSize(message)(withEqualsAndHashCode)
       .newline
-      .add(
-        "@transient val serializedSizeM = new coop.rchain.models.Memo(coop.rchain.models.ProtoM.serializedSize(this))"
-      )
+      .add("@transient var _serializedSizeM: coop.rchain.models.Memo[Int] = null")
```
ArturGajowy: I think you've meant `@volatile` here

project/StacksafeScalapbGenerator.scala
```diff
@@ -132,10 +132,16 @@ class StacksafeProtobufGenerator(params: GeneratorParams) extends ProtobufGenera
...
+      .newline
+      .add("def serializedSizeM: coop.rchain.models.Memo[Int] = if(_serializedSizeM == null) {")
```
ArturGajowy: we can either both dig into wikipedia and discuss why this is not a proper double-checked locking implementation and why it may or may not work and why us even having this discussion means that this is hard to maintain...

... or we can use `synchronized` here. Pretty please? :shrek-cat: :)

ArturGajowy: Then the volatile is not needed, I think.

rabbitonweb: forgot :)

project/StacksafeScalapbGenerator.scala
rabbitonweb: it was transient before, i though it was for a reason (when being serialized). you tell me, you wrote it.

ArturGajowy: ah, right - it's needed so that this doesn't get serialized. :+1: So it should be `@transient`, but not `@volatile` given my latter comment (let's use synchronized)

rabbitonweb:
Generates 

```
    @transient var _serializedSizeM: coop.rchain.models.Memo[Int] = null
    
    def serializedSizeM: coop.rchain.models.Memo[Int] = synchronized {
      if(_serializedSizeM == null) {
        _serializedSizeM = new coop.rchain.models.Memo(coop.rchain.models.ProtoM.serializedSize(this))
        _serializedSizeM
      } else _serializedSizeM
    }

```

lukasz-golebiewski:
It would be nice if some test guarded the desired behavior, otherwise this might come back

PR #2174, opened 2019-02-07
Improve DAG storage

itegulov:
## Overview
Various improvements to DAG storage implementation such as migration to pure IO wrappers (instead of using Java IO directly), fixing locking issues and using MonadState instead of Ref to represent internal DAG storage state.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-659

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [ ] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2102, opened 2019-01-11
Block Storage Checkpointing

itegulov:
## Overview
Title says it all.
It is relative to https://github.com/rchain/rchain/pull/2076

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-652

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2186, opened 2019-02-14
Improve DAG storage

itegulov:
## Overview
Various improvements to DAG storage implementation such as migration to pure IO wrappers (instead of using Java IO directly), fixing locking issues and using MonadState instead of Ref to represent internal DAG storage state.

A duplicate of #2174 but with dev branch as a base.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-659

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -296,13 +296,7 @@ object FileLMDBIndexBlockStore {
       storagePath: Path,
       checkpointsDirPath: Path
   ): F[StorageErr[BlockStore[F]]] = {
-    implicit val raiseIOErrorThroughSync = new FunctorRaise[F, IOError] {
-      override val functor: Functor[F] =
-        Functor[F]
-
-      override def raise[A](e: IOError): F[A] =
-        Sync[F].raiseError(IOError.ExceptionWrapper(e))
-    }
+    implicit val raiseIOError: RaiseIOError[F] = IOError.raiseIOErrorThroughSync[F]
     for {
       lock <- Semaphore[F](1)
       index <- Sync[F].delay {
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -276,6 +276,11 @@ object FileLMDBIndexBlockStore {
   ): StorageIOErrT[F, A] =
     EitherT(ioError).toStorageIOErrT
 
+  private[blockstorage] def toStorageErrT[F[_]: Sync, A](
+      ioError: F[IOErr[A]]
+  ): StorageErrT[F, A] =
+    EitherT(ioError).toStorageErrT
+
   private def loadCheckpoints[F[_]: Sync: Log](
       checkpointsDirPath: Path
   ): StorageErrT[F, List[Checkpoint]] =
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -113,9 +113,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
                                     case Some(checkpoint) =>
                                       type StorageIOErrTF[A] = StorageIOErrT[F, A]
                                       Sync[StorageIOErrTF].bracket {
-                                        toStorageIOErrT(
-                                          RandomAccessIO.open[F](checkpoint.storagePath)
-                                        )
+                                        RandomAccessIO.open[F](checkpoint.storagePath)
                                       } { storageFile =>
                                         readBlockMessageFromFile(storageFile)
                                       } { storageFile =>
@@ -207,9 +205,9 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
         blockMessageRandomAccessFile <- EitherT.liftF[F, StorageIOError, RandomAccessIO[F]](
                                          getBlockMessageRandomAccessFile
                                        )
-        _                               <- toStorageIOErrT(blockMessageRandomAccessFile.close())
-        _                               <- toStorageIOErrT(moveFile(storagePath, checkpointPath, StandardCopyOption.ATOMIC_MOVE))
-        newBlockMessageRandomAccessFile <- toStorageIOErrT(RandomAccessIO.open[F](storagePath))
+        _                               <- EitherT(blockMessageRandomAccessFile.close()).toStorageIOErrT
+        _                               <- moveFile(storagePath, checkpointPath, StandardCopyOption.ATOMIC_MOVE).toStorageIOErrT
+        newBlockMessageRandomAccessFile <- RandomAccessIO.open[F](storagePath).toStorageIOErrT
         _ <- EitherT.liftF[F, StorageIOError, Unit](
               setBlockMessageRandomAccessFile(newBlockMessageRandomAccessFile)
             )
@@ -329,7 +327,7 @@ object FileLMDBIndexBlockStore {
       index <- EitherT.liftF[F, StorageError, Dbi[ByteBuffer]](Sync[F].delay {
                 env.openDbi(s"block_store_index", MDB_CREATE)
               })
-      blockMessageRandomAccessFile <- toStorageIOErrT(RandomAccessIO.open(storagePath))
+      blockMessageRandomAccessFile <- RandomAccessIO.open(storagePath).toStorageIOErrT
       sortedCheckpoints            <- loadCheckpoints(checkpointsDirPath)
       checkpointsMap               = sortedCheckpoints.map(c => c.index -> c).toMap
       currentIndex                 = sortedCheckpoints.lastOption.map(_.index + 1).getOrElse(0)
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -92,8 +92,8 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
     def readBlockMessageFromFile(storageFile: RandomAccessIO[F]): F[BlockMessage] =
       for {
         _                      <- storageFile.seek(indexEntry.offset)
-        blockMessageSize       <- storageFile.readInt
-        blockMessagesByteArray = Array.ofDim[Byte](blockMessageSize)
+        blockMessageSizeOpt    <- storageFile.readInt
+        blockMessagesByteArray = Array.ofDim[Byte](blockMessageSizeOpt.get)
         _                      <- storageFile.readFully(blockMessagesByteArray)
         blockMessage           = BlockMessage.parseFrom(blockMessagesByteArray)
       } yield blockMessage
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -4,22 +4,21 @@ import java.io._
 import java.nio.ByteBuffer
 import java.nio.file._
 
-import cats.Monad
-import cats.data.EitherT
+import cats.{Functor, Monad}
 import cats.effect.{Concurrent, ExitCase, Sync}
 import cats.implicits._
 import cats.effect.concurrent.Semaphore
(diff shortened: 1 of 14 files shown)


itegulov:
@ArturGajowy @rabbitonweb please review this PR. This is a duplicate of #2174: I had to create a new PR because #2174 was using #2160 as a base and changing it back to dev would result in breaking our CI (this hapenned to me in #2102 before). Since #2160 has been merged, I created this PR which is using dev as a base and will be able to pass CI checks. See our discussion with @ArturGajowy on Discord #block-store for more information regarding CI issues with non-dev bases.

This PR contains mostly the same changes as #2174 with one major difference being the usage of `FunctorRaise[F, IOError]` for handling IO errors instead of `EitherT[F, IOError, A]` which has simplified the code greatly. The default instance of `FunctorRaise[F, IOError]` is being derived from `Sync[F]` and using `.raiseError` to basically kill the node on every IO error. This behavious is customizable though as one could pass a meaningful instance of `FunctorRaise[F, IOError]` instead in case we will actually want to handle IO errors in the future.

I have also addressed  your comments: now using named arguments for DAG storage state, introduced a few more error types and now treating storage errors as fatal errors (no more logging and pretending everything is ok).

@rabbitonweb regarding using state by multiple threads: yes, this is supposed to be used by multiple threads. I agree with @ArturGajowy though and would like to use the weakest typeclass for achieving the desired functionality. Unless you have a reason to suggest using `Cell` instead.

itegulov:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -4,36 +4,41 @@ import java.io.{EOFException, IOException}
...
+        case e: IOException =>
+          RaiseIOError[F].raise[A](handleIoException(e))
+        case e: SecurityException =>
+          RaiseIOError[F].raise[A](FileSecurityViolation(e))
+        case e: UnsupportedOperationException =>
+          RaiseIOError[F].raise[A](UnsupportedFileOperation(e))
+        case e: IllegalArgumentException =>
+          RaiseIOError[F].raise[A](IllegalFileOperation(e))
+        case e =>
+          RaiseIOError[F].raise[A](UnexpectedIOError(e))
```
goral09: I think that this indirection actually makes the code less understandable. `RaiseIOError` is indeed `FunctorRaise[F, E]` but `E` is a subtype of `Exception` and concrete instance of this `FunctorRaise` is derived from the `Sync` itself. Is the only gain from using `FunctorRaise` (over `Sync`) is to use the least powerful abstraction?

cc @ArturGajowy

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -4,36 +4,41 @@ import java.io.{EOFException, IOException}
...
 import cats.implicits._
 import cats.effect.Sync
-import coop.rchain.blockstorage.util.io.IOError.{IOErr, IOErrT}
+import coop.rchain.blockstorage.util.io.IOError.RaiseIOError
 
 import scala.collection.JavaConverters._
 
 package object io {
-  private[io] def handleIo[F[_]: Sync, A](
+  private[io] def handleIo[F[_]: Sync: RaiseIOError, A](
```
goral09: Since `RaiseIOError` delegates to `Sync` anyway you should be able to have just `Sync` requirement and materialize `RaiseIOError`.

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
itegulov: `E` is not a subtype of `Exception` as `IOError` is not a subtype of `Exception`. Concrete instance of `RaiseIOError` is indeed being derived from `Sync` inside of `FileLMDBIndexBlockStore` constructor which wraps `IOError`s into a throwable class and uses `Sync[F].raiseError` on them. The main gain here is that if we decide to actually care about these IO errors (in other words, handle them in some meaningful way without crashing the node) we could just remove the automatic derivation from `Sync` and replace it with our own implementation of `FunctorRaise[F, E]`.

itegulov:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
goral09: My commit is irrelevant in the light of @itegulov answer to previous one.

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
goral09: Would this work with `monix.Task` though? Since it fixes any E to be subtype of `Throwable`?

goral09: Would this work with `monix.Task` though? Since it fixes any E to be subtype of `Throwable`?

goral09: > if we decide to actually care about these IO errors (in other words, handle them in some meaningful way without crashing the node)

you can still do that with current implementation (or any `Sync`).

ArturGajowy: > Is the only gain from using `FunctorRaise` (over `Sync`) is to use the least powerful abstraction?

@goral09 are you suggesting this is not a sufficient reason? :smiling_imp:

> Would this work with `monix.Task` though?

Task has Sync, so yes.

> you can still do that with current implementation (or any `Sync`).

You're right. I think though @itegulov meant having some guidance from the typesystem.

goral09: > > Would this work with `monix.Task` though?
> 
> Task has Sync, so yes.

That's not true. I asked whether using different error type (one that doesn't inherit from `Throwable`) would work. `Sync` if fixed for subtypes of `Throwable`.

> 
> > you can still do that with current implementation (or any `Sync`).
> 
> You're right. I think though @itegulov meant having some guidance from the typesystem.

I don't think this is true either. The only way to handle errors raised by the means of `FunctorRaise` is to recover from them using `handleError`, `handleErrorWith` etc. If you change the error to something that doesn't crash the node (as I understand sth that doesn't inherit from `THrowable`) you still have to handle errors the same way. So there's no gain.

ArturGajowy: I thought you were asking if we can have `FunctorRaise[Task, IOError]`. Yes, we can. See the instance we provide based for any `F[_] : Sync`

Sure, with the errors being transferred through Sync.raise, we have no type guidance on where errors of which type can occur. It still makes sense to use `FunctorRaise` at the 'throw site', because we can then run with e.g. `F = EitehrT[IoError, Coeval]` or `F = zio.BIO` or whatever we want - and we need no changes to the 'throw site'.

As a general remark, please also mind that if a given solution doesn't meet your arbitrary ad-hoc made success criterion, it doesn't automatically mean there's no gain from it.

goral09: > As a general remark, please also mind that if a given solution doesn't meet your arbitrary ad-hoc made success criterion, it doesn't automatically mean there's no gain from it.

I didn't say anywhere it doesn't meet my success criterion. I just expressed my opinion that in the case when there's a `Sync[F]` and `FunctorRaise[F, E]` available (and the fact the `F` will inevitably become a `monix.Task`) this gives less than it would have if `Sync` was bifunctor (or how should I call it) and `E` wasn't fixed for `_ <: Throwable`.

goral09: As discussed in different comment, I think that using Defer + Applicative would partially solve this problem.

rabbitonweb:
> agree with @ArturGajowy though and would like to use the weakest typeclass for achieving the desired functionality. Unless you have a reason to suggest using `Cell` instead.

`Cell` is weaker then `MonadState`.

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -40,37 +67,50 @@ package object io {
...
-      files        <- EitherT(listInDirectory(dirPath))
-      regularFiles <- files.filterA[IOErrTF](f => EitherT(isRegularFile(f)))
-    } yield regularFiles).value
-  }
+    } yield filesList
+
+  def listRegularFiles[F[_]: Sync: RaiseIOError](dirPath: Path): F[List[Path]] =
+    for {
+      files        <- listInDirectory(dirPath)
+      regularFiles <- files.filterA[F](isRegularFile[F])
```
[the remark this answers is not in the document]
rabbitonweb: not sure about this one Artur, i honestly don't understand the rules but compiler used to freak out if `F` not explitictly provided

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -113,9 +113,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: Log] private (
                                     case Some(checkpoint) =>
                                       type StorageIOErrTF[A] = StorageIOErrT[F, A]
                                       Sync[StorageIOErrTF].bracket {
-                                        toStorageIOErrT(
-                                          RandomAccessIO.open[F](checkpoint.storagePath)
-                                        )
+                                        RandomAccessIO.open[F](checkpoint.storagePath)
```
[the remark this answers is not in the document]
rabbitonweb: I think this is an squash with wrong message commited

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -4,34 +4,61 @@ import java.io.IOException
...
-        case e                                => Left(UnexpectedIOError(e))
+        case e: IOException =>
+          RaiseIOError[F].raise[A](handleIoException(e))
+        case e: SecurityException =>
+          RaiseIOError[F].raise[A](FileSecurityViolation(e))
+        case e: UnsupportedOperationException =>
+          RaiseIOError[F].raise[A](UnsupportedFileOperation(e))
+        case e: IllegalArgumentException =>
+          RaiseIOError[F].raise[A](IllegalFileOperation(e))
+        case NonFatal(e) =>
```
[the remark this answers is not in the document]
rabbitonweb: laughing on your own jokes @ArturGajowy ? :)

rabbitonweb:
We want to get this PR merged. Here I've divided not answered comments into two sections: things we want to do before merge and things that can be done in later PRs and they  are not blocking usage file based storages:

Fixes needed before merge

- duplicates: handleIO and handleIoF functions
- Please fix or comment on the 'bad merge?' issues (see below)
- First commit "Fix RandomAccessIO opening compilation errors" - is it a wrong commit message (while squashing) or sth else?
- "This method should not be creating dirs." https://github.com/rchain/rchain/pull/2186#discussion_r257063077
- Fix DAG storage instantiation for HashSetCasperTestNode


Things that can be done later not blocking this merge:
- Let's just Sync[F].defer then https://github.com/rchain/rchain/pull/2186#discussion_r257066748
- Shouldn't that return an empty array instead, and then the Option in the return type disappears? https://github.com/rchain/rchain/pull/2186#discussion_r257067768
- the `Use named constructor args to create DAG storage state` commit should've been incorporated into the original "use a single MonadState instead of bare refs" commit. https://github.com/rchain/rchain/pull/2186#discussion_r257071618

rabbitonweb:
Superseded by https://github.com/rchain/rchain/pull/2198

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/RandomAccessIO.scala
```diff
@@ -1,44 +1,57 @@
...
+    handleIoF(file.readInt().some, {
+      case _: EOFException => none[Int].pure[F]
+      case e               => RaiseIOError[F].raise[Option[Int]](IntReadFailed(e))
+    })
 
-  def readFully(buffer: Array[Byte]): F[IOErr[Unit]] =
-    handleIo(file.readFully(buffer), ByteArrayReadFailed.apply)
+  def readFully(buffer: Array[Byte]): F[Option[Unit]] =
+    handleIoF(file.readFully(buffer).some, {
+      case _: EOFException => none[Unit].pure[F]
```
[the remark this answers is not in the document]
itegulov: No, I am using Option here to distinguish EOF (so I can't stop reading). I can't return an empty array of size N (i.e. filled with zeros) because it will be indistingushiable from actually reading N zeros.

block-storage/src/main/scala/coop/rchain/blockstorage/util/io/package.scala
```diff
@@ -4,34 +4,61 @@ import java.io.IOException
...
+          RaiseIOError[F].raise[A](UnexpectedIOError(e))
       }
-    }
+    }.flatten
```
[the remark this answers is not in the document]
itegulov: I have replaced it with Sync[F].defer for now but weakaning it does indeed make sense.

PR #2191, opened 2019-02-15
[Rchain-3048] implement RevAddress

odeac:
## Overview
Implements REV address generator & parser

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3048

### Notes

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,7 +11,6 @@ object Dependencies {
   val catsMtlVersion = "0.4.0"
 
   // format: off
-  val bitcoinjCore        = "org.bitcoinj"                % "bitcoinj-core"             % "0.14.6"
   val bouncyCastle        = "org.bouncycastle"            % "bcprov-jdk15on"            % "1.60"
   val catsCore            = "org.typelevel"              %% "cats-core"                 % catsVersion
   val catsLawsTest        = "org.typelevel"              %% "cats-laws"                 % catsVersion % "test"
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
@@ -1,0 +1,65 @@
+package coop.rchain.rholang.interpreter.util
+
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+final case class Address(val prefix: Array[Byte], val keyHash: Array[Byte])
+
+class AddressTools(prefix: Array[Byte], keyLength: Int, checksumLength: Int) {
+  private def computeChecksum(toCheck: Array[Byte]): Array[Byte] =
+    Blake2b256.hash(toCheck).take(checksumLength)
+
+  /**
+    * Creates an Address given a public key.
+    *
+    * @param pk the public key from which the address is derived
+    * @return None if the key length is invalid or Some if the address was created successfully
+    */
+  def fromPublicKey(pk: Array[Byte]): Option[String] =
+    if (keyLength == pk.length) {
+      val keyHash = Blake2b256.hash(pk)
+      val payload = prefix ++ keyHash
+      val address = payload ++ computeChecksum(payload)
+
+      Some(Base58.encode(address))
+    } else None
+
+  def isValid(address: String): Boolean = parse(address).isRight
+
+  private val checksumStart = prefix.length + Blake2b256.hashLength
+  private val addressLength = prefix.length + Blake2b256.hashLength + checksumLength
+
+  def parse(address: String): Either[String, Address] = {
+    def validateLength(bytes: Array[Byte]) =
+      if (bytes.length == addressLength)
+        Right(())
+      else
+        Left("Invalid address length")
+
+    def validateChecksum(bytes: Array[Byte]) = {
+      val (payload, checksum) = bytes.splitAt(checksumStart)
+      val computedChecksum    = computeChecksum(payload)
+
+      if (computedChecksum.deep == checksum.deep)
+        Right(payload)
+      else
+        Left("Invalid checksum")
+    }
+
+    def parsePayload(payload: Array[Byte]) = {
+      val (actualPrefix, keyHash) = payload.splitAt(prefix.length)
+
+      if (actualPrefix.deep == prefix.deep) Right(Address(actualPrefix, keyHash))
+      else Left("invalid prefix")
+    }
+
+    for {
+      decodedAddress <- Base58.decode(address).toRight("Invalid Base58 encoding")
+      _              <- validateLength(decodedAddress)
+      payload        <- validateChecksum(decodedAddress)
+      address        <- parsePayload(payload)
+    } yield address
+  }
+}
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
@@ -1,0 +1,18 @@
+package coop.rchain.rholang.interpreter.util
+import coop.rchain.crypto.codec.Base16
+
+final case class RevAddress(val keyHash: Array[Byte])
+
+object RevAddress {
+  private val coinId  = "000000"
+  private val version = "00"
+  private val prefix  = Base16.decode(coinId + version)
+
+  private val tools = new AddressTools(prefix, keyLength = 32, checksumLength = 4)
+
+  def fromPublicKey(pk: Array[Byte]): Option[String] = tools.fromPublicKey(pk)
+
+  def parse(address: String) = tools.parse(address).map(a => RevAddress(a.keyHash))
+
+  def isValid(address: String) = parse(address).isRight
+}
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/codec/Base58.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/util/codec/Base58.scala
@@ -1,0 +1,88 @@
+/**
+  * The MIT License (MIT)
+  *
+  * Copyright (c) 2016 Ratio Club, Inc.
+  *
+  * Permission is hereby granted, free of charge, to any person obtaining a copy
+  * of this software and associated documentation files (the "Software"), to deal
+  * in the Software without restriction, including without limitation the rights
+  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+  * copies of the Software, and to permit persons to whom the Software is
+  * furnished to do so, subject to the following conditions:
+  *
+  * The above copyright notice and this permission notice shall be included in all
+  * copies or substantial portions of the Software.
+  *
+  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+  * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+  * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+  * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+  * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+  * SOFTWARE.
+  *
+  * The code below was copied from https://github.com/multiformats/scala-multihash/blob/master/src/main/scala/io/mediachain/multihash/Base58.scala
+  * and some refactoring was done and decode return type was changed to Option[Array[Byte]]
+  */
+package coop.rchain.rholang.interpreter.util.codec
+import java.math.BigInteger
+
+import scala.annotation.tailrec
+
+object Base58 {
+  val alphabet = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
+  val maxDigit = BigInteger.valueOf(58L)
+  // char -> value
+  private val digitMap = alphabet.zipWithIndex.toMap
+
+  /**
+    *
+    * @param input binary data
+    * @return the base-58 representation of input
+    */
+  def encode(input: Seq[Byte]): String =
+    if (input.isEmpty) ""
+    else {
+      val big     = new BigInteger(1, input.toArray)
+      val builder = new StringBuilder
+
+      @tailrec
+      def go(current: BigInteger): Unit = current match {
+        case BigInteger.ZERO => ()
+        case _ =>
+          val Array(x, remainder) = current.divideAndRemainder(maxDigit)
+          val _                   = builder.append(alphabet.charAt(remainder.intValue))
+          go(x)
+      }
+      go(big)
+      input.takeWhile(_ == 0).foreach(_ => builder.append(alphabet.charAt(0)))
+      builder.toString().reverse
+    }
+
+  /**
+    *
+    * @param input base-58 encoded data
+    * @return the decoded data
+    */
+  def decode(input: String): Option[Array[Byte]] = {
+    val (zeroChars, trim) = input.toStream.span(_ == alphabet(0))
+    val zeroes            = zeroChars.map(_ => 0: Byte).toArray
+
+    if (trim.isEmpty)
+      Some(zeroes)
+    else {
+      def getDigit(c: Char) =
+        digitMap.get(c).map(x => BigInteger.valueOf(x.toLong))
+
+      def accumulate(aOpt: Option[BigInteger], ch: Char) =
+        for {
+          a     <- aOpt
+          digit <- getDigit(ch)
+        } yield a.multiply(maxDigit).add(digit)
+
+      trim
+        .foldLeft(Option(BigInteger.ZERO))(accumulate)
+        .map(decoded => zeroes ++ decoded.toByteArray.dropWhile(_ == 0))
+    }
+  }
+}
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/util/AddressToolsSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/util/AddressToolsSpec.scala
@@ -1,0 +1,78 @@
+package coop.rchain.rholang.interpreter.util
+import coop.rchain.crypto.hash.Blake2b256
+import org.scalacheck._
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{FlatSpec, Matchers, PropSpec}
+import cats.implicits._
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+class AddressToolsSpec extends PropSpec with GeneratorDrivenPropertyChecks with Matchers {
+  implicit val propertyCheckConfiguration =
+    PropertyCheckConfiguration(minSuccessful = 10000, sizeRange = 200)
+
+  def genByteArrayOf(n: Int) = Gen.listOfN(n, Arbitrary.arbByte.arbitrary).map(_.toArray)
+
+  val genArgs =
+    for {
+      prefixLength   <- Gen.posNum[Int]
+      keyLength      <- Gen.posNum[Int]
+      checksumLength <- Gen.choose(0, Blake2b256.hashLength)
+      prefix         <- genByteArrayOf(prefixLength)
+      publicKey      <- genByteArrayOf(keyLength)
+    } yield (prefix, keyLength, checksumLength, publicKey)
+
+  property("parse after fromPublicKey works correctly") {
+
+    forAll(genArgs) {
+      case (prefix, keyLength, checksumLength, pk) =>
+        val tools = new AddressTools(prefix, keyLength, checksumLength)
+
+        val address = tools.fromPublicKey(pk).get
+
+        val parsedAddress = tools.parse(address).right.get
+
+        parsedAddress.prefix should be(prefix.deep)
+        parsedAddress.keyHash should be(Blake2b256.hash(pk))
+    }
+  }
+
+  property("replacing a character in an address makes the parsing fail") {
+    val generator =
+      for {
+        args <- genArgs
+
+        (prefix, keyLength, checksumLength, pk) = args
+
+        tools = new AddressTools(prefix, keyLength, checksumLength)
+
+        address = tools.fromPublicKey(pk).get
+
+        idx <- Gen.choose(0, address.length - 1)
+
+        badChar <- Arbitrary.arbChar.arbitrary.filter(c => address(idx) != c)
+
+      } yield (tools, address, idx, badChar)
+
+    forAll(generator) {
+      case (tools, address, idx, newChar) =>
+        val badAddress = address.substring(0, idx) + newChar + address.substring(idx)
+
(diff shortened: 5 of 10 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
```diff
@@ -1,3 +1,17 @@
 package coop.rchain.rholang.interpreter.util
 
-object Address {}
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+object Address {
+  def fromPublicKey(prefix : String, pk : Array[Byte]) : String = {
```
JosDenmark: What is the purpose of the prefix?

JosDenmark:
Is this address scheme from another project? If so, would you provide a reference link?

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
odeac: the prefix is a human-readable component of the address and its purpose is to be obvious what coin the address is referring to.

For example we can have REVAddress which would use the prefix "rev" or "testnet-rev" or maybe "rev:" ... that's something to decide.

The important things are that the prefix is ASCII so it's human readable, that the tools can do an initial validation of the address and that the checksum validates the prefix too so one could not copy a "revtest" address and use it to transfer "rev" to it.

odeac:
> Is this address scheme from another project? If so, would you provide a reference link?

Reference: https://www.oreilly.com/library/view/mastering-bitcoin-2nd/9781491954379/ch04.html#base58

This scheme is similar to what Bitcoin is doing, except that they Base58 encode the "version" (which is the same as our prefix). The good parts in this scheme is that they have a version (i.e. prefix) and a checksum. The bad part is that encoding the whole version makes it difficult to read by humans because you have to know that 1 is bitcoin address, m or n is bitcoin testnet address etc. 

With the proposed approach, the prefix is ASCII and not Base58 encoded so it stays readable. Also, the prefix is included in the checksum computation so it cannot be separated by mistake from the address payload.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
```diff
@@ -0,0 +1,17 @@
+package coop.rchain.rholang.interpreter.util
+
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+object Address {
```
ArturGajowy: We also should have a 
`def parse(string: String): Option[Address]` that would:
1. check the `checksum`
2. take the last N bytes before the checksum, where N is our allowed key length, call it `pkHash`
3. treat the rest of the bytes on the beginning as the `prefix`
4. return `Address(prefix, pkHash, checksum)`

ArturGajowy: We should have tests showing that `fromPublicKey` only produces `parse`-able addreses, and that various invalid input strings are rejected by `parse`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
ArturGajowy: We should consider constraining the prefix length (max 16 chars?).

We also should also consider restricting it to alphanumeric characters only (so no `:` or `-` or similar) so that it's easier to validate the whole address string. If we allow anything more for the prefix than we allow for the suffix, there are two different alphabets to be checked against, and the prefix length is determined by subtracting the suffix length from total length, and the valid suffix lengths can (?) be many... I think you get my point that this starts to become complex if we allow more than base58 for prefix.

@cboscolo what are your thoughts? (re: prefix length and prefix alphabet)

project/Dependencies.scala
```diff
@@ -11,7 +11,6 @@ object Dependencies {
   val catsMtlVersion = "0.4.0"
 
   // format: off
-  val bitcoinjCore        = "org.bitcoinj"                % "bitcoinj-core"             % "0.14.6"
```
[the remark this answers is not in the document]
odeac: I did a search and I couldn't find any usages. We'll see if the code builds in CI without this line.

As for using Base58 encoder from bitcoinj, when I added it to the project I got conflicts regarding the protobuf versions.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
```diff
@@ -1,3 +1,17 @@
 package coop.rchain.rholang.interpreter.util
 
-object Address {}
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+object Address {
```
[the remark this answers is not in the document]
odeac: will do

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
```diff
@@ -1,3 +1,17 @@
...
+  def fromPublicKey(prefix : String, pk : Array[Byte]) : String = {
+    val hash = Blake2b256.hash(pk)
+    val payload = Base58.encode(hash)
+    val toCheck = prefix + payload
+    val checksum = Blake2b256.hash(toCheck.getBytes(StandardCharsets.US_ASCII)).take(4)
```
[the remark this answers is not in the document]
odeac: bitcoin uses the first 4 bytes so I guess "do what bitcoin does" should work well here

reference: https://www.oreilly.com/library/view/mastering-bitcoin-2nd/9781491954379/ch04.html#base58

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/codec/Base58Spec.scala
```diff
@@ -0,0 +1,34 @@
...
+    Base58.encode(input) should equal("")
+  }
+
+  val examples =
+    Table(
+      ("plain", "encoded"),
+      (0, "1"),
+      (58, "21"),
+      (58*58, "211"),
+      (58*58*58, "2111")
```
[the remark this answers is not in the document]
odeac: I only tested a few simple cases with unit tests because I wrote generative tests that produce random Base58 encodings.

On the other hand you have a valid point that we should test for order or digits in the Base58 alphabet. I'll write a test for that

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/codec/Base58Properties.scala
```diff
@@ -0,0 +1,32 @@
+package coop.rchain.rholang.interpreter.util.codec
+
+import org.scalacheck.Gen
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{Matchers, PropSpec}
+
+class Base58Properties extends PropSpec with GeneratorDrivenPropertyChecks with Matchers {
```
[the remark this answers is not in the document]
odeac: I wanted to put them in one single test class but I had compilation errors when I had PropSpec mixed with FlatSpec

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
odeac: > We also should have a `def parse(string: String): Option[Address]` that would:
Agreed.

I'm currently working on a parse function with the signature `String => Either[String, Address]` that produces proper error messages depending on the

odeac: > We should have tests showing that `fromPublicKey` only produces `parse`-able addreses, and that various invalid input strings are rejected by `parse`.
Aren't these generative tests from Base58Properties enough?

property("decode after encode returns the original bytes") 
property("decode a string should fail if it contains invalid Base58 chars")

project/Dependencies.scala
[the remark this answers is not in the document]
goral09: > As for using Base58 encoder from bitcoinj, when I added it to the project I got conflicts regarding the protobuf versions.

You can `exclude` protobuf package from `bitcojnCore`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,63 @@
+package coop.rchain.rholang.interpreter.util
+
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+sealed case class Address(val prefix : String, val payload : Array[Byte])
```
tymm: Why did you make it a `sealed` case class? I can't see you extending `Address` anywhere in `AddressTools.scala`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,63 @@
...
+
+class AddressTools(val prefix : String) {
```
tymm: Some methods in this class miss a return type.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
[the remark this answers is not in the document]
cboscolo: There is no requirement to CRC error correction. (Indeed we don't want to do this.)

project/Dependencies.scala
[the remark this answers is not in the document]
cboscolo: I would strongly prefer this.  Makes me nervous to code our own...

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
cboscolo: We did discuss this. But, as I ponder this thread, I think we should take a slightly more straight-forward approach that more closely follows the bitcoin approach.
To simplify decoding, I think the base58 step should only happen ONCE at the end of encoding, and the first step of decoding. (No concatenating base58 output.)
We should use a 4-byte prefix that consists of a 3-byte magic number and a 1-byte REV address version.
The 3-byte magic number should be chosen to result in an eye-pleasing base58 output.
I will update the Wiki

odeac: The 3-byte magic number will be difficult to choose. 

See the conversions below:

hex                   Base58
================
028C7B      => rev
028C7B00 => 4nALw
028C7BF0 => 4nAR5
028C7BFF => 4nARL

For this reason I think the ASCII concatenation is better

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/codec/Base58Spec.scala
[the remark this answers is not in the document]
odeac: I added an extra test including all the base58 digits

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
```diff
@@ -0,0 +1,3 @@
+package coop.rchain.rholang.interpreter.util
+
+object Address {}
```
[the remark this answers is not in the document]
odeac: done

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
odeac: I misunderstood your comment. I'll address them by adding Address tests

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
odeac: will fix

project/Dependencies.scala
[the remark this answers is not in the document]
odeac: Eventually it worked by excluding a few packages. Thanks @goral09

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
odeac: I meant `final`

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/Address.scala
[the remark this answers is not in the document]
odeac: done

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,61 @@
...
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+final case class Address(val prefix : Array[Byte], val heyHash : Array[Byte])
+
+class AddressTools(prefix : Array[Byte], keyLength : Int, checksumLength : Int) {
+  private def computeChecksum(toCheck : Array[Byte]): Array[Byte] =
+    Blake2b256.hash(toCheck).take(checksumLength)
+
+  def fromPublicKey(pk : Array[Byte]) : String = {
```
ArturGajowy: this (and upstream usage in RevAddress) should return via Option / Either whenever `pk.length != keyLength`. We should not be `throw`-ing assertion errors.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,61 @@
+package coop.rchain.rholang.interpreter.util
+
+import java.nio.charset.StandardCharsets
+
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+final case class Address(val prefix : Array[Byte], val heyHash : Array[Byte])
```
ArturGajowy: we need to fix this typo in `heyHash` :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/AddressToolsSpec.scala
```diff
@@ -0,0 +1,59 @@
+package coop.rchain.rholang.interpreter.util
+import coop.rchain.crypto.hash.Blake2b256
+import org.scalacheck._
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{FlatSpec, Matchers, PropSpec}
+import cats.implicits._
+import coop.rchain.rholang.interpreter.util.codec.Base58
+
+class AddressToolsSpec extends PropSpec with GeneratorDrivenPropertyChecks with Matchers {
```
ArturGajowy: we should probably bump the property check config to do more trials of bigger size

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/codec/Base58Properties.scala
```diff
@@ -0,0 +1,58 @@
+package coop.rchain.rholang.interpreter.util.codec
+
+import org.scalacheck.{Arbitrary, Gen}
+import org.scalatest.prop.GeneratorDrivenPropertyChecks
+import org.scalatest.{Matchers, PropSpec}
+
+class Base58Properties extends PropSpec with GeneratorDrivenPropertyChecks with Matchers {
```
ArturGajowy: let's have the property config bumped to more trials

ArturGajowy: I managed to get:
```

NumberFormatException was thrown during property evaluation.
  Message: Zero length BigInteger
  Occurred when passed generated values (
    arg0 = (,0, ) // 9 shrinks
  )
ScalaTestFailureLocation: coop.rchain.rholang.interpreter.util.codec.Base58Properties at (Base58Properties.scala:44)
org.scalatest.exceptions.GeneratorDrivenPropertyCheckFailedException: NumberFormatException was thrown during property evaluation.
  Message: Zero length BigInteger
  Occurred when passed generated values (
    arg0 = (,0, ) // 9 shrinks
  )
```

by bumping the settings to

```scala
  implicit val propertyCheckConfiguration = PropertyCheckConfiguration(minSuccessful = 10000, sizeRange = 200)
```

ArturGajowy: It's reproducible AFAICT

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
```diff
@@ -0,0 +1,18 @@
...
+import coop.rchain.crypto.codec.Base16
+
+final case class RevAddress(val keyHash : Array[Byte])
+
+object RevAddress {
+  private val coinId = "000000"
+  private val version = "00"
+  private val prefix = Base16.decode(coinId + version)
+
+  private val tools = new AddressTools(prefix, keyLength = 32, checksumLength = 4)
```
ArturGajowy: I think the keyLength should be 2048 at least. We also probably shouldn't restrict it to one possible value - rather allow 2048, and the subsequent powers of 2. Let's just have 2048 for now though.

Notice this doesn't affect the address length.

ArturGajowy: I meant 2048 bits = 256 bytes, vs 256 bits = 32 bytes we have now.
But a quick google search suggests that Bitcoins uses 32-byte public keys?
I'm pretty sure we discussed 2048-bit keys. @cboscolo should we go full-bitcoin?

odeac: I had the impression we are using Ed25519 which has a public key length of 256bits

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
odeac: nice catch! :))

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
odeac: We have runtime errors modelled by exceptions or Option or Either and we have algorithm errors where you don't want to deal with the error but instead you assert and move forward.

This is a case of the latter because I expect the code calling the fromPublicKey to make sure that the key has the proper length.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
cboscolo: The RevAddress is the output of blake2b256(Ed25519_pub_key)

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
ArturGajowy: > I expect the code calling the fromPublicKey to make sure

That's the primary issue with non-fp/impure code: assuming the context in which it is going to be called (instead of defining the possible contexts using effects and the type system).

I disagree with you. The type, and just the type, should be enough to say if I can call a function or not. In no circumstance should it `throw` and actual exception. This is impure, and without a good reason.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
ArturGajowy: @odeac I've seen a similar identifier string, but am unsure. Can you point to a specific place in code?

ArturGajowy: I've missed @cboscolo's comment. I guess your combined input closes the case, Gentlemen :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/codec/Base58Properties.scala
odeac: fixed

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/AddressToolsSpec.scala
odeac: done

odeac:
> Please expose a `rho:rev:validateAddress` system process that:
> responds on the `ret` channel with the passed input if it's a valid rev address
> responds on the `ret` channel with `Nil` otherwise

The reason I postponed this part is that it's not clear for me how we expose these tools.

I don't like your proposal because the Nil doesn't contain any information about the failure. My suggestion is to have a validateAddress entrypoint which returns Nil if the validation succeeded or an error message. Then we could wrap this primitive in a Rholang contract that returns an [Either] (https://github.com/rchain/rchain/blob/dev/casper/src/main/rholang/Either.rho)

ArturGajowy:
> Also, I would do this in a separate PR as this one is big enough already.

That's fine, let's just start with it right away as this is needed for the wallet contracts. We can discuss the details of the API offline - ping me when you're ready to start that.

ArturGajowy:
Please fix the `assert` issue and reformat both commits - and we're good to go :heavy_check_mark:

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
odeac: fair enough. I've changed the code to return Option[String] instead

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/codec/Base58.scala
```diff
@@ -0,0 +1,88 @@
...
+      input.takeWhile(_ == 0).foreach(_ => builder.append(alphabet.charAt(0)))
+      builder.toString().reverse
+    }
+
+  /**
+    *
+    * @param input base-58 encoded data
+    * @return the decoded data
+    */
+  def decode(input: String): Option[Array[Byte]] = {
```
JosDenmark: Why not have encode's input and decodes output return the same collection type? Either both `Seq` or both `Array`?

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,65 @@
...
+        Right(payload)
+      else
+        Left("Invalid checksum")
+    }
+
+    def parsePayload(payload: Array[Byte]) = {
+      val (actualPrefix, keyHash) = payload.splitAt(prefix.length)
+
+      if (actualPrefix.deep == prefix.deep) Right(Address(actualPrefix, keyHash))
+      else Left("invalid prefix")
```
JosDenmark: I don't know if uniformity is as important to you as it is to me, but this "invalid" is lowercase 😄.

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -0,0 +1,65 @@
...
+  private def computeChecksum(toCheck: Array[Byte]): Array[Byte] =
+    Blake2b256.hash(toCheck).take(checksumLength)
+
+  /**
+    * Creates an Address given a public key.
+    *
+    * @param pk the public key from which the address is derived
+    * @return None if the key length is invalid or Some if the address was created successfully
+    */
+  def fromPublicKey(pk: Array[Byte]): Option[String] =
```
JosDenmark: If we use the existing `PublicKey` type for the input parameter, we get type-level enforcements for what can be used to construct addresses.

PR #2196, opened 2019-02-19
Use GCA instead of LFB in estimator

KentShikama:
## Overview

The issue for the invalid parents right now is that the last finalized block for each validator is different. And so it is possible that even though you know that the main chain will eventually settle on a certain block B, another validator might add a block that uses something before B as a parent.
So for the purposes of calculating GHOST, we don't need a last finalized block as much as a greatest common ancestor block of all validators. The GCA block could be either before or after the last finalized block on the main chain. The GCA block may surprisingly also not be in the main chain eventually.

TODO: Implement a GCA that doesn't require the genesis block (see https://rchain.atlassian.net/browse/RCHAIN-3002)
TODO: Remove usage of BlockStore by just using BlockDagRepresentation (see https://rchain.atlassian.net/browse/RCHAIN-3003)

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-2909

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Estimator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Estimator.scala
@@ -56,7 +56,7 @@ object Estimator {
       latestMessagesHashes: Map[Validator, BlockHash]
   ): F[BlockHash] =
     for {
-      latestMessages <- latestMessagesHashes.values.toList
+      latestMessages <- latestMessagesHashes.values.toStream
                          .traverse(hash => ProtoUtil.unsafeGetBlock[F](hash))
       result <- if (latestMessages.isEmpty) {
                  genesis.pure[F]
@@ -182,10 +182,17 @@ object Estimator {
       blockDag: BlockDagRepresentation[F],
       scores: Map[BlockHash, Long]
   ): F[List[BlockHash]] =
-    blockDag.children(b).map(_.getOrElse(Set.empty[BlockHash]).filter(scores.contains)).map {
-      case c if c.nonEmpty => c.toList
-      case _               => List(b)
-    }
+    for {
+      children <- blockDag
+                   .children(b)
+                   .map(maybeChildren => maybeChildren.getOrElse(Set.empty[BlockHash]))
+      scoredChildren = children.filter(scores.contains)
+      result = if (scoredChildren.nonEmpty) {
+        scoredChildren.toList
+      } else {
+        List(b)
+      }
+    } yield result
 
   private def stillSame(blocks: List[BlockHash], newBlocks: List[BlockHash]): Boolean =
     newBlocks == blocks
--- a/casper/src/main/scala/coop/rchain/casper/Estimator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Estimator.scala
@@ -7,7 +7,7 @@ import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.util.BlockMessageUtil
 import coop.rchain.blockstorage.{BlockDagRepresentation, BlockStore}
 import coop.rchain.casper.protocol.BlockMessage
-import coop.rchain.casper.util.DagOperations
+import coop.rchain.casper.util.{DagOperations, ProtoUtil}
 import coop.rchain.casper.util.ProtoUtil.weightFromValidatorByDag
 
 import scala.collection.immutable.{Map, Set}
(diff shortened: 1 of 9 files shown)


KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -21,37 +21,53 @@ object Estimator {
 
   def tips[F[_]: Monad: BlockStore](
       blockDag: BlockDagRepresentation[F],
-      lastFinalizedBlockHash: BlockHash
+      genesis: BlockMessage
```
goral09: Just an idea - we pass `genesis` block everywhere we have `BlockStore` dependency as well. Since genesis block never changes maybe it should become part of the `BlockStore` (or `BlockDagRepresentation`) in a form o a field:
```
trait BlockStore[F] {
   def genesis: F[BlockMessage]
   …
}
```
This way we could get rid of the `genesis` argument which appear in a lot of places.

KentShikama: We will follow this PR up to remove the genesis block from the estimator. But in any case, this is still applicable to anything that uses the BlockStore and requires the genesis block. Thanks for the suggestion.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -21,37 +21,53 @@ object Estimator {
...
       sortedChildren      = maybeSortedChildren.flatten.toVector
     } yield sortedChildren
 
+  private def calculateGca[F[_]: Monad: BlockStore](
+      blockDag: BlockDagRepresentation[F],
+      genesis: BlockMessage,
+      latestMessagesHashes: Map[Validator, BlockHash]
+  ): F[BlockHash] =
+    for {
+      latestMessages <- latestMessagesHashes.values.toList
```
odeac: toStream instead of toList will avoid the creation of an intermediary list

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -166,9 +182,10 @@ object Estimator {
       blockDag: BlockDagRepresentation[F],
       scores: Map[BlockHash, Long]
   ): F[List[BlockHash]] =
-    for {
-      c <- blockDag.children(b).map(_.getOrElse(Set.empty[BlockHash]).filter(scores.contains))
-    } yield if (c.nonEmpty) c.toList else List(b)
+    blockDag.children(b).map(_.getOrElse(Set.empty[BlockHash]).filter(scores.contains)).map {
```
odeac: First of all this line is unreadable. One idea to improve it quite easily is to extract a named function with the names of the insides of the maps.

Second, this could be a private function in sortChildren so you don't have to pass explicitely  F, blockDag and the scores.

Third, I find the original `if ... else ...` less verbose and more readable than the new `case ... if ... case...`.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
odeac: I might misunderstand the issue here but my feeling is that...

If the addBlock algorithm, or, in fact, any algorithm that we run on the blockchain, cares about genesis block it will never have the right time complexity. So, if we consider this as an intermediary step that fixes a current blocker issue then it's ok. Otherwise, except for its creation, we should never care about genesis as a special block.

For this reason I wouldn't put the genesis block as a dependency but instead I would work to remove the dependencies on genesis.

KentShikama: We always have to treat genesis specially because it is a block without any parents.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
KentShikama: I like seeing what is explicitly passed in. I'll make the function more readable though.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
KentShikama: Will do

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -182,10 +182,17 @@ object Estimator {
...
+      children <- blockDag
+                   .children(b)
+                   .map(maybeChildren => maybeChildren.getOrElse(Set.empty[BlockHash]))
+      scoredChildren = children.filter(scores.contains)
+      result = if (scoredChildren.nonEmpty) {
+        scoredChildren.toList
+      } else {
+        List(b)
+      }
+    } yield result
```
odeac: The map inside for doesn't help the readability. Neither does ```for {... result = computeResult(...)} yield result``` 

Something along these lines looks more readable
```scala 
blockDag
  .children(b)
  .map { maybeChildren =>
      val scoredChildren = 
                maybeChildren
                    .getOrElse(Set.empty[BlockHash]))
                    .filter(scores.contains)

       if (scoredChildren.nonEmpty) 
            scoredChildren.toList 
       else
            List(b) 
    }

PR #2198, opened 2019-02-19
Functor raise io errors

rabbitonweb:
This is just https://github.com/rchain/rchain/pull/2186#issuecomment-465105510 but reopened on my branch so we can merge it

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
@@ -773,17 +773,17 @@ object BlockDagFileStorage {
                                        true
                                      )
       state = BlockDagFileStorageState(
-        latestMessagesMap,
-        childMap,
-        dataLookupList.toMap,
-        topoSort,
-        sortedCheckpoints.lastOption.map(_.end).getOrElse(0L),
-        sortedCheckpoints,
-        latestMessagesLogOutputStream,
-        logSize,
-        calculatedLatestMessagesCrc,
-        blockMetadataLogOutputStream,
-        calculatedDataLookupCrc
+        latestMessages = latestMessagesMap,
+        childMap = childMap,
+        dataLookup = dataLookupList.toMap,
+        topoSort = topoSort,
+        sortOffset = sortedCheckpoints.lastOption.map(_.end).getOrElse(0L),
+        checkpoints = sortedCheckpoints,
+        latestMessagesLogOutputStream = latestMessagesLogOutputStream,
+        latestMessagesLogSize = logSize,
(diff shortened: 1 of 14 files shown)


ArturGajowy:
@rabbitonweb FYI there have been some compilation errors, please fix them :)

itegulov:
@rabbitonweb I have replaced your commit with my version: `handleIoF` shouldn't actually be deleted, it is actually needed there and by replacing it with `handleIo` you have got the compilation errors @ArturGajowy mentioned. I have just reimplemented it to reduce code duplication; I hope you don't mind.

rabbitonweb:
@itegulov this is awesome, i've pushed that in my evening and went to bed :) was not aware of compilation issues. This is now ready to be  reviewed and merged.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -267,144 +347,137 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: BlockStore] private
...
-      _              <- lock.release
+      latestMessages <- getLatestMessages
+      childMap       <- getChildMap
+      dataLookup     <- getDataLookup
+      topoSort       <- getTopoSort
+      sortOffset     <- getSortOffset
     } yield FileDagRepresentation(latestMessages, childMap, dataLookup, topoSort, sortOffset)
 
+  def getRepresentation: F[BlockDagRepresentation[F]] =
+    lock.withPermit(representation)
```
KentShikama: Is it necessary to obtain a lock if we are doing only read-only operations here?

KentShikama: Is it because each individual item could be updated and so we might get some old state mixed with new state?

itegulov: Yes, you could indeed end up grabbing a mixed state. It is possible to fix this since we have removed  all`Ref`s and replaced them with a single `MonadState`. So you could get a consistent state from `MonadState` and get all individual items from it. However, this change is not as simple as just rewriting `representation` method because all other state-modifying methods assume that they "own" this state during execution and partially update it like this:

```scala
...
  _ <- setDataLookup(Map.empty)
  _ <- setChildMap(Map.empty)
  _ <- setTopoSort(Vector.empty)
  _ <- setLatestMessages(Map.empty)
...
```

So we have to rewrite all these places to modify `MonadState` atomically. Which might actually make sense but probably deserves its own PR.

goral09:
Am I right that there's only 1 file used to store the blocks and offsets are taken from the LMDB and only one `RandomAccessFile` is used to seek in the file?

PR #2199, opened 2019-02-20
RCHAIN-2846 Implement test_catch_up_next_round

zsluedem:
## Overview
Implement test_catch_up_next_round

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2846

### Notes

### DIFF
--- a/integration-tests/test/test_dag_correctness.py
+++ b/integration-tests/test/test_dag_correctness.py
@@ -1,9 +1,28 @@
+from random import Random
 import pytest
 from docker.client import DockerClient
 
+from . import conftest
+
+from .rnode import (
+    bootstrap_connected_peer,
+    docker_network_with_started_bootstrap,
+)
+from .wait import (
+    wait_for_peers_count_at_least,
+    wait_for_blocks_count_at_least,
+)
 from .common import (
     CommandLineOptions,
 )
+
+from .test_multiple_deploys import DeployThread
+
+BOOTSTRAP_NODE_KEYS = conftest.KeyPair(private_key='ff2ba092524bafdbc85fa0c7eddb2b41c69bc9bf066a4711a8a16f749199e5be', public_key='e2d64cea324a5a890326fc64c302cae1829d34ff4a584ff6a18e5026ad21f31f')
+BONDED_VALIDATOR_KEY_1 = conftest.KeyPair(private_key='597623f0b50e82008d52644983699724538b4307efbe3d0b992c651ca7f860c8', public_key='e8efbecd4f7067ec163a16e0ca6cf9b8c832b9cc9d2d1cc0ca1a642ca630aefb')
+BONDED_VALIDATOR_KEY_2 = conftest.KeyPair(private_key='9a32ff7b7c6e25527e0b4e5bec70596c6094e6529d56bf61cbd1ca26d3e92b10', public_key='fc99ed51e788bd040d4685721c46aaa580de06190cb77e51e6e3e50d38b1c6a2')
+BONDED_VALIDATOR_KEY_3 = conftest.KeyPair(private_key='af47862137d4e772f540029ae73ee01443c61288f3df9307a13d681de6ad2de4', public_key='a063cbb51f5d8b7e5b5eb9be3f8222c61287c84e0fc410e35c71c277ed12b9db')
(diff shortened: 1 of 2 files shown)


zsluedem:
* bors try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

integration-tests/test/test_dag_correctness.py
```diff
@@ -1,19 +1,113 @@
...
+                            )
+
+                            validator1_vdag = validator1.get_vdag()
+                            validator2_vdag = validator2.get_vdag()
+                            validator3_vdag = validator3.get_vdag()
+                            validator4_vdag = validator4.get_vdag()
+
+                            assert validator1_vdag == validator4_vdag
+                            assert validator2_vdag == validator4_vdag
+                            assert validator3_vdag == validator4_vdag
```
KentShikama: If any blocks get orphaned, it is possible that these vdags will look different.

zsluedem: Do you have any recommended way to compare two dag?
I tried to use networkx to parse the string but the finalized block info would be lost.

KentShikama: show-blocks should give information about finalized blocks. If you could just check that the main chain's are the same that would be enough.

zsluedem: Wait a minute. 
If any blocks get orphaned, this test should fail as expected.
Should a dag with an orphaned block be considered the same as a dag without orphaned blocks?

BTW, how to create a dag with orphaned blocks? I can not produce this kind of test case in my own machine.

KentShikama: Yeah if you remove all the orphaned blocks from both dags, then they should be the same.

It is difficult to create an orphaning when you got good network connectivity.

KentShikama: Nvm, I realized my mistake.

If a validator A has proposed a block A1 followed by block A2 and A1 is orphaned, then the resulting dag for some validators may not include A1. However, if A2 is included in all validator's dag's, then since A1 is in the justification of A2 (as you always point to your own previous block in your justification), A1 will automatically be pulled in if A2 is included.
The only case in which A1 won't exist on another validator's dag is if A2 also is orphaned (along with all the blocks above it by validator A). Hence, the only time the dags will be different is if the network continues to be down forever for a specific validator - which I think you don't have to worry about in this test.

adaszko: I've expressed my opinion on how I see it should work at https://rchain.atlassian.net/browse/RCHAIN-2925
@rabbitonweb what's the progress on that ticket?

zsluedem: @adaszko My suggestion here is that we merge this pr first since this test case won't cause an orphaned branch or something. We can deal with the more complicated feature on another pr(ticket).

KellyatPyrofex:
Discussed during March 1 standup. This PR is blocked by https://rchain.atlassian.net/browse/RCHAIN-2925. Plan is for @rabbitonweb to address RChain-2925 after completion of prioritized blockstore work.

zsluedem:
Since https://github.com/rchain/rchain/pull/2355 is merged, I would take a look again at this pr

zsluedem:
bors try

KentShikama:
bors delegate+

integration-tests/test/test_dag_correctness.py
```diff
@@ -1,19 +1,113 @@
...
+                            wait_for_blocks_count_at_least(
+                                context,
+                                validator4,
+                                expected_blocks_count + 1,
+                            )
+
+                            validator1_vdag = validator1.get_vdag()
+                            validator2_vdag = validator2.get_vdag()
+                            validator3_vdag = validator3.get_vdag()
+                            validator4_vdag = validator4.get_vdag()
```
adaszko: I think this is supposed to use `mvdag` and not `vdag` command as per https://github.com/rchain/rchain/pull/2355 and this PR was merged prematurely.

zsluedem: Sorry. I would make a quick patch.

adaszko: 👍

PR #2200, opened 2019-02-20
Make last finalized block calculation lazier

KentShikama:
## Overview

We previously attempted to update the last finalized block every block but now we only do
so when we are asked what the last finalized block is. This will make the vdag command a
bit slower but make block processing a lot faster. We also move last finalized block calculation
functions to its own object.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3006

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
@@ -1,0 +1,74 @@
+package coop.rchain.casper
+
+import cats.Monad
+import cats.effect.{Concurrent, Sync}
+import cats.implicits._
+import coop.rchain.blockstorage.{BlockDagRepresentation, BlockDagStorage, BlockStore}
+import coop.rchain.casper.util.ProtoUtil
+import coop.rchain.shared.{Cell, Log}
+import coop.rchain.catscontrib.ListContrib
+
+object LastFinalizedBlockCalculator {
+  // TODO: Extract hardcoded fault tolerance threshold
+  private val faultToleranceThreshold = 0f
+
+  def run[F[_]: Monad: Log: Sync: Concurrent: BlockStore: BlockDagStorage: SafetyOracle](
+      dag: BlockDagRepresentation[F],
+      lastFinalizedBlockHash: BlockHash
+  )(implicit state: Cell[F, CasperState]): F[BlockHash] =
+    for {
+      maybeChildrenHashes <- dag.children(lastFinalizedBlockHash)
+      childrenHashes      = maybeChildrenHashes.getOrElse(Set.empty[BlockHash]).toList
+      maybeFinalizedChild <- ListContrib.findM(
+                              childrenHashes,
+                              (blockHash: BlockHash) =>
+                                isGreaterThanFaultToleranceThreshold(dag, blockHash)
+                            )
+      newFinalizedBlock <- maybeFinalizedChild match {
+                            case Some(finalizedChild) =>
+                              removeDeploysInFinalizedBlock[F](finalizedChild) *> run[F](
+                                dag,
+                                finalizedChild
+                              )
+                            case None => lastFinalizedBlockHash.pure[F]
+                          }
+    } yield newFinalizedBlock
+
+  private def removeDeploysInFinalizedBlock[F[_]: Monad: Log: Sync: Concurrent: BlockStore: BlockDagStorage: SafetyOracle](
+      finalizedChildHash: BlockHash
+  )(implicit state: Cell[F, CasperState]): F[Unit] =
+    for {
+      block              <- ProtoUtil.unsafeGetBlock[F](finalizedChildHash)
+      deploys            = block.body.get.deploys.map(_.deploy.get).toList
+      stateBefore        <- state.read
+      initialHistorySize = stateBefore.deployHistory.size
+      _ <- state.modify { s =>
+            s.copy(deployHistory = s.deployHistory -- deploys)
+          }
+      stateAfter     <- state.read
+      deploysRemoved = initialHistorySize - stateAfter.deployHistory.size
+      _ <- Log[F].info(
+            s"Removed $deploysRemoved deploys from deploy history as we finalized block ${PrettyPrinter
+              .buildString(finalizedChildHash)}."
+          )
+    } yield ()
+
+  /*
+   * On the first pass, block B is finalized if B's main parent block is finalized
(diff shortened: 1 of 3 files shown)


KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -478,7 +478,7 @@ object BlockAPI {
         tupleSpaceHash = PrettyPrinter.buildStringNoLimit(tsHash),
         tupleSpaceDump = tsDesc,
         timestamp = timestamp,
-        faultTolerance = normalizedFaultTolerance - initialFault,
+        faultTolerance = normalizedFaultTolerance - initialFault, // TODO: Fix
```
dzajkowski: Fix what? :)

KentShikama: It is supposed to be a plus sign but I'll need to type up why.

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
```diff
@@ -0,0 +1,74 @@
...
+import coop.rchain.blockstorage.{BlockDagRepresentation, BlockDagStorage, BlockStore}
+import coop.rchain.casper.util.ProtoUtil
+import coop.rchain.shared.{Cell, Log}
+import coop.rchain.catscontrib.ListContrib
+
+object LastFinalizedBlockCalculator {
+  // TODO: Extract hardcoded fault tolerance threshold
+  private val faultToleranceThreshold = 0f
+
+  def run[F[_]: Monad: Log: Sync: Concurrent: BlockStore: BlockDagStorage: SafetyOracle](
```
lukasz-golebiewski: Is this rather complex method tested anywhere (possibly on randomly generated input)?

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
```diff
@@ -0,0 +1,74 @@
...
+      stateAfter     <- state.read
+      deploysRemoved = initialHistorySize - stateAfter.deployHistory.size
+      _ <- Log[F].info(
+            s"Removed $deploysRemoved deploys from deploy history as we finalized block ${PrettyPrinter
+              .buildString(finalizedChildHash)}."
+          )
+    } yield ()
+
+  /*
+   * On the first pass, block B is finalized if B's main parent block is finalized
```
lukasz-golebiewski: This comment looks outdated. The method it applies to does not look like it's checking whether the main parent block is finalized

lukasz-golebiewski:
I cannot see where exactly the calculation is made lazier. I guess the one and only commit should be split into two parts. The first would contain only the code extraction to a new object and the second would contain the clue

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
KentShikama: LastFinalizedBlockCalculator.run is only recursively called if there was a block that was finalized.

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
KentShikama: Basically any call to https://github.com/rchain/rchain/pull/2200/files/2c47336efce2a6afe0cd6b56abf7617a642bec36#diff-de23b4200eaecb35ba261ac3caa4d112R202 tests this

For example,

https://github.com/rchain/rchain/blob/0115d5854bcd5e31402b920a4f820b45ebaf0f1f/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala#L1676

https://github.com/rchain/rchain/blob/0115d5854bcd5e31402b920a4f820b45ebaf0f1f/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala#L1687

https://github.com/rchain/rchain/blob/0115d5854bcd5e31402b920a4f820b45ebaf0f1f/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala#L1698

https://github.com/rchain/rchain/blob/0115d5854bcd5e31402b920a4f820b45ebaf0f1f/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala#L1708

How would a randomly generated block dag testing work here?

KentShikama:
> I cannot see where exactly the calculation is made lazier

Instead of being called with every call to addBlock, it is only called when lastFinalizedBlock is called

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
lukasz-golebiewski: You mean if the "main parent block" was finalized? Or any?

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
lukasz-golebiewski: For starters you could generate random DAGs for which it is trivial to determine which blocks should be finalized and check if the algorithm properly marks them as finalized

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
KentShikama: Main parent block

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
KentShikama: If it is a random DAG, then it isn't trivial to determine which blocks should be finalized - only the genesis block can always be trivially finalized.

lukasz-golebiewski: Ok

lukasz-golebiewski:
What about splitting the commit, so the clue doesn't have to be searched for through the noise?

KentShikama:
By clue do you mean the TODO: fix stuff?

lukasz-golebiewski:
I mean the essence. The place where the calculation is made lazier

KentShikama:
@lukasz-golebiewski Done; PTAL

KentShikama:
bors try

PR #2202, opened 2019-02-20
Implement grpc API responses as a custom Either message

sebtomba:
## Overview
Instead of signalling errors inside of response messages use the Either type. Use the grpc Either message in DeployGrpcService to send the API Either over the wire to the client. Use grpc Either helpers to reassemble the API Either response. Change DeployRuntime accordingly.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2928

### Notes
These changes are **not** backward compatible. Older rnode clients will be not able to communicate with the new API.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -1,60 +1,68 @@
 package coop.rchain.casper.api
 
+import scala.collection.immutable
+
 import cats.Monad
 import cats.effect.concurrent.Semaphore
-import cats.effect.Concurrent
-import cats.effect.Sync
+import cats.effect.{Concurrent, Sync}
 import cats.implicits._
-import com.google.protobuf.ByteString
+
 import coop.rchain.blockstorage.{BlockDagRepresentation, BlockStore}
+import coop.rchain.casper._
 import coop.rchain.casper.Estimator.BlockHash
-import coop.rchain.casper.MultiParentCasperRef.MultiParentCasperRef
 import coop.rchain.casper.MultiParentCasper.ignoreDoppelgangerCheck
+import coop.rchain.casper.MultiParentCasperRef.MultiParentCasperRef
 import coop.rchain.casper.protocol._
+import coop.rchain.casper.util.rholang.RuntimeManager
+import coop.rchain.casper.util.{EventConverter, ProtoUtil}
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.graphz._
 import coop.rchain.models.Par
-import coop.rchain.rspace.StableHashProvider
-import coop.rchain.rspace.trace.{COMM, Consume, Produce}
-import coop.rchain.shared.Log
-import coop.rchain.models.serialization.implicits.mkProtobufInstance
 import coop.rchain.models.rholang.sorter.Sortable._
+import coop.rchain.models.serialization.implicits.mkProtobufInstance
+import coop.rchain.rspace.StableHashProvider
+import coop.rchain.rspace.trace._
+import coop.rchain.shared.Log
 
-import scala.collection.immutable
-import coop.rchain.casper.util.EventConverter
-import coop.rchain.casper._
-import coop.rchain.casper.util.rholang.RuntimeManager
-import coop.rchain.casper.util.ProtoUtil
+import com.google.protobuf.ByteString
 
 object BlockAPI {
 
-  def deploy[F[_]: Monad: MultiParentCasperRef: Log](d: DeployData): F[DeployServiceResponse] = {
-    def casperDeploy(implicit casper: MultiParentCasper[F]): F[DeployServiceResponse] =
+  type Error           = String
+  type ApiErr[A]       = Either[Error, A]
+  type Effect[F[_], A] = F[ApiErr[A]]
+
+  def deploy[F[_]: Monad: MultiParentCasperRef: Log](
+      d: DeployData
+  ): Effect[F, DeployServiceResponse] = {
+    def casperDeploy(
+        implicit casper: MultiParentCasper[F]
+    ): Effect[F, DeployServiceResponse] =
       for {
         r <- MultiParentCasper[F].deploy(d)
         re <- r match {
-               case Right(_)  => DeployServiceResponse(success = true, "Success!").pure[F]
-               case Left(err) => DeployServiceResponse(success = false, err.getMessage).pure[F]
+               case Right(_)  => DeployServiceResponse("Success!").asRight.pure[F]
+               case Left(err) => err.getMessage.asLeft.pure[F]
              }
       } yield re
 
     val errorMessage = "Could not deploy, casper instance was not available yet."
 
     MultiParentCasperRef
-      .withCasper[F, DeployServiceResponse](
+      .withCasper[F, ApiErr[DeployServiceResponse]](
         casperDeploy(_),
         Log[F]
           .warn(errorMessage)
-          .as(DeployServiceResponse(success = false, s"Error: $errorMessage"))
+          .as(s"Error: $errorMessage".asLeft)
       )
   }
 
   def createBlock[F[_]: Sync: Concurrent: MultiParentCasperRef: Log](
       blockApiLock: Semaphore[F]
-  ): F[DeployServiceResponse] = {
+  ): Effect[F, DeployServiceResponse] = {
     val errorMessage = "Could not create block, casper instance was not available yet."
-    MultiParentCasperRef.withCasper[F, DeployServiceResponse](
+    MultiParentCasperRef.withCasper[F, ApiErr[DeployServiceResponse]](
       casper => {
         Sync[F].bracket(blockApiLock.tryAcquire) {
           case true =>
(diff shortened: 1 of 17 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -366,39 +371,35 @@ object BlockAPI {
...
         blockQueryResponse <- maybeBlock match {
                                case Some(block) =>
                                  for {
                                    blockInfo <- getFullBlockInfo[F](block)
-                                 } yield
-                                   BlockQueryResponse(
-                                     status = "Success",
-                                     blockInfo = Some(blockInfo)
-                                   )
+                                 } yield BlockQueryResponse(blockInfo = Some(blockInfo)).asRight
```
KentShikama: Is there a reason we are getting rid of status? I guess it isn't that useful in the first place.

sebtomba: @KentShikama That is the whole point of this PR. We want to use `Either` message to signal errors instead of a `status` field.

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -50,20 +54,26 @@ class CreateBlockAPITest extends FlatSpec with Matchers {
       "for(_ <- @1){ @2!(2) }"
     ).map(ProtoUtil.sourceDeploy(_, System.currentTimeMillis(), accounting.MAX_VALUE))
 
-    implicit val logEff = new LogStub[Effect]
-    def testProgram(blockApiLock: Semaphore[Effect])(
+    def createBlock(deploy: DeployData, blockApiLock: Semaphore[Effect])(
         implicit casperRef: MultiParentCasperRef[Effect]
-    ): Effect[(DeployServiceResponse, DeployServiceResponse)] = EitherT.liftF(
+    ): Effect[ApiErr[DeployServiceResponse]] =
```
tymm: Mhh, another type called `Effect`. In my opinion another reason to get rid of the type alias in `BlockAPI.scala` :)

models/src/main/protobuf/Either.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+import "scalapb/scalapb.proto";
+import "google/protobuf/any.proto";
+
+option java_package = "coop.rchain.either";
+
+message EitherAny {
```
dckc: How do I use this from other client languages? Is there cross-language support for `type_url` in protobuf SDKs?

models/src/main/scala/coop/rchain/models/either/EitherHelper.scala
```diff
@@ -0,0 +1,48 @@
...
+import coop.rchain.models.StacksafeMessage
+import coop.rchain.shared.ThrowableOps._
+
+import scalapb.GeneratedMessageCompanion
+
+object EitherHelper {
+  def toEither[A <: StacksafeMessage[A]](
+      msg: coop.rchain.either.Either
+  )(implicit cmp: GeneratedMessageCompanion[A]): Either[Seq[String], A] = {
+    require(msg.content.isDefined, "Either message has no content")
```
dckc: Does this `require` halt the node if it fails? Does that mean I can bring down a node by sending it a bad GRPC message?

models/src/main/scala/coop/rchain/models/either/EitherAnyHelper.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  def is[A <: StacksafeMessage[A]](
+      eitherAny: EitherAny
+  )(implicit cmp: GeneratedMessageCompanion[A]): Boolean =
+    typeNameFromTypeUrl(eitherAny.typeUrl) == cmp.scalaDescriptor.fullName
+
+  def unpack[A <: StacksafeMessage[A]](
+      eitherAny: EitherAny
+  )(implicit cmp: GeneratedMessageCompanion[A]): A = {
+    require(is[A](eitherAny), s"Type of the EitherAny message does not match the given class.")
```
dckc: Does this `require` halt the node if it fails? Does that mean I can bring down a node by sending it a bad GRPC message?

models/src/main/protobuf/Either.proto
sebtomba: @dckc This message is identical to [any.proto](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/any.proto)
It shouldn't be hard in other languages to process this type the same way a Google `Any` is handled. Please look at the grpc code in your favorite client language.

models/src/main/scala/coop/rchain/models/either/EitherAnyHelper.scala
sebtomba: @dckc I don't think so. It's implemented the same way in https://github.com/scalapb/ScalaPB/blob/master/scalapb-runtime/shared/src/main/scala/scalapb/AnyMethods.scala#L13
If you send a bad message to a grpc service then your request will fail. This is true for any kind of grpc service and message.

models/src/main/scala/coop/rchain/models/either/EitherHelper.scala
sebtomba: @dckc No, this is used on the client side. The client will fail if the service (node) replies with a bad grpc message.

sebtomba:
@dckc If the `node-client` is used anywhere and not just a POC (or a just for fun project) then please update it since you are the author and (I guess) the maintainer of this Python code.

KentShikama:
I was hoping to remove node-client from this repository but I don't have the permissions to do so. I remember seeing a jira ticket for it back in November - I guess no one who has the permissions thought it was priority enough.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -99,25 +102,27 @@ object DeployRuntime {
...
 
-  private def gracefulExit[F[_]: Monad: Sync, A](program: F[Either[Throwable, String]]): F[Unit] =
+  private def gracefulExit[F[_]: Monad: Sync, A](
+      program: F[Either[Seq[String], String]]
+  ): F[Unit] =
     for {
       result <- Sync[F].attempt(program)
-      _ <- result.joinRight match {
-            case Left(ex) =>
+      _ <- processError(result).joinRight match {
```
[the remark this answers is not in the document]
sebtomba: @tymm No, this is not correct. See the `attempt` above. If we have `Left[Throwable, Either[Seq[String], A]]` this means that the whole request failed (f.e. because the node is unavailable) and we have no response (no service errors and no result). On the other hand, if we have `Right[Throwable, Either[Seq[String], A]]` then the request succeeded (no `Throwable` inside) and we can reduce the response to just `Either[Seq[String], A]`. I hope that my explanation is clear. :)

BTW, the `Left` unification happens here:
```scala
private def processError[A](error: Either[Throwable, A]): Either[Seq[String], A] =
  error.leftMap(_.toMessageList())
```

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -72,9 +82,9 @@ class CreateBlockAPITest extends FlatSpec with Matchers {
       result       <- testProgram(blockApiLock)(casperRef)
     } yield result).value.unsafeRunSync.right.get
 
-    response1.success shouldBe true
-    response2.success shouldBe false
-    response2.message shouldBe "Error: There is another propose in progress."
+    response1.isRight shouldBe true
```
lukasz-golebiewski: When this test fails you will get a rather vague `false was not true` message. A more informative error would come from `response1 shouldBe a[Right]`

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -72,9 +82,9 @@ class CreateBlockAPITest extends FlatSpec with Matchers {
...
+    response2.isRight shouldBe false
+    response2.left.get shouldBe "Error: There is another propose in progress."
```
lukasz-golebiewski: Similarly here `response2` shouldBe Left("Error:...")`

models/src/main/scala/coop/rchain/models/either/EitherAnyHelper.scala
dckc: Docs for require say:
> Tests an expression, throwing an IllegalArgumentException if false. -- [predefs](https://www.scala-lang.org/api/current/scala/Predef$.html)

Where is the `IllegalArgumentException` caught? I don't see it. What does the caller get in this case?

KellyatPyrofex:
> I was hoping to move node-client from this repository to its own repo but I don't have the permissions to do so. I remember seeing a jira ticket for it back in November - I guess no one who has the permissions thought it was priority enough.

I think you recall 
https://rchain.atlassian.net/browse/RCHAIN-1233 when we brought node-client into rchain/rchain. Moving discussion of moving this to a new repo to discord.

models/src/main/protobuf/Either.proto
dckc: ok... I put this on my todo list as https://github.com/rchain-community/rchain-api/issues/60

And I opened https://github.com/rchain-community/rchain-grpc/issues/36 for the python guys from proof.

dckc:
> I was hoping to move node-client from this repository to its own repo ...

I suppose it's subsumed by https://github.com/rchain-community/rchain-grpc

Go ahead and remove it.

PR #2203, opened 2019-02-20
Purge CasperState

rabbitonweb:
## Overview
The idea is to remove as much as possible from `CasperState` because it can leak memory + it has to sync file based storages

- [x] Remove `seenBlockHashes` from `MultiParentCasperImpl`
- [x] `blockBuffer` needs to use `BlockHash` (**in progress**, @rabbitonweb )
- [x]  for `invalidBlockTracker` we will simply replace it with a `BlockMessage` field which would mark the block as invalid
- [x] `equivocationsTracker` needs to be restored when node reboots

~~mechanism to stop receiving deploys when `deployHistory` is bigger then `X` where `X` is configurable size~~ (see https://rchain.atlassian.net/browse/RCHAIN-3016)

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3009


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -30,7 +31,6 @@ import coop.rchain.shared._
   @param equivocationsTracker: Used to keep track of when other validators detect the equivocation consisting of the base block at the sequence number identified by the (validator, base equivocation sequence number) pair of each EquivocationRecord.
   */
 final case class CasperState(
-    seenBlockHashes: Set[BlockHash] = Set.empty[BlockHash],
```
itegulov: AFAIU from our discussion with @KentShikama, this field is not exactly equivalent to `dag.contains` because `seenBlockHashes` also includes blocks which were seen but were not added just yet (i.e. the blocks from `blockBuffer`). I believe that is why it was added in the first place. @lukasz-golebiewski @KentShikama please correct me if I am wrong. The tests didn't catch that the current implementation is buggy which makes me worry.

So the correct fix here would be to replace `seenBlockHashes` with a set of block hashes from `blockBuffer` which is king of what you are already doing in the second checkpoint.

rabbitonweb:
@itegulov regarding https://github.com/rchain/rchain/pull/2203#discussion_r258826005

You are absolutely right, I've rebased and fixed that commit. Now here are few remarks:

Fact that none of the tests failed is concerning to say the least. It might be that test does not exists, but from the look of the code, maybe that feature was not even used anymore (single-threadness of casper) - I've created an issue and assigned it to myself to handle it next week https://rchain.atlassian.net/browse/RCHAIN-3015

rabbitonweb:
@KentShikama please have a look at @itegulov's commits:

Introduce `invalid` field for blocks to replace invalidBlockTracker

1. Move non-persistent equivocationTracker into BlockDagStorage
2. Introduce `invalid` field for blocks to replace invalidBlockTracker

It seems different approach for that field we've talked yesterday, but you guys seemed also talk, so maybe that's the right way to go. I will have a look myself as well, but will appreciate your quick look at it

rabbitonweb:
@itegulov changes are now rebased, `fetch` & `fast-forward merge` before you move on :)

KentShikama:
My current thought is that only the invalidBlockTracker and the equivocationsTracker needs to be persisted. The others can be wiped out. I no longer think it is enough to keep boolean fields in the BlockMetadata to keep track of invalid blocks as that would require you to go through all blocks and filter out the ones that are invalid=True to repopulate the invalidBlockTracker Set. And that would be expensive to do - maybe it would be a good background task to validate that nothing was corrupted.

rabbitonweb:
@itegulov last update for today **but** important one. I've did yet **another** rebase to `"Remove seenBlockHashes from MultiParentCasperImpl"` commit. Reason for that was that casper test started failing. It took a moment to figure it out, but it turned out that there was coupling between file based `BlockStore` and filebased `BlockDagStorage`, that I had to remove. Please see `"Remove seenBlockHashes from MultiParentCasperImpl"` commit for details.

rabbitonweb:
@itegulov ending my shift, want to sum it up:
1. seems that `invalidBlockTracker` & `equivocationsTracker` should stay in `CasperState`, we just need to have mechanism to "restore" them from storages when node reboots, see @KentShikama  for details https://github.com/rchain/rchain/pull/2203#issuecomment-465998048
2. there seemed to be some incorrect coupeling between storages, I tried to fix that but maybe you should have a look as well? see https://github.com/rchain/rchain/pull/2203#issuecomment-466143467

See you in 12h :)

rabbitonweb:
@itegulov `BlockDagFileStorageTest` is failing, because i made neccessary changes, please have a look what failed

itegulov:
@KentShikama is there any reason to actually repopulate the `invalidBlockTracker` set? As you can see from changes from my commit (Introduce `invalid` field for blocks to replace invalidBlockTracker), we don't actually need the set itself and rather need a marker whether a certain block is invalid or not (i.e. the only way `invalidBlockTracker` is used right now is by calling `.contains` on it). The only drawback I see here is introducing a new field for `BlockMessage` which is being set after checking whether the block is valid or not (which I have just realised interferes with @rabbitonweb's work on moving `BlockStore[F].put` closer to receiving the block instead of putting it this after processing the block). If you think that serializing a set of blockhashes would be a better soultion I am fine with changing it.

KentShikama:
@itegulov Yes that is a good point. We can get rid of the invalidBlockTracker entirely.

itegulov:
@rabbitonweb I have finished my work on `equivocationsTracker` so we should be ready for a review once you finish item 2. As I already said you can mark the unpassing test from `BlockDagFileStorageTest` as ignored. However, you changes have also broken two tests in `HashSetCasperTest`:
1. should allow bonding in an existing network
2. should ask peers for blocks it is missing and add them

Please have a look at them (I would have looked at them by myself but I have only realized that they are broken just now after I ran tests locally after finishing `equivocationsTracker` work).

rabbitonweb:
This is ready for review!

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -246,7 +246,8 @@ message BlockMessage {
   bytes                  sig            = 7; // signature generated by signing `hash(hash(justification) concat blockHash)`.
   string                 sigAlgorithm   = 8; // name of the algorithm used to sign
   string                 shardId        = 9; // identifier of the shard where the block was created
-  bytes extraBytes = 10;
+  bool                   invalid        = 10; // whether the block was marked as invalid
```
KentShikama: I'm worried about having the invalid slot as part of the BlockMessage. The BlockMessage should be immutable and the blockHash should be the hash of all the contents. It should never be modified after creation, and no validator would ever produce a block with invalid = True from the beginning (that would be slashing themselves).

itegulov: I agree that having `invalid` as a part of BlockMessage is a bad idea. I have removed it from the BlockMessage proto class but left it in the BlockMetadata proto class. Now when you are inserting a block into DAG storage you have to specify whether the block is invalid or not (i.e. I have added a new `invalid: Boolean` parameter to `BLockDagStorage[F].insert` method).

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -64,32 +65,36 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
...
+  ): F[BlockStatus] = {
+
+    def logAlreadyProcessed =
+      Log[F]
+        .info(
+          s"Block ${PrettyPrinter.buildString(b.blockHash)} has already been processed by another thread."
+        )
+        .as(BlockStatus.processing)
+
+    def doopelgangerAndAdd =
```
tymm: Typo.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -194,10 +195,11 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
...
   ): F[Boolean] =
     for {
-      blockStoreContains <- BlockStore[F].contains(b.blockHash)
-      state              <- Cell[F, CasperState].read
-      bufferContains     = state.blockBuffer.exists(_.blockHash == b.blockHash)
-    } yield (blockStoreContains || bufferContains)
+      dag            <- blockDag
+      dagContains    <- dag.contains(b.blockHash)
+      state          <- Cell[F, CasperState].read
+      bufferContains = state.blockBuffer.exists(_ == b.blockHash)
```
tymm: You could replace this with `state.blockBuffer.contains(b.blockHash)`.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
rabbitonweb: 👍

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
rabbitonweb: 👍

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -20,22 +20,19 @@ import coop.rchain.comm.CommError.ErrorHandler
...
   @param deployHistory
-  @param invalidBlockTracker
-  @param equivocationsTracker: Used to keep track of when other validators detect the equivocation consisting of the base block at the sequence number identified by the (validator, base equivocation sequence number) pair of each EquivocationRecord.
   */
 final case class CasperState(
     blockBuffer: Set[BlockMessage] = Set.empty[BlockMessage],
     deployHistory: Set[DeployData] = Set.empty[DeployData],
-    dependencyDag: DoublyLinkedDag[BlockHash] = BlockDependencyDag.empty,
-    equivocationsTracker: Set[EquivocationRecord] = Set.empty[EquivocationRecord]
+    dependencyDag: DoublyLinkedDag[BlockHash] = BlockDependencyDag.empty
```
tymm: Comments above `CasperState` are out of sync.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -581,16 +581,26 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
       state          <- Cell[F, CasperState].read
       dependencyFree = state.dependencyDag.dependencyFree
       dependencyFreeBlocks = state.blockBuffer
-        .filter(block => dependencyFree.contains(block.blockHash))
+        .filter(blockHash => dependencyFree.contains(blockHash))
```
tymm: Could be replaced with `intersect(dependencyFree)`.

PR #2207, opened 2019-02-24
Rspace/rchain 3022/introduce f in i store

dzajkowski:
## Overview
Introduce F in IStore and align all the pieces of `produce` and `consume` methods to be suspended in F.

### JIRA ticket:
RCHAIN-3022

### Notes

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
@@ -41,6 +41,7 @@ import coop.rchain.casper.scalatestcontrib._
 import coop.rchain.casper.util.comm.TestNetwork
 import coop.rchain.catscontrib.ski.kp2
 import coop.rchain.comm.rp.Connect.Connections
+import coop.rchain.metrics
 import coop.rchain.metrics.Metrics
 import coop.rchain.shared.Log
 import org.scalatest
@@ -1806,10 +1807,11 @@ object HashSetCasperTest {
       faucetCode: String => String,
       deployTimestamp: Long
   ): BlockMessage = {
-    val initial           = Genesis.withoutContracts(bonds, 1L, deployTimestamp, "rchain")
-    val storageDirectory  = Files.createTempDirectory(s"hash-set-casper-test-genesis")
-    val storageSize: Long = 1024L * 1024
-    implicit val log      = new Log.NOPLog[Task]
+    val initial                            = Genesis.withoutContracts(bonds, 1L, deployTimestamp, "rchain")
+    val storageDirectory                   = Files.createTempDirectory(s"hash-set-casper-test-genesis")
+    val storageSize: Long                  = 1024L * 1024
+    implicit val log                       = new Log.NOPLog[Task]
+    implicit val metricsEff: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
 
     val activeRuntime =
       Runtime.create[Task, Task.Par](storageDirectory, storageSize, StoreType.LMDB).unsafeRunSync
--- a/node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
+++ b/node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
@@ -3,18 +3,16 @@ package coop.rchain.node.diagnostics
 import java.lang.management.{ManagementFactory, MemoryType}
 
 import scala.collection.JavaConverters._
-
 import cats.effect.Sync
 import cats.implicits._
-
 import coop.rchain.catscontrib.Capture
 import coop.rchain.comm.discovery._
 import coop.rchain.comm.rp.Connect.ConnectionsCell
-import coop.rchain.metrics.Metrics
+import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.node.model.diagnostics._
-
 import com.google.protobuf.ByteString
 import com.google.protobuf.empty.Empty
+import coop.rchain.metrics.Metrics.Source
 import javax.management.ObjectName
 import monix.eval.Task
 
@@ -163,6 +161,13 @@ package object effects {
   def metrics[F[_]: Sync]: Metrics[F] =
     new Metrics[F] {
       import kamon._
+      import kamon.trace.{Span => KSpan}
+
+      case class KamonSpan(span: KSpan) extends Span[F] {
+        @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+        override def mark(name: String): F[Unit] = Sync[F].delay { span.mark(name) }
+        override def close(): F[Unit]            = Sync[F].delay { span.finish() }
+      }
 
       private val m = scala.collection.concurrent.TrieMap[String, metric.Metric[_]]()
 
@@ -226,6 +231,10 @@ package object effects {
               _ = t.stop()
             } yield r
         }
+
+      def span(source: Source): F[Span[F]] = Sync[F].delay {
+        KamonSpan(Kamon.buildSpan(source).start())
+      }
     }
 
   def grpc(
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
@@ -1,6 +1,8 @@
 package coop.rchain.rholang.interpreter
 import java.io.StringReader
 
+import coop.rchain.metrics
+import coop.rchain.metrics.Metrics
 import coop.rchain.rholang.Resources.mkRuntime
 import monix.eval.Task
 import monix.execution.Scheduler.Implicits.global
@@ -10,12 +12,13 @@ import coop.rchain.shared.Log
 import scala.concurrent.duration._
 
 class RuntimeSpec extends FlatSpec with Matchers {
-  private val mapSize          = 10L * 1024L * 1024L
-  private val tmpPrefix        = "rspace-store-"
-  private val maxDuration      = 5.seconds
-  implicit val logF: Log[Task] = Log.log[Task]
+  private val mapSize                     = 10L * 1024L * 1024L
+  private val tmpPrefix                   = "rspace-store-"
+  private val maxDuration                 = 5.seconds
+  implicit val logF: Log[Task]            = Log.log[Task]
+  implicit val noopMetrics: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
 
-//  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
+  //  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
 
   private val channelReadOnlyError = "ReduceError: Trying to read from non-readable channel."
 
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
@@ -18,8 +18,6 @@ class RuntimeSpec extends FlatSpec with Matchers {
   implicit val logF: Log[Task]            = Log.log[Task]
   implicit val noopMetrics: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
 
-  //  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
-
   private val channelReadOnlyError = "ReduceError: Trying to read from non-readable channel."
 
   "rho:io:stdout" should "not get intercepted" in {
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryOps.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryOps.scala
@@ -75,10 +75,8 @@ trait InMemoryOps[S] extends CloseOps {
       private[this] val initial = stateRef.take
       private[this] var current = initial
 
-      override def commit(): Unit = {
+      override def commit(): Unit =
         stateRef.put(current)
-        updateGauges()
-      }
 
       override def abort(): Unit = stateRef.put(initial)
 
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -61,20 +61,17 @@ class InMemoryStore[F[_], T, C, P, A, K](
       txn <- txnF
       retErr <- syncF.delay {
                  val r = f(txn)
+                 txn.readState({ state =>
+                   entriesGauge.set(state.size)
+                 })
                  txn.commit()
                  r
                }.attempt
-      _ <- syncF.delay {
-            updateGauges()
-            txn.close()
-          }
+      _   <- syncF.delay { txn.close() }
       ret <- retErr.fold(syncF.raiseError, _.pure[F])
     } yield ret
 
-  private[rspace] def updateGauges() =
-    withTxn(createTxnRead())(_.readState { state =>
-      entriesGauge.set(state.size)
-    })
+  private[rspace] def updateGauges(): Unit = ???
 
   private[rspace] def hashChannels(channels: Seq[C]): Blake2b256Hash =
     StableHashProvider.hash(channels)
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -1,14 +1,13 @@
 package coop.rchain.rspace
 
+import cats.effect.Sync
+
 import scala.collection.immutable.Seq
-
 import cats.implicits._
-
 import coop.rchain.rspace.history.{Branch, ITrieStore}
 import coop.rchain.rspace.internal._
 import coop.rchain.rspace.util.canonicalize
 import coop.rchain.shared.SeqOps.{dropIndex, removeFirst}
-
 import kamon._
 import scodec.Codec
 
@@ -36,7 +35,7 @@ object State {
 class InMemoryStore[F[_], T, C, P, A, K](
     val trieStore: ITrieStore[T, Blake2b256Hash, GNAT[C, P, A, K]],
     val trieBranch: Branch
-)(implicit sc: Serialize[C], sp: Serialize[P], sa: Serialize[A], sk: Serialize[K])
+)(implicit sc: Serialize[C], sp: Serialize[P], sa: Serialize[A], sk: Serialize[K], syncF: Sync[F])
     extends InMemoryOps[State[C, P, A, K]]
     with IStore[F, C, P, A, K] {
 
@@ -52,6 +51,25 @@ class InMemoryStore[F[_], T, C, P, A, K](
   private[this] val MetricsSource = RSpaceMetricsSource + ".in-mem"
   private[this] val refine        = Map("path" -> "inmem")
   private[this] val entriesGauge  = Kamon.gauge(MetricsSource + ".entries").refine(refine)
+
+  private[rspace] def createTxnReadF(): F[Transaction] = syncF.delay(createTxnRead())
+
+  private[rspace] def createTxnWriteF(): F[Transaction] = syncF.delay(createTxnWrite())
+
+  private[rspace] def withTxnF[R](txnF: F[Transaction])(f: Transaction => R): F[R] =
+    for {
+      txn <- txnF
+      retErr <- syncF.delay {
+                 val r = f(txn)
+                 txn.commit()
+                 r
+               }.attempt
+      _ <- syncF.delay {
+            updateGauges()
+            txn.close()
+          }
+      ret <- retErr.fold(syncF.raiseError, _.pure[F])
+    } yield ret
 
   private[rspace] def updateGauges() =
     withTxn(createTxnRead())(_.readState { state =>
@@ -276,7 +294,8 @@ object InMemoryStore {
       implicit sc: Serialize[C],
       sp: Serialize[P],
       sa: Serialize[A],
-      sk: Serialize[K]
+      sk: Serialize[K],
+      syncF: Sync[F]
   ): IStore[F, C, P, A, K] =
-    new InMemoryStore[F, T, C, P, A, K](trieStore, branch)(sc, sp, sa, sk)
+    new InMemoryStore[F, T, C, P, A, K](trieStore, branch)(sc, sp, sa, sk, syncF)
 }
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -33,12 +33,12 @@ object State {
 }
 
 @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
-class InMemoryStore[T, C, P, A, K](
+class InMemoryStore[F[_], T, C, P, A, K](
     val trieStore: ITrieStore[T, Blake2b256Hash, GNAT[C, P, A, K]],
     val trieBranch: Branch
 )(implicit sc: Serialize[C], sp: Serialize[P], sa: Serialize[A], sk: Serialize[K])
     extends InMemoryOps[State[C, P, A, K]]
-    with IStore[C, P, A, K] {
+    with IStore[F, C, P, A, K] {
 
   type TrieTransaction = T
 
@@ -269,7 +269,7 @@ object InMemoryStore {
       case Right(value) => value
     }
 
-  def create[T, C, P, A, K](
+  def create[F[_], T, C, P, A, K](
       trieStore: ITrieStore[T, Blake2b256Hash, GNAT[C, P, A, K]],
       branch: Branch
   )(
@@ -277,6 +277,6 @@ object InMemoryStore {
       sp: Serialize[P],
       sa: Serialize[A],
       sk: Serialize[K]
-  ): InMemoryStore[T, C, P, A, K] =
-    new InMemoryStore[T, C, P, A, K](trieStore, branch)(sc, sp, sa, sk)
+  ): IStore[F, C, P, A, K] =
+    new InMemoryStore[F, T, C, P, A, K](trieStore, branch)(sc, sp, sa, sk)
 }
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -31,7 +31,6 @@ object State {
   def empty[C, P, A, K]: State[C, P, A, K] = State[C, P, A, K](Map.empty, Map.empty)
 }
 
-@SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
 class InMemoryStore[F[_], T, C, P, A, K](
     val trieStore: ITrieStore[T, Blake2b256Hash, GNAT[C, P, A, K]],
     val trieBranch: Branch
@@ -237,6 +236,7 @@ class InMemoryStore[F[_], T, C, P, A, K](
   private[this] def isOrphaned(gnat: GNAT[C, P, A, K]): Boolean =
     gnat.data.isEmpty && gnat.wks.isEmpty
 
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
   protected def processTrieUpdate(update: TrieUpdate[C, P, A, K]): Unit =
     update match {
       case TrieUpdate(_, Insert, channelsHash, gnat) =>
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -56,20 +56,22 @@ class InMemoryStore[F[_], T, C, P, A, K](
 
   private[rspace] def createTxnWriteF(): F[Transaction] = syncF.delay(createTxnWrite())
 
-  private[rspace] def withTxnF[R](txnF: F[Transaction])(f: Transaction => R): F[R] =
+  private[rspace] def withTxnFlatF[R](txnF: F[Transaction])(f: Transaction => F[R]): F[R] =
     for {
-      txn <- txnF
-      retErr <- syncF.delay {
-                 val r = f(txn)
-                 txn.readState({ state =>
-                   entriesGauge.set(state.size)
-                 })
-                 txn.commit()
-                 r
-               }.attempt
-      _   <- syncF.delay { txn.close() }
-      ret <- retErr.fold(syncF.raiseError, _.pure[F])
+      txn    <- txnF
+      retErr <- f(txn).attempt
+      _ <- syncF.delay(retErr.map { _ =>
+            txn.readState({ state =>
+              entriesGauge.set(state.size)
+            })
+            txn.commit()
+          })
+      _   <- syncF.delay(txn.close())
+      ret <- syncF.fromEither(retErr)
     } yield ret
+
+  private[rspace] def withTxnF[R](txnF: F[Transaction])(f: Transaction => R): F[R] =
+    withTxnFlatF[R](txnF)(txn => syncF.delay { f(txn) })
 
   private[rspace] def updateGauges(): Unit = ???
 
--- a/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
@@ -4,6 +4,7 @@ import cats.effect.Sync
 
 import scala.collection.immutable.Seq
 import cats.implicits._
+import coop.rchain.catscontrib.ski.kp
 import coop.rchain.rspace.history.{Branch, ITrieStore}
 import coop.rchain.rspace.internal._
 import coop.rchain.rspace.util.canonicalize
@@ -57,14 +58,20 @@ class InMemoryStore[F[_], T, C, P, A, K](
 
   private[rspace] def withTxnFlatF[R](txnF: F[Transaction])(f: Transaction => F[R]): F[R] =
     for {
-      txn    <- txnF
-      retErr <- f(txn).attempt
-      _ <- syncF.delay(retErr.map { _ =>
-            txn.readState({ state =>
-              entriesGauge.set(state.size)
-            })
-            txn.commit()
-          })
+      txn <- txnF
+      retErr <- f(txn)
+                 .flatMap(
+                   r =>
+                     syncF
+                       .delay {
+                         txn.readState { state =>
+                           entriesGauge.set(state.size)
+                         }
+                         txn.commit()
+                       }
+                       .as(r)
+                 )
+                 .attempt
       _   <- syncF.delay(txn.close())
       ret <- syncF.fromEither(retErr)
     } yield ret
--- a/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
@@ -41,10 +41,6 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
   private[this] val consumeCommCounter = Kamon.counter(MetricsSource + ".comm.consume")
   private[this] val produceCommCounter = Kamon.counter(MetricsSource + ".comm.produce")
 
-  private[this] val consumeSpan   = Kamon.buildSpan(MetricsSource + ".consume")
-  private[this] val produceSpan   = Kamon.buildSpan(MetricsSource + ".produce")
-  protected[this] val installSpan = Kamon.buildSpan(MetricsSource + ".install")
-
   @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements")) // TODO remove when Kamon replaced with Metrics API
   override def consume(
       channels: Seq[C],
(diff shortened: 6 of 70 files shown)


rspace/src/test/scala/coop/rchain/rspace/concurrent/MultiLockTest.scala
```diff
@@ -93,11 +94,13 @@ class MultiLockTest extends FlatSpec with Matchers {
 
     def acquire(seq: List[String]) =
       tested.acquire(seq) {
-        for {
-          k <- seq
-          v = m.getOrElse(k, 0) + 1
-          _ = m.put(k, v)
-        } yield ()
+        seq.pure[IO].map { keys =>
```
goral09: Why are you using `cats.IO` in tests if you use `monix Task` in production?

dzajkowski: wrong person to ask...

goral09: I think you should be using the same effect. `Task` and `IO` should have similar semantics but it may not be true in every case which means you can have tests pass but have different behavior in production.

dzajkowski:
bors try

rspace/src/test/scala/coop/rchain/rspace/concurrent/MultiLockTest.scala
dzajkowski: agreed. but this PR does not address this issue :)

goral09: Sure, just sth I noticed and drew my attention.

dzajkowski:
bors try

dzajkowski:
bors try

dzajkowski:
bors try

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -60,84 +60,120 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+                         .create(channels, patterns, continuation, persist, sequenceNumber)
+                         .pure[F]
+          result <- consumeLockF(channels) {
+                     for {
+                       _ <- ().pure[F]
+                       _ = logger.debug(
+                         s"""|consume: searching for data matching <patterns: $patterns>
+                             |at <channels: $channels>""".stripMargin.replace('\n', ' ')
+                       )
+                       _ = eventLog.update(consumeRef +: _)
```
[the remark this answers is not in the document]
ArturGajowy: I paused here

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
```diff
@@ -53,6 +52,25 @@ class InMemoryStore[F[_], T, C, P, A, K](
...
+      retErr <- syncF.delay {
+                 val r = f(txn)
+                 txn.commit()
+                 r
+               }.attempt
+      _ <- syncF.delay {
+            updateGauges()
+            txn.close()
+          }
+      ret <- retErr.fold(syncF.raiseError, _.pure[F])
```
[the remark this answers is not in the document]
dzajkowski: I will change that in a later commit (which also touches these methods)

rspace/src/main/scala/coop/rchain/rspace/concurrent/MultiLock.scala
```diff
@@ -36,36 +39,24 @@ class DefaultMultiLock[K] extends MultiLock[K] {
...
+    def acquireLocks =
+      for {
+        semaphores <- keys.toSet.toList.sorted.traverse(
+                       k =>
+                         Semaphore[F](1)
+                           .map { s => // if the lock exists this is waste. how to avoid?
+                             locks.getOrElseUpdate(k, s)
+                           }
+                     )
+        acquired <- semaphores.traverse(s => s.acquireN(1).map(_ => s))
```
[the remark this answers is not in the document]
dzajkowski: "inspired" by the implementation of Semaphore, will revert

rspace/src/main/scala/coop/rchain/rspace/concurrent/MultiLock.scala
```diff
@@ -36,36 +39,24 @@ class DefaultMultiLock[K] extends MultiLock[K] {
...
-                    }))
-                    .sequence
-      res = thunk
-      _   <- openLocks.map(_.release).sequence
-    } yield (res))
-
+  def acquire[R](keys: Seq[K])(thunk: => F[R])(implicit o: Ordering[K]): F[R] = {
+    def acquireLocks =
+      for {
+        semaphores <- keys.toSet.toList.sorted.traverse(
```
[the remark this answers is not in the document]
dzajkowski: the implementation is based on `DefaultMultiLock` which is tested in production. the keys need to be unique.

rspace/src/main/scala/coop/rchain/rspace/concurrent/MultiLock.scala
```diff
@@ -36,36 +39,24 @@ class DefaultMultiLock[K] extends MultiLock[K] {
...
+                         Semaphore[F](1)
+                           .map { s => // if the lock exists this is waste. how to avoid?
+                             locks.getOrElseUpdate(k, s)
+                           }
+                     )
+        acquired <- semaphores.traverse(s => s.acquireN(1).map(_ => s))
+      } yield acquired
+    def releaseLocks(acquired: List[Semaphore[F]]) = acquired.traverse(_.release).as(())
+
+    Concurrent[F].bracket(acquireLocks)(_ => thunk)(releaseLocks)
```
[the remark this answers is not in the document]
dzajkowski: ill ticket the discovery for this thing.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-          runMatcher(commRef) match {
+          runMatcher(commRef).map {
             case Some(x) => Right(x)
             case None    => Left(commRef)
           }
         case commRef :: rem =>
-          runMatcher(commRef) match {
-            case Some(x) => Right(x)
+          runMatcher(commRef).flatMap {
+            case Some(x) => Right(x).asInstanceOf[Either[COMM, Seq[DataCandidate[C, R]]]].pure[F]
```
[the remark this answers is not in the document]
dzajkowski: please refer to a later iteration where I addressed (hopefully) all those little kinks

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -141,26 +140,37 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-          }
-        case commRef :: rem =>
-          runMatcher(commRef).flatMap {
-            case Some(x) => Right(x).asInstanceOf[Either[COMM, Seq[DataCandidate[C, R]]]].pure[F]
-            case None    => getCommOrDataCandidates(rem)
-          }
+    def getCommOrDataCandidates(comms: Seq[COMM]): F[Either[COMM, Seq[DataCandidate[C, R]]]] = {
+      type COMMOrData  = Either[COMM, Seq[DataCandidate[C, R]]]
+      type Accumulator = (Seq[COMM], Option[COMMOrData])
+      implicit class RichAccumulator(acc: Accumulator) {
```
ArturGajowy: Why not case class? If you mean to save allocations, I think there are no savings:
 - each tuple is an object instance afaik, exchangable for case class allocation
 - each RichAccumulator method call causes another allocation, defeating the effort to save allocations

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -141,26 +140,37 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
+              case Some(x) => Accumulator(Nil, Some(x.asRight[COMM]))
+              case None    => Accumulator(Nil, Some(Left(commRef)))
+            }
+          case commRef :: rem =>
+            runMatcher(commRef).map {
+              case Some(x) => Accumulator(rem, Some(x.asRight[COMM]))
+              case None    => Accumulator(rem, None)
+            }
+        }
+      concurrent.iterateWhileM(Accumulator(comms))(go)(_.result.isEmpty).map(_.result.get)
```
ArturGajowy: this is a nice improvement, but wouldn't it be clearer to use

```scala
def tailRecM[A, B](a: A)(f: A => F[Either[A, B]]): F[B]
```

directly?

I think
```scala
type Accumulator = (Seq[COMM], Option[COMMOrData])

//because of
// (A, Option[B]) === Either[A, (A, B)]

//is isomorphic to 
type Accumulator = Either[Seq[COMM], (Seq[COMM], COMMOrData)]  
```

which fits nicely the tailrecM signature.

It also seems to me that we only need the `Seq[COMM]` during iteration, so we could
tailrecM with `f: Seq[COMM] => Either[Seq[COMM], COMMOrData]` and get the COMMOrData that we need on the output without any `.get`-s or `_._2`-s

ArturGajowy: and we'd get rid of `Option.get` (yay!)

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -198,43 +208,97 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
+                case Right(produceCandidate) => Accumulator(remaining, produceCandidate)
+                case Left(_)                 => ???
+              }
+            } yield result
+        }
+      concurrent
+        .iterateWhileM(Accumulator(groupedChannels.toList))(go)(
+          acc => acc.channels.nonEmpty && acc.result.isEmpty
+        )
+        .map(_.result)
```
ArturGajowy: same thing here, we could probably use tailrecM

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
ArturGajowy: Aaand we'd not need the Accumulator type anymore

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -198,43 +208,97 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
+              case Some(x) => Accumulator(Nil, Some(Right(x)))
+              case None    => Accumulator(Nil, Some(Left(commRef)))
+            }
+          case commRef :: rem =>
+            runMatcher(commRef, produceRef, groupedChannels) map {
+              case Some(x) => Accumulator(rem, Some(Right(x)))
+              case None    => Accumulator(rem, None)
+            }
+        }
+      concurrent.iterateWhileM(Accumulator(comms))(go)(_.result.isEmpty).map(_.result.get)
```
ArturGajowy: same thing here, tailrecM with `f: Seq[COMM] => Either[Seq[COMM], COMMOrProduce]`

one can think of a type `Iter[A, B] = A => Either[A, B]` which means "a function to be called with accumulator `A` until it returns a `B` and is returning the new `A` otherwise. This is exactly what tailrecM takes as its f argument

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
ArturGajowy: Unless I'm wrong ofc

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
```diff
@@ -56,21 +56,23 @@ class InMemoryStore[F[_], T, C, P, A, K](
...
-      _   <- syncF.delay { txn.close() }
+      txn    <- txnF
+      retErr <- f(txn).attempt
+      _ = retErr.map { _ =>
+        txn.readState({ state =>
+          entriesGauge.set(state.size)
+        })
+        txn.commit()
+      }
+      _   = txn.close()
```
ArturGajowy: delay plox, u nasty hacker! :smiling_imp:

ArturGajowy: here and in the other `_ =` place

ArturGajowy: please check the other withTxn[Flat]F impls for this defect

rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
```diff
@@ -140,16 +141,20 @@ abstract class RSpaceOps[F[_]: Concurrent, C, P, E, A, R, K](
...
-        store.bulkInsert(txn, leaves.map { case Leaf(k, v) => (k, v) })
-      }
+    store.withTxnFlatF(store.createTxnWriteF()) { txn =>
+      for {
+        leaves <- store
+                   .withTrieTxn(txn) { trieTxn =>
+                     store.trieStore.validateAndPutRoot(trieTxn, store.trieBranch, root)
+                     store.trieStore.getLeaves(trieTxn, root)
+                   }
+                   .pure[F]
```
ArturGajowy: delay :) pure is almost always wrong if you can't tell it's a pure value right away

rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
```diff
@@ -140,16 +141,20 @@ abstract class RSpaceOps[F[_]: Concurrent, C, P, E, A, R, K](
...
+        _ = eventLog.update(kp(Seq.empty))
+        _ = store.getAndClearTrieUpdates()
+        _ = store.clear(txn)
+        _ <- restoreInstalls(txn)
+        _ = store.bulkInsert(txn, leaves.map { case Leaf(k, v) => (k, v) })
```
ArturGajowy: we shouldn't rely on the `_ =` hack because we lose guidance from the type system and might accidentaly `_ = x` where x is of type `F[A]` and thus lose effects.

I'd rather see a `.delay` in every line above, which has the additional value of screaming "please ~~kill~~ refactor me"

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -204,13 +213,25 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-        lockedProduce(channel, data, persist, sequenceNumber)
-      }
+      for {
+        span <- metricsF.span(produceSpanLabel)
+        _    = span.mark("before-produce-lock")
+        result <- produceLockF(channel) {
+                   lockedProduce(channel, data, persist, sequenceNumber, span)
+                 }
+        _ = span.mark("post-produce-lock")
+        _ = span.close()
```
ArturGajowy: what happens if there's an error and we don't close the span? The Bracket/Resource thing for spans might actually be necessary depending on the answer :)

rspace/src/main/scala/coop/rchain/rspace/InMemoryOps.scala
```diff
@@ -75,10 +75,8 @@ trait InMemoryOps[S] extends CloseOps {
       private[this] val initial = stateRef.take
       private[this] var current = initial
 
-      override def commit(): Unit = {
+      override def commit(): Unit =
         stateRef.put(current)
-        updateGauges()
```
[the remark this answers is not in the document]
ArturGajowy: If I get it right, you've moved the guage update into the transaction handling code? Well played :)

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
 
-    store.withTxn(store.createTxnRead()) { txn =>
-      val groupedChannels: Seq[Seq[C]] = store.getJoin(txn, channel)
+    for {
+      groupedChannels <- store.withTxnF(store.createTxnReadF()) { txn =>
+                          store.getJoin(txn, channel)
+                        }
 
-      logger.debug(s"""|produce: searching for matching continuations
+      _ = logger.debug(s"""|produce: searching for matching continuations
```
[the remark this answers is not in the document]
dzajkowski: we do!

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
```diff
@@ -19,10 +16,11 @@ package object implicits {
       override def racePair[A, B](
           fa: Id[A],
           fb: Id[B]
-      ): Id[Either[(A, Fiber[Id, B]), (Fiber[Id, A], B)]]                          = ???
-      override def async[A](k: (Either[Throwable, A] => Unit) => Unit): Id[A]      = ???
+      ): Id[Either[(A, Fiber[Id, B]), (Fiber[Id, A], B)]] = ???
+      override def async[A](k: (Either[Throwable, A] => Unit) => Unit): Id[A] =
+        asyncF(cb => syncId.delay(k(cb)))
```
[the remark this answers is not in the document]
dzajkowski: kk

rspace/src/main/scala/coop/rchain/rspace/InMemoryOps.scala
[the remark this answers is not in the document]
dzajkowski: aligned it with the other implementations

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -166,27 +168,32 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
         logF.error(msg) *> syncF.raiseError(new IllegalArgumentException(msg))
       } else
         for {
+          span <- metricsF.span(consumeSpanLabel)
+          _    = span.mark("before-consume-ref-compute")
```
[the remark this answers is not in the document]
dzajkowski: yeah I figured it's better to be conservative here.
pruned the last batch and will add new ones when they are needed.

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -166,27 +168,32 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
           consumeRef <- Consume
                          .create(channels, patterns, continuation, persist, sequenceNumber)
                          .pure[F]
           result <- consumeLockF(channels) {
                      for {
+                       _ <- span.mark("consume-lock-acquired").pure[F]
```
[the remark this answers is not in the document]
dzajkowski: agreed to F

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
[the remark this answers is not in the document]
JosDenmark: @ArturGajowy I agree that the use of `delay` versus value declarations should be used consistently, but I don't see why these need to be delayed. Sure they're effects, but they're sequenced correctly.

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -60,84 +60,120 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
                                  |storing <(patterns, continuation): ($patterns, $continuation)>
                                  |at <channels: $channels>""".stripMargin.replace('\n', ' '))
-                Right(None)
-              case Right(Some(dataCandidates)) =>
-                consumeCommCounter.increment()
-
-                eventLog.update(COMM(consumeRef, dataCandidates.map(_.datum.source)) +: _)
+                                        Right(None)
+                                      }
+                                  case Right(Some(dataCandidates)) =>
```
JosDenmark: I see why you used the syntax you did, because the `dataCandidates` traverse is the only monadic function. IMHO, this case is easier to grok if it's wrapped in a `for` comprehension.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
dzajkowski: span does not send

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -153,121 +189,179 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+      for {
+        produceRef <- Produce.create(channel, data, persist, sequenceNumber).pure[F]
+        result <- produceLockF(channel) {
+                   /*
+                    * Find produce candidate
+                    *
+                    * Could also be implemented with a lazy `foldRight`.
+                    */
+                   type MaybeCandidate = Option[ProduceCandidate[C, P, R, K]]
+                   //@tailrec <- fix tailrec
```
JosDenmark: Does it matter that this function is tail recursive, since flatMap is trampolined?

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -153,121 +189,179 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+                                     persistK,
+                                     consumeRef
+                                   ),
+                                   continuationIndex,
+                                   dataCandidates
+                                 )
+                               )
+                               ) =>
+                             syncF
+                               .delay {
```
JosDenmark: Even more so than in my previous comment, this case continuation could be expressed more clearly as a for comprehension.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-
-        commOrDataCandidates match {
-          case Left(commRef) =>
-            Right(storeWaitingContinuation(consumeRef, Some(commRef)))
-          case Right(dataCandidates) =>
-            Right(handleMatches(dataCandidates, consumeRef, comms))
-        }
-    }
+    for {
+      _          <- logger.debug(s"""|consume: searching for data matching <patterns: $patterns>
```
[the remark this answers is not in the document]
JosDenmark: In previous comprehensions that need to start with some sort of logging, you do something like `for { _ <- ().pure[F] ; _ = logger.debug(...) ...`. Whatever you choose, it should probably be consistent.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-      mats
-        .sortBy(_.datumIndex)(Ordering[Int].reverse)
-        .foreach {
-          case DataCandidate(candidateChannel, Datum(_, persistData, _), dataIndex) =>
-            if (!persistData) {
-              store.withTxn(store.createTxnWrite()) { txn =>
-                store.removeDatum(txn, Seq(candidateChannel), dataIndex)
+    ): F[Option[(ContResult[C, P, K], Seq[Result[R]])]] =
+      for {
+        _       <- consumeCommCounter.increment().pure[F]
```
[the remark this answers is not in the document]
JosDenmark: Some of the comprehensions in `RSpace` were started with `for { _ <- ().pure[F] ; _ = consumeCommCounter.increment() ...`. Whatever it is, it should be consistent.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
+    //@tailrec <- todo fix tailrec
     def runMatcher(
         comm: COMM,
         produceRef: Produce,
         groupedChannels: Seq[Seq[C]]
-    ): Option[ProduceCandidate[C, P, R, K]] =
+    ): F[Option[ProduceCandidate[C, P, R, K]]] =
       groupedChannels match {
-        case Nil => None
+        case Nil => None.asInstanceOf[Option[ProduceCandidate[C, P, R, K]]].pure[F]
```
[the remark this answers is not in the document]
JosDenmark: You can do `none[Option[ProduceCandidate[C, P, R, K]]].pure[F]` with `cats.syntax` imported.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-          }
-          dataCandidates
-            .sortBy(_.datumIndex)(Ordering[Int].reverse)
-            .foreach {
-              case DataCandidate(candidateChannel, Datum(_, persistData, _), dataIndex) =>
-                store.withTxn(store.createTxnWrite()) { txn =>
-                  if (!persistData && dataIndex >= 0) {
-                    store.removeDatum(txn, Seq(candidateChannel), dataIndex)
+          for {
+            _       <- syncF.delay { produceCommCounter.increment() }
```
JosDenmark: Not to beat a dead horse, but, again, let's be consistent with syntax here.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
+                    case DataCandidate(candidateChannel, Datum(_, persistData, _), dataIndex) =>
+                      store.withTxnF(store.createTxnWriteF()) { txn =>
+                        if (!persistData && dataIndex >= 0) {
+                          store.removeDatum(txn, Seq(candidateChannel), dataIndex)
+                        }
+                        store.removeJoin(txn, candidateChannel, channels)
+                      }
+                  }
+            r = {
+              logger.debug(s"produce: matching continuation found at <channels: $channels>")
```
JosDenmark: This entire block can be expressed more clearly as:
```
_ = logger.debug(s"produce: matching continuation found at <channels: $channels>")
  _ = removeBindingsFor(commRef)
  contSequenceNumber = commRef.nextSequenceNumber
 } yield Some((ContResult(continuation, persistK, channels, patterns, contSequenceNumber),
              dataCandidates.map(dc => Result(dc.datum.a, dc.datum.persist))))
```

casper/src/test/scala/coop/rchain/casper/HashSetCasperTest.scala
```diff
@@ -1816,10 +1817,11 @@ object HashSetCasperTest {
...
   ): BlockMessage = {
-    val initial           = Genesis.withoutContracts(bonds, 1L, deployTimestamp, "rchain")
-    val storageDirectory  = Files.createTempDirectory(s"hash-set-casper-test-genesis")
-    val storageSize: Long = 1024L * 1024
-    implicit val log      = new Log.NOPLog[Task]
+    val initial                            = Genesis.withoutContracts(bonds, 1L, deployTimestamp, "rchain")
+    val storageDirectory                   = Files.createTempDirectory(s"hash-set-casper-test-genesis")
+    val storageSize: Long                  = 1024L * 1024
+    implicit val log                       = new Log.NOPLog[Task]
+    implicit val metricsEff: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
```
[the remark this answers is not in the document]
dzajkowski: yeah...

rspace/src/main/scala/coop/rchain/rspace/concurrent/MultiLock.scala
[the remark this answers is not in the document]
dzajkowski: `Semaphore[F](1) returns F[Semaphore[F](1)]`

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -68,263 +68,283 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[F, C, P, A, K], branch:
...
-        .foreach {
-          case DataCandidate(candidateChannel, Datum(_, persistData, _), dataIndex) =>
-            if (!persistData) {
-              store.withTxn(store.createTxnWrite()) { txn =>
-                store.removeDatum(txn, Seq(candidateChannel), dataIndex)
+    ): F[Option[(ContResult[C, P, K], Seq[Result[R]])]] =
+      for {
+        _       <- consumeCommCounter.increment().pure[F]
+        commRef = COMM(consumeRef, mats.map(_.datum.source))
+        _       = assert(comms.contains(commRef), "COMM Event was not contained in the trace")
```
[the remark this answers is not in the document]
dzajkowski: as long as we can agree that BugFoundError gets bubbled?

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -153,121 +189,179 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+                    */
+                   type MaybeCandidate = Option[ProduceCandidate[C, P, R, K]]
+                   //@tailrec <- fix tailrec
+                   def extractProduceCandidate(
+                       groupedChannels: Seq[Seq[C]],
+                       batChannel: C,
+                       data: Datum[A]
+                   ): F[Either[E, MaybeCandidate]] =
+                     groupedChannels match {
+                       case Nil => Right(None).asInstanceOf[Either[E, MaybeCandidate]].pure[F]
```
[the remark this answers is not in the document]
dzajkowski: they are gone

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
```diff
@@ -61,20 +61,17 @@ class InMemoryStore[F[_], T, C, P, A, K](
...
-          }
+      _   <- syncF.delay { txn.close() }
       ret <- retErr.fold(syncF.raiseError, _.pure[F])
     } yield ret
 
-  private[rspace] def updateGauges() =
-    withTxn(createTxnRead())(_.readState { state =>
-      entriesGauge.set(state.size)
-    })
+  private[rspace] def updateGauges(): Unit = ???
```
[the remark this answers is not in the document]
dzajkowski: it comes from the interface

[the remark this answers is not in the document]
dzajkowski: it comes from the interface

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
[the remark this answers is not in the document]
ArturGajowy: Assuming a lazy/trampolining monad, they might be. I'm not sure though the laws for Sync guarantee that and the lazy behavior of `map` might be a coincidence (one could imagine a trampolining monad that does pure / map / flatMap eagerly as long as lhs is not a `Delay(f)`). I'm not sure if we always run with trampolined lazy monads and would prefer to not have to think about that. Hence delay :)

Also, see my latter comment on the same matter:
https://github.com/rchain/rchain/pull/2207#discussion_r260349246

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
ArturGajowy: I believe after the refactor to tailrecM / iterateWhileM this is a stray comment - and no matter what monad we use (as long as it has a lawful tailrecM)

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -41,6 +40,100 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+  private[this] def fetchChannelToIndexData(channels: Seq[C]) =
+    store.withTxnF(store.createTxnReadF()) { txn =>
+      channels.map { c: C =>
+        c -> Random.shuffle(
+          store.getData(txn, Seq(c)).zipWithIndex
+        )
+      }.toMap
+    }
+
+  private[this] def storeWaitingContinuation(
```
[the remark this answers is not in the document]
dzajkowski: huh I'm sure you're right. hindsight is always 20/20

rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
```diff
@@ -26,30 +26,31 @@ abstract class RSpaceOps[F[_]: Concurrent, C, P, E, A, R, K](
...
     serializeC: Serialize[C],
     serializeP: Serialize[P],
-    serializeK: Serialize[K],
-    syncF: Sync[F]
+    serializeK: Serialize[K]
 ) extends SpaceMatcher[F, C, P, E, A, R, K] {
 
   implicit val codecC = serializeC.toCodec
 
+  val syncF: Sync[F] = Concurrent[F]
```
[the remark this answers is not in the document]
dzajkowski: this `syncF: Sync[F]` comes as a dependency from traits that are composed in this class (and others) and those don't need the Concurrent typeclass. so keeping it simple seemed like the right choice.

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
dzajkowski: addressed in the "refactor" and "enable span" for rspace commit

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
dzajkowski: addressed in the "refactor" and "enable span" for rspace commit

rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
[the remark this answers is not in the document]
JosDenmark: Good enough for me 👍

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -58,325 +55,345 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[C, P, A, K], branch: Br
...
+            }
         }
+      comms.tailRecM(go)
     }
+
+    for {
+      _ <- logF.debug(s"""|consume: searching for data matching <patterns: $patterns>
+                     |at <channels: $channels>""".stripMargin.replace('\n', ' '))
+      consumeRef <- syncF.delay {
+                     Consume.create(channels, patterns, continuation, persist, sequenceNumber)
```
ArturGajowy: I think we could've skipped delaying creation of those data structures - but I guess it's not immediately obvious there are no side-effects underneath, so better safe than sorry :+1:

rspace/src/main/scala/coop/rchain/rspace/concurrent/MultiLock.scala
```diff
@@ -7,38 +7,6 @@ import scala.collection.immutable.Seq
...
-        k         <- sortedKeys
-        semaphore = locks.getOrElse(k, throw new Exception(s"Couldn't find lock for channel $k"))
-        _         = semaphore.release()
-      } yield ()
-    }
-  }
-
-}
-
 class FunctionalMultiLock[F[_]: Concurrent, K] {
```
ArturGajowy: I think we can now rename this to just Multilock :)

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
dzajkowski: I chose to be consistent. If I hear a loud "hell no" then I'll revert.

dzajkowski:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/RuntimeSpec.scala
```diff
@@ -10,12 +12,13 @@ import coop.rchain.shared.Log
...
-  private val maxDuration      = 5.seconds
-  implicit val logF: Log[Task] = Log.log[Task]
+  private val mapSize                     = 10L * 1024L * 1024L
+  private val tmpPrefix                   = "rspace-store-"
+  private val maxDuration                 = 5.seconds
+  implicit val logF: Log[Task]            = Log.log[Task]
+  implicit val noopMetrics: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
 
-//  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
+  //  val runtime = Runtime.create(Files.createTempDirectory(tmpPrefix), mapSize)
```
lukasz-golebiewski: Just delete this

rspace/src/main/scala/coop/rchain/rspace/InMemoryStore.scala
```diff
@@ -53,10 +51,28 @@ class InMemoryStore[T, C, P, A, K](
...
+
+  private[rspace] def withTxnFlatF[R](txnF: F[Transaction])(f: Transaction => F[R]): F[R] =
+    for {
+      txn    <- txnF
+      retErr <- f(txn).attempt
+      _ <- syncF.delay(retErr.map { _ =>
+            txn.readState({ state =>
+              entriesGauge.set(state.size)
+            })
+            txn.commit()
```
lukasz-golebiewski: Can the transaction be committed before updating the gauge in order to not potentially block the single writer thread?

JosDenmark:
I declare this PR approved! 🥇

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -53,266 +145,296 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
+   *
+   * Could also be implemented with a lazy `foldRight`.
+   */
+
+  private[this] def extractProduceCandidate(
+      groupedChannels: Seq[Seq[C]],
+      batChannel: C,
+      data: Datum[A]
+  )(implicit m: Match[P, E, A, R]): F[Either[E, Option[ProduceCandidate[C, P, R, K]]]] = {
+    type MaybeProduceCandidate = Option[ProduceCandidate[C, P, R, K]]
```
lukasz-golebiewski: `type RemainingChannels = Seq[C]` would make it easier to digest

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -53,266 +145,296 @@ class RSpace[F[_], C, P, E, A, R, K] private[rspace] (
...
-                .replace('\n', ' ')
-            )
-            span.mark("before-event-log-lock-acquired")
-            eventLog.update(produceRef +: _)
-            span.mark("event-log-updated")
-
+  /*
+   * Find produce candidate
+   *
+   * Could also be implemented with a lazy `foldRight`.
```
lukasz-golebiewski: This comment looks outdated

dzajkowski:
bors try

dzajkowski:
bors cancel

dzajkowski:
bors try-

dzajkowski:
bors try

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -58,325 +54,341 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[C, P, A, K], branch: Br
...
-            }
-          }
-        }
+        for {
+          span <- metricsF.span(consumeSpanLabel)
+          _    <- span.mark("before-consume-lock")
+          result <- consumeLockF(channels) {
+                     lockedConsume(channels, patterns, continuation, persist, sequenceNumber, span)
+                   }
+          _ <- span.mark("post-consume-lock")
```
lukasz-golebiewski: Are spans closed automagically?

lukasz-golebiewski: This mark is going to be placed right before the point where the span ought to finish, so it isn't very useful. Why not move it to right after the lock is acquired, so that it shows how long the thread waited for the resource?

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
```diff
@@ -163,6 +161,13 @@ package object effects {
   def metrics[F[_]: Sync]: Metrics[F] =
     new Metrics[F] {
       import kamon._
+      import kamon.trace.{Span => KSpan}
+
+      case class KamonSpan(span: KSpan) extends Span[F] {
+        @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+        override def mark(name: String): F[Unit] = Sync[F].delay { span.mark(name) }
+        override def close(): F[Unit]            = Sync[F].delay { span.finish() }
```
lukasz-golebiewski: Why the rename from `finish` to `close`?

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -58,325 +54,341 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[C, P, A, K], branch: Br
...
+              .withTxnF(store.createTxnWriteF()) { txn =>
+                store.putWaitingContinuation(
+                  txn,
+                  channels,
+                  WaitingContinuation(patterns, continuation, persist, consumeRef)
+                )
+                for (channel <- channels) store.addJoin(txn, channel, channels)
+              }
+        _ <- logF.debug(s"""|consume: no data found,
+                         |storing <(patterns, continuation): ($patterns, $continuation)>
```
lukasz-golebiewski: This looks misaligned

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
lukasz-golebiewski: The readers might start to wander why the delay was done and start searching for side effects in `Consume.create`. This is misleading, i'd opt for removal

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -393,7 +405,7 @@ class ReplayRSpace[F[_], C, P, E, A, R, K](store: IStore[C, P, A, K], branch: Br
     for {
       isEmpty <- syncF.delay(replayData.isEmpty)
       checkpoint <- isEmpty.fold(
-                     Checkpoint(store.createCheckpoint(), Seq.empty).pure[F], {
+                     syncF.delay { Checkpoint(store.createCheckpoint(), Seq.empty) }, {
```
lukasz-golebiewski: Isn't this also pure?

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
```diff
@@ -39,45 +47,52 @@ trait ReplayRSpaceTests
...
-      logger.debug("Finished produce {}", i)
-      res
+      space.produce(channelCreator(i), datumCreator(i), persist).map { r =>
+        val unpacked = r.right.get
+        logger.debug("Finished produce {}", i)
+        unpacked
+      }
     }
 
   "reset to a checkpoint from a different branch" should "work" in withTestSpaces {
```
lukasz-golebiewski: `withTestSpaces` should do the `unsafeRunSync`

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
```diff
@@ -39,45 +47,52 @@ trait ReplayRSpaceTests
...
     (space, replaySpace) =>
-      val root0 = replaySpace.createCheckpoint().root
-      replaySpace.store.isEmpty shouldBe true
+      (for {
+        root0 <- replaySpace.createCheckpoint().map(_.root)
+        _     = replaySpace.store.isEmpty shouldBe true
```
lukasz-golebiewski: Does delaying this make sense?

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
```diff
@@ -3,15 +3,34 @@ package coop.rchain.catscontrib.effect
...
-import scala.concurrent.ExecutionContext
 
-import scala.concurrent.duration.Duration
-import scala.concurrent.{Await, Promise}
 import scala.util.{Failure, Success, Try}
 import scala.util.control.NonFatal
 
 package object implicits {
 
+  // this is for testing purposes, do not use in production code!
```
lukasz-golebiewski: This has to be defined somewhere under `test`, otherwise it will be used (Murphy's law)

dzajkowski:
bors try

dzajkowski:
bors try

dzajkowski:
bors try

ArturGajowy:
That was fun! Let's

bors try

that again! :joy:

dzajkowski:
bors try

ArturGajowy:
*woosh...* :P

ArturGajowy:
bors try

dzajkowski:
bors try

dzajkowski:
bors try

dzajkowski:
bors try

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
dzajkowski: I believe `store.createCheckpoint(),` isn't.

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
dzajkowski: it's required by Address Examples

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
dzajkowski: no, it has to be an explicit 'close'.

dzajkowski: It's easier to track in zipkin if there is a closing mark.
all the methods have:
`        _          <- span.mark("before-produce-ref-computed")` just after getting the lock.

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
ArturGajowy: Then their fate is sealed. After merging this PR, we should draw lots who's gonna take the honors to remove them, along with that instance. Unless you'd like to have that privilege, @dzajkowski? (I think you deserve that glory :))

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
dzajkowski: seemed more aligned with our naming scheme

dzajkowski:
bors try

dzajkowski:
bors try

rspace/src/test/scala/coop/rchain/rspace/ReplayRSpaceTests.scala
dzajkowski: we know that task will behave as expected so... no, I don't think there is reason to do so.

shared/src/main/scala/coop/rchain/catscontrib/effect/implicits/package.scala
dzajkowski: go for it.

dzajkowski:
bors try

dzajkowski:
bors try-

dzajkowski:
bors try

PR #2214, opened 2019-02-27
Wallet contracts prerequisites

ArturGajowy:
## Overview

### JIRA ticket:

### Notes


ArturGajowy:
bors try

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
```diff
@@ -0,0 +1,20 @@
...
+trait CompiledRholangSource {
+  val term: Par
+  val code: String
+}
+
+object CompiledRholangSource {
+
+  def apply(classpath: String): CompiledRholangSource = new CompiledRholangSource {
+    override val code: String = Source.fromResource(classpath).mkString
+    override val term: Par    = Interpreter[Coeval].buildNormalizedTerm(code).value()
```
goral09: What was nice about previous way of solving this is that in case of errors in the contract you would get this in the compile time and not Runtime.

ArturGajowy: the disadvantages outweigh the benefits grossly. Also, you get the feedback as soon as you run the tests and/or run the node. And you're able to run the tests much faster than you used to be able to test:compile casper to get the error. Like, 50 ms in IDEA vs 15s in sbt.

ArturGajowy:
bors try

ArturGajowy:
bors try

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
goral09: But the generated rho files were not only used in tests but also part of the Genesis. Which, again, would fail even before you start a node rather than when you want to build a Genesis.

ArturGajowy: Right now Genesis will fail just after you start the node. No big difference. I 'll double check there is a test that will fail should any of the genesis contracts not compile.

ArturGajowy: This also means, that genesis will fail prior to (production) node start, during (integration) tests.

goral09: I understand that it will fail anyway but the compile time failure appealed to me more than runtime failure.

ArturGajowy: Yup, the bootstrap node fails very soon and terminates if there are compilation errors in the genesis contracts. The eval test in the integration tests requires a running bootstrap node. Ergo: tests will catch it.

ArturGajowy: It wasn't a compile-time failure in the previous version of this code. It was in the previous-previous, before https://github.com/rchain/rchain/pull/1623/ was merged (which was too a tremendous upgrade in dev workflow).

> compile time failure appealed to me more than runtime failure

Being able to do a runtime-check in 3s **total** appeals to me more than a compile-time check taking 15s to even start running, and yet another 15s to fail.

`/me throws chair`

I'm resolving this conversation. Feel free to reopen if there are arguments we haven't discussed yet :)

ArturGajowy:
Bors r+

ArturGajowy:
bors try

PR #2215, opened 2019-02-28
Make number of incoming message consumers configurable

sebtomba:
## Overview
Make the number of incoming message consumers configurable. Remove Casper workaround of only 1 message consumer.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3036

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala
@@ -32,7 +32,8 @@ class GrpcTransportServer(
     cert: String,
     key: String,
     maxMessageSize: Int,
-    tempFolder: Path
+    tempFolder: Path,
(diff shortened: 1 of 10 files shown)


sebtomba:
@tymm we can't be sure if this PR definitely resolves the problem. We have a queue for incoming messages. Blocks are getting streamed to a file on disk. There was an issue in Casper with memory utilization. A workaround was to limit the message consumers to 1 so only one block gets processed a time. I don't know the details but this issue seems to be resolved. So now we can have more than one message consumer. We suspect that peering of nodes fails because messages are too long in the message queue and we run into time outs. With more than one message queue consumer there is a higher probability that comm messages are handled in time. If not, we'll have to take much more radical steps and change the comm architecture to have a dedicated message queue for comm messages. Or we'll need to make the handshake protocol asynchronous.

tymm:
@sebtomba thank you for the context.

PR #2216, opened 2019-02-28
Test cost account sharing

lukasz-golebiewski:
## Overview
See ticket

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2989

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
@@ -8,6 +8,7 @@ import cats.effect.{Concurrent, ContextShift, Sync}
 import cats.implicits._
 import cats.mtl.FunctorTell
 import com.google.protobuf.ByteString
+import coop.rchain.catscontrib.mtl.implicits._
 import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.metrics.Metrics
 import coop.rchain.models.Expr.ExprInstance.GString
@@ -17,7 +18,7 @@ import coop.rchain.models._
 import coop.rchain.models.rholang.implicits._
 import coop.rchain.rholang.interpreter.Runtime.ShortLeashParams.ShortLeashParameters
 import coop.rchain.rholang.interpreter.Runtime._
-import coop.rchain.rholang.interpreter.accounting.Cost
+import coop.rchain.rholang.interpreter.accounting._
 import coop.rchain.rholang.interpreter.errors.{InterpreterError, SetupError}
 import coop.rchain.rholang.interpreter.storage.implicits._
 import coop.rchain.rspace._
@@ -276,12 +277,36 @@ object Runtime {
     )
   )
 
+  def createWithEmptyCost[F[_]: ContextShift: Concurrent: Log: Metrics, M[_]](
+      dataDir: Path,
+      mapSize: Long,
+      storeType: StoreType,
+      extraSystemProcesses: Seq[SystemProcess.Definition[F]] = Seq.empty
+  )(
+      implicit
+      P: Parallel[F, M],
+      executionContext: ExecutionContext
+  ): F[Runtime[F]] =
+    (for {
+      costAccounting <- CostAccounting.empty[F]
+      runtime <- {
+        implicit val ca: CostAccounting[F] = costAccounting
+        implicit val cost: _cost[F]        = loggingCost(ca, noOpCostLog)
+        create(dataDir, mapSize, storeType, extraSystemProcesses)
+      }
+    } yield (runtime))
+
   def create[F[_]: ContextShift: Concurrent: Log: Metrics, M[_]](
       dataDir: Path,
       mapSize: Long,
       storeType: StoreType,
       extraSystemProcesses: Seq[SystemProcess.Definition[F]] = Seq.empty
-  )(implicit P: Parallel[F, M], executionContext: ExecutionContext): F[Runtime[F]] = {
+  )(
+      implicit P: Parallel[F, M],
+      executionContext: ExecutionContext,
+      cost: _cost[F],
+      costAccounting: CostAccounting[F]
+  ): F[Runtime[F]] = {
     val errorLog                               = new ErrorLog[F]()
     implicit val ft: FunctorTell[F, Throwable] = errorLog
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
@@ -53,7 +53,7 @@ class RholangAndScalaDispatcher[M[_]] private (
 
 object RholangAndScalaDispatcher {
 
-  def create[M[_], F[_]](
+  def createWithEmptyCost[M[_], F[_]](
       tuplespace: RhoISpace[M],
       dispatchTable: => Map[Long, (Seq[ListParWithRandomAndPhlos], Int) => M[Unit]],
       urnMap: Map[String, Par]
@@ -65,6 +65,22 @@ object RholangAndScalaDispatcher {
   ): (Dispatch[M, ListParWithRandomAndPhlos, TaggedContinuation], ChargingReducer[M], Registry[M]) = {
     implicit val costAlg: CostAccounting[M] = CostAccounting.unsafe[M](Cost(0))
     implicit val cost: _cost[M]             = loggingCost(costAlg, noOpCostLog)
+    create(tuplespace, dispatchTable, urnMap)
+
+  }
+
+  def create[M[_], F[_]](
+      tuplespace: RhoISpace[M],
+      dispatchTable: => Map[Long, (Seq[ListParWithRandomAndPhlos], Int) => M[Unit]],
+      urnMap: Map[String, Par]
+  )(
+      implicit
+      cost: _cost[M],
+      costAccounting: CostAccounting[M],
+      parallel: Parallel[M, F],
+      s: Sync[M],
+      ft: FunctorTell[M, Throwable]
+  ): (Dispatch[M, ListParWithRandomAndPhlos, TaggedContinuation], ChargingReducer[M], Registry[M]) = {
 
     implicit lazy val dispatcher: Dispatch[M, ListParWithRandomAndPhlos, TaggedContinuation] =
       new RholangAndScalaDispatcher(dispatchTable)
@@ -74,7 +90,8 @@ object RholangAndScalaDispatcher {
 
     lazy val tuplespaceAlg = Tuplespace.rspaceTuplespace(chargingRSpace, dispatcher)
 
-    lazy val chargingRSpace: RhoPureSpace[M] = ChargingRSpace.pureRSpace(s, costAlg, tuplespace)
+    lazy val chargingRSpace: RhoPureSpace[M] =
+      ChargingRSpace.pureRSpace(s, costAccounting, tuplespace)
 
     val chargingReducer: ChargingReducer[M] = ChargingReducer[M]
 
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
@@ -114,7 +114,7 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks {
       }
       (result, costLog) = costsLoggingProgram
       res               <- errorLog.readAndClearErrorVector
-      _                 = Task.now(res should be(Vector.empty))
+      _                 <- Task.now(res should be (Vector.empty))
     } yield ((result, costLog))).unsafeRunSync
   }
 
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
@@ -118,10 +118,9 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks {
     } yield ((result, costLog))).unsafeRunSync
   }
 
-  case class ContractWithCost(contract: String, cost: Long)
-
-  val shortslow = ContractWithCost(
-    """new loop in {
+  val contracts = Table(
+    ("contract", "expectedTotalCost"),
+    ("""new loop in {
          contract loop(@n) = {
            match n {
              0 => Nil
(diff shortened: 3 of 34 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -276,12 +277,36 @@ object Runtime {
...
+  )(
+      implicit
+      P: Parallel[F, M],
+      executionContext: ExecutionContext
+  ): F[Runtime[F]] =
+    (for {
+      costAccounting <- CostAccounting.empty[F]
+      runtime <- {
+        implicit val ca: CostAccounting[F] = costAccounting
+        implicit val cost: _cost[F]        = loggingCost(ca, noOpCostLog)
```
ArturGajowy: I believe we should be passing the `cost` as parameter, and decide whether it's logging or not at the end of the world.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/package.scala
```diff
@@ -0,0 +1,25 @@
...
+import cats.implicits._
+import cats.mtl._
+import cats.mtl.implicits._
+
+package object utils {
+
+  def costLog[M[_]: Sync](): FunctorListen[M, Chain[Cost]] =
+    new DefaultFunctorListen[M, Chain[Cost]] {
+      override val functor: Functor[M]  = implicitly[Functor[M]]
+      private val ref                   = Ref.unsafe(Chain.empty[Cost])
```
ArturGajowy: the `costLog` method should return `M[FunctorListen[M, Chain[Cost]]]` and be called at the end of the world, at the beginning of our program.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
lukasz-golebiewski: Sounds good. Let's do that in a separate PR when `CostAccounting` is gone

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/package.scala
lukasz-golebiewski: Makes sense

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +78,95 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+    costLog.toList should contain theSameElementsAs (expectedCosts)
+  }
+
+  it should "stop the evaluation of all execution branches when one of them runs out of phlo" ignore {
+    val initialPhlo       = 1L
+    val contract          = "@1!(1) | @2!(2) | @3!(3)"
+    val expectedCosts     = List(Cost(4, "substitution"))
+    val (result, costLog) = evaluateWithCostLog(initialPhlo, contract)
+    result shouldBe Left(OutOfPhlogistonsError)
+    costLog.toList should contain theSameElementsAs (expectedCosts)
```
ArturGajowy: i'd like to see it fail during the second branch

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +78,95 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+    """new loop in {
+         contract loop(@n) = {
+           match n {
+             0 => Nil
+             _ => loop!(n-1)
+           }
+         } |
+         loop!(10)
+       }""".stripMargin,
+    1766
```
ArturGajowy: I'd like to see more ContractWithCost-s, and a test that uses contracts that have varying top-level par type (send / receive / new / expr /bundle / connective ...). While the tests in here look quite torough, they seem to be never testing a case when more than one `evalJob` is being run in parallel.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +78,95 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+    costLog.map(_.value).toList.sum shouldEqual expectedTotalCost
+  }
+
+  "Running out of phlogistons" should "stop the evaluation upon cost depletion in a single execution branch" in {
+    val initialPhlo       = 1L
+    val contract          = "@1!(1)"
+    val expectedCosts     = List(Cost(4, "substitution"))
+    val (result, costLog) = evaluateWithCostLog(initialPhlo, contract)
+    result shouldBe Left(OutOfPhlogistonsError)
+    costLog.toList should contain theSameElementsAs (expectedCosts)
```
ArturGajowy: I'd much prefer we'd just use `assert()`. the `should` dsl is way too verbose for my likes.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
ArturGajowy: ah, you're saying this thing is a functionally-dependent on costAccounting? I'm fine with that then.

ArturGajowy:
I've got mixed feeling about approving a PR focused on tests that ignores 3 of 4 tests introduced. I appreciate that this is because of dependency on @JosDenmark's work, but then my thinking is that maybe you guys should pair program on a single PR.

We can probably merge the refactorings indepenedently.

lukasz-golebiewski:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
JosDenmark: Why do we expect the cost of a substitution here when no substitution should be performed on the contract? It has no variables after-all.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -53,7 +53,7 @@ class RholangAndScalaDispatcher[M[_]] private (
 
 object RholangAndScalaDispatcher {
 
-  def create[M[_], F[_]](
+  def createWithEmptyCost[M[_], F[_]](
```
JosDenmark: Why not write a more general `createWithCost(cost: Cost, ...)`. That would cover this and other cases.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: @JosDenmark apparently sends are also being substituted: https://github.com/rchain/rchain/blob/f89a0f2e4f9036bc51d7d2ada4594a00358b1469/rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala#L167

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +72,98 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+                                par <- Interpreter[Task].buildNormalizedTerm(contract)
+                                res <- Interpreter[Task]
+                                        .evaluate(runtime, par, Cost(initialPhlo.toLong))
+                                _ <- runtime.close()
+                                _ <- Task.eval(dbDir.recursivelyDelete())
+                              } yield (res)
+                            }.attempt)
+      (result, costLog) = costsLoggingProgram
+      res               <- errorLog.readAndClearErrorVector
+      _                 = Task.now(res should be(Vector.empty))
```
JosDenmark: 😈 This should be either `_  <- Task.now(res should be(Vector.empty))` or `_  = res should be(Vector.empty)`. As it stands, the assertion is never evaluated.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: I'm going to add one more case, but I looking forward to testing this with the par generator that is currently being implemented

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +72,98 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+                             } |
+                             loop!(10)
+                           }""".stripMargin
+
+    val (result, costLog) = evaluateWithCostLog(initialPhlo, contract)
+    result shouldBe Left(OutOfPhlogistonsError)
+    val costs = costLog.map(_.value).toList
+    // The sum of all costs but last needs to be <= initialPhlo, otherwise
+    // the last cost should have not been logged
+    costs.init.sum.toLong should be <= (initialPhlo)
```
JosDenmark: So, basically, by checking that the total cost of all the operations except the last are less than the phlo limit, we can ensure that the last operation was the operation to exceed the phlo limit. Since no more costs were logged after it, we know that all execution branches were terminated?

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -145,25 +157,16 @@ class CostAccountingSpec extends FlatSpec with Matchers {
...
-    val (result, costLog) = evaluateWithCostLog(initialPhlo, contract)
-    result shouldBe Left(OutOfPhlogistonsError)
-    val costs = costLog.map(_.value).toList
-    // The sum of all costs but last needs to be <= initialPhlo, otherwise
-    // the last cost should have not been logged
-    costs.init.sum.toLong should be <= (initialPhlo)
-    // The sum of ALL costs needs to be > initialPhlo, otherwise an error
-    // should not have been reported
-    costs.sum.toLong should be > (initialPhlo)
+    check(forAllNoShrink(Gen.choose(1L, shortslow.cost - 1)) { initialPhlo =>
```
JosDenmark: So, we should generalize this test to take an arbitrary contract rather than just `shortslow`. For `n` processes running in parallel, ensure that  all terminate when one of them runs out of phlo.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Good catch! Cheers

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Yes, the end goal here is to use the par generator (when it's ready)

ArturGajowy: as discussed offline, we don't need to wait for its readiness. We can generate n processes by choosing (with repetitions) from the k predefined processes.

ArturGajowy: I think the tests we have so far are insufficient for testing actual parallel execution and short-circuiting in such an environment. I also think we won't be able to test that reliably without controlling the par shape manually. Change my mind.

lukasz-golebiewski: You're commenting an outdated version so I'm not sure whether you've seen the latest changes. The 3 contracts used in the latest version have some degree of parallelism. You claim that this is not enough for actual testing. What is enough?

The tests are going to be executed multiple times for each contract (with different max phlo). This means the contracts shouldn't be too big, otherwise the tests will take ages to execute. The fact that each test is going to be executed many times gives a good chance of finding concurrency related problems

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -145,25 +157,16 @@ class CostAccountingSpec extends FlatSpec with Matchers {
...
+      val (result, costLog) = evaluateWithCostLog(initialPhlo, shortslow.contract)
+      result shouldBe Left(OutOfPhlogistonsError)
+      val costs = costLog.map(_.value).toList
+      // The sum of all costs but last needs to be <= initialPhlo, otherwise
```
tymm: Do you mean "at least" here?

lukasz-golebiewski: No I meant "all but last"

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Actually, when I change the assertion to fail `      _                 = Task.now(res should not be (Vector.empty))` it does fail:
```
[info] Running out of phlogistons
[info] - should stop the evaluation upon cost depletion in a single execution branch *** FAILED *** (2 seconds, 572 milliseconds)
[info]   Vector() was equal to Vector() (CostAccountingSpec.scala:117)
```

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -63,4 +78,109 @@ class CostAccountingSpec extends FlatSpec with TripleEqualsSupport {
...
+          implicit val ca = costAlg
+          implicit val c  = cost
+          for {
+            runtime <- Runtime
+                        .create[Task, Task.Par](dbDir, size, StoreType.LMDB)
+            par <- Interpreter[Task].buildNormalizedTerm(contract)
+            res <- Interpreter[Task]
+                    .evaluate(runtime, par, Cost(initialPhlo.toLong))
+            _ <- runtime.close()
+            _ <- Task.eval(dbDir.recursivelyDelete())
```
ArturGajowy: It has now occurred to me, that we might be testing a code path that's not relevant to production, simply because the production uses `RuntimeManager.injAttempt`. Especially that I remember us discussing problems with reproducing the right behaviour using Runtime and Reducer.inj. I think what we should do, is expose `injAttempt` and use it here. I'd hate this to become a case of '2 unit tests, 0 integration tests'.

lukasz-golebiewski: Your concern shows an issue with the possibility of the RuntimeManager and Interpreter functionalities not being aligned. The correct solution of this problem would be to extract the common functionality of both and remove all doubt, but since this is only loosely related to testing of cost accounting I would suggest doing this in a separate PR as this one is definitely going to bloat.

ArturGajowy: Fine. Just make sure that happens before you stop working on those tests.

I've got to tell you I'm already very wary of the 'let's do that in a following PR mantra'. Since we're already going to have a follow-up for this once @JosDenmark's work lands, so be it.

ArturGajowy: Please also mind that 'correct solution of this problem would be to extract the common functionality of both and remove all doubt' would mean 'remove all traces of injAttempt from RuntimeManager'. Otherwsie we need to test via injAttempt.

lukasz-golebiewski: The "let's do that in a following PR mantra" is inevitable when working in areas of code with significant amount of debt. When the true fix for the newly uncovered and loosely related problem might cause the PR to bloat I would rather have it implemented separately instead of mixing concerns. This is only an opinion though, yours might be different.

PR #2219, opened 2019-03-01
Remove usage of BlockStore from justificationRegressions

KentShikama:
## Overview

### JIRA ticket:

### Notes


KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -483,74 +484,95 @@ object Validate {
...
+      maybePreviousBlockJustification <- dag.lookup(previousBlockJustificationHash)
+      previousBlockJustification <- maybePreviousBlockJustification match {
+                                     case Some(block) => block.pure[F]
+                                     case None =>
+                                       Sync[F].raiseError[BlockMetadata](
+                                         new Exception(
+                                           s"Missing ${PrettyPrinter.buildString(previousBlockJustificationHash)} from block dag store."
+                                         )
+                                       )
+                                   }
```
dzajkowski: maybe deduplicate?

dzajkowski:
what's concerning to me is that no test fails with this change

KentShikama:
> what's concerning to me is that no test fails with this change

Why should they? It is just a refactoring change.

KentShikama:
The only difference is instead of looking up in the BlockStore we lookup in the block dag store.

casper/src/main/scala/coop/rchain/casper/Validate.scala
itegulov: I agree. You have used this pattern a few times in the previous PRs. Maybe it would make sense to create a utility method just like `ProtoUtil.unsafeGetBlock` but using block DAG storage instead of block storage (or even change the `ProtoUtil.unsafeGetBlock` itself).

KentShikama: Yeah I was thinking I would eventually push the unsafeGetBlock to the block dag store level.

KentShikama:
bors try

dzajkowski:
the merge commit can be removed by rebasing

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -500,74 +500,95 @@ object Validate {
...
-                                        }
-      status <- if (containsJustificationRegression) {
-                 for {
-                   _ <- Log[F].warn(ignore(b, "block contains justification regressions."))
-                 } yield Left(JustificationRegression)
-               } else {
-                 Applicative[F].pure(Right(Valid))
-               }
-    } yield status
+    currentLatestMessages.toList.tailRecM {
```
dzajkowski: this is very hard to read

KentShikama: Hmm that is unfortunate. I tried really hard to somehow not duplicate `val previousBlockJustificationHash` so that I can include the value in the log message. I suppose I should revert this back and just duplicate the value?

KentShikama:
bors r+

I'll clean up in a follow up PR shortly. I want to get the debug message in.

PR #2222, opened 2019-03-02
Make estimator return block hashes instead of full block 

KentShikama:
## Overview

Some calls to the estimator don't require the full BlockMessage
to be returned. Thus we should only return the hashes and
let the caller fetch the BlockMessage based on those hashes
if they need to.

This also removes the BlockStore effect from the Estimator.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3069

### Notes


KentShikama:
bors try

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -128,17 +121,14 @@ object Estimator {
...
-                    }
-                  } yield result
+                  blockDag.lookup(childHash).map {
+                    case Some(childBlock)
+                        if childBlock.parents.size > 1 && childBlock.sender != validator =>
+                      val currScore       = acc2.getOrElse(childHash, 0L)
+                      val validatorWeight = childBlock.weightMap.getOrElse(validator, 0L)
+                      acc2.updated(childHash, currScore + validatorWeight)
+                    case _ => acc2
+                  }
```
odeac: My head hurts when I try to understand this function. Please make it more readable.

KentShikama: The change is NFC. Hopefully the after version is better than the before. If not I'll just revert this entirely.

KentShikama: I've also just considered removing this part. It was initially an idea to make the estimator a bit more fair but for the amount of thinking required verify its correctness I don't think it is worth it.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -190,17 +180,16 @@ object Estimator {
...
+    blockDag.children(b).map {
+      case Some(children) =>
+        val scoredChildren = children.filter(scores.contains)
+        if (scoredChildren.nonEmpty) {
+          scoredChildren.toList
+        } else {
+          List(b)
+        }
+      case None => List(b)
+    }
```
odeac: Trying to understand what happens here I got to the following refactoring which I find more readable (but I don't have too strong opinions):

```scala
 private  def toNonEmptyList[T](elements  :Set[T]) : Option[List[T]] =
    if (elements.isEmpty) None
    else Some(elements.toList)

  private def replaceBlockHashWithChildren[F[_]: Monad](
      b: BlockHash,
      scores: Map[BlockHash, Long]
  ): F[List[BlockHash]] =
    blockDag.children(b)
      .map (
       _.flatMap { children => toNonEmptyList(children.filter(scores.contains))}
        .getOrElse(List(b))
      )
```

casper/src/main/scala/coop/rchain/casper/Estimator.scala
KentShikama: On further thought, I'm going to remove this indeed.

casper/src/main/scala/coop/rchain/casper/Estimator.scala
KentShikama: What about

```
  private def toNonEmptyList[T](elements: Set[T]) : Option[List[T]] =
    if (elements.isEmpty) None
    else Some(elements.toList)

  private def replaceBlockHashWithChildren[F[_]: Monad](
                                                         b: BlockHash,
                                                         scores: Map[BlockHash, Long]
                                                       ): F[List[BlockHash]] =
    blockDag.children(b).map {
      case Some(children) =>
        val scoredChildren = children.filter(scores.contains)
        toNonEmptyList(scoredChildren).getOrElse(List(b))
      case None =>
        List(b)
    }
```

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -320,21 +320,27 @@ object ProtoUtil {
...
-            acc.pure[F]
-          )
-      }
-      .map(_.reverse)
+    for {
+      blocks <- blockHashes.toList.traverse(hash => ProtoUtil.unsafeGetBlock[F](hash))
+      result <- blocks
+                 .foldM(List.empty[BlockMessage]) {
+                   case (acc, b) =>
+                     acc.forallM(nonConflicting(b)).map { isNonConflictingWithAll =>
```
dzajkowski: the previous version seemed easier to read. why the change?

KentShikama: I thought this version was easier to read; the Monad[F].ifM syntax just doesn't read cleanly to me.

odeac: I agree with Dom that the previuos was cleaner. And even if ifM is difficult to read, because of the fact that it's a standard operator it is easy to learn once and then expect the same behaviour everywhere

KentShikama: Ok I'll revert to using ifM.

KentShikama:
bors try

PR #2224, opened 2019-03-04
Limit deploy history duplication check depth

KentShikama:
## Overview

We currently prevent replay attacks by traversing the entire ancestor chain to figure out whether there is a duplicate transaction. If there is no duplicate transaction with the same (user, timestamp) pair, then we are safe. However, this process becomes computationally expensive as the dag grows.

Thus to improve performance, we make the following change. In each transaction, we add a new field: valid after block number. Each transaction must be included in a block with a block number that is greater than or equal to the valid after block number value. We also check that each transaction in the about-to-be-proposed block has a valid after block number that is less than the block number minus some expiration threshold. This allows us to not have to traverse back more blocks than the expiration threshold to check for duplicate transactions. If there was a block that contains a transaction that is in a block that is further back in the past than the expiration threshold compared to the current block, that transaction cannot be duplicated in the current block as it would be expired (not be valid).

Note that we are using block numbers as a really rough proxy of time. We assume each block contains some minimum amount of gas executed and transaction fee spent, effectively giving us a guarantee that the block took some non-zero time and work to create.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3070

### Notes


KentShikama:
bors try

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -75,6 +75,7 @@ message DeployData {
   int64 phloPrice     = 7; //phlo price
   int64 phloLimit     = 8; //phlo limit for the deployment
   int32 nonce         = 9; //nonce for transaction made against `from` wallet
+  int64 validAfterBlockNumber = 10;
```
tymm: The deploy is already valid for that block number, right? If so, the naming might be a little confusing.

KentShikama: Hmm but `validAfterOrOnBlockNumber` is a bit long. I think I'll just change the code so it only validates if it is strictly after.

casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
```diff
@@ -45,6 +45,8 @@ final case object InvalidTransaction      extends InvalidBlock with Slashable
 final case object InvalidBondsCache       extends InvalidBlock with Slashable
 final case object InvalidBlockHash        extends InvalidBlock with Slashable
 final case object InvalidDeployCount      extends InvalidBlock with Slashable
+final case object ContainsExpiredDeploy   extends InvalidBlock with Slashable
+final case object ContainsFutureDeploy    extends InvalidBlock with Slashable
```
tymm: Does that mean it's a slashable offense for a validator to include a deploy that's already in the log? Wouldn't that be an attack vector? I'm thinking about a situation where an evil validator tries to front-run a good validator by getting a certain deploy in before.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -243,33 +252,36 @@ object Validate {
...
+                              .exists(d => deployKeySet.contains((d.user, d.timestamp)))
+                          }
       result <- duplicatedBlock match {
                  case Some(b) =>
                    for {
                      _ <- Log[F].warn(
                            ignore(
                              block,
-                             s"found deploy by the same (user, millisecond timestamp) produced in the block(${b.blockHash})"
+                             s"found deploy by the same (user, millisecond timestamp) produced in the block ${b.blockHash}"
```
odeac: Instead of this `for-comprehension` you can do this:
```scala
                   Log[F].warn(
                           ignore(
                             block,
                             s"found deploy by the same (user, millisecond timestamp) produced in the block ${b.blockHash}"
                           )
                         ) *> Left(InvalidRepeatDeploy).pure
```

The operator *> produces the effects from its both sides and returns the value on its right side.

Even more, where possible, I try to avoid the matches because they are error prone. For this reason I would replace the whole match like this:

```scala
      result <- duplicatedBlock
                 .traverse( b =>
                     Log[F].warn(
                       ignore(
                         block,
                         s"found deploy by the same (user, millisecond timestamp) produced in the block ${b.blockHash}"
                       )
                     )
                     *> InvalidRepeatDeploy.pure[F]
                 )
    } yield result.toLeft(Valid)
```

casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
tymm: Nevermind, I see it now.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -348,6 +360,45 @@ object Validate {
...
+                ignore(
+                  b,
+                  s"block contains an future deploy with valid after block number of ${futureDeploy.validAfterBlockNumber}: ${futureDeploy.term}"
+                )
+              )
+        } yield Left(ContainsFutureDeploy)
+      case None =>
+        Applicative[F].pure(Right(Valid))
+    }
+  }
```
odeac: same comment as above

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -348,6 +360,45 @@ object Validate {
...
+          _ <- Log[F].warn(
+                ignore(
+                  b,
+                  s"block contains an expired deploy with valid after block number of ${expiredDeploy.validAfterBlockNumber}: ${expiredDeploy.term}"
+                )
+              )
+        } yield Left(ContainsExpiredDeploy)
+      case None =>
+        Applicative[F].pure(Right(Valid))
+    }
```
odeac: same

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -281,6 +281,51 @@ class ValidateTest
...
+        deployData        = deploy.deploy.get
+        updatedDeployData = deployData.withValidAfterBlockNumber(Long.MaxValue)
+        blockWithFutureDeploy <- createBlock[Task](
+                                  Seq.empty[BlockHash],
+                                  deploys = Seq(deploy.withDeploy(updatedDeployData))
+                                )
+        status2 <- Validate.futureTransaction[Task](blockWithFutureDeploy)
+        _       = status2 should be(Left(ContainsFutureDeploy))
+      } yield ()
+  }
```
odeac: This test can be split in two: one for valid scenario and one for the invalid one

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -281,6 +281,51 @@ class ValidateTest
...
+        updatedDeployData = deployData.withValidAfterBlockNumber(Long.MinValue)
+        blockWithExpiredDeploy <- createBlock[Task](
+                                   Seq.empty[BlockHash],
+                                   deploys = Seq(deploy.withDeploy(updatedDeployData))
+                                 )
+        status2 <- Validate
+                    .transactionExpiration[Task](blockWithExpiredDeploy, expirationThreshold = 10)
+        _ = status2 should be(Left(ContainsExpiredDeploy))
+      } yield ()
+  }
```
odeac: same as above

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -212,29 +214,40 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Capture: ConnectionsCell: Tr
...
-      deploy = d.flatten
-      result = hist -- deploy
-    } yield (result.toSeq)
+      state               <- Cell[F, CasperState].read
+      currentBlockNumber  = maxBlockNumber + 1
+      earliestBlockNumber = currentBlockNumber - expirationThreshold
+      deploys             = state.deployHistory
+      validDeploys = deploys.filter(
+        d =>
+          d.validAfterBlockNumber <= currentBlockNumber && d.validAfterBlockNumber >= earliestBlockNumber
```
dzajkowski: maybe naming this check would make it easier to read?

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
bors try

rabbitonweb:
@KentShikama I've added the changed you've asked for

```
./node/target/universal/stage/bin/rnode deploy --help
  (....)
  -v, --valid-after-block-number  <arg>   [TBD]
  -h, --help                              Show help message
```

rabbitonweb:
```
  -v, --valid-after-block-number  <arg>   Set this value to one less than the
                                          current block height: you have 50
                                          blocks to get this transaction into
                                          the chain.

```

KentShikama:
bors try

PR #2225, opened 2019-03-04
Towards long-running Whiteblock tests

adaszko:
This PR just adds logback config files that will be used in whiteblock.  Rest are small updates to the documentation and code formatting.


adaszko:
bors try

PR #2230, opened 2019-03-05
Add BatchInfluxDBReporter

sebtomba:
## Overview
Add BatchInfluxDBReporter

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3075

### Notes
Uses `kamon-influxdb` [reference.conf](https://github.com/kamon-io/kamon-influxdb/blob/master/src/main/resources/reference.conf)

An additional setting is `kamon.influxdb.batch-interval` which defaults to 10 seconds.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -147,7 +147,7 @@ class NodeRuntime private[node] (
 
       _ <- Task.delay {
             Kamon.reconfigure(conf.underlying.withFallback(Kamon.config()))
-            if (conf.kamon.influxDb) Kamon.addReporter(new kamon.influxdb.InfluxDBReporter())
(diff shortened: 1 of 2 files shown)


sebtomba:
The code is 90% copy&paste from https://github.com/kamon-io/kamon-influxdb/blob/master/src/main/scala/kamon/influxdb/InfluxDBReporter.scala

The difference is batching of the data so we can configure Kamon to report metrics every 500ms, collect the data and push it every 10 seconds to InfluxDB.

rabbitonweb:
where is it different from the one that is replaced?

rabbitonweb:
> The difference is batching of the data so we can configure Kamon to report metrics every 500ms, collect the data and push it every 10 seconds to InfluxDB.

ok, i see

tymm:
Since a lot of the functionality is the same as in `InfluxDBReporter`, why not just extend from `InfluxDBReporter`?

sebtomba:
@tymm Unfortunately most of the required methods in the original `InfluxDBReporter` are `private` and can't be used in a subclass.

PR #2232, opened 2019-03-05
fix getDeployParams

odeac:
## Overview
fix system process getDeployParams

The implementation is obviously wrong as it should fail with invalid argument exception if the isContract doesn't match.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2966

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
@@ -208,14 +208,11 @@ object SystemProcesses {
       ): (Seq[ListParWithRandomAndPhlos], Int) => F[Unit] = {
         case isContractCall(produce, Seq(ack)) =>
           for {
-            parameters <- shortLeashParams.getParams
-            _ <- parameters match {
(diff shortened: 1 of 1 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -212,10 +212,10 @@ object SystemProcesses {
             _ <- parameters match {
                   case ShortLeashParameters(codeHash, phloRate, userId, timestamp) =>
                     produce(Seq(codeHash, phloRate, userId, timestamp), ack)
```
KentShikama: Probably something like

```suggestion
ShortLeashParameters(codeHash, phloRate, userId, timestamp) = parameters
_ <- produce(Seq(codeHash, phloRate, userId, timestamp), ack)
```

Also add a TODO to add a test for the case _ => below

odeac: Actually that can be fixed even nicer, by using the case class fields instead of the match

PR #2239, opened 2019-03-06
Align runtime manager with the interpreter

lukasz-golebiewski:
## Overview
See title and ticket description

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3078

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -34,6 +34,8 @@ trait Interpreter[F[_]] {
 
   def execute(runtime: Runtime[F], reader: Reader): F[Runtime[F]]
 
+  def evaluate(runtime: Runtime[F], term: String, initialPhlo: Cost): F[EvaluateResult]
+
   def evaluatePar(runtime: Runtime[F], par: Par): F[EvaluateResult]
 
   def evaluatePar(runtime: Runtime[F], par: Par, initialPhlo: Cost): F[EvaluateResult]
@@ -84,6 +86,39 @@ object Interpreter {
     def evaluatePar(runtime: Runtime[F], par: Par): F[EvaluateResult] = {
       val initialPhlo = Cost(Integer.MAX_VALUE) //This is OK because evaluate is not called on deploy
       evaluatePar(runtime, par, initialPhlo)
+    }
+
+    def evaluate(runtime: Runtime[F], term: String, initialPhlo: Cost): F[EvaluateResult] = {
+      implicit val rand: Blake2b512Random = Blake2b512Random(128)
+      for {
+        checkpoint <- runtime.space.createCheckpoint()
+        _          <- runtime.reducer.setPhlo(initialPhlo)
+        res        <- injAttempt(runtime, term)
+        _          <- if (res.errors.nonEmpty) runtime.space.reset(checkpoint.root) else F.unit
+      } yield res
+    }
+
+    def injAttempt(
+        runtime: Runtime[F],
+        term: String
+    )(implicit rand: Blake2b512Random): F[EvaluateResult] = {
+      val parsingCost = accounting.parsingCost(term)
+      Interpreter[F].buildNormalizedTerm(term).attempt.flatMap {
+        case Right(parsed) =>
+          for {
+            result    <- runtime.reducer.inj(parsed).attempt
+            phlosLeft <- runtime.reducer.phlo
+            oldErrors <- runtime.errorLog.readAndClearErrorVector()
+            newErrors = result.swap.toSeq.toVector
+            allErrors = oldErrors |+| newErrors
+            cost      = phlosLeft - parsingCost
+          } yield EvaluateResult(cost, allErrors)
+        case Left(error) =>
+          for {
+            phlosLeft <- runtime.reducer.phlo
+            cost      = phlosLeft - parsingCost
+          } yield EvaluateResult(cost, Vector(error))
+      }
     }
 
     private def buildAST(reader: Reader): F[Proc] =
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -38,6 +38,13 @@ trait Interpreter[F[_]] {
 
   def evaluatePar(runtime: Runtime[F], par: Par): F[EvaluateResult]
   def evaluatePar(runtime: Runtime[F], par: Par, initialPhlo: Cost): F[EvaluateResult]
+
+  def injAttempt(
+      reducer: ChargingReducer[F],
+      errorLog: ErrorLog[F],
+      term: String,
+      initialPhlo: Cost
+  )(implicit rand: Blake2b512Random): F[EvaluateResult]
 }
 
 object Interpreter {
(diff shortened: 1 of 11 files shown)


lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -34,6 +34,8 @@ trait Interpreter[F[_]] {
 
   def execute(runtime: Runtime[F], reader: Reader): F[Runtime[F]]
 
+  def evaluate(runtime: Runtime[F], term: String, initialPhlo: Cost): F[EvaluateResult]
```
JosDenmark: This should probably be `evaluateSource` or `evaluateString` since, as you've noted, we have at least three different evaluation sources, i.e., `String`, `Proc`, and `Par`.

rholang/src/test/scala/coop/rchain/rholang/InterpreterSpec.scala
```diff
@@ -112,17 +112,25 @@ class InterpreterSpec extends FlatSpec with Matchers {
     StoragePrinter.prettyPrint(runtime.space.store)
 
   private def success(runtime: Runtime[Task], rho: String): Task[Unit] =
-    execute(runtime, rho).map(_.swap.foreach(error => fail(s"""Execution failed for: $rho
-                                               |Cause:
-                                               |$error""".stripMargin)))
+    execute(runtime, rho).map(
+      res =>
+        assert(
```
dzajkowski: mark todo "should raise instead of assert"?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -60,30 +68,55 @@ object Interpreter {
...
-                 else F.raiseError(new RuntimeException(mkErrorMsg(errors)))
-      } yield result
+    def evaluatePar(runtime: Runtime[F], par: Par): F[EvaluateResult] = {
+      val initialPhlo = Cost.Max
+      evaluatePar(runtime, par, initialPhlo)
+    }
+
+    def evaluatePar(runtime: Runtime[F], par: Par, initialPhlo: Cost): F[EvaluateResult] = {
+      val term = PrettyPrinter().buildString(par)
+      evaluate(runtime, term, initialPhlo)
```
ArturGajowy: we really shouldn't be pretty-printing an already parsed par and parsing it again in evaluate. Not until we have property-based tests checking a `.rho -> Par -> .rho -> Par` roundtrip to arrive at the same Par twice. I'm pretty sure the PrettyPrinter is not robust enough for such usage, even if we take aside the inefficiency this causes.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -60,30 +68,55 @@ object Interpreter {
...
-        checkpoint    <- runtime.space.createCheckpoint()
-        _             <- runtime.reducer.setPhlo(initialPhlo)
-        _             <- runtime.reducer.inj(par)(rand)
-        errors        <- runtime.readAndClearErrorVector()
-        remainingPhlo <- runtime.reducer.phlo
-        cost          = initialPhlo - remainingPhlo
-        _             <- if (errors.nonEmpty) runtime.space.reset(checkpoint.root) else F.unit
-      } yield EvaluateResult(cost, errors)
+        checkpoint <- runtime.space.createCheckpoint()
+        res        <- injAttempt(runtime.reducer, runtime.errorLog, term, initialPhlo)
```
ArturGajowy: I like how you moved 'setPhlo' inside the injAttempt, and are returning EvaluateResult from it. Those are deduplications long due :+1:

rholang/src/test/scala/coop/rchain/rholang/InterpreterSpec.scala
```diff
@@ -112,17 +112,25 @@ class InterpreterSpec extends FlatSpec with Matchers {
     StoragePrinter.prettyPrint(runtime.space.store)
 
   private def success(runtime: Runtime[Task], rho: String): Task[Unit] =
-    execute(runtime, rho).map(_.swap.foreach(error => fail(s"""Execution failed for: $rho
-                                               |Cause:
-                                               |$error""".stripMargin)))
+    execute(runtime, rho).map(
```
ArturGajowy: fwiw, the previous version (using fail) did not print `true was false` and just cut to the interesting part :) No big deal though, will not bikeshed here :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
lukasz-golebiewski: How would you approach .rho file generation?

rholang/src/test/scala/coop/rchain/rholang/InterpreterSpec.scala
lukasz-golebiewski: This behavior was preserved

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
ArturGajowy: Tymm is working on it

ArturGajowy: while he is though, we need to change this - if not for just the inefficiency this introduces

lukasz-golebiewski: This function is used only in tests and both repls so efficiency shouldn't be top concern here

ArturGajowy: the real issue is correctness though. Also, this is just a matter of reversing the delegation order. So I don't see why are we even discussing this.

lukasz-golebiewski: It's not a matter of reversing the delegation order because `injAttempt` expects `term: String`

ArturGajowy: That's exactly my point. Make it expect a Par (and make it private?), and introduce a public wrapper that does the parsing to retain the callsites.

lukasz-golebiewski: I won't be able to calculate the parsing cost from an already parsed term

ArturGajowy: Then (1) pass it from the outside (could be dummy if in test code), or - better yet - (2) replace the usages of evaluatePar with evaluate. All call sites but `CostAccountingPropertyTest` have the unparsed rho as well not so far above them. So please do (1) for `CostAccountingPropertyTest` and (2) for the other places.

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -60,30 +65,45 @@ object Interpreter {
...
+              _         <- reducer.setPhlo(phloAfterParsing)
+              result    <- reducer.inj(parsed).attempt
+              phlosLeft <- reducer.phlo
+              oldErrors <- errorLog.readAndClearErrorVector()
+              newErrors = result.swap.toSeq.toVector
+              allErrors = oldErrors |+| newErrors
+            } yield EvaluateResult(initialPhlo - phlosLeft, allErrors)
+          case Left(error) =>
+            for {
+              phlosLeft <- reducer.phlo
```
JosDenmark: @lukasz-golebiewski I just noticed the value `phlosLeft` isn't used 😢

PR #2251, opened 2019-03-09
RChain-2990: Share the _cost capability with the rspace matcher

JosDenmark:
## Overview
This PR shares standardizes the `_cost[F]` effect across the RSpace matcher and the rest of the interpreter, including the `ChargingRSpace` type. As a result, the `CostAccounting[F]` implicit parameters are eliminated and only used to instantiate `_cost[F]`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2990

### Notes
`ChargingRSpace` used to charge for matching based on the size of the data being matched. This PR eliminates that charge since the complexity of matching data does not influence the space complexity of persistent storage. Now, the only matching charges are those that exist within the matcher itself.


JosDenmark:
bors try

JosDenmark:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpaceTest.scala
```diff
@@ -334,27 +344,41 @@ class ChargingRSpaceTest extends fixture.FlatSpec with TripleEqualsSupport with
...
+      }
 }
 
 object ChargingRSpaceTest {
   type ChargingRSpace = RhoPureSpace[Task]
-  final case class TestFixture(chargingRSpace: ChargingRSpace, costAlg: CostAccounting[Task])
+  final case class TestFixture(chargingRSpace: ChargingRSpace, cost: _cost[Task])
 
   val RSPACE_MATCH_PCOST     = 100L
   val RSPACE_MATCH_COST      = Cost(RSPACE_MATCH_PCOST)
```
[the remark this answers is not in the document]
ArturGajowy: The above are no longer relevant after removing ListParWithRandomAndPhlos

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
```diff
@@ -27,46 +26,43 @@ object implicits {
...
+            case (freeMap, caughtRem) =>
+              val remainderMap = pattern.remainder match {
+                case Some(Var(FreeVar(level))) =>
+                  freeMap + (level -> VectorPar().addExprs(EList(caughtRem.toVector)))
+                case _ => freeMap
+              }
+              ListParWithRandomAndPhlos(
+                toSeq(remainderMap, pattern.freeCount),
+                data.randomState,
+                (init - left).value
```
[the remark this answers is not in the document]
ArturGajowy: I've deleted the `ListParWithRandomAndPhlos`, please have a look at those commits :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -90,18 +85,11 @@ object ChargingRSpace {
         result match {
           case Left(oope) =>
             // if we run out of phlos during the match we have to zero phlos available
-            costAlg.get().flatMap(cost => costAlg.charge(cost)) >> Sync[F].raiseError(oope)
+            cost.get.flatMap(charge[F]) >> Sync[F].raiseError(oope)
```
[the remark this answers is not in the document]
ArturGajowy: there's so much self-contradictory in this line... the cost should be negative already if it's oope, we shouldn't be charging for "all there's left", because there's none already, and we could probably (hopefully?) replace this with just `rethrow(result)`. See my other comment though: https://github.com/rchain/rchain/pull/2251#discussion_r264175656

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Chargeable.scala
```diff
@@ -2,6 +2,9 @@ package coop.rchain.rholang.interpreter.accounting
 
 import coop.rchain.models.{ProtoM, StacksafeMessage}
 
+/* TODO: Make Chargeable instances for requisite rspace type parameters. Then, create an instance of PureRSpace
+         that uses the generic instances, _cost, and _error for a single, charging PureRSpace. */
+
 trait Chargeable[A] {
```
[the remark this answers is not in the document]
ArturGajowy: > (which might be wrong in itself, but let the code admit that first).
I'd say it's good enough though, for sure for now.

rholang/src/main/scala/coop/rchain/rholang/interpreter/ChargingReducer.scala
```diff
@@ -31,13 +31,13 @@ object ChargingReducer {
...
-      costAccounting: CostAccounting[F]
+      cost: _cost[F]
   ): ChargingReducer[F] = new ChargingReducer[F] {
     def phlo: F[Cost] =
-      costAccounting.get()
+      cost.get
 
     def setPhlo(limit: Cost): F[Unit] =
-      costAccounting.set(limit)
+      cost.set(limit)
```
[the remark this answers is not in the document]
goral09: AFAIK in the Tagless Final world this is sometimes called "vertical stacking of interpreters" (vs "horizontal" as in `F[_]: Log: Console`). https://github.com/SystemFw/Befunge-93/blob/master/src/main/scala/semantics/language.scala#L16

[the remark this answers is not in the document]
ArturGajowy: No matter how you call it, it's a useless instance of what you're calling it - sorry :) The example you're referring to provides substantial abstraction and useful buildup without exposing what's used underneath to build it. What we have here is horizontal stacking in poor disguise :)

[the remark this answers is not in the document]
goral09: You should be able to construct an instance of `Reducer` without worrying about cost accounting (if you do then you use `ChargingReducer` that adds this capability). You should be able to test `Reducer` functionality without costs and similarly you should be able to test cost accounting logic without knowing anything about `Par`s. That's btw the reason why there are `substitute*[charge]` methods. 

If all you ever do is interpret `Par`s and charge for them then why not fixate for `private val costLimit = AtomicInteger` that is reachable from everywhere and don't try to write generic interpreter? 

You've done great job with refactoring this mess and you should be able to finally build these abstractions (similarly to what can be found in the link) but you've decided to fix cost accounting and interpreter logic together.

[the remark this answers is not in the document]
ArturGajowy: `able to construct an instance of Reducer without worrying about cost accounting` - Reducer is full of `charge` calls, I don't see what you're suggesting without further substantial refactoring, on a totally orthogonal axis

`test Reducer functionality without costs` - this should be achieved by providing a no-op cost account

`should be able to test cost accounting logic without knowing anything about Pars` - ideally, maybe. The cost should though be a function of (what data is being used, what aspect of its usage we charge for). A Par has different execution, substitution, and storage costs. So Chargable's current API is not enough for that, and at the same time it's obvious its only purpose is being able to Substitute[A] (it's not used elsewhere, effectively). I mean, maybe the intent was different - but intent doesn't compile, and for the abstraction to bring us any value it needs to be used instead of the concretion.

`the reason why there are substitute*[charge] methods.` - as noted, I understand that charging 'inside' Substitute[A] required a Chargable[A]. The problems I have with that is the above + what I mentioned in the other comment + it's being done on the surface of `Substitute` really - so it's cheaper (mental costs -wise) to do that outside of Substitute.

`why not fixate for private val costLimit = AtomicInteger` - we might do that, after having all usages hidden behind the `_cost[F]` capability.

`You've done great job with refactoring this mess and you should be able to finally build these abstractions (similarly to what can be found in the link) but you've decided to fix cost accounting and interpreter logic together.` - contrary to what one might think, the refactorings aren't being made for their own sake. The abstractions you mention, OTOH, don't solve any of our immediate goals, nor are they easily achievable. So we're taking a step 'back' in this axis, just to make a couple steps forward along the other axes, and then maybe retake a step towards an abstract 'charge reason' in a better way :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -29,79 +30,66 @@ object ChargingRSpace {
...
           result: Either[InterpreterError, Option[
             (
                 ContResult[Par, BindPattern, TaggedContinuation],
-                Seq[Result[ListParWithRandomAndPhlos]]
+                Seq[Result[ListParWithRandom]]
             )
           ]]
       ): F[Unit] =
         result match {
           case Left(oope) =>
```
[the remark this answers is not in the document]
JosDenmark: I concur! I think that's going to be the next PR. By moving `E` into `F`, we can actually make `ChargingRSpace` polymorphic over all parameters without much effort.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -182,11 +182,16 @@ object Runtime {
             freeCount = arity
           )
         )
-        val continuation                   = TaggedContinuation(ScalaBodyRef(ref))
-        implicit val MATCH_UNLIMITED_PHLOS = matchListPar(Cost(Integer.MAX_VALUE))
+        val continuation = TaggedContinuation(ScalaBodyRef(ref))
+        val cost0: _cost[F] =
```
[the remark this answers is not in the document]
JosDenmark: So, I did this because I was trying to preserve the original semantics, for some reason thinking reading `MATCH_UNLIMITED_PHLOS` as a `def` and not a `val`. As a `def`, it would provide both `space.install` and `replaySpace.install` with maximum gas. Now, I'm realizing that we don't even need to instantiate new cost accounts with maximum gas. 

Lukasz. A special cost type would require some way to indicate when you're entering and exiting its usage. So, maybe a bracket operation like `def free[A, B](fa: A => F[B]): F[B]` that doesn't charge for computing `fa` would suffice.

To Artur's point, I agree. Ideally, we shouldn't need to provide `Cost(Integer.MAX_VALUE)` to simulate free operations. We should be able to just not charge in the code and pass `_cost[F]` implicitly. The only problem with doing this for `install` is that it eventually calls `m.get(pattern, matchCandidate)`, which costs something. I'll have to think on this for a longer term solution.

I'll revise to de-duplicate the cost accounts and change `Cost(Integer.MAX_VALUE)` to `Cost.UNSAFE_UNLIMITED`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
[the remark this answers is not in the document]
JosDenmark: Yes. It's redundant now. I'll refactor to just re-throw the error. That's the best that can be done until `Either` is eliminated from `ISpace`.

JosDenmark:
bors try

JosDenmark:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
```diff
@@ -29,23 +30,26 @@ import cats.implicits._
...
       sequenceNumber: Int
   )(values: Seq[Par], ch: Par): F[Unit] =
     for {
-      produceResult <- space.produce(
-                        ch,
-                        ListParWithRandom(values, rand),
-                        persist = false,
-                        sequenceNumber
-                      )(UNLIMITED_MATCH_PHLO)
+      costAlg <- CostAccounting.of(Cost(0))
```
lukasz-golebiewski: Use the `CostAccounting.empty` alias

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -71,7 +71,7 @@ object Runtime {
       BindPattern,
       InterpreterError,
       ListParWithRandom,
-      ListParWithRandomAndPhlos,
+      ListParWithRandom,
```
lukasz-golebiewski: Just double-checking. Do we need 2x `ListParWithRandom` here?

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
```diff
@@ -29,6 +29,14 @@ package object accounting extends Costs {
       override def set(s: Cost): F[Unit]            = state.set(s)
     }
 
+  def noCharge[F[_]: Monad, A](fa: F[A])(implicit cost: _cost[F]): F[A] =
```
lukasz-golebiewski: Wouldn't it make more sense to instantiate a one-off `_cost` instance here instead?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
ArturGajowy: Technically, we could probably delete one of the holes in here, and in RSpace. I'd wait with that though.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
lukasz-golebiewski: I fear that the code below isn't thread-safe

ArturGajowy: That's a very interesting idea! This would also cause the `charge`-s conducted in those blocks to not show up in the upstream cost log. Which is desirable, I think.

ArturGajowy: We'd have to make sure though that the newly-spawned _cost instance is being used actaully. Functions have no implicit parameters, so I guess we'd have an explicit one and define an `implicit val` in the function block?

lukasz-golebiewski: Another idea that comes to my mind would be to instantiate here a `noOpCost` which would never go below zero and thus never raise errors

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
```diff
@@ -29,24 +30,26 @@ import cats.implicits._
...
     for {
-      produceResult <- space.produce(
-                        ch,
-                        ListParWithRandom(values, rand),
-                        persist = false,
-                        sequenceNumber
-                      )(matchListPar)
+      costAlg <- CostAccounting.of(Cost(0))
+      cost    = loggingCost(costAlg, noOpCostLog[F])
+      produceResult <- noCharge {
```
ArturGajowy: having 
```scala
noCharge { costAlg =>
  implicit val cost = loggingCost(costAlg, noOpCostLog[F])
  ...
}
```
we could make it clear that the only part doing any cost accounting is inside `noCharge`, and have one line of code less :sunglasses: (not counting braces :))

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -58,7 +58,9 @@ object ChargingRSpace {
           patterns: Seq[BindPattern],
           continuation: TaggedContinuation
       ): F[Option[(TaggedContinuation, Seq[ListParWithRandom])]] =
-        space.install(channels, patterns, continuation)
+        noCharge {
+          space.install(channels, patterns, continuation)
+        }
```
ArturGajowy: this looks a bit redundant, but it doesn't matter too much. On one hand I understand the intent to show it's not doing any charging, OTOH the compiler doesn't care... Let's leave it though :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -425,27 +419,30 @@ object Runtime {
...
+          )(matchListPar)
+        }, noCharge {
+          replaySpace.produce(
+            Registry.registryRoot,
+            ListParWithRandom(Seq(Registry.emptyMap), rand),
+            false,
+            0
+          )(matchListPar)
+        })
+      }
```
ArturGajowy: maybe we could have 
```scala
_ <- noCharge { costAlg =>
  implicit val cost0: _cost[F] = loggingCost(costAlg, noOpCostLog)
  space.produce(...)
}
_ <- noCharge { costAlg =>
  implicit val cost0: _cost[F] = loggingCost(costAlg, noOpCostLog)
  replaySpace.produce(...)
}
```
here?

Since we don't charge anyway, not-sharing the cost account between the not-charging parts makes sense - especially given the boilerplate spared...

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
ArturGajowy: No matter what we spawn, I'd do that inside the `def noCharge`, as it seems to clean-up some boilerplate. See my other comments :)

ArturGajowy: I'd also say a throw-away local-to-noCharge CostAccount.of(Cost.Max) is good enough :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Chargeable.scala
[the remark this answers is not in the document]
ArturGajowy: How do we address the discrepancies between that TODO and my comment above? :)
Can we agree to elliminate Chargeable in a future PR?
Am I right we agree to rename PureRspace to ChargingRspace and drop the non-charging PureRspace intacne in a future PR?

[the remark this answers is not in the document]
ArturGajowy: I think my comment here got lost, re posting:

how do we address the discrepancies between that TODO comment and my comment above?

Can we agree that Chargable should be elliminated in a future PR?

Do we agree that PureRSpace should be renamed to ChargingRSpace and its non-chargin instance should be dropped?

Do we agree that we're dropping CharingReducer in an upcoming PR?

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
JosDenmark: Semantically, they're equivalent, but I think we should aim for a unified `_cost` instance. This makes future changes less finicky because we won't have to worry about synchronizing `_cost` instances. It has the added benefit of using less memory. It uses one instance of `_cost` rather than 1 + (# of `noCharge` calls) instances.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
JosDenmark: No. Good catch.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Chargeable.scala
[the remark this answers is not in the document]
JosDenmark: @ArturGajowy I didn't see this! So, I think it's valuable to have a parametrically polymorphic `ChargingRSpace`. For that, we still need `Chargeable`. I also want to eliminate all of the boiler-plate delegating calls to `ISpace`, and to do that we need a smaller store interface that `ISpace`. Maybe something that only has `install`, `produce`, and `consume`?

I agree that PureRSpace should be removed.

I agree we're dropping ChargingReducer in an upcoming PR.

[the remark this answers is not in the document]
ArturGajowy: >  parametrically polymorphic ChargingRSpace. For that, we still need Chargeable

We don't :) PureRSpace (to be ChargingRSpace) is already polymorphic in terms of what it stores (and thus: charges for). Unless you want to make the ChargingRSpace **instance** polymorphic with regard to charged-for things. It doesn't make sense though, as even instances of RSpace know what they're storing, and similarly the PureRSpace instance in `ChargingRSpace.scala:32` knows what it's charging for.

> eliminate all of the boiler-plate delegating calls to ISpace

I'd say the instance at `ChargingRSpace.scala:32` is the place to do that. PureRSpace already has the minimal interface needed by the interpreter. If by 'boilerplate' you mean `handleResult` in that instance, all the noise will disappear once you ellimiate `Either` and `E` from RSpace :)

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Chargeable.scala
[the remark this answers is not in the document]
JosDenmark: > We don't :) PureRSpace (to be ChargingRSpace) is already polymorphic in terms of what it stores (and thus: charges for)

We do. We're talking about the cost of storage. A polymorphic `ChargingRSpace` can't know how to charge for polymorphic types unless we tell it how. I don't see what the big difference is between using `Chargeable` and using `Match[F, P, A, R` (a single op, lawless typeclass) in `RSpace`. It's the same in principle. I agree with you that `PureRSpace` can go. In the PR I do that, I'll show you the `ChargingRSpace` instance and we can revisit.

> I'd say the instance at ChargingRSpace.scala:32 is the place to do that. PureRSpace already has the minimal interface needed by the interpreter.

I mean the boilerplate of `createCheckpoint`, `reset`, and `close`, which only call the same methods on the underlying `ISpace` instance.

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
JosDenmark: I like that. I'd rather make those changes in a subsequent PR though. I'd like to get a `noCharge` operation that didn't require instantiating new instances of `CostAccounting` and `_cost` 🤔

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
JosDenmark: I'm probably going to just `Cost.UNSAFE_MAX` for now and write `noCharge` into a subsequent PR 👍

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
JosDenmark: Outdated

JosDenmark:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
lukasz-golebiewski: How is this outdated?

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Chargeable.scala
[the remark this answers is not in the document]
ArturGajowy: So you're saying the instance should be polymorphic with the data stored. I don't see that - not without having a `Serializable[A]` typeclass required as well. Re: 'Match is no better' - having a broken window doesn't mandate breaking more of them. That being said,

> I'll show you the ChargingRSpace instance and we can revisit.

sounds good :) My only request is that you rename PureRspace to ChargingRspace first (and remove the non-charging instance), and only then make a commit making it polymorphic using Chargable. This way we'll see the outcomes clearly.

> I mean the boilerplate of createCheckpoint, reset, and close

We can think of a finer-grained interface, but I don't really see an issue with having them on ChargingRspace.

JosDenmark:
bors try

PR #2255, opened 2019-03-11
Rchain 3088: implement Lockbox

odeac:
## Overview
Implement Lockbox contract

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3088

### Notes

### DIFF
--- a/casper/src/main/resources/Lockbox.rho
+++ b/casper/src/main/resources/Lockbox.rho
@@ -1,0 +1,55 @@
+//scalapackage coop.rchain.rholang
+
+//Registry info:
+//  sk: 78623a45fb0de7c4bc35c7099e8037c09721fff814e58d419b0e1b21f5a630bb
+//  pk: 7e59141cbc0dfb24ca0e4fe57666895190740778871744be3cec262147e02f56
+//  user == pk
+//  timestamp: 1552131653401
+//Resulting unforgable name: Unforgeable(0xbcfcd4f7ab0b4e1f6c1018f38c42d9721180b9fc92ba4bbe8f00b5010513f80e)
+//  ==> signature data == 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20d7da018b7065def0035d962198cf1581a6aae7509aff8ca75a4035ac5a2ce49c1001
+//  ==> signature == 5af8389f2e933cc7e3e701dbf53b75ac0a3ba05265a3fb166f010cf3b301a4837351f0ec752b334620f7efd56b9971ebedd9e6c137efcefe2bfa44627d866d0c
+//URI derived from pk == `rho:id:9dsr55z1js13x346yhhecx66ns486i3yqf6jafrd9p9hdrrbxjqmyu`
+new
+  Lockbox,
+  rs(`rho:registry:insertSigned:ed25519`),
+  uriOut,
+  getDeployData(`rho:deploy:params`)
+in {
+  contract Lockbox(@"lock", channel, retCh) = {
+    new privateCh, locker, getIdentity in {
+      contract getIdentity(retCh) = {
+        new ch in {
+          getDeployData ! (*ch) |
+          for (@_,@_,@userId, @_ <- ch) {
+            retCh ! (userId)
+          }
+        }
+      } |
+
+      getIdentity!(*privateCh) |
+      for(@deployer <- privateCh) {
+        contract locker(@"unlock", retCh) = {
+          new callerCh in {
+            getIdentity!(*callerCh) |
+            for(@caller <- callerCh) {
+              if (caller == deployer) {
+                retCh ! (*channel)
+              }
+              else {
+                retCh ! (Nil)
+              }
+            }
+          }
+        } |
+
+        retCh ! (bundle+{*locker})
+      }
+    }
+  } |
+  rs!(
+    "7e59141cbc0dfb24ca0e4fe57666895190740778871744be3cec262147e02f56".hexToBytes(),
+    (9223372036854775807, bundle+{*Lockbox}),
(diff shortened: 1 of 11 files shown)


casper/src/main/resources/Lockbox.rho
```diff
@@ -0,0 +1,55 @@
...
+          getDeployData ! (*ch) |
+          for (@(_,_,userId, _) <- ch) {
+            retCh ! (userId)
+          }
+        }
+      } |
+
+      getIdentity!(*privateCh) |
+      for(@deployer <- privateCh) {
+        contract locker(retCh) = {
```
ArturGajowy: Let's have `contract locker(@"unlcok", retCh)`, so that it's more clear on each call that care should be taken not to leak the returned result

ArturGajowy: Also, it's just struck me: we should probably make names consistent - either `lockbox` or `locker`. Either is fine for me - let's just use one of them.

odeac: I've added method names lock/unlock for clarity

casper/src/test/scala/coop/rchain/casper/helper/DeployDataContract.scala
```diff
@@ -0,0 +1,41 @@
...
+  )(message: (Seq[ListParWithRandomAndPhlos], Int)): F[Unit] = {
+
+    val isContractCall = new ContractCall(ctx.space, ctx.dispatcher)
+    message match {
+      case isContractCall(
+          produce,
+          Seq(RhoType.String("userId"), (pk @ RhoType.ByteArray(_)), ackCh)
+          ) =>
+        for {
+          params    <- ctx.shortLeashParams.getParams
```
JosDenmark: So, `set` followed by `get` on `Ref` is not atomic. You have to use `getAndSet` or `modify` for atomicity, otherwise you introduce race conditions.

odeac: This is only used in testing but that's still a very good point. I'll fix this asap.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -84,7 +84,10 @@ object Runtime {
       private val params: Ref[F, ShortLeashParameters]
   ) {
     def setParams(codeHash: Par, phloRate: Par, userId: Par, timestamp: Par): F[Unit] =
-      params.set(ShortLeashParameters(codeHash, phloRate, userId, timestamp))
+      setParams(ShortLeashParameters(codeHash, phloRate, userId, timestamp))
+
+    def setParams(parameters: ShortLeashParameters): F[Unit] =
```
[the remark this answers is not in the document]
odeac: There is already a usage of setParams  :(Par,Par,Par,Par) => F[Unit] and I didn't want to change those call sites. n the other hand getParams returns F[ShortLeashParameters] so, for symmetry I added this version of setParams.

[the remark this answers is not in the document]
odeac: Anyway, I remove this and replaced it with an updateFunction as requested in your other comment

casper/src/main/resources/Lockbox.rho
```diff
@@ -0,0 +1,56 @@
...
+in {
+  contract Lockbox(@"lock", @ownerPubKeyBytes, protected, ret) = {
+    new lockbox, getIdentity in {
+
+      contract getIdentity(retCh) = {
+        new ch in {
+          getDeployData ! (*ch) |
+          for (@_,@_,@userId, @_ <- ch) {
+            retCh ! (userId)
+          }
```
JoshOrndorff: This is equivalent to ethereum's msg.sender. I would prefer not to have that available everywhere in the rholang. I'm not opposed to it happening "at the door" though.

JoshOrndorff: This is equivalent to ethereum's msg.sender. I would prefer not to have that available everywhere in the rholang. I'm not opposed to it happening "at the door" though.

ArturGajowy: What's wrong with msg.sender? Why would you prefer to not have it available everywhere?

I imagine the Lockbox to be a basic building block for safe, private storage of unforgable names / OCAPs. One could trivially implement a PO Box contract based on that, and/or a 'private locker' (basically a map in a lockbox).

JoshOrndorff: The trouble with `msg.sender` is that it allows users to work _around_ ocap discipline and instead implement access control lists. In ethereum it is ubiquitous to have a ACL of length one, the so-called "owner".

There are a few discussions in pub-rchain#wallet about this.

I'm not claiming this is a disastrous design decision, but I think if we make it, we should stop promoting rholang as "enforcing ocap discipline" and maybe pivot to "permitting ocap discipline".

Here's a design I proposed on discord
<blockquote>
I'm imagining a phase that happens between when the user sends new code to a validator, and when that code gets parred into the tuplespace that would be the "door" phase. The workings are not conceptually hard. Anytime code contains rho:io:stdout or rho:lockbox:<some pubkey> or whatever, the node checks to see whether the deploying pubkey (msg.sender) has access to those channels. If so the new names are instantiated to the correct unforgeable IDs. If not... I'm open to suggestions. They are just arbitrary new names? Deploy fails?
So how do we know whether a particular deployer "has access" to a given capability? If it's in the registry or their, then eveyone has access to it.  There is also a lockbox/safe/locker associated with each pubkey (implicitly empty until first use) only accessible in this "door" phase. If a capability is in the deployers locker, they have access to it.
</blockquote>

ArturGajowy: I see. I think we should revisit this before mainnet, it might make sense to stick to ocap-only. It's a bit less user friendly though (it's easier to leak/lose track of an ocap than a public key).

Funny though how this was introduced to implement safe ocap storage in the first place (a lockbox is a safe/locker of size 1, and when discussing this change I had a safe in mind as well).

PR #2212, opened 2019-02-27
Remove use of BlockStore from Validate.sequenceNumber

KentShikama:
## Overview
<sup>_What this PR does, and why it's needed_</sup>

We can make use of the blockMeta's seqNum instead

### JIRA ticket:
<sup>_Create it if there isn't one already._</sup>

https://rchain.atlassian.net/browse/RCHAIN-3046

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2257, opened 2019-03-12
Clean up usages of dag.lookup and Sync[F].raise in Validate

KentShikama:
## Overview

This is a follow up from #2212 

### JIRA ticket:

### Notes


KentShikama:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -193,6 +193,15 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: BlockStore: RaiseIO
...
       dataLookup.get(blockHash).pure[F]
+    def unsafeLookup(blockHash: BlockHash): F[BlockMetadata] =
+      dataLookup.get(blockHash) match {
+        case Some(block) => block.pure[F]
+        case None =>
+          val blockHashString = Base16.encode(blockHash.toByteArray)
+          Sync[F].raiseError[BlockMetadata](
+            new Exception(s"Missing block hash $blockHashString in block dag.")
+          )
+      }
```
odeac: this is the same as the code below but the match is more error prone than using the standard operators like `map` and `getOrElse`

```scala
    def unsafeLookup(blockHash: BlockHash): F[BlockMetadata] =
      dataLookup
        .get(blockHash)
        .map(_.pure[F])
        .getOrElse {
          val blockHashString = Base16.encode(blockHash.toByteArray)
          Sync[F].raiseError[BlockMetadata](
            new Exception(s"Missing block hash $blockHashString in block dag.")
          )
      }
```

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockDagStorage.scala
```diff
@@ -35,6 +35,15 @@ final class InMemBlockDagStorage[F[_]: Concurrent: Sync: Log: BlockStore](
...
       dataLookup.get(blockHash).pure[F]
+    def unsafeLookup(blockHash: BlockHash): F[BlockMetadata] =
+      dataLookup.get(blockHash) match {
+        case Some(block) => block.pure[F]
+        case None =>
+          val blockHashString = Base16.encode(blockHash.toByteArray)
+          Sync[F].raiseError[BlockMetadata](
+            new Exception(s"Missing block hash $blockHashString in block dag.")
+          )
+      }
```
odeac: same as above  + the copy+paste looks bad

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -569,26 +551,8 @@ object Validate {
...
-                                     case Some(block) => block.pure[F]
-                                     case None =>
-                                       Sync[F].raiseError[BlockMetadata](
-                                         new Exception(
-                                           s"Missing ${PrettyPrinter.buildString(previousBlockJustificationHash)} from block dag store."
-                                         )
-                                       )
-                                   }
+      currentBlockJustification  <- dag.unsafeLookup(currentBlockJustificationHash)
+      previousBlockJustification <- dag.unsafeLookup(previousBlockJustificationHash)
```
odeac: The function unsafeLookup seems to be a way to remove this duplication. If that's the case, I don't think it's a good idea. You could easily have a function with a signature like this to eliminate the duplication:
```scala
def failOnNone : F[Option[A]] => F[A]
```

KentShikama: I prefer hiding the fact that it can throw an exception as an implementation detail

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
KentShikama: Hmm I find this code harder to read...

I'm trying to imagine myself as someone who just learned Scala: .map on an Option and _.pure[F] are confusing concepts imo.

KentShikama:
```scala
      dataLookup.get(blockHash) match {
        case Some(block) => block.pure[F]
        case None =>
          val blockHashString = Base16.encode(blockHash.toByteArray)
          Sync[F].raiseError[BlockMetadata](
            new Exception(s"Missing block hash $blockHashString in block dag.")
          )
      }
```

Here is how I read this code:

1. `dataLookup.get(blockHash)` - "Oh we lookup a map"
2. `case Some(block) =>` vs `case None =>` - "Maybe the block exists, if it does, go on with this case. Otherwise go on with the None case"
3. `=> block.pure[F]` - "Ok I have no idea what this .pure[F] thing does but I'm assuming it is some minor modification to block, but I'm assuming the general idea is that when the block exists, I just return it"
4. 
```
          val blockHashString = Base16.encode(blockHash.toByteArray)
          Sync[F].raiseError[BlockMetadata](
            new Exception(s"Missing block hash $blockHashString in block dag.")
          )
``` 
"Ok so there is this weird `Sync[F].raiseError` syntax but from the name I guess it throws some kind of error - and this makes sense, when the blockHash doesn't exist we need to throw an error"

vs

```scala
    def unsafeLookup(blockHash: BlockHash): F[BlockMetadata] =
      dataLookup
        .get(blockHash)
        .map(_.pure[F])
        .getOrElse {
          val blockHashString = Base16.encode(blockHash.toByteArray)
          Sync[F].raiseError[BlockMetadata](
            new Exception(s"Missing block hash $blockHashString in block dag.")
          )
      }
```

1. `dataLookup.get(blockHash)`- "Oh we lookup a map"
2. `.map(_.pure[F])` - "We map over the result? Does it return a list? What is this magical underscore pure thing? Searches for an hour to learn that you can map over Options or any Monad. And then spends another day reading up all the tutorials on Monads."

KentShikama:
Status: Low priority but may need discussion

casper/src/main/scala/coop/rchain/casper/Validate.scala
ArturGajowy: That is never an implementation detail. As soon as you do that you've got no local reasoning and no types can help you. I second @odeac's remarks. No more `unsafe*` stuff should get into our codebase.

ArturGajowy: Sorry I got mislead by the name. I thought we're talking about an effectful function not returning via `F[_]`. I think it's fine to not parameterize with the exception. The name should be changed though, as it causes an instant trigger :) Maybe `tryLookup`?

PR #2260, opened 2019-03-12
RCHAIN-3094 vdag bug fixed

rabbitonweb:
## Overview
Fix vdag bug were more then two blocks were on the same level in single validator

After fix you can expect vdags that look like this:

Screenshot 2019-03-12 at 16 42 53

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3094

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
@@ -20,23 +20,11 @@ final case class ValidatorBlock(
     justifications: List[String]
 )
 
-object ValidatorBlock {
-  implicit val validatorBlockMonoid: Monoid[ValidatorBlock] = new Monoid[ValidatorBlock] {
-    def empty: ValidatorBlock = ValidatorBlock("", List.empty[String], List.empty[String])
-    def combine(vb1: ValidatorBlock, vb2: ValidatorBlock): ValidatorBlock =
-      ValidatorBlock(
-        vb2.blockHash,
-        vb1.parentsHashes ++ vb2.parentsHashes,
-        vb1.justifications ++ vb2.justifications
-      )
-  }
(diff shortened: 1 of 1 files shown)


ArturGajowy:
From what I've understood from your discord chat, it would be great to base the y-coordinate of the block on `sequenceNumber` though?

KentShikama:
@ArturGajowy I think they have their pros and cons. Maybe we could have both. This graph is less "jumpy" than the sequence number version.

ArturGajowy:
You mean there would be just one column per validator (vs many as is in this PR) had we used the sequenceNumber? Maybe we could have it as is (many columns per validator), just with offset based on sequence number?

I obviously leave that all up to you guys though :)

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -168,7 +156,7 @@ object GraphzGenerator {
       validators: Map[String, ValidatorsBlocks]
   ): G[Unit] =
     validators.values.toList
-      .flatMap(_.values.toList)
+      .flatMap(_.values.toList.flatten)
```
odeac: obviously an `extraFlatMap` operator is needed :))

rabbitonweb: hehe :)

KentShikama:
bors r+

https://drone.rchain-dev.tk/rchain/rchain/2195/7

```
[error] java.lang.IllegalArgumentException: requirement failed: Source file '/drone/src/github.com/rchain/rchain/crypto/target/scala-2.12/classes/coop/rchain/crypto/hash/Blake2b512Random$.class' does not exist.
[error] at scala.Predef$.require(Predef.scala:277)
```

seems transient

PR #2262, opened 2019-03-13
RCHAIN-3007 Fix

rabbitonweb:
## Overview
Postpone Casper initialization until comm receives first connection

CasperPacketHandler is seriously the weirdest set of design choices
I've seen in the past few weeks. I'm planning to refactor hell out of
it in incoming days, but for now this commit is focusing on fixing bug
RCHAIN-3007.

After this change, when initializing node (which is not bootstrap),
casper code will not be triggered until comm layer grabs at least
first connection. In 99% of cases the flow of the program will be
following:

1. node boots up
2. node puts bootstrap into its kademlia table
3. node waits for connection to bootstrap node
4. once connection is established:
a) CasperPachketHandler protocol is run (init method)
b) casperloop is triggered

After this change, new node will start catching up with the network
even if nobody is proposing to it (or other nodes) at the given moment

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3007

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
@@ -34,7 +34,7 @@ import coop.rchain.p2p.effects.PacketHandler
 import coop.rchain.shared.{Log, LogSource, Time}
 import monix.eval.Task
 import monix.execution.Scheduler
-
+import coop.rchain.catscontrib.ski._
 import scala.concurrent.duration.FiniteDuration
 import scala.util.Try
 
@@ -125,8 +125,6 @@ object CasperPacketHandler extends CasperPacketHandlerInstances {
                       conf.approveGenesisDuration,
                       conf.approveGenesisInterval
                     )
-            // TOOD this returns a fiber, should we consider cancelling on shutdown?
-            _ <- Concurrent[F].start(abp.run())
             standalone <- Ref.of[F, CasperPacketHandlerInternal[F]](
                            new StandaloneCasperHandler[F](abp)
                          )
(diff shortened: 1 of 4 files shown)


node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -240,10 +240,16 @@ class NodeRuntime private[node] (
 
     val loop: Effect[Unit] =
       for {
-        _ <- time.sleep(1.minute).toEffect
         _ <- dynamicIpCheck.toEffect
         _ <- Connect.clearConnections[Effect]
         _ <- Connect.findAndConnect[Effect](Connect.connect[Effect])
+        _ <- time.sleep(1.minute).toEffect
```
ArturGajowy: :heart: :joy:

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -240,10 +240,16 @@ class NodeRuntime private[node] (
...
+      } yield ()
+
+    def waitForFirstConnetion: Effect[Unit] =
+      for {
+        _ <- time.sleep(10.second).toEffect
+        _ <- ConnectionsCell[Effect].read.map(_.isEmpty).ifM(waitForFirstConnetion, ().pure[Effect])
```
ArturGajowy: shouldn't we check-then-wait here as well?

ArturGajowy: Also, I'd appreciate a shorter check interval - 3 s at most

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
```diff
@@ -125,8 +125,6 @@ object CasperPacketHandler extends CasperPacketHandlerInstances {
                       conf.approveGenesisDuration,
                       conf.approveGenesisInterval
                     )
-            // TOOD this returns a fiber, should we consider cancelling on shutdown?
-            _ <- Concurrent[F].start(abp.run())
```
KentShikama: How come you can remove this?

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
```diff
@@ -148,25 +146,23 @@ object CasperPacketHandler extends CasperPacketHandlerInstances {
...
             validatorId <- ValidatorIdentity.fromConfig[F](conf)
-            bootstrap <- Ref.of[F, CasperPacketHandlerInternal[F]](
+            bootstrap <- Ref.of[F, CasperPacketHandlerInternal[F]] {
                           new BootstrapCasperHandler(
                             runtimeManager,
                             conf.shardId,
                             validatorId,
-                            validators
+                            validators,
+                            CommUtil.requestApprovedBlock[F](delay)
```
KentShikama: So does the requestApprovedBlock now get run inside BootstrapCasperHandler?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
sebtomba: @rabbitonweb
BUT WHY?

PR #2264, opened 2019-03-14
Add gRPC API max message size option

sebtomba:
## Overview
ATM we use server max message size option for both client and server. However, the API client can't use the `--max-message-size` command line option and uses always the default value for this option. Add gRPC API max message command line option `--grpc-max-message-size`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3101

### DIFF
--- a/node/src/main/resources/reference.conf
+++ b/node/src/main/resources/reference.conf
@@ -78,6 +78,9 @@ rnode {
 
     # Port used for external gRPC API
     port-internal = 40402
+
+    # Maximum size of message that can be sent via gRPC API
+    max-message-size = 4M
   }
 
   # Casper configuration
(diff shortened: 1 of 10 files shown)


node/src/main/resources/reference.conf
```diff
@@ -78,6 +78,9 @@ rnode {
 
     # Port used for external gRPC API
     port-internal = 40402
+
+    # Maximum size of message that can be sent via gRPC API
+    max-message-size = 256K
```
KentShikama: Do you not need to delete the max-message-size above (on line 49)?

KentShikama: Ah ok makes sense. In that case can you make the default for API messages 10x the current value?

PR #2265, opened 2019-03-14
RCHAIN-3103 Change Connect to work asynchronously

rabbitonweb:
## Overview
Change Connect to work asynchronously. After this we are **thhhiissss** close to remove `roundTrip`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3103

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
@@ -25,24 +25,24 @@ object HandleMessages {
     Metrics.Source(CommMetricsSource, "rp.handle")
 
   def handle[F[_]: Monad: Sync: Log: Time: Metrics: TransportLayer: ErrorHandler: PacketHandler: ConnectionsCell: RPConfAsk](
-      protocol: Protocol,
-      defaultTimeout: FiniteDuration
+      protocol: Protocol
   ): F[CommunicationResponse] =
     ProtocolHelper.sender(protocol) match {
       case None =>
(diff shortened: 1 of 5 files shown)


rabbitonweb:
This is ready for review

comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
```diff
@@ -74,29 +74,25 @@ object HandleMessages {
...
-      hbrErr       <- TransportLayer[F].roundTrip(peer, ProtocolHelper.heartbeat(local), defaultTimeout)
-      commResponse <- hbrErr.fold(error => notHandledHandshake(error), _ => handledHandshake(local))
-    } yield commResponse
-  }
+      local    <- RPConfAsk[F].reader(_.local)
+      response = ProtocolHelper.protocolHandshakeResponse(local)
+      _        <- TransportLayer[F].send(peer, response) >>= ErrorHandler[F].fromEither
+      _        <- Log[F].info(s"Responded to protocol handshake request from $peer")
+      _        <- ConnectionsCell[F].flatModify(_.addConn[F](peer))
+    } yield handledWithoutMessage
```
KentShikama: If it is async, then shouldn't the return be Unit? Is it because the heartbeat has to remain roundTrip?

rabbitonweb: "handle **without ** message" - Once we remove `roundrripw` the `CommResponse` will be removed, and all those message will return unit

PR #2268, opened 2019-03-15
Wallet contracts

ArturGajowy:
## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -15,7 +15,61 @@
 new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lookup`), ListOpsCh in {
   rl!(`rho:id:dputnspi15oxxnyymjrpu7rkaok3bjkiwq84z7cqcrx4ktqfpyapn4`, *ListOpsCh) |
   for(@(_, ListOps) <- ListOpsCh) {
-    //Right-biased flatMap; f must also return an Either
+
+    // Like fromNillable, except it takes `valueCh` and reads it to obtain the `value` parameter
+    contract Either(@"fromNillable <-", valueCh, @nilValue, return) = {
+      for (@value <- valueCh) {
+        Either!("fromNillable", value, nilValue, *return)
+      }
+    } |
+
+    // Lifts a value into Either, putting Nil-s into Left and using Right otherwise.
+    contract Either(@"fromNillable", @value, @nilValue, return) = {
+      match value {
+        Nil => return!(("Left", nilValue))
+        _   => return!(("Right", value))
+      }
+    } |
+
+    // Like fromBoolean, except it takes `successCh` and reads it to obtain the `success` parameter
+    contract Either(@"fromBoolean <-", successCh, @nilSuccess, return) = {
+      for (@success <- successCh) {
+        Either!("fromBoolean", success, nilSuccess, *return)
+      }
+    } |
+
+    // Lifts a Boolean into Either, making `false` into Left and using Right otherwise.
+    contract Either(@"fromBoolean", @success, @failureMessage, return) = {
+      match success {
+        Nil => return!(("Left", failureMessage))
+        _   => return!(("Right", Nil))
+      }
+    } |
+
+    // Like map2, except it takes `aCh` and `bCh` and reads them to obtain the `a` and `b` parameters
+    contract Either(@"map2 <-", aCh, bCh, f, return) = {
+      for (@a <- aCh; @b <- bCh) {
+        Either!("map2", a, b, *f, *return)
+      }
+    } |
+
+    // Composes two eithers `a: A` and `b: B` using the function `f`. See cats.Apply.map2 for details.`
+    contract Either(@"map2", @a, @b, f, return) = {
+      match (a, b) {
+        (("Left", _), _)               => return!(a)
+        (_, ("Left", _))               => return!(b)
+        (("Right", va), ("Right", vb)) => {
+          new ret in {
+            f!(va, vb, *ret) |
+            for (@v <- ret) {
+              return!(("Right", v))
+            }
+          }
+        }
+      }
+    } |
+
+    // Right-biased flatMap; f must also return an Either
     contract Either(@"flatMap", @either, f, return) = {
       match either {
         ("Right", value) => { f!(value, *return) }
@@ -23,7 +77,14 @@ new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lo
       }
     } |
 
-    //Right-biased map to transform an Either
+    // Like map, except it takes `eitherCh` and reads them to obtain the `either` parameter
+    contract Either(@"map <-", eitherCh, f, return) = {
+      for (@either <- eitherCh) {
+        Either!("map", either, *f, *return)
+      }
+    } |
+
+    // Right-biased map to transform an Either
     contract Either(@"map", @either, f, return) = {
       match either {
         ("Right", value) => {
@@ -38,14 +99,14 @@ new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lo
       }
     } |
 
-    //Composes a list of functions returning Either into a
-    //single result by using flatMap. Similar to following Scala code:
-    //for {
-    //  r0 <- Right(input)
-    //  r1 <- f1(r0)
-    //  r2 <- f2(r1)
-    //  ...
-    //} yield fn
+    // Composes a list of functions returning Either into a
+    // single result by using flatMap. Similar to following Scala code:
+    // for {
+    //   r0 <- Right(input)
+    //   r1 <- f1(r0)
+    //   r2 <- f2(r1)
+    //   ...
+    // } yield fn
     contract Either(@"compose", @input, @functions, return) = {
       new combinator in {
         contract combinator(@head, @accumulatedValue, return) = {
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -41,8 +41,57 @@ new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lo
     // Lifts a Boolean into Either, making `false` into Left and using Right otherwise.
     contract Either(@"fromBoolean", @success, @failureMessage, return) = {
       match success {
-        Nil => return!(("Left", failureMessage))
-        _   => return!(("Right", Nil))
+        false => return!(("Left", failureMessage))
+        true  => return!(("Right", Nil))
+      }
+    } |
+
+    // Like fromSingletonList, except it takes `listCh` and reads it to obtain the `list` parameter
+    contract Either(@"fromSingletonList <-", listCh, @failureMessage, return) = {
+      for (@list <- listCh) {
+        Either!("fromSingletonList", list, failureMessage, *return)
+      }
+    } |
+
+    // Converts a singleton list into Either, making the single element into Right
+    // and using Left if it's not a singleton list.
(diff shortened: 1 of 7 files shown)


casper/src/test/resources/RevAddressTest.rho
```diff
@@ -29,10 +31,41 @@ in {
...
+        (("11112comWwT626KfXkvnWRHDwPDm4VefqSr6ijvinhVHGTuXe6wWmk", "== <-", *r2), "correct RevAddress"),
+        (("11112ipm5NY9BqyQ3p3sUx8pdFWV5jFQ5qqMcUzZWe19hwz3USbHBG", "== <-", *r3), "correct RevAddress"),
+      ], *ackCh)
+    }
+  } |
+
+  contract test_fromPublicKey_invalid(rhoSpec, _, ackCh) = {
+    new retCh in {
+      revAddress!("fromPublicKey", "F0F0".hexToBytes(), *retCh) |
+      rhoSpec!("assert", (Nil, "== <-", *retCh), "correct RevAddress", *ackCh)
```
odeac: the clue is the information displayed with the test when it fails so it should explain why the tests failed. 

In this case it should be something like "the key length is incorrect so the result must be Nil"

casper/src/test/resources/RevAddressTest.rho
```diff
@@ -29,10 +31,41 @@ in {
...
+      ) |
+      revAddress!(
+        "fromPublicKey",
+        "2222222222222222222222222222222222222222222222222222222222222222".hexToBytes(),
+        *r3
+      ) |
+      rhoSpec!("assertMany", [
+        (("111123jwD7acNy6vfYwzWXAbnSQ8z8Bv24cCkVcVbd6SuP2LtFK7w7", "== <-", *r1), "correct RevAddress"),
+        (("11112comWwT626KfXkvnWRHDwPDm4VefqSr6ijvinhVHGTuXe6wWmk", "== <-", *r2), "correct RevAddress"),
+        (("11112ipm5NY9BqyQ3p3sUx8pdFWV5jFQ5qqMcUzZWe19hwz3USbHBG", "== <-", *r3), "correct RevAddress"),
```
odeac: In this case three tests don't bring more value than one test

casper/src/main/resources/Either.rho
```diff
@@ -23,6 +58,12 @@ new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lo
       }
     } |
 
+    contract Either(@"map <-", eitherCh, f, return) = {
+      for (@either <- eitherCh) {
+        Either!("map", either, *f, *return)
+      }
+    } |
+
```
odeac: please add tests for all these new methods

casper/src/main/resources/RevVault.rho
```diff
@@ -102,7 +104,7 @@ in {
         contract _newVault(@ownRevAddress, purse, ret) = {
           new revVault in {
 
-            ret!(bundle+{*revVault}) |
+            ret!(("Right", bundle+{*revVault})) |
```
odeac: when does this _newVault contract return a "Left" value?

ArturGajowy: This changes in the next commit, that was a bandaid so that I could have the tests pass without changing everything at once.

JosDenmark:
@ArturGajowy Make sure to provide scaladoc style documentation for each of the contracts you define here. Rholang is nowhere near self-documenting.

casper/src/main/resources/RevVault.rho
```diff
@@ -0,0 +1,145 @@
...
+                      ret!(("Left", "Insufficient funds"))
+                    [p] => {
+                      @{revAddress | bundle0{*_revVault}}!("_deposit", p, *ret)
+                    }
+                  }
+                }
+              }
+            } |
+
+            contract @{ownRevAddress | bundle0{*_revVault}}(@"_deposit", deposit, retCh) = {
```
JosDenmark: Instead of `deposit`, it should probably be `purse` or something similar.

casper/src/test/resources/RevVaultTest.rho
```diff
@@ -1,36 +1,128 @@
...
+                ((Nil, "== <-", *unlockFail), "can't unlock without proper identity"),
+              ], *ackCh
+            )
+          }
+        }
+      } |
+
+      contract testCreateAliceVaultUnlocked(rhoSpec, RevVault, ackCh) = {
+        new balanceCh, aliceVaultCh in {
+          withAliceVault!(*RevVault, *aliceVaultCh) |
```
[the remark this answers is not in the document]
ArturGajowy: It makes sense to have those tests in a single suite. The duplication is properly extracted using the withVault method (Alice was eliminated from the name and body)

casper/src/test/resources/RevAddressTest.rho
ArturGajowy: I think the test name and the compared values in the assertion message do a good enough job here

casper/src/test/resources/RevAddressTest.rho
ArturGajowy: Having all those has the advantage of making the magic strings used in the header of RevVaultTest.rho searchable in the codebase, which saves work to anybody willing to validate if the strings used there are indeed valid.

casper/src/main/resources/Either.rho
ArturGajowy: Will do either in this PR or in a follow up :+1: For now, I think you can tell from the implementations and the RevVaultTest.rho passing that those are OK

ArturGajowy:
bors try

casper/src/main/resources/RevVault.rho
```diff
@@ -13,15 +13,139 @@
...
+              new vaultCh in {
+                _makeVault!(ownerPubKeyBytes, 0, *vaultCh) |
+                for (@vault <- vaultCh) {
+
+                  @Lockbox!("lock", ownerPubKeyBytes, vault, *retCh)
+                }
+              }
+            }
+          }
+        } |
```
KentShikama: I don't quite understand this method. We create a vault at `vaultCh` but we don't insert it into the `vaultMapStore`?

ArturGajowy: we create a `vault`, by calling `_makeVault`, which will return it on `vaultCh`. We listen on the `vaultCh`, take the `vault`, `"lock"` it in a `Lockbox`, return the new lockbox via `retCh`.

KentShikama: So when is it inserted into the `vaultMapStore`?

ArturGajowy: re `vaultMapStore`: the `_getOrCreate` method handles all insertions into the vaultMap.
So either there is a vault for such public key - and then the `createLockedVault` continuation/lambda isn't called at all - or there isn't, in which case `createLockedVault` creates the vault and returns the control flow to `_getOrCreate` by sending the result to `retCh`.

casper/src/main/resources/RevVault.rho
```diff
@@ -0,0 +1,154 @@
...
+
+              new
+                revAddressValid, revAddressValidEither, amountNonNegative, split,
+                eitherPurseCh, purseReadyCh, doDeposit
+              in {
+                RevAddress!("validate", revAddress, *revAddressValid) |
+                @Either!("fromNillableError <-", *revAddressValid, *revAddressValidEither) |
+                @Either!("fromBoolean", amount >= 0, "Amount must be non-negative", *amountNonNegative) |
+                @Either!("flatMap <-", *amountNonNegative, *split, *eitherPurseCh) |
+                for (_, ret <- split) {
```
odeac: This is consuming two consecutive values from the split channel (not a Tuple) while `Either.flatMap` sends a tuple to the `f` parameter.

Why is this working? I'm expecting it to block while waiting for `ret` parameter. 

Am I missing something?

ArturGajowy: afaiu it does an arity-2 receive, complementary to an arity-2 send done in https://github.com/rchain/rchain/pull/2268/files#diff-ca522a4895f4cbde9cde7606fe2bc704R131. Notice none of them are sending/receiving a tuple directly (which would be arity-1). Rather, both of them send/receive two values, the first being a 2-tuple (the either) and the second being the return channel

ArturGajowy: Also, AFAIK there's no such ting as 'consecutive values in a channel', as channel receive/send order is non-deterministic

casper/src/main/resources/RevVault.rho
```diff
@@ -0,0 +1,154 @@
...
+                for (@vault, retCh <- lock) {
+                  @Lockbox!("lock", ownerPubKeyBytes, vault, *retCh)
+                }
+              }
+            }
+          }
+        } |
+
+        contract _getOrCreate(@key, constructor, ret) = {
+          for (@vaultMap <- vaultMapStore) {
```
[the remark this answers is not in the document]
ArturGajowy: I think this is such a basic idiom and so easy to do with the just the language (without libs), that adding a lib contract would defeat the purpose. It might make sense if we find some substantial functionality to build on top of that, but as it is now - it's just a send + receive on a channel that only you know, and that stores at most 1 value. Doing anything around that requires the user to start using the 'method call convention', and a single 'method call' is as complex to set up as the whole idiom. So I really don't see any benefit for now (without a method call protocol built into rholang)

casper/src/main/resources/RevVault.rho
```diff
@@ -0,0 +1,154 @@
...
+            } |
+
+            contract @{ownRevAddress | bundle0{*_revVault}}(@"_deposit", depositPurse, retCh) = {
+
+              new amountCh, depositSuccessCh in {
+                depositPurse!("getBalance", *amountCh) |
+                for (@amount <- amountCh) {
+
+                  purse!("deposit", amount, *depositPurse, *depositSuccessCh) |
+                  @Either!("fromBoolean <-", *depositSuccessCh, "BUG FOUND: purse deposit failed", *retCh)
```
odeac: The purse deposit can fail in case of overflow. In such situation we need to revert the split.

ArturGajowy: Yeah, we should at least change the message. I'm not sure though what should happen when one transfers to an already-full ~~wallet~~ vault:
 - should the funds be lost (easiest one)
 - should they be refunded (can cause a cascade and is subject to concurrency bugs)
 - should the vault keep a pile of purses and prevent overflow proactively? (require substantial development effort)

Given that it's unclear how to proceed, and that this is admittedly a rare scenario - are you OK with me scheduling a JIRA for fixing that later?

casper/src/main/resources/RevVault.rho
```diff
@@ -0,0 +1,154 @@
...
+            RevAddress!("fromPublicKey", ownerPubKeyBytes, *revAddrCh) |
+            @Either!("fromNillable <-", *revAddrCh, "Invalid public key", *eitherRevAddrCh) |
+            mint!("makePurse", initialAmount, *purseCh) |
+            @Either!("fromNillable <-", *purseCh, "Couldn't create purse", *eitherPurseCh) |
+            @Either!("map2 <-", *eitherRevAddrCh, *eitherPurseCh, *_newVault, *ret)
+          }
+
+        } |
+
+        contract _newVault(@ownRevAddress, purse, ret) = {
```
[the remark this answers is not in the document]
ArturGajowy: I'd rather not nest the scopes any deeper. I think of it as two private methods on the same class, one calling the other.

casper/src/main/resources/RevVault.rho
odeac: Indeed, this is not trivial fix so I'm ok with a new JIRA issue

PR #2276, opened 2019-03-15
RCHAIN-2729: Return an error if phloLimit passed to deploy command is less than intrinsic phlo

JosDenmark:
## Overview
This PR changes `RuntimeInterpreter` to throw an error if the deploy being evaluated provides phlo less than the minimum cost of evaluating a deploy. It is the first of a series of validity checks that deploys will be subject to.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2729

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -8,8 +8,6 @@ import coop.rchain.casper.PrettyPrinter
 import coop.rchain.models.EquivocationRecord.SequenceNumber
 import coop.rchain.casper.Estimator.{BlockHash, Validator}
 import coop.rchain.casper.protocol.{DeployData, _}
-import coop.rchain.casper.util.ProtoUtil.basicDeployData
-import coop.rchain.casper.util.rholang.InterpreterUtil
 import coop.rchain.casper.util.implicits._
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b256
@@ -506,6 +504,16 @@ object ProtoUtil {
       timestamp = timestamp,
       term = source,
       phloLimit = phlos
+    )
+
+  def sourceDeployNowWithPhlo[F[_]: Monad: Time](source: String, phlos: Long): F[DeployData] =
+    Time[F].currentMillis.map(
+      now =>
+        sourceDeploy(
+          source,
+          now,
+          phlos
+        )
     )
 
   def sourceDeployNow(source: String): DeployData =
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -205,14 +205,12 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
         terms match {
           case deploy +: rem =>
             for {
-              _ <- runtime.space.reset(hash)
-              (codeHash, phloPrice, userId, timestamp) = ProtoUtil.getRholangDeployParams(
-                deploy
-              )
-              _                                      <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
-              injResult                              <- doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
-              EvaluateResult(evaluationCost, errors) = injResult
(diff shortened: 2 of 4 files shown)


JosDenmark:
bors try

JosDenmark:
> We want to replay deploys with an insufficient phlo - otherwise a malicious validator can mark a bunch of valid transactions as invalid.

I'm unsure what you mean. The insufficient phlo check is done in `doInj`, which is used both in `doReplayEval` and `doEval`. The phlo check is repeated independently during replay.

KentShikama:
`   "replayComputeState" should "not replay a deploy that throws an InsufficientPhloError" in {`

I interpreted that test title to mean that we don't replay deploys with an insufficient phlo.

JosDenmark:
Ah. I see. I'll rename to `"replayComputeState" should "rethrow InsufficientPhloError"`.

JosDenmark:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -293,19 +291,27 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
     implicit val rand: Blake2b512Random = Blake2b512Random(
       DeployData.toByteArray(ProtoUtil.stripDeployData(deploy))
     )
-    Interpreter[F].injAttempt(
-      reducer,
-      errorLog,
-      deploy.term,
-      Cost(deploy.phloLimit)
-    )
+    val intrinsicGas: Long = parsingCost(deploy.term).value
```
KentShikama: Why not just `parsingCost` or `parsingGas`?

casper/src/main/scala/coop/rchain/casper/util/SignatureAlgorithms.scala
```diff
@@ -4,10 +4,19 @@ import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
...
+  def lookupSigAlg(algorithm: String): SignatureAlgorithm =
     algorithm match {
       case "ed25519"   => Ed25519.sign
       case "secp256k1" => Secp256k1.sign
       case _           => throw new Exception(s"Unknown signature algorithm $algorithm")
     }
+
+  type SignatureVerifyAlgorithm = (Array[Byte], Array[Byte], Array[Byte]) => Boolean
+  @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
+  def lookupSigVerifyAlg(algorithm: String): SignatureVerifyAlgorithm =
```
KentShikama: So where do you use this?

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -508,6 +508,13 @@ object ProtoUtil {
       phloLimit = phlos
     )
 
+  def sourceDeployNowWithPhlo(source: String, phlos: Long): DeployData =
+    sourceDeploy(
+      source,
+      System.currentTimeMillis(),
```
KentShikama: I prefer if we could use `Time[F].currentMillis` so we can inject logicalTime for tests

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: In the future, it's possible that it includes costs besides parsing. For example, while we don't have charging in the normalizer, we might include a cost for normalization.

casper/src/main/scala/coop/rchain/casper/util/SignatureAlgorithms.scala
JosDenmark: This was part of a larger PR at first which used it. It should be removed.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
KentShikama: I see; I say we rename it then

JosDenmark: Rename it to what? The terminology is taken from Ethereum, which has a somewhat analogous minimum.

KentShikama: To intrinsic gas

PR #2278, opened 2019-03-18
RCHAIN-3110 remove roundtrip

rabbitonweb:
## Overview
Remove roundTrip!!!!1111

### JIRA ticket:
RCHAIN-3110


rabbitonweb:
`Pong` is gone because `roundtrip no longer exists.
I think I will rename `heartbeat` to `poke`

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransport.scala
```diff
@@ -56,7 +56,6 @@ object GrpcTransport {
       .flatMap(
         tlr =>
           tlr.payload match {
-            case p if p.isProtocol   => Right(Some(tlr.getProtocol))
             case p if p.isNoResponse => Right(None)
```
sebtomba: Please refactor the functions `processResponse` and `transport` to return `CommErr[Unit]`. Why does it still return `CommErr[Option[Protocol]]`? I thought that `Protocol` is gone?

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransport.scala
```diff
@@ -74,26 +73,11 @@ object GrpcTransport {
...
-
   def send(peer: PeerNode, msg: Protocol)(implicit metrics: Metrics[Task]): Request[Unit] =
     for {
       _ <- ReaderT.liftF(metrics.incrementCounter("send"))
       result <- transport(peer)(
-                 _.tell(TLRequest(msg.some))
+                 _.send(TLRequest(msg.some))
                    .timer("send-time")
                ).map(_.flatMap {
                  case Some(p) =>
```
sebtomba: This will change once you change `transport` to return `CommErr[Unit]`

rabbitonweb: no , it will not

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransport.scala
rabbitonweb: I will be removing `CommunicationResponse` and refactoring `PacketHandler` in next PR (small steps)

PR #2279, opened 2019-03-18
_cost synchronization

lukasz-golebiewski:
## Overview
~~Replaces the underlying Ref in _cost with an MVar~~. Introduces a Semaphore and Bracket in the _cost capability for synchronization. This needs to be done due to the fact that Ref does not provide synchronization by itself, which means that concurrent branches of execution would not always see the latest amount of available phlogistons and continue executing also if the account was in the meantime depleted 

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3077

### Notes
Caveat: MVar doesn't work well with Bracket. Upon task cancellation the release function is provided with an outdated value

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
+++ b/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
@@ -64,7 +64,7 @@ object TestUtil {
       _ <- runtime.reducer.setPhlo(Cost.UNSAFE_MAX)
       _ <- runtime.reducer.inj(term)
       _ <- runtime.reducer.phlo
-    } yield ()).runSyncUnsafe(5.seconds)
+    } yield ()).runSyncUnsafe(10.seconds)
 
   def eval(
       code: String,
--- a/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
+++ b/casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
@@ -44,26 +44,27 @@ object TestUtil {
       runtime <- TestRuntime.create[Task, Task.Par](extraServices)
       _       <- Runtime.injectEmptyRegistryRoot[Task](runtime.space, runtime.replaySpace)
     } yield (runtime)
-  }.unsafeRunSync
+  }.runSyncUnsafe(5.seconds)
 
   def evalDeploy(deploy: DeployData, runtime: Runtime[Task])(
       implicit scheduler: Scheduler
   ): Unit = {
-    runtime.reducer.setPhlo(Cost.UNSAFE_MAX).runSyncUnsafe(1.second)
     implicit val rand: Blake2b512Random = Blake2b512Random(
       DeployData.toByteArray(ProtoUtil.stripDeployData(deploy))
     )
     val term = mkTerm(deploy.term).right.get
-    runtime.reducer.inj(term).unsafeRunSync
+    evalTerm(term, runtime)
   }
 
   def evalTerm(
       term: Par,
       runtime: Runtime[Task]
-  )(implicit scheduler: Scheduler, rand: Blake2b512Random): Unit = {
-    runtime.reducer.setPhlo(Cost.UNSAFE_MAX).runSyncUnsafe(1.second)
-    runtime.reducer.inj(term).unsafeRunSync
-  }
+  )(implicit scheduler: Scheduler, rand: Blake2b512Random): Unit =
+    (for {
+      _ <- runtime.reducer.setPhlo(Cost.UNSAFE_MAX)
+      _ <- runtime.reducer.inj(term)
+      _ <- runtime.reducer.phlo
+    } yield ()).runSyncUnsafe(5.seconds)
 
   def eval(
       code: String,
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -61,7 +61,7 @@ object Interpreter {
                       case Right(parsed) =>
                         for {
                           result    <- reducer.inj(parsed).attempt
-                          phlosLeft <- C.get
+                          phlosLeft <- C.inspect(identity)
                           oldErrors <- errorLog.readAndClearErrorVector()
                           newErrors = result.swap.toSeq.toVector
                           allErrors = oldErrors |+| newErrors
@@ -72,6 +72,7 @@ object Interpreter {
                   case Left(error) =>
                     EvaluateResult(parsingCost, Vector(error)).pure[F]
                 }
+          _ <- C.get
         } yield res
 
       }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
@@ -1,12 +1,12 @@
 package coop.rchain.rholang.interpreter
 
 import cats._
-import cats.effect.Sync
+import cats.effect._
 import cats.implicits._
 import coop.rchain.catscontrib.mtl.implicits._
 import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.rholang.interpreter.accounting._
-import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
+import coop.rchain.rholang.interpreter.errors._
 
 final case class EvaluateResult(cost: Cost, errors: Vector[Throwable])
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
@@ -1,7 +1,7 @@
 package coop.rchain.rholang.interpreter.accounting
 
-import cats.effect.Sync
-import cats.effect.concurrent.Ref
+import cats.effect.Concurrent
+import cats.effect.concurrent.MVar
 import cats.implicits._
 import cats.mtl._
 import cats.{FlatMap, Monad}
@@ -9,20 +9,28 @@ import coop.rchain.rholang.interpreter.errors.OutOfPhlogistonsError
 
 object CostAccounting {
 
-  def of[F[_]: Sync](init: Cost): F[MonadState[F, Cost]] =
-    Ref[F]
+  def of[F[_]: Concurrent](init: Cost): F[MonadState[F, Cost]] =
+    MVar[F]
       .of(init)
-      .map(
-        state =>
-          new DefaultMonadState[F, Cost] {
-            val monad: cats.Monad[F]                      = implicitly[Monad[F]]
-            def get: F[Cost]                              = state.get
-            def set(s: Cost): F[Unit]                     = state.set(s)
-            override def modify(f: Cost => Cost): F[Unit] = state.update(f)
-          }
-      )
+      .map(defaultMonadState)
 
-  def empty[F[_]: Sync]: F[MonadState[F, Cost]] =
-    this.of(Cost(0, "init"))
+  def empty[F[_]: Concurrent]: F[MonadState[F, Cost]] =
+    MVar[F]
+      .empty[Cost]
+      .map(defaultMonadState)
 
+  private[this] def defaultMonadState[F[_]: Monad] =
+    (state: MVar[F, Cost]) =>
+      new DefaultMonadState[F, Cost] {
+        val monad: cats.Monad[F]                    = implicitly[Monad[F]]
+        def get: F[Cost]                            = state.take
+        def set(s: Cost): F[Unit]                   = state.put(s)
+        override def inspect[A](f: Cost => A): F[A] = state.read.map(f)
+        override def modify(f: Cost => Cost): F[Unit] =
+          for {
+            s <- get
+            _ <- set(f(s))
+          } yield ()
+
+      }
 }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
@@ -1,5 +1,6 @@
 package coop.rchain.rholang.interpreter.accounting
 
+import cats.data._
 import cats.effect.Concurrent
 import cats.effect.concurrent._
 import cats.implicits._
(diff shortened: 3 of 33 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
```diff
@@ -648,13 +649,13 @@ class VarMatcherSpec extends FlatSpec with Matchers with TimeLimits with TripleE
...
-    *
-    * here we match on the entire bundle, not its components.
-    */
+ *
+ * bundle { @7!(42) | @"0"!(43) } match {
+ *   w => …
+ * }
+ *
+ * here we match on the entire bundle, not its components.
+ */
```
dzajkowski: formatting?

dzajkowski:
seems that the 1st commit is redundant

lukasz-golebiewski:
Yes, it was added to check whether introducing the `Bracket` is absolutely necessary

lukasz-golebiewski:
bors try

lukasz-golebiewski:
bors try

lukasz-golebiewski:
bors try

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
```diff
@@ -189,6 +190,29 @@ trait StreamTInstances2 {
...
       def zero[A](state: Stream[A]): Boolean = state.isEmpty
     }
+
+  implicit def streamTSync[F[_]](implicit F0: Sync[F]): Sync[StreamT[F, ?]] =
+    new StreamTSync[F]() {
+      implicit val F = F0
+    }
+}
+
+private trait StreamTSync[F[_]] extends Sync[StreamT[F, ?]] with StreamTMonadError[F, Throwable] {
```
ArturGajowy: Please remember to add laws tests for this Sync instance. See StreamTLawsSpec. The laws are defined on https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/discipline/SyncTests.scala. We seem to need a cats-effect-laws dependency

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
lukasz-golebiewski: I'm afraid it is impossible to construct a lawful `Sync` instance for `StreamT`. It will always be a bit special - that is - possible to use only when the stream is not empty. Otherwise it won't be possible to `acquire` an element from it and provide it to the `use` function

lukasz-golebiewski:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -35,27 +36,36 @@ class CostAccountingReducerTest extends FlatSpec with Matchers with TripleEquals
...
-      .value
-    assert(res.isLeft)
-    assert(costAlg.get() === (initCost - Cost(originalTermCost)))
+
+    (for {
+      cost <- CostAccounting.initialCost[Task](initCost)
+      res <- {
+        implicit val c = cost
+        Substitute
+          .charge(Task.raiseError[Par](new RuntimeException("")), Cost(originalTermCost))
```
KentShikama: Lets at least add a TODO to insert a suitable exception message

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
```diff
@@ -44,26 +44,27 @@ object TestUtil {
...
       runtime: Runtime[Task]
-  )(implicit scheduler: Scheduler, rand: Blake2b512Random): Unit = {
-    runtime.reducer.setPhlo(Cost.UNSAFE_MAX).runSyncUnsafe(1.second)
-    runtime.reducer.inj(term).unsafeRunSync
-  }
+  )(implicit scheduler: Scheduler, rand: Blake2b512Random): Unit =
+    (for {
+      _ <- runtime.reducer.setPhlo(Cost.UNSAFE_MAX)
+      _ <- runtime.reducer.inj(term)
+      _ <- runtime.reducer.phlo
```
KentShikama: So what does this line do?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -61,7 +61,7 @@ object Interpreter {
                       case Right(parsed) =>
                         for {
                           result    <- reducer.inj(parsed).attempt
-                          phlosLeft <- C.get
+                          phlosLeft <- C.inspect(identity)
```
KentShikama: What is the difference?

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
lukasz-golebiewski: I was wrong

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
```diff
@@ -2,52 +2,84 @@ package coop.rchain.rholang.interpreter
...
-      _ <- error.ensure(cost.get)(OutOfPhlogistonsError)(_.value >= 0)
+      _ <- error.bracket(cost.acquire)(
+            _ =>
+              for {
+                c <- cost.get
+                _ <- Monad[F].ifM[Unit]((c.value < 0).pure[F])(
+                      ifTrue = {
+                        for {
+                          _ <- error.raiseError[Unit](OutOfPhlogistonsError)
+                        } yield (())
```
KentShikama: Just `error.raiseError[Unit](OutOfPhlogistonsError)`?

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
lukasz-golebiewski: This would read the phlo value thus emptying the MVar and unblocking the threads awaiting to re-populate it

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
```diff
@@ -191,27 +192,45 @@ trait StreamTInstances2 {
...
       release: (A, ExitCase[Throwable]) => StreamT[F, Unit]
   ): StreamT[F, B] =
-    StreamT.liftF(
-      F.bracketCase(acquire)(use.andThen(unlift(_)))((a, b) => release(a, b))
+    StreamT.fromStream(
+      F.bracketCase[Stream[A], Stream[B]](StreamT.run(acquire)) {
+        case Stream.Empty => F.pure[Stream[B]](Stream.empty)
+        case cons: Cons[A] =>
+          StreamT
+            .run(cons.map(use).fold(StreamT.empty[F, B])(AL.combineK))
```
ArturGajowy: We must not be using `.run` in here (or anywhere except the end of the world, really), because it forces the computation - and part of the point in having a StreamT is to have it lazy.

Instead of running the entire computation, we should be 'running just the first step, adapting it as needed, and returning a computation that will do the same in its next step'.
The method to 'run a single step is `StreamT#next`.

See the implementation of combineK https://github.com/rchain/rchain/blob/e319d584b7287e7cb7ee7da71f182c4fb36e1dd6/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala#L95

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
```diff
@@ -1,28 +1,46 @@
...
+    for {
+      s <- Semaphore(1)
+      c <- empty
+    } yield (loggingCost(c, L, s))
+
+  def initialCost[F[_]: Concurrent](
+      init: Cost
+  )(implicit L: FunctorTell[F, Chain[Cost]]): F[_cost[F]] =
+    for {
+      s <- Semaphore(1)
```
ArturGajowy: why `Fall back to Ref, use Semaphore for synchronization`? I assume Sync is still needed when we do the fall back? If so - I'm not sure I see the benefit?

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
lukasz-golebiewski: The message is irrelevant

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
odeac: Then I guess it's not a good idea to construct a Sync[StreamT] instance

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
ArturGajowy: kudos for delegating the Sync to the underlying effect though :+1:

ArturGajowy: I mean especially the fact that it's a `MonadError[Throwable`

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
lukasz-golebiewski: Sync provides the `bracketCase` operation

ArturGajowy: This doesn't answer why a fallback to Ref + Semaphore is needed / better tham MVar. I don't have preference, I want to understand (and then I might have preference)

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
```diff
@@ -931,15 +929,6 @@ class VarMatcherSpec extends FlatSpec with Matchers with TimeLimits with TripleE
     assertSpatialMatch(failTarget, pattern, None)
   }
```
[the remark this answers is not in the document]
lukasz-golebiewski: It was duplicated

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
lukasz-golebiewski: `get` can be blocking if the underlying `MonadState` is an MVar

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/CostAccounting.scala
lukasz-golebiewski: It's described in the "caveats" section of this PR

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
```diff
@@ -194,38 +194,37 @@ trait StreamTInstances2 {
...
       release: (A, ExitCase[Throwable]) => StreamT[F, Unit]
   ): StreamT[F, B] =
-    StreamT.fromStream(
-      F.bracketCase[Stream[A], Stream[B]](StreamT.run(acquire)) {
-        case Stream.Empty => F.pure[Stream[B]](Stream.empty)
-        case cons: Cons[A] =>
-          StreamT
-            .run(cons.map(use).fold(StreamT.empty[F, B])(AL.combineK))
+    StreamT(
+      F.bracketCase[Step[F, A], Step[F, B]](acquire.next) {
```
ArturGajowy: you can probably avoid those `[Step[F, A], Step[F, B]]` if you declare 

```scala
val next: F[Step[F, B]] = F.bracketCase() { ... } { ... }
StreamT(next)
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
```diff
@@ -194,38 +194,37 @@ trait StreamTInstances2 {
...
+        case SNil() => F.pure(SNil())
+        case SCons(head, tail) => {
+          AL.combineK(use(head), M.flatMap(tail)(use)).next
```
ArturGajowy: With obtaining the lock once, then doing the whole stream, and then releasing it, we exclude exchanging the lock usage between the elements of two streams using it simultaneously. I think it's not a problem in our case though, and the laws seem to agree.

ArturGajowy: We really need to Nil-out the result if release returns a Nil. See the equivalent code for OptionT or Either:

```
{
            case (None, _) => F.unit //Nothing to release
            case (Some(a), ExitCase.Completed) =>
              release(a, ExitCase.Completed).value.flatMap {
                case None => ref.set(true)
                case Some(_) => F.unit
              }
            case (Some(a), res) => release(a, res).value.void
          }.flatMap[Option[B]] {
            case s @ Some(_) => ref.get.map(b => if (b) None else s)
            case None => F.pure(None)
          }
```

ArturGajowy: Please add a test for that, and maybe see how do they test that in cats-effect

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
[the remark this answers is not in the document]
lukasz-golebiewski: by `  "spatialMatchAndCharge" should "short-circuit when runs out of phlo in the middle of matching" in {
`

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
lukasz-golebiewski: Why do you think "we really need to Nil-out the result if release returns a Nil"?

ArturGajowy:
Because otherwise we're ingoring the fact that realease returned a Nil. And
to be consistent with what other instances do.

rholang/src/test/scala/coop/rchain/rholang/interpreter/matcher/MatchTest.scala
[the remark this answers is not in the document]
ArturGajowy: In the future it'd help if you provided this info in the commit message.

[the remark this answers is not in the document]
ArturGajowy: We should also probably remove the `spatialMatchAndCharge` method (inline it)

lukasz-golebiewski:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/StreamT.scala
lukasz-golebiewski: Please do not respond outside of the thread, it makes the conversation harder to follow.
This should by no means be a blocker for this pull request, but ok. It is implemented

rholang/src/test/scala/coop/rchain/rholang/interpreter/RegistrySpec.scala
```diff
@@ -58,7 +55,9 @@ trait RegistryTester extends PersistentStoreTester {
   ): R =
     withTestSpace(errorLog) {
       case TestFixture(space, _) =>
-        val _                                        = errorLog.readAndClearErrorVector().runSyncUnsafe(1.second)
+        val _             = errorLog.readAndClearErrorVector().runSyncUnsafe(1.second)
```
odeac: we keep adding runSyncUnsafe calls. At some moment we should group these in a task on which we only call runSyncUnsafe once. Maybe that moment is now, if you feel like it. No pressure. https://www.youtube.com/watch?v=JfnddMpzPsM

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
```diff
@@ -2,52 +2,78 @@ package coop.rchain.rholang.interpreter
...
   )(implicit cost: _cost[F], error: _error[F]): F[Unit] =
     for {
-      _ <- cost.tell(Chain.one(amount))
-      _ <- cost.modify(_ - amount)
-      _ <- error.ensure(cost.get)(OutOfPhlogistonsError)(_.value >= 0)
+      _ <- error.bracket(cost.acquire)(
+            _ =>
+              for {
+                c <- cost.get
+                _ <- Monad[F].ifM[Unit]((c.value < 0).pure[F])(
```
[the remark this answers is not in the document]
lukasz-golebiewski: There isn't one, I got carried away

PR #2284, opened 2019-03-19
Vaults demo

ArturGajowy:
## Overview

### JIRA ticket:

### Notes


rholang/examples/vault_demo/1.create_alcie_vault.rho
```diff
@@ -0,0 +1,52 @@
...
+      stdout(`rho:io:stdout`), log
+    in {
+
+      rl!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *RevVaultCh) |
+      for (@(_, RevVault) <- RevVaultCh) {
+        log!(("1.create_alcie_vault.rho")) |
+
+        @RevVault!("findOrCreateGenesisVault", genesisPubKey, 9000, *genesisVaultCh ) |
+
+        @RevVault!("findOrCreate", alicePubKey, *aliceVaultLockedCh) |
```
cboscolo: @ArturGajowy this call needs to be the REV Address, not alicePublicKey.  The REV Address is what they exchange in the real world outside of any RChain software.

ArturGajowy: The exchange does not need to create the wallet to make the transfer, nor does the target wallet need to exist for that. It can be created later based on the pub key used to derive the revaddress, at which point the funds from the exchange will make its way to it. Also, the funds will be deducted from the exchange even if the wallet at the address doesn't exist.

cboscolo: perhaps I should ask a different question.  What is the purpose of this "findOrCreate" call? Below, there is a call to:
`genesisVault!("transfer", aliceRevAddress, 1000, *transferDone)`
which should should hide the details of this "findOrCreate" logic.

casper/src/main/resources/RevVault.rho
```diff
@@ -31,11 +31,11 @@ in {
       @MakeMint!(*mintCh) |
       for (mint <- mintCh) {
 
-        //FIXME the result of createGenesisVault is not protected, nor is the call to it. Both should be.
-        contract RevVault(@"createGenesisVault", @ownerPubKeyBytes, @initialAmount, ret) = {
-          new createGenesisVault in {
-            _getOrCreate!(ownerPubKeyBytes, *createGenesisVault, *ret) |
-            for (retCh <- createGenesisVault) {
+        //FIXME the result of findOrCreateGenesisVault is not protected, nor is the call to it. Both should be.
```
cboscolo: Do we even need a "findOrCreateGenesisVault"?  I think we would be better to initialize the RevAddress Map with the proper address balances so they end up in the genesis block.

rholang/examples/vault_demo/1.create_alcie_vault.rho
ArturGajowy: If `transfer` would create the target vault when it doesn't exist, there would be no way of transferring funds across different 'kinds' of vaults, or we'd need to parameterize vaults with a 'vault creation strategy'. We could be targeting a multisig, after all.

You're right that `findOrCreate` is redundant here. I intended to show the balance, but I guess it makes more sense to decouple that into two deploys.

casper/src/main/resources/RevVault.rho
ArturGajowy: The map doesn't store the balances. While hardcoding the genesis vaults creation along with their corresponding pubKeys could be an option, more questions remain:
1. how to secure access to those vaults?
2. how to put funds in them without providing a pre-filled purse to be used internally? (we probably could call _deposit directly from the revvault contract, but this needs some thought)

Long story short, this aspect hasn't yet been addressed. Do we need those sorted out for the demo?

ArturGajowy: Oh I see I've just answered two of my questions :) The last one remains though :)

cboscolo: I was sloppy with my words and didn't intend to imply balances are kept anywhere but the purse in the vault. But before I clarify, and repose a suggestion I have a question:

In your implementation is there a map storing **Vaults** (key being a PublicKey):
https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/RevVault.rho#L30
And a tuple storing purses (uniquely keyed by the Rev Addresses):
https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/RevVault.rho#L130

ArturGajowy: I'm not sure I get the question. There is a pubKey -> vault, but there isn't another map. The second thing you're pointing to is a contract, registered according to the address scheme. We might leverage the `_deposit` contract by addressing it directly (not from within `transfer`) in some init code in the RevVault. What's your suggestion?

rholang/examples/vault_demo/0.transfer_from_genesis.rho
```diff
@@ -0,0 +1,60 @@
...
+    log!(("Using public key:", pubKey, "to obtain a `deployer's-pubkey-bound` vault")) |
+
+    @RevVault!("findOrCreateGenesisVault", pubKey, 9000, *vaultCh ) |
+
+    // most Vault methods return an `Either[String, A] = ("Left", String) \/ ("Right", A)`
+    for (@("Right", vault) <- vaultCh) {
+
+      log!(("Genesis vault obtained")) |
+
+      match ("%REV_ADDR", 100) {
```
tymm: It's already pretty obvious but it might add some extra clarity if you use `%RECIPIENT_REV_ADDR` or something like that.

ArturGajowy: I'd rather not, it's easier to type during the demo that way :)

PR #2285, opened 2019-03-19
RCHAIN-3113: Add curly brackets around receive in match case when pretty printing

tymm:
## Overview
Without this fix the PrettyPrinter outputs Rholang that is not parsable.

Another solution might be to change the grammar in `rholang_mercury.cf` from
```
-- Match Cases
CaseImpl. Case ::= Proc13 "=>" Proc3 ;
```
to
```
-- Match Cases
CaseImpl. Case ::= Proc13 "=>" Proc1 ;
```
This would result in `match Nil { _ => for(_ <- @Nil) {Nil} }` to be parsable.
I didn't go that route because it looks to me as if there is *some* reason that the bnfc rule is what it is.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3113

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
@@ -440,8 +440,12 @@ class ProcPrinterSpec extends FlatSpec with Matchers {
     checkRoundTrip(
       """for( @{x0}, @{x1} <- @{0} ) {
         |  match x0 {
-        |    =x0 => Nil
-        |    =x1 => Nil
+        |    =x0 => {
+        |      Nil
+        |    }
+        |    =x1 => {
+        |      Nil
+        |    }
         |  }
         |}""".stripMargin
     )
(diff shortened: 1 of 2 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
```diff
@@ -745,6 +745,18 @@ class ProcPrinterSpec extends FlatSpec with Matchers {
...
 
+  it should "Print receive in curly brackets" in {
+    checkRoundTrip(
+      """match Nil {
+        |  _ => {
+        |    for( @{_} <- @{Nil} ) {
+        |      Nil
+        |    }
+        |  }
+        |}""".stripMargin
```
ArturGajowy: We should actually always put braces around the process that is the body of a match case

tymm: Except for sends, right?

tymm: And out of interest: why do you think so?

ArturGajowy: why except for sends?
I think so because of consistency :)

ArturGajowy: And implementation ease

ArturGajowy: and then, maintainance ease (I really hope we'll rewrite the pretty-printer in the future)

tymm: Nevermind, it was late. I somehow thought you were talking about the grammar and not the PrettyPrinter.

PR #2287, opened 2019-03-20
RCHAIN- 3129 sig validation function

rabbitonweb:
## Overview

1.  Align messages in DeployData
2. Introduce SignatreuAlg as a common interface to (currently) Ed25519 a…
3. Reuse SignaturesAlg in SignatureAlgorithms  …

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3129

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/SignatureAlgorithms.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/SignatureAlgorithms.scala
@@ -1,13 +1,10 @@
 package coop.rchain.casper.util
-import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
+import coop.rchain.crypto.signatures.SignaturesAlg
 
 object SignatureAlgorithms {
   type SignatureAlgorithm = (Array[Byte], Array[Byte]) => Array[Byte]
   @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
   def lookup(algorithm: String): SignatureAlgorithm =
-    algorithm match {
-      case "ed25519"   => Ed25519.sign
-      case "secp256k1" => Secp256k1.sign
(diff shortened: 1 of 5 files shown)


casper/src/main/scala/coop/rchain/casper/util/SignatureAlgorithms.scala
```diff
@@ -1,13 +1,10 @@
...
   type SignatureAlgorithm = (Array[Byte], Array[Byte]) => Array[Byte]
   @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
   def lookup(algorithm: String): SignatureAlgorithm =
-    algorithm match {
-      case "ed25519"   => Ed25519.sign
-      case "secp256k1" => Secp256k1.sign
-      case _           => throw new Exception(s"Unknown signature algorithm $algorithm")
-    }
+    SignaturesAlg(algorithm)
+      .fold(throw new Exception(s"Unknown signature algorithm $algorithm"))(_.sign)
```
KentShikama: Are you sure this won't eagerly throw?

rabbitonweb: it will, is it a bad thing?

KentShikama: I'm not sure how fold is implemented but I thought it does something like calculate the default value and then pass that into the accumulator. `_.sign` then just ignores the value of the accumulator but it doesn't change the fact that the `throw new Exception` was executed. So I guess the exception isn't thrown unless it is returned?

odeac: scala> val opt = Some(10)
opt: Some[Int] = Some(10)
scala> opt.fold(throw new Exception("aaa")) (identity)
res3: Int = 10

PR #2291, opened 2019-03-20
RCHAIN-3129 sign deploydata

rabbitonweb:
## Overview
Utility function to sign `DeployData`

### JIRA ticket:
RCHAIN-3129

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
+++ b/casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
@@ -1,0 +1,30 @@
+package coop.rchain.casper
+
+import coop.rchain.casper.protocol.DeployData
+import coop.rchain.crypto.{PrivateKey, PublicKey}
+import coop.rchain.crypto.hash.Blake2b256
+import coop.rchain.crypto.signatures._
+import com.google.protobuf.ByteString
+
+object SignDeployment {
+
+  private def fill(
+      deployData: DeployData
+  )(user: PublicKey, sig: Array[Byte], sigAlgorithm: String): DeployData =
+    deployData
+      .withUser(ByteString.copyFrom(user.bytes))
+      .withSig(ByteString.copyFrom(sig))
+      .withSigAlgorithm(sigAlgorithm)
+
+  private def clear(deployData: DeployData): DeployData =
+    fill(deployData)(PublicKey(Array.empty[Byte]), Array.empty[Byte], "")
+
+  def apply(key: PrivateKey, deployData: DeployData): DeployData = {
+    val toSign    = clear(deployData).toByteString.toByteArray
+    val hash      = Blake2b256.hash(toSign)
+    val signature = Ed25519.sign(hash, key.bytes)
+
+    fill(deployData)(PublicKey(Ed25519.toPublic(key.bytes)), signature, "Ed25519".toLowerCase)
+  }
+
+}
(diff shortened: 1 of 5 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -132,7 +132,7 @@ object SystemProcesses {
         case isContractCall(
             produce,
             Seq(RhoType.String("fromPublicKey"), RhoType.ByteArray(publicKey), ack)
-        ) =>
+            ) =>
```
JosDenmark: lololololol

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  private def clear(deployData: DeployData): DeployData =
+    fill(deployData)(PublicKey(Array.empty[Byte]), Array.empty[Byte], "")
+
+  def apply(key: PrivateKey, deployData: DeployData, alg: SignaturesAlg = Ed25519): DeployData = {
+    val toSign    = clear(deployData).toByteString.toByteArray
+    val hash      = Blake2b256.hash(toSign)
+    val signature = alg.sign(hash, key.bytes)
+
+    fill(deployData)(PublicKey(alg.toPublic(key.bytes)), signature, alg.name)
```
odeac: please either change the signature of toPublic to `PrivateKey => PublicKey` or add a new overload with this signature

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
```diff
@@ -0,0 +1,30 @@
...
+      .withSig(ByteString.copyFrom(sig))
+      .withSigAlgorithm(sigAlgorithm)
+
+  private def clear(deployData: DeployData): DeployData =
+    fill(deployData)(PublicKey(Array.empty[Byte]), Array.empty[Byte], "")
+
+  def apply(key: PrivateKey, deployData: DeployData, alg: SignaturesAlg = Ed25519): DeployData = {
+    val toSign    = clear(deployData).toByteString.toByteArray
+    val hash      = Blake2b256.hash(toSign)
+    val signature = alg.sign(hash, key.bytes)
```
odeac: Since the long term goal is to avoid using Array[Byte] but to use Private/PublicKey, please add an overload sign : (Array[Byte], PrivateKey) => Array[Byte]

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
rabbitonweb: Todays's discussion! That's why I'm so frustrated with the whole formatting discussion :)

rabbitonweb:
```
method should not take no assumptions of what's already in DeployData
this way solution is context freee
you give me any DeployData, and I sign it, and it is valid after that
i dont have to do any checks what youy've potentially filled there before
purge, sign, fill, return
```

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
rabbitonweb: fill free to update this @odeac "allow eidts form mainneres" is checked
i'm after 12h watch, i'm signing off for today

ArturGajowy:
Guess we can close this, It's going to be merged with 2293 anyway. Closing now, reopen if needed.

PR #2293, opened 2019-03-20
use Public/PrivateKey instead of Array[Byte]

odeac:
~~Note: this is built on top of #2291  Please review that first. The commit log will be cleaner as soon as the parent PR is merged~~

Just review this insterad of #2291


rabbitonweb:
@odeac you dont have to create new PR to build on top, just edit existing branch

PR #2298, opened 2019-03-21
Reject peers if the maximum number of connection is exceeded

sebtomba:
## Overview
Reject peers if the maximum number of connection is exceeded.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3133

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
@@ -1,6 +1,7 @@
 package coop.rchain.comm.rp
 
 import scala.concurrent.duration._
+import scala.util.Random
 
 import cats._
 import cats.effect._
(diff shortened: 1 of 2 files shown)


comm/src/main/scala/coop/rchain/comm/rp/Connect.scala
```diff
@@ -127,20 +130,23 @@ object Connect {
...
-          }
-    } yield peersAndResonses.filter(_._2.isRight).map(_._1)
+  ): F[List[PeerNode]] = {
+    def find(): F[List[PeerNode]] =
+      for {
+        connections      <- ConnectionsCell[F].read
+        tout             <- RPConfAsk[F].reader(_.defaultTimeout)
+        peers            <- NodeDiscovery[F].peers.map(p => (p.toSet -- connections).toList)
+        responses        <- peers.traverse(p => ErrorHandler[F].attempt(conn(p, tout)))
+        peersAndResonses = peers.zip(responses)
```
KentShikama: So if attempting to connect is asynchronous how come you're waiting for a response here?

Also you have a typo in `peersAndResonses`

sebtomba: @KentShikama Actually, this code is not new and not the significant change of the PR. ;-)
We are using `TransportLayer.send` (Handshake) to connect to other peers. The current implementation of `send` is not _fire-and-forget_. There is a confirmation of delivery. If the message can't be delivered to the remote peer then `TransportLayer.send` returns a `CommError`. But there is no guarantee that the delivered message has been or will get processed by the remote peer (perhaps besides that the message has been added to a message queue).

sebtomba: But now I see that this will be a problem
`NodeDiscovery[F].peers.map(p => (p.toSet -- connections).toList)`

We should take only so many peers we have space left in the connection list. I'll fix that.

sebtomba: bors r+

PR #2302, opened 2019-03-22
Slow cooking

adaszko:
* Updates `.drone.yml` to take into account slow-cooking Drone instance
* Adds a dev dependency on python language server to integration tests

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -26,6 +26,9 @@ __pr_build_condiditions: &pr_build_conditions
 
 __buildenv: &buildenv
   image: rchain/buildenv:latest
+  when:
+    instance:
+      - drone.rchain-dev.tk
 
 __sbtcompileenv: &sbtcompileenv
   <<: *buildenv
@@ -138,7 +141,6 @@ pipeline:
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
 
-
   notify:
     <<: *buildenv
     commands:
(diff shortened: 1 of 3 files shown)


adaszko:
bors try

.drone.yml
```diff
@@ -91,6 +100,9 @@ pipeline:
       - /var/cache/rchain-build/.sbt:/root/.sbt
       - /var/cache/rchain-build/.ivy2:/root/.ivy2
       - /var/run/docker.sock:/var/run/docker.sock
+    when:
+      instance:
+        - drone.rchain-dev.tk
```
ArturGajowy: Please de-duplcate the when blocks. One way to do that is to put them all into the `__buildenv` struct. The only place where an override is needed is the slowcooker build.

Alternatively: isn't there a way to tell drone which .yml to use?

adaszko: > Please de-duplcate the when blocks. One way to do that is to put them all into the __buildenv struct. The only place where an override is needed is the slowcooker build.

It is clearer to just duplicate.  You can't tell immediately that you're looking at an unused pipeline step if you hide away the `when` rule in a distant dict.

> Alternatively: isn't there a way to tell drone which .yml to use?

 There is not AFAIK.

ArturGajowy: > It is clearer to just duplicate. Y

I **vehemently** disagree :) I acknowledge that yaml (in drone's variant) is a shite format for de-duplication, but: added structural noise will simply be ignored by the readers.

> can't tell immediately that you're looking at an unused pipeline step if you hide away the when rule in a distant dict.

If you need the 'where it's run' distinction in steps , change the names so they tell you, or add comments as appropriate.

woky: Since this all is added just for one additional task, it makes sense to de-duplicate it.

PR #2305, opened 2019-03-25
Remove block buffer and instead use dependencyDag

KentShikama:
## Overview

It no longer makes sense to keep the block buffer and
dependencyDag as two separate data structures. Previously,
we needed the block buffer because it stored entire
blocks but now since it only stores hashes, we can extract
the same information from dependencyDag through 

```scala
  private def seenBlockHashes(state: CasperState): Set[BlockHash] =
    state.dependencyDag.parentToChildAdjacencyList.values.toSet.flatten
```

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3147

### Notes


KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/DoublyLinkedDagOperations.scala
```diff
@@ -8,6 +8,9 @@ trait DoublyLinkedDag[A] {
   val parentToChildAdjacencyList: Map[A, Set[A]]
   val childToParentAdjacencyList: Map[A, Set[A]]
   val dependencyFree: Set[A]
+  def contains(element: A): Boolean =
```
rabbitonweb: this one is new?

KentShikama:
bors try

KentShikama:
Unfortunately, the dependencyDag still needs to keep
information about which blocks actually came in and which
hashes are in the dependencyDag because they are dependencies
of other blocks but aren't yet seen. However, inserting this logic into the DoublyLinkedDag feels off. I'm going to close this for now.

PR #2307, opened 2019-03-26
Logging improvements

woky:
This doesn't add any additional logging library/framework. SLF4J is merely now direct, instead of transitive dependency. jul-to-SLF4j is bridge from java.util.logging to SLF4J, see the relevant commit for details. Logstash encoder is to allow for sending log messages to Logstash over network for processing (alerting).

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -23,10 +23,6 @@ import org.slf4j.bridge.SLF4JBridgeHandler
 
 object Main {
 
-  // https://www.slf4j.org/legacy.html#jul-to-slf4j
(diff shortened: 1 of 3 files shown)


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -26,6 +28,11 @@ object Main {
 
   @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
   def main(args: Array[String]): Unit = {
+    // Catch-all for unhandled exceptions. Use only JDK and SLF4J.
+    Thread.setDefaultUncaughtExceptionHandler((thread, ex) => {
+      LoggerFactory.getLogger(getClass).error("Unhandled exception in thread " + thread.getName, ex)
```
lukasz-golebiewski: Nice!

PR #2308, opened 2019-03-27
WIP: RCHAIN-3136

JosDenmark:
## Overview
The last bit of `PoS.rho` that has to be done is generation of the unforgeable name. In the past, the `RegistrySigGen` has been used, but @odeac mentioned that this doesn't work. Instead, we deploy the contract with a print statement to determine the actual value of the first name generated, then use that in the registry insert. After that, tests have to be written for the contract. Hence, WIP.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3136

### DIFF
--- a/casper/src/main/resources/PoS.rho
+++ b/casper/src/main/resources/PoS.rho
@@ -1,0 +1,179 @@
+ /*
+ The table below describes the required computations and their dependencies
+
+ No. | Dependency | Computation method | Result
+ ----+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------
+ 1.  |            | given              | sk = d70491c21a5ba44ffb60a97d9bfbf3f3839b997bde107a3a8ab6872594282740
+ 2.  |            | given              | timestamp = 1553862086627
+ 3.  |            | lastNonce          | nonce = 9223372036854775807
+ 4.  | 1,         | ed25519            | pk = cc87bf7747a8c176714b417ca14a63897b07572876c5e38a7896b6007738ef81
+ 5.  | 4, 2,      | genIds             | uname = Unforgeable(0xf535bfc7781a46c6e5ef5d24beffb62fb5ac8c2f8bd1f7f55abeea2e416100da)
+ 6.  | 3, 5,      | registry           | value = (9223372036854775807, bundle+ {   Unforgeable(0xf535bfc7781a46c6e5ef5d24beffb62fb5ac8c2f8bd1f7f55abeea2e416100da) })
+ 7.  | 6,         | protobuf           | toSign = 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20f535bfc7781a46c6e5ef5d24beffb62fb5ac8c2f8bd1f7f55abeea2e416100da1001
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -210,7 +210,7 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
                 deploy
               )
               _                                      <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
-              injResult                              <- doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
+              injResult                              <- doInj(deploy, runtime)
```
ArturGajowy: :ok_hand:

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1910,4 +1911,136 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+        } yield space.store.toMap
+        Await.result(inspectTask.runToFuture, 3.seconds)
+    }
+    result should be(mutable.HashMap.empty)
+    errorLog.readAndClearErrorVector().unsafeRunSync should be(
+      Vector(MethodArgumentNumberMismatch("toList", 0, 1))
+    )
+  }
+
+  it should "transform Set(1, 2, 3) into a [1, 2, 3]" in {
```
ArturGajowy: All those tests should be expressed in rholang.

ArturGajowy:
I'll make sure the PoS.rho contract is available in the registry in tests and genesis (production).

ArturGajowy:
I've changed the identifier used a in PoS.rho to `PoS` (was MakePoS) to avoid confusion

ArturGajowy:
> The last bit of PoS.rho that has to be done is generation of the unforgeable name. In the past, the RegistrySigGen has been used, but @odeac mentioned that this doesn't work.

I'm not sure what @odeac told you, but RegistrySigGen does work. It only needs the actual unforgable name to be passed - but it's enough to print it from a contract that's been deployed with known `sk, timestamp`, and pass all those to the tool.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -250,7 +250,7 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
             for {
               _         <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
               _         <- runtime.replaySpace.rig(hash, log.toList)
-              injResult <- doInj(deploy, runtime.replayReducer, runtime.errorLog)(runtime.cost)
+              injResult <- doInj(deploy, runtime)
```
lukasz-golebiewski: This line used to forward the `replayReducer` into the `doInj` method. After your change the `doInj` defaults to using the `reducer`

ArturGajowy: good catch! :+1:

PR #2310, opened 2019-03-27
Vaults API revamp

ArturGajowy:
## Overview

This is the first step to align Vaults API to be presented on the demo to what we've discussed during the internal demo.

Demo script updates pending.

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -1,7 +1,7 @@
 //scalapackage coop.rchain.rholang.collection
 
 //Requires ListOps
-//Represents Either[A, B] as ("Left", A) or ("Right", B)
+//Represents Either[A, B] as (false, A) or (true, B)
 
 //Registry info:
 //  sk: 7545b56525751bf79daa6cf3ee5757632ea6f37bfa61e93aafc4449d8510df9b
@@ -26,8 +26,8 @@ new Either, rs(`rho:registry:insertSigned:ed25519`), uriOut, rl(`rho:registry:lo
     // Lifts a value into Either, putting Nil-s into Left and using Right otherwise.
     contract Either(@"fromNillable", @value, @nilValue, return) = {
       match value {
-        Nil => return!(("Left", nilValue))
-        _   => return!(("Right", value))
+        Nil => return!((false, nilValue))
+        _   => return!((true, value))
       }
     } |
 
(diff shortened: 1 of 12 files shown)


casper/src/main/resources/RevVault.rho
```diff
@@ -32,60 +32,64 @@ in {
...
-                    match eitherVault {
-                      ("Right", vault) => vaultMapStore!(vaultMap.set(key, vault))
-                      ("Left",  _    ) => vaultMapStore!(vaultMap)
-                    } |
-                    ret!(eitherVault)
+        contract _getOrCreate(@revAddress, constructor, retCh) = {
+          new revAddressValidCh, revAddressValidEitherCh, getOrCreate in {
+            RevAddress!("validate", revAddress, *revAddressValidCh) |
+            @Either!("fromNillableError <-", *revAddressValidCh, *revAddressValidEitherCh) |
+            @Either!("flatMap <-", *revAddressValidEitherCh, *getOrCreate, *retCh) |
```
tymm: You don't really need `retCh` here, do you?

ArturGajowy: WDYM? I need to signal to the caller that flatMap finished. It might be that it's a left, in which case the 'lambda' won't even be called and won't have a chance to signal a return. Not to mention I should never rely on that :)

tymm: Or the next line should be: `for (_, ret <- retCh; @vaultMap <- vaultMapStore) {` ?

ArturGajowy: no, it shouldn't :) `getOrCreate` is the handle under which a "lambda" to be called by `flatMap` is going to listen. As per the "method call convention" we seem to have developed, the labmda takes as many parameters it needs, plus a return channel. But this is a different return channel than `flatMap` will return to. In fact, each "method caller" must spawn its own return channel and pass it to the "called method".

tymm: Thanks, I see it now.

casper/src/main/resources/Either.rho
```diff
@@ -1,7 +1,7 @@
 //scalapackage coop.rchain.rholang.collection
 
 //Requires ListOps
-//Represents Either[A, B] as ("Left", A) or ("Right", B)
+//Represents Either[A, B] as (false, A) or (true, B)
```
tymm: What was your reasoning behind this change?

PR #2316, opened 2019-03-28
pull-rhoc-balances: Allow minor updates of packages

woky:
Address CVE-2018-16492 in pull-rhoc-balances.


ArturGajowy:
JS packages change so often, that I think having a package-lock makes a lot of sense. Let's update just the lib in question, and any others that need to be updated.

Notice that we most likely have no tests for this, so our only hope of it working is s/b (you?) to test it before/after merging and for it to not change during subsequent builds (due to lack of package-lock).

PR #2325, opened 2019-04-01
Split gRPC Kademlia RPC into client and server

sebtomba:
## Overview
Split `GrpcKademliaRPC` into client and server.

Extract `GrpcServer` class to `shared` project.

Remove caching of gRPC channels in the instance `GrpcKademliaRPC`. Since we can now have up to 3200 peers in the Kademlia table, the caching of many connections may lead to large memory consumption, unexpected infrastructure errors and even memory leaks. The caching of channels introduced some accidental complexity (cache invalidation problem). On the other hand, it is not clear if the caching of channels brings any benefits to the overall performance of the system.

Refactor `NodeDiscovery`:
- Extract `KademliaStore` typeclass and `KademliaHandleRPC` object
- Make `PeerTable` pure.
- Kademlia RPC service (server) is started in NodeRuntime.
- Kademlia discover gets triggered in the loop of `NodeRuntime`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3159


node/src/main/scala/coop/rchain/node/api/package.scala
```diff
@@ -0,0 +1,69 @@
...
+
+import io.grpc.netty.NettyServerBuilder
+import monix.eval.Task
+import monix.execution.Scheduler
+
+package object api {
+
+  // 16 MB is max message size allowed by HTTP2 RFC 7540
+  // grpc and netty can however work with bigger values
+  val maxMessageSize: Int = 16 * 1024 * 1024
```
tymm: So we are restricted to 16 MB messages?

sebtomba: Good catch @tymm ! We must use the gRPC max message size from the configuration (min 256KB, max 10MB and default 4MB). I don't know if these values are sensible. But that's how it is for now in the `Configuration`.

comm/src/main/scala/coop/rchain/comm/discovery/package.scala
```diff
@@ -0,0 +1,36 @@
...
+          KademliaGrpcMonix
+            .bindService(new GrpcKademliaRPCServer(pingHandler, lookupHandler), scheduler)
+        )
+        .build
+    )
+
+  def toPeerNode(n: Node): PeerNode =
+    PeerNode(NodeIdentifier(n.id.toByteArray), Endpoint(n.host.toStringUtf8, n.tcpPort, n.udpPort))
+
+  def toNode(n: PeerNode): Node =
```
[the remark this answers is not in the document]
sebtomba: @rabbitonweb Those are for Kademlia gRPC generated classes. The `ProtocolHelper' deals with generated protocol gRPC classes.

PR #2327, opened 2019-04-02
Fix RegistrySigGen to derive the unforgable name properly

ArturGajowy:
Previously, using RegistrySigGen to obtain the proper registry signature
required a multi-step process of:

1. fixing the (timestamp, sk) pair in the deploy to be signed
2. printing out the resulting unforgable name used for registration
3. passing the uname as parameter to RegistrySigGen

Currently, it's enough to pick a (timestamp, sk) pair, the uname
is going to be properly derived based on their values.

## Overview

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3204

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
@@ -1,8 +1,8 @@
 package coop.rchain.casper.util.rholang
 
 import com.google.protobuf.ByteString
-
 import coop.rchain.casper.protocol.DeployData
+import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.{Blake2b256, Blake2b512Random}
 import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
@@ -177,7 +177,7 @@ object RegistrySigGen {
         .withDeployer(ByteString.copyFrom(deployer))
         .withTimestamp(timestamp)
 
-    val rnd = Blake2b512Random(DeployData.toByteArray(seed))
+    val rnd = Blake2b512Random(DeployData.toByteArray(ProtoUtil.stripDeployData(seed)))
 
     rnd.next()
   }


casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
```diff
@@ -177,7 +177,7 @@ object RegistrySigGen {
         .withDeployer(ByteString.copyFrom(deployer))
         .withTimestamp(timestamp)
 
-    val rnd = Blake2b512Random(DeployData.toByteArray(seed))
+    val rnd = Blake2b512Random(DeployData.toByteArray(ProtoUtil.stripDeployData(seed)))
```
lukasz-golebiewski: The `seed` instantiated in line 175 contains only the deployer and timestamp fields. I cannot see any value in passing it to `stripDeployData` which isn't going to strip anything in this case, but return an identical copy of `seed`

ArturGajowy: This aligns the RNG initialization with production code and makes the unforgable name inserted into the registry computed correctly. So the value, as the commit says, is fixing the RegistrySigGen tool

lukasz-golebiewski: This function generates an unforgeable name, it doesn't insert anything into the registry. 
Does `ProtoUtil.stripDeployData(seed)` return anything different than `seed` in this case?

dckc: I don't understand the failure mode that's being fixed here either.

ArturGajowy: Try singing a contract from scratch in a way that actually allows to resolve it from a the registry. You'll see that without this change, you need to pass to RegistriSigGen all of: timestamp, sk, uname. If you don't pass the uname, it's going to differ from what is actually registered, and the registry url provided by RegSigGen is going to be wrong.

Try with this change and uname is no longer needed.

lukasz-golebiewski: Sounds like a test that could be automated and should be a part of this PR

odeac: I can confirm that Artur is right. The RegistrySigGen stopped working at some point.

Also I agree with Lukasz that we should automate its testing. It might be complicated to do so and I would be happy with just logging a JIRA issue

ArturGajowy:
@lukasz-golebiewski you were right, I somehow glossed over what you state in the comment, being sure "it must change things, because it works now". I've tested again and it works without this change, as one would expect given what you've written.

I think the error we might've been making is using 'sk' as the deploy user, while it is the 'pk' that should go there.

PR #2329, opened 2019-04-02
RCHAIN-3184: Use strongly typed keys

JosDenmark:
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3184

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -1,22 +1,21 @@
 package coop.rchain.casper
 
 import cats._
-import cats.implicits._
-import coop.rchain.catscontrib.ski._
-import coop.rchain.catscontrib.{BooleanF, Catscontrib}, BooleanF._, Catscontrib._
 import cats.effect.concurrent.{Ref, Semaphore}
 import cats.effect.{Concurrent, Sync}
+import cats.implicits._
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.{BlockDagRepresentation, BlockDagStorage, BlockStore}
-import coop.rchain.catscontrib._
-import coop.rchain.casper.protocol._
 import coop.rchain.casper.DeployError._
+import coop.rchain.casper.protocol._
 import coop.rchain.casper.util.ProtoUtil._
 import coop.rchain.casper.util._
 import coop.rchain.casper.util.comm.CommUtil
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
 import coop.rchain.casper.util.rholang._
-import coop.rchain.catscontrib.ListContrib
+import coop.rchain.catscontrib.BooleanF
+import coop.rchain.catscontrib.BooleanF._
+import coop.rchain.catscontrib.ski._
 import coop.rchain.comm.CommError.ErrorHandler
 import coop.rchain.comm.rp.Connect.{ConnectionsCell, RPConfAsk}
 import coop.rchain.comm.transport.TransportLayer
@@ -76,7 +75,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Sync: ConnectionsCell: Trans
         dag <- blockDag
         _ <- validatorId match {
               case Some(ValidatorIdentity(publicKey, _, _)) =>
-                val sender = ByteString.copyFrom(publicKey)
+            val sender = ByteString.copyFrom(publicKey.bytes)
                 handleDoppelganger(b, sender)
               case None => ().pure[F]
             }
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
@@ -1,14 +1,14 @@
 package coop.rchain.casper.util.rholang
 
 import com.google.protobuf.ByteString
-
 import coop.rchain.casper.protocol.DeployData
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.{Blake2b256, Blake2b512Random}
-import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
-import coop.rchain.models.Expr.ExprInstance.{GBool, GByteArray, GInt, GString}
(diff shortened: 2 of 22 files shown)


JosDenmark:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
```diff
@@ -109,8 +109,8 @@ object Args {
 
     val id =
       unforgeableNameStr
-        .map(Base16.unsafeDecode(_))
-        .getOrElse(RegistrySigGen.generateUnforgeableNameId(keyPair._1, timestamp))
+        .map(Base16.unsafeDecode)
+        .getOrElse(RegistrySigGen.generateUnforgeableNameId(keyPair._2, timestamp))
```
KentShikama: What is ._1 and ._2 here?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -76,7 +75,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Sync: ConnectionsCell: Trans
         dag <- blockDag
         _ <- validatorId match {
               case Some(ValidatorIdentity(publicKey, _, _)) =>
-                val sender = ByteString.copyFrom(publicKey)
+            val sender = ByteString.copyFrom(publicKey.bytes)
```
ArturGajowy: please reformat this commit :)

crypto/src/main/scala/coop/rchain/crypto/signatures/Ed25519.scala
```diff
@@ -10,11 +11,11 @@ object Ed25519 extends SignaturesAlg {
   val name: String = "Ed25519".toLowerCase
 
   //TODO: Make use of strongly typed keys
-  def newKeyPair: (Array[Byte], Array[Byte]) = {
+  def newKeyPair: (PrivateKey, PublicKey) = {
```
ArturGajowy: Shouldn't other methods of SignaturesAlg take and return the PrivateKey/PublicKey wrappers as well?

ArturGajowy: I guess I'm fine with sticking to what's done in this PR and scheduling this work for later though.

casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
ArturGajowy: So, are you saying the types won't allow otherwise?

I would be surprised if this change didn't break the RegistrySigGen tool to the extent where it can't produce the right registry url *at all* (vs "you need to provide timestamp, sk, and the uname that's going to be used for registration, after you've fixed timestamp and sk").

TL;DR did you test that you can register a new contract / an old contract under new timestamp + sk + uname, and see the tests pass?

crypto/src/main/scala/coop/rchain/crypto/util/Sorting.scala
```diff
@@ -0,0 +1,13 @@
+package coop.rchain.crypto.util
+
+import coop.rchain.crypto.PublicKey
+import coop.rchain.crypto.codec.Base16
+
+object Sorting {
+
+  implicit val byteArrayOrdering: Ordering[Array[Byte]] =
```
ArturGajowy: can we make this private, and maybe not even implicit?

crypto/src/main/scala/coop/rchain/crypto/util/Sorting.scala
```diff
@@ -0,0 +1,13 @@
...
+    Ordering.by(Base16.encode)
+
+  implicit val publicKeyOrdering: Ordering[PublicKey] =
+    (x: PublicKey, y: PublicKey) => byteArrayOrdering.compare(x.bytes, y.bytes)
```
ArturGajowy: Since Ordering is a ContravariantFunctor, you can make this better like this:

```scala
  import cats.implicits._ //waaaay up in the imports

  val o: Ordering[PublicKey] = Ordering.by(Base16.encode).contramap(_.bytes)
```

Besides satisfying one's FP kink, we also get to not call `.bytes` twice (which begs for an error!) and have no way of swapping the args involuntarily.

:sunglasses:

casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
JosDenmark: Yes, it's a type error otherwise. This is the same bug that I mentioned to you earlier today. I haven't tested it, no. I guess there should be a test or two for `RegistrySigGen`. I'll add some.

crypto/src/main/scala/coop/rchain/crypto/signatures/Ed25519.scala
JosDenmark: They do, actually. They're implemented in `SignaturesAlg`, since they're just wrappers.

casper/src/main/scala/coop/rchain/casper/util/rholang/RegistrySigGen.scala
JosDenmark: @KentShikama `._1` is `PrivateKey`, and `._2` is `PublicKey`.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
JosDenmark: 👍

crypto/src/main/scala/coop/rchain/crypto/util/Sorting.scala
JosDenmark: Just because it isn't used anywhere?

crypto/src/main/scala/coop/rchain/crypto/util/Sorting.scala
JosDenmark: 👍

crypto/src/main/scala/coop/rchain/crypto/util/Sorting.scala
JosDenmark: 👍

PR #2330, opened 2019-04-02
RCHAIN-3181: add toList method

JosDenmark:
This is a result of work done last week that is no longer being used. I don't want this to go to waste though.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3181

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
@@ -29,6 +29,7 @@ import monix.execution.Scheduler.Implicits.global
 import org.scalatest.{Assertion, FlatSpec, Matchers}
 
 import scala.collection.immutable.BitSet
+import scala.collection.mutable
 import scala.collection.mutable.HashMap
 import scala.concurrent.Await
 import scala.concurrent.duration._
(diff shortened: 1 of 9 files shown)


JosDenmark:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1910,4 +1911,136 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
     )
   }
 
+  it should "return an error when `toList` is called with arguments" in {
```
ArturGajowy: we really shouldn't be testing rholang methods from AST level. Maybe in rare cases, where there's no possibility to write given rholang in source form (I'm not even sure we have such cases) - but for sure not for this. Those tests might actually be better done using RhoSpec.

JosDenmark: RhoSpec is too cumbersome to justify the extra effort at this point. I checked that the methods work in the example files.

ArturGajowy: Then we should make those tests take rholang and produce AST instead of writing the AST by hand. Guess we can handle that in the future though, as it's the issue for most tests in that class.

PR #2335, opened 2019-04-04
Rspace/hot store rewrite

lukasz-golebiewski:
## Overview
Part of the RSpace rewrite epic. Aims to provide a new in-memory HotStore lazily reading data from history when needed

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3141

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -12,6 +12,7 @@ import scala.collection.concurrent.TrieMap
 trait HotStore[F[_], C, P, A, K] {
   def getContinuations(channels: Seq[C]): F[Seq[WaitingContinuation[P, K]]]
   def putContinuation(channels: Seq[C], wc: WaitingContinuation[P, K]): F[Unit]
+  def removeContinuation(channels: Seq[C], index: Int): F[Unit]
 
   def getData(channel: C): F[Seq[Datum[A]]]
   def putDatum(channel: C, d: Datum[A]): F[Unit]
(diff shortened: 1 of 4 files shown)


rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -95,6 +110,17 @@ private class InMemHotStore[F[_]: Sync, C, P, A, K](
               case Some(joins) => Applicative[F].pure(joins)
             }
     } yield (res)
+
+  private def removeIndex[E](col: Vector[E], index: Int): F[Vector[E]] =
+    if (col.isDefinedAt(index)) {
+      val (l1, l2) = col splitAt index
+      (l1 ++ (l2 drop 1)).pure[F]
```
dzajkowski: `2 drop 1` tail?

lukasz-golebiewski: ?

dzajkowski: `(l1 ++ (l2.tail)).pure[F]` ?

lukasz-golebiewski: No, this isn't going to work with an empty list

ArturGajowy:
Is there a way we can enable this mechanism right now? My thinking is: if it's ready, let's enable it (guarding it with a flag, maybe, for easy disabling). If it's not, it shouldn't be merged, is dead code, and can prove to require substantial changes for its integration.

lukasz-golebiewski:
If this work isn't divided into smaller, independent chunks we'll end up with an enormous PR which is against one of our rules "is at most 200 lines of code (excluding tests)"

ArturGajowy:
This already does too. All I'm sayin is we should find a way to integrate this as soon as possible.

lukasz-golebiewski:
It doesn't. 
I agree that integrating fast would be the best approach, but unfortunately I don't see how it could be done in this case without a huge PR

PR #2337, opened 2019-04-05
Rspace/history trie implementation

dzajkowski:
## Overview
Introduces a trie store that tracks only key structures without data.

### JIRA ticket:
RCHAIN-3140

### Notes


rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,334 @@
...
+  def commonBytes(l: List[Byte], r: List[Byte]): List[Byte] =
+    (l.view, r.view).zipped.takeWhile { case (ll, rr) => ll == rr }.unzip._1.toList
+
+  def skip(t: Trie, affix: List[Byte]): Skip =
+    Skip(ByteVector(affix), Trie.hash(t))
+
+  implicit class RichByteList(l: List[Byte]) {
+
+    @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+    def splitAtHead: (Byte, List[Byte]) = l match {
```
lukasz-golebiewski: This functionality can be achieved by doing `val x :: xs = list` at call site

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,334 @@
...
+      t
+    } else {
+      skip(t, path)
+    }
+
+  // this mapping is kept explicit on purpose
+  @inline
+  def toInt(b: Byte): Int = {
+    import java.lang.{Byte => JByte}
+    JByte.toUnsignedInt(b)
```
lukasz-golebiewski: `java.lang.Byte.toUnsignedInt(b)` is less cognitively demanding (no import, no rename, one line)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,334 @@
...
+    @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+    def splitAtHead: (Byte, List[Byte]) = l match {
+      case h :: tail => (h, tail)
+      case _         => throw new RuntimeException("missing elements")
+    }
+
+    def toByteVector = ByteVector(l)
+  }
+
+  def isPrefix(affix: ByteVector, path: List[Byte]): Boolean =
```
lukasz-golebiewski: This could be moved to `RichByteList`

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/LeafStore.scala
```diff
@@ -0,0 +1,3 @@
+package coop.rchain.rspace.nextgenrspace.history
+
+trait LeafStore {}
```
lukasz-golebiewski: This isn't used anywhere

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
```diff
@@ -0,0 +1,123 @@
...
+class HistorySpec
+    extends FlatSpec
+    with Matchers
+    with OptionValues
+    with GeneratorDrivenPropertyChecks
+    with InMemoryHistoryTestBase {
+
+  implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny
+
+  "history" should "accept new leafs (insert, update, delete)" in forAll {
```
lukasz-golebiewski: This test looks quite thorough but it seems to me it doesn't test the internal Trie structure. It would also pass if the Trie was implemented without pointer blocks

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -52,214 +51,259 @@ final case class History[K](
...
+          key    = currentPath.drop(common.size)
+          s      = skipOrValue(key, newLeaf)
+          newPB  = pastPB.updated((toInt(common.last), s) :: Nil)
+          node   <- storePointerBlock(newPB)
+          result <- rehash(common.init, elems, node, newLeaf :: s :: Nil)
+        } yield result
+
+      case TriePath(elems :+ (_: Leaf), _, path) => // update value
+        rehash(path, elems, newLeaf, Nil)
+      case _ => throw new RuntimeException("malformed trie")
```
lukasz-golebiewski: Shouldn't you `raiseError`?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -52,214 +51,259 @@ final case class History[K](
...
+                )
+              )
+              .map(_.asLeft)
+          case Vector() => Applicative[F].pure(currentAcc :+ nextLastSeen).map(_.asRight)
+        }
     }
+    FlatMap[F].tailRecM((path, rest, lastSeen, acc))(go)
+  }
 
   @SuppressWarnings(Array("org.wartremover.warts.Throw"))
```
lukasz-golebiewski: These can go away

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistoryTest.scala
```diff
@@ -0,0 +1,113 @@
...
+    } yield ()
+  }
+
+  "leafs under same prefix" should "live in one pointer block" in withEmptyTrie { emptyHistory =>
+    val data = List.range(0, 10).map(zerosAnd).map(k => InsertAction(k, randomBlake))
+    for {
+      newHistory <- emptyHistory.process(data)
+      results    <- data.traverse(action => newHistory.traverseTrie(action.key.bytes.toSeq.toList))
+      _          = results should have size 10
+      _          = results.head._2.partialPath.tail.head shouldBe a[Node]
```
lukasz-golebiewski: `(_, path) = results.head` would read clearer

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistoryTest.scala
```diff
@@ -0,0 +1,114 @@
...
+import TestData._
+import coop.rchain.rspace.Blake2b256Hash
+import coop.rchain.rspace.nextgenrspace.history.History.Trie
+import scodec.bits.ByteVector
+import History.codecTrie
+import cats.implicits._
+
+import scala.util.Random
+
+class HistoryTest extends FlatSpec with Matchers with OptionValues with InMemoryHistoryTestBase {
```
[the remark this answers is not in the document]
dzajkowski: let's discuss as sidebar

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistoryTest.scala
```diff
@@ -0,0 +1,114 @@
...
+  "first leaf in trie" should "create a Skip -> Leaf" in withEmptyTrie { emptyHistory =>
+    val data = insert(_zeros) :: Nil
+    for {
+      newHistory <- emptyHistory.process(data)
+      result     <- newHistory.traverseTrie(_zeros)
```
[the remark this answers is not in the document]
dzajkowski: it's package private. it's part of the internal solution and not the external api.

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
```diff
@@ -0,0 +1,147 @@
...
+import monix.execution.Scheduler.Implicits.global
+
+class HistorySpec
+    extends FlatSpec
+    with Matchers
+    with OptionValues
+    with GeneratorDrivenPropertyChecks
+    with InMemoryHistoryTestBase {
+
+  implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny
```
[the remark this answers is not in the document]
ArturGajowy: I remember you telling me that this is b/c of the database state being shared across shrinks. Given how lightweight the setup is (literally 2 x TrieMap.empty), we could put the setup into the property itself, and then benefit from shrinking.

Just pointing this out, do whatever makes more sense :) I'd need to see how well shrinking works for a big failure scenario in here to tell what I'd actually do.

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistoryTest.scala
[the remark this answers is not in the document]
ArturGajowy: How do you get anything out of the history though? I'd expect there to be at least a `get(key): Trie` method - but maybe I'm wrong :)

Also, I'm so brain-dead after all the traverses I've done that when you say tree.traverse, I expect the tree to contain Option-s and expect an Option[Tree] to be returned. Because, you know... Trees are Foldables :sweat_smile:

How about `findPath(key)`, and having the terminal node be the head of the path instead of duplicated in the tuple?

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
dzajkowski: The next step is to attach it to lmdb.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+          } yield Trie.hash(elements.last)
+
+        case (currentRoot, DeleteAction(k)) =>
+          val currentPath = asBytes(k)
+          for {
+            traverseResult <- traverseTrie(currentPath, currentRoot)
+            (_, triePath)  = traverseResult
+            elements <- triePath match {
+                         case TriePath(elems :+ (_: Leaf), None, path) if currentPath == path =>
+                           rebalanceDelete(currentPath, elems, None)
```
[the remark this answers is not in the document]
ArturGajowy: Of course this just for your consideration as a separate PR

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
```diff
@@ -0,0 +1,138 @@
...
+      Gen
+        .listOfN(32, Arbitrary.arbitrary[Int])
+        .map(ints => TestKey32.create(ints))
+    })
+
+  val arbitrary2TestKey32: Arbitrary[TestKey32] =
+    Arbitrary(Gen.sized { _ =>
+      Gen
+        .listOfN(3, Arbitrary.arbitrary[Int])
+        .map(ints => TestKey32.create(List.fill(29)(0) ++ ints))
```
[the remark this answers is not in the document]
dzajkowski: what's the gain?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+                         case _ =>
+                           Applicative[F].pure[List[Trie]](Nil)
+                       }
+            _ <- historyStore.put(elements)
```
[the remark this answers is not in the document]
dzajkowski: elements are all the created trie nodes and they need to be persisted.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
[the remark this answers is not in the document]
dzajkowski: I'm sure you're right, need to think about it.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def create(): PointerBlock = new PointerBlock(Vector.fill(length)(EmptyTrie))
+
+  def create(first: (Int, Trie)): PointerBlock =
+    PointerBlock.create().updated(List(first))
+
+  def create(first: (Int, Trie), second: (Int, Trie)): PointerBlock =
+    PointerBlock.create().updated(List(first, second))
+
+  def fromVector(vector: Vector[Trie]): PointerBlock = new PointerBlock(vector)
```
ArturGajowy: this method should validate the keys of the vector to be in 0.255, and that none of the values is null

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
```diff
@@ -0,0 +1,43 @@
...
+import coop.rchain.rspace.nextgenrspace.history.History.codecTrie
+import scodec.Codec
+import scodec.bits.BitVector
+import scodec.codecs.{provide, vectorOfN}
+import coop.rchain.shared.AttemptOps._
+
+final case class PointerBlock private (toVector: Vector[Trie]) {
+  def updated(tuples: List[(Int, Trie)]): PointerBlock =
+    new PointerBlock(tuples.foldLeft(toVector) { (vec, curr) =>
+      vec.updated(curr._1, curr._2)
```
ArturGajowy: This should validate that the keys of each `tuple` are valid:
 - index in 0..255
 - value non-null (are there other exclusions?)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
```diff
@@ -0,0 +1,43 @@
...
+  def create(): PointerBlock = new PointerBlock(Vector.fill(length)(EmptyTrie))
+
+  def create(first: (Int, Trie)): PointerBlock =
+    PointerBlock.create().updated(List(first))
+
+  def create(first: (Int, Trie), second: (Int, Trie)): PointerBlock =
+    PointerBlock.create().updated(List(first, second))
+
+  def fromVector(vector: Vector[Trie]): PointerBlock = new PointerBlock(vector)
+  implicit val codecPointerBlock: Codec[PointerBlock] =
```
ArturGajowy: a blank line above would help :)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+  def toInt(b: Byte): Int =
+    java.lang.Byte.toUnsignedInt(b)
+
+  // this mapping is kept explicit on purpose
+  @inline
+  def toByte(i: Int): Byte =
+    i.toByte
+
+  def commonPrefix(l: List[Byte], r: List[Byte]): List[Byte] =
+    (l.view, r.view).zipped.takeWhile { case (ll, rr) => ll == rr }.unzip._1.toList
```
ArturGajowy: For some reason IDEA suggests `.map(_._1).toList` instead `.unzip.toList`. Guess it saves some allocations?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+
+  // this mapping is kept explicit on purpose
+  @inline
+  def toInt(b: Byte): Int =
+    java.lang.Byte.toUnsignedInt(b)
+
+  // this mapping is kept explicit on purpose
+  @inline
+  def toByte(i: Int): Byte =
+    i.toByte
```
ArturGajowy: this will wrap on overflow. E.g. `257.toByte == 1`. I'd expect we want an error in such a case?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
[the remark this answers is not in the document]
ArturGajowy: It's really appealing how the 'pointer block' trick resembles the `Fix` trick in recursion schemes. Also, I feel that finding the target place in the Trie (sometimes creating a new path in it) and rebuilding the tree while coming back are ... the 'pre' and 'post' parts of a single recursive descent (that sounds way more obvious than I suspected :D).

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
ArturGajowy: Us not assuming a specific generator in tests, and the actual ability to test the 'top parts' of the trie and its edge cases (should there be any)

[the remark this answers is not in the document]
dzajkowski: the ability is still there, the test just uses a particular version of the key.

I can add a second test which uses a broader key.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+  type KeyPath = List[Byte]
+
+  def asBytes[K](k: K)(implicit codecK: Codec[K]): List[Byte] =
+    codecK.encode(k).get.bytes.toSeq.toList
+
+  type PointerBlockPointer = Blake2b256Hash
+  type TriePointer         = Blake2b256Hash
+  type LeafPointer         = Blake2b256Hash
```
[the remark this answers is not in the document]
dzajkowski: not really. introducing parameters makes it harder to follow what is going on between the layered abstractions (coming from rspace + istore + itriestore)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+        case (currentRoot, DeleteAction(k)) =>
+          val currentPath = asBytes(k)
+          for {
+            traverseResult <- traverseTrie(currentPath, currentRoot)
+            (_, triePath)  = traverseResult
+            elements <- triePath match {
+                         case TriePath(elems :+ (_: Leaf), None, path) if currentPath == path =>
+                           rebalanceDelete(currentPath, elems, None)
+                         case _ =>
+                           Applicative[F].pure[List[Trie]](Nil)
```
ArturGajowy: Shouldn't this case require the path to end at an `Empty` node, and raise an error if it doesn't? Or at least log.warn? After all, in such a case, somebody was able to address the middle of the tree using the 32-byte key passed through public API, or the implementation has a bug...

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: this isn't a complex util method. it's a helper in a very specific context. making it complex will not help in any way.

I can make it private.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+    FlatMap[F].tailRecM((p1, r, a))(go)
+  }
+
+  def process(actions: List[HistoryAction[K]]): F[History[F, K]] =
+    // TODO this is an intermediate step to reproduce all the trie behavior
+    // will evolve to a fold based implementation with partial tries
+    actions
+      .foldLeftM(this.root) {
+        case (currentRoot, InsertAction(k, value)) =>
+          val currentPath = asBytes(k)
```
ArturGajowy: I'd find `remainingPath` clearer - both here, and in the parameter name in traverseTeee

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
dzajkowski: I'm not convinced about it. we have another implementation of this PointerBlock (this is a shameless copy) which does not do any validation and there were no errors. my impression is that this is not a general purpose pointer block. it's a very particular piece of code that serves a particular purpose.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: does it? I find it harder to follow.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+import cats.implicits._
+sealed trait Trie
+sealed trait NonEmptyTrie                                   extends Trie
+case object EmptyTrie                                       extends Trie
+final case class Skip(affix: ByteVector, hash: TriePointer) extends NonEmptyTrie
+final case class Node(hash: PointerBlockPointer)            extends NonEmptyTrie
+final case class Leaf(hash: LeafPointer)                    extends NonEmptyTrie
+
+final case class TriePath(partialPath: Vector[Trie], conflicting: Option[Trie], path: KeyPath) {
+  def append(affix: KeyPath, t: Trie): TriePath =
```
ArturGajowy: I'd make it:

```scala
TriePath(nodes: Vector[Trie], edges: KeyPath, conflictingSkip: Option[Skip]) {
```

Isn't there an invariant that `nodes.size == edges.size + 1`?

BTW why to we use a Vector for `nodes`? Do we need sparsity or random indexing? Wouldn't a List suffice?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+    this.copy(partialPath = this.partialPath :+ t, path = this.path ++ affix)
+}
+
+object TriePath {
+  def root(conflicting: Trie): TriePath =
+    TriePath(Vector(), Some(conflicting), Nil)
+
+  def start(pp: Vector[Trie], path: KeyPath): TriePath =
+    TriePath(pp, None, path)
+
```
ArturGajowy: The above two methods are not used

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: IOW the domain will not breach 0-255 values so there is little reason to quadruple check if a value is valid (first level - blake hashing, second level - store bytes, third level - read bytes, fourth level convert to list of bytes)

dzajkowski: there is a case to be made that this should not deal in `Int`

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+            case Some(pb) =>
+              Applicative[F]
+                .pure((pb.toVector(toInt(h)), tail, path.append(h :: Nil, node)))
+                .map(_.asLeft)
+            case None => Sync[F].raiseError(new RuntimeException("malformed trie"))
+          }
+        case _ => Sync[F].raiseError(new RuntimeException("malformed trie"))
+      }
+    for {
+      hash   <- historyStore.get(start)
```
ArturGajowy: s/hash/node?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
```diff
@@ -0,0 +1,42 @@
...
+      vec.updated(curr._1, curr._2)
+    })
+
+  def countNonEmpty: Int = toVector.count(_ != EmptyTrie)
+}
+
+object PointerBlock {
+  val length = 256
+
+  def create(): PointerBlock = new PointerBlock(Vector.fill(length)(EmptyTrie))
```
ArturGajowy: Now, contrary to the logical structure - which I very much like - we should probably make sure that EmptyTrie-s are not serialized and that we indeed benefit from the PointerBlock's sparsity.

More importantly though, we should make sure that the EmptyTrie case object is not being instantiated by reflection during deserialization, as this was leading to veeeeery strange hashCode/equals behavior... 

We've seen it in @lukasz-golebiewski's kryo PRs and sth similar seems to be described in https://blog.softwaremill.com/is-your-scala-object-always-a-singleton-cb3fd24a2fd9

ArturGajowy: So I guess making the serialized structure sparse (only including NonEmptyTrie-s) would kill two birds with one stone

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
ArturGajowy: GAH! I lost my comment...

Guess we're lucky to have this copy, now we can have a review for preexisting parts of the design as well :)

> there were no errors.

They might've slipped, they may still be lurking...

> not a general purpose pointer block

:100: and so it makes even more sense to encode usage-specific invariants in it, to catch bugs.

> a very particular piece of code that serves a particular purpose

This doesn't mandate assuming that only valid calls/usages of it are going to be made.

I'm fine with doing an `assert` if you want to avoid F[_]-ing with it ;)

Having said that all, I leave it up to you :)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+      }
+      .map(newRoot => this.copy(root = newRoot))
+
+  private[history] def findPath(key: KeyPath, start: Blake2b256Hash): F[(Trie, TriePath)] = {
+    type Params = (Trie, List[Byte], TriePath)
+    //t: Trie, currentPath: List[Byte], path: TriePath
+    def traverse(params: Params): F[Either[Params, (Trie, TriePath)]] =
+      params match {
+        case (EmptyTrie, _, path) => Applicative[F].pure((EmptyTrie, path)).map(_.asRight)
+        case (l: Leaf, _, path)   => Applicative[F].pure((l, path append (Nil, l))).map(_.asRight)
```
ArturGajowy: I think we should be requiring an empty `key` / `remainingPath` (middle tuple element) in the first two cases. Otherwise we've lost track during navigation and it's a malformed tree / bug

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: random indexing was the thought.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: I assumed that it's not a bug to try delete an element that does not exist.

let me ponder if it makes sense to push warn here

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: edges is a bad name, it's a path. edges indicate that there is no relation between them.

dzajkowski: the more I think about it: nodes is also a bad name. partial path or partial prefix nodes is more precise.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
```diff
@@ -0,0 +1,42 @@
+package coop.rchain.rspace.nextgenrspace.history
+import coop.rchain.rspace.Blake2b256Hash
+import coop.rchain.rspace.nextgenrspace.history.History.codecTrie
+import scodec.Codec
+import scodec.bits.BitVector
+import scodec.codecs.{provide, vectorOfN}
+import coop.rchain.shared.AttemptOps._
+
+final case class PointerBlock private (toVector: Vector[Trie]) {
+  def updated(tuples: List[(Int, Trie)]): PointerBlock =
```
ArturGajowy: It seems that all the Int-s in here should be replaced with Byte-s

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+                  .toVector
+                  .zipWithIndex
+                  .filter(v => v._1 != EmptyTrie)
+                  .head
+
+              val (other, idx) = getOther
+              Applicative[F].pure((keyInit, init, Some(pointTo(toByte(idx), other)))).map(_.asLeft)
+            case Some(pb) => // pb contains 3+ elements, drop one, rehash
+              val updated = pb.updated((toInt(keyLast), EmptyTrie) :: Nil)
+              storePointerBlock(updated)
```
ArturGajowy: why do we `storePointerBlock` here, but not in the previous case? Both cases update the pointerBlock in question

ArturGajowy: also, why doesn't the previous case need a rehash?

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
ArturGajowy: That was not my intention, but I won't be stopping you :)

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+          nextLastSeen: Trie,
+          currentAcc: List[Trie]
+          ) =>
+        currentRest match {
+          case head :+ Node(ptr) =>
+            for {
+              pointerBlock <- pointerBlockStore
+                               .get(ptr)
+                               .map(_.get.updated((toInt(currentPath.last), nextLastSeen) :: Nil))
+              nodeToPB <- storePointerBlock(pointerBlock)
```
ArturGajowy: Since we return all the modified Trie-s to be saved in `process`, Trie has its recursion broken by the PointerBlocks, maybe it'd make sense to move all the `storePointerBlock` calls in there?

Oh wait... There isn't a PointerBlock in the Node, there's a PointerBlockPointer...

Maybe making the PointerBlock sparse would allow its fusion with the Trie.Node. But that's just an IDEA for future.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
ArturGajowy: My problem is that it's unclear which is the labels, and which is the nodes. My proposal is to have path.nodes and path.edges :)

re: random indexing - I guess that's only needed because we operate on the end of the list instead of its head? That's not a big deal, but maybe worth considering a change in the future.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
ArturGajowy: > not a bug to try delete an element that does not exist.

The non-existence should be known based on the traverse result and then we can make `delete` expect the deleted node to be a Leaf.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: but why? is vector a lesser collection?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3226

ArturGajowy: Some dead cell in my mind assumed that Vector is sparse. There drivers for making PointerBlock use a sparse representation that I've mentioned throught the review sitll hold though. We could probably even get rid of PointerBlock and just make Node have a sparse structure of trie pointers

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+                  path.append(affix.toSeq.toList, s)
+                ).asLeft
+            )
+        case (s: Skip, _, path) => //not a prefix
+          Applicative[F].pure((EmptyTrie, path.copy(conflicting = Some(s)))).map(_.asRight)
+        case (node @ Node(ptr), h :: tail, path) =>
+          pointerBlockStore.get(ptr).flatMap {
+            case Some(pb) =>
+              Applicative[F]
+                .pure((pb.toVector(toInt(h)), tail, path.append(h :: Nil, node)))
```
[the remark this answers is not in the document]
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3227

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
dzajkowski: In my opinion "Us not assuming a specific generator in tests" is pointless. there is no gain in having abstract tests in a case where the possible extensions consist of using a different key (which no one should ever do, we want a working system for keys that align with production, not for shorter ones).

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
dzajkowski: Completely disagree.
Every invariant, check, assert costs. We know how this class is being used. 
It's purpose is known. 
It will not change. 
Adding asserts, validations, checks only bloats a perfectly simple piece of code that's easy to comprehend and reason about.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: nodes and edges don't point to "nodes" and "labels"

sorry, I don't follow the need. it's a case class that has pieces of named and typed data. the names express their role. what's the gain of using less specific names?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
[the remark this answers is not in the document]
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3228

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/PointerBlock.scala
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3227

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3227

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
```diff
@@ -0,0 +1,375 @@
...
+                )
+              )
+              .map(_.asLeft)
+          case Vector() => Applicative[F].pure(currentAcc :+ nextLastSeen).map(_.asRight)
+        }
+    }
+    FlatMap[F].tailRecM((path, rest, lastSeen, acc))(go)
+  }
+
+  private def rebalanceDelete(p1: KeyPath, r: Vector[Trie], a: Option[Trie]): F[List[Trie]] = {
```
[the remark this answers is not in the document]
dzajkowski: `Applicative[F].pure((keyInit, init, Some(pointTo(toByte(idx), other)))).map(_.asLeft)`?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: because it's not done. the returned Left will continue the loop

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: because an empty trie is EmptyTrie. it's an edge case.

dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3228

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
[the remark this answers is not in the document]
ArturGajowy: I've no idea what you mean. where does this contain an `a`? Or is it to the other point?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
ArturGajowy: My point was I don't see a difference in terms of path in traversing to a Leaf vs traversing to an Empty. I'd expect both cases to enlarge the path.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
ArturGajowy: guess I'll never learn. My issue with the original names - `path` and `partialPath` is:
 - how is one partial, and the other isn't? Aren't they both 'partial'? (partial as in 'from the start to wherever we managed to get so far')
 - which one has the nodes on the path, and which one their indices in the PointerBlock?

With nodes + edges it's clear which is which. Partiality is kind of expected for me when we talk about paths. And if it applies to both paths, maybe let's have PartialTriePath or just PartialPath instead of TriePath.

Anyway, I voiced why this was causing problems in understanding. What to do with that is up to you.

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
ArturGajowy: I wouldn't be proposing polymorphism if I didn't see potential gains in testability and understandability and non-attachment-to-the-impls of that code. The latter two area at a tradeoff with having it concrete -> easier to grok the whole.

The gain in testability is: it's easier and more robust to grab a default Arbitrary for a smaller type than to tinker with the generator to test at the bottom of the tree.

I don't care, really. You asked, so I explained myself. IMO "pointless" is a bit harsh, and not really (I hope I explained why).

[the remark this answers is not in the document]
dzajkowski: I feel we're running into deep waters of philosophical debate.
let's agree that it's more important to have a close to production test now and I will drop a ticket to revisit this topic.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/History.scala
dzajkowski: ok, I can see that the current names are not expressive enough. I don't agree that the proposed ones are better.

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/history/HistorySpec.scala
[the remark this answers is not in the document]
dzajkowski: https://rchain.atlassian.net/browse/RCHAIN-3229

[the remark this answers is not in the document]
ArturGajowy: That ticket is only needed if you feel my point makes a difference :) I'm aware that it might not after the code has been written with a different approach :)

PR #2355, opened 2019-04-12
RCHAIN-2925 machine verifiable dag

rabbitonweb:
## Overview
New tool that does following: 

```
./bin/rnode --grpc-port 10401 mvdag          
75b0f982aef51f44b767cd257c64749b641d5b67896d4d72479a7550a9927553 cffdadbf5ef137ae85ecf5245a949bb910eb16ff65f8a7eb9c2c56a1bfe00287
cffdadbf5ef137ae85ecf5245a949bb910eb16ff65f8a7eb9c2c56a1bfe00287 f5474499577894cdd1ec78f2d970542f32317d1d770694325df2644977a867e2
```

Some refactoring were made during the process. Next PR will introduce property based test to check that DAGs are exactly the same no matter what and that they can be compared line by line. Also the DAG should be transported in stream fashion if it is to work with bigger DAGs in a future.

### JIRA ticket:
RCHAIN-2925

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -248,11 +248,9 @@ object BlockAPI {
 
     def casperResponse(implicit casper: MultiParentCasper[F]): Effect[F, A] =
       for {
-        dag         <- MultiParentCasper[F].blockDag
-        maxHeight   <- dag.topoSort(0L).map(_.length - 1)
-        startHeight = math.max(0, maxHeight - depth)
-        topoSort    <- dag.topoSortTail(depth)
-        result      <- doIt(casper, topoSort)
+        dag      <- MultiParentCasper[F].blockDag
+        topoSort <- dag.topoSortTail(depth)
+        result   <- doIt(casper, topoSort)
       } yield result
 
     MultiParentCasperRef.withCasper[F, ApiErr[A]](
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -236,34 +236,46 @@ object BlockAPI {
     }
   }
 
-  def visualizeDag[
+  def temp[
       F[_]: Monad: Sync: MultiParentCasperRef: Log: SafetyOracle: BlockStore,
-      G[_]: Monad: GraphSerializer
-  ](
-      d: Option[Int] = None,
-      visualizer: (Vector[Vector[BlockHash]], String) => F[G[Graphz[G]]],
-      stringify: G[Graphz[G]] => String
-  ): Effect[F, VisualizeBlocksResponse] = {
+      A
+  ](depth: Int)(
+      doIt: (MultiParentCasper[F], Vector[Vector[BlockHash]]) => F[ApiErr[A]]
+  ): Effect[F, A] = {
 
     val errorMessage =
       "Could not visualize graph, casper instance was not available yet."
 
-    def casperResponse(implicit casper: MultiParentCasper[F]): Effect[F, VisualizeBlocksResponse] =
+    def casperResponse(implicit casper: MultiParentCasper[F]): Effect[F, A] =
       for {
-        dag                <- MultiParentCasper[F].blockDag
-        maxHeight          <- dag.topoSort(0L).map(_.length - 1)
-        depth              = d.getOrElse(maxHeight)
-        startHeight        = math.max(0, maxHeight - depth)
-        topoSort           <- dag.topoSortTail(depth)
-        lastFinalizedBlock <- MultiParentCasper[F].lastFinalizedBlock
-        graph              <- visualizer(topoSort, PrettyPrinter.buildString(lastFinalizedBlock.blockHash))
-      } yield VisualizeBlocksResponse(stringify(graph)).asRight
+        dag         <- MultiParentCasper[F].blockDag
+        maxHeight   <- dag.topoSort(0L).map(_.length - 1)
+        startHeight = math.max(0, maxHeight - depth)
+        topoSort    <- dag.topoSortTail(depth)
+        result      <- doIt(casper, topoSort)
+      } yield result
 
-    MultiParentCasperRef.withCasper[F, ApiErr[VisualizeBlocksResponse]](
+    MultiParentCasperRef.withCasper[F, ApiErr[A]](
       casperResponse(_),
       Log[F].warn(errorMessage).as(errorMessage.asLeft)
     )
   }
+
+  def visualizeDag[
+      F[_]: Monad: Sync: MultiParentCasperRef: Log: SafetyOracle: BlockStore,
+      G[_]: Monad: GraphSerializer
+  ](
+      depth: Int,
+      visualizer: (Vector[Vector[BlockHash]], String) => F[G[Graphz[G]]],
+      stringify: G[Graphz[G]] => String
+  ): Effect[F, VisualizeBlocksResponse] =
+    temp[F, VisualizeBlocksResponse](depth) {
+      case (casper, topoSort) =>
+        for {
+          lastFinalizedBlock <- casper.lastFinalizedBlock
+          graph              <- visualizer(topoSort, PrettyPrinter.buildString(lastFinalizedBlock.blockHash))
+        } yield VisualizeBlocksResponse(stringify(graph)).asRight[Error]
+    }
 
   def showBlocks[F[_]: Monad: MultiParentCasperRef: Log: SafetyOracle: BlockStore](
       depth: Int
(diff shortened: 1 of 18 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -239,73 +239,80 @@ object BlockAPI {
...
-        startHeight        = math.max(0, maxHeight - depth)
-        topoSort           <- dag.topoSortTail(depth)
-        lastFinalizedBlock <- MultiParentCasper[F].lastFinalizedBlock
-        graph              <- visualizer(topoSort, PrettyPrinter.buildString(lastFinalizedBlock.blockHash))
-      } yield VisualizeBlocksResponse(stringify(graph)).asRight
-
-    MultiParentCasperRef.withCasper[F, ApiErr[VisualizeBlocksResponse]](
+        dag       <- MultiParentCasper[F].blockDag
+        maxHeight <- dag.topoSort(0L).map(_.length - 1)
+        depth     = maybeDepth.getOrElse(maxHeight)
```
odeac: `dag.topoSort` sounds expensive and you shouldn't be doing it unless maybeDepth is None.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,44 @@
...
+    import VerifiableEdge._
+
+    toposort
+      .foldM(List.empty[VerifiableEdge]) {
+        case (acc, blockHashes) =>
+          for {
+            parents          <- blockHashes.toList.traverse(fetchParents)
+            blocks           = blockHashes.toList.map(_.show)
+            blocksAndParents = blocks.zip(parents.map(_.map(p => p.show)))
+            entries = blocksAndParents.flatMap {
```
odeac: I think this can be simplified. The zip seems unnecessary. Something like the code below should also work but I haven't tried to compile it

```scala
blockHashes.toStream
    .traverse ( h=> fetchParents(h).map(p => (h.show, p.show)))
    .map(blocksAndParents => 
        val entries = blockAndParents.flatMap {...}
        entries ++ acc
    )
...
```

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -79,9 +77,19 @@ private[api] object DeployGrpcService {
...
+        defer(BlockAPI.machineVerifiableDag[F])
+
       override def showBlocks(request: BlocksQuery): Observable[GrpcEither] =
         Observable
-          .fromTask(deferList(BlockAPI.showBlocks[F](request.depth)))
+          .fromTask(
+            deferList[BlockInfoWithoutTuplespace](
+              Functor[F].map(BlockAPI.showBlocks[F](Some(request.depth))) {
+                case Right(list) => list
+                case Left(_)     => List.empty
```
odeac: Bifunctor's leftMap(_ => List.empty) is less verbose and less error prone than the explicit match

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
rabbitonweb: 👍

itegulov: We should not be calling topoSort for finding out which block is the highest anyway (this is a ridicilous overkill). I have ticketed this https://rchain.atlassian.net/browse/RCHAIN-3253

rabbitonweb: Yes, I've even added TODO for that. Good there is a ticket. But for now we need to know the depth. This part of code is just extracting what was common in two methods already (`showBlocks` and `vdag`) and was about to hit the third method (`machineVerifiableDag`)

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -16,4 +16,8 @@ object ByteStringOps {
       buffer
     }
   }
+
+  implicit class RichBytesArray(arr: Array[Byte]) {
```
tymm: `RichByteArray`

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
rabbitonweb: 👍 not sure, if more readable but commit was added

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
rabbitonweb: 👍

casper/src/test/scala/coop/rchain/casper/ConstructDeploy.scala
```diff
@@ -4,7 +4,7 @@ import cats._
 import cats.implicits._
 import com.google.protobuf.ByteString
 import coop.rchain.casper.protocol._
-import coop.rchain.crypto.PrivateKey
+import coop.rchain.crypto.{PrivateKey, PublicKey}
```
tymm: I think this import is not needed.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
+package coop.rchain.casper.api
+
+import coop.rchain.casper._
+import coop.rchain.casper.protocol._
```
tymm: Not needed.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
...
+import coop.rchain.casper.util.ProtoUtil
```
tymm: Not needed.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
...
+import coop.rchain.crypto.codec.Base16
+import coop.rchain.blockstorage.BlockStore
```
tymm: Not needed.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
...
+import cats._, cats.data._, cats.implicits._
+
+final case class VerifiableEdge(from: String, to: String)
+
+object VerifiableEdge {
+
+  implicit def showBlockHash: Show[BlockHash] = new Show[BlockHash] {
```
tymm: You could remove the `new Show[BlockHash]` but personally don't mind.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
...
+    def show(blockHash: BlockHash): String =
+      Base16.encode(blockHash.toByteArray)
+  }
+
+  implicit def showVerifiableEdge: Show[VerifiableEdge] = new Show[VerifiableEdge] {
```
tymm: Same as with `showBlockHash`.

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -0,0 +1,47 @@
+package coop.rchain.casper.api
+
+import coop.rchain.casper._
+import coop.rchain.casper.protocol._
+import coop.rchain.casper.util.ProtoUtil
+import coop.rchain.crypto.codec.Base16
+import coop.rchain.blockstorage.BlockStore
+import cats._, cats.data._, cats.implicits._
```
tymm: IntelliJ says `cats.data._` is not needed as well.

tymm: Same in `MachineVerifiableDagSpec.scala` btw.

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -79,9 +77,18 @@ private[api] object DeployGrpcService {
...
 
+      override def machineVerifiableDag(q: MachineVerifyQuery): Task[GrpcEither] =
+        defer(BlockAPI.machineVerifiableDag[F])
+
       override def showBlocks(request: BlocksQuery): Observable[GrpcEither] =
         Observable
-          .fromTask(deferList(BlockAPI.showBlocks[F](request.depth)))
+          .fromTask(
+            deferList[BlockInfoWithoutTuplespace](
+              Functor[F].map(BlockAPI.showBlocks[F](Some(request.depth)))(
```
tymm: Weird that this hack is needed to make `map` work. Let me know when you find out why that is :)

casper/src/test/scala/coop/rchain/casper/ConstructDeploy.scala
rabbitonweb: 👍

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
rabbitonweb: 👍

rabbitonweb:
I've pushed changes to adhere to review, rebase was needed. @odeac pls have a second look. @itegulov and @tymm thx for review

PR #2356, opened 2019-04-12
Rchain 3136: implement PoS contract based on RevVault

odeac:
## Overview
see title

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3136

### Notes


casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,162 @@
...
+
+            contract PoS (@"bond", @amount, returnCh) = {
+              new userCh, depositCh in {
+                getUser!(*userCh) |
+
+                for(@bonds <- bondsCh;
+                    @userPk <- userCh) {
+
+                  if (bonds.contains(userPk)) {
+                    returnCh!((false, "Public key is already bonded."))
```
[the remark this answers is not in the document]
odeac: good catch!

odeac:
bors try

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
-                    match depositResult {
-                      (true, Nil) => { paymentStore!([codeHash, userPk, timestamp, phloRate * amount]) | return!(phloRate *amount) }
-                      (false, err) => { paymentStore!([codeHash, userPk, timestamp, 0]) | return!(0) }
-                    }
-                  }
+            }
+          }
+        } |
+
+        contract deposit (@fromUserPk, @amount, returnCh) = {
```
KentShikama: Interesting; so we do a "push" transfer (as opposed to "pull" transfer) from within the PoS contract. Basically making use of the transitive dependency attack.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
-                        @RevVault!("deployerAuthKey", *authKeyCh)
-                        | for (userAuthKey <- authKeyCh){
-                          @vault!("transfer", posRevAddress, amount, *userAuthKey, *return)
-                        }
-                      }
-                    }
-                  }
-                }
-              }
+            contract PoS (@"pay", @amount, returnCh) = {
```
KentShikama: What does this PoS.pay do? Is this supposed to be the basic template for rewards that will be expanded later?

KentShikama: Oh nvm...; this will conflict with

https://github.com/rchain/rchain/blob/f4980d38f183d80b02ecd9a3fe59ca64d2552f93/casper/src/main/resources/PoS.rho#L155

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+  registryLookup!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
+
+  for(@(_, Either) <- eitherCh;
+      @(_, RevVault) <- revVaultCh) {
+    new posRevAddressCh in {
+  //TODO - set the user properly when running PoSTest.rho
+  //    getCurrentUserAddress!(*posRevAddressCh)|
+
+      revAddressOps!("fromPublicKey", "0000000000000000000000000000000000000000000000000000000000000000".hexToBytes(), *posRevAddressCh) |
+
```
JosDenmark: This doesn't match the generated public key.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+    registryLookup(`rho:registry:lookup`),
+    stdout(`rho:io:stdout`),
+    getParameters(`rho:deploy:params`),
+    revAddressOps(`rho:rev:address`),
+    posRevAddressCh,
+    eitherCh,
+    revVaultCh,
+    getUser,
+    getCurrentUserAddress,
+    getCurrentUserVault,
```
JosDenmark: You should introduce names in the lexical scope that directly encloses their usage. You don't need all of these at the very top of the file because quite a few aren't used.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+      @(_, RevVault) <- revVaultCh) {
+    new posRevAddressCh in {
+  //TODO - set the user properly when running PoSTest.rho
+  //    getCurrentUserAddress!(*posRevAddressCh)|
+
+      revAddressOps!("fromPublicKey", "0000000000000000000000000000000000000000000000000000000000000000".hexToBytes(), *posRevAddressCh) |
+
+      for(@posRevAddress <- posRevAddressCh) {
+
+        contract MakePoS(@minimumBond, @maximumBond, @initialBonds, posReturnCh) = {
```
JosDenmark: Why did you reintroduce the `initialBonds` argument?

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+                    returnCh!((false, "Public key is already bonded."))
+                  } else if (amount < minimumBond) {
+                    bondsCh!(bonds) |
+                    returnCh!((false, "Bond less than minimum!"))
+                  } else if (amount > maximumBond) {
+                    bondsCh!(bonds) |
+                    returnCh!((false, "Bond greater than maximum!"))
+                  } else {
+                    deposit!(userPk, amount, *depositCh) |
+                    for(@depositResult <- depositCh) {
```
JosDenmark: In this implementation, the bonds map is always updated, regardless of whether or not the deposit fails. You should only update the bonds map if the deposit succeeds.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+        } |
+
+        contract deposit (@fromUserPk, @amount, returnCh) = {
+          new vaultCh,
+              revAddressCh,
+              authKeyCh in {
+                stdout!("get deployerauthkey") |
+            @RevVault!("deployerAuthKey", *authKeyCh) |
+            for (@authKey <- authKeyCh) {
+              stdout!({"deployerauthkey" : authKey}) |
```
JosDenmark: Please remove all unnecessary print statements.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+4.  | 1,         | ed25519            | pk = cc87bf7747a8c176714b417ca14a63897b07572876c5e38a7896b6007738ef81
+5.  | 4, 2,      | genIds             | uname = Unforgeable(0xf974f799067b34be1acb7b565e0b160edb23df3e3ad58f200af183ae48b458cd)
+6.  | 3, 5,      | registry           | value = (9223372036854775807, bundle+ {   Unforgeable(0xf974f799067b34be1acb7b565e0b160edb23df3e3ad58f200af183ae48b458cd) })
+7.  | 6,         | protobuf           | toSign = 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20f974f799067b34be1acb7b565e0b160edb23df3e3ad58f200af183ae48b458cd1001
+8.  | 7, 1,      | ed25519            | sig = 7ff160accda7f76f14f2ae132aaae140ac22b2671e561a58a566bff1a01e49cf372201a5df678aba5ff09171b6af966906deffbf8933efeeddb69a4c204b0503
+9.  | 4,         | registry           | uri = rho:id:cnec3pa8prp4out3yc8facon6grm3xbsotpd4ckjfx8ghuw77xadzt
+----+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------
+*/
+
+new MakePoS,
```
JosDenmark: Why did you name this contract `MakePoS` when there is already a `MakePoS` contract? Also, you use `PoS` in the body of the contract, so why not just keep it as it was originally?

casper/src/main/resources/PoS.rho
JosDenmark: It looks like `lastPayment` in the `MakePoS` contract is only used here [https://github.com/rchain/rchain/blob/f4980d38f183d80b02ecd9a3fe59ca64d2552f93/casper/src/test/scala/coop/rchain/casper/MultiParentCasperDeploySpec.scala#L188](url). Once we have vaults and gas linked, we won't need this test anymore, so it's ok to remove.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
-                      return!((true, bondAmount))
-                    }
-                  }
-
-                  // TODO: figure out rate-limiting of bonding
-                  | contract validateBondingRate (@bondAmount, return) = {
-                    return!((true, bondAmount))
-                  }
-
-                  | contract validatePublicKey (@bondAmount, return) = {
```
JosDenmark: Did you remove this check because key verification is performed by `RevVault`?

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
               }
             }
           }
         }
+      } |
+
+      contract getUser (returnCh) = {
+        new parametersCh in {
+          getParameters!(*parametersCh) |
+          for (@codeHash, @phloRate, @userPk, @timestamp <- parametersCh) {
```
JosDenmark: Please replace these unused variables with wildcards.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+            returnCh!(userPk)
+          }
+        }
+      } |
+
+      contract getCurrentUserAddress(returnCh) = {
```
JosDenmark: Do you use this? It looks like you intended to on line 39, but your TODO doesn't explain why you didn't.

casper/src/main/resources/PoS.rho
```diff
@@ -1,180 +1,165 @@
...
+    } |
+
+    new uriCh in {
+      registryInsert!(
+        "cc87bf7747a8c176714b417ca14a63897b07572876c5e38a7896b6007738ef81".hexToBytes(),
+        (9223372036854775807, bundle+{*MakePoS}),
+        "7ff160accda7f76f14f2ae132aaae140ac22b2671e561a58a566bff1a01e49cf372201a5df678aba5ff09171b6af966906deffbf8933efeeddb69a4c204b0503".hexToBytes(),
+        *uriCh
+      ) |
+      for (@uri <- uriCh) {
```
JosDenmark: What is the point of this continuation?

casper/src/main/resources/PoS.rho
odeac: fixed

casper/src/main/resources/PoS.rho
odeac: I didn't realise it was removed. I'll remove that

casper/src/main/resources/PoS.rho
odeac: I moved getUser and getCurrentUserAddress out of the minimal lexical scope because they are utility functions and they could be used as utility functions

casper/src/main/resources/PoS.rho
odeac: First of all this is just a local name so it doesn't affect anybody else except the reader of this file.

I named the top-level contract like this because it "makes a PoS contract". The returned value is the actual "PoS".  For me it reads better but I don't have a strong opinion about it.

casper/src/main/resources/PoS.rho
odeac: The bondsCh acts as a MVar so I take the map value, use it and then I put it back in the channel. If it changes I put back in the channel the updated value. If it stays the same I put back in the channel the original value.

If I don't put the value back even on errors the pos instance would become unusable after an error.

casper/src/main/resources/PoS.rho
odeac: This contract checks if the key is already bonded. This check exists in my implementation too. See line 65

casper/src/main/resources/PoS.rho
odeac: I don't understand this thread. Please explain.

casper/src/main/resources/PoS.rho
odeac: fixed

casper/src/main/resources/PoS.rho
odeac: it's also used in getCurrentUserVault which is used to create the vault when the PoS instance is created

casper/src/main/resources/PoS.rho
odeac: fixed

casper/src/main/resources/PoS.rho
odeac: It's for debugging reasons. I had some situations when the contract was not registered in the registry and because if this the tests failed without too much clues for the poor developer. Then I started to add these continuations that print to the stdout what contracts were registered and their URIs. On a second thought, maybe I should move this log message in scala implementation of registry insertSigned.

I'll remove it though.

casper/src/main/resources/PoS.rho
JosDenmark: I'll rephrase. You're returning an updated map to the `bondsCh` whether or not `deposit` succeeds. See the lines below. Explicitly, if deposit fails, you should return `bonds`. If it succeeds, you should should return what you're returning now to `bonds`.

odeac: good point! I'll add a test for that case too

casper/src/main/resources/PoS.rho
```diff
@@ -1,181 +1,176 @@
...
+7.  | 6,         | protobuf           | toSign = 2a3eaa013b0a0d2a0b10feffffffffffffffff010a2a5a280a243a220a20f974f799067b34be1acb7b565e0b160edb23df3e3ad58f200af183ae48b458cd1001
+8.  | 7, 1,      | ed25519            | sig = 7ff160accda7f76f14f2ae132aaae140ac22b2671e561a58a566bff1a01e49cf372201a5df678aba5ff09171b6af966906deffbf8933efeeddb69a4c204b0503
+9.  | 4,         | registry           | uri = rho:id:cnec3pa8prp4out3yc8facon6grm3xbsotpd4ckjfx8ghuw77xadzt
+----+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------
+*/
+
+new MakePoS,
+    registryInsert(`rho:registry:insertSigned:ed25519`),
+    registryLookup(`rho:registry:lookup`),
+    getParameters(`rho:deploy:params`),
```
JosDenmark: `getParameters` is only used in `getUser`. Please introduce it in with `parametersCh`.

casper/src/main/resources/PoS.rho
```diff
@@ -1,181 +1,176 @@
...
+    deposit in {
+
+  registryLookup!(`rho:id:j6trahbxycumerwpr5qype7j43b7eqh8auebwsa9nn68if47gswh73`, *eitherCh) |
+  registryLookup!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
+
+  for(@(_, Either) <- eitherCh;
+      @(_, RevVault) <- revVaultCh) {
+    new posRevAddressCh in {
+  //TODO - set the user properly when running PoSTest.rho
+  //    getCurrentUserAddress!(*posRevAddressCh)|
```
JosDenmark: This is never used. Should it be removed?

casper/src/main/resources/PoS.rho
```diff
@@ -1,181 +1,176 @@
...
+    getMVar,
+    runMVar,
+    deposit in {
+
+  registryLookup!(`rho:id:j6trahbxycumerwpr5qype7j43b7eqh8auebwsa9nn68if47gswh73`, *eitherCh) |
+  registryLookup!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
+
+  for(@(_, Either) <- eitherCh;
+      @(_, RevVault) <- revVaultCh) {
+    new posRevAddressCh in {
```
JosDenmark: Case in point. You've duplicated the `posRevAddressCh` here. Please de-duplicate the name and ensure that all names are introduced in a minimal context.

casper/src/main/resources/PoS.rho
```diff
@@ -1,181 +1,176 @@
...
+    getCurrentUserAddress,
+    getCurrentUserVault,
+    getMVar,
+    runMVar,
+    deposit in {
+
+  registryLookup!(`rho:id:j6trahbxycumerwpr5qype7j43b7eqh8auebwsa9nn68if47gswh73`, *eitherCh) |
+  registryLookup!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
+
+  for(@(_, Either) <- eitherCh;
```
JosDenmark: It looks like this dependency can be eliminated 🎉. Please remove this continuation and `eitherCh`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -210,25 +210,20 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
   private def deployPaymentSource(amount: Long, name: String = "__SCALA__"): String =
-    s"""
-       | new rl(`rho:registry:lookup`), poSCh in {
-       |   rl!(`rho:id:cnec3pa8prp4out3yc8facon6grm3xbsotpd4ckjfx8ghuw77xadzt`, *poSCh) |
-       |   for(@(_, PoS) <- poSCh) {
-       |     @PoS!("pay", $amount, "$name")
-       |   }
-       | }
-       """.stripMargin
+    s"""@"$name"!((true, Nil))"""
```
JosDenmark: Why was this eliminated/changed? Why are you planning to pull `PoS` from `SystemInstances`? It should be taken from the registry like the existing implementation suggests. Please revert this change.

casper/src/test/resources/PoSTest.rho
```diff
@@ -4,31 +4,248 @@
...
+  "2222222222222222222222222222222222222222222222222222222222222222".hexToBytes()
 ) {
 
-  (genesisPubKey, genesisRevAddress) => {
+  (genesisPubKey, user1PubKey, user2PubKey) => {
+    new
+      rl(`rho:registry:lookup`),
+      stdlog(`rho:io:stdlog`),
+      stdout(`rho:io:stdout`),
+      getParameters(`rho:deploy:params`),
```
JosDenmark: Please remove `getParameters`.

casper/src/test/resources/PoSTest.rho
```diff
@@ -4,31 +4,248 @@
...
+        stdlog!("info", {"posRevAddress":posRevAddress, "user1RevAddress":user1RevAddress}) |
 
-    new rl(`rho:registry:lookup`),
-        RhoSpecCh,
-        PosCh,
-        setup,
-        test,
-        test_different_deployer,
-        ackCh in {
+        @RevVault!("findOrCreateGenesisVault", posRevAddress, 10000, *posVaultCh) |
```
JosDenmark: I don't think it's necessary to instantiate the `PoS` vault here. REV won't be leaving the `PoS` vault, and it's instantiated in the `PoS` constructor.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -71,7 +71,7 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
     result.status.isFailed should be(true)
   }
 
-  "computeDeployPayment" should "transfer REV from the deployer's vault to the PoS vault" in withTestFixture {
+  "computeDeployPayment" should "transfer REV from the deployer's vault to the PoS vault" ignore withTestFixture {
```
JosDenmark: Please explain why the changes to runtime manager are needed.

casper/src/test/resources/PoSTest.rho
JosDenmark: It is created. It's created in line 43 of `PoS.rho`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: No. The existing code does not rely on `SystemRegistry`. The changes you made should leave everything in tact, including the RuntimeManager. If your changes make something break in RuntimeManager, then there's something wrong with your changes. Please revert this and all other changes to RuntimeManager.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
JosDenmark: There shouldn't be any changes to runtime manager. It's working against the `PoS.rho` interface, which hasn't changed from what it was before you took over.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
```diff
@@ -41,4 +41,12 @@ object RhoType {
       }
   }
 
+  object Tuple2 {
+    def apply(b: Boolean) = Expr(GBool(b))
```
JosDenmark: I'm confused why the Tuple2 `apply` method takes a boolean and returns a Par that's a boolean. It should take a pair `(Par, Par)` and return a `Par`.

odeac:
closed by mistake

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
odeac: You are right. This is a copy paste error.

odeac:
bors try

casper/src/main/resources/PoS.rhox
```diff
@@ -0,0 +1,164 @@
...
+    getCurrentUserVault,
+    getMVar,
+    runMVar,
+    deposit in {
+  registryLookup!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
+
+  for(@(_, RevVault) <- revVaultCh) {
+    new posRevAddressCh in {
+
+      revAddressOps!("fromPublicKey", "cc87bf7747a8c176714b417ca14a63897b07572876c5e38a7896b6007738ef81".hexToBytes(), *posRevAddressCh) |
```
ArturGajowy: shouldn't we parameterize the rev address as well?

casper/src/main/resources/PoS.rhox
```diff
@@ -0,0 +1,164 @@
...
+
+      for(@posRevAddress <- posRevAddressCh) {
+        new bondsCh in {
+          bondsCh!({}) |
+          getCurrentUserVault!(Nil) |
```
ArturGajowy: We should not use Nil as a return channel - it's going to linger there. Also, what's the purpose of this call?

ArturGajowy: If it is creation of the PoS vault, let's just do `findOrCreate!(revAddress)` directly without resolving the address yet again

casper/src/main/resources/PoS.rhox
```diff
@@ -0,0 +1,164 @@
...
+
+          contract PoS (@"getBonds", returnCh) = {
+            getMVar!(*bondsCh, *returnCh)
+          } |
+
+          contract PoS (@"bond", @amount, returnCh) = {
+            new userCh, depositCh, processCh in {
+              runMVar!(*bondsCh, *processCh, *returnCh) |
+
+              getUser!(*userCh) |
```
ArturGajowy: the `getUser` method will not be possible in the future, so the PoS contract should be accepting an explicit `posAuth` token, similar to RevVault.

casper/src/main/resources/PoS.rhox
```diff
@@ -0,0 +1,164 @@
...
+                  resultCh!(bonds, (false, "Bond is less than minimum!"))
+                } else if (amount > $$maximumBond$$) {
+                  resultCh!(bonds, (false, "Bond is greater than maximum!"))
+                } else {
+                  deposit!(userPk, amount, *depositCh) |
+                  for(@depositResult <- depositCh) {
+                    match depositResult {
+                      (true, _) => {
+                        resultCh!(
+                          bonds.set(userPk,[amount, bonds.size() + 1]),
```
ArturGajowy: does the `bonds.size() + 1` act as a sequence number for things in the bonds map? Does the bonds map ever shrink?

casper/src/main/resources/PoS.rhox
```diff
@@ -0,0 +1,164 @@
...
+          }
+        } |
+
+        contract deposit (@fromUserPk, @amount, returnCh) = {
+          new vaultCh,
+              revAddressCh,
+              authKeyCh in {
+            @RevVault!("deployerAuthKey", *authKeyCh) |
+            for (@authKey <- authKeyCh) {
+              revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
```
ArturGajowy: this relies direclty on the 'drain vault attack' that is going to be prevented. Instead of doing the deposit for the bonded party, we should let them do that, notify us when it's done, check balance, and proceed accordingly.

This also means each bonding request should receive a new, one-time-only, unforgrable-name-bound RevVault to which the bonding amount would be transferred by the party, and from it to the PoS vault.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -32,16 +34,31 @@ final case class ProofOfStake(minimumBond: Long, maximumBond: Long, validators:
...
   lazy val term: Par = ParBuilder[Coeval].buildNormalizedTerm(code).value()
 
+  /**
+    * Loads code from a resource file while doing macro substitution with the provided values.
+    * The macros have the format $$macro$$
+    * @param file
+    * @param env a sequence of pairs macro -> value
+    * @return
+    */
+  private def loadTemplateFromResources(file: String, env: (String, Any)*): String = {
```
ArturGajowy: let's move that to CompileRholangSource, it's a general-purpose util

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -32,16 +34,31 @@ final case class ProofOfStake(minimumBond: Long, maximumBond: Long, validators:
...
+    val originalContent = Source.fromResource(file).mkString
+
+    val finalContent = env.foldLeft(originalContent) {
+      case (content, (name, value)) => content.replace(s"""$$$$$name$$$$""", value.toString)
+    }
+
+    s"""//Loaded from resource file <<$file>>
+       |$finalContent
+       |""".stripMargin
```
ArturGajowy: this should probably be

```scala
    s"""//Loaded from resource file <<$file>>
       #$finalContent
       #""".stripMargin('#')
```

or else we have veeeeery weird errors as soon as s/b leaves a `|` as a first character in a line (no matter how deeply indented). We've had quite some debugging with @JosDenmark around exactly this.

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+                    @user1Vault!("balance", *finalUserBalanceCh) |
+                    for(@finalPosBalance <- finalPosBalanceCh;
+                        @finalUserBalance <- finalUserBalanceCh;
+                        @bonds <- retCh) {
+                      stdlog!("info", {"finalPosBalance": finalPosBalance, "finalUserBalance":finalUserBalance, "bonds":bonds}) |
+                      match bonds.get(user1PubKey) {
+                        [bondAmount, _] => {
+                          rhoSpec!("assertMany",
+                            [
+                              ((100, "==", bondAmount), "the new bond is expected in the final map"),
```
ArturGajowy: That's a lot of nesting that can be avoided by ways similar to `assertBalances` in RevVaultTest.rho (we pass a map of balances to assert, we pass a "now's your turn" channel as the first parameter and wait till triggered to do the assertions

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+                }
+              }
+            }
+          } |
+
+          contract test_multiple_bonding_succeeds(rhoSpec, _, ackCh) = {
+            new setupCh, retCh, initialPosBalanceCh, finalPosBalanceCh,
+                initialUser1BalanceCh, finalUser1BalanceCh,
+                initialUser2BalanceCh, finalUser2BalanceCh in {
+              prepareUser!("2222222222222222222222222222222222222222222222222222222222222222", 10000, *setupCh) |
```
ArturGajowy: having those looooong literals replaced with variables ("constatns") by using a `match` on the top would help

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+              for (_, _, _ <- setupCh) {
+                @PoS!("getBonds", *bondsCh) |
+                for (@initialBonds <- bondsCh) {
+                  @PoS!("bond", 20000, *retCh) |
+
+                  for ( @(result, msg) <- retCh) {
+                    @PoS!("getBonds", *bondsCh) |
+                    rhoSpec!("assertMany",
+                      [
+                        ((result, "==", false), "the bond should fail"),
```
ArturGajowy: would be prudent to test the failure msg too

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+              prepareUser!("6666666666666666666666666666666666666666666666666666666666666666", 10000, *setupCh) |
+              for (_, _, _ <- setupCh) {
+                @PoS!("getBonds", *bondsCh) |
+                for (@initialBonds <- bondsCh) {
+                  @PoS!("bond", -1, *retCh) |
+                  for ( @(result, _) <- retCh) {
+                    @PoS!("getBonds", *bondsCh) |
+                    rhoSpec!("assertMany",
+                      [
+                        ((result, "==", false), "the bond should fail"),
```
ArturGajowy: ditto re failure cause

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+              prepareUser!("7777777777777777777777777777777777777777777777777777777777777777", 10000, *setupCh) |
+              for (_, _, _ <- setupCh) {
+                @PoS!("getBonds", *bondsCh) |
+                for (@initialBonds <- bondsCh) {
+                  @PoS!("bond", 100000, *retCh) |
+                  for ( @(result, _) <- retCh) {
+                    @PoS!("getBonds", *bondsCh) |
+                    rhoSpec!("assertMany",
+                      [
+                        ((result, "==", false), "the bond should fail"),
```
ArturGajowy: ditto re failure cause

casper/src/test/resources/PoSTest.rho
```diff
@@ -3,32 +3,290 @@
...
+                  @PoS!("getBonds", *initialBondsCh) |
+                  for ( @initialBonds <- initialBondsCh) {
+                    @PoS!("bond", 150, *bond2Ch) |
+                    for ( @(result2, _) <- bond2Ch) {
+                      @PoS!("getBonds", *bondsCh) |
+                      rhoSpec!("assertMany",
+                        [
+                          ((result1, "==", true), "the first bond should succeed"),
+                          ((result2, "==", false), "the second bond should fail"),
+                          ((initialBonds, "== <-", *bondsCh), "the bonds map remains unchanged")
```
ArturGajowy: The 'bonding fails' scenario could be extracted into a method and reused 4 times. In this case, we'd do the initial successful bonding, an assertion about its success, and then we'd delegate to `testBondFailure(150, "Already bonded!", ackCh)`

ArturGajowy:
We should remove `pos` from `system instances registry`, maybe even along that registry, in this pr

casper/src/test/resources/PoSTest.rho
odeac: It can be reused 3 times, indeed. The `test_bonding_fails_if_already_bonded` is different

casper/src/test/resources/PoSTest.rho
odeac: Not really. They are only used once. and then I would have user1PublicKey, user2PublicKey... user8PublicKey. Not fun! 

This way they are only used in one single place i.e. the call to prepareUser

casper/src/main/resources/PoS.rhox
odeac: This complicates the API. I wouldn't worry about the PoS contract receiving a rootAuth token when needed because it's a "system contract"

Let's fix the vault drain issue first and then we will see how this changes.

casper/src/main/resources/PoS.rhox
odeac: The map doesn't shrink for now. 

Also I don't know why this sequence number is needed in the first place.

casper/src/main/resources/PoS.rhox
odeac: let's address that in the future

casper/src/main/resources/PoS.rhox
ArturGajowy: > Also I don't know why this sequence number is needed in the first place.

@KentShikama or @JosDenmark, any ideas? Maybe we could remove it?

casper/src/main/resources/PoS.rhox
odeac: are you saying that this call is not really needed and the vault will be created when it's used the first time?

casper/src/test/resources/PoSTest.rho
ArturGajowy: I guess it's a matter of preference until they're repeated

casper/src/main/resources/PoS.rhox
odeac: True. Even better, we should get the current user public key instead of hardcoding it twice.

casper/src/test/resources/PoSTest.rho
odeac: I'm not happy with testing based on error messages but I agree that we should make sure the reason for the failure is the right one

casper/src/test/resources/PoSTest.rho
odeac: same

casper/src/test/resources/PoSTest.rho
odeac: same

odeac:
> We should remove `pos` from `system instances registry`, maybe even along that registry, in this pr

We cannot do that now. I have a separate Jira issue in progress which addresses exactly that

casper/src/main/resources/PoS.rhox
ArturGajowy: Related issue: https://rchain.atlassian.net/browse/RCHAIN-3425

casper/src/main/resources/PoS.rhox
ArturGajowy: I'm saying let's use the interpolated posVaultRevAddress to spawn the vault, instead of inferring the revAddress from the pubkey of the deployer of this system contract.

No big deal though, I think - it should not be a public-key-bound vault anyway.

Here's a ticket for having that API https://rchain.atlassian.net/browse/RCHAIN-3427

And here's one for using that API https://rchain.atlassian.net/browse/RCHAIN-3428

If you could pull in some of those tickets and fix a couple of them in a single PR we'd be able to avoid some churn :)

casper/src/main/resources/PoS.rhox
ArturGajowy: :sob:

casper/src/main/resources/PoS.rhox
ArturGajowy: Please create a new issue for the one-time-stake-transfer rev vaults and put it into sprint-30 :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
```diff
@@ -42,7 +42,7 @@ object RhoType {
   }
 
   object Tuple2 {
-    def apply(b: Boolean) = Expr(GBool(b))
+    def apply(tuple: (Par, Par)): Par = Expr(ETupleBody(ETuple(Seq(tuple._1, tuple._2))))
```
JosDenmark: Same here. Should be done in a fixup.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -71,7 +69,7 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
     result.status.isFailed should be(true)
   }
 
-  "computeDeployPayment" should "transfer REV from the deployer's vault to the PoS vault" ignore withTestFixture {
+  "computeDeployPayment" should "transfer REV from the deployer's vault to the PoS vault" in withTestFixture {
```
JosDenmark: Same here. Should be done in a fixup.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -222,8 +222,14 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     }
 
   private def deployPaymentSource(amount: Long, name: String = "__SCALA__"): String =
-    s"""@"$name"!((true, Nil))"""
-  //TODO replace the dummy implementation with an actual call to PoS.pay where PoS is taken from SystemInstances
+    s"""
```
JosDenmark: This change, for example, should be done in a fixup.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
odeac: there are conflicts and the fix-up it only works if I put everything in one big commit but if you insist...

odeac:
> You're getting close.
> 
> A lot is going on in the last commit. The final commit should really be broken into fixups to previous commits, so that we just see the necessary changes. The commit history should be such that we don't see corrections made in later commits to mistakes made in earlier commits. You can use `git reset HEAD~` to restore to previous ref while keeping changes. From there, just do partial commits, and do `git rebase -i --autosquash` for each fixup [https://github.com/embox/embox/wiki/Contributing:-Git-cheatsheet](url).
> 
> Once you've done the fixups, you should add the changes that bring `PoS.rho` under `ProofOfStake.scala` in a separate commit. After that, I'll approve the PR 😄.

Because of the conflicts I can only do it if I squash everything in one big commit. It's a waste of time to break this into small commits at this point

casper/src/main/resources/PoS.rhox
odeac: let's do it simple, like it is implemented now and we'll figure out the final solution as we go. Those two tickets are not defined yet and there is no need to postpone the merge of this PR

casper/src/main/resources/PoS.rhox
odeac: I'm not sure we actually need that one-time-stake-transfer. If we consider acceptable to have the rootAuth passed to the RevVault contract then we can also have it passed to other system contracts like the PoS contract.

ArturGajowy: The one time stake transfer is needed, because two bonding requests could be processed in parallel, leading to the balance not matching what we expect.

`If we consider acceptable to have the rootAuth passed to the RevVault contract ` - this is only for creating a RevVault-specific auth ocap. After having on-chain-deployer-id we could just as well remove the `deployerAuthKey` from RevVault and make the user create that token themselves based on a `bundle0{_revVault}` exposed by the RevVault contract. So they'd call
`@AuthKey!("make", (*_revVaultOcapBundled, bundle0{*deployerId}), *ret)` to create a key themselves. I'm not sure why you bring this up in this thread though.

ArturGajowy: Ah, I see why you bring that up. In no case should passing a `deployerId` to a contract should be utilized to do things in other contracts. If the user wants to give us that capability, they should provide the related tokens.

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
```diff
@@ -20,3 +20,28 @@ object CompiledRholangSource {
...
+
+  val finalContent = env.foldLeft(originalContent) {
+    case (content, (name, value)) => content.replace(s"""$$$$$name$$$$""", value.toString)
+  }
+
+  override val path: String = classpath
+  override val code: String =
+    s"""//Loaded from resource file <<$classpath>>
+         |$finalContent
+         |""".stripMargin
```
ArturGajowy: You forgot the '#' instead of '|'

casper/src/main/resources/PoS.rhox
odeac: How are you going to prevent that? The deployerId is, in fact, a `su - root`. With this in mind we should be aware of the risks involved and we could take advantage of it, one of the situations being this one.

Anyway, we should discuss this in detail: https://rchain.atlassian.net/browse/RCHAIN-3437

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
odeac: fixed

PR #2359, opened 2019-04-14
A bit faster casper tests and CI

ArturGajowy:
## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -71,7 +71,16 @@ pipeline:
     group: parallel
     <<: *sbttestenv
     commands:
-      - sbt test it:test
+      - cp -r . ../step.run-unit-tests && cd ../step.run-unit-tests
+      - sbt "set test in casper := {}" test it:test
+
+  pr-run-casper-tests:
+    <<: *pr_build_conditions
+    group: parallel
+    <<: *sbttestenv
+    commands:
+      - cp -r . ../step.run-casper-tests && cd ../step.run-casper-tests
+      - sbt casper/test
 
   # full build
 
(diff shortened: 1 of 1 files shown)


ArturGajowy:
bors try

ArturGajowy:
bors try

casper/src/test/scala/coop/rchain/casper/MultiParentCasperTestUtil.scala
```diff
@@ -100,11 +100,12 @@ object MultiParentCasperTestUtil {
...
     (for {
+      //TODO consider reverting back to the production-used variant
       activeRuntime <- Runtime
                         .createWithEmptyCost[Task, Task.Par](
-                          storageDirectory,
-                          storageSize,
-                          StoreType.LMDB
+                          Paths.get("/not/a/path"),
+                          -1,
+                          StoreType.InMem
```
adaszko: Is InMem this something like a Java HashTable?  If so, I think it is risky to use that.  I'd rather try to disable `fsync` in LMDB.

adaszko: See `metasync` and `sync` options at https://lmdb.readthedocs.io/en/release/#environment-class

ArturGajowy: WDYM by risky? Do you expect more errors? Or more false-positives?

If the latter, I'd say having 200 tests not targeted at the false-positives exercise the tupplespace is an overkill and a waste.

Re concurrency semantics - @dzajkowski WDYT?

adaszko: > WDYM by risky? Do you expect more errors? Or more false-positives?

I mean any difference in behavior compared to actual LMDB.  Take for instance the long standing bug of segfaulting at shutdown.  Will it happen with InMem?  I think it should and I bet it won't.

ArturGajowy:
Let's try this without the InMem change and see how it rolls :)

ArturGajowy:
bors try

ArturGajowy:
> Let's try this without the InMem change and see how it rolls :)

Yup, no visible difference. Let's use the production setting then.

PR #2360, opened 2019-04-14
Filter duplicate deploys in current chain based on (deployer, timestamp) 

KentShikama:
## Overview

See commit messages

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3244

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -113,17 +113,27 @@ object BlockCreator {
       validDeploys = distinctDeploys.filter(
         d => notFutureDeploy(currentBlockNumber, d) && notExpiredDeploy(earliestBlockNumber, d)
       )
-      deploysInCurrentChain <- DagOperations
-                                .bfTraverseF[F, BlockMessage](parents.toList)(
-                                  b =>
-                                    ProtoUtil
-                                      .unsafeGetParentsAboveBlockNumber[F](b, earliestBlockNumber)
-                                )
-                                .map { b =>
-                                  ProtoUtil.deploys(b).flatMap(_.deploy)
-                                }
-                                .toList
-    } yield (validDeploys -- deploysInCurrentChain.flatten).toSeq
+      deploysInCurrentChainList <- DagOperations
+                                    .bfTraverseF[F, BlockMessage](parents.toList)(
+                                      b =>
+                                        ProtoUtil
+                                          .unsafeGetParentsAboveBlockNumber[F](
+                                            b,
+                                            earliestBlockNumber
+                                          )
+                                    )
+                                    .map { b =>
+                                      ProtoUtil.deploys(b).flatMap(_.deploy)
+                                    }
+                                    .toList
+      deploysInCurrentChain = deploysInCurrentChainList.flatten.toSet
+      deploysInCurrentChainAsTuple = deploysInCurrentChain.map(
+        deploy => (deploy.deployer, deploy.timestamp)
+      )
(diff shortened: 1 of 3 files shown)


KentShikama:
bors try

ArturGajowy:
Shouldn't we filter based on deploy signature though?

KentShikama:
Deploy signature includes the contents of the deploy. If two deploys have the same (deployer, timestamp) but have different contents then they shouldn't be both in the same block (or chain).

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -107,20 +107,33 @@ object BlockCreator {
       currentBlockNumber  = maxBlockNumber + 1
       earliestBlockNumber = currentBlockNumber - expirationThreshold
       deploys             = state.deployHistory
-      validDeploys = deploys.filter(
+      distinctDeploys = deploys
+        .groupBy(deploy => (deploy.deployer, deploy.timestamp))
+        .map(_._2.head)
```
odeac: this would be a candidate for a distinctBy function

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -107,20 +107,33 @@ object BlockCreator {
...
+                                          .unsafeGetParentsAboveBlockNumber[F](
+                                            b,
+                                            earliestBlockNumber
+                                          )
+                                    )
+                                    .map { b =>
+                                      ProtoUtil.deploys(b).flatMap(_.deploy)
+                                    }
+                                    .toList
+      deploysInCurrentChain = deploysInCurrentChainList.flatten.toSet
```
odeac: if you remove the toList and the flatten you get the same result

casper/src/test/scala/coop/rchain/casper/MultiParentCasperDeploySpec.scala
```diff
@@ -114,6 +114,22 @@ class MultiParentCasperDeploySpec extends FlatSpec with Matchers with Inspectors
...
+    val casper           = node.casperEff
+    implicit val timeEff = new ConstantTime[Effect]
+
+    for {
+      deploy1           <- ConstructDeploy.basicDeployData[Effect](0)
+      deploy2           <- ConstructDeploy.basicDeployData[Effect](1)
+      _                 <- casper.deploy(deploy1)
+      _                 <- casper.deploy(deploy2)
+      createBlockResult <- casper.createBlock
+      Created(block)    = createBlockResult
```
tymm: Is the `Created(...)` part needed?

tymm:
> Deploy signature includes the contents of the deploy. If two deploys have the same (deployer, timestamp) but have different contents then they shouldn't be both in the same block (or chain).

But is there a good reason to not filter based on the deploy signature?
I think there might be situations where allowing for two or more deployments from one deployer at the same timestamp could make sense.

KentShikama:
@tymm Yeah on further thought, I agree there is no good reason to prevent the same (deployer, timestamp). If we check signatures, we still won't have to look back past what is expired due to the valid after block number.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
KentShikama: Isn't the flatten necessary? Does toSet also flatten?

ArturGajowy:
@KentShikama is this still intended to be merged, given your last comment?

KentShikama:
@ArturGajowy Superseded by https://github.com/rchain/rchain/pull/2377

PR #2361, opened 2019-04-15
Add network version ID

sebtomba:
## Overview
Add a network ID to all messages in node discovery and RChain protocol. Reject peers from other networks. The default network ID is ATM `testnet`. A network ID can be provided in the configuration file (`rchain.server.network-id`) or as a command line option `--network <id>`.
 
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3230

### Notes
This change is not backward compatible. The new rnode will reject to talk to all previous versions of the rnode.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
+++ b/comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
@@ -50,11 +50,10 @@ class GrpcKademliaRPC(networkId: String, timeout: FiniteDuration)(
                       _.sendLookup(lookup)
                         .timer("lookup-time")
                     ).attempt
-    } yield
-      responseErr.fold(
-        kp(Seq.empty[PeerNode]),
-        r => if (r.networkId == networkId) r.nodes.map(toPeerNode) else Seq.empty[PeerNode]
-      )
+    } yield responseErr.fold(
+      kp(Seq.empty[PeerNode]),
+      r => if (r.networkId == networkId) r.nodes.map(toPeerNode) else Seq.empty[PeerNode]
+    )
 
   private def withClient[A](peer: PeerNode, timeout: FiniteDuration, enforce: Boolean = false)(
       f: KademliaRPCServiceStub => Task[A]
--- a/comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
+++ b/comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
@@ -17,7 +17,7 @@ import io.grpc.netty._
 import monix.eval._
 import monix.execution._
 
-class GrpcKademliaRPC(timeout: FiniteDuration)(
+class GrpcKademliaRPC(networkId: String, timeout: FiniteDuration)(
     implicit
     scheduler: Scheduler,
     peerNodeAsk: PeerNodeAsk[Task],
@@ -31,23 +31,30 @@ class GrpcKademliaRPC(timeout: FiniteDuration)(
     for {
       _     <- Metrics[Task].incrementCounter("ping")
       local <- peerNodeAsk.ask
-      ping  = Ping().withSender(toNode(local))
+      ping  = Ping().withSender(toNode(local)).withNetworkId(networkId)
       pongErr <- withClient(peer, timeout)(
                   _.sendPing(ping)
                     .timer("ping-time")
                 ).attempt
-    } yield pongErr.fold(kp(false), kp(true))
+    } yield pongErr.fold(kp(false), _.networkId == networkId)
 
   def lookup(key: Seq[Byte], peer: PeerNode): Task[Seq[PeerNode]] =
     for {
-      _      <- Metrics[Task].incrementCounter("protocol-lookup-send")
-      local  <- peerNodeAsk.ask
-      lookup = Lookup().withId(ByteString.copyFrom(key.toArray)).withSender(toNode(local))
+      _     <- Metrics[Task].incrementCounter("protocol-lookup-send")
+      local <- peerNodeAsk.ask
+      lookup = Lookup()
+        .withId(ByteString.copyFrom(key.toArray))
+        .withSender(toNode(local))
+        .withNetworkId(networkId)
       responseErr <- withClient(peer, timeout)(
                       _.sendLookup(lookup)
                         .timer("lookup-time")
                     ).attempt
-    } yield responseErr.fold(kp(Seq.empty[PeerNode]), _.nodes.map(toPeerNode))
+    } yield
+      responseErr.fold(
+        kp(Seq.empty[PeerNode]),
+        r => if (r.networkId == networkId) r.nodes.map(toPeerNode) else Seq.empty[PeerNode]
+      )
 
   private def withClient[A](peer: PeerNode, timeout: FiniteDuration, enforce: Boolean = false)(
       f: KademliaRPCServiceStub => Task[A]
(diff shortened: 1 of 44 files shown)


comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
```diff
@@ -31,23 +31,30 @@ class GrpcKademliaRPC(timeout: FiniteDuration)(
...
       _     <- Metrics[Task].incrementCounter("ping")
       local <- peerNodeAsk.ask
-      ping  = Ping().withSender(toNode(local))
+      ping  = Ping().withSender(toNode(local)).withNetworkId(networkId)
       pongErr <- withClient(peer, timeout)(
                   _.sendPing(ping)
                     .timer("ping-time")
                 ).attempt
-    } yield pongErr.fold(kp(false), kp(true))
+    } yield pongErr.fold(kp(false), _.networkId == networkId)
```
KentShikama: So here the "default case" in the fold, the pong fails. And the pong only has succeeded if the pong's network ID is equal to the current's?

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -69,33 +67,43 @@ object StreamHandler {
...
+          val readSoFar = stmd.readSoFar + array.length
+          if (circuitBreaker(readSoFar))
+            Right(stmd.copy(circuitBroken = true))
+          else
+            Left(stmd.copy(readSoFar = readSoFar))
+      }
 
     EitherT(collectStream.attempt.map {
+      case Right(stmd) if stmd.wrongNetwork =>
+        new RuntimeException("Wrong network id").asLeft
```
KentShikama: Are you sure we want to throw an RuntimeException here? It would be a common thing once we start running multiple networks

comm/src/main/scala/coop/rchain/comm/discovery/GrpcKademliaRPC.scala
sebtomba: @KentShikama yes, this is true. The pinged node responses with a `Pong` which contains a network ID. We consider a node as _dead_ if the `Pong` has the wrong network ID. However, please keep in mind that Kademlia pings remote peers only when a bucket full is.

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
sebtomba: Actually, this exception won't get thrown. The private function `collect` is used only by `StreamHandler.handleStream` which returns a `Task[Either[Throwable, StreamMessage]]`. This is used then as
```scala
(StreamHandler.handleStream(networkId, tempFolder, observable, neverBreak) >>= {
  case Left(ex)   => logger.error(s"Could not receive stream! Details: ${ex.getMessage}", ex)
  case Right(msg) => Task.delay(bufferBlobMessage.pushNext(msg)).as(())
}).as(ChunkResponse())
```

On the other hand, once all nodes are updated to this version we should never see streams coming from other networks because they will get filtered at the node discovery layer. However, I implemented also the network ID check at the RChain protocol layer to prevent traffic from older nodes.

sebtomba:
bors try

ArturGajowy:
bors try

comm/src/main/scala/coop/rchain/comm/transport/SslSessionServerInterceptor.scala
```diff
@@ -1,17 +1,18 @@
...
 import io.grpc._
 import javax.net.ssl.SSLSession
 
 /**
-  * This wart exists because that's how grpc works. They looooovveee throwing exceptions
+  * This wart exists because that's how gRPC works
   */
-@SuppressWarnings(Array("org.wartremover.warts.Throw"))
-class SslSessionServerInterceptor() extends ServerInterceptor {
+@SuppressWarnings(Array("org.wartremover.warts.Var"))
```
rabbitonweb: I love that you've removed `Throw`, why did you added `Var`?

comm/src/main/scala/coop/rchain/comm/transport/SslSessionServerInterceptor.scala
```diff
@@ -27,42 +28,55 @@ class SslSessionServerInterceptor() extends ServerInterceptor {
       call: ServerCall[ReqT, RespT]
   ) extends ServerCall.Listener[ReqT] {
 
-    override def onHalfClose(): Unit = next.onHalfClose()
-    override def onCancel(): Unit    = next.onCancel()
-    override def onComplete(): Unit  = next.onComplete()
-    override def onReady(): Unit     = next.onReady()
+    @volatile
+    private var closeWithStatus = Option.empty[Status]
```
rabbitonweb: What's the idea behind this?

comm/src/main/scala/coop/rchain/comm/transport/TcpServerObservable.scala
```diff
@@ -44,18 +48,21 @@ class TcpServerObservable(
...
         request.protocol
           .fold(internalServerError("protocol not available in request").pure[Task]) { protocol =>
-            Task.delay(bufferTell.pushNext(Send(protocol))).map {
-              case false => internalServerError("message dropped")
-              case true  => noResponse
-            }
+            rPConfAsk.reader(_.local) >>= (
+                src =>
+                  Task.delay(bufferTell.pushNext(Send(protocol))).map {
+                    case false => internalServerError("message dropped")
```
rabbitonweb: you only need `src` when `case true`, this should be rather sth like

```
Task.delay(bufferTell.pushNext(Send(protocol))) >>= {
     case false => internalServerError("message dropped").pure[Task]
     case true => rPConfAsk.reader(_.local).map(noResponse)
```

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -177,6 +177,9 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
     val secureRandomNonBlocking =
       opt[Flag](descr = "Use a non blocking secure random instance")
 
+    val network =
+      opt[String](descr = "ID of the RChain network")
```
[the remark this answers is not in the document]
sebtomba: The defaults are now in `reference.conf`

comm/src/main/scala/coop/rchain/comm/transport/TcpServerObservable.scala
sebtomba: @rabbitonweb `NoResponse` has a `header` but we never sent one. Since we removed roundtrip all response messages had no headers and were ignored them in the `SslSessionClientInterceptor`. On the other hand, we need to cope with a (potential) dynamic IP address of the local node. That's why we need to ask `RPConf` every time we want to use the `local` node.

comm/src/main/scala/coop/rchain/comm/transport/SslSessionServerInterceptor.scala
sebtomba: @rabbitonweb A `ServerCall` is kind of a state machine. It goes thru the states (omitting `Cancel`):
- `Ready`: connection established, ready to receive messages
- `Message`: receive one or more messages
- `HalfClose`: client signals that all messages have been sent and he's waiting for a response
- `Complete`: the response has been sent to the client

We inspect all incoming messages in the interceptor hook `OnMessage`. However, we can't send a response (in our case a failure status) to the client right away. We need to wait for the `OnHalfClose` hook to send the response. That's why we need a `var`. TBH, I think the `@volatile` annotation is not required but it does no harm. Please keep in mind that an `InterceptionListener` instance is created for every single `ServerCall` so the `var` is not shared between calls.

PR #2365, opened 2019-04-16
New rspace/plug in hot store

lukasz-golebiewski:
## Overview
Adapts RSpace for use with the newly created HotStore abstraction

### JIRA ticket:
Main part of https://rchain.atlassian.net/browse/RCHAIN-3142

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2366, opened 2019-04-16
Do not allow installing/inserting duplicate joins

lukasz-golebiewski:
## Overview
Fixes a bug which surfaced when porting RSpace tests to work with the new HotStore (within https://github.com/rchain/rchain/pull/2365)
Duplicate joins per channel should not be allowed

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3141

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -167,17 +167,19 @@ private class InMemHotStore[F[_]: Sync, C, P, A, K](
   def putJoin(channel: C, join: Seq[C]): F[Unit] =
     for {
       joins <- getJoins(channel)
-      _ <- S.flatModify { cache =>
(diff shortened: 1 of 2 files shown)


rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -167,17 +167,19 @@ private class InMemHotStore[F[_]: Sync, C, P, A, K](
   def putJoin(channel: C, join: Seq[C]): F[Unit] =
     for {
       joins <- getJoins(channel)
-      _ <- S.flatModify { cache =>
+      _ <- if (!joins.contains(join)) S.flatModify { cache =>
```
dzajkowski: thought: would the order of `join: Seq[C]` affect this check?

lukasz-golebiewski: You're probably right, but this would mean that the current `IStore` implementations are also affected

PR #2367, opened 2019-04-16
Reproduce IgnorableEquivocation scenario in unit tests

itegulov:
## Overview
Reproduces "Invalid block: IgnorableEquivocation" bug scenario fixed in https://github.com/rchain/rchain/pull/2353

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3235

### Notes

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
@@ -62,18 +62,20 @@ class CreateBlockAPITest extends FlatSpec with Matchers {
 
     def testProgram(blockApiLock: Semaphore[Effect])(
         implicit casperRef: MultiParentCasperRef[Effect]
-    ): Effect[(ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse])] =
+    ): Effect[(ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse])] =
       EitherT.liftF(
         for {
           t1 <- createBlock(deploys.head, blockApiLock).value.start
           _  <- Time[Task].sleep(2.second)
           t2 <- createBlock(deploys.last, blockApiLock).value.start //should fail because other not done
+          t3 <- createBlock(deploys.last, blockApiLock).value.start //should fail because other not done
           r1 <- t1.join
           r2 <- t2.join
-        } yield (r1.right.get, r2.right.get)
+          r3 <- t3.join
+        } yield (r1.right.get, r2.right.get, r3.right.get)
       )
 
-    val (response1, response2) = (for {
+    val (response1, response2, response3) = (for {
       casperRef    <- MultiParentCasperRef.of[Effect]
       _            <- casperRef.set(casper)
       blockApiLock <- Semaphore[Effect](1)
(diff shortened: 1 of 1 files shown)


casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -62,18 +62,21 @@ class CreateBlockAPITest extends FlatSpec with Matchers {
 
     def testProgram(blockApiLock: Semaphore[Effect])(
         implicit casperRef: MultiParentCasperRef[Effect]
-    ): Effect[(ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse])] =
+    ): Effect[(ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse], ApiErr[DeployServiceResponse])] =
       EitherT.liftF(
         for {
           t1 <- createBlock(deploys.head, blockApiLock).value.start
-          _  <- Time[Task].sleep(2.second)
+          _  <- Time[Task].sleep(1.second)
```
odeac: why did you choose 1 second?

itegulov: It was 2 seconds before (and, admittedly, I don't know why) and I just split it between two calls

odeac: why not remove the sleep entirely?

itegulov: I guess because we want to simulate a close-to-real-world scenario where the user tries to create a block 3 times in a row with some delays between the calls. I don't have a strong opinion here and it is not really necessary for the purpose of testing the IgnorableEquivocation bug so I am fine with removing it if you insist.

odeac: I think the 1s sleep is arbitrary. If the reason why you reduced the time between two events was to simulate a collision then you will certainly get the collision without the sleep

itegulov: The first sleep is necessary to "ensure" that the first createBlock attempt will succeed. The second sleep is unnecessary, you are right. I will bring the original 2s sleep back and remove the second sleep so the code will look like this:
```scala
  t1 <- createBlock(...).value.start
  _  <- Time[Task].sleep(2.second)
  t2 <- createBlock(...).value.start
  t3 <- createBlock(...).value.start
```
Does this sound good?

rabbitonweb: I think I would remove it if it's not failing the test

odeac: If the 2s sleep is required to ensure that the createBlock will succeed then maybe we shouldn't start the task asynchronously. I'm thinking to wait for the create block to finish and then start t2 and t3.

This way we can remove the sleep entirely

itegulov: No, the whole point of the test is to test how BlockAPI handles concurrent calls. When I said "ensure that the first createBlock attempt will succeed" I meant "ensure that the first createBlock attempt will acquire all required locks/resources to be at the point when it will definitely succeed in the near future". So after waiting 2 seconds we are trying to acquire the necessary locks and fail since the first attempt already acquired them and did not release them yet. If we don't wait 2 seconds then the second attempt could overtake the first attempt and acquire the locks instead.

PR #2371, opened 2019-04-18
Bump version to 0.9.3

woky:


build.sbt
```diff
@@ -252,7 +252,7 @@ lazy val node = (project in file("node"))
   .settings(commonSettings: _*)
   .enablePlugins(RpmPlugin, DebianPlugin, JavaAppPackaging, BuildInfoPlugin)
   .settings(
-    version := "0.9.2" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
```
lukasz-golebiewski: Maybe it would make sense to store the "0.9.3" version in a constant so that it's less probable that some unwanted change in the `git.gitHeadCommit...` gets in unnoticed?

woky: Sure, though I don't know what would be the good place to put it.

PR #2375, opened 2019-04-19
raise error in RuntimeManager.captureResults on unsuccessful deploy

odeac:
## Overview
Without this change the errors are silent and can go unnoticed

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3261

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -52,7 +52,7 @@ trait RuntimeManager[F[_]] {
   def emptyStateHash: ByteString
 }
 
-class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
+class RuntimeManagerImpl[F[_]: Concurrent : Sync] private[rholang] (
     val emptyStateHash: ByteString,
     runtimeContainer: MVar[F, Runtime[F]]
 ) extends RuntimeManager[F] {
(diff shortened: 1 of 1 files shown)


lukasz-golebiewski:
It's sad that no test had to be adapted. Do you think you could write a fresh one?

ArturGajowy:
Don't you have to handle this error somewhere in casper as well? Did you run casper tests?

It should be the case that such an error during a propose doesn't terminate the propose. Rather, the failed deploys should be propagated for other nodes to know (or ignored, if we don't do the propagation yet).

ArturGajowy:
Also, I've observed that RhoSpec doesn't report rholang evaluation errors neither in test nor in production code. Is this change intended to fix that?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -71,12 +71,13 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
-          result <- if (processedDeploy.status.isFailed) Seq.empty[Datum[ListParWithRandom]].pure[F]
-                   else {
-                     val r: F[Seq[Datum[ListParWithRandom]]] =
-                       runtime.space.getData(name).map(_.toSeq)
-                     r
-                   }
+          result <- if (processedDeploy.status.isFailed)
+                      Sync[F].raiseError(new Exception("Evaluation error:" + processedDeploy.status))
+                    else {
+                      val r: F[Seq[Datum[ListParWithRandom]]] =
```
tymm: Not related to your change but looks like this is a leftover from debugging and can be removed.

odeac:
the rebase is too dificult.If we decide to do this again I'll redoit

PR #2380, opened 2019-04-23
Slashing invalid blocks

KentShikama:
## Overview

Add "slash" method in PoS and `rho:casper:invalidBlocks`

In the block creator, we queue up a slashing deploy
that calls the "slash" method in the new PoS contract.
Note this doesn't actually do anything yet as our source
of truth for bonds right now is still the old PoS contract

`ProtoUtil.unseenBlockHashes` (used in rho:casper:invalidBlocks)

When we pass the list of slashable blocks to Rholang,
we need to exclude all the blocks that the slasher
hasn't justified as seen yet.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-2381

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -83,6 +83,23 @@ new  PoS,
                   deposit!(userPk, amount, *returnCh)
                 }
               }
+            } |
+
+            contract PoS(@"slash", @blockHash, return) = {
+              new userCh, invalidBlocksCh, getInvalidBlocks(`rho:casper:invalidBlocks`) in {
+                getInvalidBlocks!(*invalidBlocksCh) |
+                getUser!(*userCh) |
+                for (@invalidBlocks <- invalidBlocksCh; @bonds <- bondsCh; @userPk <- userCh) {
+                  new toBeSlashed in {
+                    toBeSlashed!(invalidBlocks.getOrElse(blockHash, userPk)) |
+                    for (@validator <- toBeSlashed) {
+                      // TODO: Transfer to coop wallet instead of just simply setting bonds to 0
+                      bondsCh!(bonds.set(validator, [0, bonds.get(validator).nth(1)])) |
+                      return!(true)
+                    }
+                  }
+                }
+              }
             }
           }
         } |
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -94,7 +94,7 @@ new  PoS,
                     toBeSlashed!(invalidBlocks.getOrElse(blockHash, userPk)) |
                     for (@validator <- toBeSlashed) {
                       // TODO: Transfer to coop wallet instead of just simply setting bonds to 0
-                      bondsCh!(bonds.set(validator, [0, bonds.get(validator).nth(1)])) |
+                      bondsCh!(bonds.set(validator, (0, bonds.get(validator).nth(1)))) |
                       return!(true)
                     }
                   }
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -436,13 +436,22 @@ object ProtoUtil {
       result <- latestMessages.toList
                  .traverse {
                    case (validator, latestMessage) =>
-                     val lastSeenBlockHash = latestMessagesOfBlock(validator)
-                     DagOperations
-                       .bfTraverseF(List(latestMessage))(
-                         block => getCreatorJustificationUnlessGoal(dag, block, lastSeenBlockHash)
-                       )
-                       .map(_.blockHash)
-                       .toSet
+                     latestMessagesOfBlock.get(validator) match {
+                       case Some(latestMessageOfBlockByValidator) =>
+                         DagOperations
+                           .bfTraverseF(List(latestMessage))(
+                             block =>
+                               getCreatorJustificationUnlessGoal(
+                                 dag,
+                                 block,
+                                 latestMessageOfBlockByValidator
+                               )
+                           )
+                           .map(_.blockHash)
+                           .toSet
+                       case None =>
+                         Set.empty[BlockHash].pure[F]
+                     }
                  }
                  .map(_.flatten.toSet)
     } yield result -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -434,28 +434,41 @@ object ProtoUtil {
     for {
       latestMessages        <- dag.latestMessages
       latestMessagesOfBlock <- ProtoUtil.toLatestMessage(block.justifications, dag)
-      result <- latestMessages.toList
-                 .traverse {
-                   case (validator, latestMessage) =>
-                     latestMessagesOfBlock.get(validator) match {
-                       case Some(latestMessageOfBlockByValidator) =>
-                         DagOperations
-                           .bfTraverseF(List(latestMessage))(
-                             block =>
-                               getCreatorJustificationUnlessGoal(
-                                 dag,
-                                 block,
-                                 latestMessageOfBlockByValidator
-                               )
-                           )
-                           .map(_.blockHash)
-                           .toSet
-                       case None =>
-                         Set.empty[BlockHash].pure[F]
-                     }
-                 }
-                 .map(_.flatten.toSet)
-    } yield result -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
+      unseenBlockHashesAndLatestMessages <- latestMessages.toStream
+                                             .traverse {
+                                               case (validator, latestMessage) =>
+                                                 getJustificationChainFromLatestMessageToBlock(
+                                                   dag,
+                                                   latestMessagesOfBlock,
+                                                   validator,
+                                                   latestMessage
+                                                 )
+                                             }
+                                             .map(_.flatten.toSet)
+    } yield unseenBlockHashesAndLatestMessages -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
+
+  private def getJustificationChainFromLatestMessageToBlock[F[_]: Sync: BlockStore](
+      dag: BlockDagRepresentation[F],
+      latestMessagesOfBlock: Map[Validator, BlockMetadata],
+      validator: Validator,
+      latestMessage: BlockMetadata
+  ): F[Set[BlockHash]] =
+    latestMessagesOfBlock.get(validator) match {
+      case Some(latestMessageOfBlockByValidator) =>
+        DagOperations
+          .bfTraverseF(List(latestMessage))(
+            block =>
+              getCreatorJustificationUnlessGoal(
+                dag,
+                block,
+                latestMessageOfBlockByValidator
+              )
+          )
+          .map(_.blockHash)
+          .toSet
+      case None =>
+        Set.empty[BlockHash].pure[F]
+    }
 
   private def getCreatorJustificationUnlessGoal[F[_]: Sync: BlockStore](
       dag: BlockDagRepresentation[F],
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -98,11 +98,7 @@ object ProtoUtil {
       }
 
   def creatorJustification(block: BlockMetadata): Option[Justification] =
-    block.justifications
-      .find {
-        case Justification(validator: Validator, _) =>
-          validator == block.sender
-      }
+    block.justifications.find(justification => justification.validator == block.sender)
 
   /**
     * Since the creator justification is unique
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -21,6 +21,7 @@ import coop.rchain.models._
 import coop.rchain.rholang.interpreter.DeployParameters
 
 import scala.collection.immutable
+import scala.collection.immutable.Map
 
 object ProtoUtil {
 
@@ -477,5 +478,25 @@ object ProtoUtil {
         }
       case None =>
         List.empty[BlockMetadata].pure[F]
+    }
+
+  def invalidLatestMessages[F[_]: Monad](
+      dag: BlockDagRepresentation[F]
+  ): F[Map[Validator, BlockHash]] =
+    dag.latestMessages.flatMap(
+      latestMessages =>
+        invalidLatestMessages(dag, latestMessages.map {
+          case (validator, block) => (validator, block.blockHash)
+        })
+    )
+
+  def invalidLatestMessages[F[_]: Monad](
+      dag: BlockDagRepresentation[F],
+      latestMessagesHashes: Map[Validator, BlockHash]
+  ): F[Map[Validator, BlockHash]] =
+    dag.invalidBlocks.map { invalidBlocks =>
+      latestMessagesHashes.filter {
+        case (_, blockHash) => invalidBlocks.map(_.blockHash).contains(blockHash)
+      }
     }
 }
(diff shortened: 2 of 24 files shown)


casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -411,4 +418,48 @@ object ProtoUtil {
...
+              List(creatorJustification).pure[F]
+            }
+          case None =>
+            Sync[F].raiseError[List[BlockMetadata]](
+              new RuntimeException(s"Missing block hash $hash in block dag.")
+            )
+        }
+      case None =>
+        List.empty[BlockMetadata].pure[F]
+    }
```
odeac: This function with two layers of matches is unreadable. The version below reads better and should work the same

```scala
  private def getCreatorJustificationUnlessGoal[F[_]: Sync: BlockStore](
      block: BlockMetadata,
      goal: BlockMetadata
  ): F[List[BlockMetadata]] = {
    def findByHash(hash : ByteString) =
      for {
        maybeBlock  <- dag.lookup(hash)
        maybeFoundBlock = maybeBlock.filter(_ == goal)
        result <- maybeFoundBlock
          .fold(
            Sync[F].raiseError[Option[BlockMetadata]](new RuntimeException(s"Missing block hash $hash in block dag.")))(
            j => Option(j).pure[F])
      } yield result

    ProtoUtil.creatorJustification(block)
      .map(_.latestBlockHash)
      .traverse(findByHash)
      .map(_.flatten.toList)
  }
```

casper/src/test/scala/coop/rchain/casper/util/ProtoUtilTest.scala
```diff
@@ -24,4 +27,59 @@ class ProtoUtilTest extends FlatSpec with Matchers with GeneratorDrivenPropertyC
...
     }
   }
+
+  import MultiParentCasperTestUtil._
+
+  implicit val timeEff = new LogicalTime[Effect]
+
+  private val (validatorKeys, validatorPks) = (1 to 4).map(_ => Ed25519.newKeyPair).unzip
+  private val genesis = buildGenesis(
+    buildGenesisParameters(4, createBonds(validatorPks))
```
odeac: I would use validatorKeys.length instead of hardcoding 4 here

casper/src/test/scala/coop/rchain/casper/util/ProtoUtilTest.scala
```diff
@@ -24,4 +27,59 @@ class ProtoUtilTest extends FlatSpec with Matchers with GeneratorDrivenPropertyC
...
+    implicit val timeEff = new LogicalTime[Effect]
+
+    for {
+      deploy            <- ConstructDeploy.basicDeployData[Effect](0)
+      casper            = node.casperEff
+      _                 <- casper.deploy(deploy)
+      createBlockResult <- casper.createBlock
+      Created(block0)   = createBlockResult
+      _                 <- casper.addBlock(block0, ignoreDoppelgangerCheck[Effect])
+      _ <- (1 to 8).toList.traverse_[Effect, Unit] { i =>
```
odeac: why 8? 1 would be enough to show that the unseen blocks are returned properly. If you want to test more items would it make sense to have a generative test?

casper/src/test/scala/coop/rchain/casper/util/ProtoUtilTest.scala
```diff
@@ -24,4 +27,59 @@ class ProtoUtilTest extends FlatSpec with Matchers with GeneratorDrivenPropertyC
...
+            for {
+              deploy            <- ConstructDeploy.basicDeployData[Effect](i)
+              createBlockResult <- casper.deploy(deploy) *> casper.createBlock
+              Created(block)    = createBlockResult
+              _                 <- casper.addBlock(block, ignoreDoppelgangerCheck[Effect])
+            } yield ()
+          }
+      dag               <- casper.blockDag
+      unseenBlockHashes <- ProtoUtil.unseenBlockHashes[Effect](dag, block0)
+      _                 = unseenBlockHashes.size should be(8)
```
odeac: you should check that the collection also contains the expected hashes

casper/src/test/scala/coop/rchain/casper/util/ProtoUtilTest.scala
```diff
@@ -24,4 +27,59 @@ class ProtoUtilTest extends FlatSpec with Matchers with GeneratorDrivenPropertyC
...
+      deploy            <- ConstructDeploy.basicDeployData[Effect](0)
+      casper            = node.casperEff
+      _                 <- casper.deploy(deploy)
+      createBlockResult <- casper.createBlock
+      Created(block0)   = createBlockResult
+      _                 <- casper.addBlock(block0, ignoreDoppelgangerCheck[Effect])
+      _ <- (1 to 8).toList.traverse_[Effect, Unit] { i =>
+            for {
+              deploy            <- ConstructDeploy.basicDeployData[Effect](i)
+              createBlockResult <- casper.deploy(deploy) *> casper.createBlock
```
odeac: I don't have a preference here but I've seen some discussion saying that we prefer `>>` instead of `*>`. Not a big deal though.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
KentShikama: Personally, I find your version much harder to read. In fact it isn't the same - when the creatorJustification is None, it should return an empty list.

KentShikama:
bors try

KentShikama:
@odeac  PTAL

KentShikama:
bors try

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -411,4 +418,48 @@ object ProtoUtil {
...
+                       .toSet
+                 }
+                 .map(_.flatten.toSet)
+    } yield result -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
+
+  private def getCreatorJustificationUnlessGoal[F[_]: Sync: BlockStore](
+      dag: BlockDagRepresentation[F],
+      block: BlockMetadata,
+      goal: BlockMetadata
+  ): F[List[BlockMetadata]] =
```
itegulov: Why you returning `List` instead of `Option` here considering this function can only return 0 or 1 blocks? Is it because `DagOperations.bfTraverseF` expects `F[List[BlockMetadata]]`?

KentShikama: See https://discordapp.com/channels/391435113415573504/410182072054054912/570082282359226388

itegulov: Sorry I missed the conversation. I think your best bet was implementing such function with `tailRecM` like this:
```scala
def kent[F[_]: Monad](
    start: BlockMetadata
  )(nextFunction: BlockMetadata => F[BlockMetadata])(until: BlockMetadata => Boolean): F[Set[BlockMetadata]] =
    Monad[F].tailRecM((start, Set(start))) {
      case (block, result) =>
        if (!until(block)) {
          for {
            x <- nextFunction(block)
          } yield (x, result + x).asLeft[Set[BlockMetadata]]
        } else {
          result.asRight[(BlockMetadata, Set[BlockMetadata])].pure[F]
        }
    }
```

I am fine with keeping the code as it is though if you want; just wanted to share my thoughts on this

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
bors try-

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
bors try-

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
bors try-

KentShikama:
bors try

KentShikama:
Note I still owe a unit test here.

KentShikama:
bors try

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,4 +426,77 @@ object ProtoUtil {
...
+                               )
+                           )
+                           .map(_.blockHash)
+                           .toSet
+                       case None =>
+                         Set.empty[BlockHash].pure[F]
+                     }
+                 }
+                 .map(_.flatten.toSet)
+    } yield result -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
```
odeac: The body of the traverse, being so big, deserves a name and a type signature in order to be easier to understand. I find something like the code below to be more readable if you replace `computeSomething` and `result` with proper names.

```scala
  def unseenBlockHashes[F[_]: Sync: BlockStore](
      block: BlockMessage
  ): F[Set[BlockHash]] = {
    def computeSomething(
        latestMessagesOfBlock : Map[Validator, BlockMetadata], latestMessages : Map[Validator, BlockMetadata]) (
        validator : Validator, latestMessage : BlockMetadata
    ) : F[Set[ByteString]] =
      latestMessagesOfBlock
        .get(validator)
        .map(latestMessageOfBlockByValidator =>
          DagOperations
            .bfTraverseF(List(latestMessage))(
              block =>
                getCreatorJustificationUnlessGoal(
                  dag,
                  block,
                  latestMessageOfBlockByValidator
                )
            )
            .map(_.blockHash)
            .toSet
        )
        .getOrElse(Set.empty[BlockHash])

    for {
      latestMessages <- dag.latestMessages
      latestMessagesOfBlock <- ProtoUtil.toLatestMessage(block.justifications, dag)
      result <- 
        latestMessages
          .toStream
          .traverse(Function.tupled(computeSomething(latestMessagesOfBlock, latestMessages)))
    } yield result.flatten.toSet -- latestMessagesOfBlock.values.map(_.blockHash) - block.blockHash
  }
```

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,4 +426,77 @@ object ProtoUtil {
...
 
+  // Return hashes of all blocks that are yet to be seen by the passed in block
+  def unseenBlockHashes[F[_]: Sync: BlockStore](
+      dag: BlockDagRepresentation[F],
+      block: BlockMessage
+  ): F[Set[BlockHash]] =
+    for {
+      latestMessages        <- dag.latestMessages
+      latestMessagesOfBlock <- ProtoUtil.toLatestMessage(block.justifications, dag)
+      result <- latestMessages.toList
```
odeac: result is a bad name. Please use use something that describes the content

odeac: toList is inferior to toStream because it has to create an actual List object containing the data while the Stream is lazy and it only iterates through the underlying data structure

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -422,4 +426,77 @@ object ProtoUtil {
...
+        dag.lookup(hash).flatMap {
+          case Some(creatorJustification) =>
+            if (creatorJustification == goal) {
+              List.empty[BlockMetadata].pure[F]
+            } else {
+              List(creatorJustification).pure[F]
+            }
+          case None =>
+            Sync[F].raiseError[List[BlockMetadata]](
+              new RuntimeException(s"Missing block hash $hash in block dag.")
```
odeac: when would a block hash be missing from the dag?

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
odeac: OK. If the behavior of my code is different then the case is closed :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -136,6 +144,29 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+          Expr(
+            Expr.ExprInstance.EMapBody(
+              ParMap(SortedParMap(invalidBlocks.map {
+                case (validator, blockHash) =>
+                  (byteStringToPar(validator), byteStringToPar(blockHash))
+              }))
+            )
+          )
+        )
+      )
```
odeac: isn't this a case for a new RhoType.Map? You could create one with only the apply function and leave the unapply for when we have the first usage

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -136,6 +144,29 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+    runtime.invalidBlocks.setParams(invalidBlocksPar)
+  }
+
+  private def byteStringToPar(b: ByteString): Par =
+    Par(exprs = Seq(Expr(Expr.ExprInstance.GByteArray(b))))
```
odeac: we alreay have a RhoType.ByteArray that does the same job

casper/src/test/scala/coop/rchain/casper/genesis/contracts/PoSSpec.scala
```diff
@@ -9,5 +9,5 @@ class PoSSpec
     extends RhoSpec(
       CompiledRholangSource("PoSTest.rho"),
       Seq.empty,
-      20.seconds
+      60.seconds
```
odeac: it looks like we should have a default timeout for the timeout parameter in RhoSpec so we only make one change in situations like this

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -32,7 +32,7 @@ object RhoSpec {
       testResultCollector: TestResultCollector[F]
   ): Seq[SystemProcess.Definition[F]] = {
     val testResultCollectorService =
-      Seq((5, "assertAck", 25), (1, "testSuiteCompleted", 26))
+      Seq((5, "assertAck", 26), (1, "testSuiteCompleted", 27))
```
odeac: I assume you had to change these because of the conflicts with the production system processes. If that's the case I think it was discussed in the standup that we should use a higher number. What about if the test ones start at 1000?

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
KentShikama: It should never be (unless you directly fiddled with the filesystem) and thus the raiseError

casper/src/main/resources/PoS.rhox
```diff
@@ -83,6 +83,23 @@ new  PoS,
...
+
+            contract PoS(@"slash", @blockHash, return) = {
+              new userCh, invalidBlocksCh, getInvalidBlocks(`rho:casper:invalidBlocks`) in {
+                getInvalidBlocks!(*invalidBlocksCh) |
+                getUser!(*userCh) |
+                for (@invalidBlocks <- invalidBlocksCh; @bonds <- bondsCh; @userPk <- userCh) {
+                  new toBeSlashed in {
+                    toBeSlashed!(invalidBlocks.getOrElse(blockHash, userPk)) |
+                    for (@validator <- toBeSlashed) {
+                      // TODO: Transfer to coop wallet instead of just simply setting bonds to 0
```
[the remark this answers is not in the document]
KentShikama: https://rchain.atlassian.net/browse/RCHAIN-3532

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
KentShikama: We can create it if we need to build a second instance

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
KentShikama: Bytes only go until 127 but sounds good

KentShikama:
PTAL

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
odeac: I don't agree. We have rhotype objects for the other types so, for consistency reasons we should have one for map too. It's trivial to do so I don't think it's worth postponing

PR #2385, opened 2019-04-23
[Rchain 3135] - Pay for gas with Rev

JosDenmark:
## Overview
This PR adds balance verification and REV/phlo payment to the `computeState` function in `RuntimeManager`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3135

### Notes
This is the first of two PR's required to fully integrate the execution model with the phlo framework. The next PR should add balance verification and REV/phlo payment to the `replayComputeState` function of `RuntimeManager`. This would allow validation of said payments and is a prerequisite for integrating vaults, phlo, and consensus.

### DIFF
--- a/casper/src/main/resources/NonNegativeNumber.rho
+++ b/casper/src/main/resources/NonNegativeNumber.rho
@@ -17,7 +17,7 @@ new NonNegativeNumber, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
       contract this(@"add", @x, success) = {
         if (x >= 0) {
           for(@v <- valueStore){
-            if (v + x > v) {
+            if (v + x >= v) {
               valueStore!(v + x) | success!(true)
             } else {
               //overflow
--- a/casper/src/main/resources/PoS.rho
+++ b/casper/src/main/resources/PoS.rho
@@ -18,7 +18,8 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
 
  for (@minimumBond, @maximumBond <- PoS) {
 
-    match "11112GofKy1v5mPWZjZEYBByvpZ4ahYJp6vFN1hgGCn7L543rFSpyA" {
+ // TODO: Move this code into genesis.contracts.ProofOfStake to parameterize it over public key and address.
+    match "1111jXe5SseJpLMkUbXQN9k9aNHifUVgKm1pew3fwydaahdtvRRo1" {
 
       posRevAddress => {
 
--- a/casper/src/main/resources/PoS.rho
+++ b/casper/src/main/resources/PoS.rho
@@ -121,7 +121,7 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
               }
 
               //Assuming that user public key is valid, a vault should always be returned.
-              | contract deposit (@amount, return) = {
+              | contract deposit (@revAmount, return) = {
                 new parametersCh,
                   vaultCh,
                   revAddressCh,
@@ -134,7 +134,7 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
                       | for (@(true, vault) <- vaultCh) {
                         @RevVault!("deployerAuthKey", *authKeyCh)
                         | for (userAuthKey <- authKeyCh){
-                          @vault!("transfer", posRevAddress, amount, *userAuthKey, *return)
+                          @vault!("transfer", posRevAddress, revAmount, *userAuthKey, *return)
                         }
                       }
                     }
@@ -152,14 +152,14 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
                 }
               }
 
-              | contract PoS (@"pay", @amount, return) = {
+              | contract PoS (@"pay", @revAmount, return) = {
                 new parametersCh, depositRet in {
-                  deposit!(amount, *depositRet)
+                  deposit!(revAmount, *depositRet)
                   | getParameters!(*parametersCh)
                   | for (@depositResult <- depositRet; @codeHash, @phloRate, @userPk, @timestamp <- parametersCh; lastPayment <- paymentStore) {
                     match depositResult {
-                      (true, Nil) => { paymentStore!([codeHash, userPk, timestamp, phloRate * amount]) | return!(phloRate *amount) }
-                      (false, err) => { paymentStore!([codeHash, userPk, timestamp, 0]) | return!(0) }
+                      (true, Nil) => { paymentStore!([codeHash, userPk, timestamp, revAmount]) | return!(true) }
+                      (false, err) => { paymentStore!([codeHash, userPk, timestamp, 0]) | return!(err) }
                     }
                   }
                 }
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -74,6 +74,20 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
       evaluateResult                           <- doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
     } yield evaluateResult
 
+  private def replayComputeEffect(
+      runtime: Runtime[F]
+  )(start: Blake2b256Hash, processedDeploy: InternalProcessedDeploy): F[EvaluateResult] = {
+    import processedDeploy._
+    for {
+      _ <- runtime.replaySpace.rig(start, log.toList)
+      (codeHash, phloPrice, userId, timestamp) = ProtoUtil.getRholangDeployParams(
+        deploy
+      )
+      _         <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
+      injResult <- doInj(deploy, runtime.replayReducer, runtime.errorLog)(runtime.cost)
+    } yield injResult
+  }
+
   /**
     * @note `replayEval` does not need to reset the evaluation store,
     *       merely the replay store. Hence, `replayComputeState` uses
@@ -226,16 +240,12 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     ): F[Either[(Option[DeployData], Failed), StateHash]] =
       Concurrent[F].defer {
         terms match {
-          case InternalProcessedDeploy(deploy, _, log, status) +: rem =>
-            val (codeHash, phloPrice, userId, timestamp) = ProtoUtil.getRholangDeployParams(
-              deploy
-            )
+          case processedDeploy +: rem =>
+            import processedDeploy._
             for {
-              _         <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
-              _         <- runtime.replaySpace.rig(hash, log.toList)
-              injResult <- doInj(deploy, runtime.replayReducer, runtime.errorLog)(runtime.cost)
+              replayEvaluateResult <- replayComputeEffect(runtime)(hash, processedDeploy)
               //TODO: compare replay deploy cost to given deploy cost
-              EvaluateResult(cost, errors) = injResult
+              EvaluateResult(cost, errors) = replayEvaluateResult
               cont <- DeployStatus.fromErrors(errors) match {
                        case int: InternalErrors => Left(Some(deploy) -> int).pure[F]
                        case replayStatus =>
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -41,7 +41,7 @@ trait RuntimeManager[F[_]] {
   def computeBonds(hash: StateHash): F[Seq[Bond]]
   def getData(hash: StateHash)(channel: Par): F[Seq[Par]]
   def getContinuation(hash: StateHash)(
-      channels: immutable.Seq[Par]
+      channels: Seq[Par]
   ): F[Seq[(Seq[BindPattern], Par)]]
   def emptyStateHash: StateHash
 }
@@ -164,7 +164,7 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
 
   def getContinuation(
       hash: StateHash
-  )(channels: immutable.Seq[Par]): F[Seq[(Seq[BindPattern], Par)]] =
+  )(channels: Seq[Par]): F[Seq[(Seq[BindPattern], Par)]] =
     withResetRuntime(hash)(
       _.space
         .getWaitingContinuations(channels)
@@ -352,9 +352,7 @@ object RuntimeManager {
       override def getData(hash: StateHash)(channel: Par): T[F, scala.Seq[Par]] =
         runtimeManager.getData(hash)(channel).liftM[T]
 
-      override def getContinuation(hash: StateHash)(
-          channels: scala.collection.immutable.Seq[Par]
-      ): T[F, scala.Seq[(scala.Seq[BindPattern], Par)]] =
+      override def getContinuation(hash: StateHash)(channels: Seq[Par]): T[F, scala.Seq[(scala.Seq[BindPattern], Par)]] =
         runtimeManager.getContinuation(hash)(channels).liftM[T]
 
       override val emptyStateHash: StateHash = runtimeManager.emptyStateHash
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -72,7 +72,7 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
       for {
         evaluateResult <- computeEffect(runtime)(deploy)
         result <- if (evaluateResult.errors.isEmpty) getData(runtime)(name)
-               else Seq.empty[Par].pure[F]
+                 else Seq.empty[Par].pure[F]
       } yield result
     }
 
@@ -134,27 +134,28 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
         case Left(_)      => None
       }
 
-  def computeBonds(hash: StateHash): F[Seq[Bond]] = {
-    val bondsQuery =
-      """new rl(`rho:registry:lookup`), SystemInstancesCh, posCh in {
-        |  rl!(`rho:id:wdwc36f4ixa6xacck3ddepmgueum7zueuczgthcqp6771kdu8jogm8`, *SystemInstancesCh) |
-        |  for(@(_, SystemInstancesRegistry) <- SystemInstancesCh) {
-        |    @SystemInstancesRegistry!("lookup", "pos", *posCh) |
-        |    for(pos <- posCh){ pos!("getBonds", "__SCALA__") }
-        |  }
-        |}""".stripMargin
-
-    captureResults(hash, ConstructDeploy.sourceDeployNow(bondsQuery))
+  def computeBonds(hash: StateHash): F[Seq[Bond]] =
+    captureResults(hash, ConstructDeploy.sourceDeployNow(bondsQuerySource()))
       .ensureOr(
         bondsPar =>
           new IllegalArgumentException(
-            s"Incorrect number of results from query of current bonds: ${bondsPar.size}"
-          )
+    s"Incorrect number of results from query of current bonds: ${bondsPar.size}"
+  )
       )(bondsPar => bondsPar.size == 1)
       .map { bondsPar =>
         toBondSeq(bondsPar.head)
       }
-  }
+
+  private def bondsQuerySource(name: String = "__SCALA__"): String =
+    s"""
+       | new rl(`rho:registry:lookup`), SystemInstancesCh, posCh in {
+       |   rl!(`rho:id:wdwc36f4ixa6xacck3ddepmgueum7zueuczgthcqp6771kdu8jogm8`, *SystemInstancesCh) |
+       |   for(@(_, SystemInstancesRegistry) <- SystemInstancesCh) {
+       |     @SystemInstancesRegistry!("lookup", "pos", *posCh) |
+       |     for(pos <- posCh){ pos!("getBonds", "$name") }
+       |   }
+       | }
+     """.stripMargin
 
   def computeBalance(start: StateHash)(user: ByteString): F[Long] =
     withResetRuntime(start)(computeBalance(_)(user))
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -17,13 +17,12 @@ import coop.rchain.models._
 import coop.rchain.rholang.interpreter.accounting._
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
 import coop.rchain.rholang.interpreter.{
-  ChargingReducer,
-  ErrorLog,
-  EvaluateResult,
-  Interpreter,
-  Runtime
-}
-import coop.rchain.rspace.internal.Datum
+    ChargingReducer,
+    ErrorLog,
+    EvaluateResult,
+    Interpreter,
+    Runtime
+  }
 import coop.rchain.rspace.{Blake2b256Hash, ReplayException}
 
 trait RuntimeManager[F[_]] {
@@ -62,12 +61,19 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     withResetRuntime(start) { runtime =>
       //TODO: Is better error handling needed here?
       for {
-        evalR                     <- newEval(deploy :: Nil, runtime, start)
-        (_, Seq(processedDeploy)) = evalR
-        result <- if (processedDeploy.status.isFailed) Seq.empty[Datum[ListParWithRandom]].pure[F]
-                 else runtime.space.getData(name).map(_.toSeq)
-      } yield result.flatMap(_.a.pars)
+        evaluateResult <- computeEffect(runtime)(deploy)
+        result <- if (evaluateResult.errors.isEmpty) getData(runtime)(name) else Seq.empty[Par].pure[F]
+      } yield result
     }
+
+  private def computeEffect(runtime: Runtime[F])(deploy: DeployData): F[EvaluateResult] =
+    for {
+      runtimeParameters                        <- ProtoUtil.getRholangDeployParams(deploy).pure[F]
+      (codeHash, phloPrice, userId, timestamp) = runtimeParameters
+      _                                        <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
+      evaluateResult                           <- doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
+    } yield evaluateResult
+
   /**
     * @note `replayEval` does not need to reset the evaluation store,
     *       merely the replay store. Hence, `replayComputeState` uses
@@ -128,7 +134,7 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
         bondsPar =>
           new IllegalArgumentException(
             s"Incorrect number of results from query of current bonds: ${bondsPar.size}"
-        )
+          )
       )(bondsPar => bondsPar.size == 1)
       .map { bondsPar =>
         toBondSeq(bondsPar.head)
@@ -155,7 +161,10 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     }.toList
 
   def getData(hash: StateHash)(channel: Par): F[Seq[Par]] =
-    withResetRuntime(hash)(_.space.getData(channel).map(_.flatMap(_.a.pars)))
+    withResetRuntime(hash)(getData(_)(channel))
+
+  private def getData(runtime: Runtime[F])(channel: Par): F[Seq[Par]] =
+    runtime.space.getData(channel).map(_.flatMap(_.a.pars))
 
   def getContinuation(
       hash: StateHash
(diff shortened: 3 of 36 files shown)


JosDenmark:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -158,11 +162,82 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+  def computeBalance(start: StateHash)(user: ByteString): F[Long] =
+    withResetRuntime(start)(computeBalance(_)(user))
+
+  private def balanceQuerySource(user: ByteString): String =
+    s"""
+       | new rl(`rho:registry:lookup`), revAddressOps(`rho:rev:address`), revVaultCh in {
+       |   rl!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh)
+       |   | for (@(_, RevVault) <- revVaultCh) {
+       |     new vaultCh, revAddressCh in {
+       |       revAddressOps!("fromPublicKey", "${Base16.encode(user.toByteArray)}".hexToBytes(), *revAddressCh)
```
lukasz-golebiewski: Wasn't `"${Base16.encode(user.toByteArray)}"` supposed to be interpolated?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -158,11 +162,82 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
-      val blakeHash = Blake2b256Hash.fromByteArray(hash.toByteArray)
-      runtime.space.reset(blakeHash).flatMap(_ => block(runtime))
-    }(runtime => runtimeContainer.put(runtime))
+  def computeBalance(start: StateHash)(user: ByteString): F[Long] =
+    withResetRuntime(start)(computeBalance(_)(user))
+
+  private def balanceQuerySource(user: ByteString): String =
+    s"""
+       | new rl(`rho:registry:lookup`), revAddressOps(`rho:rev:address`), revVaultCh in {
+       |   rl!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh)
```
lukasz-golebiewski: The `rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z` could be extracted to a constant, otherwise it's a magic string

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -174,69 +249,130 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
-                DeployStatus.fromErrors(errors)
-              )
-              cont <- if (errors.isEmpty)
-                       doEval(rem, newCheckpoint.root, acc :+ deployResult)
-                     else doEval(rem, hash, acc :+ deployResult)
-            } yield cont
-
-          case _ => (ByteString.copyFrom(hash.bytes.toArray), acc).pure[F]
+      runtime: Runtime[F]
+  )(deploys: Seq[DeployData], withPayment: Boolean): F[(StateHash, Seq[InternalProcessedDeploy])] =
```
lukasz-golebiewski: The cyclomatic complexity of this function makes it incomprehensible. Could you extract some clarifying functions?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -174,69 +249,130 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+    withResetRuntime(start)(
+      _.space
+        .getWaitingContinuations(channels)
+        .map(
+          _.filter(_.continuation.taggedCont.isParBody)
+            .map(result => (result.patterns, result.continuation.taggedCont.parBody.get.body))
+        )
     )
 
   private def newEval(
```
ArturGajowy: The commit name `Eliminate superfluous reset from newEval` is a bit misleading. "Superfluous" would suggest it being removed whatsoever, while it's just been moved up to `computeState`.

Also, separating the changes to `replayComputeState` would prevent me bing mislead about its behavior (not) being changed in this commit.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: Porbably having a separate `newEval / computeState` for genesis would help. Then this one could always charge, and the 'genesis' one would never charge. Similarly, the genesis one could avoid a dozen checkpoints done in between the genesis deploys, most likely speeding up test setup and genesis block creation. It's "all or nothing" for genesis, after all.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -158,11 +138,13 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
-    Sync[F].bracket(runtimeContainer.take) { runtime =>
-      val blakeHash = Blake2b256Hash.fromByteArray(hash.toByteArray)
-      runtime.space.reset(blakeHash).flatMap(_ => block(runtime))
-    }(runtime => runtimeContainer.put(runtime))
+  private def withRuntime[A](f: Runtime[F] => F[A]): F[A] =
+    Sync[F].bracket(runtimeContainer.take)(f)(runtimeContainer.put)
+
+  private def withResetRuntime[A](start: StateHash)(f: Runtime[F] => F[A]): F[A] =
+    withRuntime(
+      runtime => runtime.space.reset(Blake2b256Hash.fromByteArray(start.toByteArray)) *> f(runtime)
```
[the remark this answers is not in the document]
ArturGajowy: Please make sure all the usages of `*>` in this PR are changed to `>>`, unless you have good rationale for using `*>` in any places. I'm pretty sure we've investigated bugs around `*>` usage with @rabbitonweb in the past :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: Yup, commenting out the mid-deploy checkpointing in newEval drops the RevVaultSpec startup time from 20s to 14s. So I guess having a separate 'genesisEval' is worth giving a shot for both performance and complexity reasons.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -65,44 +65,62 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+    } yield evaluateResult
+
+  private def replayComputeEffect(
+      runtime: Runtime[F]
+  )(start: Blake2b256Hash, processedDeploy: InternalProcessedDeploy): F[EvaluateResult] = {
+    import processedDeploy._
+    for {
+      _                                        <- runtime.replaySpace.rig(start, deployLog.toList)
+      (codeHash, phloPrice, userId, timestamp) = ProtoUtil.getRholangDeployParams(deploy)
+      _                                        <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
```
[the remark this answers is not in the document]
JosDenmark: I agree. I'll add a `TODO`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -158,11 +162,82 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
 
-  private def withResetRuntime[R](hash: StateHash, block: Runtime[F] => F[R]) =
-    Sync[F].bracket(runtimeContainer.take) { runtime =>
-      val blakeHash = Blake2b256Hash.fromByteArray(hash.toByteArray)
-      runtime.space.reset(blakeHash).flatMap(_ => block(runtime))
-    }(runtime => runtimeContainer.put(runtime))
+  def computeBalance(start: StateHash)(user: ByteString): F[Long] =
+    withResetRuntime(start)(computeBalance(_)(user))
+
+  private def balanceQuerySource(user: ByteString): String =
```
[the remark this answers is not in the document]
JosDenmark: 👍

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: @lukasz-golebiewski My argument is the same as Artur's. Are you convinced?

casper/src/main/resources/PoS.rho
```diff
@@ -18,7 +18,7 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
 
  for (@minimumBond, @maximumBond <- PoS) {
 
-    match "11112GofKy1v5mPWZjZEYBByvpZ4ahYJp6vFN1hgGCn7L543rFSpyA" {
+    match "1111jXe5SseJpLMkUbXQN9k9aNHifUVgKm1pew3fwydaahdtvRRo1" {
```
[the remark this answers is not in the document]
JosDenmark: It's hardcoded because the public key from which the address is derived is hardcoded. This could be parameterized if we unified `PoS.rho` with the `ProofOfStake` generated code.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: @lukasz-golebiewski 

> The cyclomatic complexity of this function makes it incomprehensible. Could you extract some clarifying functions?

Can do.

@ArturGajowy 

> Porbably having a separate newEval / computeState for genesis would help.

I've considered this option, but it seemed sloppy. After all, that function would only be used once. It seems like those deploys should be evaluated in the constructor of RuntimeManager, and `emptyState` should be set to the result of those evaluations.

JosDenmark: At any rate, a separate function for initializing the manager would eliminate the need for the `withPayment` flag. That would greatly simplify this `newEval`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
lukasz-golebiewski: Of course it is, my bad

casper/src/main/resources/NonNegativeNumber.rho
```diff
@@ -17,7 +17,7 @@ new NonNegativeNumber, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
       contract this(@"add", @x, success) = {
         if (x >= 0) {
           for(@v <- valueStore){
-            if (v + x > v) {
+            if (v + x >= v) {
```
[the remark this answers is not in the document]
ArturGajowy: Yup! I think such test should fail before your change

casper/src/test/scala/coop/rchain/casper/MultiParentCasperTestUtil.scala
```diff
@@ -65,13 +65,30 @@ object MultiParentCasperTestUtil {
...
+          maximumBond = Long.MaxValue,
+          validators = bonds.toSeq.map(Validator.tupled)
+        ),
+        faucet = false,
+        genesisPk = Ed25519.newKeyPair._2,
+        vaults = bonds.toList.map {
+          case (pk, stake) =>
+            RevAddress.fromPublicKey(pk).map(Vault(_, stake))
+        }.flattenOption,
+        supply = Long.MaxValue
```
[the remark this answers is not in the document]
ArturGajowy: I'm fine with removing the commit as long as the resulting merge resolutions adheres to the 'Either bonds or validatorPKs - there can be only one' comment, and does not reintroduce the duplication pointed out above :) My point is: we've already decreased the entropy in those parts of code, let's not reintroduce it.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
lukasz-golebiewski: This contract id is already used in many places. I think it would make sense to introduce a `rho:contracts:vault` mapping instead of using an obscure hash in a separate PR

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: I'm not sure we're talking about the same commit. I also think we can leave the commit I mentioned above as is (unless I made other comments to it).

Trying to answer your question: the changes to `replayComputeState` in that commit would best be split as a separate commit saying sth like 'rename the `hash` parameter to `start` in `replayComputeState`', because of the reason I gave above (I thought the reset was also eliminated from replay for quite some time, partly because of the comment you added too - which could easily be yet another commit.)

All this conversation is only useful if it makes our future PR reviews more understandable - no need for changes now :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -65,44 +65,62 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+      runtimeParameters                        <- ProtoUtil.getRholangDeployParams(deploy).pure[F]
+      (codeHash, phloPrice, userId, timestamp) = runtimeParameters
+      _                                        <- runtime.shortLeashParams.setParams(codeHash, phloPrice, userId, timestamp)
+      evaluateResult                           <- doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
+    } yield evaluateResult
+
+  private def replayComputeEffect(
+      runtime: Runtime[F]
+  )(start: Blake2b256Hash, processedDeploy: InternalProcessedDeploy): F[EvaluateResult] = {
+    import processedDeploy._
```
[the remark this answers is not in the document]
ArturGajowy: How about extracting two vals for `log` and `deploy`? It'd make obvious what parts of InternalProcessedDeploy are used.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -65,44 +65,62 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+    withResetRuntime(start) { runtime =>
+      //TODO: Is better error handling needed here?
+      for {
+        evaluateResult <- computeEffect(runtime)(deploy)
+        result <- if (evaluateResult.errors.isEmpty) getData(runtime)(name)
+                 else Seq.empty[Par].pure[F]
+      } yield result
+    }
+
+  private def computeEffect(runtime: Runtime[F])(deploy: DeployData): F[EvaluateResult] =
```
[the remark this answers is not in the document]
ArturGajowy: Either that, or just the rename would be OK :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -2,58 +2,58 @@ package coop.rchain.casper.util.rholang
...
 import cats.effect.concurrent.MVar
+import cats.effect.{Sync, _}
 import cats.implicits._
 import com.google.protobuf.ByteString
 import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
-import coop.rchain.catscontrib.Catscontrib._
+import coop.rchain.casper.util.{ConstructDeploy, ProtoUtil}
+import coop.rchain.catscontrib.Catscontrib.ToMonadOps
```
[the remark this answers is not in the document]
ArturGajowy: Was this IDEA-only then? Pointing this out in the commit message would spare these comments :)

casper/src/main/resources/PoS.rho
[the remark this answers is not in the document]
ArturGajowy: We should un-hardcode both of those places, probably before we progress with the next PR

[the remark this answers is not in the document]
ArturGajowy: Having this hardcoded could be hiding quite a lot of dependencies which in turn might be making my understanding even more partial than it is already :wink:

casper/src/main/resources/PoS.rho
```diff
@@ -158,8 +158,8 @@ new PoS, rs(`rho:registry:insertSigned:ed25519`), uriOut in {
                   | getParameters!(*parametersCh)
                   | for (@depositResult <- depositRet; @codeHash, @phloRate, @userPk, @timestamp <- parametersCh; lastPayment <- paymentStore) {
                     match depositResult {
-                      (true, Nil) => { paymentStore!([codeHash, userPk, timestamp, phloRate * amount]) | return!(phloRate *amount) }
-                      (false, err) => { paymentStore!([codeHash, userPk, timestamp, 0]) | return!(0) }
+                      (true, Nil) => { paymentStore!([codeHash, userPk, timestamp, amount]) | return!(true) }
+                      (false, err) => { paymentStore!([codeHash, userPk, timestamp, 0]) | return!(err) }
```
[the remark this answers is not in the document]
ArturGajowy: By 'convention' I meant 'a way of encoding error results', not 'the way that we're supposed to use' :smile: 

>  It also makes value extraction more difficult in the in the computeDeployPayment function.

Yeah, I haven't got to it yet. Let's park this for now, it's probbly OK.

The benefit of having the bonding deploy's signature stored on-chain would be being able to cross-check and debug better. The deploy's signature is the closest we have to a canonical deployId.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: I'm not sure about the `emptyState`, but having two funcitons instead of one with a boolean flag would help greatly. Having one of them skip the in-the-middle checkpoints would further make things clearer, and faster at the same time.

JosDenmark: That introduces the issue that I presented last time we spoke. Casper doesn't discriminate between the genesis deploys and the deploys that follow. All of Casper assumes that the genesis deploys are validated like any other deploys, and so they're replayed like any other deploys. That means that we have to checkpoint them like any other deploys. Changing Casper to recognize the genesis deploys seems complex. I'd appreciate input from @KentShikama here.

casper/src/main/resources/PoS.rho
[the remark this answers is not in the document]
JosDenmark: I think that verification is performed implicitly by the signature check in the node runtime. At the point this contract is running, we've verified that `userPk` is the true identity of the deployer.

casper/src/main/resources/NonNegativeNumber.rho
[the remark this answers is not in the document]
JosDenmark: 👍

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: It looks like `computeBlockCheckpointFromDeploys` in `InterpreterUtil` [https://github.com/rchain/rchain/blob/05facc171003298ea95d5f640aa9082ac9708e93/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala#L264](url) is a viable option for discriminating between genesis and non-genesis deploys. It calls `computeDeploysCheckpoint`, which calls `computeState`. This would be the place to call an alternative method that just ran the genesis deploys.

casper/src/main/resources/PoS.rho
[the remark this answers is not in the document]
ArturGajowy: My point is to store the signature just in case, for better debugging ability, and the ability to do a **cross-**check - so that's another check should the previous pass unjustly. Also, deploySignature is verifying more than the deployer identity - it verifies a deploy.

I'm not saying we need it in the PoS contract - but at the same time I find it strange that the PoS contract doesn't need it.

@KentShikama WDYT?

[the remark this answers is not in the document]
KentShikama: The deployer's identity can be accessed by the userPk right?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: That's a good one, in that it's actually only used in tests... I'll push a commit moving it to src/test.

Anyway, I don't think the intermediate checkpoints (hashes) are needed for replay. Are they?

ArturGajowy: Nope, I won't. Please cherry-pick commit 31e83fb2 from my remote.

casper/src/main/resources/PoS.rho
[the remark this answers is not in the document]
ArturGajowy: The question is whether it makes sense to store the bonding deploy's signature,  so that we can later easily point to 'you've singed with your blood, here'.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -32,18 +38,205 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
...
+        val deploy0 =
+          SignDeployment.sign(
+            userSk,
+            ConstructDeploy
+              .sourceDeployNow(program)
+              .withDeployer(ByteString.copyFrom(userPk.bytes))
+              .withPhloPrice(phloPrice)
+              .withPhloLimit(phloLimit)
+          )
+        forAll(minSuccessful(1)) { withPayment: Boolean =>
```
ArturGajowy: instead of a generative test over a Boolean domain, with an if/match inside - let's have two tests for two cases. Especially that those should be two different methods for with and without payment.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -32,18 +38,205 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
...
+    val genesisTerms = Seq(
+      StandardDeploys.listOps,
+      StandardDeploys.either,
+      StandardDeploys.nonNegativeNumber,
+      StandardDeploys.makeMint,
+      StandardDeploys.lockbox,
+      StandardDeploys.PoS,
+      StandardDeploys.authKey,
+      StandardDeploys.revVault,
+      StandardDeploys.revGenerator(
```
ArturGajowy: This test setup needs deduplication and denoising

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -158,11 +162,82 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+
+  def computeDeployPayment(start: StateHash)(user: ByteString, amount: Long): F[Unit] =
+    withResetRuntime(start)(computeDeployPayment(_)(user, amount))
+
+  private def deployPaymentSource(amount: Long): String =
+    s"""
+       | new rl(`rho:registry:lookup`), poSCh in {
+       |   rl!(`rho:id:cnec3pa8prp4out3yc8facon6grm3xbsotpd4ckjfx8ghuw77xadzt`, *poSCh) |
+       |   for(@(_, PoS) <- poSCh) {
+       |     @PoS!("pay", $amount, "__SCALA__")
```
ArturGajowy: this should go after `computeDeployPayment`, and not have `__SCALA__` hardcoded

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -174,69 +249,130 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
-          case _ => (ByteString.copyFrom(hash.bytes.toArray), acc).pure[F]
+      runtime: Runtime[F]
+  )(deploys: Seq[DeployData], withPayment: Boolean): F[(StateHash, Seq[InternalProcessedDeploy])] =
+    Concurrent[F]
+      .defer {
+        runtime.space.createCheckpoint() >>= { checkpoint0 =>
+          deploys.toList.foldM((checkpoint0.root, Seq.empty[InternalProcessedDeploy])) {
+            case ((prev, processedDeploys), deploy) =>
+              import deploy._
+              if (withPayment) {
```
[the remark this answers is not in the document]
JosDenmark: I'm fine with that. The reason it's not used in production code is because `replayComputeState` does not yet replay the payment. Thus, the replay state would never match the computed state.

casper/src/main/resources/PoS.rho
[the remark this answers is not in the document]
JosDenmark: @ArturGajowy The signature check in the node verifies the validity of the deploy, which includes all of the runtime parameters used above, so the signature is implicitly included. We didn't have signature verification until recently, so we couldn't remove the signature until recently. That we can remove it demonstrates progress :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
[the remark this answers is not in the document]
JosDenmark: 👎

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -144,6 +146,48 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+       |         }
+       |       }
+       |     }
+       |   }
+       | }
+     """.stripMargin
+
+  // TODO: Determine if accumulated values on @"__SCALA__" interfere with balance/payment processing.
+  private def computeBalance(runtime: Runtime[F])(user: ByteString): F[Long] =
+    computeEffect(runtime)(
```
KentShikama: So is the call to computeBalance followed by a createCheckpoint call before we call reset - as in are we permanently changing the tuplespace trie?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
KentShikama: We need the replay log for the genesis block because some genesis deploys are non-deterministic.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -276,41 +276,103 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+  )(deploys: Seq[DeployData], withPayment: Boolean): F[(StateHash, Seq[InternalProcessedDeploy])] =
+    Concurrent[F]
+      .defer {
+        runtime.space.createCheckpoint() >>= { checkpoint0 =>
+          deploys.toList.foldM((checkpoint0.root, Seq.empty[InternalProcessedDeploy])) {
+            case ((prev, processedDeploys), deploy) =>
+              import deploy._
+              if (withPayment) {
+                computeBalance(runtime)(deployer) >>= { balance =>
+                  runtime.space.reset(prev) *> {
```
KentShikama: I think this answers my previous question

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -276,41 +276,103 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+                      computeEffect(runtime)(deploy) >>= {
+                        case EvaluateResult(cost, errors) =>
+                          errors.isEmpty
+                            .pure[F]
+                            .ifM(
+                              ifTrue = ()
+                                .pure[F],
+                              ifFalse = runtime.space.reset(prev)
+                            ) *> {
+                            runtime.space.createCheckpoint() >>= { checkpoint1 =>
```
KentShikama: Maybe name this `deployEvalCheckpoint` and the below one `deployPaymentCheckpoint`

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosDenmark: No, the balance check will not change the tuplespace trie.

JosDenmark: @ArturGajowy Be sure to note the above as you continue the implementation.

casper/src/main/resources/NonNegativeNumber.rho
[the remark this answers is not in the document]
JosDenmark: So, just to be clear, I'd introduce a test prior to this commit that would fail for all refs prior to this commit. If all commits in the history should compile, shouldn't we try our best to have all commits pass tests as well?

JosDenmark:
bors try

JosDenmark:
@ArturGajowy I've made the all of the changes you requested, except for the test for the NonNegativeNumber bug (see my comment). The branch is set to allow edits from maintainers 😄. Feel free to `bors r+` it if/when the build passes.

casper/src/main/resources/NonNegativeNumber.rho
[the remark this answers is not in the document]
ArturGajowy: You would not :) You'd introduce the test with the fix, or just after it.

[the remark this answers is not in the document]
ArturGajowy: > I think such test should fail before your change

This was just supposed to mean "that's a valid test for your change"

ArturGajowy:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: @KentShikama thx for bringing that up. Can you point to concrete examples?

ArturGajowy: Also, notice we didn't discuss disabling replay - we discussed disabling mid-deploy checkpointing for the genesis block. AFAIU the intermediate hashes / checkpoints are not needed in any way

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: @KentShikama @JosDenmark I believe the contrary should be the case. If we're interacting with the tuplespace in any way, we must leave a `trace` of that. This way we constrain the potential for the validators cheating and make our lives easier when debugging any controversial behavior on chain.

PR #2383, opened 2019-04-23
RCHAIN-3250 Refactor CasperPacketHandler

rabbitonweb:
## Overview
Refactor CasperPacketHandler to better understand what's going on.

### JIRA ticket:
RCHAIN-3250

PR #2393, opened 2019-04-26
RCHAIN-3250 further refactoring

rabbitonweb:
## Overview
Most significant change in this PR is work around CasperMessage. There is now a sealed `CasperMessage` defined and so we have a proper ADT (e.g if you forget to include one of the messages in pattern matching, compiler will complain with an error)

Bunch of stuff was moved in and out of `CasperPacketHandler`, so that future generations have better grasp of understanding "what on earth is going on here".

Based on those changes, I can finally go bug hunting.

### JIRA ticket:
RCHAIN-3397


rabbitonweb:
WIP till https://github.com/rchain/rchain/pull/2383 gets merged

rabbitonweb:
@ArturGajowy @KentShikama this is now ready for review

casper/src/main/scala/coop/rchain/casper/protocol/package.scala
```diff
@@ -0,0 +1,58 @@
...
+      packetToForkChoiceTipRequest(packet) orElse
+      packetToApprovedBlock(packet) orElse
+      packetToApprovedBlockRequest(packet) orElse
+      packetToBlockMessage(packet) orElse
+      packetToBlockApproval(packet) orElse
+      packetToUnapprovedBlock(packet) orElse
+      packetToNoApprovedBlockAvailable(packet)
+
+  private def packetToBlockMessage(msg: Packet): Option[BlockMessage] =
+    if (msg.typeId == transport.BlockMessage.id)
```
lukasz-golebiewski: The `if (condition) Try(...) else None` logic is repeated many times - it would make sense to extract it to a helper function

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandlerInternal.scala
```diff
@@ -0,0 +1,371 @@
...
+import coop.rchain.comm.{transport, PeerNode}
+import coop.rchain.metrics.Metrics
+import coop.rchain.shared.{Log, LogSource, Time}
+import monix.eval.Task
+import monix.execution.Scheduler
+
+import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
+
+trait CasperPacketHandlerInternal[F[_]] {
```
lukasz-golebiewski: If this trait is internal, shouldn't this be enforced by a package private access modifier?

rabbitonweb: It's not internal. Name is confusing (I did not created the name). Wanted to talk / bike-shed about it yesterday - no one was interested... remember? :)

casper/src/main/scala/coop/rchain/casper/protocol/package.scala
rabbitonweb: 👍

PR #2403, opened 2019-05-03
RCHAIN-3338 Introduce func deploy_contract_with_substitution

zsluedem:
## Overview
Introduce func deploy_contract_with_substitution

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3338

### Notes

### DIFF
--- a/integration-tests/pylintrc
+++ b/integration-tests/pylintrc
@@ -64,8 +64,6 @@ disable=
     format,
     line-too-long,
     logging-format-interpolation,
-
-    # Node generate this
     too-many-public-methods,
 
     # pytest fixtures generate this
--- a/integration-tests/pylintrc
+++ b/integration-tests/pylintrc
@@ -65,6 +65,9 @@ disable=
     line-too-long,
     logging-format-interpolation,
 
+    # Node generate this
+    too-many-public-methods,
+
     # pytest fixtures generate this
     redefined-outer-name,
 
--- a/integration-tests/test/rnode.py
+++ b/integration-tests/test/rnode.py
@@ -166,7 +166,7 @@ def extract_block_hash_from_propose_output(propose_output: str) -> str:
     return match.group(1)
 
 
-class Node:  # pylint: disable=R0904
+class Node:
     def __init__(self, *, container: Container, deploy_dir: str, command_timeout: int, network: str) -> None:
         self.container = container
         self.local_deploy_dir = deploy_dir
--- a/integration-tests/test/rnode.py
+++ b/integration-tests/test/rnode.py
@@ -336,19 +336,21 @@ class Node:  # pylint: disable=R0904
         return Node.__log_message_rx.split(log_content)
 
     def deploy_contract_with_substitution(self, substitute_dict: Dict[str, str], rho_file_path: str, private_key: str) -> str:
-        # Supposed that you have a contract with content like below.
-        #
-        # new x in { x!("#DATA") }
-        #
-        # If you pass a dict {'#DATA': "123456"} as substitute_dict args in this func,
-        # this method would substitute the string #DATA in the contract with 123456, which turns out to be
-        #
-        # new x in { x!("123456") }
-        #
-        # And then deploy the contract in the node
+        """
+        Supposed that you have a contract with content like below.
+
+        new x in { x!("#DATA") }
+
+        If you pass a dict {'#DATA': "123456"} as substitute_dict args in this func,
+        this method would substitute the string #DATA in the contract with 123456, which turns out to be
+
+        new x in { x!("123456") }
+
+        And then deploy the contract in the node
+        """
         shutil.copyfile(rho_file_path, os.path.join(self.local_deploy_dir, os.path.basename(rho_file_path)))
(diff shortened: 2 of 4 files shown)


zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

zsluedem:
@adaszko ^

integration-tests/test/rnode.py
```diff
@@ -165,7 +166,7 @@ def extract_block_hash_from_propose_output(propose_output: str) -> str:
     return match.group(1)
 
 
-class Node:
+class Node:  # pylint: disable=R0904
```
adaszko: Is this `too-many-public-methods (R0904)`?  This class has just one public method.

integration-tests/test/rnode.py
```diff
@@ -334,6 +335,30 @@ def log_lines(self) -> List[str]:
...
+        # If you pass a dict {'#DATA': "123456"} as substitute_dict args in this func,
+        # this method would substitute the string #DATA in the contract with 123456, which turns out to be
+        #
+        # new x in { x!("123456") }
+        #
+        # And then deploy the contract in the node
+        shutil.copyfile(rho_file_path, os.path.join(self.local_deploy_dir, os.path.basename(rho_file_path)))
+        container_contract_file_path = os.path.join(self.remote_deploy_dir, os.path.basename(rho_file_path))
+        substitute_rules = ';'.join(['s/{}/{}/g'.format(key, value) for key, value in substitute_dict.items()])
+        self.shell_out(
```
adaszko: `/` should be probably escaped in `key` and `value`.

integration-tests/test/rnode.py
```diff
@@ -334,6 +335,30 @@ def log_lines(self) -> List[str]:
...
+        # Supposed that you have a contract with content like below.
+        #
+        # new x in { x!("#DATA") }
+        #
+        # If you pass a dict {'#DATA': "123456"} as substitute_dict args in this func,
+        # this method would substitute the string #DATA in the contract with 123456, which turns out to be
+        #
+        # new x in { x!("123456") }
+        #
+        # And then deploy the contract in the node
```
adaszko: Very instructive comment.  Would you mind doing it into a docstring?

integration-tests/test/rnode.py
zsluedem: Yes . It is `too-many-public-methods`. 
All the method not started with `_` is public methods in `Node`. You must count it wrong.

adaszko: How many are there in your opinion?

integration-tests/test/rnode.py
adaszko: Why has this been resolved?

integration-tests/test/rnode.py
adaszko: ^

zsluedem: 26? What do you want to do with it?

adaszko: Oh, I got fooled by GitHub's folding of code 🤦‍♂ 

So I have 2 remarks: First, let's please not use crypting codes like R0904 but full names instead like I do everywhere else.  Second, in this case, the warning doesn't seem that useful so let's put it in `pylintrc`: https://github.com/rchain/rchain/blob/fbddb0d04573267159411c439e58bc0410003b7a/integration-tests/pylintrc#L55

integration-tests/pylintrc
```diff
@@ -65,6 +65,9 @@ disable=
     line-too-long,
     logging-format-interpolation,
 
+    # Node generate this
```
adaszko:
```suggestion
```

adaszko: This comment isn't necessary

integration-tests/test/rnode.py
zsluedem: `'s/{}/{}/g'` -> `r's/{}/{}/g'`
Actually `/` doesn't have to be escaped while `\` should be escaped.
I put a `r` in front of the string which means this is the raw string and even the `\` doesn't have to be escaped.

zsluedem: https://stackoverflow.com/questions/2081640/what-exactly-do-u-and-r-string-flags-do-and-what-are-raw-string-literals

woky: Try to run
```
x=/
echo /foo/bar | sed "s/$x/;/"
```
You'll get an error. You can use different delimiters in sed, e.g. `|` or `#` instead of `/` to avoid this.

zsluedem: What about `x=#`? I don't think that's the way to solve the error.

adaszko: `/` as delimiters are fine.  It's just that they need to be escaped or else sed will fail on cases like `'s/{}/{}/g'.format('foo/bar', 'anything')`

adaszko: https://docs.python.org/3/library/shlex.html#shlex.quote should work better than str.replace

PR #2404, opened 2019-05-03
Change the gRPC API for visualising a graph to stream the resulting vdag

sebtomba:
## Overview
Change the gRPC API for visualising a graph to stream the resulting vdag

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3118

### DIFF
--- a/graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
+++ b/graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
@@ -1,10 +1,10 @@
 package coop.rchain.graphz
 
+import java.io.FileOutputStream
+
 import cats._, cats.data._, cats.implicits._
 import cats.effect.Sync
 import cats.mtl._
-
-import java.io.FileOutputStream
 
 trait GraphSerializer[F[_]] {
   def push(str: String, suffix: String = "\n"): F[Unit]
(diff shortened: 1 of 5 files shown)


graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -1,11 +1,13 @@
 package coop.rchain.graphz
 
+import java.io.FileOutputStream
+
+import scala.collection.mutable.ListBuffer
```
lukasz-golebiewski: What's the motivation behind using a mutable collection inside the `MonadState`?

sebtomba: @lukasz-golebiewski The reason for a mutable `ListBuffer` is that we want to append and not prepend lines of strings. Of course, we could prepend and then reverse the whole list later. But this seems error-prone to me.

lukasz-golebiewski: Why not use `Vector`?

sebtomba: @lukasz-golebiewski np

PR #2413, opened 2019-05-13
RCHAIN-3250 fix the bug

rabbitonweb:
## Overview

This PR fixes the RCHAIN-3250 bug. Bug existed because genesis validator was continuously responding to ALL UnapprovedBlocks.
Fix is straightforward, once genesis validator validated UnapprovedBlock, it immediately jumps to `Initializing` state, ignoring all other `UnapprovedBlock` messages

Screenshot 2019-05-14 at 16 47 46

Because it no longer validates each Unapproved Block message, the whole process is a lot faster. Here output from tests

Before `test_successful_genesis_ceremony` took:

```
1 passed, 27 deselected in 1087.99 seconds
```

Now it takes:

```
1 passed, 29 deselected in 449.71 second
```

### JIRA ticket:
RCHAIN-3250

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
@@ -167,14 +167,12 @@ object CasperLaunch {
     for {
       validators  <- CasperConf.parseValidatorsFile[F](init.conf.knownValidatorsFile)
       validatorId <- ValidatorIdentity.fromConfig[F](init.conf)
-      _ <- EngineCell[F].set(
-            new Initializing(
-              init.runtimeManager,
-              init.conf.shardId,
-              validatorId,
-              validators,
-              CommUtil.requestApprovedBlock[F]
-            )
+      _ <- Engine.tranistionToInitializing(
+            init.runtimeManager,
+            init.conf.shardId,
+            validatorId,
+            validators,
+            CommUtil.requestApprovedBlock[F]
           )
     } yield ()
 }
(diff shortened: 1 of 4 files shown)


rabbitonweb:
@KentShikama @dzajkowski this is now ready for review

rabbitonweb:
bors try

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -167,14 +167,12 @@ object CasperLaunch {
...
       validatorId <- ValidatorIdentity.fromConfig[F](init.conf)
-      _ <- EngineCell[F].set(
-            new Initializing(
-              init.runtimeManager,
-              init.conf.shardId,
-              validatorId,
-              validators,
-              CommUtil.requestApprovedBlock[F]
-            )
+      _ <- Engine.tranistionToInitializing(
```
KentShikama: transitionToInitializing

lukasz-golebiewski:
Is there a way of preserving this behavior using a test?

lukasz-golebiewski:
Or at least reducing the timeout on an existing one

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
rabbitonweb: Thx for catching this. I have few typos on my todolist, I will fix that one as well.

rabbitonweb:
Thx @lukasz-golebiewski , I will consult @adaszko and @woky about it

PR #2420, opened 2019-05-15
Implement circuit breaker

sebtomba:
## Overview
Implement a circuit breaker with a default message size limit 200MB.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2792

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
@@ -6,8 +6,8 @@ import java.nio.file.{Files, Path}
 import cats.data._
 import cats.implicits._
 
-import coop.rchain.shared.GracefulClose._
 import coop.rchain.shared.{Log, _}
+import coop.rchain.shared.GracefulClose._
 import coop.rchain.shared.PathOps._
 import Compression._
 import coop.rchain.comm.PeerNode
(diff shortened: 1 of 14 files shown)


comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -22,6 +22,10 @@ object StreamHandler {
 
   type CircuitBreaker = Long => Boolean
 
+  sealed trait StreamFault
+  final case class StreamError(message: String)      extends StreamFault
```
rabbitonweb: please change those names, they are meaning less

sebtomba: @rabbitonweb you didn't propose any better names I should use instead ;)

sebtomba: A stream error is when the other side did something wrong. It's a user-side mistake. On the other hand, a stream exception is an unexpected error. So it's a server-side problem. Both errors together form a disjoint stream fault algebra.

itegulov: I agree that the error names make little sense. You can probably take `CommError`'s approach here and split user-level errors (`StreamError`) into different classes (e.g. `WrongNetworkId`, `CircuitWasBroken`). Also, it seems to me that server-side errors mostly come from unexpected IO errors so you could introduce `StreamUnexpectedIOError` or something similar.

PR #2424, opened 2019-05-16
Update RevIssuanceTest to use RevVaults

KentShikama:
## Overview

Assumes RevAddress = hash(EthAddress)

`preWalletUnlockDeploy` which uses the old wallet
system is moved out of RevIssuanceTest. Note we can't
delete it just yet as other tests that still are using
the old wallet system still need to be updated.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3422

### Notes

- Invalid auth key was resolved by changing the implementation of fromPubKey 
- If RevVault isn't created yet, sending to that RevAddress and trying to check balance will silently fail

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
@@ -1,18 +1,21 @@
 package coop.rchain.casper.genesis.contracts
 
-import cats.Traverse
 import cats.effect.Concurrent
 import cats.implicits._
+import com.google.protobuf.ByteString
 import coop.rchain.casper.MultiParentCasperTestUtil.createBonds
+import coop.rchain.casper.SignDeployment
 import coop.rchain.casper.genesis.Genesis
 import coop.rchain.casper.protocol.DeployData
-import coop.rchain.casper.util.rholang.RuntimeManager
+import coop.rchain.casper.util.rholang.{InterpreterUtil, RuntimeManager}
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
 import coop.rchain.casper.util.{BondingUtil, ConstructDeploy}
 import coop.rchain.catscontrib.TaskContrib._
+import coop.rchain.crypto.{PrivateKey, PublicKey}
 import coop.rchain.crypto.codec.Base16
-import coop.rchain.crypto.signatures.Ed25519
 import coop.rchain.metrics.Metrics
+import coop.rchain.crypto.hash.Keccak256
+import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
 import coop.rchain.models.Expr.ExprInstance.GString
 import coop.rchain.models._
 import coop.rchain.rholang.Resources._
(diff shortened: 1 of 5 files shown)


KentShikama:
bors try

ArturGajowy:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -24,13 +26,19 @@ class AddressTools(prefix: Array[Byte], keyLength: Int, checksumLength: Int) {
     * @return None if the key length is invalid or Some if the address was created successfully
     */
   def fromPublicKey(pk: PublicKey): Option[Address] =
-    if (keyLength == pk.bytes.length) {
-      val keyHash = Blake2b256.hash(pk.bytes)
-      val payload = prefix ++ keyHash
-
+    if (keyLength == pk.bytes.length || pk.bytes.length == 65) { // TODO: Clean up for secp256k1
+      val ethAddress = Base16.encode(Keccak256.hash(pk.bytes.drop(1))).takeRight(40)
+      val keyHash    = Keccak256.hash(Base16.unsafeDecode(ethAddress))
```
ArturGajowy: we should use fromEthAddress here

casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
```diff
@@ -45,91 +48,102 @@ class RevIssuanceTest extends FlatSpec with Matchers {
...
+
       val genesisDeploys =
         Genesis.defaultBlessedTerms(
           0L,
           ProofOfStake(1L, Long.MaxValue, posValidators),
-          wallet :: Nil,
+          Nil,
           Faucet.noopFaucet,
           genesisPk,
           vaults,
```
ArturGajowy: The custom `vaults` in here mean that this test is testing its own setup rather than production

We should reimplement this test using RhoSpec, along the lines of:

```scala
class RevIssueanceSpec extends RhoSpec(
      CompiledRholangSource("RevIssueanceTest.rho"),
      TestUtil.defaultGenesis.copy(
        wallets = Seq(("0x041e1eec23d118f0c4ffc814d4f415ac3ef3dcff", 42))
      )
      Seq.empty,
      30.seconds
)

//inside RevIssueanceTest.rho:
 - findOrCreate the vault for eth address above
 - get its balance
 - check the balance is as expected
```

Notice the `Genesis` object passed above does not allow for customisation of adding / replacing the genesis deploys - it only allows for parameterizing them, just like the config files do in production setting.

To be clear:
- RhoSpec constructor would start taking a Genesis parameter
- TestUtil would have the Genesis object that's created in `TestUtil.defaultGenesisSetup` exposed as a method/constant

casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevIssuanceTest.scala
```diff
@@ -29,107 +32,106 @@ class RevIssuanceTest extends FlatSpec with Matchers {
...
+         |        for (@(true, destinationVault) <- destinationCh) {
+         |          @destinationVault!("balance", *balanceCh) |
+         |          for (@balance <- balanceCh) {
+         |            @"$transferStatusOut"!(balance)
+         |          }
+         |        }
+         |      }
+         |    }
+         |  }
+         |}""".stripMargin
```
odeac: In the light of the yesterday discussions about moving non-trivial Rholang code to resource files, I think this qualifies as non-trivial.

I would move create RevIssuanceTest/Transfer.rhox with content:
```

new rl(`rho:registry:lookup`), revVaultCh, vaultCh, revVaultKeyCh, stdout(`rho:io:stdout`), resultCh, destinationCh, balanceCh in {
  rl!(`rho:id:1o93uitkrjfubh43jt19owanuezhntag5wh74c6ur5feuotpi73q8z`, *revVaultCh) |
  for(@(_, RevVault) <- revVaultCh) {
  @RevVault!("findOrCreate", "$$from$$", *vaultCh) |
    @RevVault!("deployerAuthKey", *revVaultKeyCh) |
    for (@(true, vault) <- vaultCh; key <- revVaultKeyCh) {
      @RevVault!("findOrCreate", "${destination.toBase58}", *destinationCh) | // Creates vault in case it doesn't exist
      @vault!("transfer", "$$to$$", $$amount$$, *key, *resultCh) |
      for (@(true, Nil) <- resultCh) {
        for (@(true, destinationVault) <- destinationCh) {
          @destinationVault!("balance", *balanceCh) |
          for (@balance <- balanceCh) {
            @"$$transferStatusOut$$"!(balance)
          }
        }
      }
    }
  }
}

```

...and the code here would look like
```val transferDeployCode = CompiledRholangTemplate("RevIssuanceTest/Transfer.rhox", 
   "from" -> from.toBase58,
   "to" -> destination.toBase58,
   "amount" -> amount,
   "transferStatusOut" -> transferStatusOut
).code```

ArturGajowy: @odeac see my comment above

odeac: I agree that, if this can be written as a RhoSpec test, it would be better.

KentShikama:
Superseded by https://github.com/rchain/rchain/pull/2528

PR #2428, opened 2019-05-17
RCHAIN-3405: Implement an on-chain deployer id

tymm:
## Overview
https://rchain.atlassian.net/browse/RCHAIN-3405


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -52,7 +53,8 @@ trait Reduce[M[_]] {
 
 class DebruijnInterpreter[M[_], F[_]](
     tuplespaceAlg: Tuplespace[M],
-    urnMap: Map[String, Par]
+    urnMap: Map[String, Par],
+    deployParametersRef: Ref[M, DeployParameters]
```
ArturGajowy: I'd love to avoid having yet another ref shared. Ideally, we'd create the Reduce instance just before interpretation, and pass the deploy parameters directly. We should be able to reuse the tuplespace between the interpreters. I'll open a ticket to clean that up.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -305,6 +323,14 @@ object implicits {
     def connectiveUsed(s: Send)          = s.connectiveUsed
     def locallyFree(s: Send, depth: Int) = s.locallyFree
   }
+
+  // TODO
```
ArturGajowy: is it done yet? :)

models/src/test/scala/coop/rchain/models/HashMSpec.scala
```diff
@@ -37,7 +37,9 @@ class HashMSpec extends FlatSpec with PropertyChecks with Matchers {
   sameResultAsReference[ESet]
   sameResultAsReference[ETuple]
   sameResultAsReference[EVar]
+  sameResultAsReference[GUnforgeable]
   sameResultAsReference[GPrivate]
+  sameResultAsReference[GDeployerAuth]
```
ArturGajowy: This one seems redundant, please confirm

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -589,6 +601,15 @@ trait SpatialMatcherInstances {
       } else
         MonoidK[F].empty
 
+  implicit def gdeployerAuthSpatialMatcherInstance[F[_]: Splittable: Alternative: Monad: _error: _cost: _freeMap]
+      : SpatialMatcher[F, GDeployerAuth, GDeployerAuth] =
+    (target, pattern) =>
+      if (target == pattern) {
+        val cost = equalityCheckCost(target, pattern)
+        charge[F](cost) *> ().pure[F]
```
ArturGajowy: Charging should be done regardless of the comparison result.

Do other matching rules repeat this error?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -416,16 +416,34 @@ class DebruijnInterpreter[M[_], F[_]](
...
-          case Some(p) => newEnv.put(p).asRight[ReduceError]
-          case None    => ReduceError(s"Unknown urn for new: $urn").asLeft[Env[Par]]
-        }
+      def addUrn(
+          deployParameters: DeployParameters
+      )(newEnv: Env[Par], urn: String): Either[ReduceError, Env[Par]] =
+        if (urn == "rho:deployer:auth") {
+          val bs = deployParameters.userId.exprs.headOption match {
+            case Some(Expr(GByteArray(byteString))) => byteString
+            case Some(_) | None                     => ByteString.EMPTY
```
ArturGajowy: `case _ =>`

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -416,16 +416,34 @@ class DebruijnInterpreter[M[_], F[_]](
...
+          }
+          Right(
+            newEnv.put(
+              Par(
+                unforgeables = Vector(GDeployerAuth(bs))
+              )
+            )
+          )
```
ArturGajowy: Please join those lines, and use `.asRight` consistently with surrounding code

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: this should be raising an error. We don't want the Auth-s to equal accidentally, and there's no use case for them outside of deploy (where they must be defined)

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
```diff
@@ -0,0 +1,117 @@
...
+    result.size should be(1)
+    result.head should be(GDeployerAuth(pk): Par)
+  }
+
+  it should "make drain vault attacks impossible" in {
+    val skDeployer = PrivateKey(
+      Base16.unsafeDecode("b18e1d0045995ec3d010c387ccfeb984d783af8fbb0f40fa7db126d889f6dadd")
+    )
+    val skAttacker = PrivateKey(
+      Base16.unsafeDecode("c18e1d0045995ec3d010c387ccfeb984d783af8fbb0f40fa7db126d889f6dadd")
```
ArturGajowy: please make those two keys obviously different. E.g. 00000 and 11111

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
```diff
@@ -0,0 +1,117 @@
...
+    )
+    val deploy = ConstructDeploy.sourceDeploy(
+      source =
+        s"""new auth(`rho:deployer:auth`) in { contract @"checkAuth"(input, ret) = { ret!(*input == *auth) }}""",
+      timestamp = System.currentTimeMillis(),
+      accounting.MAX_VALUE,
+      sec = skDeployer
+    )
```
ArturGajowy: Please extract a `deploy(deployer, rho)` private method in this test that hides the irrelevant timestamp and phloLimmimt parameters

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
```diff
@@ -0,0 +1,117 @@
...
+              val hash = result._1
+              mgr
+                .captureResults(
+                  hash,
+                  attack,
+                  captureChannel
+                )
+            }
+        }
+        .runSyncUnsafe(10.seconds)
```
ArturGajowy: Instead of organizing the code by technical aspect (source1, source2, compute1, compute2, assert1, assert2), please reorganize by domain aspect (sourceAttack, computeAttack, assertAttack, similar for non-attack) and then extract the duplication between the two blocks into a method. The whole test should be similar in length to:

```
checkAccessGranted(deployer, deployer, true)
checkAccessGranted(deployer, attacker, false)
```

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
tymm: Yes :)

models/src/test/scala/coop/rchain/models/HashMSpec.scala
tymm: Hmm, why does it seem redundant?

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
tymm: Yes, other matching rules repeat this.

The reasoning might be that checking for equality on two things that are equal is much more expensive than checking for equality on two things that are different.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tymm: How would they equal accidentally? `deployParamteres` is always available in deploys.
If you want to raise an error here, then we have to provide a dummy auth value for the RholangCLI and probably for a bunch of tests.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/util/rholang/DeployerAuthTest.scala
tymm: Done

models/src/test/scala/coop/rchain/models/HashMSpec.scala
ArturGajowy: The test for GUnforgrable should explore these as well. No big deal though :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
tymm: Maybe there should be a constant small cost for the inequality case.

ArturGajowy: That would be a wrong reasoning, as I can maliciously make the compared things differ at the last bits :)

I've reviewed the usages of equalityCheckCost and this pattern might make some sense in cases where we delegate further to `spatialMatch` when two lists of things are equal (see e.g. ReceiveBind or Match). Spatial match is going to do the charging for each of the elements in that case.

It doesn't make sense for ground terms though (and this includes all GUnforgrable-s)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: > How would they equal accidentally

That's the thing with accidents, you never know ;)

> deployParamteres is always available in deploys.

The types don't guard that, so there are chances this will change without anybody noticing - until s/b expliots it.

> If you want to raise an error here, then we have to provide a dummy auth value for the RholangCLI and probably for a bunch of tests.

We don't, because none of them use the "rho:deployer:auth" uri.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
tymm: > That would be a wrong reasoning, as I can maliciously make the compared things differ at the last bits :)
I was under the assumption that this is not possible. How would you do that for `GPrivate` or `GDeployerAuth`?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tymm: Yeah, makes sense.

tymm: Done

tymm:
bors try

tymm:
bors try

KentShikama:
```python
started_standalone_bootstrap_node = <Node(name='rchain-djfsl.bootstrap')>
random_generator = <random.Random object at 0x1b7d598>

    def test_propose_cost(started_standalone_bootstrap_node: Node, random_generator: Random) -> None:
        rho_contract, contract_cost = random_generator.choice(list(FIX_COST_RHO_CONTRACTS.items()))
        shutil.copyfile(os.path.join('resources/cost', rho_contract), os.path.join(started_standalone_bootstrap_node.local_deploy_dir, rho_contract))
        container_contract_file_path = os.path.join(started_standalone_bootstrap_node.remote_deploy_dir, rho_contract)
        started_standalone_bootstrap_node.deploy(container_contract_file_path, DEPLOY_KEY.private_key)
        block_hash = started_standalone_bootstrap_node.propose()
        output = started_standalone_bootstrap_node.show_block(block_hash)
        block_info = parse_show_block_output(output)
        cost = extract_validator_stake_from_deploy_cost_str(block_info['deployCost'])
>       assert contract_cost == cost[DEPLOY_KEY.public_key]
E       assert 1936 == 1970
E         -1936
E         +1970
```
from https://drone.rchain-dev.tk/rchain/rchain/3613/7

Looks like you need to update the integration test to reflect the new costs.

tymm:
bors try

tymm:
@KentShikama Thanks for pointing that out. I was fixated on looking for stale signatures...

tymm:
@KentShikama the signatures had to be changed because `rho:registry:insertSigned` wants a signature over a tuple of a nonce and an unforgeable name wrapped in a write-only bundle. The way unforgeable names are encoded changed (see the `GUnforgeable.writeTo` function) and therefore the signatures had to change.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
ArturGajowy: You're right, that would require rolling two very similar public keys. That's not the real point though:

> Maybe there should be a constant small cost for the inequality case.

It can't be constant. In case of GDeployerAuth and GPrivate both outcomes of the comparison have the same cost lower bound. Granted, there are cases where the comparison short-circuits after 2 or 3 bytes - or even the first one! - but almost none of them do that without incurring any computation. So we can't be saying "if they're not equal, charge nothing".

So we either charge both outcomes equally, or need to integrate cost accounting into the equals method (at which point it shadows the actual computation cost of the equality).

tymm: > So we can't be saying "if they're not equal, charge nothing".

But I still think we could say "if they're not equal, charge a small fixed amount" since on average that small fixed amount would reflect the actual cost (and I don't see how it could be an attack vector).
However I'm perfectly fine with the better-safe-than-sorry option for now :)

ArturGajowy: > since on average that small fixed amount would reflect the actual cost

One of the main points of cost accounting is securing against the ones that play **against** the average. We must employ a pessimistic, upper-bound based mindset in this area. It can't be an attack vector only because of the particular case we have here.

Also, we really should rather make the cost model simple and uniform than try to make it "optimal" in terms of how little we charge. That's not the point, and actually against the interests of validators, and - to a certain degree - against the interests of the community / coop / platform itself.

PR #2435, opened 2019-05-20
Rchain 3430 remove error handler and effect

rabbitonweb:
## Overview
ErrorHandler is mostly not used in comm (or used poorly), but that constraint pushes a lot complexity to other layers that use comm. Once ErrorHandler is removed, a lot of that complexity will be removed

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3430

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
@@ -75,16 +75,21 @@ object HandleMessages {
       _ <- ConnectionsCell[F].flatModify(_.addConnAndReport[F](peer))
     } yield handledWithoutMessage
 
-  def handleProtocolHandshake[F[_]: Monad: TransportLayer: Log: ErrorHandler: ConnectionsCell: RPConfAsk: Metrics](
+  def handleProtocolHandshake[F[_]: Monad: TransportLayer: Log: ConnectionsCell: RPConfAsk: Metrics](
       peer: PeerNode,
       protocolHandshake: ProtocolHandshake
   ): F[CommunicationResponse] =
     for {
       conf     <- RPConfAsk[F].ask
       response = ProtocolHelper.protocolHandshakeResponse(conf.local, conf.networkId)
-      _        <- TransportLayer[F].send(peer, response) >>= ErrorHandler[F].fromEither
-      _        <- Log[F].info(s"Responded to protocol handshake request from $peer")
-      _        <- ConnectionsCell[F].flatModify(_.addConnAndReport[F](peer))
+      resErr   <- TransportLayer[F].send(peer, response)
+      _ <- resErr.fold(
+            kp(().pure[F]),
+            kp(
+              Log[F].info(s"Responded to protocol handshake request from $peer") >>
+                ConnectionsCell[F].flatModify(_.addConnAndReport[F](peer))
+            )
+          )
     } yield handledWithoutMessage
 
   def handleHeartbeat[F[_]: Monad: ConnectionsCell](
--- a/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
+++ b/comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
@@ -22,7 +22,7 @@ object HandleMessages {
   implicit private val metricsSource: Metrics.Source =
     Metrics.Source(CommMetricsSource, "rp.handle")
 
-  def handle[F[_]: Monad: Sync: Log: Time: Metrics: TransportLayer: ErrorHandler: PacketHandler: ConnectionsCell: RPConfAsk](
+  def handle[F[_]: Monad: Sync: Log: Time: Metrics: TransportLayer: PacketHandler: ConnectionsCell: RPConfAsk](
       protocol: Protocol
   ): F[CommunicationResponse] =
     ProtocolHelper.sender(protocol) match {
@@ -31,7 +31,7 @@ object HandleMessages {
       case Some(sender) => handle_[F](protocol, sender)
     }
 
-  private def handle_[F[_]: Monad: Sync: Log: Time: Metrics: TransportLayer: ErrorHandler: PacketHandler: ConnectionsCell: RPConfAsk](
+  private def handle_[F[_]: Monad: Sync: Log: Time: Metrics: TransportLayer: PacketHandler: ConnectionsCell: RPConfAsk](
       proto: Protocol,
       sender: PeerNode
   ): F[CommunicationResponse] =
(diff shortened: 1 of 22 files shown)


lukasz-golebiewski:
Not sure if the last commit is necessary. Keeping the effect aliased would mean that less changes are needed when the time comes to move away from monix Task

comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
```diff
@@ -75,16 +75,21 @@ object HandleMessages {
...
-      _        <- TransportLayer[F].send(peer, response) >>= ErrorHandler[F].fromEither
-      _        <- Log[F].info(s"Responded to protocol handshake request from $peer")
-      _        <- ConnectionsCell[F].flatModify(_.addConnAndReport[F](peer))
+      resErr   <- TransportLayer[F].send(peer, response)
+      _ <- resErr.fold(
+            kp(().pure[F]),
+            kp(
+              Log[F].info(s"Responded to protocol handshake request from $peer") >>
+                ConnectionsCell[F].flatModify(_.addConnAndReport[F](peer))
+            )
```
ArturGajowy: I'm sure we can have it simpler than `.fold(kp(().pure[F]), X)`. I think `.map(X)`?

node/src/main/scala/coop/rchain/node/diagnostics/client/Runtime.scala
```diff
@@ -35,18 +35,6 @@ object Runtime {
...
     } yield ()
 
-    for {
-      result <- program.attempt
-      _ <- result match {
-            case Left(ex) => ConsoleIO[F].println(s"Error: ${processError(ex).getMessage}")
-            case _        => ().pure[F]
-          }
-    } yield ()
-  }
```
ArturGajowy: This seems like an unrelated change for `Remove everywhere ErrorHandler as typeclass constraint`. Is the delete intentional?

node/src/main/scala/coop/rchain/node/NodeEnvironment.scala
```diff
@@ -20,24 +20,24 @@ object NodeEnvironment {
 
   class InitializationException(msg: String) extends RuntimeException
 
-  def create(conf: Configuration)(implicit log: Log[Task]): Effect[NodeIdentifier] =
+  def create(conf: Configuration)(implicit log: Log[Task]): Task[NodeIdentifier] =
     for {
-      dataDir <- Task.delay(conf.server.dataDir.toFile).toEffect
+      dataDir <- Task.delay(conf.server.dataDir.toFile)
```
ArturGajowy: I'd remove the `.toEffect` calls, but maybe leave the `Effect` type alias and its usage? Just in case we'd want to migrate to ZIO or sth...

OTOH ZIO has a Task type of the same shape, so just the imports would change.
If we were to migrate to a bifunctor, there's no big difference between `Task[A] -> ZIO[E, A]` and `Effect[A] -> Effect[E, A]`

WDYT? When's a type alias worth it?

node/src/main/scala/coop/rchain/node/diagnostics/client/Runtime.scala
rabbitonweb: yes, program is task so you no longer call attempt on it. you will not get those `CommErrors` return here.

comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
rabbitonweb: that gived you back an Either[, X] as a result  which is not what we want, right?

rabbitonweb:
@lukasz-golebiewski I thought about it. The issue here is that people were not using `Effect` instantiation properly. If you look at that last what I had to remove, a lot of those changes were things like `leftT` or `rightT`. I also hope that we will always use just `Task` (or equivalent like like `ZIO`). But I'm happy to to revert last commit.

node/src/main/scala/coop/rchain/node/NodeEnvironment.scala
rabbitonweb: see my reply to @lukasz-golebiewski

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala
```diff
@@ -137,27 +137,6 @@ class TransportServer(server: GrpcTransportServer) {
...
-        case Right(m) => Task.now(m)
-      }
-    val hb: Blob => Task[Unit] = b =>
-      handleStreamed(b).value.flatMap {
-        case Left(err) =>
-          Log[Task].error(s"Error while handling streamed Packet message. Error: ${err.message}")
-        case Right(_) => Task.unit
-      }
-    EitherT.liftF(start(dis, hb))
-  }
```
sebtomba: Please add `attemptAndLog` to `GrpcTransportServer.receive.dispatchInternal` in case `handleStreamed()` fails.
```scala
case msg: StreamMessage =>
  StreamHandler.restore(msg) >>= {
    case Left(ex) =>
      Log[Task].error("Could not restore data from file while handling stream", ex)
    case Right(blob) => handleStreamed(blob)
  }
```

node/src/main/scala/coop/rchain/node/NodeEnvironment.scala
rabbitonweb: If we do things right chainging the `IO` type should be matter of modifing module `node` - but prooject is not that state yet (see `comm` queue machanism, see casper tests).

Global rename `s/monix.Task/zio.Task` will be least of our concerns

rabbitonweb:
bors try

comm/src/main/scala/coop/rchain/comm/rp/HandleMessages.scala
ArturGajowy: right, that won't do. Feels awkwardly over-verbose, but having thought more of it, I don't see a way to make it simpler ;/

node/src/main/scala/coop/rchain/node/NodeEnvironment.scala
ArturGajowy: FWIW it's been quite easy to port one of the tests executing end-to-end through rholang :)
https://github.com/rchain/rchain/compare/dev...ArturGajowy:try_zio

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala
rabbitonweb: as discussed in call, this was only handling the `CommError` (not throwables) so nothing was removed really.

PR #2439, opened 2019-05-20
 Logging when queue is full and looping

sebtomba:
## Overview
Log retries of adding messages to the client stream queue every 60 seconds.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3240

### Notes
Can fill logs at a rate of 20x/second when the queue for streaming messages (packets) to other peers is full. The default queue size is 1000. We don't drop messages on the client side. Instead, we block Casper and loop forever until the queue accepts new messages again. We'll now log every 60 retries for each worker. With a delay of 1 second between retries we will log every 60 seconds for each worker.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
@@ -18,6 +18,9 @@ final case class StreamToPeers(peers: Seq[PeerNode], path: Path, sender: PeerNod
 class StreamObservable(bufferSize: Int, folder: Path)(implicit log: Log[Task], scheduler: Scheduler)
     extends Observable[StreamToPeers] {
 
+  val LogRetryEvery: Int         = 60 // With a delay of 1 second log every minute
+  val RetryDelay: FiniteDuration = 1.second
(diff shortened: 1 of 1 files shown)


comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -31,15 +34,20 @@ class StreamObservable(bufferSize: Int, folder: Path)(implicit log: Log[Task], s
...
+    def retry(file: Path, retryCount: Int): Task[Unit] =
       Task
-        .defer(log.warn("Retrying push to client stream") >> propose(file))
-        .delayExecution(1.second)
+        .defer(logRetry(retryCount) >> propose(file, retryCount + 1))
+        .delayExecution(RetryDelay)
+
+    def logRetry(retryCount: Int): Task[Unit] =
+      if (retryCount % LogRetryEvery == 0)
+        log.warn(s"Retrying push to client stream. Retry $retryCount")
```
rabbitonweb: Information does not say anything about queue being full. Maybe it's worth mentioning.

PR #2442, opened 2019-05-20
RCHAIN-872 Move the API that supports the propose call to an internal port

rabbitonweb:
## Overview
Move the API that supports the propose call to an internal port

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-872

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/ProposeService.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/ProposeService.scala
@@ -1,0 +1,50 @@
+package coop.rchain.casper.util.comm
+
+import java.io.Closeable
+import java.util.concurrent.TimeUnit
+
+import scala.util.Either
+
+import coop.rchain.casper.protocol._
+import coop.rchain.models.either.implicits._
+
+import com.google.protobuf.empty.Empty
+import io.grpc.{ManagedChannel, ManagedChannelBuilder}
+import monix.eval.Task
+
+trait ProposeService[F[_]] {
+  def propose(): F[Either[Seq[String], String]]
+}
+
+object ProposeService {
+  def apply[F[_]](implicit ev: ProposeService[F]): ProposeService[F] = ev
+}
+
+class GrpcProposeService(host: String, port: Int, maxMessageSize: Int)
+    extends ProposeService[Task]
+    with Closeable {
+
+  private val channel: ManagedChannel =
+    ManagedChannelBuilder
+      .forAddress(host, port)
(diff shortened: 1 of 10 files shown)


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -70,6 +70,12 @@ object Main {
         conf.grpcServer.portExternal,
         conf.grpcServer.maxMessageSize
       )
+    implicit val proposeService: GrpcProposeService =
+      new GrpcProposeService(
+        conf.grpcServer.host,
+        conf.grpcServer.portInternal,
+        conf.grpcServer.maxMessageSize
+      )
```
itegulov: You also need to close `proposeService` in `program.doOnFinish(`

casper/src/main/scala/coop/rchain/casper/util/comm/ProposeService.scala
```diff
@@ -0,0 +1,50 @@
...
+  private val stub = ProposeServiceGrpcMonix.stub(channel)
+
+  def propose(): Task[Either[Seq[String], String]] =
+    stub.propose(Empty()).map(_.toEither[DeployServiceResponse].map(_.message))
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  override def close(): Unit = {
+    val terminated = channel.shutdown().awaitTermination(10, TimeUnit.SECONDS)
+    if (!terminated) {
+      println(
```
kayvank: curious as to why not use log.warn?

node/src/main/scala/coop/rchain/node/api/package.scala
```diff
@@ -6,7 +6,7 @@ import cats.effect.concurrent.Semaphore
 import coop.rchain.blockstorage.BlockStore
 import coop.rchain.casper.MultiParentCasperRef.MultiParentCasperRef
 import coop.rchain.casper.SafetyOracle
-import coop.rchain.casper.protocol.DeployServiceGrpcMonix
+import coop.rchain.casper.protocol.{DeployServiceGrpcMonix, ProposeServiceGrpcMonix}
 import coop.rchain.catscontrib._
 import coop.rchain.comm.discovery._
```
kayvank: Don't think these imports are required

casper/src/main/scala/coop/rchain/casper/util/comm/ProposeService.scala
rabbitonweb: because it is effectful and we don't control that part (api that sucks)

PR #2444, opened 2019-05-21
RCHAIN-3282 Separate genesis eval from standard deploy eval (part 1)

ArturGajowy:
## Overview

Without replay for now

### JIRA ticket:
RCHAIN-3282

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -43,6 +43,10 @@ trait RuntimeManager[F[_]] {
       terms: Seq[DeployData],
       blockTime: Long
   ): F[(StateHash, Seq[InternalProcessedDeploy])]
+  def computeGenesis(
+      terms: Seq[DeployData],
+      blockTime: Long
+  ): F[(StateHash, StateHash, Seq[InternalProcessedDeploy])]
   def storageRepr(hash: StateHash): F[Option[String]]
   def computeBonds(hash: StateHash): F[Seq[Bond]]
   def computeBalance(start: StateHash)(user: ByteString): F[Long]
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -116,6 +120,19 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+  def computeGenesis(
+      terms: Seq[DeployData],
+      blockTime: Long
+  ): F[(StateHash, StateHash, Seq[InternalProcessedDeploy])] = {
+    val startHash = emptyStateHash
+    withResetRuntime(startHash) { runtime =>
+      for {
+        _          <- setBlockTime(blockTime, runtime)
+        evalResult <- newEval(terms, runtime, startHash)
+      } yield (startHash, evalResult._1, evalResult._2)
```
JosDenmark: The fact that you're returning `startHash` means that (at some point) you'd like the flexibility to have `startHash != emptyStateHash`?

ArturGajowy: Quite the contrary: notice `computeState` took `startHash` as parameter - which brings the flexibility you mention. Since there must be no flexibility on the start hash for genesis, and the caller might not necessarily know that, they have no way of passing it in, and are prompted to use the right value by it being returned :)

JosDenmark: Hmm. I still don't understand. Under what conditions would we call `computeGenesisState` with `startHash != emptyStateHash`? I can't think of any, and if `startHash` isn't a parameter, then it doesn't have to be a return value, right?

ArturGajowy: Right. Though making it a return value allows the user to not ask for the startHash (emptyRootHash) separately - and they need it to create the resulting InternalProcessedDeploy.

PR #2447, opened 2019-05-21
Event Log for crucial events

sebtomba:
## Overview
Introduce an Event Log by writing the events to a separate Logback logger
`coop.rchain.shared.EventLogger`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3056

### Notes
The `EventLogger` is disabled by default in the `logback.xml` file (node project).
Currently, all events are simple strings. We can augment the events and serialize them f.e. to Json in the next step.

For the integration tests, the mapping of logs to events:

Log | Event
---- | ----- 
coop.rchain.node.NodeRuntime - Listening for traffic on rnode | `NodeStarted`
Making a transition to Running state. | `EnteredRunningState`
Valid ApprovedBlock received! | `ApprovedBlockReceived`
c.r.c.u.c.ApproveBlockProtocol$ApproveBlockProtocolImpl - APPROVAL: Sent UnapprovedBlock  | `SentUnapprovedBlock`
c.r.c.u.c.ApproveBlockProtocol$ApproveBlockProtocolImpl - APPROVAL: received block approval from | `BlockApprovalReceived`
c.r.c.u.c.ApproveBlockProtocol$ApproveBlockProtocolImpl - APPROVAL: Sent ApprovedBlock | `SentApprovedBlock`

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
@@ -140,7 +140,14 @@ object ApproveBlockProtocol {
                 Metrics[F].incrementCounter("genesis")
               else ().pure[F]
           _ <- Log[F].info(s"APPROVAL: received block approval from $sender")
-          _ <- EventLog[F].publish(shared.Event.BlockApprovalReceived)
+          _ <- EventLog[F].publish(
+                shared.Event.BlockApprovalReceived(
+                  a.candidate
+                    .flatMap(_.block.map(b => PrettyPrinter.buildStringNoLimit(b.blockHash)))
+                    .getOrElse(""),
+                  sender
+                )
+              )
         } yield (),
         Log[F].warn(s"APPROVAL: ignoring invalid block approval from $sender")
       )
@@ -169,7 +176,7 @@ object ApproveBlockProtocol {
         _ <- Log[F].info(s"APPROVAL: Beginning send of UnapprovedBlock $candidateHash to peers...")
         _ <- CommUtil.streamToPeers[F](transport.UnapprovedBlock, serializedUnapprovedBlock)
         _ <- Log[F].info(s"APPROVAL: Sent UnapprovedBlock $candidateHash to peers.")
-        _ <- EventLog[F].publish(shared.Event.SentUnapprovedBlock)
+        _ <- EventLog[F].publish(shared.Event.SentUnapprovedBlock(candidateHash))
       } yield ()
 
     private def completeIf(time: Long, signatures: Set[Signature]): F[Unit] =
@@ -194,7 +201,7 @@ object ApproveBlockProtocol {
                       )
                   _ <- CommUtil.streamToPeers[F](transport.ApprovedBlock, serializedApprovedBlock)
                   _ <- Log[F].info(s"APPROVAL: Sent ApprovedBlock $candidateHash to peers.")
-                  _ <- EventLog[F].publish(shared.Event.SentApprovedBlock)
+                  _ <- EventLog[F].publish(shared.Event.SentApprovedBlock(candidateHash))
                 } yield ()
             }
       } yield ()
--- a/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
@@ -20,6 +20,7 @@ import coop.rchain.comm.transport.TransportLayer
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b256
 import coop.rchain.metrics.Metrics
+import coop.rchain.shared
 import coop.rchain.shared._
 
 import com.google.protobuf.ByteString
@@ -51,7 +52,7 @@ object ApproveBlockProtocol {
   def apply[F[_]](implicit instance: ApproveBlockProtocol[F]): ApproveBlockProtocol[F] = instance
 
   //For usage in tests only
-  def unsafe[F[_]: Sync: ConnectionsCell: TransportLayer: Log: Time: Metrics: RPConfAsk: LastApprovedBlock](
+  def unsafe[F[_]: Sync: ConnectionsCell: TransportLayer: Log: EventLog: Time: Metrics: RPConfAsk: LastApprovedBlock](
       genesisBlock: BlockMessage,
       trustedValidators: Set[ByteString],
       requiredSigs: Int,
@@ -70,7 +71,7 @@ object ApproveBlockProtocol {
       sigsF
     )
 
-  def of[F[_]: Sync: ConnectionsCell: TransportLayer: Log: Time: Metrics: RPConfAsk: LastApprovedBlock](
+  def of[F[_]: Sync: ConnectionsCell: TransportLayer: Log: EventLog: Time: Metrics: RPConfAsk: LastApprovedBlock](
       genesisBlock: BlockMessage,
       trustedValidators: Set[ByteString],
       requiredSigs: Int,
@@ -90,7 +91,7 @@ object ApproveBlockProtocol {
       sigsF
     )
 
-  private class ApproveBlockProtocolImpl[F[_]: Sync: ConnectionsCell: TransportLayer: Log: Time: Metrics: RPConfAsk: LastApprovedBlock](
+  private class ApproveBlockProtocolImpl[F[_]: Sync: ConnectionsCell: TransportLayer: Log: EventLog: Time: Metrics: RPConfAsk: LastApprovedBlock](
       val genesisBlock: BlockMessage,
       val requiredSigs: Int,
       val trustedValidators: Set[ByteString],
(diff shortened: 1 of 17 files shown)


adaszko:
bors try

shared/src/main/scala/coop/rchain/shared/EventLog.scala
```diff
@@ -0,0 +1,26 @@
+package coop.rchain.shared
+
+sealed trait Event
+object Event {
+  object NodeStarted           extends Event
+  object EnteredRunningState   extends Event
+  object ApprovedBlockReceived extends Event
+  object SentUnapprovedBlock   extends Event
+  object BlockApprovalReceived extends Event
+  object SentApprovedBlock     extends Event
```
ArturGajowy: I'd say those should be case classes parameterized with the data allowing for discerning the event instances, and doing debugging.

E.g. any Block-related event should contain its hash
any receive-related eevent should contain the sender

ArturGajowy: I'm also missing a 'Block received' event in here (the above events only cover genesis).

Also, an event should most likely be emitted whenever `.propose` is triggered on a node

casper/src/test/scala/coop/rchain/casper/engine/ApproveBlockProtocolTest.scala
```diff
@@ -210,9 +209,10 @@ class ApproveBlockProtocolTest extends FlatSpec with Matchers {
   }
 
   it should "not accept BlockApproved messages signed by not trusted validators" in {
-    implicit val ctx         = TestScheduler()
-    implicit val logStub     = new LogStub[Task]()
-    implicit val metricsTest = new MetricsTestImpl[Task]()
+    implicit val ctx          = TestScheduler()
+    implicit val logStub      = new LogStub[Task]()
+    implicit val eventLogStub = new EventLogStub[Task]()
+    implicit val metricsTest  = new MetricsTestImpl[Task]()
```
ArturGajowy: Why don't we move all those repeated implicit vals to the test's fields?

shared/src/main/scala/coop/rchain/shared/EventLog.scala
adaszko: > E.g. any Block-related event should contain its hash
any receive-related eevent should contain the sender

By event you mean a Scala representation of an event or the thing that gets emitted on a logger?

ArturGajowy: I'd go with the former - the data in the events might be useful for processing otherwise than dumping it into the logs. It also seems prudent to defer string serialization as long as possible.

ArturGajowy: > useful for processing otherwise
not that we're going to store that data now - but having the API be data-bound and not string-bound will make things easier if/when we start storing the data / outputting json

sebtomba: @ArturGajowy I augmented the events with the block hash. We can add other events later. For now, this will be used only in the integration tests.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -284,8 +285,9 @@ class NodeRuntime private[node] (
             pm => HandleMessages.handle[Task](pm),
             blob => packetHandler.handlePacket(blob.sender, blob.packet)
           )
-      address = s"rnode://$id@$host?protocol=$port&discovery=$kademliaPort"
+      address = local.toAddress
       _       <- Log[Task].info(s"Listening for traffic on $address.")
+      _       <- EventLog[Task].publish(Event.NodeStarted(address))
```
ArturGajowy: It just occurred to me: where are those events used? I think the PR introducing the EventLog should provide its first usage as well. Otherwise we might be very wrong about what's **atually** needed.

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
```diff
@@ -139,6 +140,14 @@ object ApproveBlockProtocol {
                 Metrics[F].incrementCounter("genesis")
               else ().pure[F]
           _ <- Log[F].info(s"APPROVAL: received block approval from $sender")
+          _ <- EventLog[F].publish(
+                shared.Event.BlockApprovalReceived(
+                  a.candidate
+                    .flatMap(_.block.map(b => PrettyPrinter.buildStringNoLimit(b.blockHash)))
+                    .getOrElse(""),
```
ArturGajowy: That's a lot of ugly code due to protobufs and their optionality. Let's extract that 3-times duplicated chunk into a util at least? Also, most likely "<NONE>" or "<MISSING>" would be better than "".

Moreover, I'll stand by my previous claim that the events should contain raw data rathter than strings, and the toString should happen at the usage site.

shared/src/main/scala/coop/rchain/shared/EventLog.scala
```diff
@@ -0,0 +1,26 @@
...
+  def publish(event: Event): F[Unit]
+}
+
+object EventLog {
+  def apply[F[_]](implicit E: EventLog[F]): EventLog[F] = E
+
+  implicit val EventLoggerLogSource: LogSource = LogSource(classOf[EventLogger])
+
+  def eventLogger[F[_]: Log]: EventLog[F] =
+    (event: Event) => Log[F].info(event.getClass.getSimpleName.stripSuffix("$"))
```
ArturGajowy: oh it is used, just generically! We should probably do a pattern match in here, without a 'catch-all' case so that we have the compiler yell at us on new unsupported event types.

Also, we should be leveraging the fields you've just added

adaszko:
@ArturGajowy Are your remarks absolutely necessary or are there something we can iterate on later?  I'd benefit from having this PR merged ASAP.

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
sebtomba: @ArturGajowy I see your point. However, I'm not convinced that events should carry huge balls of raw data. In my opinion, events should carry as little data as possible and as much data as required. I don't like the idea that we should use the protobuf auto-generated classes in our events. On the other hand, the processing of these classes requires domain logic (from `casper`) which shouldn't leak into the `EventLog` instance somewhere else. I admit, that just using `toString` wouldn't require any domain logic. But that is an awful serialization strategy.

shared/src/main/scala/coop/rchain/shared/EventLog.scala
sebtomba: What format do you propose? If it's Json then we need some further specification, f.e.
```json
{
  "name": "BlockApprovalReceived",
  "ts": 12345,
  "body": {
    "block-hash": "abb989aabcc"
  }
}
```

casper/src/test/scala/coop/rchain/casper/engine/ApproveBlockProtocolTest.scala
sebtomba: AFAIK the `*Stub`s are used locally to accumulate f.e. logs and verify them in the particular test. ATM I didn't verify the collected events in the `EventLogStub`. A refactoring of Casper unit test is not in the scope of this PR

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
ArturGajowy: > as much data as required

Yes, but not in string form. `base64(byteArray)` is bigger than `byteArray`, not to mention reference sharing.

> I don't like the idea that we should use the protobuf auto-generated classes in our events. 

I'm not arguing for that, I'm simply against strings :) In that particular case I'd use an Option[Array[Byte]] to represent the hash.

> toString wouldn't require any domain logic. But that is an awful serialization strategy.

I concur :smile: 

Now that I look at it again, I'm a bit partial on all the events being defined in the `EventLog` file. Suddenly it's the case that this is a common dependency of all modules that need to publish events. On one hand it's nice to have them all in one place, OTOH, morally, each of the events is part of the modules having the domain for them.

Anyway, no matter where we put the events, if we keep them at 'small case class containing just the data they need, either using primitives or custom-event-tree-case-classes using the former', we can probably have generic serialization to whatever format we invent, including JSON.

shared/src/main/scala/coop/rchain/shared/EventLog.scala
ArturGajowy: I'd say the specification should be 'dump the event case class structure into JSON generically, that's the JSON the output specification defines' :joy: 

If we're not happy with the JSON structure thus achieved, we should probably not be happy with the event case classes either :)

casper/src/test/scala/coop/rchain/casper/engine/ApproveBlockProtocolTest.scala
ArturGajowy: If they need to be local, then they can be re-declared locally. In all ohter cases they should use the same subs, noop-stubs.

> A refactoring of Casper unit test is not in the scope of this PR

Honestly, whenever a change requires changing a dozen of places in one file, I'd say we should require first doing the refactoring that reduces the number of places to one. Especially if it's as simple as in this case.

Even more: you could've declared that new implicit once, globally.

ArturGajowy: Notice I approved the PR, I'm just making a case for a better future :)

PR #2453, opened 2019-05-23
Rholang Error Handling Pt. 1

JosDenmark:
## Overview

Pay no mind to errors thrown in the `Reduce` tests. They're easily fixed.

Starting with the last commit, an error arises in `InterpreterSpec` when all of the Rholang tests are run using sbt. The error doesn't arise when the Rholang tests are run using IntelliJ, and the error doesn't arise when only `InterpreterSpec` is run in sbt.

In `InterpreterSpec` the error arises because a produce (`@1!(1)`) occurs between the reset and the next checkpoint, which is induced by the error `@2!(3.noSuchMethod())`. The question is, why is `@1!(1)` executing after the reset? It should be the case that all events are logged and completed before the reset. 

Given `@1!(1) | @2!(3.noSuchMethod())`, there are two possible evaluation orders:

1) If `@1!(1)` finishes first, it should add `1` to the `@1`, then trigger the exception for `@2!(3.noSuchMethod())`.

2) If `@2!(3.noSuchMethod())` finishes first, it should trigger the exception and subsequently cancel `@1!(1)`.

Finally, it seems that `StackSafetySpec` contributes to the test failing. Enabling `logback` in the rholang tests (even logging to a file) causes `StackSafetySpec` to fail from a stack overflow error, which affects the result of the `InterpreterSpec`. If one could find a way to test `StackSafetySpec` and `InterpreterSpec`, but only log during `InterpreterSpec`, we'd have better insight into the behavior of RSpace during the failure.

### JIRA ticket:
[https://rchain.atlassian.net/browse/RCHAIN-3395](url)
[https://rchain.atlassian.net/browse/RCHAIN-3394](url)

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1,12 +1,10 @@
 package coop.rchain.rholang.interpreter
 
-import cats._
 import cats.effect.Sync
 import cats.implicits._
 import cats.mtl.FunctorTell
 import cats.{Applicative, FlatMap, Foldable, Parallel, Eval => _}
 import com.google.protobuf.ByteString
-import coop.rchain.catscontrib.mtl._
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.models.Expr.ExprInstance._
@@ -15,15 +13,13 @@ import coop.rchain.models.Var.VarInstance.{BoundVar, FreeVar, Wildcard}
 import coop.rchain.models.rholang.implicits._
 import coop.rchain.models.serialization.implicits._
 import coop.rchain.models.{Match, MatchCase, _}
-import coop.rchain.rholang.interpreter._
 import coop.rchain.rholang.interpreter.Substitute.{charge => _, _}
 import coop.rchain.rholang.interpreter.accounting._
 import coop.rchain.rholang.interpreter.errors._
-import coop.rchain.rholang.interpreter.matcher._
+import coop.rchain.rholang.interpreter.matcher.SpatialMatcher.spatialMatchAndCharge
 import coop.rchain.rholang.interpreter.storage.Tuplespace
 import coop.rchain.rspace.Serialize
 import monix.eval.Coeval
-import SpatialMatcher.spatialMatchAndCharge
 
 import scala.collection.immutable.BitSet
 import scala.util.Try
@@ -261,6 +257,7 @@ object Reduce {
       * 3. Evaluate any top level expressions in the data being sent.
       * 4. Call produce
       * 5. If produce returned a continuation, evaluate it.
+      *
       * @param send An output process
       * @param env An execution context
       * @return
@@ -276,7 +273,7 @@ object Reduce {
         unbundled <- subChan.singleBundle() match {
                       case Some(value) =>
                         if (!value.writeFlag) {
-                          s.raiseError(ReduceError("Trying to send on non-writeable channel."))
+                          s.raiseError[Par](ReduceError("Trying to send on non-writeable channel."))
                         } else {
                           s.pure(value.body)
                         }
@@ -487,7 +484,7 @@ object Reduce {
         unbndl <- subst.singleBundle() match {
                    case Some(value) =>
                      if (!value.readFlag) {
-                       s.raiseError(ReduceError("Trying to read from non-readable channel."))
+                       s.raiseError[Par](ReduceError("Trying to read from non-readable channel."))
                      } else {
                        s.pure(value.body)
                      }
@@ -528,7 +525,7 @@ object Reduce {
             evaledArgs   <- arguments.toList.traverse(expr => evalExpr(expr))
             resultPar <- methodLookup match {
                           case None =>
-                            s.raiseError(ReduceError("Unimplemented method: " + method))
+                            s.raiseError[Par](ReduceError("Unimplemented method: " + method))
                           case Some(f) => f(evaledTarget, evaledArgs)
                         }
           } yield resultPar
@@ -557,7 +554,7 @@ object Reduce {
                      case (GString(s1), GString(s2)) =>
                        Applicative[M].pure(GBool(relops(s1, s2)))
                      case _ =>
-                       s.raiseError(ReduceError("Unexpected compare: " + v1 + " vs. " + v2))
+                       s.raiseError[GBool](ReduceError("Unexpected compare: " + v1 + " vs. " + v2))
                    }
         } yield result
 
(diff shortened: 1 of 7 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -15,20 +12,18 @@ import coop.rchain.models.Var.VarInstance.{BoundVar, FreeVar, Wildcard}
...
 import coop.rchain.rholang.interpreter.storage.Tuplespace
 import coop.rchain.rspace.Serialize
 import monix.eval.Coeval
-import SpatialMatcher.spatialMatchAndCharge
 
 import scala.collection.immutable.BitSet
 import scala.util.Try
 
-// Notes: Caution, a type annotation is often needed for Env.
+// TODO: Are out of phlo errors the only fatal errors?
```
ArturGajowy: no, all errors are fatal (if you ask rholang, at least :))

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -56,973 +51,869 @@ trait Reduce[M[_]] {
...
-                case (k, _) => current.startsWith("${" + k + "}")
-              } match {
-                case Some((k, v)) =>
-                  result ++= v
-                  current = current.drop(k.length + 3)
-                case None =>
-                  result += current.head
-                  current = current.tail
-              }
+      case ELtBody(ELt(p1, p2)) => // TODO: Replace <* with <<
```
ArturGajowy: I'm a bit partial on your leaving of TODO comments in a commit that does an unrelated thing. OFC, TODO are always unrelated to the thing you work on when you spot them, but it hurts even more when you leave the same TODO multiple times, and then get rid of all (but one) occurrences in the next commits.

At the same time, I see the workflow that emerges from such practice and am tempted to employ it myself. Can we though agree that - when there's multiple occurences of the same TODO, and the occurrences can be searched moderately easily - we don't include more than one copy of a TODO?

ArturGajowy:
To your last commit - It's also very neat (when we ignore the printlns), and doesn't do anything suspicious TBH. What I'd do is clean the commit up, and proceed with obliteration of the ErrorLog until the compiler / context of removal tells me sth. Then, most likely, anything that needs to change will need to change in a separate commit that will be put before `Remove ErrorLog from Reduce.`

ArturGajowy:
I haven't stressed this enough: great job on this @JosDenmark! :smiley:

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosDenmark: Yes. We can agree :)

JosDenmark:
bors try

JosDenmark:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -52,7 +52,7 @@ trait Reduce[M[_]] {
 
 class DebruijnInterpreter[M[_], F[_]](
     tuplespaceAlg: Tuplespace[M],
-    private val urnMap: Map[String, Par]
+    urnMap: Map[String, Par]
```
lukasz-golebiewski: What's the reason for relaxing the access constraints to `urnMap`?

rholang/src/test/scala/coop/rchain/rholang/Resources.scala
```diff
@@ -71,18 +66,10 @@ object Resources {
...
-            runtime <- {
-              implicit val c = cost
-              Runtime.create[F, M](tmpDir, storageSize, StoreType.LMDB, additionalSystemProcesses)
-            }
-          } yield (runtime)
-        }(
-          rt => rt.close()
-        )
-      }
+    mkTempDir[F](prefix).flatMap { tmpDir =>
```
lukasz-golebiewski: Nice one!

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -467,7 +467,8 @@ object Runtime {
     def checkCreateDataDir: F[Unit] =
       for {
         notexists <- Sync[F].delay(Files.notExists(dataDir))
-        _         <- if (notexists) Sync[F].delay(Files.createDirectories(dataDir)) else ().pure[F]
+        _ <- if (notexists) Sync[F].delay(Files.createDirectories(dataDir)) >> ().pure[F]
```
lukasz-golebiewski: What's the point of adding `>> ().pure[F]` here?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -276,7 +273,7 @@ object Reduce {
         unbundled <- subChan.singleBundle() match {
                       case Some(value) =>
                         if (!value.writeFlag) {
-                          s.raiseError(ReduceError("Trying to send on non-writeable channel."))
+                          s.raiseError[Par](ReduceError("Trying to send on non-writeable channel."))
```
lukasz-golebiewski: I'm failing to see the benefit of removing `import cats._` in favor of specifying the types explicitly here

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosDenmark: The `private val` modifier is redundant because constructor arguments aren't accessible in class instances without a `val` modifier.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
JosDenmark: So that the return type of the statement is consistent and the editor will stop screaming. Prior to the change, the affirmative case returned `F[Path]` while the negative case returned `F[Unit]`. I'm actually not sure why it compiled in the first place.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosDenmark: In a subsequent commit, I convert these to direct method calls on the errors. In that commit, type inference fails without the type annotations even if `cats._` is imported.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
ArturGajowy: This could be `Sync[F].delay(Files.createDirectories(dataDir)).as(())`, but NBD :)

ArturGajowy:
Guess I'll trigger a merge so that we don't get conflicts by chance...

bors r+

PR #2455, opened 2019-05-23
Make validator private key an encrypted file instead of CLI option

itegulov:
## Overview
- Hides the command line option `rnode run --validator-private-key`
- The file with the validator private key should be encrypted. The node operator must provide the password in the environment variable `RNODE_VALIDATOR_PASSWORD`
- The node fails if the PK file doesn't exist, the file is not encrypted, the env var is not set or the file can't be decrypted.

Client side
- When doing `rnode keygen`, asks the user for a password and encrypts the private key with the password
- The deployers private key must be stored in an encrypted file
- Hides `rnode deploy --private-key`
- Adds a command line option `rnode deploy --private-key-path` for the encrypted file with the private key
- When a user does `rnode deploy` asks him for the password (readline) and decrypts the file which contains the private key
- Alternatively the deployer can provide the password in the environment variable RNODE_DEPLOYER_PASSWORD

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-1608

### Notes

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -68,15 +68,44 @@ object Main {
     val program = conf.command match {
       case Eval(files) => new ReplRuntime().evalProgram[Task](files)
       case Repl        => new ReplRuntime().replProgram[Task].as(())
-      case Deploy(phlo, phloPrice, validAfterBlock, privateKey, location) =>
-        DeployRuntime
-          .deployFileProgram[Task](
-            phlo,
-            phloPrice,
-            validAfterBlock,
-            privateKey,
-            location
-          )
+      case Deploy(
+          phlo,
+          phloPrice,
+          validAfterBlock,
+          maybePrivateKey,
+          maybePrivateKeyPath,
+          location
+          ) =>
+        (maybePrivateKey, maybePrivateKeyPath) match {
+          case (Some(privateKey), _) =>
+            DeployRuntime.deployFileProgram[Task](
+              phlo,
+              phloPrice,
+              validAfterBlock,
+              Some(privateKey),
+              location
+            )
+          case (None, Some(privateKeyPath)) =>
+            for {
+              password   <- ConsoleIO[Task].readPassword("Password for private key file: ")
+              privateKey <- Secp256k1.parsePemFile[Task](privateKeyPath, password)
+              _ <- DeployRuntime.deployFileProgram[Task](
+                    phlo,
+                    phloPrice,
+                    validAfterBlock,
+                    Some(privateKey),
+                    location
+                  )
+            } yield ()
+          case (None, None) =>
+            DeployRuntime.deployFileProgram[Task](
+              phlo,
+              phloPrice,
+              validAfterBlock,
+              None,
+              location
+            )
+        }
       case FindDeploy(deployId) => DeployRuntime.findDeploy[Task](deployId)
       case Propose              => DeployRuntime.propose[Task]()
       case ShowBlock(hash)      => DeployRuntime.getBlock[Task](hash)
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -9,6 +9,7 @@ import cats.implicits._
 import coop.rchain.casper.util.comm._
 import coop.rchain.casper.util.BondingUtil
 import coop.rchain.catscontrib.TaskContrib._
+import coop.rchain.crypto.PrivateKey
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.signatures.{Ed25519, Secp256k1}
 import coop.rchain.metrics
@@ -24,6 +25,7 @@ import org.slf4j.LoggerFactory
 import org.slf4j.bridge.SLF4JBridgeHandler
 
 object Main {
+  private val RNodeDeployerPasswordEnvVar = "RNODE_DEPLOYER_PASSWORD"
 
   implicit private val logSource: LogSource     = LogSource(this.getClass)
   implicit private val log: Log[Task]           = effects.log
@@ -65,6 +67,8 @@ object Main {
 
     implicit val time: Time[Task] = effects.time
 
+    implicit val envVars: EnvVars[Task] = EnvVars.envVars
+
     val program = conf.command match {
       case Eval(files) => new ReplRuntime().evalProgram[Task](files)
       case Repl        => new ReplRuntime().replProgram[Task].as(())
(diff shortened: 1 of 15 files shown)


itegulov:
@ArturGajowy you can't provide an empty password to create an unencrypted private key. There is no point in generating an unencrypted private key since, starting from this PR, the node will not work with unencrypted private keys (i.e. passing an unencrypted private key will raise an error).

itegulov:
@ArturGajowy this is mentioned by @sebtomba in one of the comments under the Jira issue if that makes a difference for you.

I am open to discussion but you need to elaborate a little on your opinion here. Notice that you **don't have** to type in your password every time you are making a deploy. You can set the environment variables (`RNODE_VALIDATOR_PASSWORD` and `RNODE_DEPLOYER_PASSWORD`) once and use rnode without having to type anything.

ArturGajowy:
That being said, I anticipate a lot of users churning at this. The no-encryption option should be available at least for the deployers. I'm speaking from a developer's perspective, and I say that again: literally never did I encrypt my private key if I weren't forced to. I do that choice consciously, because the inconvenience of passing the password / env variable on every command execution is not worth it imo.

Maybe if there's an easy way of integrating a password manager with environment variables, and we documented that - maybe then that would be bearable.

odeac:
> You can set the environment variables (RNODE_VALIDATOR_PASSWORD and RNODE_DEPLOYER_PASSWORD) once and use rnode without having to type anything.
I assume the validator admin wants to protect the private key with a password in order to add another level of security such that a process that runs with its user privileges is not able to read the private key.

If we use environment variables to pass the private key password to rnode, this defeats the purpose of having a encrypted key because any process running with current user privileges will be able to read that environment variable by looking at /proc/<pid>/environ

cboscolo:
We need to wrap up this work. Can we get a second approval on this? @odeac

Regarding the ENV approach, it is agreed that it has some security concerns, but for some cloud environments is sufficient. I will ticket adding integration for cloud HSM/KMS integration.

Regarding allowing the use of rnode without needing a password, `--validator-private-key` will remain as a hidden option.

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -65,28 +80,73 @@ object Main {
...
-          )
+      case Deploy(
+          phlo,
+          phloPrice,
+          validAfterBlock,
+          maybePrivateKey,
+          maybePrivateKeyPath,
+          location
+          ) =>
+        (maybePrivateKey, maybePrivateKeyPath) match {
```
odeac: This can be rewritten without code duplication and quite readable, I would say:

```scala
        def decryptPrivateKey(encryptedPrivateKeyPath : Path) =
          for {
            maybePassword <- EnvVars[Task].get(RNodeDeployerPasswordEnvVar)
            password <- maybePassword.map(p => p.pure[Task])
              .getOrElse(ConsoleIO[Task].readPassword("Password for private key file: "))

            privateKey <- Secp256k1.parsePemFile[Task](encryptedPrivateKeyPath, password)
          } yield privateKey

        val getPrivateKey =
          maybePrivateKey.map(_.pure[Task])
            .combineK(maybePrivateKeyPath.map(decryptPrivateKey))
            .sequence

        for {
          privateKey <- getPrivateKey
          _ <- DeployRuntime.deployFileProgram[Task](
            phlo,
            phloPrice,
            validAfterBlock,
            privateKey,
            location
          )
        } yield()
```

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -106,18 +166,68 @@ object Main {
...
+               val provider         = new BouncyCastleProvider()
+               val keyPairGenerator = KeyPairGenerator.getInstance("ECDSA", provider)
+               keyPairGenerator.initialize(
+                 new ECGenParameterSpec("secp256k1"),
+                 SecureRandomUtil.secureRandomNonBlocking
+               )
+               val kp = keyPairGenerator.generateKeyPair
+               kp.getPrivate.pure[Task]
+             case _ => Task.raiseError(new IllegalStateException("Invalid algorithm name"))
+           }
```
odeac: For readability reasons this can be extracted in a `generateSk` function

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -106,18 +166,68 @@ object Main {
...
+      password       <- ConsoleIO[Task].readPassword("Password for generated private key file: ")
+      passwordRepeat <- ConsoleIO[Task].readPassword("Repeat password: ")
+      _ <- if (password != passwordRepeat) {
+            ConsoleIO[Task].println("Passwords do not match. Try again.") >> generateKey(
```
odeac: you don't need to re-generate the key here. You only have to re-read the password

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -106,18 +166,68 @@ object Main {
...
+              conf,
+              algorithm,
+              privateKeyPath
+            )
+          } else if (password.isEmpty) {
+            ConsoleIO[Task].println("Password is empty. Try again.") >> generateKey(
```
odeac: same here

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -106,18 +166,68 @@ object Main {
...
+            Resource
+              .make(
+                Task.delay(new JcaPEMWriter(new FileWriter(privateKeyPath.toFile)))
+              )(writer => Task.delay(writer.close()))
+              .use { writer =>
+                Task.delay(writer.writeObject(pemObj))
+              } >> ConsoleIO[Task].println(
+              s"Successfully generated private key: ${privateKeyPath.toAbsolutePath}"
+            )
+          }
```
odeac: I would extract this block into a named function for readability reasons

itegulov:
Thanks for your review @odeac! PTAL.

itegulov:
bors try

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -1,29 +1,51 @@
...
+import org.bouncycastle.asn1.x509.AlgorithmIdentifier
+import org.bouncycastle.asn1.x9.X9ObjectIdentifiers
+import org.bouncycastle.crypto.ec.CustomNamedCurves
+import org.bouncycastle.crypto.util.{PrivateKeyFactory, PrivateKeyInfoFactory}
+import org.bouncycastle.jcajce.provider.asymmetric.util.ECUtil
+import org.bouncycastle.jce.ECNamedCurveTable
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import org.bouncycastle.jce.spec.{ECParameterSpec, ECPrivateKeySpec}
+import org.bouncycastle.openssl.jcajce.{JcaMiscPEMGenerator, JcaPEMWriter, JcePEMEncryptorBuilder}
+import org.bouncycastle.util.io.pem.PemObject
```
sebtomba: @itegulov All these new imports show that `Main` does things out of the scope of launching the application. Please move all the encryption related logic into the `crypto` sub-project.

itegulov: Good point

PR #2456, opened 2019-05-23
Check on node startup if all ports are free and exit if they are not

sebtomba:
## Overview
Check on node startup if all ports are free and exit if they are not

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-732

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -126,11 +126,34 @@ object Main {
     SLF4JBridgeHandler.removeHandlersForRootLogger()
     SLF4JBridgeHandler.install()
     for {
+      _       <- checkPorts(conf)
       _       <- log.info(VersionInfo.get)
       _       <- logConfiguration(conf)
       runtime <- NodeRuntime(conf)
       _       <- runtime.main
     } yield ()
+  }
+
+  private def checkPorts(conf: Configuration): Task[Unit] = {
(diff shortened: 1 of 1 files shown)


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -121,13 +119,36 @@ object Main {
...
 
+  private def checkPorts(conf: Configuration): Task[Unit] = {
+    def isLocalPortAvailable(port: Int): Task[Boolean] =
+      Task
+        .delay(new java.net.ServerSocket(port).close())
+        .attempt
+        .map(_.isRight)
+        .ifM(
+          true.pure[Task],
+          Task.delay(System.err.println(s"Port $port is already in use!")).as(false)
```
itegulov: Why use `System.err.println` instead of `Log[F].error`?

sebtomba: @adaszko what do you prefer? On startup of the node, if something is wrong with the configuration we print a message to the `err` out and exit. Now we do the same with the ports. If one of the ports is not free we (would) print a message to `err` out and exit.

adaszko: Actually, it is the process exit code that is most important in this case.  It should be nonzero.

Assuming the exit code is nonzero, I'd tilt slightly towards `Log[F].error`, just for consistency (i.e. not to multiply output channels unnecessarily).

PR #2461, opened 2019-05-27
Charging for deploy prerequisites part 3

ArturGajowy:
## Overview

Mostly refactorings, all needed for https://rchain.atlassian.net/browse/RCHAIN-3281

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3281

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
@@ -65,7 +65,7 @@ object InterpreterUtil {
   ): F[Either[BlockException, Option[StateHash]]] =
     possiblePreStateHash match {
       case Left(ex) =>
-        Left(BlockException(ex)).rightCast[Option[StateHash]].pure[F]
+        BlockException(ex).asLeft[Option[StateHash]].pure[F]
       case Right(computedPreStateHash) =>
         if (preStateHash == computedPreStateHash) {
           processPreStateHash[F](
@@ -99,33 +99,29 @@ object InterpreterUtil {
         case Left((Some(deploy), status)) =>
           status match {
             case InternalErrors(exs) =>
-              Left(
-                BlockException(
-                  new Exception(s"Internal errors encountered while processing ${PrettyPrinter
-                    .buildString(deploy)}: ${exs.mkString("\n")}")
-                )
-              ).rightCast[Option[StateHash]].pure[F]
+              BlockException(
+                new Exception(s"Internal errors encountered while processing ${PrettyPrinter
+                  .buildString(deploy)}: ${exs.mkString("\n")}")
+              ).asLeft[Option[StateHash]].pure[F]
             case UserErrors(errors: Vector[Throwable]) =>
-              Log[F].warn(s"Found user error(s) ${errors.map(_.getMessage).mkString("\n")}") *> Right(
-                none[StateHash]
-              ).leftCast[BlockException].pure[F]
+              Log[F].warn(s"Found user error(s) ${errors.map(_.getMessage).mkString("\n")}") *>
+                none[StateHash].asRight[BlockException].pure[F]
             case ReplayStatusMismatch(replay: DeployStatus, orig: DeployStatus) =>
               Log[F].warn(
                 s"Found replay status mismatch; replay failure is ${replay.isFailed} and orig failure is ${orig.isFailed}"
-              ) *> Right(none[StateHash]).leftCast[BlockException].pure[F]
+              ) *>
+                none[StateHash].asRight[BlockException].pure[F]
             case UnknownFailure =>
-              Log[F].warn(s"Found unknown failure") *> Right(none[StateHash])
-                .leftCast[BlockException]
-                .pure[F]
+              Log[F].warn(s"Found unknown failure") *>
+                none[StateHash].asRight[BlockException].pure[F]
             case UnusedCommEvent(_) =>
               Sync[F].raiseError(new RuntimeException("found UnusedCommEvent"))
           }
         case Left((None, status)) =>
           status match {
             case UnusedCommEvent(ex: ReplayException) =>
-              Log[F].warn(s"Found unused comm event ${ex.getMessage}") *> Right(none[StateHash])
-                .leftCast[BlockException]
-                .pure[F]
+              Log[F].warn(s"Found unused comm event ${ex.getMessage}") *>
+                none[StateHash].asRight[BlockException].pure[F]
             case InternalErrors(_) => throw new RuntimeException("found InternalErrors")
             case ReplayStatusMismatch(_, _) =>
               throw new RuntimeException("found ReplayStatusMismatch")
@@ -135,13 +131,14 @@ object InterpreterUtil {
         case Right(computedStateHash) =>
           if (tsHash.contains(computedStateHash)) {
             // state hash in block matches computed hash!
-            Right(Option(computedStateHash)).leftCast[BlockException].pure[F]
+            computedStateHash.some.asRight[BlockException].pure[F]
           } else {
             // state hash in block does not match computed hash -- invalid!
             // return no state hash, do not update the state hash set
             Log[F].warn(
               s"Tuplespace hash ${tsHash.getOrElse(ByteString.EMPTY)} does not match computed hash $computedStateHash."
-            ) *> Right(none[StateHash]).leftCast[BlockException].pure[F]
+            ) *>
+              none[StateHash].asRight[BlockException].pure[F]
 
           }
       }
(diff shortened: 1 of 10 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -80,7 +80,7 @@ object InterpreterUtil {
           Log[F].warn(
             s"Computed pre-state hash ${PrettyPrinter.buildString(computedPreStateHash)} does not equal block's pre-state hash ${PrettyPrinter
               .buildString(preStateHash)}"
-          ) *> Right(none[StateHash]).leftCast[BlockException].pure[F]
+          ) >> Right(none[StateHash]).leftCast[BlockException].pure[F]
```
tymm: I wonder what's the reasoning behind this change since `*>` is only dangerous when there is recursion, right?

ArturGajowy: Also when rhs depends on the (side)effect of lhs (in `lhs *> rhs` there's no guarantee which of the effects is going to be executed first). I'd like to use `>>` by default simply because the gotchas are subtle and can surface without being noticed.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -299,57 +299,55 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
...
+        cont <- DeployStatus.fromErrors(errors) match {
+                 case int: InternalErrors =>
+                   (deploy.some, int: Failed).asLeft[Blake2b256Hash].pure[F]
+                 case replayStatus =>
+                   if (status.isFailed != replayStatus.isFailed)
+                     (deploy.some, ReplayStatusMismatch(replayStatus, status): Failed)
+                       .asLeft[Blake2b256Hash]
+                       .pure[F]
+                   else if (errors.nonEmpty)
+                     hash.asRight[(Option[DeployData], Failed)].pure[F]
```
tymm: This seems wrong.

ArturGajowy: Why is that? What seems wrong?

tymm: I guess I just don't understand the function well enough but it seems wrong that we return when errors are non-empty.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -49,7 +49,6 @@ trait RuntimeManager[F[_]] {
   ): F[(StateHash, StateHash, Seq[InternalProcessedDeploy])]
   def storageRepr(hash: StateHash): F[Option[String]]
   def computeBonds(hash: StateHash): F[Seq[Bond]]
-  def computeBalance(start: StateHash)(user: ByteString): F[Long]
```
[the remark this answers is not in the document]
ArturGajowy: When would the balance check take place? Before execution? Notice that the deploy can also access the deployers vault and spend it to a place where there's nothing left. So there needs to be a transfer before the deploy being executed, and then a transfer of the leftovers back to the deployer's vault.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: Notice we're returning the previous initial hash. So the meaning of `processDeploy` is `try process, and return the resulting hash. If the execution was not successful, return the previous hash. We'll reset to it before continuing`. Notice also this is merely a refactoring of the direct recursion that took place before (IOW it maintains behavior).

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -349,6 +310,48 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     result.nested.map(_.toByteString).value
   }
 
+  private def replayDeploy(runtime: Runtime[F])(
+      hash: Blake2b256Hash,
+      processedDeploy: InternalProcessedDeploy
+  ): F[Either[(Option[DeployData], Failed), Blake2b256Hash]] = {
+    import processedDeploy._
+    for {
```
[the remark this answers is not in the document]
ArturGajowy: `replayComputeEffect` calls `rig` underneath, which calls `reset`, so *technically* the reset is there...

Good point though, I'll try to make it more obvious. I first scratched my head in the very same place, then seen the `rig` doing the reset, and then proceeded to refactoring and numbed a bit :sweat_smile:

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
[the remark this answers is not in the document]
JosDenmark: > Notice that the deploy can also access the deployers vault and spend it to a place where there's nothing left.

Would you explain this a bit more? The balance check would occur before the deploy is executed.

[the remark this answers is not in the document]
JosDenmark: Nvm. I see.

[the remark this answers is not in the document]
ArturGajowy: Say Bob deploys with phloLimit 100 and phloRate 1. His balance is 200. His deploy does a transfer of 150 rev from his vault to Alice. His deploy costs 75 phlo (= 75 rev)

With two transfers, Bob is going to hit a 'insufficient funds' before attempting the transfer (only 125 is left at deploy start). No further processing is needed after an OOPE. We need to call a secured method from PoS contract to do the refund, but we need to figure that out anyway for other casperwork.

With balance check + transfer after deploy, Bob's deploy won't be seen by the network at all - but the original validator will need to process its entirety. When OOPE happens, additional rholang execution is needed.
When the deploy finishes, we need to do another balance check, 
reset the deploy's state change if not enough is left to cover execution costs, and transfer the cost of phlo used. Otherwise the example discussed above ends up with an 'unexpected error during payment deploy'.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -99,33 +99,29 @@ object InterpreterUtil {
...
               Log[F].warn(
                 s"Found replay status mismatch; replay failure is ${replay.isFailed} and orig failure is ${orig.isFailed}"
-              ) *> Right(none[StateHash]).leftCast[BlockException].pure[F]
+              ) >>
+                none[StateHash].asRight[BlockException].pure[F]
             case UnknownFailure =>
-              Log[F].warn(s"Found unknown failure") *> Right(none[StateHash])
-                .leftCast[BlockException]
-                .pure[F]
+              Log[F].warn(s"Found unknown failure") >>
```
dzajkowski: `s""` seems no longer needed.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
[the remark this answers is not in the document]
dzajkowski: would it make sense to make this discussion part of a less volatile piece of data? (wiki page?)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -468,23 +429,20 @@ object RuntimeManager {
...
       )(user: ByteString, amount: Long): T[F, StateHash] =
         runtimeManager.computeDeployPayment(start)(user, amount).liftM[T]
 
-      override def getData(hash: StateHash)(channel: Par): T[F, scala.Seq[Par]] =
+      override def getData(hash: StateHash)(channel: Par): T[F, Seq[Par]] =
         runtimeManager.getData(hash)(channel).liftM[T]
 
       override def getContinuation(
           hash: StateHash
-      )(channels: Seq[Par]): T[F, scala.Seq[(scala.Seq[BindPattern], Par)]] =
```
dzajkowski: what's this, what's this? it's magic in the air...

please tell me it's the default Seq used.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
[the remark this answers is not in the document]
ArturGajowy: I'll make the renames in a further PR, b/c I'd like reviewers to be able to comment on the names chosen. I'll leverage my 2 :heavy_check_mark:-s in the meantime :D

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
ArturGajowy: WDYM? I hope so too. Lemme see. Yup, there's no imports from `scala.collection` in the file, so it's those `scala.` prefixes were just redundant.

PR #2465, opened 2019-05-29
RCHAIN-3490: Introduce payment log in ProcessedDeploy

JosDenmark:
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3490

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -1,17 +1,16 @@
 package coop.rchain.casper
 
 import cats.effect.{Concurrent, Sync}
-import cats.{Applicative, Monad}
 import cats.implicits._
-
+import cats.{Applicative, Monad}
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.{BlockDagRepresentation, BlockStore}
 import coop.rchain.casper.protocol.Event.EventInstance
 import coop.rchain.casper.protocol.{ApprovedBlock, BlockMessage, Justification}
-import coop.rchain.casper.util.{DagOperations, ProtoUtil}
-import coop.rchain.casper.util.ProtoUtil.{blockNumber, bonds}
-import coop.rchain.casper.util.rholang.{InterpreterUtil, RuntimeManager}
+import coop.rchain.casper.util.ProtoUtil.bonds
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
+import coop.rchain.casper.util.rholang.{InterpreterUtil, RuntimeManager}
+import coop.rchain.casper.util.{DagOperations, ProtoUtil}
 import coop.rchain.crypto.hash.Blake2b256
 import coop.rchain.crypto.signatures.Ed25519
 import coop.rchain.metrics.{Metrics, Span}
@@ -19,7 +18,8 @@ import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.BlockMetadata
 import coop.rchain.models.Validator.Validator
 import coop.rchain.shared._
-import scala.util.{Failure, Success, Try}
+
+import scala.util.{Success, Try}
 
 object Validate {
   type PublicKey = Array[Byte]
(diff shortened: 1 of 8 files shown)


JosDenmark:
bors try

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -139,8 +139,9 @@ message DeployData {
 message ProcessedDeploy {
     DeployData deploy = 1;
     PCost cost = 2 ;
-    repeated Event log = 3; //the new terms and comm. rule reductions from this deploy
-    bool errored = 4; //true if deploy encountered a user error
+    repeated Event deployLog = 3; //the new terms and comm. rule reductions from this deploy
+    repeated Event paymentLog = 4; //the comm. rule reductions from the payment for the deploy
```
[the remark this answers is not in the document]
ArturGajowy: I guess we'd be better off supporting it being non-deterministic than proving the contrary :D

ArturGajowy:
@KentShikama said in chat:

>I think you want to sort both logs. I expect to see one line added instead of one line removed and added.
Hopefully that makes sense without context.

It seems to me too that both logs need sorting

ArturGajowy:
Also, @KentShikama do we do log sorting in other places than the Genesis? I guess we should do that before sending the block out into the wild?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -151,7 +151,9 @@ object Validate {
       for {
         _ <- Log[F].warn(ignore(b, s"block post state is missing."))
       } yield false
-    } else if (b.body.get.deploys.flatMap(_.log).exists(_.eventInstance == EventInstance.Empty)) {
+    } else if (b.body.get.deploys
+                 .flatMap(_.deployLog)
+                 .exists(_.eventInstance == EventInstance.Empty)) {
```
ArturGajowy: same should be done for payment log

casper/src/main/scala/coop/rchain/casper/util/rholang/InternalProcessedDeploy.scala
```diff
@@ -30,7 +30,7 @@ object InternalProcessedDeploy {
     for {
       d <- pd.deploy
       c <- pd.cost
-      l = pd.log.map(EventConverter.toRspaceEvent)
+      l = pd.deployLog.map(EventConverter.toRspaceEvent)
       s = if (pd.errored) UnknownFailure else Succeeded
     } yield InternalProcessedDeploy(d, c, l, Seq.empty[trace.Event], s)
```
ArturGajowy: we should be reading the payment log in here instead of passing seq.empty

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -762,7 +757,8 @@ class ValidateTest
                    genesis.withBody(
                      genesis.body.get
                        .withDeploys(
-                         genesis.body.get.deploys.map(_.withLog(List(Event(EventInstance.Empty))))
+                         genesis.body.get.deploys
+                           .map(_.withDeployLog(List(Event(EventInstance.Empty))))
```
ArturGajowy: a copy of this assertion for paymentLog is needed

ArturGajowy:
One more thing: make sure to reformat all commits. Mind the PR build failure because of that (it won't stop merging (yet... :smiling_imp:), but is there for us to know)

casper/src/main/scala/coop/rchain/casper/Validate.scala
JosDenmark: Hmm. Right now, all of the payment logs are empty because no payment is occurring. Also, in `RuntimeManager`, it looks like the `InternalProcessedDeploy` corresponding to the `DeployData` is added to the output sequence even if an error is detected. What if a deploy doesn't have enough REV to pay for the deploy? Then, no deploy is processed, so no payment is made, so the payment log is empty. Is that something worth logging?

JosDenmark: Actually, doesn't that mean that we should allow empty event instances for the regular deploy log as well? With payments, it's no longer the case that deploys always result in a non-empty deploy log.

JosDenmark: Come to think of it, was never the case that deploys result in non-empty deploy logs. Consider the deployment `1 + 1`. Since it doesn't affect the store, there are no criteria against which we can validate it. Still, it's a valid deploy - albeit one that does not have a non-empty deploy log.

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
JosDenmark: See my arguments above.

casper/src/main/scala/coop/rchain/casper/Validate.scala
ArturGajowy: You're confusing the log case (en empty list - which is fine) with a non-empty list containing the nonsensical protobuf garbage that FooInstance.Empty is. If this ever happens, this simply means we're talking to an incompatible node.

PR #2469, opened 2019-05-29
Charging for deploy part 3.5

ArturGajowy:
## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -86,11 +86,9 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     runtime.deployParametersRef.set(ProtoUtil.getRholangDeployParams(deploy)) >>
       doInj(deploy, runtime.reducer, runtime.errorLog)(runtime.cost)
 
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -74,21 +74,19 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
 
   def captureResults(start: StateHash, deploy: DeployData, name: Par): F[Seq[Par]] =
     withResetRuntime(start) { runtime =>
-      computeEffect(runtime)(deploy)
+      computeEffect(runtime, runtime.reducer)(deploy)
```
JosDenmark: I hate that our choices are to either duplicate methods, one for evaluation the other for replay, or to duplicate a parameter, as is done here. We should basically make `RuntimeManager` a two-dimensional action indexed state machine, either in state 1) evaluation or in state 2) replay (distinguishing between the two statically with `Runtime` and `ReplayRuntime` types) where the actions are deploys. Then we can stop this duplication nonsense.

ArturGajowy: IDEA initially proposed `computeEffect(runtime, _.reducer)` - but this still allows for accessing the other reducer inside the method. I think a single facade parameter should actually be passed over runtime, reducer, and rspace.

PR #2473, opened 2019-05-30
Casper TLA+ spec

rabbitonweb:
## Overview
As show this week on standup. To generate PDF you need TLA+Toolbox.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3218


rabbitonweb:
Don't you love when PR that **does not** change a single line of code... fails on compilation phase? :D

dckc:
How serious is this spec effort? Is it (aimed to be) adequate to mechanically prove safety?

dckc:
Kent tells me the scope is genesis, not the main protocol. Please add a comment to that effect and/or change the filename.

PR #2474, opened 2019-05-30
Check and filter private host addresses

sebtomba:
## Overview
- By default forbid host addresses from a private address space.
- Filter private (and invalid) addresses from Kademlia lookup responses.
- Check on startup if the `--host` option and configuration are valid addresses.
- The address `0.0.0.0` (all IPv4 addresses on the local machine) is not a valid `--host` option.
- Introduce a new option `--allow-private-addresses`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3435


adaszko:
`The address 0.0.0.0 (all IPv4 addresses on the local machine) is not a valid --host option.`

Why?  I think it should be.

sebtomba:
@adaszko the purpose of the `--host` option is to specify manually an exact hostname or address. The `0.0.0.0` address is the default behavior of the node if the `--host` option/configuration is not provided. The node guesses then the host address. On the other hand, a node listens on all network interfaces in either case. But it needs to communicate to other nodes only one address. So `all IPv4 addresses` is not an option here.

comm/src/main/scala/coop/rchain/comm/package.scala
```diff
@@ -12,4 +17,20 @@ package object comm {
   }
 
   val CommMetricsSource: Metrics.Source = Metrics.Source(Metrics.BaseSource, "comm")
+
+  def isValidInetAddress(host: String): Boolean =
+    Try(InetAddress.getByName(host))
```
lukasz-golebiewski: I don't think `InetAddress.getByName` is a pure function, so it should be enclosed in `F` and delayed

adaszko:
> The 0.0.0.0 address is the default behavior of the node if the --host option/configuration is not provided. The node guesses then the host address.

The semantics of 0.0.0.0 aren't to guess an address but to use all addresses.

comm/src/main/scala/coop/rchain/comm/package.scala
sebtomba: @lukasz-golebiewski done

PR #2476, opened 2019-05-30
RCHAIN-3214 deploy not signed

rabbitonweb:
## Overview
Slash blocks that arrived with deploys whose signatures can not be verified

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3214


casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -190,60 +190,56 @@ object Validate {
...
-                                           Validate.justificationRegressions[F](block, genesis, dag)
-                                       )
-      _ <- span.mark("before-shrad-identifier-validation")
-      shardIdentifierStatus <- justificationRegressionsStatus.joinRight.traverse(
-                                _ => Validate.shardIdentifier[F](block, shardId)
-                              )
-    } yield shardIdentifierStatus.joinRight
+  ): F[Either[BlockStatus, ValidBlock]] = {
+    val getBlockSummary: EitherT[F, BlockStatus, ValidBlock] = for {
+      _ <- EitherT.liftF(span.mark("before-block-hash-validation"))
```
lukasz-golebiewski: The explicit lifting boilerplate looks horrible. Isn't it somehow possible to do the lifting implicitly?

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -792,6 +792,52 @@ class ValidateTest
       } yield result
   }
 
+  "Block validDeploySignatures" should "return Ok if all deploys are signed correctly" in withStorage {
```
lukasz-golebiewski: sed 's/return Ok/succeed' because no `Ok` is being returned anywhere

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -792,6 +792,52 @@ class ValidateTest
...
+    implicit blockStorey =>
```
lukasz-golebiewski: :-)

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -792,6 +792,52 @@ class ValidateTest
...
+      implicit blockDagStorage =>
+        // given
+        val (sk, pk) = Ed25519.newKeyPair
+        val block    = MultiParentCasperTestUtil.createGenesis(Map(pk -> 1))
```
lukasz-golebiewski: Why is building the heavy genesis block needed to test such a simple scenario?

casper/src/main/scala/coop/rchain/casper/Validate.scala
EncodePanda: I can only think of writing an `Ops` that cats does not have and instead of doing 

```
EitherT.liftF(span.mark("before-block-hash-validation"))
```

do

```
span.mark("before-block-hash-validation").liftF
```

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
EncodePanda: I've used existing test API that requires the genesis block to create block

```
trait BlockGenerator {
(...)
  def createBlock[F[_]: Monad: Time: BlockStore: IndexedBlockDagStorage](
      parentsHashList: Seq[BlockHash],
      genesis: BlockMessage,
     (...)

```

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
```diff
@@ -28,10 +30,10 @@ object SignDeployment {
   }
 
   def verify(deployData: DeployData): Option[Boolean] =
-    SignaturesAlg(deployData.sigAlgorithm).map { alg =>
+    SignaturesAlg(deployData.sigAlgorithm) >>= { alg =>
       val toVerify = clear(deployData).toByteString.toByteArray
       val hash     = Blake2b256.hash(toVerify)
-      alg.verify(hash, deployData.sig.toByteArray, PublicKey(deployData.deployer.toByteArray))
+      Try(alg.verify(hash, deployData.sig.toByteArray, PublicKey(deployData.deployer.toByteArray))).toOption
```
odeac, line 43: It doesn't make sense that alg.verify throws even if the underlying implementation does throw. 

Ed25519 implementation already has a try/catch around the code that could fail.

So, instead of adding a Try here, we should add a try in the implementation of Secp256k1

casper/src/main/scala/coop/rchain/casper/Validate.scala
odeac: but what about an implicit conversionl as @lukasz-golebiewski says. Doesn't that work?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -190,60 +190,56 @@ object Validate {
...
+      _ <- EitherT(Validate.justificationFollows[F](block, genesis, dag))
+      _ <- EitherT.liftF(span.mark("before-parents-validation"))
+      _ <- EitherT(Validate.parents[F](block, genesis, dag))
+      _ <- EitherT.liftF(span.mark("before-sequence-number-validation"))
+      _ <- EitherT(Validate.sequenceNumber[F](block, dag))
+      _ <- EitherT.liftF(span.mark("before-justification-regression-validation"))
+      _ <- EitherT(Validate.justificationRegressions[F](block, genesis, dag))
+      _ <- EitherT.liftF(span.mark("before-shrad-identifier-validation"))
+      _ <- EitherT(Validate.shardIdentifier[F](block, shardId))
+    } yield Valid
```
odeac: isn't this a case where the operator `>>` would make the code cleaner than the for comprehension?

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
EncodePanda: @odeac you are not correct, it was actuallt `Ed25519` that was throwing erros! That's why the Try was added in the first place!

casper/src/main/scala/coop/rchain/casper/Validate.scala
EncodePanda: it potentially would, matter of style

casper/src/main/scala/coop/rchain/casper/Validate.scala
EncodePanda: definiettly not a good idea to add some magical convertions in some global scope that would lift Fs to EitherT[F. good luck with figuring out compiler issues

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
EncodePanda: 👍 fixed

rabbitonweb:
bors try

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -791,6 +791,52 @@ class ValidateTest
...
+  "Block validDeploySignatures" should "succeed if all deploys are signed correctly" in withStorage {
+    implicit blockStorey =>
+      implicit blockDagStorage =>
+        // given
+        val (sk, pk) = Secp256k1.newKeyPair
+        val block    = MultiParentCasperTestUtil.createGenesis(Map(pk -> 1))
+        for {
+          dag     <- blockDagStorage.getRepresentation
+          deploy1 <- ConstructDeploy.basicProcessedDeploy[Task](0, sk)
+          deploy2 <- ConstructDeploy.basicProcessedDeploy[Task](0, sk)
```
KentShikama: If you keep the deploy contents all as 0, then it will be as if you only deployed 1.

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
```diff
@@ -791,6 +791,52 @@ class ValidateTest
...
+                    deploys = Seq(deploy1, deploy2, deploy3)
+                  )
+          // when
+          validationResult <- Validate.validDeploySignatures[Task](block).value
+          // then
+        } yield validationResult shouldBe Right(())
+  }
+
+  it should "fail if one of deploys is not properly signed" in withStorage {
+    implicit blockStorey => implicit blockDagStorage =>
```
KentShikama: blockStore

EncodePanda: 👍

casper/src/test/scala/coop/rchain/casper/ValidateTest.scala
EncodePanda: see i never knew what `0` stands for, good to know, thx!

KentShikama:
bors try

KentShikama:
@rabbitonweb Is the only issue with this PR that it needs to pass bors?

KentShikama:
I see; this PR needs to update the integration tests to have the deploys be signed.

KellyatPyrofex:
Discussed during today's standup. @rabbitonweb, please create a ticket if this PR is blocked by changes needed in an integration test. If that is the case, please label this PR as *blocked*.

KentShikama:
bors try

KentShikama:
I've gone ahead and done a rebase for this PR

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/SignDeployment.scala
```diff
@@ -28,10 +30,10 @@ object SignDeployment {
   }
 
   def verify(deployData: DeployData): Option[Boolean] =
-    SignaturesAlg(deployData.sigAlgorithm).map { alg =>
+    SignaturesAlg(deployData.sigAlgorithm) >>= { alg =>
       val toVerify = clear(deployData).toByteString.toByteArray
```
zsluedem: Just curious. Why is `clear` here?

All the integration tests related to the `deploy and propose` failed return with `Failure! Invalid block: DeployNotSigned`.

KentShikama: `clear` might be a misleading name; maybe `normalize` is better

The integration tests related to `deploy` `propose` are failing because we aren't signing the deploys, which needs to be done off-chain.

zsluedem: So every deploy is signed with empty bytes? It doesn't look right to me.

zsluedem: Oh, there is some other data in `DeployData`. My bad.

KentShikama: `verify` is used to verify transactions. Maybe you meant to look at

```
  def sign(sec: PrivateKey, deployData: DeployData, alg: SignaturesAlg = Secp256k1): DeployData = {
    val toSign    = clear(deployData).toByteString.toByteArray
    val hash      = Blake2b256.hash(toSign)
    val signature = alg.sign(hash, sec)
    fill(deployData)(alg.toPublic(sec), signature, alg.name)
  } 
```

KentShikama: Yeah

KentShikama:
bors try

sebtomba:
bors try

sebtomba:
r+

PR #2481, opened 2019-05-31
Uncaught FileNotFoundException when deleting bts file

sebtomba:
## Overview
This problem is very mysterious. The logs show that the file has been streamed to a peer and then deleted. And then 4 seconds later node tries to stream and delete the same file again. The only explanation that comes to my mind it that the file has been enqueued twice to the stream queue.

For now, I did:
1. Improved the code so it anticipates the case when the file is missing and don’t throws an exception.
2. Added a debug message each time a packet file gets enqueued. Next time we can check if a packet got enqueued twice for streaming.

Caveat:
Since there will be no exception thrown anymore we need to monitor the logs for stream errors.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3443

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/PacketOps.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/PacketOps.scala
@@ -1,52 +1,58 @@
 package coop.rchain.comm.transport
 
-import coop.rchain.shared.GracefulClose._
-import coop.rchain.comm.CommError, CommError._
-import coop.rchain.comm.protocol.routing._
-import java.nio.file._
 import java.io._
-import cats._, cats.data._, cats.implicits._
-import cats.effect.Sync
-import java.util.Date
+import java.nio.file._
 import java.text.SimpleDateFormat
+import java.util.Date
 
 import scala.util.Random
 
+import cats.data._
+import cats.effect.Sync
+import cats.implicits._
+
+import coop.rchain.catscontrib.ski.kp
+import coop.rchain.comm.CommError
+import coop.rchain.shared.GracefulClose._
+import CommError._
+import coop.rchain.comm.protocol.routing._
 import coop.rchain.crypto.codec.Base16
 
 object PacketOps {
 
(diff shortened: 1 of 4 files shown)


comm/src/main/scala/coop/rchain/comm/transport/PacketOps.scala
```diff
@@ -1,52 +1,58 @@
...
-                   gracefullyClose(fin) >> Left(unableToRestorePacket(file, th)).pure[F]
-                 case Right(packet) => gracefullyClose(fin) >> Right(packet).pure[F]
-               }
-    } yield resErr
+  def restore[F[_]: Sync](file: Path): F[CommErr[Packet]] = {
+    def fileInputStream: F[CommErr[FileInputStream]] =
+      Sync[F]
+        .delay(new FileInputStream(file.toFile))
+        .attempt
+        .map(_.fold(unableToRestorePacket(file, _).asLeft, _.asRight))
```
itegulov: This should be a `Resource`

comm/src/main/scala/coop/rchain/comm/transport/PacketOps.scala
```diff
@@ -1,52 +1,58 @@
...
-                   case Left(th) =>
-                     gracefullyClose(fos) >> Left(unableToStorePacket(packet, th)).pure[F]
-                   case Right(_) =>
-                     gracefullyClose(fos) map {
-                       case Left(th) => Left(unableToStorePacket(packet, th))
-                       case Right(_) => Right(file)
-                     }
-                 }
-      } yield resErr
+        closeErr <- gracefullyClose(fos)
```
itegulov: I would say `createPacketFile` should return `Resource[F, PacketFile]` as well but it probably falls out of scope of this particular PR

sebtomba: `StreamHandler` is using this function too. So maybe next time.

PR #2489, opened 2019-06-04
Add a command line option for using random ports

itegulov:
## Overview
Adds a command line option for using random ports when RChain Protocol port and/or Kademlia port are not free.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3342

### Notes


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -160,15 +173,44 @@ object Main {
...
+                                if (configuration.server.useRandomPorts)
+                                  for {
+                                    port <- getFreePort
+                                    _    <- log.info(s"Using random port $port as $portName port")
+                                  } yield setPort(configuration, port)
+                                else
+                                  Task.delay(System.exit(1)).as(configuration)
+                              )
+                          }
+    } yield confWithFreePorts
```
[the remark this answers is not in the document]
itegulov: Yeah, I am not entirely happy with this piece of code. My reasoning for writing without code duplication was a possibility of using random ports with other ports as well (e.g. http port, grpc ports).

I don't have a particularly strong opinion here so if you think code duplication is the way to go here, I am happy to change it.

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -167,15 +180,41 @@ object Main {
...
+    def checkRChainProtocolPort(configuration: Configuration): Task[Configuration] =
+      isLocalPortAvailable(configuration.server.port).ifM(
+        configuration.pure[Task],
+        if (configuration.server.useRandomPorts)
+          for {
+            port <- getFreePort
+            _    <- log.info(s"Using random port $port as RChain Protocol port")
+          } yield configuration.copy(server = configuration.server.copy(port = port))
+        else
+          Task.delay(System.exit(1)).as(configuration)
```
odeac: Is it ok exist just like this? Any resources to be cleaned up?

I would rather have a clean exit point from the main function instead of many exit points from these functions. Would that be difficult to achieve?

We should also print some error message.

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -167,15 +180,41 @@ object Main {
...
+    def checkKademliaPort(configuration: Configuration): Task[Configuration] =
+      isLocalPortAvailable(configuration.server.kademliaPort).ifM(
+        configuration.pure[Task],
+        if (configuration.server.useRandomPorts)
+          for {
+            port <- getFreePort
+            _    <- log.info(s"Using random port $port as Kademlia port")
+          } yield configuration.copy(server = configuration.server.copy(kademliaPort = port))
+        else
+          Task.delay(System.exit(1)).as(configuration)
```
odeac: same

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -167,15 +180,41 @@ object Main {
...
+      )
+
+    for {
+      _ <- List(
+            conf.server.httpPort,
+            conf.grpcServer.portExternal,
+            conf.grpcServer.portInternal
+          ).traverse(isLocalPortAvailable)
+            .map(_.forall(identity))
+            .ifM(Task.unit, Task.delay(System.exit(1)))
```
odeac: same

node/src/main/scala/coop/rchain/node/Main.scala
itegulov: @sebtomba any specific reason why you decided to use `System.exit(1)` originally here?

rabbitonweb: @itegulov I would create a seperate ticket in jira for that to get this cleaned up asap

sebtomba: We have a growing number of `System.exit(1)` in the codebase. Maybe it's time to find a better approach in `Main`

itegulov: Ticketed: https://rchain.atlassian.net/browse/RCHAIN-3558

PR #2502, opened 2019-06-10
Add spans and span markers to Casper RuntimeManager

sebtomba:
## Overview
Add spans and span markers to Casper `RuntimeManager`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3507

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -5,13 +5,16 @@ import cats.data.EitherT
 import cats.effect.concurrent.MVar
 import cats.effect.{Sync, _}
 import cats.implicits._
+
 import com.google.protobuf.ByteString
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
 import coop.rchain.casper.util.{ConstructDeploy, ProtoUtil}
+import coop.rchain.casper.CasperMetricsSource
 import coop.rchain.catscontrib.Catscontrib.ToMonadOps
 import coop.rchain.catscontrib.MonadTrans
 import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.models.Expr.ExprInstance.GString
 import coop.rchain.models.Var.VarInstance.FreeVar
 import coop.rchain.models._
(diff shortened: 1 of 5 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -98,8 +108,10 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
   ): F[Either[ReplayFailure, StateHash]] =
     withRuntimeLock { runtime =>
       for {
+        span   <- Metrics[F].span(replayComputeStateLabel)
         _      <- setBlockTime(blockTime, runtime)
-        result <- replayDeploys(startHash, terms, replayDeploy(runtime))
+        _      <- span.mark("before-replay-deploys")
+        result <- replayDeploys(startHash, terms, replayDeploy(runtime, span))
```
odeac: passing the span to the called function feels wrong to me. I would have a Metrics effect with operations like: beginSpan, mark.
If you do that you won't have to pass the span and the functions will just place their own marks.

Just my 2c

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -121,8 +135,10 @@ class RuntimeManagerImpl[F[_]: Concurrent] private[rholang] (
     val startHash = emptyStateHash
     withRuntimeLock { runtime =>
       for {
+        span       <- Metrics[F].span(computeGenesisLabel)
         _          <- setBlockTime(blockTime, runtime)
-        evalResult <- processDeploys(startHash, terms, processDeploy(runtime))
+        _          <- span.mark("before-process-deploys")
```
odeac: ...for example here you are calling span.mark here because you don't want to pass the span in processDeploys. 

Because of this you have to duplicate the mark message

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
KentShikama: I prefer having the UI display all related spans in one table.

Note the above sentence may not be relevant.

sebtomba: @odeac I'm not sure I understand what you are proposing. We don't want to keep in `Metrics` track of all spans. Our current implementation in `Metrics[F[_]]` is
```scala
def span(source: Metrics.Source): F[Span[F]]
```

sebtomba: @KentShikama `RuntimeManager` is a `trait`. I don't want to introduce _spans_ into the traits methods parameter list. So `RuntimeManagerImpl` starts its own spans.

PR #2504, opened 2019-06-10
RCHAIN-3503: Implement basic lock

JosDenmark:
## Overview
This is the basic lock needed to write the readers-writer lock (upcoming PR).

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3503

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
@@ -1,0 +1,41 @@
+package coop.rchain.rholang.interpreter.error
+
+import cats.effect.{Bracket, Concurrent}
+import cats.implicits._
+import cats.effect.concurrent.MVar
+
+final class MLock[F[_]](lock: MVar[F, Unit])(implicit bracketF: Bracket[F, Throwable]) {
+
+  def acquire: F[Unit] = lock.take
+
+  /**
+    * Attempts to release lock.
+    *
+    * Throws an error if lock is already released.
+    *
+    */
+  def release: F[Unit] =
+    lock
+      .tryPut(())
+      .ifM(
+        ().pure[F],
+        LockReleaseException("attempted to release released lock").raiseError[F, Unit]
+      )
+
+  def withLock[A](fa: F[A]): F[A] = bracketF.bracket[Unit, A](acquire)(_ => fa)(_ => release)
+
+  /**
+    * Blocks asynchronously until lock is released.
+    *
+    * The next call to release will unblock all readers simultaneously.
+    *
+    */
+  def waitForLock: F[Unit] = lock.read
+
+}
+
+object MLock {
+
+  def apply[F[_]: Concurrent]: F[MLock[F]] = MVar.of[F, Unit](()).map(new MLock[F](_))
+
+}
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/error/MLockSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/error/MLockSpec.scala
@@ -1,0 +1,85 @@
+package coop.rchain.rholang.interpreter.error
+
+import monix.eval.Task
+import monix.execution.schedulers.TestScheduler
+import org.scalatest.FlatSpec
+
+import scala.concurrent.duration._
+import scala.util.{Failure, Success}
+
+class MLockSpec extends FlatSpec {
+
+  implicit val scheduler: TestScheduler = TestScheduler()
+
+  private val MAX_DURATION = 5.seconds
+
+  def withFixture[A](f: MLock[Task] => A): A = f(MLock[Task].runSyncUnsafe(MAX_DURATION))
+
+  behavior of "MLock"
+
+  "release" should "throw error when released lock is released" in withFixture { lock =>
+    val a = lock.release.runToFuture
+
+    scheduler.tick()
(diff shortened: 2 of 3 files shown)


JosDenmark:
bors try

JosDenmark:
@KentShikama It's not that different. It's basically a purely functional wrapper around a SyncVar that holds `Unit`.

lukasz-golebiewski:
Isn't this code functionally equivalent to what a `Semaphore` with a single permit offers?

rholang/src/test/scala/coop/rchain/rholang/interpreter/error/MLockSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import scala.concurrent.duration._
+
+class MLockSpec extends FlatSpec with Matchers {
+
+  import monix.execution.Scheduler.Implicits.global
+
+  private val MAX_DURATION = 10.seconds
+
+  def withFixture[A](f: MLock[Task] => Task[A]): A =
```
lukasz-golebiewski: `in fixture` reads better than `in withFixture`

rholang/src/test/scala/coop/rchain/rholang/interpreter/error/MLockSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+      .map(_ should be(Left(LockReleaseException("attempted to release released lock"))))
+  )
+
+  "release" should "unblock lock acquisitions in FIFO order" in withFixture { lock =>
+    for {
+      ref <- Ref.of[Task, Int](0)
+      _   <- lock.acquire
+      result <- List(
+                 lock.acquire >> ref.getAndSet(1),
+                 Task.sleep(0.1.seconds) >> lock.acquire >> ref.getAndSet(2),
```
lukasz-golebiewski: Using `sleep` can lead to potential race conditions and test failures in rare cases. Consider using this approach instead: https://monix.io/docs/3x/execution/scheduler.html#injecting-time-and-tests

JosDenmark:
@lukasz-golebiewski What would be the analogue to `waitForLock`?

rholang/src/test/scala/coop/rchain/rholang/interpreter/error/MLockSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+    } yield assert(result.head == 0 && result(1) == 0 && result(2) == 1)
+  }
+
+  "release" should "unblock all waiting threads simultaneously" in withFixture { lock =>
+    for {
+      ref   <- Ref.of[Task, Int](0)
+      _     <- lock.acquire
+      count = 999
+      _ <- List(
+            (0 to count).toList.parTraverse_(_ => lock.waitForLock >> ref.update(_ + 1)),
```
lukasz-golebiewski: I'll answer to your comment https://github.com/rchain/rchain/pull/2504#issuecomment-500872011 here as this is the only usage of `lock.waitForLock`. If you remove `lock.waitForLock >>` from the test above it still passes. This is due to the nature of `ref.update` as stated in the docs:
```/**
   * Modifies the current value using the supplied update function. If another modification
   * occurs between the time the current value is read and subsequently updated, the modification
   * is retried using the new value. Hence, `f` may be invoked multiple times.
   *
   * Satisfies:
   *   `r.update(_ => a) == r.set(a)`
   */
  def update(f: A => A): F[Unit]
```

Maybe the functionality of `waitForLock` isn't absolutely necessary?

JosDenmark: Its functionality is required in a following PR:

```
final class RWLock[F[_]](stateRef: MVar[F, State], readLock: MLock[F], writeLock: MLock[F])(
    implicit bracketF: Bracket[F, Throwable]
) {

  /**
    * Acquires read lock.
    *
    * Blocks (asynchronously) if another thread has write lock.
    *
    */
  def acquireRead: F[Unit] =
    stateRef.take >>= {
      case Free        => readLock.acquire >> stateRef.put(Read(1))
      case Write       => stateRef.put(Write) >> writeLock.waitForLock >> acquireRead
      case Read(count) => stateRef.put(Read(count + 1))
    }
...
}

File: rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala [hunk 4]
```diff
@@ -0,0 +1,41 @@
+package coop.rchain.rholang.interpreter.error
+
+import cats.effect.{Bracket, Concurrent}
+import cats.implicits._
+import cats.effect.concurrent.MVar
+
+final class MLock[F[_]](lock: MVar[F, Unit])(implicit bracketF: Bracket[F, Throwable]) {
+
+  def acquire: F[Unit] = lock.take
```
odeac: Do we really want this low level API? What happens is we only expose a resource like API?

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
```diff
@@ -0,0 +1,41 @@
...
+    */
+  def release: F[Unit] =
+    lock
+      .tryPut(())
+      .ifM(
+        ().pure[F],
+        LockReleaseException("attempted to release released lock").raiseError[F, Unit]
+      )
+
+  def withLock[A](fa: F[A]): F[A] = bracketF.bracket[Unit, A](acquire)(_ => fa)(_ => release)
```
odeac: Using cats resource would be better for composability

PR #2509, opened 2019-06-11
Fixed a bug in data generated to test RSpace Next History (#RCHAIN-3544)

dzajkowski:
## Overview
The keys passed to History process should not be duplicated. The generated data did not honour this assumption.

### JIRA ticket:
RCHAIN-3544

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/HistoryInstances.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/HistoryInstances.scala
@@ -332,6 +332,12 @@ object HistoryInstances {
         )
 
       for {
+        _ <- Sync[F].ifM((actions.map(_.key).toSet.size == actions.size).pure[F])(
+              ifTrue = Applicative[F].unit,
+              ifFalse = Sync[F].raiseError(
+                new RuntimeException("Cannot process duplicate actions on one key")
+              )
+            )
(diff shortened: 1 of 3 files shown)


rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/history/HistoryInstances.scala
```diff
@@ -294,6 +294,7 @@ object HistoryInstances {
 
     def process(actions: List[HistoryAction]): F[History[F]] = {
       type LastModification = (KeyPath, Trie)
+      assert(actions.map(_.key).toSet.size == actions.size) // expect the list of actions to modify distinct keys
```
lukasz-golebiewski: This assertion will throw an `AssertionError` when it fails. It would make sense to raise an error in `F` in such case. An even better idea would be to enforce correctness using the type system, that is by requiring `def process(actions: Set[HistoryAction])`

dzajkowski: the type system will not guarantee uniqueness in this case. two `HistoryAction` items on the same key but different values will pass this simplistic test.
regarding using `F`: we had a discussion regarding this topic and the conclusion was to not to utilise `F` as a carrier of critical errors.
I don't feel strongly about it and will modify with `raiseError`.

PR #2513, opened 2019-06-11
RCHAIN-3504: Implement readers-writer lock

JosDenmark:
### JIRA ticket:

The RWLock will be used to guard access to a single `Option[InterpreterError]` resource. We want error reporting to be such that any number of threads can see if there has been an error, but only one thread at a time can report such an error. 

A subsequent PR should modify the reducer such that each thread checks the error resource before executing. If an error has been reported, it doesn't execute. Otherwise, it does. This usage implies that the error resource will only be modified once for a given term evaluation. The first error is the last.

https://rchain.atlassian.net/browse/RCHAIN-3504

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
@@ -1,0 +1,41 @@
+package coop.rchain.rholang.interpreter.error
+
+import cats.effect.{Bracket, Concurrent}
+import cats.implicits._
+import cats.effect.concurrent.MVar
+
+final class MLock[F[_]](lock: MVar[F, Unit])(implicit bracketF: Bracket[F, Throwable]) {
+
+  def acquire: F[Unit] = lock.take
+
+  /**
+    * Attempts to release lock.
+    *
+    * Throws an error if lock is already released.
+    *
+    */
+  def release: F[Unit] =
+    lock
+      .tryPut(())
+      .ifM(
+        ().pure[F],
+        LockReleaseException("attempted to release released lock").raiseError[F, Unit]
+      )
+
+  def withLock[A](fa: F[A]): F[A] = bracketF.bracket[Unit, A](acquire)(_ => fa)(_ => release)
+
+  /**
+    * Blocks asynchronously until lock is released.
+    *
+    * The next call to release will unblock all readers simultaneously.
+    *
+    */
+  def waitForLock: F[Unit] = lock.read
+
+}
+
+object MLock {
+
+  def apply[F[_]: Concurrent]: F[MLock[F]] = MVar.of[F, Unit](()).map(new MLock[F](_))
+
+}
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/RWLock.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/error/RWLock.scala
@@ -1,0 +1,166 @@
+package coop.rchain.rholang.interpreter.error
+
+import cats.effect.concurrent.MVar
+import cats.effect.{Bracket, Concurrent}
+import cats.implicits._
+
+/**
+  * [[Free]]: Read or write lock can be acquired without blocking.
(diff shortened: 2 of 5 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/error/RWLock.scala
```diff
@@ -0,0 +1,163 @@
...
+  /**
+    * Acquires read lock.
+    *
+    * Blocks (asynchronously) if another thread has write lock.
+    *
+    */
+  def acquireRead: F[Unit] =
+    stateRef.take >>= {
+      case Free        => readLock.acquire >> stateRef.put(Read(1))
+      case Write       => stateRef.put(Write) >> writeLock.waitForLock >> acquireRead
```
JosDenmark: @lukasz-golebiewski This is why the wait functionality is required. This isn't infinite recursion as you suggested. In the case that the state reference is in the `Write` state, the `waitForLock` function asynchronously blocks the `acquireRead` function until the write lock is released. The only time one call to `acquireRead` recurses more than once is if another write permit is granted before it calls `stateRef.take`.

JosDenmark: The wait functionality is required because we want all readers to wake up simultaneously when the write lock is returned.

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
```diff
@@ -0,0 +1,41 @@
+package coop.rchain.rholang.interpreter.error
+
+import cats.effect.{Bracket, Concurrent}
+import cats.implicits._
+import cats.effect.concurrent.MVar
+
+final class MLock[F[_]](lock: MVar[F, Unit])(implicit bracketF: Bracket[F, Throwable]) {
```
odeac: Why implement a lock when you have the less error prone MVar?

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def acquire: F[Unit] = lock.take
```
odeac: I'm not happy with the interface of MLock eposing low-level primitives like aquire and release. Can't we just expose resource?

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/RWLock.scala
lukasz-golebiewski: I never suggested this *was* infinite recursion, I merely stated that this is what I see in the snippet you provided in https://github.com/rchain/rchain/pull/2504#discussion_r292542311 mainly due to the lack of context

lukasz-golebiewski: This PR and the tests make things clearer. Thanks!

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
JosDenmark: Their behaviors are different. Take `release` for example. It is an error to release a released lock. I also don't need to expose the entire interface of `MVar`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/MLock.scala
JosDenmark: The `withLock` function provides something similar to resource. It's not that simple when you want to use `MLock` as a primitive in more complex locking structures. See the subsequent PR.

odeac: then we should replace withLock with a proper Resource that is composable with other resources.

And we should also hide acquire/release because they are error prone

ArturGajowy:
Also, at least half of http://degoes.net/articles/zio-cats-effect seems related

PR #2515, opened 2019-06-11
RCHAIN-2747 remove fauce (part 2)

rabbitonweb:
**Death to Faucet!**

This is part 2/2 of faucet removal. Nothing fancy.

JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2747


integration-tests/test/test_heterogenous_validators.py
```diff
@@ -37,11 +37,6 @@ def test_heterogenous_validators(command_line_options: CommandLineOptions, rando
                     bonded_validator.propose()
 
                 with bootstrap_connected_peer(context=context, bootstrap=bootstrap_node, name='joining-validator', keypair=JOINING_VALIDATOR_KEYPAIR) as joining_validator:
-                    joining_validator.generate_faucet_bonding_deploys(
-                        bond_amount=123,
-                        private_key=JOINING_VALIDATOR_KEYPAIR.private_key,
-                        public_key=JOINING_VALIDATOR_KEYPAIR.public_key,
-                    )
                     wait_for_peers_count_at_least(context, bonded_validator, 2)
```
KentShikama: Does this test still pass?

KentShikama:
bors try

integration-tests/test/test_heterogenous_validators.py
rabbitonweb: this test was already marked `@pytest.mark.skip(reason="Skipped because we don't support bonding/unbonding in the new PoS yet")`

KentShikama: In that case, I would just leave this test intact as a reference until we replace it with the new PoS contract. Or if we feel like we don't need this reference, just remove the entire test.

rabbitonweb: but I've removed `generate_faucet_bonding_deploys` from testing framework

KentShikama: Yeah my point is that you've destroyed the entire test so it doesn't make sense to only remove `generate_faucet_bonding_deploys`

rabbitonweb: Ok, then I vote for leaving just method name (empty implementation) set to pending. This way somebody who will implement that they will work from scratch (they can always ask git to see how it looked like before). Makes sense?

KentShikama: Yeah sounds good

ArturGajowy:
There are unused imports in the integration test you've changed, pylint fails the build :(

rabbitonweb:
what was the definition of insanity?

rabbitonweb:
@ArturGajowy omg !

PR #2520, opened 2019-06-12
Peek implementation via a list of peek indexes

dzajkowski:
## Overview

### JIRA ticket:
RCHAIN-1587

### Notes

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
@@ -115,7 +115,8 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
                 channels,
                 bindPatterns,
                 TaggedContinuation(ParBody(body)),
-                false
+                false,
+                Seq.empty
               )
             )
           )
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
@@ -35,6 +35,7 @@ import scala.collection.mutable
 import scala.collection.mutable.HashMap
 import scala.concurrent.Await
 import scala.concurrent.duration._
+import scala.collection.SortedSet
 
 final case class TestFixture(space: RhoISpace[Task], reducer: ChargingReducer[Task])
 
@@ -116,7 +117,7 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
                 bindPatterns,
                 TaggedContinuation(ParBody(body)),
                 false,
-                Seq.empty
+                SortedSet.empty
               )
             )
           )
--- a/rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
+++ b/rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
@@ -82,11 +82,12 @@ abstract class ExamplesSerializerBenchState {
     phone = "555-6969"
   )
 
-  val channel      = Channel("colleagues")
-  val channels     = List(channel, Channel("friends"))
-  val datum        = Datum.create(channel, data, false)
-  val patterns     = Seq[Pattern](CityMatch(city = "Crystal Lake"))
-  val continuation = WaitingContinuation.create(channels, patterns, new EntriesCaptor(), false)
+  val channel  = Channel("colleagues")
+  val channels = List(channel, Channel("friends"))
+  val datum    = Datum.create(channel, data, false)
+  val patterns = Seq[Pattern](CityMatch(city = "Crystal Lake"))
+  val continuation =
+    WaitingContinuation.create(channels, patterns, new EntriesCaptor(), false, Seq.empty)
 
   def gnat() = GNAT[Channel, Pattern, Entry, EntriesCaptor](
     channels,
@@ -103,7 +104,9 @@ abstract class ExamplesSerializerBenchState {
       channels,
       range.map(i => Datum.create(channels(i - 1), data, false)),
       range.map(
-        i => WaitingContinuation.create(channels.take(i), patterns, new EntriesCaptor(), false)
+        i =>
+          WaitingContinuation
+            .create(channels.take(i), patterns, new EntriesCaptor(), false, Seq.empty)
       )
     )
   }
--- a/rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
+++ b/rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
@@ -11,6 +11,7 @@ import coop.rchain.rspace.bench.serialization._
 
 import org.openjdk.jmh.annotations.{State => BenchState, _}
 import org.openjdk.jmh.infra.Blackhole
+import scala.collection.SortedSet
 
 class ExamplesSerializerBench {
 
@@ -87,7 +88,7 @@ abstract class ExamplesSerializerBenchState {
   val datum    = Datum.create(channel, data, false)
   val patterns = Seq[Pattern](CityMatch(city = "Crystal Lake"))
   val continuation =
-    WaitingContinuation.create(channels, patterns, new EntriesCaptor(), false, Seq.empty)
+    WaitingContinuation.create(channels, patterns, new EntriesCaptor(), false, SortedSet.empty)
 
   def gnat() = GNAT[Channel, Pattern, Entry, EntriesCaptor](
     channels,
@@ -106,7 +107,7 @@ abstract class ExamplesSerializerBenchState {
       range.map(
         i =>
           WaitingContinuation
-            .create(channels.take(i), patterns, new EntriesCaptor(), false, Seq.empty)
+            .create(channels.take(i), patterns, new EntriesCaptor(), false, SortedSet.empty)
       )
     )
   }
--- a/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
@@ -77,6 +77,7 @@ class RSpace[F[_], C, P, A, R, K] private[rspace] (
                 patterns,
                 continuation,
                 persist,
+                Seq.empty,
                 consumeRef
               )
             )
@@ -312,6 +313,7 @@ class RSpace[F[_], C, P, A, R, K] private[rspace] (
             patterns,
             continuation,
             persistK,
+            _,
             consumeRef
           ),
           continuationIndex,
--- a/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
@@ -16,6 +16,7 @@ import coop.rchain.shared.SyncVarOps._
 import org.lmdbjava.Txn
 import scodec.Codec
 
+import scala.collection.SortedSet
 import scala.concurrent.ExecutionContext
 import scala.util.Random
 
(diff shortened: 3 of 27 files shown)


rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/ReplayRSpaceTests.scala
```diff
@@ -789,6 +790,501 @@ trait ReplayRSpaceTests extends ReplayRSpaceTestsBase[String, Pattern, String, S
       }
   }
 
+  "Picking a datum from 100 waiting datums" should "replay correctly" in fixture {
```
lukasz-golebiewski: I'm failing to see the value of having this test along with "Picking n datums from m waiting datums"

lukasz-golebiewski: This whole commit looks unnecessary

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -115,7 +115,8 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
                 channels,
                 bindPatterns,
                 TaggedContinuation(ParBody(body)),
-                false
+                false,
```
lukasz-golebiewski: Judging from the commit name the previous one does not compile or it's tests are failing. These should be squashed then

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -77,6 +77,7 @@ class RSpace[F[_], C, P, A, R, K] private[rspace] (
                 patterns,
                 continuation,
                 persist,
+                Seq.empty,
```
lukasz-golebiewski: Why have you suddenly switched to using `Seq` instead of `Set`?

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/RSpace.scala
```diff
@@ -210,10 +197,19 @@ class RSpace[F[_], C, P, A, R, K] private[rspace] (
                                       _ <- metricsF.incrementCounter(consumeCommLabel)
                                       _ <- syncF.delay {
                                             eventLog.update(
-                                              COMM(consumeRef, dataCandidates.map(_.datum.source)) +: _
+                                              COMM(
+                                                consumeRef,
+                                                dataCandidates.map(_.datum.source),
+                                                peeks.toSeq.sorted
```
lukasz-golebiewski: I see now. This looks costly. Better enforce `SortedSet` in the method's signature and use the same in `COMM`

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/RSpace.scala
```diff
@@ -344,7 +340,10 @@ class RSpace[F[_], C, P, A, R, K] private[rspace] (
                   Datum(_, persistData, _),
                   dataIndex
                   ) => {
-                (if (!persistData && dataIndex >= 0) {
+                val idx = channelsToIndex(candidateChannel)
+                (if (!persistData && dataIndex >= 0 || (dataIndex >= 0 && !peeks.contains(
```
lukasz-golebiewski: `(!persistData || !peeks.contains(idx)) && dataIndex >= 0`

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/ReplayRSpace.scala
```diff
@@ -342,7 +354,10 @@ class ReplayRSpace[F[_]: Sync, C, P, A, R, K](
                   .sortBy(_.datumIndex)(Ordering[Int].reverse)
                   .traverse {
                     case DataCandidate(candidateChannel, Datum(_, persistData, _), dataIndex) =>
-                      (if (!persistData && dataIndex >= 0) {
+                      val idx = channelsToIndex(candidateChannel)
+                      (if ((!persistData && dataIndex >= 0) || (dataIndex >= 0 && !peeks.contains(
```
lukasz-golebiewski: I've seen this before. Better extract with a clarifying name

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/StorageActionsTests.scala
```diff
@@ -150,6 +150,58 @@ trait StorageActionsTests[F[_]]
...
+
+      for {
+        r1 <- space.produce(channel, "datum", persist = false)
+        d1 <- store.getData(channel)
+        _  = d1 shouldBe List(Datum.create(channel, "datum", false))
+        c1 <- store.getContinuations(key)
+        _  = c1 shouldBe Nil
+        _  = r1 shouldBe None
+
+        r2            <- space.consume(key, List(Wildcard), new StringsCaptor, persist = false, peeks = Set(0))
```
lukasz-golebiewski: This always tests peek on the first channel. Perhaps it would make sense to write a generative spec in which the peeked channel would be chosen randomly?

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/ReplayRSpaceTests.scala
```diff
@@ -142,6 +142,214 @@ trait ReplayRSpaceTests extends ReplayRSpaceTestsBase[String, Pattern, String, S
...
+        _ = replayResultProduce shouldBe None
+        _ = replayResultConsume shouldBe resultConsume
+        _ = finalPoint.root shouldBe rigPoint.root
+        _ = replaySpace.replayData shouldBe empty
+      } yield ()
+    }
+
+  "Creating COMM Events on many channels with peek" should "replay correctly" in
+    fixture { (store, replayStore, space, replaySpace) =>
+      val channels3 = List("ch1", "ch2")
```
lukasz-golebiewski: Why `channels3`?

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/ReplayRSpaceTests.scala
```diff
@@ -133,7 +133,7 @@ trait ReplayRSpaceTests extends ReplayRSpaceTestsBase[String, Pattern, String, S
 
         replayResultConsume <- replaySpace.consume(channels, patterns, continuation, false)
         replayResultProduce <- replaySpace.produce(channels(0), datum, false)
-        finalPoint          <- space.createCheckpoint()
+        finalPoint          <- replaySpace.createCheckpoint()
```
lukasz-golebiewski: Good catch!

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
dzajkowski: serialization. I'm working on a version that works with sorted sets.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
dzajkowski: I don't mind. My focus was to have small commits that introduce concepts step by step. changing a lot of unrelated files seemed like purposefully making the previous commit less obvious.

rspace/src/main/scala/coop/rchain/rspace/nextgenrspace/RSpace.scala
dzajkowski: on it

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/ReplayRSpaceTests.scala
dzajkowski: huh, reasons ;)... will change

rspace/src/test/scala/coop/rchain/rspace/nextgenrspace/StorageActionsTests.scala
dzajkowski: yep, thinking how to build one

rspace/src/main/scala/coop/rchain/rspace/internal.scala
```diff
@@ -88,12 +93,16 @@ object internal {
   implicit def codecDatum[A](implicit codecA: Codec[A]): Codec[Datum[A]] =
     (codecA :: bool :: Codec[Produce]).as[Datum[A]]
 
+  def sortedSet[A](codecA: Codec[A])(implicit O: Ordering[A]): Codec[SortedSet[A]] =
+    codecSeq[A](codecA).xmap[SortedSet[A]](s => SortedSet(s: _*), _.toSeq.sorted)
```
lukasz-golebiewski: Is it really necessary to do `.sorted`?

```
scala> SortedSet(0,10,5,6,1).toSeq
res6: Seq[Int] = Vector(0, 1, 5, 6, 10)

scala> SortedSet(0,10,5,6,1).toSeq.sorted
res7: Seq[Int] = Vector(0, 1, 5, 6, 10)
```

rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
```diff
@@ -11,6 +11,7 @@ import coop.rchain.rspace.bench.serialization._
```
kayvank: import on line 10 is not required

dzajkowski:
bors try

rspace-bench/src/test/scala/coop/rchain/rspace/bench/serialization/ExamplesSerializerBench.scala
dzajkowski: good catch!
I will do that in a separate one if that's alright with you

rspace/src/main/scala/coop/rchain/rspace/internal.scala
```diff
@@ -36,6 +39,7 @@ object internal {
         patterns: Seq[P],
         continuation: K,
         persist: Boolean,
+        peek: SortedSet[Int],
```
ArturGajowy: I think it'd make sense to have the Channels be a `Seq[(C, Boolean)]`, or a Seq[ChannelRead], where ChannelRead ~ (C, Boolean)

dzajkowski: cost is: we have to encode the tuple
what's the gain?

ArturGajowy: We have the info about the channels on one field vs two. We're explicit in saying whether a given channel is a peek-channel, and don't rely on indexes that could be de-sync from the indexes of the channels.

Speaking of which: are there places where we sort the channels list? If yes, do the peek indices get mapped through the resulting permutation?

dzajkowski: I don't follow. why is it that `patterns` is fine being extracted as a separate parameter (which is dense, every channel has to have a pattern) and peek (which is sparse) has to be part of the channels structure? this seems counter intuitive to me.

we are not sorting the channels as they are basis for the key in looking up a continuation.

ArturGajowy: You have a point that the pattern should also be a part of a `readChannel / bindPattern / channelRead / whatever we call it`, and I :100: agree with you on that ;)

re: sparse vs dense - this matters mostly during serialization, which is (luckily) decoupled from the ADT - at least in rspace.

I'm just suggesting a data representation that avoids indexing and instead aggregates data that constitute one whole in tuples/products, to make invalid states irrepresentable (different number of patterns  than there are channels in the consume / elements in the `peek` set that are not indexes in patterns / source.channels).

I guess though when we add this all this is a big refactoring effort, so let's just ponder it for a while and maybe consider next time. I hope the fields of WaitingContinuation now irk you just as much as they irk me :)

dzajkowski:
bors try

rspace/src/main/scala/coop/rchain/rspace/internal.scala
dzajkowski: ah! so it's a bigger point! I started with that approach (there is a closed pr with that somewhere) and it's a mess.

PR #2521, opened 2019-06-12
Log Rholang output

woky:
```
    Messages sent to stdout (stderr) are currently directly printed into
    stdout (stderr) of node process. This commit also logs them via SLF4J
    logger under category coop.rchain.rholang.stdout (.stderr). This is to
    allow for programmatic contract output processing via Logback structured
    remote appenders[1].
    
    Log level is set to debug in order to not print Rholang output twice by
    default and confuse users. Corresponding *Ack() functions were also
    updated.

    [1] https://github.com/logstash/logstash-logback-encoder/
```

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
@@ -3,6 +3,7 @@ package coop.rchain.rholang.interpreter
 import cats.effect.Concurrent
 import cats.effect.concurrent.Ref
 import cats.implicits._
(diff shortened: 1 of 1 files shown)


woky:
> LGTM as long as we only use it for our internal tests. The users must not attach to that behavior. Ideally, the kibana instance exposing the logged messages would not be accessible for them. Is the above consistent with our intents?

Yeah, I do not want this to be official supported interface. I plan to collect output via Logstash similar to:
- https://github.com/rchain/rchain-testnet-node/blob/dev/logback.xml#L28
- https://github.com/rchain/rchain-logstash

PR #2524, opened 2019-06-13
[WIP] RCHAIN-3505: Implement error handler

JosDenmark:


JosDenmark:
At the behest of @odeac and @ArturGajowy, this is a WIP for the usage of the readers-writer lock described in https://github.com/rchain/rchain/pull/2513 and https://github.com/rchain/rchain/pull/2504.

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala
```diff
@@ -41,28 +39,19 @@ package object accounting extends Costs {
...
-              } yield ()
-          ) { _ =>
-            cost.release
-          }
-      _ <- error.ensure(cost.inspect(identity))(OutOfPhlogistonsError)(_.value >= 0)
-    } yield ()
+  )(implicit cost: _cost[F], bracketF: Bracket[F, Throwable]): F[Unit] =
+    bracketF.bracket(cost.acquire) { _ =>
+      for {
+        phlo0 <- cost.get
```
odeac: probably initialPhlo and finalPhlo would be better names than phlo0 and phlo1

ArturGajowy:
Looks promising, I guess seeing some tests for cancellation in rholang that would fail without those changes would be somewhat convincing :)

I still think though the surface area of the RWLock API is too big. Having all the methods except `withRead/WriteLock` be `private[error]` would scratch that itch.

Also the MLock class should be private. I'm partial on its existence, but as long as the actual usages are integrated as cleanly as in the error handler, and it all does the job for us - I'm fine having it :)

ArturGajowy:
I also think we should close the other two PRs and work here

rholang/src/main/scala/coop/rchain/rholang/interpreter/error/package.scala
```diff
@@ -0,0 +1,27 @@
...
+import cats.{Applicative, Functor, Monad}
+import cats.mtl.{ApplicativeAsk, FunctorTell, MonadState}
+
+package object error {
+
+  type _error[F[_]] = ApplicativeAsk[F, Option[Throwable]] with FunctorTell[F, Throwable]
+
+  def _error[F[_], E](implicit ev: _error[F]): _error[F] = ev
+
+  def errorHandler[F[_]: Monad, E](
```
odeac: where is this used? 

We are adding RWLock and justify it by this function but we don't have a usage of this function. The bottom-up approach can lead to unnecessary constructions. I would like to see top-down how this is used.

JosDenmark:
@ArturGajowy One thing to note is that something like:

```
_error[M].reader {
      case None => _error[M].tell(Throwable)
      case _ => ().pure[F]
}
```

will actually deadlock because the reader lock can only be released when the write lock has been released and the write can only occur after the read lock has been released. We have to do something like

```
_error[M].ask >>= {
      case None => _error[M].tell(Throwable)
      case _ => ().pure[F]
}
```

but I'm just realizing that this won't actually fix the issue at hand. The real problem here is that Par evaluation doesn't behave like a sink. It doesn't wait for all of the threads to complete before returning. It returns before they've completed, which is an issue. I don't see how this would fix that. This lock doesn't enforce that all threads complete before `Reducer.inj` returns.

ArturGajowy:
For the record, after a long chat on discord with @JosDenmark :

the rw-lock approach as discussed above won't work, but for a bit of a different reasons than the one stated above. TL;DR: reductions are recursive, so we must not treat every reduction as a reader. We'll then make the only places that can cause pollution - consume and produce - the readers in the RWLock solution (long term).

My current understanding is that we should:

- put a trivial single-permit lock around 'consume', 'produce' and 'raiseError' in the reducer, and progress with that.
This way we'll:
 - achieve correctness
 - see any other errors there might be lurking
 - have 'error in any par stops the deploy' semantics which is our original goal
 - have the cost accounting short-circuiting work correctly as well (AFAIU, an OOPE error could lead to tuplespace corruption just as well currently)
- in a separate PR later we'll make the lock a reader-writer one, so that only raising errors actually stops the world.

PR #2526, opened 2019-06-14
Warn log unknown keys in the configuration file

sebtomba:
## Overview
Warn log unknown keys in the configuration file

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3551

### DIFF
--- a/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
@@ -4,14 +4,13 @@ import java.nio.file.{Path, Paths}
 
 import collection.JavaConverters._
 import scala.util.Try
+
 import coop.rchain.blockstorage.{BlockDagFileStorage, FileLMDBIndexBlockStore}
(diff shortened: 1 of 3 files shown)


node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -219,6 +234,7 @@ object Configuration {
 
 final case class Configuration(
     underlying: Config,
+    unknownConfigurationKeys: Set[String],
```
lukasz-golebiewski: Wouldn't `unknownKeys` be enough given the fact that this is part of the `Configuration`?

sebtomba: true

sebtomba: Well, now I think `unknownKeys` is not sufficient. I'll rename it to `unknownConfigKeys` since the keys from in the `underlying: Config`.

PR #2528, opened 2019-06-14
Change RevAddress to be a derivative of the eth address 

KentShikama:
## Overview

Also updates genesis vaults to be parsed from wallets.txt 

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3422

### Notes


KentShikama:
bors try

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,66 @@ object Genesis {
...
+          Sync[F].raiseError(new Exception(s"Specified wallets file $walletsPath does not exist"))
+        )
+      case None =>
+        Monad[F].ifM(exists(defaultWalletPath))(
+          Log[F].info(s"Using default file $defaultWalletPath") >> walletFromFile(
+            defaultWalletPath
+          ),
+          Log[F]
+            .warn(
+              "No wallets file specified and no default file found. No wallets will exist at genesis."
```
odeac: is this a warning situation? Isn't this a fatal error? Can RChain work without genesis wallets?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,66 @@ object Genesis {
...
+            )
+            .map(_ => Seq.empty[Vault])
+        )
+    }
+  }
+
+  def fromLine(line: String): Either[String, Vault] = line.split(",").filter(_.nonEmpty) match {
```
odeac: if I understand correctly ```filter(_.nonEmpty)``` will silently accept empty fields in the wallets.txt 

I think this is a case where we should fail instead of silently going forward

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/AddressTools.scala
```diff
@@ -25,12 +26,16 @@ class AddressTools(prefix: Array[Byte], keyLength: Int, checksumLength: Int) {
...
-
-      Some(Address(prefix, keyHash, computeChecksum(payload)))
+      val ethAddress = Base16.encode(Keccak256.hash(pk.bytes.drop(1))).takeRight(40)
+      Some(fromEthAddress(ethAddress))
     } else None
 
+  def fromEthAddress(ethAddress: String): Address = {
+    val keyHash = Keccak256.hash(Base16.unsafeDecode(ethAddress))
+    val payload = prefix ++ keyHash
+    Address(prefix, keyHash, computeChecksum(payload))
```
odeac, line 36: There was a discussion with @cboscolo about the address algorithm and he said he will update the spec wiki page but I don't see any change: https://rchain.atlassian.net/wiki/spaces/CORE/pages/652640311/Wallet+proposal+Note+we+decided+to+rename+terms+such+that+wallet+is+not+used+for+on-chain+components

So, are we keeping the algorithm the same or will we change it? 

The downside of the current algorithm is the readability of the resulting address. My preference is to have a human readable ascii prefix like `revxx` instead of the current one.

If we decide to change, we should do it in a future Jira/PR, of course, but I'm just pointing at the issue

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,66 @@ object Genesis {
...
+      for {
+        maybeLines <- SourceIO.open(walletsPath).use(_.getLines).attempt
+        wallets <- maybeLines match {
+                    case Right(lines) =>
+                      lines
+                        .traverse(fromLine(_) match {
+                          case Right(wallet) => wallet.some.pure[F]
+                          case Left(errMsg) =>
+                            Log[F]
+                              .warn(s"Error in parsing wallets file: $errMsg")
```
odeac: is this a warning situation? Shouldn't it be a critical failure?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,66 @@ object Genesis {
...
+                              .map(_ => none[Vault])
+                        })
+                        .map(_.flatten)
+                    case Left(ex) =>
+                      Log[F]
+                        .warn(
+                          s"Failed to read ${walletsPath.toAbsolutePath} for reason: ${ex.getMessage}"
```
odeac: same here

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: Eventually but as of now the SRE team isn't forced to pass in wallets.txt which is convenient.

KentShikama:
bors try

KentShikama:
@odeac PTAL

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,66 @@ object Genesis {
                      )
     } yield genesisBlock
 
+  def getWallets[F[_]: Sync: Log: RaiseIOError](
+      maybeWalletsPath: Option[String],
+      defaultWalletPath: Path
+  ): F[Seq[Vault]] = {
+    def walletFromFile(walletsPath: Path): F[Seq[Vault]] =
```
lukasz-golebiewski: This function is not a closure so it could be extracted as a standalone private function for better readability

lukasz-golebiewski:
Is there a good reason for keeping the wallets information in a plain text file and parsing it by hand instead of putting it in a lightbend config file and getting parsing for free? https://github.com/lightbend/config

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,67 @@ object Genesis {
...
+                        .traverse(fromLine(_) match {
+                          case Right(wallet) =>
+                            wallet.some.pure[F]
+                          case Left(errMsg) =>
+                            Sync[F].raiseError[Option[Vault]](
+                              new RuntimeException(s"Error in parsing wallets file: $errMsg")
+                            )
+                        })
+                        .map(_.flatten)
+                    case Left(ex) =>
```
odeac: now that you are raising errors, you can use `adaptError` and `fromEither` to simplify the flow:
 ```scala
 def getWallets[F[_]: Sync: Log: RaiseIOError](
      maybeWalletsPath: Option[String],
      defaultWalletPath: Path
  ): F[Seq[Vault]] = {
    def walletFromFile(walletsPath: Path): F[Seq[Vault]] = {
      def parseLine(line : String) : F[Vault] =
        Sync[F].fromEither(
          fromLine(line)
            .leftMap(errMsg => new RuntimeException(s"Error in parsing wallets file: $errMsg")))

      for {
        lines <-
        SourceIO.open(walletsPath)
          .use(_.getLines)
          .adaptError { 
            case ex: Throwable => 
              new RuntimeException(s"Failed to read ${walletsPath.toAbsolutePath} for reason: ${ex.getMessage}") 
          }

        vaults <- lines.traverse(parseLine)
      } yield vaults
    }
```

File: casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala (hunk 1 above)
Comment (odeac) [reply casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala]: @adaszko Is it a real problem to have the wallets.txt mandatory?

File: casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala [hunk 8]
```diff
@@ -103,6 +94,67 @@ object Genesis {
...
+                      Sync[F].raiseError(
+                        new RuntimeException(
+                          s"Failed to read ${walletsPath.toAbsolutePath} for reason: ${ex.getMessage}"
+                        )
+                      )
+                  }
+      } yield wallets
+
+    maybeWalletsPath match {
```
odeac: we should stick with the `vaults` term instead of `wallet`. For future team members it can be confusing to have to use both. Maybe we could also rename `wallets.txt` to `vaults.txt`

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
adaszko: 1) If the `--wallets-file` has been passed and the file isn't usable (i.e. parsing error, the file can't be opened, etc), it's OK — and even desirable — for the node to fail
2) If the `--wallets-file` hasn't been passed and there's not wallets.txt file at the default path, the node shouldn't fail yet, as we would need to make some preparations for that case first, in integration tests, devnet and testnet. /CC @woky 

I've created a ticket for 2): https://rchain.atlassian.net/browse/RCHAIN-3573

The way to update integration tests is to prepare a PR where the `--wallets-file` option is mandatory, run integration tests on it and see what failed.  I can then patch up the problems on that branch and we'll be golden.

odeac: ok. thanks @adaszko for clarifications

KentShikama:
@lukasz-golebiewski There isn't besides the fact that it will take work to refactor.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: I agree eventually we should rename the file to vaults.txt but for now let's not surprise the SRE team.

KentShikama:
bors try

KentShikama:
@odeac @lukasz-golebiewski @ArturGajowy PTAL

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,62 @@ object Genesis {
...
+    Sync[F].fromEither(
+      fromLine(line)
+        .leftMap(errMsg => new RuntimeException(s"Error in parsing vaults file: $errMsg"))
+    )
+
+  private def fromLine(line: String): Either[String, Vault] = line.split(",") match {
+    case Array(ethAddress, initRevBalanceStr, _) =>
+      Try(BigInt(initRevBalanceStr)) match {
+        case Success(initRevBalance) =>
+          Right(Vault(RevAddress.fromEthAddress(ethAddress), initRevBalance.toLong)) // TODO: Remove BigInt conversion
```
lukasz-golebiewski: I don't understand the motivation behind introducing the `BigInt` two lines above and adding a todo note to remove it here. Since you need a `long` at the very end why not do `initRevBalanceStr.toLong`?

KentShikama:
@lukasz-golebiewski PTAL

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,62 @@ object Genesis {
...
+  ): F[Seq[Vault]] =
+    maybeVaultPath match {
+      case Some(vaultsPathStr) =>
+        val vaultsPath = Paths.get(vaultsPathStr)
+        Monad[F].ifM(exists(vaultsPath))(
+          vaultFromFile[F](vaultsPath),
+          Sync[F].raiseError(new Exception(s"Specified vaults file $vaultsPath does not exist"))
+        )
+      case None =>
+        Monad[F].ifM(exists(defaultVaultPath))(
```
ArturGajowy, line 113: What's the purpose behind the defaultVaultPath? How is it set?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,62 @@ object Genesis {
...
+    } yield vaults
+
+  private def parseLine[F[_]: Sync: Log: RaiseIOError](line: String): F[Vault] =
+    Sync[F].fromEither(
+      fromLine(line)
+        .leftMap(errMsg => new RuntimeException(s"Error in parsing vaults file: $errMsg"))
+    )
+
+  private def fromLine(line: String): Either[String, Vault] = line.split(",") match {
+    case Array(ethAddress, initRevBalanceStr, _) =>
```
ArturGajowy, line 146: what's on the 3rd coordinate? Why do we ignore it?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -103,6 +94,62 @@ object Genesis {
                      )
     } yield genesisBlock
 
+  private def getVaults[F[_]: Sync: Log: RaiseIOError](
```
ArturGajowy, line 97: We should not be using sync here. raiseError is available on `FunctorRaise`

ArturGajowy: We should probably also have some custom exception types (at least one - ConfigurationError?) for the error cases we have here

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: I think by default it is `~/.rnode/genesis/wallets.txt`. @sebtomba might know more.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: I don't think there is anything in the 3rd+ coordinate.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: Hmm; do you mean I should use `raiseIOErrorThroughSync`?

itegulov: I think @ArturGajowy means that since you have `RaiseIOError`, your can use `RaiseIOError[F].raise(...)` instead of `Sync[F].raiseError(...)`. And get rid of dependency on `Sync` entirely.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
itegulov: `case Array(a, b, _)` means that you are matching three coordinates. In https://files.rchain-dev.tk/wallets_all.txt I can see that there are some 0s and 1s in the third coordinate. Do you know what do they stand for?

KentShikama: Now I remember: 0 or 1 corresponds to whether that address is a contract. We need to handle those that are contracts in a special fashion - we probably need to have a larger discussion for this but any adjustments that result should go in a future PR.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
ArturGajowy: My point is that we don't need two paths from which we'll read the wallets file. The config option could have default value, but if not - we go with what the user passed. Notice this would happen in the config code, not here.

Then we read the file from whatever the config value resolved to (default or not).

Then we terminate if the file is not there.

KentShikama: I've created a ticket for this: https://rchain.atlassian.net/browse/RCHAIN-3590

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: How should I treat

```scala
      lines <- SourceIO
                .open(vaultsPath)
                .use(_.getLines)
                .adaptError {
                  case ex: Throwable =>
                    new RuntimeException(
                      s"Failed to read ${vaultsPath.toAbsolutePath} for reason: ${ex.getMessage}"
                    )
                }
```
?

KentShikama: Also

```scala
  private def parseLine[F[_]: Sync: Log: RaiseIOError](line: String): F[Vault] =
    Sync[F].fromEither(
      fromLine(line)
        .leftMap(errMsg => new RuntimeException(s"Error in parsing vaults file: $errMsg"))
    )
```

KentShikama: @itegulov Wouldn't I still need `Sync` for `ifM`?

itegulov: Yeah, seems like you need at least `ApplicativeError` here for what you are trying to do. `FunctorRaise` is not enough. Maybe try to change `RaiseIOError`'s type to `ApplicativeError`? If it is not as simple/does not make sense I would just stick with `Sync` here...

Regarding `ifM`: you only need `Monad` for it which far less powerful than `Sync`.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
ArturGajowy: Please create a JIRA issue describing why they need special treatment, so that we don't forget to (desing) and implement it

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
ArturGajowy: The Sync is only needed, because we don't yet have a RaiseCasperError type that could inject errors from downstream modules into its own error type. There's not even the error type, as signified by the `new RuntimeException`. It seems to me that we should discuss an approach to error handling at some time in the future, and do small steps refactoring towards it while we progress with PRs.

ArturGajowy:
Forgot to mention: the fact that we have parsing logic of a data type mixed with the data types logic and internals (IOW: everything lives in the Genesis.scala file) should also be revisited at some time. @lukasz-golebiewski's proposal to lean on the config mechanism for that is one possible solution, another would be having a separate 'GenesisParser / GenesisConfigReader / GenesisConfigRepository / what have you` service (TF trait + impl) for that.

KentShikama:
@ArturGajowy Ticket for "isContract" flag/bit: https://rchain.atlassian.net/browse/RCHAIN-3592

PR #2529, opened 2019-06-17
Get tmp dir from env

zsluedem:
## Overview
I need this to make tests running right in windows system. I believe it doesn't hurt the current implementation.

### JIRA ticket:
No

### Notes

### DIFF
--- a/integration-tests/test/common.py
+++ b/integration-tests/test/common.py
@@ -17,7 +17,6 @@ if TYPE_CHECKING:
     from .wait import PredicateProtocol
 
 
-
 @dataclasses.dataclass(eq=True, frozen=True)
 class KeyPair:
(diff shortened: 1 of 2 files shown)


adaszko:
Is this needed at all?

https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp
> If dir is not None, the file will be created in that directory; otherwise, a default directory is used. The default directory is chosen from a platform-dependent list, but the user of the application can control the directory location by setting the TMPDIR, TEMP or TMP environment variables. There is thus no guarantee that the generated filename will have any nice properties, such as not requiring quoting when passed to external commands via os.popen().

zsluedem:
@adaszko You are right. I just remove all the `dir` args

PR #2479, opened 2019-05-31
Charging for deploy 5

ArturGajowy:
## Overview
<sup>_What this PR does, and why it's needed_</sup>

This is a work in progress.
Some changes are @JosDenmark's, will do a rebase later to add co-authorship info to commits.

### JIRA ticket:
<sup>_Create it if there isn't one already._</sup>

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2535, opened 2019-06-21
[WIP] Charging for deploy 5

JosDenmark:
## Overview
@ArturGajowy , @KentShikama asked me to open a PR with the changes that you and I made to your original WIP.

All but 13 tests are working:
- the last commit shows some work that I did on tracking down the bug causing the test `"Blocks" should "conflict if they use the same deploys in different histories"` to fail with the error `replayData multimap has 158 elements left`. This is occurring because deploy payments aren't succeeding in replay evaluation. More specifically, the registry lookup of `PoS` on which they depend are not succeeding in replay. As I said, I left some notes.
- the errors that say `Expected response message was not received` means that the test is attempting to process a deploy with a payment when the genesis block has not been correctly constructed. The fix probably involves some usage of `TestUtil.defaultGenesisSetup`.
- the errors that say `installing can only be done on startup` and `BlockStore is missing hash f52bc80e0a`, I have yet to investigate.

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -30,7 +30,8 @@ new  PoS,
     runMVar,
     deposit,
     rs(`rho:registry:insertSigned:secp256k1`),
-    uriOut in {
+    uriOut,
+    stdout(`rho:io:stdout`) in {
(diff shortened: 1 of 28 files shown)


ArturGajowy:
Let's continue in https://github.com/rchain/rchain/pull/2479

PR #2537, opened 2019-06-21
RCHAIN-3424: Make RevVault use the on-chain deployer ID

tymm:
## Overview
Not removing the old version of `deployerAuthKey` yet since it is still needed for the PoS contract and removing it from the PoS contract is another ticket.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3424

### Notes

### DIFF
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -33,6 +33,17 @@ in {
       @MakeMint!(*mintCh) |
       for (mint <- mintCh) {
 
+        contract RevVault(@"deployerAuthKey", deployerId, ret) = {
+
+          new RevAddress(`rho:rev:address`), revAddrCh in {
+            RevAddress!("fromDeployerId", *deployerId, *revAddrCh)|
+            for (@deployerRevAddress <- revAddrCh) {
+
+              @AuthKey!("make", (*_revVault, deployerRevAddress), *ret)
+            }
+          }
+        } |
+
         contract RevVault(@"deployerAuthKey", ret) = {
 
           new deployerPubKeyBytesCh in {
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -35,11 +35,15 @@ in {
 
         contract RevVault(@"deployerAuthKey", deployerId, ret) = {
 
-          new RevAddress(`rho:rev:address`), revAddrCh in {
-            RevAddress!("fromDeployerId", *deployerId, *revAddrCh)|
-            for (@deployerRevAddress <- revAddrCh) {
+          new RevAddress(`rho:rev:address`), DeployerIdOps(`rho:rchain:deployerId:ops`), revAddrCh, deployerPubKeyBytesCh in {
+            DeployerIdOps!("pubKeyBytes", *deployerId, *deployerPubKeyBytesCh) |
+            for (@deployerPubKeyBytes <- deployerPubKeyBytesCh) {
 
-              @AuthKey!("make", (*_revVault, deployerRevAddress), *ret)
+              RevAddress!("fromPublicKey", deployerPubKeyBytes, *revAddrCh)|
+              for (@deployerRevAddress <- revAddrCh) {
+
+                @AuthKey!("make", (*_revVault, deployerRevAddress), *ret)
+              }
             }
           }
         } |
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -44,6 +44,8 @@ in {
           }
         } |
 
+        // Deprecated version of `deployerAuthKey`
+        // Please use the version above where a deployer id has to be passed
         contract RevVault(@"deployerAuthKey", ret) = {
 
           new deployerPubKeyBytesCh in {
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
@@ -1,12 +1,10 @@
 package coop.rchain.casper.genesis.contracts
 
+import cats.implicits._
 import coop.rchain.crypto.codec.Base16
-import coop.rchain.models.Par
-import coop.rchain.rholang.build.{CompiledRholangSource, CompiledRholangTemplate}
-import coop.rchain.rholang.interpreter.ParBuilder
-import monix.eval.Coeval
-
-import scala.io.Source
+import coop.rchain.crypto.signatures.Secp256k1
+import coop.rchain.rholang.build.CompiledRholangTemplate
+import coop.rchain.rholang.interpreter.NormalizerEnv
 
 // TODO: Eliminate validators argument if unnecessary.
 // TODO: eliminate the default for epochLength. Now it is used in order to minimise the impact of adding this parameter
@@ -17,6 +15,7 @@ final case class ProofOfStake(
     epochLength: Int = 10000
 ) extends CompiledRholangTemplate(
       "PoS.rhox",
+      NormalizerEnv(none, ProofOfStake.pk.some),
       "minimumBond"  -> minimumBond,
       "maximumBond"  -> maximumBond,
       "initialBonds" -> ProofOfStake.initialBonds(validators),
@@ -30,6 +29,7 @@ final case class ProofOfStake(
 }
 
 object ProofOfStake {
+  val (_, pk) = Secp256k1.newKeyPair
 
   // TODO: Determine how the "intial bonds" map can simulate transferring stake into the PoS contract
   //       when this must be done during genesis, under the authority of the genesisPk, which calls the
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
@@ -1,8 +1,6 @@
 package coop.rchain.casper.genesis.contracts
 
-import cats.implicits._
 import coop.rchain.crypto.codec.Base16
-import coop.rchain.crypto.signatures.Secp256k1
 import coop.rchain.rholang.build.CompiledRholangTemplate
 import coop.rchain.rholang.interpreter.NormalizerEnv
 
@@ -15,7 +13,7 @@ final case class ProofOfStake(
     epochLength: Int = 10000
 ) extends CompiledRholangTemplate(
       "PoS.rhox",
-      NormalizerEnv(none, ProofOfStake.pk.some),
+      NormalizerEnv.Empty,
       "minimumBond"  -> minimumBond,
       "maximumBond"  -> maximumBond,
       "initialBonds" -> ProofOfStake.initialBonds(validators),
@@ -29,8 +27,6 @@ final case class ProofOfStake(
 }
 
 object ProofOfStake {
-  val (_, pk) = Secp256k1.newKeyPair
-
   // TODO: Determine how the "intial bonds" map can simulate transferring stake into the PoS contract
   //       when this must be done during genesis, under the authority of the genesisPk, which calls the
   //       linear receive in PoS.rho
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
@@ -42,7 +42,11 @@ final case class RevGenerator(
        # }
      """.stripMargin('#')
 
-  val term: Par = ParBuilder[Coeval].buildNormalizedTerm(code, NormalizerEnv(deployId = None, deployerPk = genesisPk.some)).value()
+  val normalizerEnv: NormalizerEnv = NormalizerEnv(deployId = none, deployerPk = genesisPk.some)
+
+  val term: Par = ParBuilder[Coeval]
+    .buildNormalizedTerm(code, normalizerEnv)
+    .value()
 
   private def findOrCreate(userVault: Vault): String =
     s""" 
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
@@ -1,14 +1,19 @@
 package coop.rchain.casper.genesis.contracts
 
+import cats.implicits._
+import coop.rchain.crypto.PublicKey
 import coop.rchain.models.Par
 import coop.rchain.rholang.build.CompiledRholangSource
 import coop.rchain.rholang.interpreter.{NormalizerEnv, ParBuilder}
 import coop.rchain.rholang.interpreter.util.RevAddress
-import coop.rchain.rholang.interpreter.util.codec.Base58
 import monix.eval.Coeval
 
-final case class RevGenerator(genesisAddress: RevAddress, userVaults: Seq[Vault], supply: Long)
-    extends CompiledRholangSource {
+final case class RevGenerator(
+    genesisPk: PublicKey,
+    genesisAddress: RevAddress,
+    userVaults: Seq[Vault],
+    supply: Long
+) extends CompiledRholangSource {
 
   val path: String = "<synthetic in Rev.scala>"
 
@@ -24,8 +29,8 @@ final case class RevGenerator(genesisAddress: RevAddress, userVaults: Seq[Vault]
        #         *genesisVaultCh
        #       )
        #       | for (@(true, genesisVault) <- genesisVaultCh) {
-       #         new genesisAuthKeyCh in {
-       #           @RevVault!("deployerAuthKey", *genesisAuthKeyCh)
+       #         new genesisAuthKeyCh, deployerId(`rho:rchain:deployerId`) in {
+       #           @RevVault!("deployerAuthKey", *deployerId, *genesisAuthKeyCh)
        #           | for (genesisVaultAuthKey <- genesisAuthKeyCh) {
        #             ${concatenate(findOrCreate)} |
        #             ${concatenate(transfer)}
@@ -37,7 +42,7 @@ final case class RevGenerator(genesisAddress: RevAddress, userVaults: Seq[Vault]
        # }
      """.stripMargin('#')
 
-  val term: Par = ParBuilder[Coeval].buildNormalizedTerm(code, NormalizerEnv.Empty).value()
+  val term: Par = ParBuilder[Coeval].buildNormalizedTerm(code, NormalizerEnv(deployId = None, deployerPk = genesisPk.some)).value()
 
   private def findOrCreate(userVault: Vault): String =
     s""" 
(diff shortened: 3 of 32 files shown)


tymm:
bors try

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
```diff
@@ -34,8 +35,11 @@ object CompiledRholangSource {
   * @param env a sequence of pairs macro -> value
   * @return
   */
-abstract class CompiledRholangTemplate(classpath: String, env: (String, Any)*)
-    extends CompiledRholangSource {
+abstract class CompiledRholangTemplate(
+    classpath: String,
+    deployerPk: Option[PublicKey],
```
odeac: Why is this an option? Can't you make it mandatory?

tymm: Please take a look at which `CompiledRholangSource`s actually take some public key. It is just one out of many.
And I think it is valuable to make it explicit which rholang needs the public key and which doesn't for at least debugging reasons.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/AuthKeySpec.scala
```diff
@@ -5,7 +5,8 @@ import coop.rchain.rholang.build.CompiledRholangSource
 
 class AuthKeySpec
     extends RhoSpec(
-      CompiledRholangSource("AuthKeyTest.rho"),
+      CompiledRholangSource("AuthKeyTest.rho", None),
       Seq.empty,
+      None,
```
ArturGajowy: I think we have a functional dependency between RhoSpec's paramaters, meaning we effectively always pass the same value of deployerPk twice. We should remove the 3rd parameter from RhoSpec and obtain the deployerPk from the passed CompiledRholangSource

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
```diff
@@ -30,20 +26,21 @@ object TestUtil {
...
+      otherLibs: Seq[DeployData],
+      deployerPk: Option[PublicKey]
   ): F[Runtime[F]] =
     for {
       runtimeManager <- RuntimeManager.fromRuntime(runtime)
       _              <- genesisSetup(runtimeManager)
-      _              <- evalDeploy(rhoSpecDeploy, runtime)
-      _              <- otherLibs.toList.traverse(evalDeploy(_, runtime))
+      _              <- evalDeploy(rhoSpecDeploy, runtime, deployerPk)
+      _              <- otherLibs.toList.traverse(evalDeploy(_, runtime, deployerPk))
```
ArturGajowy: for otherLibs we should be using the deployerPk from the DeployData that represents a given otherLib. I think otherLibs are never used though, so it might make sense to remove the parameter whatsoever.

casper/src/main/resources/RevVault.rho
```diff
@@ -33,6 +33,17 @@ in {
       @MakeMint!(*mintCh) |
       for (mint <- mintCh) {
 
+        contract RevVault(@"deployerAuthKey", deployerId, ret) = {
+
+          new RevAddress(`rho:rev:address`), revAddrCh in {
+            RevAddress!("fromDeployerId", *deployerId, *revAddrCh)|
```
ArturGajowy: Ideally, we'd obtain `pubKeyBytes` from a `deployerId`, and use the preexisting `RevAddress!fromPubKeyBytes` method. ~~I think you even have~~ sth. like `DeployerId!pubKeyBytes` ~~in a latter commit as a test contract~~ - do you see a problem having it generally available?

casper/src/main/resources/RevVault.rho
```diff
@@ -33,6 +33,17 @@ in {
...
+            for (@deployerRevAddress <- revAddrCh) {
+
+              @AuthKey!("make", (*_revVault, deployerRevAddress), *ret)
```
ArturGajowy: (this is kind of a note to self, but maybe you'll have some ideas?)

I'm a bit torn on using revAddress as part of the AuthKey.shape, but at the same time `findOrCreate` only receives a revAddress...

In my ideal world, AuthKey creation can be moved out of RevVault and be decoupled from it - the users would create an AuthKey of an agreed shape, based on a publicly available, bundled _revVault unf name (that encodes the contract we're creating an AuthKey for). 

That isn't safe though, if the shape is just `(bundle0{_revVault}, revAddressString)` (everybody has the rev address).

One direction for plugging that would be a `shape` of `(_deployerAuthKeyMarker, bundle0{_revVault}, revAddressString)`, where a key with a shape containing `_deployerAuthKeyMarker` would only be obtainable from a safe `AuthKey!deployersKey(deployerId, contractMarker, payload)`. This doesn't align seamlessly with unf-name based auth-keys though, as RevVault no longer knows the whole shape and would need to differentiate how to check the AuthKey (for deployerId vs unf-name based vaults).

Another approach would be somehow changing the internal account token from `revAddressString` to `bundle0{ deployerId }` / `bundle0 { unfName }`. The bundle is needed for RevVault to never know the `deployerId`, but at the same time it makes 3rd-party RevVault creation impossible... The corresponding revAddresses would be derived from deployerId / unfName, except not possible because a bundle wraps them....

It seems as though we need to leave `RevVault!deployerAuthKey` coupled with RevVault for now, and base our trust on it doing the right thing...

casper/src/test/resources/RevVaultTest.rho
```diff
@@ -22,6 +22,7 @@ match (
       ListOpsCh,
       stdout(`rho:io:stdout`),
       setDeployData(`rho:test:deploy:set`),
+      getDeployerId(`rho:test:deployerId:get`),
```
ArturGajowy, line 25: Nice one! :)

casper/src/test/scala/coop/rchain/casper/helper/DeployerIdContract.scala
```diff
@@ -0,0 +1,27 @@
+package coop.rchain.casper.helper
+import cats.effect.Concurrent
+import coop.rchain.models.rholang.implicits._
+import coop.rchain.models.{GDeployerId, ListParWithRandom, Par}
+import coop.rchain.rholang.interpreter.Runtime.SystemProcess
+import coop.rchain.rholang.interpreter.{ContractCall, RhoType}
+import coop.rchain.shared.ByteStringOps._
+
+object DeployerIdContract {
```
ArturGajowy, line 10: We should probably add a fat warning saying this should under no circumstances be available in production :)

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -24,6 +22,7 @@ final case class ProofOfStake(minimumBond: Long, maximumBond: Long, validators:
 }
 
 object ProofOfStake {
+  val (_, pk) = Secp256k1.newKeyPair
```
ArturGajowy: This is going to differ across different node runs. We must hardcode it.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
```diff
@@ -77,19 +74,21 @@ object TestUtil {
 
   private def evalDeploy[F[_]: Sync](
       deploy: DeployData,
-      runtime: Runtime[F]
+      runtime: Runtime[F],
+      deployerPk: Option[PublicKey]
```
ArturGajowy: There's no need for `deployerPk` argument, and in fact there's a need for there to not be one :) It should be taken from `deploy: DeployData`

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -57,6 +58,7 @@ object RhoSpec {
   def getResults(
       testObject: CompiledRholangSource,
       otherLibs: Seq[DeployData],
+      deployerPk: Option[PublicKey],
```
ArturGajowy: No need for this paramter, it's in testObject

integration-tests/resources/wallets/transfer_funds.rho
```diff
@@ -16,9 +16,9 @@ in {
     ) {
       (from, to, amount) => {
 
-        new vaultCh, revVaultkeyCh in {
```
ArturGajowy, line 19: Interesting! I wasn't aware this made it to integration tests :D

casper/src/main/resources/RevVault.rho
```diff
@@ -33,6 +33,17 @@ in {
...
+            }
+          }
+        } |
+
         contract RevVault(@"deployerAuthKey", ret) = {
```
ArturGajowy, line 53: Well done on migrating all the usages of this method but the one in PoS! :tada: Please add a fat "deprecated, don't use - use the other one instead" comment on this method :)

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
ArturGajowy: We should use the PK corresponding to the genesis deploy, or make that one refer to this one

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
odeac: I'm thinking all the Rholang deployed should need a deployer identity but then we would have a few predefined ones like genesis identity and tester identity.

This would help us get rid of the Option above and will simplify the logic

ArturGajowy: That's a good point! I'd leave this for a separate PR though. It seems that CompliedRholangSource has less and less sense on its own, and needs to pull in basicaly the whole DeployData? I'd like to see how this evolves.

casper/src/main/resources/RevVault.rho
ArturGajowy: I think this can be solved by making AuthKey create sth. like a pair of (priv, pub)-key - each of them would be a contract, pub would be used to verify the priv. We could have different methods for obtaining (parts of) the keypair (deployerId-based, unf-name-based) and only the public part of the keyparir would be needed upon creation.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/AuthKeySpec.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/helper/DeployerIdContract.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
tymm: Done

casper/src/main/resources/RevVault.rho
tymm: Done assuming you are fine with the system process living under `rho:rchain:deployerId:ops` and not overloading `rho:rchain:deployerId` for that.

casper/src/main/resources/RevVault.rho
tymm: Done

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
tymm: I still think having defaults could make it harder to debug rholang where deployer/deploy ids are involved. Happy to create a ticket though if I am overruled :) Should I?

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
tymm: There are at least two places (`Genesis.scala` and `BlockApproverProtocol.scala`) where we generate the genesis pk in this way.
Therefore I'm not sure if hardcoding it just here or passing it from the genesis deploy would help much.
How about adding a comment and creating a ticket where we solve the problem in its entirety?

casper/src/main/resources/RevVault.rho
tymm: >I think this can be solved by making AuthKey create sth. like a pair of (priv, pub)-key - each of them would be a contract, pub would be used to verify the priv. We could have different methods for obtaining (parts of) the keypair (deployerId-based, unf-name-based) and only the public part of the keyparir would be needed upon creation.

Sounds like an interesting idea.
This would however force the user to use public keys instead of rev addresses for `findOrCreate`. Not sure if this is what we want.

tymm:
bors try

casper/src/main/resources/RevVault.rho
ArturGajowy: it's "pub key-like contract from AuthKey", it doesn't have to use anything else than the provided shape (which is arbitrary), and an AuthKey-private token to work. The private counterpart would require passing deployId + shape for construction, and would respond with `bundle0{(_authKey, shape)}` when `challenge`d by the pub part :)

ArturGajowy: There would be many variants of the `_authKey` token, to allow AuthKey-pairs derived from at least:
 - deployId
 - whatever unforgable name

tymm: But how does the *create*-part of `findOrCreate` would work with this when the user that wants to create the rev vault is not the owner? The user would have to know about the "pub key-like contract from AuthKey" of the owner, correct? That might be a problem from a UX perspective?

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
odeac: I'm ok with a separate PR for this

casper/src/test/scala/coop/rchain/casper/helper/DeployerIdContract.scala
```diff
@@ -0,0 +1,30 @@
...
+  )(message: (Seq[ListParWithRandom], Int)): F[Unit] = {
+
+    val isContractCall = new ContractCall(ctx.space, ctx.dispatcher)
+    message match {
+      case isContractCall(
+          produce,
+          Seq(RhoType.String("deployerId"), RhoType.ByteArray(pk), ackCh)
+          ) =>
+        for {
+          _ <- produce(Seq(GDeployerId(pk.toByteString): Par), ackCh)
```
odeac: You could use Rhotype.DeployerId.apply here

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
```diff
@@ -60,4 +61,13 @@ object RhoType {
 
     def apply(s: String): Par = GUri(s)
   }
+
+  object GDeployerId {
```
odeac: In order to be consistent with the other RhoType names this should be called `DeployerId`

casper/src/main/resources/RevVault.rho
ArturGajowy: The find-or-create method would create the public key. Creating the public authkey is safe ocap-leak-wise. It is the private part that is sensitive - we don't want to be passing the deployerId to too many places.

The above discussion's goal is to make the AuthKey!create*Private methods be the only places one has to pass such a high-grade auth ocap and obtain an attenuated, purpose specific ocap that can be checked using the corresponding public part, without the risk of leaking the high-grade ocap.

rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
```diff
@@ -54,5 +55,6 @@ abstract class CompiledRholangTemplate(
         #$finalContent
         #""".stripMargin('#')
 
-  override val term: Par = ParBuilder[Coeval].buildNormalizedTerm(code, normalizerEnv).value()
+  override val term: Par                    = ParBuilder[Coeval].buildNormalizedTerm(code, normalizerEnv0).value()
+  override val normalizerEnv: NormalizerEnv = normalizerEnv0
```
ArturGajowy, line 59: normalizerEnv contains the deployId, which is a signature over the deployed code, which is a parameter of the compiledRholangTemplate....

I think it might make sense to no longer inherit RholangSource in the Template, and instead make the Template a (:drum:) factory, that will produce Source when provided with a NormalizerEnv.

ArturGajowy: nvm the above, since in both the Source and the Template case we don't have means to verify the passed Env is correct (no access to deploy timestamp).

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
ArturGajowy: I think this pk is currently unused (we don't use deployerId in PoS.rhox). And we shouldn't be, really. The PoS vault should be an unf-name based vault, with a multisig wrapper over it - tho PoS would use the unf-name based one though. Please make the corresponding RholangSource/Template use EmptyEnv

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -24,8 +29,8 @@ final case class RevGenerator(genesisAddress: RevAddress, userVaults: Seq[Vault]
        #         *genesisVaultCh
        #       )
        #       | for (@(true, genesisVault) <- genesisVaultCh) {
-       #         new genesisAuthKeyCh in {
-       #           @RevVault!("deployerAuthKey", *genesisAuthKeyCh)
+       #         new genesisAuthKeyCh, deployerId(`rho:rchain:deployerId`) in {
```
ArturGajowy, line 32: this too should be migrated to an unf-name based genesis vault (the only one having rev supply upon its creation). So there's no need for genesisPk in RevGenerator. Not until we have the unf-name based vaults though. Can you link this comment to the relevant jira issue?

tymm: Done: https://rchain.atlassian.net/browse/RCHAIN-3617

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
tymm: Done

tymm:
bors try

PR #2545, opened 2019-06-25
RCHAIN-3530 drop if message notneeded

rabbitonweb:
## Overview
We are dropping changes that we were hoping to provide with RCHAIN-3530 and will apply ideas that were discussed on two last sprints (RCHAIN-3582), but the refactorings I did for RCHAIN-3530 are still worth having in dev.

In this PR we are modifing the `CircuitBreaker`. Instead of returning `Boolean` it returns ADT `Closed` or `Open(error)`. This way all functionality that was breaking streams before was now moved to the `CircuitBreaker`. 
Additional some classes were extracted (like `Header`) and few other minor improvements

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3530

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
@@ -56,8 +56,11 @@ object GrpcTransportReceiver {
 
           val circuitBreaker: StreamHandler.CircuitBreaker = {
             case streamed =>
-              streamed.header.map(_.networkId != networkId).getOrElse(false) ||
-                streamed.readSoFar > maxStreamMessageSize
+              if (streamed.header.map(_.networkId != networkId).getOrElse(false))
+                Broken(StreamHandler.StreamError.wrongNetworkId)
+              else if (streamed.readSoFar > maxStreamMessageSize)
+                Broken(StreamHandler.StreamError.circuitBroken)
+              else Closed
           }
 
           (handleStream(tempFolder, observable, circuitBreaker) >>= {
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
@@ -54,7 +54,9 @@ object GrpcTransportReceiver {
           import StreamHandler._
           import StreamError.StreamErrorToMessage
 
-          val circuitBreaker: StreamHandler.CircuitBreaker = _ > maxStreamMessageSize
+          val circuitBreaker: StreamHandler.CircuitBreaker = {
+            case streamed => streamed.readSoFar > maxStreamMessageSize
+          }
 
           (handleStream(networkId, tempFolder, observable, circuitBreaker) >>= {
             case Left(error @ StreamError.Unexpected(t)) => logger.error(error.message, t)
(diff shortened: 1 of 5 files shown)


comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -102,37 +114,40 @@ object StreamHandler {
...
+                .Header(ChunkHeader(Some(sender), typeId, compressed, cl, nid))
             )
-          } else
-            Right(stmd.copy(wrongNetwork = true))
+            ) =>
+          val newStmd = stmd.copy(
+            header = Some(Header(ProtocolHelper.toPeerNode(sender), typeId, cl, nid, compressed))
+          )
+          val circuit = circuitBreaker(newStmd)
+          if (circuit.broken) Right(newStmd.copy(circuit = circuit))
```
dzajkowski, line 126: tripped

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
```diff
@@ -54,7 +54,9 @@ object GrpcTransportReceiver {
           import StreamHandler._
           import StreamError.StreamErrorToMessage
 
-          val circuitBreaker: StreamHandler.CircuitBreaker = _ > maxStreamMessageSize
+          val circuitBreaker: StreamHandler.CircuitBreaker = {
+            case streamed => streamed.readSoFar > maxStreamMessageSize
```
ArturGajowy: I think there's no need for `case` here

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -120,10 +120,10 @@ object StreamHandler {
           stmd.fos.write(array)
           stmd.fos.flush()
           val readSoFar = stmd.readSoFar + array.length
-          if (circuitBreaker(readSoFar))
-            Right(stmd.copy(circuitBroken = true))
-          else
-            Left(stmd.copy(readSoFar = readSoFar))
+          val newStmd   = stmd.copy(readSoFar = readSoFar)
+          if (circuitBreaker(newStmd))
+            Right(newStmd.copy(circuitBroken = true))
```
ArturGajowy: this looks like a logic change - previously the Right variant didn't have readSoFar updated. should a test cahnge / be added?

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -104,17 +103,16 @@ object StreamHandler {
...
-          } else
-            Right(stmd.copy(wrongNetwork = true))
+          val newStmd = stmd.copy(
+            sender = sender.map(ProtocolHelper.toPeerNode),
+            typeId = Some(typeId),
+            contentLength = Some(cl),
+            networkId = Some(nid),
+            compressed = compressed
+          )
+          if (circuitBreaker(newStmd)) Right(newStmd.copy(circuitBroken = true))
```
ArturGajowy: shouldn't that if be inside a method of CircuitBreaker?

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
EncodePanda: ok

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
EncodePanda: because if it was `opened/broken` there was no point, but this is correct i guess: we open circuit because we've read to much data. number of data read is not passed to error message, maybe it should?

EncodePanda:
bors try

PR #2555, opened 2019-06-26
RCHAIN-3172: Add support for byte array interpolation.

JosDenmark:
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3172

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -20,7 +20,6 @@ import coop.rchain.rholang.interpreter.matcher.SpatialMatcher.spatialMatchAndCha
 import coop.rchain.rholang.interpreter.storage.Tuplespace
 import coop.rchain.rspace.Serialize
 import monix.eval.Coeval
-import cats.effect.concurrent.Ref
 
 import scala.collection.immutable.BitSet
 import scala.util.Try
@@ -651,8 +650,8 @@ class DebruijnInterpreter[M[_], F[_]](
               (keyString -> valueBool.toString).pure[M]
             case (GString(keyString), GUri(uri)) =>
               (keyString -> uri).pure[M]
-            // TODO: Add cases for other ground terms as well? Maybe it would be better
-            // to implement cats.Show for all ground terms.
+            case (GString(keyString), GByteArray(byteString)) =>
+              (keyString -> Base16.encode(byteString.toByteArray)).pure[M]
             case (_: GString, value) =>
               ReduceError(s"Error: interpolation doesn't support ${value.typ}")
                 .raiseError[M, (String, String)]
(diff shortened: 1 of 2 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1475,6 +1473,26 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+          interpolate(
+            "${key}",
+            List[(Par, Par)](
+              (GString("key"), GByteArray(ByteString.copyFrom(Base16.unsafeDecode("012345"))))
+            )
+          )
+        )
+        Await.result(inspectTask.runToFuture, 3.seconds)
+    }
+    result.exprs should be(Seq(Expr(GString("012345"))))
```
odeac: Can you use RhoType.String here for improved readability?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -651,8 +650,8 @@ class DebruijnInterpreter[M[_], F[_]](
               (keyString -> valueBool.toString).pure[M]
             case (GString(keyString), GUri(uri)) =>
               (keyString -> uri).pure[M]
-            // TODO: Add cases for other ground terms as well? Maybe it would be better
-            // to implement cats.Show for all ground terms.
+            case (GString(keyString), GByteArray(byteString)) =>
+              (keyString -> Base16.encode(byteString.toByteArray)).pure[M]
```
ArturGajowy: I'm a bit partial on having a default format of serializing a byte array to string. Especially that the format is Base16. I'm afraid that this is going to be (ab)used by end users leading to gross inefficiencies. OTOH, it's more rev for the validators, right? :thinking:

Why do we need this?

JosDenmark: > I’m a bit partial on having a default format of serializing a byte array to string

Honestly, I feel the same way. The only reason it’s Base16 is because that’s Rholang’s default for _serializing_ data in the first place.

> Why do we need this?

It’s a ticket that’s assigned to me and I haven’t issued a PR in a while :) No, in seriousness, @odeac would have a better idea than me. He reported the ticket.

odeac: Before this the byte array  interpolation was failing silently, which was a bit unexpected. We should support byte array interpolation or, at least, fail with a big fat error saying that we don't support interpolation for this data type.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: This seems to work
```scala
result should be(RhoType.String("012345"))```

Comment (odeac) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala:1492]: This seems to work
```scala
result should be(RhoType.String("012345"))
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: @JosDenmark is working on the failures no longer being silent. Let's close this then.

PR #2557, opened 2019-06-27
Rchain 3464: distribute rewards at closeBlock

odeac:
## Overview
With this PR the rewards are kept in a separate data structure. Only when the block is closed they are moved to the rewards map and become visible from the outside

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3464

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -26,7 +26,6 @@ new  PoS,
     getUser,
     getCurrentUserAddress,
     getCurrentUserVault,
-    initialRewardsCh, computeInitialRewards,
     getMVar,
     runMVar,
     deposit, distributeRewards,
@@ -43,20 +42,14 @@ new  PoS,
       for(@posRevAddress <- posRevAddressCh) {
         new stateCh, posVaultCh in {
 
-          @ListOps!("fold", $$initialBonds$$.toList(), {}, *computeInitialRewards, *initialRewardsCh) |
-          contract computeInitialRewards(@(pk, _), @acc, resultCh) = {
-            resultCh!(acc.set(pk, 0))
-          } |
-
           @RevVault!("findOrCreate", posRevAddress, *posVaultCh) |
-          for (@(true, _) <- posVaultCh;
-               @initialRewards <- initialRewardsCh ) {
+          for (@(true, _) <- posVaultCh ) {
 
             // State structure:
             // pendingRewards : Map[PublickKey, Int] - are accummulated by calling "pay"
             // committedRewards : Map[PublickKey, Int] - are moved from pendingRewards at each closeBlock
             // bonds : Map[PublickKey, Int] - each validator stake
-            stateCh!(({}, initialRewards, $$initialBonds$$)) |
+            stateCh!(({}, {}, $$initialBonds$$)) |
 
             contract PoS (@"getBonds", returnCh) = {
               new tmpCh in {
@@ -99,7 +92,7 @@ new  PoS,
                       match depositResult {
                         (true, _) => {
                           resultCh!(
-                            (pending, rewards.set(userPk, 0), bonds.set(userPk,amount)),
+                            (pending, rewards, bonds.set(userPk,amount)),
                             depositResult
                           )
                         }
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -67,6 +67,10 @@ new  PoS,
               }
             } |
 
+            /**
+             * Returns a Map[PublidKey, Int] containing the rewards accumulated for each validator.
+             * The returned map contains only the committed rewards after the last closeBlock
+             */
             contract PoS (@"getRewards", returnCh) = {
               new tmpCh in {
                 getMVar!(*stateCh, *tmpCh) |
@@ -113,21 +117,18 @@ new  PoS,
             contract PoS (@"pay", @amount, returnCh) = {
               new vaultCh, transferCh, userCh, processPayCh,
                   depositCh,
-                  computeSum, totalStakeCh,
-                  computeDelta, rewardsDeltaCh,
-                  computeMergeMap, newRewardsCh in {
+                  newPendingCh in {
                 getUser!(*userCh) |
-
                 runMVar!(*stateCh, *processPayCh, *returnCh) |
                 for(@(pending, rewards, bonds), payResultCh <- processPayCh;
                     @userPk <- userCh) {
                   deposit!(userPk, amount, *depositCh) |
 
-                  distributeRewards!(amount, bonds, rewards, *newRewardsCh) |
+                  distributeRewards!(amount, bonds, pending, *newPendingCh) |
 
                   for(@depositResult <- depositCh;
-                      @newRewards <- newRewardsCh) {
-                    payResultCh!((pending, newRewards, bonds), depositResult)
+                      @newPending <- newPendingCh) {
+                    payResultCh!((newPending, rewards, bonds), depositResult)
                   }
                 }
               }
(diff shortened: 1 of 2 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -79,22 +79,8 @@ object Runtime {
   type BodyRef   = Long
 
   final case class BlockData(timeStamp: Long, blockNumber: Long)
-
```
KentShikama: Nitpick: timestamp

KentShikama (approved):

casper/src/main/resources/PoS.rhox
```diff
@@ -76,27 +81,27 @@ new  PoS,
...
-                    resultCh!((rewards, bonds), (false, "Bond is greater than maximum!"))
+                    resultCh!((pending, rewards, bonds), (false, "Bond is greater than maximum!"))
                   } else {
                     deposit!(userPk, amount, *depositCh) |
                     for(@depositResult <- depositCh) {
                       match depositResult {
                         (true, _) => {
                           resultCh!(
-                            (rewards.set(userPk, 0), bonds.set(userPk,amount)),
+                            (pending, rewards.set(userPk, 0), bonds.set(userPk,amount)),
```
JosephDenman: Why isn't the first component of this tuple `pending.set(userPk, 0)`?

casper/src/main/resources/PoS.rhox
```diff
@@ -133,12 +135,14 @@ new  PoS,
...
                 getInvalidBlocks!(*invalidBlocksCh) |
                 getUser!(*userCh) |
-                for (@invalidBlocks <- invalidBlocksCh; @(rewards, bonds) <- stateCh; @userPk <- userCh) {
+                for (@invalidBlocks <- invalidBlocksCh; @(pending, rewards, bonds) <- stateCh; @userPk <- userCh) {
                   new toBeSlashed in {
                     toBeSlashed!(invalidBlocks.getOrElse(blockHash, userPk)) |
                     for (@validator <- toBeSlashed) {
                       // TODO: Transfer to coop wallet instead of just simply setting bonds to 0
-                      stateCh!((rewards.set(validator, 0), bonds.set(validator, 0))) |
+                      stateCh!((pending.set(validator, 0),
```
JosephDenman, line 140: Why does slashing not result in the validator being removed entirely from the `pending`, `rewards`, and `bonds` maps? Why are we wasting storage keeping track of the zero balances?

casper/src/main/resources/PoS.rhox
```diff
@@ -65,7 +70,7 @@ new  PoS,
             contract PoS (@"getRewards", returnCh) = {
               new tmpCh in {
                 getMVar!(*stateCh, *tmpCh) |
-                for (@(rewards, _) <- tmpCh) {
+                for (@(_, rewards, _) <- tmpCh) {
```
JosephDenman, line 70: Please document that `getRewards` returns the `pending + committed` rewards that a validator had _before_ the last block was closed.

casper/src/test/resources/PoSTest.rho
```diff
@@ -193,37 +193,39 @@ match (
...
-                        contract isPositive(@n, resultCh) = { resultCh!(n > 0)} |
-
-                        @ListOps!("fold", deltaRewards.toList(), 0, *sumDelta, *deltaSumCh) |
-                        contract sumDelta(@(_, delta), @acc, resultCh) = {
-                          resultCh!(acc + delta)
+                    @PoS!("closeBlock", *ackCh) |
+                    for (_ <- ackCh) {
+                      @posVault!("balance", *finalPosBalanceCh) |
+                      @user1Vault!("balance", *finalUser1BalanceCh) |
+                      @PoS!("getRewards", *finalRewardsCh) |
```
JosephDenman: Will you add an assertion that the reward value of each validator in the `initialRewards` map is zero?

casper/src/test/resources/PoSTest.rho
```diff
@@ -193,37 +193,39 @@ match (
...
+                      for(@finalPosBalance <- finalPosBalanceCh;
+                          @finalUser1Balance <- finalUser1BalanceCh;
+                          @finalRewards <- finalRewardsCh) {
+                        @ListOps!("fold", finalRewards.toList(), {}, *computeDelta, *deltaRewardsCh) |
+                        contract computeDelta(@(pk, finalReward), @acc, resultCh) = {
+                          resultCh!(acc.set(pk, finalReward - initialRewards.get(pk)))
                         } |
+                        for (@deltaRewards <- deltaRewardsCh) {
+                          @ListOps!("forall", deltaRewards, *isPositive, *allRewardsPositiveCh) |
```
JosephDenman: Please change this test to assert that each validator's reward has increased by the proportion of the payment amount corresponding to its weight.

JosephDenman (changes requested):
Good stuff. After you address my comments, I'm happy to approve.

I'm curious, are there no tests for slashing because the definition of `slash` is incomplete?

casper/src/main/resources/PoS.rhox
odeac: OK. I'll add a comment to describe what the "getRewards" actually does.

casper/src/main/resources/PoS.rhox
odeac: That's a good point. Setting the bonds to zero was the original implementation and then I just followed without questioning it.

@KentShikama is there any reason to keep the slashed validators in the map?

casper/src/test/resources/PoSTest.rho
odeac: The problem here is that the tests are not truly isolated. Since we have a singleton that is instantiated at genesis, I cannot re-instantiate it for each test. So, at this point, I can't really know what the initial reward is for a validator because it depends on what tests have been executed so far.

PS: These tests are pretty ugly because of this reason and I am thinking to suggest a refactoring such that we have the PoS factory and the PoS instance separated. This is something to discuss in the future.

casper/src/test/resources/PoSTest.rho
odeac: That becomes too complicated. What if I test that only the current validator reward is increased by the expected amount?

casper/src/main/resources/PoS.rhox
odeac: Yeah, I can do that. For me it just felt more natural to say "if it's not found then it's zero". But you are right that I should do that in order to keep it consistent with the rewards map.

casper/src/main/resources/PoS.rhox
KentShikama: https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/Validate.scala#L741

casper/src/main/resources/PoS.rhox
odeac: Now, after debugging some situation where the keys are not found and the Map.get fails silently I kind of changed my mind. I would keep the pending code as it is and change the rewards code to use getOrElse because relying on get is more brittle than using getOrElse. 

What do you think?

JosephDenman: I have no preference as long as the usage is consistent.

casper/src/test/resources/PoSTest.rho
JosephDenman: > I cannot re-instantiate it for each test

It's not that you can't. You would just have to go through an entire test setup to do so (construct `RhoSpec`, `PoS`, `RevVault`, etc.) and that's cumbersome. There's a question of whether one `RhoSpec` instance should rerun the test setup for each test, or whether one `RhoSpec` instance should be created for each test.

When we have to manually calculate what the final `PoS` state for a previous test is to inform how we write a subsequent test, we have a problem.

casper/src/test/resources/PoSTest.rho
JosephDenman: Is it more complicated than running a `forall` on all elements of the `finalRewards` map?

casper/src/main/resources/PoS.rhox
JosephDenman: @KentShikama I don't see where `getBonds` in the `PoS` contract is accessed in that function.

KentShikama: https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/Validate.scala#L761

KentShikama: It is cached and validated once. And that cached valued in the block is used.

casper/src/main/resources/PoS.rhox
odeac: ok. I use getOrElse for both

casper/src/main/resources/PoS.rhox
odeac: @KentShikama looking at https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/Validate.scala#L741 the result is the same if the stake is zero or missing from the bonds map. Does this mean that, at slashing, we can remove the entry from the bonds map?

casper/src/test/resources/PoSTest.rho
odeac: The test is too big anyway and both with yours and my proposal we test the same call paths but mine is simpler.

odeac:
> I'm curious, are there no tests for slashing because the definition of `slash` is incomplete?

Not sure. Somehow slash just got merged without the tests. Anyway, when the future logic will be added to slash we will certainly have to add tests for the whole method

JosephDenman (approved):

PR #2560, opened 2019-06-28
RCHAIN-3220: Rholang interpreter prints out unmatched sends

tymm:
## Overview
This is part 1.
Part 2 will add support for RhoSpec.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3672

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -21,6 +21,7 @@ import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.Par
 import coop.rchain.models.rholang.sorter.Sortable._
 import coop.rchain.models.serialization.implicits.mkProtobufInstance
+import coop.rchain.rholang.interpreter.storage.StoragePrinter
 import coop.rchain.rspace.StableHashProvider
 import coop.rchain.rspace.trace._
 import coop.rchain.shared.Log
@@ -65,7 +66,8 @@ object BlockAPI {
   }
 
   def createBlock[F[_]: Sync: Concurrent: MultiParentCasperRef: Log: Span](
-      blockApiLock: Semaphore[F]
+      blockApiLock: Semaphore[F],
+      printUnmatchedSends: Boolean = false
   ): Effect[F, DeployServiceResponse] = Span[F].trace(CreateBlockSource) {
     val errorMessage = "Could not create block, casper instance was not available yet."
     MultiParentCasperRef.withCasper[F, ApiErr[DeployServiceResponse]](
@@ -79,8 +81,12 @@ object BlockAPI {
                            s"Error while creating block: $err".asLeft[DeployServiceResponse].pure[F]
                          case Created(block) =>
                            casper
-                             .addBlock(block, ignoreDoppelgangerCheck[F])
-                             .map(addResponse(_, block))
+                             .addBlock(block, ignoreDoppelgangerCheck[F]) >>= (addResponse(
+                             _,
+                             block,
+                             casper,
+                             printUnmatchedSends
+                           ))
                        }
             } yield result
           case false =>
@@ -565,21 +571,42 @@ object BlockAPI {
       case None             => none[BlockMessage]
     }
 
-  private def addResponse(
+  private def addResponse[F[_]: Concurrent](
       status: BlockStatus,
-      block: BlockMessage
-  ): ApiErr[DeployServiceResponse] =
+      block: BlockMessage,
+      casper: MultiParentCasper[F],
+      printUnmatchedSends: Boolean
+  ): Effect[F, DeployServiceResponse] =
     status match {
       case _: InvalidBlock =>
-        s"Failure! Invalid block: $status".asLeft
+        s"Failure! Invalid block: $status".asLeft[DeployServiceResponse].pure[F]
       case _: ValidBlock =>
-        val hash = PrettyPrinter.buildString(block.blockHash)
-        DeployServiceResponse(s"Success! Block $hash created and added.").asRight
+        val hash    = PrettyPrinter.buildString(block.blockHash)
+        val deploys = block.body.get.deploys.map(_.deploy.get)
+        val maybeUnmatchedSendsOutputF =
+          if (printUnmatchedSends) prettyPrintUnmatchedSends(casper, deploys).map(_.some)
+          else none[String].pure[F]
+        maybeUnmatchedSendsOutputF >>= (
+            maybeOutput =>
+              DeployServiceResponse(
+                s"Success! Block $hash created and added.${maybeOutput.map("\n" + _).getOrElse("")}"
+              ).asRight[Error].pure[F]
+          )
       case BlockException(ex) =>
-        s"Error during block processing: $ex".asLeft
+        s"Error during block processing: $ex".asLeft[DeployServiceResponse].pure[F]
       case Processing =>
-        "No action taken since other thread is already processing the block.".asLeft
+        "No action taken since other thread is already processing the block."
+          .asLeft[DeployServiceResponse]
+          .pure[F]
     }
+
+  private def prettyPrintUnmatchedSends[F[_]: Concurrent](
+      casper: MultiParentCasper[F],
+      deploys: Seq[DeployData]
+  ): F[String] =
+    casper.getRuntimeManager >>= (
+      _.withRuntimeLock(runtime => StoragePrinter.prettyPrintUnmatchedSends(deploys, runtime))
+    )
 
   def previewPrivateNames[F[_]: Monad: Log](
       deployer: ByteString,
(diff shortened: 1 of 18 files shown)


node/src/main/scala/coop/rchain/node/api/ReplGrpcService.scala
```diff
@@ -67,7 +70,7 @@ private[api] class ReplGrpcService(runtime: Runtime[Task], worker: Scheduler)
                   .map(_.toString())
                   .mkString("Errors received during evaluation:\n", "\n", "\n")
             s"Deployment cost: $cost\n" +
-              s"${errorStr}Storage Contents:\n ${StoragePrinter.prettyPrint(runtime.space)}"
+              s"${errorStr}Storage Contents:\n$prettyStorage"
```
kayvank: why an empty line?

tymm: Not sure what you mean?
Before the first term was offset by one space while all following terms were not.

node/src/main/scala/coop/rchain/node/effects/ReplClient.scala
```diff
@@ -43,15 +46,18 @@ class GrpcReplClient(host: String, port: Int, maxMessageSize: Int)
...
+  def eval(
+      fileNames: List[String],
+      unmatchedSends: Boolean
+  ): Task[List[Either[Throwable, String]]] =
     fileNames
-      .traverse(eval)
+      .traverse(eval(_, unmatchedSends))
 
-  def eval(fileName: String): Task[Either[Throwable, String]] = {
+  def eval(fileName: String, printUnmatchedSendsOnly: Boolean): Task[Either[Throwable, String]] = {
```
kayvank, line 56: should this private?

kayvank (approved):

node/src/main/scala/coop/rchain/node/effects/ReplClient.scala
tymm: Nope, it's called from `ReplClient.scala`.

node/src/main/scala/coop/rchain/node/api/ReplGrpcService.scala
kayvank: just a formatting question. Why keep the empty lines, line 74 & line 75?

tymm: Oh, I see. Just removed them.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
```diff
@@ -50,6 +50,30 @@ object StoragePrinter {
...
+      pars = mapped.map {
+        case (
+            channels: Seq[Par],
+            row: Row[BindPattern, ListParWithRandom, TaggedContinuation]
+            ) =>
+          row match {
+            case Row(data: Seq[Datum[ListParWithRandom]], _) =>
+              toSends(data)(channels)
+          }
+      }.toList
```
odeac: do you need the toList here since the next step is to fold it?

node/src/main/scala/coop/rchain/node/api/ReplGrpcService.scala
```diff
@@ -57,7 +57,9 @@ private[api] class ReplGrpcService(runtime: Runtime[Task], worker: Scheduler)
               implicit val c = runtime.cost
               Interpreter[Task].evaluate(runtime, source, NormalizerEnv.Empty)
             }
-            prettyStorage                <- StoragePrinter.prettyPrint(runtime.space)
+            prettyStorage <- if (printUnmatchedSendsOnly)
```
odeac, line 60: So we can choose to either print only the tuplespace or only the unmatched sends. Does it makes sense to print both in some situations?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RholangCLI.scala
```diff
@@ -226,7 +252,12 @@ object RholangCLI {
     })
   }
 
-  def evaluatePar(runtime: Runtime[Task], source: String, quiet: Boolean = false)(
+  def evaluatePar(
+      runtime: Runtime[Task],
+      source: String,
+      quiet: Boolean = false,
```
odeac: are the default values  required for the parameters quiet and unmatchedSendsOnly? My impression is that, at call site, we provide values for both

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -566,20 +572,44 @@ object BlockAPI {
...
+        "No action taken since other thread is already processing the block."
+          .asLeft[DeployServiceResponse]
+          .pure[F]
+    }
+
+  private def prettyPrintUnmatchedSends[F[_]: Concurrent](
+      casper: MultiParentCasper[F],
+      deploys: Seq[DeployData]
+  ): F[Option[String]] =
+    casper.getRuntimeManager >>= {
```
odeac: you should be able to use traverse instead of flatMap + match

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -566,20 +572,44 @@ object BlockAPI {
...
+      case Some(rm) =>
+        rm.withRuntimeLock(runtime => StoragePrinter.prettyPrintUnmatchedSends(deploys, runtime))
```
odeac: now I see the usage of withRuntimeLock but why do you need to lock the runtime? I'm thinking that the printing of unmatched sends happen after the execution ended so it's not executed concurrently with other runtime operations

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
```diff
@@ -40,39 +45,79 @@ object StoragePrinter {
...
+      deploys: Seq[DeployData],
+      runtime: Runtime[F]
+  ): F[String] =
+    deploys
+      .map(
+        deploy =>
+          prettyPrintUnmatchedSends(deploy, runtime)
+            .map(unmatchedSends => (deploy, unmatchedSends))
+      )
+      .toList
```
odeac: I assume if you need to traverse the collection so you can use toStream instead of toList. That would be better because it doesn't build the intermediary list

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
```diff
@@ -40,39 +45,79 @@ object StoragePrinter {
...
+      .sequence
```
odeac: map followed by sequence can be replaced by traverse

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tymm: `StoragePrinter.prettyPrintUnmatchedSends` executes Rholang. Therefore I think we should use `withRuntimeLock`.

node/src/main/scala/coop/rchain/node/api/ReplGrpcService.scala
tymm: When you print the tuplespace, you also get the unmatched sends. Therefore I would argue that it doesn't make sense to print both at the same time.

odeac: ok

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tymm: Hmm, how would that work? We would have to add the `Traverse` typeclass to `F`. Do we want that?

odeac: I'm sorry. I wasn't paying attention at the types when I wrote my comment.

This is what I meant:

```scala
  private def prettyPrintUnmatchedSends[F[_]: Concurrent](
      casper: MultiParentCasper[F],
      deploys: Seq[DeployData]
  ): F[Option[String]] =
    casper.getRuntimeManager >>= (maybeRuntime =>
      maybeRuntime.traverse( rm =>
        rm.withRuntimeLock(
          runtime => StoragePrinter.prettyPrintUnmatchedSends(deploys, runtime))))
```

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
odeac: ok

odeac (approved):

tymm:
bors try

tymm:
bors retry

tymm:
bors try

PR #2564, opened 2019-07-02
Metric: received, enqueued or dropped messages

sebtomba:
## Overview
Add metrics how many incoming messages/packets have been received, enqueued or dropped. Also, add a counter of how many messages/packets have been dispatched.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3579

### Notes
Bug fixed:
When a packet gets dropped because the message queue is full then delete the packet file.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
@@ -6,11 +6,12 @@ import scala.concurrent.duration._
 
 import cats.implicits._
 
-import coop.rchain.comm.PeerNode
+import coop.rchain.comm.{CommMetricsSource, PeerNode}
 import coop.rchain.comm.protocol.routing._
 import coop.rchain.comm.rp.Connect.RPConfAsk
 import coop.rchain.comm.rp.ProtocolHelper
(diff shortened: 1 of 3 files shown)


comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
```diff
@@ -43,31 +50,46 @@ object GrpcTransportReceiver {
...
+              metrics.incrementCounter("received.packets") >>
+                Task
+                  .delay(blobBuffer.pushNext(msg))
+                  .ifM(
+                    List(
+                      logger.debug(s"Enqueued for handling packet ${msg.path}"),
+                      metrics.incrementCounter("enqueued.packets")
+                    ).sequence,
+                    List(
+                      logger.debug(s"Dropped packet ${msg.path}"),
```
woky, line 88: Could you attach information from `msg` (such as `sender`) to these messages? It'd be best if network layer had its own logging mechanism so that you wouldn't have to put it into arguments on each log call. The attached peer info could possibly be structured via https://logback.qos.ch/manual/mdc.html.

kayvank (approved):

EncodePanda (approved):
LGTM

PR #2566, opened 2019-07-03
Rholang peek

lukasz-golebiewski:
## Overview
Adds peek semantics in binds to rholang: `for (a <<- @1) { ... }`
At present it is only possible to use binds of only the same type in a receive, so for consistency this is also the case for peek binds

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2107

### Notes

### DIFF
--- a/models/src/main/protobuf/RhoTypes.proto
+++ b/models/src/main/protobuf/RhoTypes.proto
@@ -134,7 +134,8 @@ message Receive {
     repeated ReceiveBind binds = 1;
     Par body = 2 [(scalapb.field).no_box = true];
     bool persistent = 3;
-    int32 bindCount = 4;
+    bool peek = 4;
+    int32 bindCount = 5;
     bytes locallyFree = 6 [(scalapb.field).type = "coop.rchain.models.AlwaysEqual[scala.collection.immutable.BitSet]"];
     bool connective_used = 7;
 }
--- a/models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
@@ -29,8 +29,9 @@ object ReceiveSortMatcher extends Sortable[Receive] {
   // This function will then sort the insides of the preordered binds.
   def sortMatch[F[_]: Sync](r: Receive): F[ScoredTerm[Receive]] =
     for {
-      sortedBinds         <- r.binds.toList.traverse(sortBind[F])
-      persistentScore     = if (r.persistent) 1L else 0L
+      sortedBinds     <- r.binds.toList.traverse(sortBind[F])
+      persistentScore = if (r.persistent) 1L else 0L
+      // TODO: score peek
       connectiveUsedScore = if (r.connectiveUsed) 1L else 0L
       sortedBody          <- Sortable.sortMatch(r.body)
     } yield ScoredTerm(
@@ -38,6 +39,7 @@ object ReceiveSortMatcher extends Sortable[Receive] {
         sortedBinds.map(_.term),
         sortedBody.term,
         r.persistent,
+        r.peek,
         r.bindCount,
         r.locallyFree,
         r.connectiveUsed
--- a/models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
@@ -29,9 +29,9 @@ object ReceiveSortMatcher extends Sortable[Receive] {
   // This function will then sort the insides of the preordered binds.
   def sortMatch[F[_]: Sync](r: Receive): F[ScoredTerm[Receive]] =
     for {
-      sortedBinds     <- r.binds.toList.traverse(sortBind[F])
-      persistentScore = if (r.persistent) 1L else 0L
-      // TODO: score peek
+      sortedBinds         <- r.binds.toList.traverse(sortBind[F])
+      persistentScore     = if (r.persistent) 1L else 0L
+      peekScore           = if (r.peek) 1L else 0L
       connectiveUsedScore = if (r.connectiveUsed) 1L else 0L
       sortedBody          <- Sortable.sortMatch(r.body)
     } yield ScoredTerm(
@@ -46,7 +46,9 @@ object ReceiveSortMatcher extends Sortable[Receive] {
       ),
       Node(
         Score.RECEIVE,
-        Seq(Leaf(persistentScore)) ++ sortedBinds.map(_.score) ++ Seq(sortedBody.score)
+        Seq(Leaf(persistentScore), Leaf(peekScore)) ++ sortedBinds.map(_.score) ++ Seq(
+          sortedBody.score
+        )
           ++ Seq(Leaf(r.bindCount.toLong)) ++ Seq(Leaf(connectiveUsedScore)): _*
       )
     )
--- a/models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
+++ b/models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
@@ -384,7 +384,7 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
     result.term should be(sortedParExpr)
   }
 
-  it should "sort Receives based on persistence, channels, patterns and then body" in {
+  it should "sort Receives based on persistence, peek, channels, patterns and then body" in {
     val parExpr =
       Par(
         receives = List(
@@ -417,6 +417,14 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             Par(),
             true,
             false,
+            0,
+            BitSet()
+          ),
+          Receive(
+            List(ReceiveBind(List(GInt(0)), GInt(3))),
+            Par(),
+            true,
+            true,
             0,
             BitSet()
           ),
@@ -465,7 +473,15 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             0,
             BitSet()
           ),
-          Receive(List(ReceiveBind(List(GInt(0)), GInt(3))), Par(), true, false, 0, BitSet())
+          Receive(List(ReceiveBind(List(GInt(0)), GInt(3))), Par(), true, false, 0, BitSet()),
+          Receive(
+            List(ReceiveBind(List(GInt(0)), GInt(3))),
+            Par(),
+            true,
+            true,
+            0,
+            BitSet()
+          )
         )
       )
     val result = sort(parExpr)
--- a/models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
+++ b/models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
@@ -392,6 +392,7 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(1)), GInt(3))),
             Par(),
             false,
+            false,
             0,
             BitSet()
           ),
@@ -399,12 +400,14 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(0)), GInt(3))),
             EVar(BoundVar(0)),
             false,
+            false,
             0,
             BitSet()
           ),
           Receive(
             List(ReceiveBind(List(GInt(0)), GInt(3))),
             Par(),
+            false,
             false,
             0,
             BitSet()
@@ -413,12 +416,14 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(0)), GInt(3))),
             Par(),
             true,
+            false,
             0,
             BitSet()
           ),
           Receive(
             List(ReceiveBind(List(GInt(100)), GInt(2))),
             Par(),
+            false,
             false,
             0,
             BitSet()
@@ -432,12 +437,14 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(100)), GInt(2))),
             Par(),
             false,
+            false,
             0,
             BitSet()
           ),
           Receive(
             List(ReceiveBind(List(GInt(0)), GInt(3))),
             Par(),
+            false,
             false,
             0,
             BitSet()
@@ -446,6 +453,7 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(0)), GInt(3))),
             EVar(BoundVar(0)),
             false,
+            false,
             0,
             BitSet()
           ),
@@ -453,10 +461,11 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
             List(ReceiveBind(List(GInt(1)), GInt(3))),
             Par(),
             false,
+            false,
             0,
             BitSet()
           ),
-          Receive(List(ReceiveBind(List(GInt(0)), GInt(3))), Par(), true, 0, BitSet())
+          Receive(List(ReceiveBind(List(GInt(0)), GInt(3))), Par(), true, false, 0, BitSet())
         )
       )
     val result = sort(parExpr)
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -747,6 +747,7 @@ object ProcNormalizeMatcher {
               ),
               body = bodyResult.par,
               persistent = true,
+              peek = false,
               bindCount = boundCount,
               locallyFree = ParLocallyFree
                 .locallyFree(nameMatchResult.chan, input.env.depth) | formalsResults._3
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -861,7 +861,10 @@ object ProcNormalizeMatcher {
                       (lbi.listname_.toList, lbi.name_, lbi.nameremainder_).pure[M]
                   }
                   .map(x => (x, false, true))
+              case default =>
+                sync.raiseError(NormalizerError(s"Unknown receipt impl type $default"))
             }
+          case default => sync.raiseError(NormalizerError(s"Unknown receipt type $default"))
         }
 
         for {
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -839,7 +839,7 @@ object ProcNormalizeMatcher {
                     case lbi: LinearBindImpl =>
                       (lbi.listname_.toList, lbi.name_, lbi.nameremainder_).pure[M]
                   }
-                  .map(x => (x, false))
+                  .map(x => (x, false, false))
             }
           case rl: ReceiptRepeated =>
             rl.receiptrepeatedimpl_ match {
@@ -849,13 +849,23 @@ object ProcNormalizeMatcher {
                     case lbi: RepeatedBindImpl =>
                       (lbi.listname_.toList, lbi.name_, lbi.nameremainder_).pure[M]
                   }
-                  .map(x => (x, true))
+                  .map(x => (x, true, false))
+            }
+          case rl: ReceiptPeek =>
+            rl.receiptpeekimpl_ match {
+              case ls: PeekSimple =>
+                ls.listpeekbind_.toList
+                  .traverse {
+                    case lbi: PeekBindImpl =>
+                      (lbi.listname_.toList, lbi.name_, lbi.nameremainder_).pure[M]
+                  }
+                  .map(x => (x, false, true))
             }
         }
 
         for {
           res                                                              <- resM
-          (bindingsRaw, persistent)                                        = res
+          (bindingsRaw, persistent, peek)                                  = res
           sourcesP                                                         <- processSources(bindingsRaw)
           (sources, thisLevelFree, sourcesLocallyFree, sourcesConnectives) = sourcesP
           bindingsProcessed                                                <- processBindings(sources)
@@ -896,6 +906,7 @@ object ProcNormalizeMatcher {
               binds,
               bodyResult.par,
               persistent,
+              peek,
               bindCount,
               sourcesLocallyFree | bindingsFree | (bodyResult.par.locallyFree
                 .from(bindCount)
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -12,6 +12,8 @@ import coop.rchain.rholang.interpreter.storage.implicits.matchListPar
 import coop.rchain.rspace.util._
 import coop.rchain.rspace.{Blake2b256Hash, Checkpoint, ContResult, Result, Match => StorageMatch}
 
+import scala.collection.SortedSet
+
 object ChargingRSpace {
   def storageCostConsume(
       channels: Seq[Par],
@@ -40,14 +42,25 @@ object ChargingRSpace {
           patterns: Seq[BindPattern],
           continuation: TaggedContinuation,
           persist: Boolean,
-          sequenceNumber: Int
+          sequenceNumber: Int,
+          peek: Boolean
       ): F[
         Option[(ContResult[Par, BindPattern, TaggedContinuation], Seq[Result[ListParWithRandom]])]
       ] =
         for {
-          _       <- charge[F](storageCostConsume(channels, patterns, continuation))
-          consRes <- space.consume(channels, patterns, continuation, persist, sequenceNumber)
-          _       <- handleResult(consRes)
+          _ <- charge[F](storageCostConsume(channels, patterns, continuation))
+          peekChannels = if (peek) {
+            SortedSet((0 to channels.size - 1): _*)
+          } else SortedSet.empty[Int]
+          consRes <- space.consume(
+                      channels,
+                      patterns,
+                      continuation,
+                      persist,
+                      sequenceNumber,
+                      peekChannels
+                    )
+          _ <- handleResult(consRes)
         } yield consRes
 
       override def install(
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
@@ -70,7 +70,7 @@ object StoragePrinter {
           patterns: Seq[BindPattern],
           continuation: TaggedContinuation,
           persist: Boolean,
-          _: SortedSet[Int],
+          peeks: SortedSet[Int],
           _: Consume
           ) =>
         val receiveBinds: Seq[ReceiveBind] = (channels zip patterns).map {
@@ -88,6 +88,7 @@ object StoragePrinter {
               receiveBinds,
               p.body,
               persist,
+              peeks.nonEmpty,
               patterns.map(_.freeCount).sum
             )
           case _ => Receive(receiveBinds, Par.defaultInstance, persist)
--- a/rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
@@ -16,6 +16,7 @@ import scala.concurrent.duration._
 class PeekSpec extends FlatSpec with Matchers {
 
   import Resources._
+  import InterpreterUtil._
 
   implicit val logF: Log[Task]            = new Log.NOPLog[Task]
   implicit val noopMetrics: Metrics[Task] = new Metrics.MetricsNOP[Task]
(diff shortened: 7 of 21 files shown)


models/src/main/protobuf/RhoTypes.proto
```diff
@@ -134,7 +134,8 @@ message Receive {
     repeated ReceiveBind binds = 1;
     Par body = 2 [(scalapb.field).no_box = true];
     bool persistent = 3;
-    int32 bindCount = 4;
+    bool peek = 4;
```
lukasz-golebiewski, line 137: TODO: replace presistent & peek with a single field holding an ADT, e.g. `BindType = Normal | Persistent | Peek`

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -849,13 +850,23 @@ object ProcNormalizeMatcher {
...
                   }
-                  .map(x => (x, true))
+                  .map(x => (x, true, false))
+            }
+          case rl: ReceiptPeek =>
+            rl.receiptpeekimpl_ match {
+              case ls: PeekSimple =>
+                ls.listpeekbind_.toList
+                  .traverse {
+                    case lbi: PeekBindImpl =>
```
odeac, line 860: can you use a total function here?

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
```diff
@@ -29,15 +29,17 @@ object ReceiveSortMatcher extends Sortable[Receive] {
   // This function will then sort the insides of the preordered binds.
   def sortMatch[F[_]: Sync](r: Receive): F[ScoredTerm[Receive]] =
     for {
-      sortedBinds         <- r.binds.toList.traverse(sortBind[F])
-      persistentScore     = if (r.persistent) 1L else 0L
+      sortedBinds     <- r.binds.toList.traverse(sortBind[F])
+      persistentScore = if (r.persistent) 1L else 0L
+      // TODO: score peek
```
dzajkowski: is there a ticket for this?

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -39,14 +41,24 @@ object ChargingRSpace {
...
-          consRes <- space.consume(channels, patterns, continuation, persist, sequenceNumber)
-          _       <- handleResult(consRes)
+          _ <- charge[F](storageCostConsume(channels, patterns, continuation))
+          consRes <- space.consume(
+                      channels,
+                      patterns,
+                      continuation,
+                      persist,
+                      sequenceNumber,
+                      if (peek) {
```
dzajkowski: nitpick: don't you think that an if merits a separate line in the for?

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -39,14 +41,24 @@ object ChargingRSpace {
...
+                        SortedSet((0 to channels.size - 1): _*)
+                      } else SortedSet.empty
```
dzajkowski: is there a ticket for expanding the `for` syntax in rholang to accommodate mixed consumes?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -70,17 +71,43 @@ trait PersistentStoreTester {
...
+  }
 }
 
 class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
   implicit val rand: Blake2b512Random = Blake2b512Random(Array.empty[Byte])
 
+  /**
+    Do not use this method, see explanation below.
+    Use mapData and assert at call site
+    */
```
dzajkowski, line 104: perhaps mark is as deprecated?

dzajkowski:
would it make sense to provide a test that runs two continuations and depends on the 1st one peeking the data they depend on?

dzajkowski (approved):
small nits

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -39,14 +41,24 @@ object ChargingRSpace {
           patterns: Seq[BindPattern],
           continuation: TaggedContinuation,
           persist: Boolean,
-          sequenceNumber: Int
+          sequenceNumber: Int,
+          peek: Boolean
```
odeac, line 46: The flag "peek" which selects different behaviours is a bad smell.

Another issue is that the name of the method "consume" is no longer valid if we are talking about peeks. 

I think this can be fixed by having two methods: `consume` and `peek` in RhoPureSpace

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
lukasz-golebiewski: This will cause compilation errors

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
lukasz-golebiewski: I don't think there is one, we should discuss whether this is needed for mercury (or at all)

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
lukasz-golebiewski: There is now https://rchain.atlassian.net/browse/RCHAIN-3600

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
lukasz-golebiewski: Can do

lukasz-golebiewski:
@dzajkowski sure, this makes sense. I was thinking of adding such a test but didn't find an appropriate spot. Any suggestions?

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
lukasz-golebiewski: This might make sense temporarily, but if we extend Rholang syntax to allow mixing standard/persistent/peek reads in binds under the same receive this change will have to be reverted

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
lukasz-golebiewski: No

models/src/main/protobuf/RhoTypes.proto
lukasz-golebiewski: I'm hesitating with introducing this change because it would have to be reverted if we decided to move the peek flag to `ReceiveBind`

ArturGajowy: we are, I believe

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
ArturGajowy: This PR is incomplete without this change, that's a no-go for merging. Rholang terms will no longer be sortable until this is fixed, and we rely on this propoerty in multiple places.

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -646,6 +649,49 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
...
+      new PeekBindImpl(listBindings, new NameRemainderEmpty(), new NameQuote(new PNil()))
+    )
+    val linearSimple = new PeekSimple(listLinearBinds)
+    val receipt      = new ReceiptPeek(linearSimple)
+
+    val listSend = new ListProc()
+    listSend.add(new PEval(new NameVar("y")))
+    val body       = new PSend(new NameVar("x"), new SendSingle(), listSend)
+    val basicInput = new PInput(receipt, body)
+    val bindCount  = 2
```
ArturGajowy: I know this is not consistent with code elsewhere, but we should be using the parser to construct the input for this test. Basically we should be using ParBuilder.buildAST, which should probably be private[rholang.interpreter]. This way we're going to test the grammar somewhat as well.

So, use ParBuilder.buildAST here

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -1194,6 +1243,7 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
           ),
           Par(),
           persistent = false,
+          peek = false,
```
ArturGajowy, line 1211: This is more of a general remark for anybody reading this than a request for this particular PR, but:

the amount of places where the expected test result changed is hideous.
Tests should only assert over the thing they test, nothing more. Unless we want concrete boots for our future refactorings.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
```diff
@@ -88,6 +88,7 @@ object StoragePrinter {
               receiveBinds,
               p.body,
               persist,
+              peeks.nonEmpty,
```
ArturGajowy, line 91: throw/raise an error here if the peek flag differs for different channels

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -646,6 +649,49 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
...
+            ReceiveBind(List(EVar(FreeVar(0)), EVar(FreeVar(1))), Par(), freeCount = 2)
+          ),
+          Send(
+            EVar(BoundVar(1)),
+            List[Par](EVar(BoundVar(0))),
+            false,
+            BitSet(0, 1)
+          ),
+          persistent = false,
+          peek = true,
```
ArturGajowy: this test should only care about the `peek` property being true (see my latter comment for rationale).

Don't compare the whole structure, just assert the `peek` property

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -747,6 +747,7 @@ object ProcNormalizeMatcher {
               ),
               body = bodyResult.par,
               persistent = true,
+              peek = false,
```
ArturGajowy, line 750: where's the normalizer change for peek? I mean, it's not the case that all normalizer outputs are going to be non-peeks. They are in syntax already.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: yes, peek is going to be part of BindPattern

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: It is needed, please ticket that. I assumed that to be the scope of this change, but a JIRA will do for now. Not needed for mercure most likely.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
ArturGajowy: we also need changes in rholang PrettyPrinter, and a test for them

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
ArturGajowy: also, I think there are tests for the sorter / normalizer that check that fields are not omitted from the input. We need a similar test for peek in the reduce

ArturGajowy (changes requested):
A couple comments for now, will add more later

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
lukasz-golebiewski: I didn't assume that, so maybe we should have been more explicit about this. From my perspective the amount of changes needed to implement thist would fan-out wide, and as it isn't absolutely necessary it could be done in a separate PR. Here's the ticket:
https://rchain.atlassian.net/browse/RCHAIN-3607

models/src/main/protobuf/RhoTypes.proto
lukasz-golebiewski: Ok, in that case I'll skip this change

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
lukasz-golebiewski: What do you mean? Peeks should be available in standard receives, a.k.a. `PInput` before normalization. My understanding is that peeks in contract binds should not be allowed, hence `false` here (here a `PContr` is matched)

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
lukasz-golebiewski: Not sure which test you have in mind here

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
lukasz-golebiewski: Agreed. This comment belongs rather somewhere here: https://github.com/rchain/rchain/pull/2540/files#diff-8b18b7dd8293005720929ac43eb7f90dR7

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
lukasz-golebiewski: 1) I don't think we need to be so defensive in the printer given that the current rholang syntax doesn't allow it
2) Yeah, makes sense - added

lukasz-golebiewski:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
```diff
@@ -212,6 +212,33 @@ class ProcPrinterSpec extends FlatSpec with Matchers {
     result shouldBe target
   }
 
+  "Receive" should "print peek" in {
```
tymm, line 215: Could you make use of the `checkRoundTrip` function in this test? Makes it much more concise and readable :)

tymm (changes requested):

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
odeac: so if the case doesn't match we get a match failure... not nice

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
odeac: I understand. So we need one method to cover a mixture of peek and consumes. Still, the flag is a bad smell as we should not pass a flag to switch between the two behaviours internally. 

What if we pass the list of "peeked channels" and the list of "consumed channels"? This way we won't need the flag

rholang/src/test/scala/coop/rchain/rholang/interpreter/PrettyPrinterTest.scala
lukasz-golebiewski: True, done

odeac (approved):
I still have a problem with the peek flag in consume but that can be addressed later

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
JosephDenman: we should raise an error if the receipt type matches none of these.

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+  val tmpPrefix = "peek-spec-"
+
+  "peek" should "not remove read data" in {
+    mkRuntime[Task](tmpPrefix)
+      .use { runtime =>
+        implicit val c = runtime.cost
+        for {
+          res1 <- InterpreterUtil.evaluate[Task](runtime, """@1!("v1") | for(_ <<- @1) { Nil }""")
+          _    = res1.errors shouldBe empty
+          res2 <- InterpreterUtil.evaluate[Task](runtime, """for(_ <- @1) { @2!("v2") }""")
```
JosephDenman: It makes sense to add a similar test for persistent sends. One that asserts that peek does not duplicate v1 at @1.

JosephDenman (dismissed):

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -429,6 +459,72 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+
+    val channel: Par       = GString("channel")
+    val resultChannel: Par = GString("result")
+
+    val splitRand0 = rand.splitByte(0)
+    val splitRand1 = rand.splitByte(1)
+    val mergeRand  = Blake2b512Random.merge(Seq(splitRand1, splitRand0))
+    val send =
+      Send(channel, List(GInt(7L), GInt(8L), GInt(9L)), false, BitSet())
+    val receive = Receive(
```
tymm, line 496: Nitpick: named arguments would make it so much easier to read for me.

tymm (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
ArturGajowy: If the syntax doesn't allow that - what's wrong with throwing the exception? The point of being defensive is so that we spot things immediately when the outside context / callsite changes. I insist :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
ArturGajowy: Ah, true. I've simply not seen the whole PR, and didn't see the context for this line.

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
ArturGajowy: The one in SortTest:251. You should set `peek = true` in s2.

Also, make sure the property specs in that class (search: forAll) fail without your changes to the sorter. They should start failing as soon as you added the `peek` field. If they don't, we need to tweak the implicit PropertyCheckConfiguration (see scalatest docs).

ArturGajowy: >as you added the peek field

and the normalizer changes

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
```diff
@@ -46,7 +46,9 @@ object ReceiveSortMatcher extends Sortable[Receive] {
       ),
       Node(
         Score.RECEIVE,
-        Seq(Leaf(persistentScore)) ++ sortedBinds.map(_.score) ++ Seq(sortedBody.score)
+        Seq(Leaf(persistentScore)) ++ Seq(Leaf(peekScore)) ++ sortedBinds.map(_.score) ++ Seq(
```
ArturGajowy: Merge the neighoboring seq-s into Seq(persistentScore, peeksCore)

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
```diff
@@ -432,31 +437,35 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
...
           ),
           Receive(
             List(ReceiveBind(List(GInt(1)), GInt(3))),
             Par(),
             false,
+            false,
             0,
             BitSet()
           ),
-          Receive(List(ReceiveBind(List(GInt(0)), GInt(3))), Par(), true, 0, BitSet())
```
ArturGajowy, line 459: at least one `false` added in this test should be `true`

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  val tmpPrefix = "peek-spec-"
+
+  "peek" should "not remove read data" in {
+    mkRuntime[Task](tmpPrefix)
+      .use { runtime =>
+        implicit val c = runtime.cost
+        for {
+          res1 <- InterpreterUtil.evaluate[Task](runtime, """@1!("v1") | for(_ <<- @1) { Nil }""")
+          _    = res1.errors shouldBe empty
```
ArturGajowy, line 33: Please extract a `success` method to reduce the noise of these two tests.
Ideally, we'd have it in a util class in `rholang` tests.

Actually, InterpreterUtil should be in `test` scope - it's only used there. Please move it to test scope, then we can embed the 'no errors when evaulating' asseriton in its method. To discern between expected results, let's have two methods:

`evaulate(rho: String): F[Unit]`
`evaulateResult(rho: String): F[EvaulateResult]`

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -646,6 +649,19 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
...
+  it should "handle peek" in {
+    val basicInput = ParBuilder[Coeval]
+      .buildAST(
+        new StringReader("""for ( x, y <<- @Nil ) { x!(*y) }""")
+      )
+      .value()
+
+    val result = ProcNormalizeMatcher.normalizeMatch[Coeval](basicInput, inputs).value
+    result.par.receives.head.peek shouldBe true
+
```
ArturGajowy: much better, altough this empty line is redundant :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -646,6 +649,19 @@ class ProcMatcherSpec extends FlatSpec with Matchers {
     )
     result.knownFree should be(inputs.knownFree)
   }
+
+  it should "handle peek" in {
+    val basicInput = ParBuilder[Coeval]
+      .buildAST(
+        new StringReader("""for ( x, y <<- @Nil ) { x!(*y) }""")
```
ArturGajowy: Get rid of the StringReader in here. Options:

1. add an overload in the ParBuilder object that takes a string (notice the ParBuilder trait has such an overload)

2. Make `buildAST` a util method in the InterpreterUtil, and hide the `Coeval` in it as well

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -70,17 +71,43 @@ trait PersistentStoreTester {
...
+    Use mapData and assert at call site
+    */
   def checkData(
       result: Map[
         Seq[Par],
         Row[BindPattern, ListParWithRandom, TaggedContinuation]
       ]
   )(channel: Par, data: Seq[Par], rand: Blake2b512Random): Assertion =
+    // The 'should' assertion will fail on this line making it harder to find
+    // the exact spot where the error manifested itself
```
ArturGajowy, line 112: Your newly introduced tests don't differ in their shape from the existing ones. Please either:

1. fix all the tests w.r.t. to checkData usage and remove the method
2. extract a common core from the tests, fix the `checkData` issue once
   (either write your tests in the old style first and refactor all, or refactor and then add your tests in the new style)

In any case, do retain consistency, so that others can later do further cleanup more easily,

ArturGajowy (dismissed):
A couple more comments, I think we'll be fine to merge after these.

models/src/main/scala/coop/rchain/models/rholang/sorter/ReceiveSortMatcher.scala
lukasz-golebiewski: I think you meant `SortTest:387`, as peek is defined on the `Receive` level.

As for the generators, this can be done in a separate PR

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
lukasz-golebiewski: I don't think moving assertions to "library" methods is a good idea, because when they eventually fail the reported information about the line at which this happened is useless. Following the DRY principle at all cost doesn't always make sense in tests

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
lukasz-golebiewski: :)

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
ArturGajowy: > when they eventually fail the reported information about the line at which this happened is useless

It depends on how you implement them. Often it's enough to display the assertion inputs in the error message, and not have a 'true was false' assertion. In this case we should have the rholang code that failed.

If that's not enough for you, you can thread source position information using Li Haoyi's `sourcecode`, which we have on classpath. See `shared.Debug` for example usage.

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
lukasz-golebiewski: I'm going to do 3. which is add `buildAST(s: String)` to `ParBuilderUtil`

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
lukasz-golebiewski: The assertion checks whether execution was successful, so displaying rholang alone doesn't make sense here because the result of the execution doesn't depend solely on the passed in code.

I don't see now how I could use this library for that specific purpose. Let's discuss with the team whether we want to go ahead with this approach

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
lukasz-golebiewski: This test class is huge so 1) is going to make this PR explode. Better if this is done in a separate PR.
2a) is not possible due to how `checkData` is currently written. 
Looks to me that 2b) = 1)

ArturGajowy: >  make this PR explode

not really, and even if - it's fine as long as you keep the commits fine-grained.

> done in a separate PR.

I'm fine with that if you do that immediately after merging this. Otherwise it'd be better if you kept the new tests consistent with the (granted, poor) style the other tests have.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/StoragePrinter.scala
lukasz-golebiewski: A better question would be "if the syntax doesn't allow that, what's wrong with not throwing the exception"? :-)
If you insist though, perhaps a better place to throw would be `WaitingContinuation.create`?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
lukasz-golebiewski: > I'm fine with that if you do that immediately after merging this.

Yeah, can do
> Otherwise it'd be better if you kept the new tests consistent with the (granted, poor) style the other tests have.

I cannot use `checkData` in it's current form, so that's not possible

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
ArturGajowy: > displaying rholang alone doesn't make sense

feel free to include whatever makes sense, or what the inline assertions include.

> I don't see now how I could use this library

Have you seen its [readme](https://github.com/lihaoyi/sourcecode#sourcecode---)? It's pretty self-explanatory.

lukasz-golebiewski: Yes, it looks pretty clear, thanks. I'm still not convinced we should manually do what `scalatest` does automatically for us.
Currently we have:

```
[info] - should not remove read data *** FAILED *** (2 seconds, 664 milliseconds)                            
[info]   Vector(coop.rchain.rholang.interpreter.errors$SyntaxError: syntax error(): @ at 1:2-1:3) was not empty (PeekSpec.scala:33)
```
After your suggestion is implemented:
```
[info] - should not remove read data *** FAILED *** (2 seconds, 606 milliseconds)
[info]   Vector(coop.rchain.rholang.interpreter.errors$SyntaxError: syntax error(): @ at 1:2-1:3) was not empty (InterpreterUtil.scala:20) 
```
One simply cannot tell which evaluation caused the assertion to fail. 

After passing in implicit `sourcecode` parameters to `evaluate` the following could be displayed, but the info at the very end to which everybody is used would probably make it confusing:
```
[info] - should not remove read data *** FAILED *** (2 seconds, 391 milliseconds)
[info]   The failure was in: /home/lukasz/git/coop/rchain/rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala:32 Vector(coop.rchain.rholang.interpreter.errors$SyntaxError: syntax error(): @ at 1:2-1:3) was not empty (InterpreterUtil.scala:24)
```

lukasz-golebiewski:
bors try

rholang/src/test/scala/coop/rchain/rholang/PeekSpec.scala
ArturGajowy: Consider this: tests are read more often than the assertion messages themselves. Notice also there isn't really a tradeoff between the two readabilities, and I'd argue we can even do much better than the message that would be provided by the inline assertion.

Re: familiarity - each and every assertion error needs careful attention anyway. I'd start by catching the original error and rethrowing new one with source position or/and executed rholang info in the new message, and the old message as the `cause` of the new AssertionError.

lukasz-golebiewski: Ok, you've convinced me. I'll change this in the follow-up PR

PR #2567, opened 2019-07-03
Let syntax

JosephDenman:
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2330?jql=text%20~%20%22let%20syntax%22

### DIFF
--- a/rholang/examples/let.rho
+++ b/rholang/examples/let.rho
@@ -1,0 +1,17 @@
+new stdout(`rho:io:stdout`) in {
+
+  // prints 12
+  let a = 1 * 2 * 3, b = 1 + 2 + 3 in {
+    stdout!(a + b)
+  }
+
+  // prints (1, 2, 3)
+  | let [a, b, c] = [1, 2, 3] in {
+    stdout!((a, b, c))
+  }
+
+  // prints (1, 2)
+  | let {[1, 2] \/ ["a", "b"]} /\ [x, y] = [1, 2] in {
+    stdout!((x, y))
+  }
+}
--- a/rholang/src/main/bnfc/rholang_mercury.cf
+++ b/rholang/src/main/bnfc/rholang_mercury.cf
@@ -64,6 +64,7 @@ PContr.          Proc2  ::= "contract" Name "(" [Name] NameRemainder")" "=" "{" 
 PInput.          Proc2  ::= "for" "(" Receipt ")" "{" Proc "}" ;
 PChoice.         Proc2  ::= "select" "{" [Branch] "}" ;
 PMatch.          Proc2  ::= "match" Proc4 "{" [Case] "}" ;
+PLet.            Proc2  ::= "let" [Decl] "in" "{" Proc "}" ;
 PBundle.         Proc2  ::= Bundle "{" Proc "}" ;
 PIf.             Proc1  ::= "if" "(" Proc ")" Proc2 ;
 -- Use precedence to force braces around an interior if.
@@ -72,6 +73,10 @@ PNew.         Proc1 ::= "new" [NameDecl] "in" Proc1 ;
 PPar.         Proc  ::= Proc "|" Proc1 ;
 
 separator Proc "," ;
+
+-- Single Declaration
+DeclarationImpl. Decl ::= Proc13 "=" Proc4 ;
+separator nonempty Decl "," ;
 
 -- Process variables
 ProcVarWildcard. ProcVar ::= "_" ;
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
@@ -23,7 +23,9 @@ import scala.collection.immutable.{BitSet, Vector}
 import scala.util.Try
 
 sealed trait VarSort
+
 case object ProcSort extends VarSort
+
 case object NameSort extends VarSort
 
 object BoolNormalizeMatcher {
@@ -50,12 +52,14 @@ object GroundNormalizeMatcher {
       case gs: GroundString => Expr(GString(stripString(gs.stringliteral_))).pure[M]
       case gu: GroundUri    => Expr(GUri(stripUri(gu.uriliteral_))).pure[M]
     }
+
   // This is necessary to remove the backticks. We don't use a regular
   // expression because they're always there.
   def stripUri(raw: String): String = {
     require(raw.length >= 2)
     raw.substring(1, raw.length - 1)
   }
+
   // Similarly, we need to remove quotes from strings, since we are using
   // a custom string token
   def stripString(raw: String): String = {
@@ -290,6 +294,7 @@ object NameNormalizeMatcher {
 object ProcNormalizeMatcher {
   // FIXME before adding any more implicits, or fields to the `*VisitInputs` classes, make the normalizer use
   // ApplicativeAsk / MonadState instead
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
   def normalizeMatch[M[_]](p: Proc, input: ProcVisitInputs)(
       implicit sync: Sync[M],
       env: Map[String, Par]
@@ -1009,6 +1014,7 @@ object ProcNormalizeMatcher {
 
         def missingEnvElement(name: String, uri: String) =
           NormalizerError(s"`$uri` was used in rholang usage context where $name is not available.")
+
         if (requiresDeployId && env.get(deployIdUri).forall(_.singleDeployId().isEmpty))
           missingEnvElement("DeployId", deployIdUri).raiseError[M, ProcVisitOutputs]
         else if (requiresDeployerId && env.get(deployerIdUri).forall(_.singleDeployerId().isEmpty))
@@ -1032,6 +1038,7 @@ object ProcNormalizeMatcher {
           val errMsg = {
             def at(variable: String, sourcePosition: SourcePosition): String =
               variable + " line: " + sourcePosition.row + ", column: " + sourcePosition.column
+
             val wildcardsPositions = targetResult.knownFree.wildcards.map(at("", _))
             val freeVarsPositions = targetResult.knownFree.levelBindings.map {
               case (n, LevelContext(_, _, sourcePosition)) => at(s"`$n`", sourcePosition)
@@ -1128,6 +1135,16 @@ object ProcNormalizeMatcher {
         )
       }
 
+      case p: PLet =>
+        p.listdecl_.toList.reverse.foldM(p.proc_) {
+          case (proc, decl: DeclarationImpl) =>
+            val listCase = new ListCase()
+            listCase.add(new CaseImpl(decl.proc_1, proc))
+            (new PMatch(decl.proc_2, listCase): Proc).pure[M]
+          case _ =>
+            BugFoundError("Unexpected declaration implementation").raiseError[M, Proc]
+        } >>= (normalizeMatch(_, input))
+
       case p: PIf =>
         normalizeIfElse(p.proc_1, p.proc_2, new PNil(), input.copy(par = VectorPar()))
           .map(n => n.copy(par = n.par ++ input.par))
@@ -1153,14 +1170,17 @@ final case class ProcVisitInputs(
     env: IndexMapChain[VarSort],
     knownFree: DeBruijnLevelMap[VarSort]
 )
+
 // Returns the update Par and an updated map of free variables.
 final case class ProcVisitOutputs(par: Par, knownFree: DeBruijnLevelMap[VarSort])
 
 final case class NameVisitInputs(env: IndexMapChain[VarSort], knownFree: DeBruijnLevelMap[VarSort])
+
 final case class NameVisitOutputs(chan: Par, knownFree: DeBruijnLevelMap[VarSort])
 
 final case class CollectVisitInputs(
     env: IndexMapChain[VarSort],
     knownFree: DeBruijnLevelMap[VarSort]
 )
+
 final case class CollectVisitOutputs(expr: Expr, knownFree: DeBruijnLevelMap[VarSort])


dckc:
Shouldn't this let syntax turn into match in normalize? Is there something novel about the way it's reduced?

I stopped working on let when I realized back in May that I could write all the same things with match.
https://github.com/rchain/rchain/pull/2055#issuecomment-480853406

As I noted there, it would be much more interesting with the [pict][]-style `let x = f(a)` function call syntax.

[pict]: https://en.wikipedia.org/wiki/Pict_(programming_language)

JosephDenman:
Dan, it’s better to make add let to the Par ADT than to compile to a Match for printing purposes. The compilation process is the same though. 

As far as function syntax, when we eventually change the built-in methods to functions, we’ll be able to do exactly what you wrote. For now, you have to do something like let x = {}.toList() in { P }.

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -1013,6 +1013,129 @@ object ProcNormalizeMatcher {
...
+          (patternProcs, targetProcs)                                                        = liftedDeclarations.unzip
+          normalizedTargets                                                                  <- normalizeTargets(targetProcs)
+          (targetPars, targetParsKnownFree, targetParsLocallyFree, targetParsConnectiveUsed) = normalizedTargets
+          normalizedPatterns                                                                 <- normalizePatterns(patternProcs)
+          (patternPars, patternParKnownFreeList, patternParLocallyFreeList)                  = normalizedPatterns.unzip3
+          patternParsLocallyFree                                                             = patternParLocallyFreeList.foldLeft(BitSet())(_ | _)
+          declarationsWithLocallyFree = patternPars
+            .zip(targetPars.reverse)
+            .zip(patternParKnownFreeList)
+          sortedDeclarationsWithLocallyFree <- DeclarationsSortMatcher
```
tymm: Is there a need to sort declarations in the normalization step?

models/src/main/protobuf/RhoTypes.proto
```diff
@@ -163,6 +164,20 @@ message Match {
...
 
+message Declaration {
+    Par pattern = 1 [(scalapb.field).no_box = true];
+    Par value = 2 [(scalapb.field).no_box = true];
+    int32 freeCount = 3;
+}
+
+message Let {
+    repeated Declaration declarations = 1;
+    Par continuation = 2 [(scalapb.field).no_box = true];
```
tymm: I would vote for calling this `body` since we call it that elsewhere.

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
```diff
@@ -500,6 +504,77 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
...
+          declarations = List(
+            Declaration(EVar(FreeVar(0)), GBool(true), freeCount = 1),
+            Declaration(EVar(FreeVar(1)), GInt(5), freeCount = 1)
+          ),
+          continuation = GInt(4),
+          bindCount = 2
+        )
+      )
+    )
+    val sortedParLet = Par(
```
tymm: Please add a `Let` that has a `continuation` other than `GInt(4)` so that we can see that the `continuation` part is actually considered.

tymm (changes requested):
Not even half way through but have to stop for now because it's bedtime :)
Looks great in general but I'm not sold yet on why we want patterns in the declaration?

KentShikama (commented):
So how does this affect the matcher?

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
JosephDenman: Yes. We have to pre-sort declarations in the same way that we pre-sort receive bindings.

JosephDenman:
@KentShikama Ah.That's right. Without a match instance for let, we can't write `let let x = y in { x } = let x = 1 in { x } in { y }` lol

JosephDenman:
@tymm We want patterns in declarations so that we can do fun stuff like `let match { x | y } { 47 => Nil } = match { 1 | 2 } { 47 => Nil } in { x + y }`.

rholang/examples/let.rho
```diff
@@ -0,0 +1,6 @@
+
+new stdout(`rho:io:stdout`) in {
+  let match { x | y } { 47 => Nil } = match { 1 | 2 } { 47 => Nil } in {
```
odeac: With such an example of using "let", basically, we encourage people to write unreadable code :)

I would add a couple of simple examples and then one that shows that you can go crazy with the pattern matching

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -150,7 +150,7 @@ final case class PrettyPrinter(
   private def buildChannelStringM(p: Par, indent: Int): Coeval[String] = {
     def quoteIfNotNew(s: String): String = {
       val isBoundNew = p match {
-        case Par(_, _, _, Seq(Expr(EVarBody(EVar(Var(BoundVar(level)))))), _, _, _, _, _, _) =>
+        case Par(_, _, _, Seq(Expr(EVarBody(EVar(Var(BoundVar(level)))))), _, _, _, _, _, _, _) =>
```
odeac: thinking out loud: this line change clearly shows that there's nothing wrong with our AST :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -171,14 +171,45 @@ final case class PrettyPrinter(
...
+    t match {
       case v: Var => buildStringM(v)
       case s: Send =>
         buildChannelStringM(s.chan) |+| pure {
           if (s.persistent) "!!("
           else "!("
         } |+| buildSeq(s.data) |+| pure(")")
 
+      case l: Let =>
+        val (totalFree, declarationsString) = ((0, pure("")) /: l.declarations.zipWithIndex) {
```
odeac: For readability reasons, I prefer foldLeft instead of /: but, if others disagree, I don't mind that much. Maybe this is one item for the coding guidelines

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -171,14 +171,45 @@ final case class PrettyPrinter(
...
+                  boundShift = 0,
+                  freeId = boundId,
+                  baseId = setBaseId()
+                )
+                .buildParPattern(Seq(declaration.pattern))
+
+            (
+              declaration.freeCount + previousFree,
+              string |+| patternString |+| pure(" = ") |+| buildStringM(declaration.value) |+| pure {
+                if (i != l.declarations.length - 1) " , " else ""
```
odeac: Is this a usecase for Foldable.intercalate?

The code below seems to compile:
```scala
    val xs = List("1".pure[Coeval], "2".pure[Coeval])
    xs.intercalate(",".pure[Coeval])
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -171,14 +171,45 @@ final case class PrettyPrinter(
...
+              }
+            )
+        }
+
+        this
+          .copy(boundShift = boundShift + totalFree)
+          .buildStringM(l.continuation, indent + 1) >>= { continuationString =>
+          if (continuationString.nonEmpty) {
```
odeac: Something like this reads better:
```scala
val indentedContinuationString = if (continuationString.empty) "{ }" else {...}
pure("let ") |+| declarationsString |+| pure(indentedContinuationString)
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -1013,6 +1013,129 @@ object ProcNormalizeMatcher {
...
+            knownFreeList: List[DebruijnLevelMap[VarSort]]
+        ): M[DebruijnLevelMap[VarSort]] =
+          knownFreeList.foldM(DebruijnLevelMap[VarSort]()) {
+            case (acc, knownFree) =>
+              acc.merge(knownFree) match {
+                case (declarationsKnownFree1, Nil) =>
+                  declarationsKnownFree1.pure[M]
+                case (_, (shadowingVariable, line, col) :: _) =>
+                  val Some((_, _, firstUsageLine, firstUsageCol)) =
+                    acc.get(shadowingVariable)
```
odeac: is this guaranteed to match?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -655,6 +655,42 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+    implicit val errorLog = new ErrorLog[Task]()
+    val splitRand         = rand.splitByte(0)
+    val result = withTestSpace(errorLog) {
+      case TestFixture(space, reducer) =>
+        val pattern = Send(EVar(FreeVar(0)), List(GInt(7L), EVar(FreeVar(1))), false, BitSet())
+          .withConnectiveUsed(true)
+        val target =
+          Send(EVar(BoundVar(1)), List(GInt(7L), EVar(BoundVar(0))), false, BitSet(0, 1))
+        val matchTerm = Let(
+          Seq(Declaration(pattern, target, freeCount = 2)),
```
odeac: I would also add a test that shows let working with two declarations. Or maybe change this test and split the two bindings in two separate declarations. 

Also, I don't think the pattern matching part is interesting for this test so I would remove it in order to make the test simpler

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -330,6 +363,16 @@ final case class PrettyPrinter(
         }
     }
 
+  private def buildParPattern(patterns: Seq[Par]): Coeval[String] =
+    (pure("") /: patterns.zipWithIndex) {
+
+      case (string, (pattern, i)) =>
+        string |+| buildStringM(pattern) |+| pure {
+          if (i != patterns.length - 1) ", "
```
odeac: intercalate again

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
JosephDenman: What do you mean?

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
JosephDenman: No, because we need to carry this state of the pretty printer as we build the declarations. A simple intercalation can't do that. You'd need a `foldIntercalate` or something.

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
JosephDenman: Yes. The fact that `acc.merge(knownFree)` returns `(shadowingVariable, line, col)` means that `acc` already contains `shadowingVariable`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
JosephDenman: See comment above.

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
odeac: I was being sarcastic. I just find it difficult to read such a line

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
```diff
@@ -177,14 +177,45 @@ final case class PrettyPrinter(
...
-    val content = t match {
+    t match {
       case v: Var => buildStringM(v)
       case s: Send =>
         buildChannelStringM(s.chan) |+| pure {
           if (s.persistent) "!!("
           else "!("
         } |+| buildSeq(s.data) |+| pure(")")
 
+      case l: Let =>
```
tymm: Please add a test for pretty printing `Let`s.

tymm (changes requested):

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
```diff
@@ -500,6 +504,77 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
...
+        ),
+        Let(
+          declarations = List(
+            Declaration(
+              Connective(ConnAndBody(ConnectiveBody(List(EVar(FreeVar(0)), EVar(FreeVar(1)))))),
+              GInt(5),
+              freeCount = 2
+            )
+          ),
+          body = GString("Test"),
```
tymm: Can you change this to `GInt(5)`? Makes it easier to understand the test.
Sorry, should have said that in my initial comment.

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
```diff
@@ -500,6 +504,77 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
...
+          body = GBool(false),
+          bindCount = 2
+        )
+      )
+    )
+    val sortedParLet = Par(
+      lets = List(
+        Let(
+          declarations = List(
+            Declaration(EVar(FreeVar(0)), GBool(true), freeCount = 1),
```
tymm: Can you change `GBool(true)` to `GInt(3)`?

models/src/main/scala/coop/rchain/models/rholang/sorter/LetSortMatcher.scala
```diff
@@ -0,0 +1,40 @@
...
+      sortedTarget  <- Sortable.sortMatch(declaration.value)
+    } yield ScoredTerm(
+      Declaration(sortedPattern.term, sortedTarget.term, declaration.freeCount),
+      Node(Seq(sortedTarget.score) ++ Seq(sortedPattern.score))
+    )
+
+  def sortMatch[F[_]: Sync](term: Let): F[ScoredTerm[Let]] =
+    for {
+      sortedDeclarations  <- term.declarations.toList.traverse(sortMatchDeclaration(_))
+      sortedContinuation  <- Sortable.sortMatch(term.body)
```
tymm: Please rename to `sortedBody` for consistency.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -280,6 +282,44 @@ object Substitute {
         substituteNoSort(term).flatMap(mat => Sortable.sortMatch(mat)).map(_.term)
     }
 
+  /**
+    * Similarly to substituting the continuation of a receive, to substitute the continuation of a let expression the
+    * input environment must be shifted by the total number of variables bound by declarations.
+    *
+    */
+  implicit def substituteLet[M[_]: Sync]: Substitute[M, Let] =
```
tymm: Can you please add a test to `SubstituteTest.scala` for this?

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
odeac: Why do you say that? I don't see any dependency between the declarations of the let-construction.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
```diff
@@ -280,6 +282,44 @@ object Substitute {
...
+                              case Declaration(pattern, value, freeCount) =>
+                                for {
+                                  subPattern <- substitutePar[M].substituteNoSort(pattern)(
+                                                 depth + 1,
+                                                 env
+                                               )
+                                  subValue <- substitutePar[M].substituteNoSort(value)
+                                } yield Declaration(subPattern, subValue, freeCount)
+                            }
+          continuationSub <- substitutePar[M].substituteNoSort(term.body)(
```
tymm: Please rename to `bodySub` :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -359,6 +360,50 @@ class DebruijnInterpreter[M[_], F[_]](
...
+                                    )
+                                }
+      env1 <- substitutedDeclarations.foldM(env) {
+               case (env0, Declaration(pattern, target, freeCount)) =>
+                 spatialMatchAndCharge[M](target, pattern).map {
+                   case Some((freeMap, ())) => updateEnvironment(env0, freeMap, freeCount)
+                   case None                => env0
+                 }
+             }
+      substitutedContinuation <- substituteNoSortAndCharge(
```
tymm: `substitutedBody` :)

tymm (changes requested):
Second and last part of my review :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/PrettyPrinter.scala
JosephDenman: This is _inside_ of a fold over the declaration list. The `previousFree` value is a dependency between the declarations. I need to carry that in the fold over the declarations to correctly calculate the pattern string. I may as well build the declarations string while I'm folding over them already.

models/src/main/scala/coop/rchain/models/rholang/sorter/LetSortMatcher.scala
```diff
@@ -0,0 +1,40 @@
...
+      Let(
+        sortedDeclarations.map(_.term),
+        sortedContinuation.term,
+        term.bindCount,
+        term.locallyFree,
+        term.connectiveUsed
+      ),
+      Node(
+        Score.LET,
+        sortedDeclarations.map(_.score) ++ Seq(sortedContinuation.score) ++ Seq(
```
lukasz-golebiewski: No need to instantiate 3 `Seq`s here

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
```diff
@@ -500,6 +504,77 @@ class ParSortMatcherSpec extends FlatSpec with Matchers {
     result.term should be(sortedParMatch)
   }
 
+  it should "sort Lets based on declarations and then body" in {
```
lukasz-golebiewski: Looks to me like sorting on the body is missing

JosephDenman: You want a separate test that sorts some number of `let` statements with the same declarations but different bodies?

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
JosephDenman: What would be the purpose of doing so?

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
JosephDenman: If another review agrees that it's easier to understand the test with `GInt(5)` instead, then I'll change it 👍 Otherwise, I think that metric is too subjective to warrant modification.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -655,6 +655,78 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+        val matchTask = reducer.eval(matchTerm)(Env(), splitRand)
+        val inspectTask = for {
+          _   <- matchTask
+          res <- space.toMap
+        } yield res
+        Await.result(inspectTask.runToFuture, 3.seconds)
+    }
+    val channel: Par = GString("result")
+
+    checkData(result)(channel, Seq(GInt(1), GInt(2)), splitRand)
```
lukasz-golebiewski: Could you refrain from using `checkData`? It makes finding the failure, should it occur, much harder.
Check out the peek test in this file (after rebasing with dev)

rholang/examples/let.rho
```diff
@@ -0,0 +1,23 @@
+
+new stdout(`rho:io:stdout`) in {
```
lukasz-golebiewski, line 1: Could you add an example with both connectives?

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
lukasz-golebiewski: No, I think adding one more let here with a declaration equal to some other let but with a different body in a wrong position should be enough

rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/Costs.scala
```diff
@@ -111,6 +110,8 @@ trait Costs {
 
   final val MATCH_EVAL_COST = Cost(12, "match eval")
 
+  final val DECLARATION_EVAL_COST = Cost(2, "declaration")
```
lukasz-golebiewski: Where is this used?

JosephDenman: Good catch. It's not.

rholang/examples/let.rho
JosephDenman: Hmm. Both of which?

lukasz-golebiewski: `/\` & `\/`

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
JosephDenman: I think doing so would be redundant. Substitution is inductive, and we already know that it works for `Par`, `FreeVar`, and `BoundVar`. Additionally, we know through production usage that `Receive` is substituted correctly, despite having no tests checking that it does. `Let` uses a similar substitution pattern to `Receive`. @lukasz-golebiewski and @odeac , what do you think?

models/src/test/scala/coop/rchain/models/rholang/SortTest.scala
tymm: The purpose would be that then it is easier to see that the order of `Let`s is actually correct since `GInt(3) < GInt(4)` is easier to see than `GBool(true) < GInt(4)`.

JosephDenman: I'm adding an additional `let x = 4, y = 5 in { false }` to this test to satisfy Lukasz's request that two lets with the same declarations and different bodies are sorted in this test. This should satisfy your requirements since all integers are scored the same.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Substitute.scala
lukasz-golebiewski: I would be in favor of adding such test

tymm (approved):

lukasz-golebiewski (approved):
LGTM

rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -100,6 +101,10 @@ LinearSimple. ReceiptLinearImpl ::= [LinearBind] ;
 -- for (x <- a; y <- b if *x)
 -- LinearCond.   Linear ::= [LinearBind] "if" Proc ;
 
+-- Single Declaration
+DeclarationImpl. Decl ::= Proc "=" Proc ;
```
ArturGajowy: this is far too permissive. Since
```
let p1 = e1, ..., pn = en in { Q }

//is syntax sugar for

match (e1, ..., en) {
  (p1, ..., pn) => Q
}
```

this should be sth like
`DeclarationImpl. Decl ::= Proc13 "=" Proc4 ;`

(it should use the same parts of grammar the match uses)

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -773,6 +773,91 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
     errorLog.readAndClearErrorVector.unsafeRunSync should be(Vector.empty[InterpreterError])
   }
 
+  "let x = 1, y = 2 in { @\"result\"!(x, y) }" should "return (1, 2) to \"result\"" in {
```
ArturGajowy: these tests should not be needed, as there should be no `let` in the ast by the time it reaches the reducer.
on the other hand, an "e2e interpreter test" executing some examples of let usage in rholang should be added.

ArturGajowy (changes requested):
I don't understand the amount of code added in the reducer and other places - this should probably only touch the normalizer, and boil down to translating the AST to one using `match`-es. Granted, this takes away pretty printing and direct execution. I don't think we have the resources to cope with potential bugs in there though, or with review this big.

nzpr:
@JosDenmark what should happen if code with `let` hit the node that is not updated? I wonder should this go into 0.9.26 or better to keep in feature branch?

PR #2574, opened 2019-07-04
Rchain 3461: distribute rewards to active validators

odeac:
## Overview
At bond time the validator is added to the bonds map without becoming active. At epoch change it has the chance to become an active validator. Only the active validators are the ones who actually do validation work.

For now, all the validators become active but in the future some percentage will change at each epoch.

Note that, at first look, this implementation might seem different than what was discussed originally but, in fact, it behaves the same.

Also note that, for the moment, for testing purposes the epoch length is set to 1.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3461

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -2,6 +2,7 @@
 // minimumBond - the minimum bond allowed by the PoS
 // maximumBond - the maximum bond allowed by PoS
 // initialBonds - the initial bonds map
+// epochLength - the length of the validation epoch in blocks
 
 /*
  The table below describes the required computations and their dependencies
@@ -168,7 +169,7 @@ new  PoS,
                 for(_, @blockNumber <- blockDataCh;
                     @(pendingRewards, rewards, activeValidators, allBonds), mvarResultCh <- mvarProcessCh) {
 
-                  if (blockNumber % 1000 == 0) {
+                  if (blockNumber % $$epochLength$$ == 0) {
                     pickActiveValidators!(allBonds, *newValidatorsCh)
                   } else {
                     newValidatorsCh!(activeValidators)
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -48,14 +48,14 @@ new  PoS,
             // State structure:
             // pendingRewards : Map[PublickKey, Int] - are accummulated by calling "pay"
             // committedRewards : Map[PublickKey, Int] - are moved from pendingRewards at each closeBlock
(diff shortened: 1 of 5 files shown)


casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -9,12 +9,14 @@ import monix.eval.Coeval
 import scala.io.Source
 
 // TODO: Eliminate validators argument if unnecessary.
-final case class ProofOfStake(minimumBond: Long, maximumBond: Long, validators: Seq[Validator])
+// TODO: eliminate the default for epochLength. Now it is used in order to minimmize the impact of adding this parameter
```
KentShikama: minimize

casper/src/main/resources/PoS.rhox
```diff
@@ -40,22 +43,25 @@ new  PoS,
 
       getCurrentUserAddress!(*posRevAddressCh) |
       for(@posRevAddress <- posRevAddressCh) {
-        new stateCh, posVaultCh in {
+        new stateCh, posVaultCh, initialActiveCh in {
+          pickActiveValidators!($$initialBonds$$, *initialActiveCh) |
```
KentShikama, line 47: I think this interpolation to make all validators active as first makes this according to plan.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/TestUtil.scala
```diff
@@ -58,6 +58,7 @@ object TestUtil {
         proofOfStake = ProofOfStake(
           minimumBond = 0L,
           maximumBond = Long.MaxValue,
+          epochLength=epochLength,
```
KentShikama: Missing a space here. Did you run the tests?

KentShikama (approved):

KentShikama:
bors try

casper/src/main/resources/PoS.rhox
```diff
@@ -240,6 +253,10 @@ new  PoS,
         }
       } |
 
+      contract fst(@(first, _), resultCh) = {
+        resultCh!(first)
```
JosephDenman, line 257: Can we use an in-line match statement instead?

JosephDenman (approved):

casper/src/main/resources/PoS.rhox
odeac: not really. I need a lambda-like contract to pass to the ListOps.map

PR #2575, opened 2019-07-04
RCHAIN-3582 casper comm feedbackloop

EncodePanda:
### TL;DR

When Casper needs a block, instead of requesting that block right away from set of random peer, it will ask set of peers if they have it, and then request to the first one that answers. Solution tries to cover all corner cases - scenarios when things go wrong.

This should dramatically limit number of messages being exchanged between nodes.

### Description

When `Casper` is missing some dependencies (missing some blocks), it will ask for it. Underneath it is  using a method `CommUtil.sendBlockRequest` to request blocks.

```
[info] CommUtil
[info]   sendBlockRequest
[info]     if given block was not yet requested
[info]     - broadcast HasBlockRequest to random peers (402 milliseconds)
[info]     - should log to INFO that request was made (2 milliseconds)
[info]     if given block was already requested
[info]     - should do nothing (3 milliseconds)
```

Only node in state `Running` will answer messages of type `HasBlockRequest`

```
[info] Running
[info]   handleHasBlockRequest
[info]     if given block is stored
[info]     - should send back HasBlock message to the sender (275 milliseconds)
[info]     if given block is not stored in BlockStore
[info]     - should do nothing (42 milliseconds)
```

Only node in state `Running` will answer messages of type `HasBlock`

```
[info] Running.handleHasBlock
[info] - should not request block if casper contains block (259 milliseconds)
[info] - should ignore if requested before for the same peer (9 milliseconds)
[info] - should store on a waiting list and don't request if requested by different peer (9 milliseconds)
[info] - should request block and store information about requested block (126 milliseconds)
```

Entries from `requested blocks` must be "maintained". If a block was requested, but data is not received in some defined timeout, a request must be sent to a peer from a waiting-list.

```
[info] Running
[info]   maintainRequestedBlocks, for every block that was requested
[info]     if block request is still within a timeout
[info]     - should keep the request not touch (304 milliseconds)
[info]     if block was not delivered within given timeout
[info]       if waiting list is not empty
[info]       - should request block from first peer on a waiting list (128 milliseconds)
[info]       - should move that peer from the waiting list to the requested set (3 milliseconds)
[info]       - timestamp is reset (2 milliseconds)
[info]       if waiting list IS empty
[info]       - should log a warning to the user (25 milliseconds)
[info]       - should NOT send requests to other peers (1 millisecond)
[info]       - should remove the entry from the requested block lists (1 millisecond)
```

Entries from `requested blocks` can be removed once the block is added by casper

```
[info] Running
[info]   handleBlockMessage
[info]   - should remove entry from requested blocks once attempted to add to Casper (290 milliseconds)
```

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3582


casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -108,7 +108,7 @@ private class SleepingMultiParentCasperImpl[F[_]: Monad: Time](underlying: Multi
       b: BlockMessage,
       handleDoppelganger: (BlockMessage, Validator) => F[Unit]
   ): F[BlockStatus]                                           = underlying.addBlock(b, ignoreDoppelgangerCheck[F])
-  def contains(b: BlockMessage): F[Boolean]                   = underlying.contains(b)
```
ArturGajowy, line 103: Why this change? ~~Please provide an overload that allows for passing the block itself too and delegates to the BlockHash-taking one.~~

Please also restore the previous usages of the block-taking method - I find the change from `.contains(block)` to `.contains(block.blockHash)` is downgrade in readability ~~and a needless change~~.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperAddBlockSpec.scala
```diff
@@ -313,30 +313,32 @@ class MultiParentCasperAddBlockSpec extends FlatSpec with Matchers with Inspecto
...
         signedBlock2 <- nodes(1).addBlock(deployDatas(2))
         signedBlock3 <- nodes(2).addBlock(deployDatas(3))
         _            <- nodes(2).transportLayerEff.clear(nodes(2).local) //nodes(2) ignores block2
-        _            <- nodes(1).receive() // receives block3; asks for block1'
-        _            <- nodes(2).receive() // receives request for block1'; sends block1'
+        _            <- nodes(1).receive() // receives block3; asks if has block1'
+        _            <- nodes(2).receive() // receives request has block1'; sends i have block1'
+        _            <- nodes(1).receive() // receives has block1 ack; asks for block1'
+        _            <- nodes(2).receive() // receives request block1'; sends block1'
         _            <- nodes(1).receive() // receives block1'; adds both block3 and block1'
```
ArturGajowy, line 314: Ideally, we'd have a method `exchangeTillSilence(nodes)` that would call receive on all passed nodes in round-robin fashion until all of them have empty message queues. It could have a default iterations threshold/we could lean on the test timeout for cases where silence doesn't happen.

Less ideal, but another option would be to exchange a `requestAndReceiveBlock(node1, node2)` that would cover the 4 or 5 receives above, along with their comments

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
ArturGajowy: LOL sorry it occurred to me only under the shower that you're making a production code change (which is most likely justified - didin't notice it previously) - and I'm referring to test code.

I think what should happen is: we get a `HashSetCasperTestNode#contains(BlockMessage)` method that does the convenience of grabbing the hash and delegates to `casperEff.contains(hash)`.

So a change to all the test call-sites will be needed anyway. I'd just prefer the change to be from:
`node.casperEff.contains(block)`
to
`node.contains(block)`
vs
`node.casperEff.contains(block.blockHash)`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperAddBlockSpec.scala
EncodePanda: ideally we dont want to have tests that look like above

ArturGajowy: Baby steps, my mate, baby steps :joy:

Had a broader look and I admit the .clear usage seem quite hacky. Re: validateBlockStore part - this will be simplified once you rebase on dev :)

EncodePanda:
@ArturGajowy @sebtomba this is now ready for review

EncodePanda:
bors try

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
EncodePanda: Well, the change was motivated by the fact that Casper only needs `hash` to check if the block exists or not. Providing the whole `BlockMessage` as an argument only to check if there is a block with given hash was bad design. Simple as that.

EncodePanda:
bors try

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -22,14 +22,19 @@ option (scalapb.options) = {
...
 
 message BlockRequest {
   option (scalapb.message).extends = "CasperMessage";
-  //Fields are redundant to allow for validity check.
-  //Such a check can definitively distinguish this
-  //from other message types with similar serializations.
-  string base16Hash = 1;
-  bytes  hash       = 2;
+  bytes  hash       = 1;
 }
```
zsluedem, line 38: You change the protocol.
Right now the integration test would simulate the way how a rnode request a block and send a block. So if you change the protocol, the test would fail.
According to the log https://drone.rchain-dev.tk/rchain/rchain/4641/7 line 1230, the integration test can not request the block right because the `hash` is messing around with the `base16Hash`.

But right now, the integration test is using protocol which stays in  [pyrchain](https://github.com/rchain/pyrchain). I need to updated the lib first. So I RECOMMEND you to mark the test xfail first. I create the ticket for that to fix it.

zsluedem: https://rchain.atlassian.net/browse/RCHAIN-3663 ticket for that.

EncodePanda: Thx @zsluedem

EncodePanda:
bors try

EncodePanda:
it's surprising, but bors did not use latest two commits 🤔

EncodePanda:
bors try

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -8,84 +8,217 @@ import cats.{Applicative, Monad}
...
+            RequestedBlocks[F].modify(
+              _ + (hb.hash -> Requested(timestamp = ts, peers = Set(peer)))
+            )
+          }
+      )
+
+    def addToWaitingList(requested: Requested): F[Unit] =
+      Log[F].info(s"Adding $peer to waiting list of $hashStr request") >> RequestedBlocks[F]
+        .modify { requestedBlocks =>
+          requestedBlocks + (hb.hash -> requested.copy(waitingList = peer :: requested.waitingList))
```
KentShikama, line 140: We should add a TODO or a ticket to append last

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -8,84 +8,217 @@ import cats.{Applicative, Monad}
...
+      casperContains: BlockHash => F[Boolean],
+      casperAdd: BlockMessage => F[BlockStatus]
+  ): F[Unit] =
+    casperContains(b.blockHash)
+      .ifM(
+        Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
+        for {
+          _ <- Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.")
+          _ <- casperAdd(b)
+          _ <- RequestedBlocks[F].modify(_ - b.blockHash)
```
KentShikama, line 179: We should add a TODO or a ticket to check BlockStatus for inDag and only remove from RequestedBlocks if so

KentShikama (approved):

EncodePanda:
bors try

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
EncodePanda: https://rchain.atlassian.net/browse/RCHAIN-3675

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
EncodePanda: https://rchain.atlassian.net/browse/RCHAIN-3675

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -76,10 +76,10 @@ object MultiParentCasper extends MultiParentCasperInstances {
   def ignoreDoppelgangerCheck[F[_]: Applicative]: (BlockMessage, Validator) => F[Unit] =
     kp2(().pure[F])
 
-  def forkChoiceTip[F[_]: MultiParentCasper: Monad: BlockStore]: F[BlockMessage] =
+  def forkChoiceTip[F[_]: Monad: BlockStore](casper: MultiParentCasper[F]): F[BlockMessage] =
```
sebtomba, line 80: Todo: We need to check where else `MultiParentCasper` is used like a typeclass.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -85,26 +85,32 @@ object Running {
       Map(filtered: _*)
     }
 
+    def tryRerequest(hash: BlockHash, requested: Requested): F[(BlockHash, Option[Requested])] =
+      if (requested.waitingList.nonEmpty) {
```
sebtomba, line 89: Please consider `headOption.fold` here.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -8,84 +8,217 @@ import cats.{Applicative, Monad}
...
-class Running[F[_]: RPConfAsk: BlockStore: Monad: ConnectionsCell: TransportLayer: Log: Time](
-    private val casper: MultiParentCasper[F],
+  **/
+object Running {
+
+  val timeout: FiniteDuration = 240 seconds
+
+  final case class Requested(
+      timestamp: Long,
+      peers: Set[PeerNode] = Set.empty,
```
sebtomba, line 46: It seems `peers` are not used anywhere. You are moving peers from the waiting list to this set. But you could also just drop the peers.

sebtomba (approved):
LGTM. But see my comments.

casper/src/main/scala/coop/rchain/casper/Casper.scala
EncodePanda: https://rchain.atlassian.net/browse/RCHAIN-3676

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
EncodePanda: 👍

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
EncodePanda:
```

    "ignore if requested before for the same peer" in {
      // given
      val sender = peerNode("somePeer", 40400)
      val requestedBefore =
        Map(hash -> Requested(timestamp = System.currentTimeMillis, peers = Set(sender)))
      implicit val requestedBlocks = initRequestedBlocks(init = requestedBefore)
      val casperContains           = alwaysFalse
      // when
      Running.handleHasBlock[Coeval](sender, hb)(casperContains).apply()
      // then
      transport.requests shouldBe empty
    }
```

PR #2576, opened 2019-07-05
Remove ChunkResponse from the streaming API

sebtomba:
## Overview
Remove ChunkResponse from the streaming API because it's misleading. Instead, use `TLResponse` and respond with `NoResponse` when the streaming was successful and with `InternalCommunicationError` otherwise.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3605

## Note
This is a breaking change in the node to node protocol.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
@@ -68,13 +68,15 @@ object GrpcTransportReceiver {
             Opened(StreamHandler.StreamError.circuitOpened)
           else Closed
 
-        def stream(observable: Observable[Chunk]): Task[ChunkResponse] = {
+        def stream(observable: Observable[Chunk]): Task[TLResponse] = {
           import StreamHandler._
           import StreamError.StreamErrorToMessage
 
-          (handleStream(tempFolder, observable, circuitBreaker) >>= {
-            case Left(error @ StreamError.Unexpected(t)) => logger.error(error.message, t)
-            case Left(error)                             => logger.warn(error.message)
+          handleStream(tempFolder, observable, circuitBreaker) >>= {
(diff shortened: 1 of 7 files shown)


EncodePanda (approved):
rename `NoResponse` to `Ack` because since we don't hafve roundtrip anymore '"no response"' is misleading. otherwise 👍

comm/src/test/scala/coop/rchain/comm/transport/GrpcTransportSpec.scala
```diff
@@ -56,10 +56,10 @@ class GrpcTransportSpec extends WordSpecLike with Matchers with Inside {
       sendMessages += request
       response
     }
-    def stream(input: Observable[Chunk]): Task[ChunkResponse] =
+    def stream(input: Observable[Chunk]): Task[TLResponse] =
       input.toListL.map { l =>
         streamMessages += l
-        ChunkResponse()
+        noResponse
```
tymm: I think this should be renamed to `ack` as well.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
```diff
@@ -89,8 +91,8 @@ object GrpcTransportReceiver {
                       metrics.incrementCounter("dropped.packets"),
                       msg.path.deleteSingleFile[Task]
                     ).sequence
-                  )
-          }).as(ChunkResponse())
+                  ) >> rPConfAsk.reader(c => noResponse(c.local))
```
tymm: See previous comment.

tymm (changes requested):

comm/src/test/scala/coop/rchain/comm/transport/GrpcTransportSpec.scala
sebtomba: @tymm done

tymm (approved):

PR #2578, opened 2019-07-05
Modify conflicts so that volatile events do not conflict 

KentShikama:
## Overview

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3611

For example, two blocks with the contents

```scala
@"hi"!(5) | for (_ <- @"hi") { Nil }
```

no longer conflict as it has no effect on the trie.

### Notes


casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,13 +86,47 @@ object EstimatorHelper {
       ancestors = maybeAncestors.flatten
       ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
         ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))
+      commEvents = ancestorEvents.filter {
+        case Event(Comm(CommEvent(Some(_: ConsumeEvent), _))) =>
+          true
+        case _ => false
+      }
+      producesInCommEvents = commEvents.flatMap {
+        case Event(Comm(CommEvent(Some(_: ConsumeEvent), produces))) =>
```
odeac: this repeating pattern is not nice. At line 89 you can extract the commEvent from the Event(Comm(CommEvent(_))) pattern and then partition them in produces and consumes. Then you only work with the CommEvent objects, which would be simpler

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,11 +61,19 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
 
   it should "handle multi-parent blocks correctly when they operate on stdout" ignore effectTest {
     def echoContract(no: Int) = s"""new stdout(`rho:io:stdout`) in { stdout!("Contract $no") }"""
-    val time                  = System.currentTimeMillis()
+    testConflictOnSimpleDiamondDag(echoContract)
+  }
+
+  it should "handle multi-parent blocks correctly when they operate on volatile produce/consume pairs" in effectTest {
+    testConflictOnSimpleDiamondDag((no: Int) => s"""@"hi"!($no) | for (_ <- @"hi") { Nil }""")
```
odeac: I might misunderstand the concept of "volatile" but, since the messages are sent to a forgeable channel, is this still volatile?

ArturGajowy (commented):
Just a heads-up for others: I think there are cases when this is not correct, and I'll try to provide counterexamples today. I'd also like us to have more tests for that, partly based on Mike Stay's [Mergeable?](https://docs.google.com/spreadsheets/d/1pABqArF9e8HRTO9zSefp93mIVUm91avekeDgqSEw0R8/edit?ts=5c818e78#gid=0) spreadhseet

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,13 +86,47 @@ object EstimatorHelper {
...
         case Event(Consume(consume: ConsumeEvent)) =>
-          consume.channelsHashes.toSet
+          if (consumeInCommEvents.contains(consume)) {
+            Set.empty[ByteString] // Volatile consume
+          } else {
+            consume.channelsHashes.toSet
+          }
         case Event(Comm(CommEvent(Some(consume: ConsumeEvent), produces))) =>
-          consume.channelsHashes.toSet ++ produces.map(_.channelsHash).toSet
+          val consumeChannelHashes = if (consumeInCommEvents.contains(consume)) {
```
ArturGajowy: this condition is always true, and so is the one for produces in line 126. (that's because we gather consumeInCommEvents in exactly the sane way). This means no `acnestorChannels` are going to be reported from CommEvents. This is wrong, as a COMM-event could span across two deploys, and in this case the channels from the side that's no from this deploy (be it produce or consume) should be added to 'ancestorE, vents' to cause conflicts.

Example test cases:
```scala
checkConflicts("@0!(0)", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }")
checkNoConflicts("0", "for (_ <- @0) { 0 } | @0!(1)", "for (_ <- @0) { 0 } | @0!(2)")
checkConflicts("for (_ <- @0) { 0 }", "@0!(1)", "@0!(2)")
checkNoConflicts("for (_ <- @0) { 0 }", "@0!(1)", "for (@2 <- @0) { 0 } | @0!(2)")

//cases that are conflicting now but should not once we have more merge support:
checkConflicts("0", "@0!(0)", "@0!(0)")
checkConflicts("0", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }")
```

Both test methods take the deploy code for blocks BASE, A, B respectively in a dag:
```
A  B
 \/
BASE
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,11 +61,19 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+  }
+
+  private def testConflictOnSimpleDiamondDag(contract: Int => String): Effect[Unit] = {
+    val time = System.currentTimeMillis()
     HashSetCasperTestNode.networkEff(genesis, networkSize = 2).use { nodes =>
       val deploys = Vector(
-        ConstructDeploy.sourceDeploy(echoContract(1), time + 1, accounting.MAX_VALUE),
-        ConstructDeploy.sourceDeploy(echoContract(2), time + 2, accounting.MAX_VALUE)
+        ConstructDeploy.sourceDeploy(contract(1), time + 1, accounting.MAX_VALUE),
+        ConstructDeploy.sourceDeploy(contract(2), time + 2, accounting.MAX_VALUE)
```
ArturGajowy: We've got way too few tests for conflict detection, and they should not be made in this class - at least not primarily. Each such test takes 1.5 s here, could take 15ms in the other class.

Please introduce the test methods I've menioned above in the EstimatorHelperTest, and go through each case of first 5 column's of Mike Stay's spreadsheet I've linked above - this will amount to ~30 cases:

15 cells from the spreadsheet * 2 for whether the 'existing' side of the match is in BASE or in one of the children (let's always make it A in the test data and make the test methods test twice by swapping the test data).

ArturGajowy (changes requested):
We need much more tests, both for positive and negative cases, including the cases that are going to change after better merge support.

We need to better discern whether a COMM-event stays local to a deploy. Currently no COMM event can contribute to a conflict (see below), which is obviously wrong.

We also need to structure the prod code better to avoid duplication, noise, and throwing exceptions - I save that for the second round though.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
KentShikama: Why does

`checkConflicts("@0!(0)", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }")` conflict?

KentShikama:
bors try

KentShikama:
bors try

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
KentShikama: Good catch; it was careless on my part to not take the produces/consumes from the event log and instead take them from the comm events.

KentShikama:
bors try-

KentShikama:
bors try

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
KentShikama: I think the merge spec is the place these tests should be as long as they are using networkEff.

KentShikama:
The checks aren't labelled by the mergeable spreadsheet yet but PTAL @ArturGajowy

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
KentShikama: Volatile just mean that it doesn't affect the trie at the end. Volatility depends on the execution - if someone else sent to the forgeable name then it may not be volatile.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
KentShikama: PTAL

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,32 +63,72 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      checkConflicts("0", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }")
+  }
+
+  private def checkConflicts(base: String, b1: String, b2: String): Effect[Unit] = {
+    val time = System.currentTimeMillis()
+    val deploys = Vector(
+      ConstructDeploy.sourceDeploy(base, time + 1, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy(b1, time + 2, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy(b2, time + 3, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy("Nil", time + 4, accounting.MAX_VALUE)
```
ArturGajowy: Please no longer use `System.currentTimeMillis` in the test deploys, as it makes the test fail in weird ways once they start depending on the time being consistent. If controlling the time is not the topic of that test, we should always use `sourceDeployNowF`.

Similarly with the phloLimit. This is another aspect that used to get in the way *all the time* when working on charging for deploy. `sourceDeployNowF` takes care of that properly, by passing a phloLimit that is within reach of the default deployer, but high enough for most deploys.

The general takeaway from this is: tests should never specify a single value they don't care about. If a value is not crucial in the test setup and just needs a default - the test factory method (like `sourceDeployNowF`) should provide it.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,13 +86,54 @@ object EstimatorHelper {
...
+        case ((produces, consumes, commEvents), ancestorEvent) =>
+          ancestorEvent match {
+            case Event(Produce(produce: ProduceEvent)) =>
+              (produces :+ produce, consumes, commEvents)
+            case Event(Consume(consume: ConsumeEvent)) =>
+              (produces, consumes :+ consume, commEvents)
+            case Event(Comm(commEvent: CommEvent)) =>
+              (produces, consumes, commEvents :+ commEvent)
+            case _ =>
+              throw new RuntimeException("Unexpected ancestor event")
```
ArturGajowy: This should say: `s'Unknown ancestorEvent type for: $ancestorEvent'`.
Most of the time we should include in the error message the data that causes it.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,32 +63,72 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+  }
+
+  it should "handle multi-parent blocks correctly when they operate on volatile produce/consume pairs" in effectTest {
+    checkNoConflicts(
+      "Nil",
+      """@"hi"!(1) | for (_ <- @"hi") { Nil }""",
+      """@"hi"!(2) | for (_ <- @"hi") { Nil }"""
+    ) >>
+      checkConflicts("@0!(0)", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }") >>
+      checkNoConflicts("0", "for (_ <- @0) { 0 } | @0!(1)", "for (_ <- @0) { 0 } | @0!(2)") >>
```
ArturGajowy: Sorry for the commotion, but let's name those:
```
checkConflict
checkMerge
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,32 +63,72 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    val deploys = Vector(
+      ConstructDeploy.sourceDeploy(base, time + 1, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy(b1, time + 2, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy(b2, time + 3, accounting.MAX_VALUE),
+      ConstructDeploy.sourceDeploy("Nil", time + 4, accounting.MAX_VALUE)
+    )
+    diamondConflictCheck(deploys, numberOfParentsForDiamondTip = 1)
+  }
+
+  private def checkNoConflicts(base: String, b1: String, b2: String): Effect[Unit] = {
```
ArturGajowy: we need both our `check` methods to test the dags:
```
Nil
/  \
A  B
\  /
BASE
```
and 
```
Nil
/  \
B  A
\  /
BASE
```
so that we see the mergability relation is symmetric (we cut the number of test cases to specify explicitly by 2 this way)

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,13 +86,54 @@ object EstimatorHelper {
...
+          }
+      }
+      producesInCommEvents = commEvents.flatMap {
+        case CommEvent(Some(_: ConsumeEvent), produces) =>
+          produces
+        case _ =>
+          throw new RuntimeException("Unexpected comm event")
```
ArturGajowy: I don't think this case is needed here, as we have a list of CommEvents here - the type system guards this, and the totality checker should see that. There must be a way to have a total function commEvents -> producesInCommEvents
...
Ah, I see... Damn protobuf and its option-ridden garbage output.

The error should then say `Invalid comm event $c`, where `c` is part of the pattern in the first case `c @ CommEvent(...)`

ArturGajowy (changes requested):
I like where this is going :) Please add the remaining test cases though, and the labels as discussed in discord.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
odeac: Still I'm not happy with the output. I tried some refactoring on my machine and I think it looks readable and it seems to compile but I haven't run the tests. Do you think this works?

Also we could move the extractors to some common place where they could be reused.

```scala
  val extractProduce : Event => Option[ProduceEvent] = {
    case Event(Produce(produce: ProduceEvent)) => Some(produce)
    case _ => None
  }


  val extractConsume : Event => Option[ConsumeEvent] = {
    case Event(Consume(consume: ConsumeEvent)) => Some(consume)
    case _ => None
  }

  val extractComm : Event => Option[CommEvent] = {
    case Event(Comm(commEvent: CommEvent)) => Some(commEvent)
    case _ => None
  }

  private def buildBlockAncestorChannels[F[_]: Monad: BlockStore](
      blockAncestorsMeta: List[BlockMetadata]
  ): F[Set[ByteString]] =
    for {
      maybeAncestors <- blockAncestorsMeta.traverse(
                         blockAncestorMeta => BlockStore[F].get(blockAncestorMeta.blockHash)
                       )
      ancestors = maybeAncestors.flatten
      ancestorEvents =
        ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
        ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))

      produceEvents = ancestorEvents.map(extractProduce).flatten
      consumeEvents = ancestorEvents.map(extractConsume).flatten
      commEvents = ancestorEvents.map(extractComm).flatten

      producesInCommEvents = commEvents.flatMap(_.produces)

      consumeInCommEvents = commEvents.flatMap(_.consume)

      ancestorProduceChannels =
        produceEvents
          .filter(producesInCommEvents.contains(_))
          .map(p => p.channelsHash)
          .toSet

      ancestorConsumeChannels =
        consumeEvents
          .filter(consumeInCommEvents.contains(_))
          .flatMap(c => c.channelsHashes)
          .toSet

      ancestorCommChannels =
        commEvents
          .filter(ce => consumeEvents.contains(ce.consume) && ce.produces.forall(produce => produceEvents.contains(produce)))
          .flatMap(c => c.consume.toSeq.map(_.channelsHashes) ++ c.produces.map(_.channelsHash))
          .toSet

      ancestorChannels = ancestorProduceChannels ++ ancestorConsumeChannels ++ ancestorCommChannels
    } yield ancestorChannels
```

KentShikama:
bors try

KentShikama:
bors try

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,32 +63,111 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+  }
 
-        //multiparent block joining block0 and block1 since they do not conflict
-        multiparentBlock <- nodes(0).addBlock(deploys(1))
-        _                <- nodes(1).receive()
+  // TODO: Peek rows/column
+  // Note this skips pairs that lead to infinite loops
+  it should "handle multi-parent blocks correctly when they operate on volatile produce/consume pairs" in effectTest {
+    // 4! 4!
+    checkConflicts("@0!(0)", "for (_ <- @0) { 0 }", "for (_ <- @0) { 0 }") >>
```
ArturGajowy: Just to not forget it: we should also have:
```scala
checkConflict("@0!(0)", "for (@x <- @0) { @0!(x) }", "for (@x <- @0) { @0!(x) }")
checkMerge("@0!(0) | @0!(1)", "for (@0 <- @0) { 0 }", "for (@1 <- @0) { 0 }")
checkMerge("@0!(0)", "for (@x <- @0) { @0!(x) }", "@0!(1) | for (@1 <- @0) { 0 }")

//will merge or conlflict in the future - we'll need to split that into more test cases then
checkConflict("@0!(0)", "for (@0 <- @0) { @0!(x) }", "@0!(0) | for (@0 <- @0) { @0!(0) }") 

//will merge in the future
checkConflict("@0!(0)", "for (@x <- @0) { @0!(x) }", "@0!(1) | for (@1 <- @0) { @0!(1) }")
checkMerge("0", "@0!(0) | for (@0 <- @0) { @0!(x) }", "@0!(0) | for (@0 <- @0) { @0!(0) }")
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -58,32 +63,111 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      // !4 4!!
+      checkMerge("for (_ <- @0) { 0 } | @1!!(0)", "@0!(0)", "for (_ <- @1) { 0 }") >>
+      // !4 !!X
+      checkMerge("for (_ <- @0) { 0 }", "@0!(0)", "@1!!(0)") >>
+      // !4 !!4
+      checkConflicts("for (_ <- @0) { 0 } | for (_ <- @0) { 0 }", "@0!(0)", "@0!!(0)") >>
+      // !4 CX (TODO: Check if this could conflict)
+      checkMerge("for (_ <- @0) { 0 }", "@0!(1)", "contract @1(id) = { 0 }") >>
+      // !4 C!
+      checkMerge("for (_ <- @0) { 0 } | for (_ <- @1) { 0 }", "@0!(0)", "contract @1(id) = { 0 }")
```
ArturGajowy: I didn't mean for you to go through the whole sheet :) It's good to have more tests though.

At the same time, we should for now probably focus the combinations:
`{!X, !4, !C , 4X, 4!}^2`, maybe extending it to `{!X, !4, !C , 4X, 4!, CX, C!}^2`

ArturGajowy: Notice there's no `!!` in the combinations above. Let's keep the tests for `!!` you've added, but not add more - unless you see a need.

KentShikama:
bors try

KentShikama:
bors try

KentShikama:
PTAL

KentShikama:
bors try

ArturGajowy:
@Kent please rebase on upstream/dev to avoid the spurious merge commit. Thx!

KentShikama:
It isn't spurious: I've resolved some conflicts that resulted

KentShikama:
I prefer rebases before the PR review process starts but when it is towards the end a rebase makes it such that you have to go through all the commits again.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -61,30 +61,39 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
-        _      <- nodes(1).receive()
+        _ <- nodes(0).addBlock(deploys(0))
+        _ <- nodes(0).receive()
+        _ <- nodes(1).receive()
+        _ <- nodes(0).addBlock(deploys(1))
+        _ <- nodes(1).addBlock(deploys(2))
+        _ <- nodes(0).receive()
+        _ <- nodes(1).receive()
+        _ <- nodes(0).receive()
+        _ <- nodes(1).receive()
```
ArturGajowy: I think this can be reduced to

```scala
        _ <- nodes(0).addBlock(deploys(0))
        _ <- nodes(1).receive()
        _ <- nodes(0).addBlock(deploys(1))
        _ <- nodes(1).addBlock(deploys(2))
        _ <- nodes(0).receive()
        multiparentBlock <- nodes(0).addBlock(deploys(3))
```

the mutliparent block is not needed on node1

ArturGajowy: Or we could add one more `nodes(1).receive()` and keep testing it on both nodes - still, some of those receives are not needed

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
-            Set.empty[ByteString] // Volatile consume & produces
-          } else {
-            consume.channelsHashes.toSet ++ produces.map(_.channelsHash).toSet
-          }
-        case _ => throw new RuntimeException("incorrect ancestor events")
-      }.toSet
+
+      produceEvents = ancestorEvents.map(extractProduce).flatten
+      consumeEvents = ancestorEvents.map(extractConsume).flatten
+      commEvents    = ancestorEvents.map(extractComm).flatten
```
ArturGajowy, line 107: `map(...).flatten` is `flatMap`

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
+
+      producesInCommEvents = commEvents.flatMap(_.produces)
+      consumeInCommEvents = commEvents.flatMap(_.consume)
+
+      ancestorProduceChannels = ancestorEvents
+        .map(extractProduce)
+        .flatten
```
ArturGajowy, line 114: this is `produceEvents`

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
+        .filterNot(producesInCommEvents.contains(_))
+        .map(p => p.channelsHash)
+        .toSet
+
+      ancestorConsumeChannels = ancestorEvents
+        .map(extractConsume)
+        .flatten
```
ArturGajowy, line 121: this is `consumeEvents`

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
+        .filterNot(consumeInCommEvents.contains(_))
+        .flatMap(c => c.channelsHashes)
+        .toSet
+
+      ancestorCommChannels = ancestorEvents
+        .map(extractComm)
+        .flatten
```
ArturGajowy, line 128: this is `commEvents`

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
+        .filterNot {
+          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+            consumeEvents.contains(consume) && produces.forall(
+              produce => produceEvents.contains(produce)
+            )
+          case _ => false
```
ArturGajowy, line 134: please extract `isVolatile(commEvent: CommEvent) and use it here.

The `case _ =>` should be throwing an exception.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,55 +101,46 @@ object EstimatorHelper {
...
+        }
+        .flatMap {
+          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+            consume.channelsHashes ++ produces.map(_.channelsHash)
+          case e @ CommEvent(None, _) =>
+            throw new RuntimeException(s"Unexpected comm event $e")
```
ArturGajowy, line 140: this is somewhat late to raise this exception. Also, the message should say what was wrong with that comme vent (i didn't have consumes, so was gibberish)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -66,6 +66,8 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
     checkMerge("Nil", echoContract(1), echoContract(2))
   }
 
+  // TODO: Peek rows/column
+  // Note this skips pairs that lead to infinite loops
```
ArturGajowy, line 68: we should include tests for those ultimately, as the merge result for them is clear: they're mregable. So this should be another TOOO :)

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,14 +101,46 @@ object EstimatorHelper {
...
-          consume.channelsHashes.toSet ++ produces.map(_.channelsHash).toSet
-        case _ => throw new RuntimeException("incorrect ancestor events")
-      }.toSet
+
+      produceEvents = ancestorEvents.map(extractProduce).flatten
+      consumeEvents = ancestorEvents.map(extractConsume).flatten
+      commEvents    = ancestorEvents.map(extractComm).flatten
+
+      producesInCommEvents = commEvents.flatMap(_.produces)
+      consumeInCommEvents  = commEvents.flatMap(_.consume)
```
ArturGajowy, line 110: those two should not simply 'flatMap' - they should raise errors wherever the consume or the produces are empty

ArturGajowy (changes requested):
I've found some issues with the prod code, going to go trhough tests now

ArturGajowy:
@KentShikama I've just noticed we won't be able to have volatile events merging without the ability to discern persistent sends and receives. For example:

```scala
// notice I've moved the 'base' parameter to the last position (!)
      // !X (!C)
      checkConflicts("@0!(0)", "@0!(1) | contract @0(id) = { 0 }", "0"),

      checkMerge("@0!(0)", "for (_ <- @0) { 0 } | @0!!(0)", "0"),
      //FIXME it might be crucial that we consider interaction with persistent things, e.g.
      // 4X 4!! vs 4X (4!!) - the former merges safely, the latter must not merge, as the !! could match both 4s
```

Long story short, when a persistent send/receive in a COMM event means the COMM is no longer volatile, because the send/receive is going to be added to the TS.

But this is good, we're going to have the merge rule for contracts with the same effort.

ArturGajowy:
So I guess this ticket merges with https://rchain.atlassian.net/browse/RCHAIN-3638, @KellyatPyrofex

ArturGajowy:
I'll push an update to tests that makes working with them easier later this day

KellyatPyrofex:
> So I guess this ticket merges with https://rchain.atlassian.net/browse/RCHAIN-3638, @KellyatPyrofex

@ArturGajowy, do you mean the work described in RChain-3638 is covered in this PR and could be closed as done once this PR merges?

ArturGajowy:
I mean for this PR to be merged without introducing bugs, the work in that ticket needs to happen in this PR

ArturGajowy:
I've modified the test in several aspects:

1. only the `@0` channel should be used - interactions on different channels are trivially mergable. Adding some tests for that (preferably: generative) makes sense, but can wait for a separate PR.
2. In some cases the rholang was not in line with the test case (e.g. C! C! without a single `contract` in rho)
3. In plenty cases, our assertion for now needs to be 'checkConflicts'.
4. I've mainly been adding cross-deploy variants (covered by Mike's sheet directly). We need to add in-deploy variants to test volatile events, e.g.:
  - 4X !X - has no in-deploy variants
  - 4! CX - has one in-deploy variant, (4!) CX
  - 4! 4! - has two in-deploy variants: (4!) 4! and (4!) (4!)
5. As mentioned above, in cases where a persistent read/write happens in-deploy, it's not a volatile event (TS gets a new persistent thingy). Those test cases are marked as 'checkMerges', but this is a bug and a `FIXME: THIS NEEDS TO CONFLICT` stands near them. **These must be fixed before we merge this PR**, or we end up with corrupted TS.
6. I've parallelized the test for it to run faster, and added better assertion message, with a code pointer. Make sure to install `Awesome Console` in idea to be able to click the pointer to the failing line.
7. I've trimmed down unnecessary execution that is going to be covered by randomizing the side/node on which deploy A and B happen. (see the WIP commit)

Please continue work from here, I'll do the needed rebases before we merge.

ArturGajowy:
One more thing: to steer which sends and receives match together, data needs to be used. See for example tests for `4! C!`:
```scala
     // 4! C! may merge or not
      // different ! (TODO: Eventually should merge)
      checkConflicts("for (@0 <- @0) { 0 }", "contract @0(@1) = { 0 }", "@0!(0) | @0!(1)"),
      // same !
      checkConflicts("for (_ <- @0) { 0 }", "contract @0(id) = { 0 }", "@0!(0)"),
```

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
KentShikama: @odeac

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
odeac: that's what I also thought but flatMap didn't compile here

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
odeac: the original implementation didn't fail on empty consumes or produces. I only suggested a way to write that such that it's more readable

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +59,153 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    // For-s (linear receives)
+    val F_ = Rho("for (_ <- @0) { 0 }")
+    val F0 = Rho("for (@0 <- @0) { 0 }")
+    val F1 = Rho("for (@1 <- @0) { 0 }")
+    // Contracts (persistent receives)
+    val C_ = Rho("contract @0(id) = { 0 }")
+    val C0 = Rho("contract @0(@0) = { 0 }")
+    val C1 = Rho("contract @0(@1) = { 0 }")
+    // Ground term
+    val Z = Rho("0")
```
lukasz-golebiewski, line 85: I would rather define `object Nil extends Rho("Nil")` and use this instead of `Z`

lukasz-golebiewski:
Closing in favor of https://github.com/rchain/rchain/pull/2606

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: How? Infinite loops are going to run out of phlo and any changes they make on the TS are going to be reverted. I see no point in testing such cases

ArturGajowy: Having such a test costs near to 0, and prevents us from unexpected results - like a block conflicting despite there being a way to merge it.

Notice that running out of phlo doesn't mean the deploy won't make it to the network. It will, because we need to charge for it. We might not replay it (or we may) - but those are all issues separate from what the result of 'is this a conflict' is.

Finally, I've been wrong in my `result is clear: they're mergable` statement: `CX !!X` leads to an infinite loop, yet must be a conflict.

Hope this convinces you that it's worth adding those cases?

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
lukasz-golebiewski: I'm going to re-write this and use RSpace events instead of those protobufs. Thanks to this the `Option`s will go away

PR #2583, opened 2019-07-05
Rchain 3465: add PoS.withdraw method

odeac:
## Overview
Add logic for validator withdrawal

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3465
https://rchain.atlassian.net/browse/RCHAIN-3466

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -57,7 +57,8 @@ new  PoS,
             // pendingRewards : Map[PublickKey, Int] - are accummulated by calling "pay"
             // committedRewards : Map[PublickKey, Int] - are moved from pendingRewards at each closeBlock
             // activeValidators : List[PublicKey] - the active validators
-            // withdrawers : Map[PublicKey, Int] - the validators willing to withdraw and the blocknumber for their withdrawal
+            // withdrawers : Map[PublicKey, Int] - the validators willing to withdraw and the blocknumber when their quarantine ends
+            //                                     This is the beggining of the next epoch after their withdrawal + quarantineLength
             // allBonds : Map[PublickKey, Int] - each validator stake
             stateCh!({
               "pendingRewards":{},
@@ -138,9 +139,12 @@ new  PoS,
                     @userPk <- userCh;
                     _, @blockNumber <- blockDataCh) {
                   if (state.get("allBonds").contains(userPk)) {
+                    // the withdwal is in effect starting from the next epoch
                     resultCh!(
                       state.set("withdrawers",
-                                state.get("withdrawers").set(userPk, blockNumber)),
+                                state.get("withdrawers")
+                                     .set(userPk,
+                                       $$quarantineLength$$ + $$epochLength$$ * (1 + blockNumber / $$epochLength$$))),
                       (true, Nil))
                   } else {
                     resultCh!(state, (false, "User is not bonded"))
@@ -333,7 +337,7 @@ new  PoS,
                     } |
 
                     contract isQuarantineFinished(@(_, blockNumber), resultCh) = {
-                      resultCh!(blockNumber + $$quarantineLength$$ >= currentBlockNumber)
+                      resultCh!(blockNumber >= currentBlockNumber)
                     }
                   }
                 }
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -3,6 +3,7 @@
 // maximumBond - the maximum bond allowed by PoS
 // initialBonds - the initial bonds map
 // epochLength - the length of the validation epoch in blocks
+// quarantineLength - the length of the quarantine time in blocks
 
 /*
  The table below describes the required computations and their dependencies
@@ -185,7 +186,8 @@ new  PoS,
             contract PoS(@"closeBlock", ackCh) = {
               new blockDataCh, mvarProcessCh,
                   getBlockData(`rho:block:data`),
-                  commitReward, rewardsCh, newValidatorsCh in {
+                  commitReward, rewardsCh, newValidatorsCh,
+                  newWithdrawersCh, newBondsCh in {
                 getBlockData!(*blockDataCh) |
 
                 runMVar!(*stateCh, *mvarProcessCh, *ackCh) |
@@ -198,6 +200,53 @@ new  PoS,
                     newValidatorsCh!(activeValidators)
                   } |
 
+                  new quarantinedValidatorsCh,
+                      validatorsToWithdrawListCh,
+                      validatorsToWithdrawSetCh,
+                      isQuarantineFinished,
+                      isWithdrawn,
+                      accumulateToSet,
+                      accumulateToMap,
+                      newBondsListCh,
+                      newWithdrawersListCh,
+                      test1ch, test2ch
+                  in {
+                    @ListOps!("filter", withdrawers.toList(), *isQuarantineFinished, *quarantinedValidatorsCh) |
+                    for (@quarantinedValidators <- quarantinedValidatorsCh) {
+
+                      @ListOps!("map", quarantinedValidators, *fst, *validatorsToWithdrawListCh) |
+                      for (@validatorsToWithdrawList <- validatorsToWithdrawListCh) {
+                        @ListOps!("fold", validatorsToWithdrawList, Set(), *accumulateToSet, *validatorsToWithdrawSetCh) |
+
+                        for (@validatorsToWithdrawSet <- validatorsToWithdrawSetCh) {
+                          @ListOps!("filter", allBonds.toList(), *isWithdrawn, *newBondsListCh) |
+
+                          for (@newBondsList <- newBondsListCh) {
+                            @ListOps!("fold", newBondsList, {}, *accumulateToMap, *newBondsCh)
+                          } |
+
+                          @ListOps!("filter", withdrawers.toList(), *isWithdrawn, *newWithdrawersListCh) |
+                          contract isWithdrawn(@(pk, _), resultCh) = {
+                            resultCh!(not validatorsToWithdrawSet.contains(pk))
+                          } |
+
+                          for (@newWithdrawersList <- newWithdrawersListCh) {
+                            @ListOps!("fold", newWithdrawersList, {}, *accumulateToMap, *newWithdrawersCh)
+                          } |
+
+                          contract isWithdrawn(@(pk, _), resultCh) = { resultCh!(not validatorsToWithdrawSet.contains(pk))}
+                        }
+                      }
+                    } |
+
+                    contract isQuarantineFinished(@(_, blockTime), resultCh) = {
+                      resultCh!(blockTime >= $$quarantineLength$$)
+                    } |
+
+                    contract accumulateToSet(@x, @acc, resultCh) = {resultCh! (acc.union(Set(x)))} |
+
+                    contract accumulateToMap(@(k, v), @acc, resultCh) = {resultCh!(acc.set(k, v))}
+                  } |
 
                   @ListOps!("fold", pendingRewards.toList(), rewards, *commitReward, *rewardsCh) |
                   contract commitReward(@(pk, pending), @acc, resultCh) = {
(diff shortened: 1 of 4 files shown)


JosephDenman:
Now that `filter` is merged, you can probably rebase onto dev and eliminate the first commit.

casper/src/main/resources/PoS.rhox
```diff
@@ -160,30 +201,119 @@ new  PoS,
...
 
+                    contract isQuarantineFinished(@(_, blockNumber), resultCh) = {
+                      resultCh!(blockNumber + $$quarantineLength$$ >= currentBlockNumber)
+                    } |
 
-                  @ListOps!("fold", pendingRewards.toList(), rewards, *commitReward, *rewardsCh) |
-                  contract commitReward(@(pk, pending), @acc, resultCh) = {
-                    resultCh!(acc.set(pk, acc.getOrElse(pk, 0) + pending))
-                  } |
+                    contract accumulateToSet(@x, @acc, resultCh) = {resultCh! (acc.union(Set(x)))} |
```
KentShikama: nitpick: extra space

casper/src/main/resources/PoS.rhox
```diff
@@ -160,30 +201,119 @@ new  PoS,
...
+                      accumulateToMap,
+                      newBondsListCh,
+                      newWithdrawersListCh
+                  in {
+                    @ListOps!("filter", withdrawers.toList(), *isQuarantineFinished, *quarantinedValidatorsCh) |
+                    for (@quarantinedValidators <- quarantinedValidatorsCh) {
+
+                      @ListOps!("map", quarantinedValidators, *fst, *validatorsToWithdrawListCh) |
+                      for (@validatorsToWithdrawList <- validatorsToWithdrawListCh) {
+                        @ListOps!("fold", validatorsToWithdrawList, Set(), *accumulateToSet, *validatorsToWithdrawSetCh) |
```
KentShikama: so is this just a .toSet operation?

casper/src/main/resources/PoS.rhox
```diff
@@ -160,30 +201,119 @@ new  PoS,
...
+                      state.get("activeValidators"),
+                      state.get("withdrawers"),
+                      *newValidatorsCh)|
+
+                    removeQuarantinedWithdrawers!(
+                      blockNumber,
+                      state.get("allBonds"),
+                      state.get("withdrawers"),
+                      *newWithdrawersCh,
+                      *newBondsCh) |
```
KentShikama: So I see this "method" gives back the new withdraw and new bond set but where does the withdrawals actually go to?

casper/src/main/resources/PoS.rhox
odeac: Yeah, I was also thinking this would better go to a list .toSet() method. And we should also have .toMap() method that works on lists of tuples

casper/src/main/resources/PoS.rhox
odeac: For now, they just go to a better place :)

Now I realise that they should receive their bonds and their rewards. I have to  fix this in this PR

casper/src/main/resources/PoS.rhox
odeac: this is done

casper/src/main/resources/PoS.rhox
odeac: fixed

casper/src/main/resources/PoS.rhox
```diff
@@ -39,29 +40,38 @@ new  PoS,
 
   for(@(_, RevVault) <- revVaultCh;
       @(_, ListOps) <- listOpsCh) {
-    new posRevAddressCh in {
+    new posRevAddressCh, posAuthKeyCh, posVaultCh, initialActiveCh in {
 
+      @RevVault!("deployerAuthKey", *posAuthKeyCh) |
```
KentShikama, line 45: I forgot why this works even though the current user is the validator that is executing this logic.

KentShikama (approved):

casper/src/main/resources/PoS.rhox
JosephDenman: The code is only called when the `PoS` contract is first deployed. That deploy uses the `PoS` contract's secret key.

casper/src/main/resources/PoS.rhox
```diff
@@ -271,49 +271,42 @@ new  PoS,
...
+                            for (@newBondsList <- newBondsListCh) {
+                              newBondsCh!(newBondsList.toMap())
+                            } |
 
-                          @ListOps!("filter", withdrawers.toList(), *isWithdrawn, *newWithdrawersListCh) |
-                          contract isWithdrawn(@(pk, _), resultCh) = {
-                            resultCh!(not validatorsToWithdrawSet.contains(pk))
-                          } |
+                            @ListOps!("filter", withdrawers.toList(), *isWithdrawn, *newWithdrawersListCh) |
+                            contract isWithdrawn(@(pk, _), resultCh) = {
```
JosephDenman: It looks like this "function" is duplicated below.

casper/src/main/resources/PoS.rhox
```diff
@@ -113,6 +113,22 @@ new  PoS,
               }
             } |
 
+            contract PoS (@"withdraw", returnCh) = {
+              new userCh, depositCh, processCh in {
```
tymm: `depositCh` seems to be unused.

casper/src/test/resources/PoSTest.rho
```diff
@@ -144,6 +146,54 @@ match (
...
+                        @posVault!("balance", *finalPosBalanceCh) |
+                        @user1Vault!("balance", *finalUserBalanceCh) |
+                        @PoS!("getRewards", *finalRewardsCh) |
+                        for(@finalPosBalance <- finalPosBalanceCh;
+                            @finalUserBalance <- finalUserBalanceCh;
+                            @finalBonds <- finalBondsCh;
+                            @finalRewards <- finalRewardsCh) {
+                          rhoSpec!("assertMany",
+                            [
+                              ((finalBonds, "==", initialBonds), "bonds don't change"),
```
JosephDenman, line 184: Do these not change because the validator withdrawing has not yet completed the required quarantine length?

casper/src/main/resources/PoS.rhox
```diff
@@ -203,50 +205,72 @@ new  PoS,
...
-                          "pendingRewards":{},
-                          "committedRewards":newRewards,
-                          "activeValidators":newValidators,
-                          "withdrawers":newWithdrawers,
-                          "allBonds":newBonds
-                        },
-                        Nil)
+                      *commitRewardsCh) |
+
+                    for (@committedRewards <- commitRewardsCh) {
```
JosephDenman, line 238: Is there a reason that you call `changeEpoch` after `commitRewards` returns? Is this a concurrency issue or just coincidence?

casper/src/main/resources/PoS.rhox
```diff
@@ -268,37 +292,39 @@ new  PoS,
...
+                      newRewardsListCh,
+                      paymentsCh, computePay
                   in {
                     @ListOps!("filter", withdrawers.toList(), *isQuarantineFinished, *quarantinedValidatorsCh) |
                     for (@quarantinedValidators <- quarantinedValidatorsCh) {
                       @ListOps!("map", quarantinedValidators, *fst, *validatorsToWithdrawListCh) |
                       for (@validatorsToWithdrawList <- validatorsToWithdrawListCh) {
+                        @ListOps!("map", validatorsToWithdrawList, *computePay, *paymentsCh) |
+                        contract computePay(@pk, resultCh) = {
+                          resultCh!((pk, allBonds.get(pk) + rewards.getOrElse(pk, 0)))
```
JosephDenman, line 317: Is `withDraw` basically un-bonding?

casper/src/test/resources/PoSTest.rho
JosephDenman: Shouldn't there be a test that verifies these values _do_ change?

JosephDenman (changes requested):
The fourth from last commit should be squashed into the fourth commit. This PR will be much easier to review after you've done so.

casper/src/main/resources/PoS.rhox
odeac: it is. I'm not sure why the name changed. Maybe @KentShikama knows

casper/src/main/resources/PoS.rhox
odeac: It's just coincidence. There is no dependency between commitRewards and changeEpoch. I'll add a new commit to remove this apparent dependency

casper/src/test/resources/PoSTest.rho
odeac: There is no change because no payment happens between bond and withdraw. I'll add a test for a situation where things change

casper/src/main/resources/PoS.rhox
odeac: true. And also the name is wrong.

This is fixed in the following commits

casper/src/main/resources/PoS.rhox
```diff
@@ -203,50 +205,72 @@ new  PoS,
...
+                            Nil)
+                        }
+                      }
+                    }
+                  }
+                } |
+
+                contract payValidators(@payments, ackCh) = {
+                  new payValidator in {
+                    @ListOps!("foreach", payments, *payValidator, *ackCh) |
```
JosephDenman: I don't see why payments should be sequentialized. You could use something like `unorderedParMap` to get your money's worth.

casper/src/test/resources/PoSTest.rho
```diff
@@ -195,6 +197,58 @@ match (
...
+                @validatorVault!("balance", *initialValidatorBalanceCh) |
+                @PoS!("getBonds", *initialBondsCh) |
+                for(@initialValidatorBalance <- initialValidatorBalanceCh;
+                    @initialBonds <- initialBondsCh) {
+                  @PoS!("bond", 100, *retCh) |
+                  for ( @(true, _) <- retCh) {
+                    @PoS!("closeBlock", *closeBlockCh) |
+                    for (_ <- closeBlockCh) {
+                      prepareUser!("cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc", 10000, *setupCh) |
+                      for (@userPubKey, _, @userVault <- setupCh) {
```
JosephDenman, line 217: While you were writing these tests, did you witness these "methods" returning values? If, for example, `"pay"` returned something other than `@(true, _)` on `ret`, we wouldn't know that the test failed. The same goes if nothing is returned on `closeBlockCh` or other similar receives. It could be useful to verify with print statements.

casper/src/test/resources/PoSTest.rho
```diff
@@ -195,6 +197,58 @@ match (
...
+                initialBondsCh, finalBondsCh,
+                withdrawCh, closeBlockCh,
+                setDeployData(`rho:test:deploy:set`), identitySet in {
+              prepareUser!("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb", 10000, *setupCh) |
+              for (@validatorPubKey, _, @validatorVault <- setupCh) {
+                @validatorVault!("balance", *initialValidatorBalanceCh) |
+                @PoS!("getBonds", *initialBondsCh) |
+                for(@initialValidatorBalance <- initialValidatorBalanceCh;
+                    @initialBonds <- initialBondsCh) {
+                  @PoS!("bond", 100, *retCh) |
```
JosephDenman, line 212: This test should also include a usage of `"pay"` to ensure that validator and `PoS` vault balances change appropriately?

JosephDenman (changes requested):

casper/src/test/resources/PoSTest.rho
odeac: if pay returns something else than true the test suite will timeout and we will certainly notice that situation

casper/src/test/resources/PoSTest.rho
odeac: It does call PoS.pay at line 218

JosephDenman (approved):
LGTM

PR #2585, opened 2019-07-09
Base integration tests on pyenv and not on system Python

adaszko:
[Here's why](https://rchain.atlassian.net/browse/RCHAIN-3549)

### DIFF
--- a/integration-tests/README.md
+++ b/integration-tests/README.md
@@ -1,41 +1,39 @@
 # Installing the prerequisites
 ## Step 1. Install Docker
 
-Follow the instructions on [this page](https://docs.docker.com/install/) for platform specific instalation steps
+If you don't know better, refer to the platform-specific installation
+instructions at [Docker website](https://docs.docker.com/install/).
 
-## Step 2. Python v3.6+
+## Step 2. Install pyenv
 
-The integration tests run on **Python v3**, which is incompatible with Python v2.
+`pyenv` allows for installing a standalone, isolated, specific CPython
+version, making integration tests work irrespective of what CPython version
+you have installed in your operating system.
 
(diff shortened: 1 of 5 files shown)


adaszko:
bors try

adaszko:
@zsluedem I had to revert 38d1bf20cb49b0780748adf4f62a87b752df93bb to make the tests work on macOS again.  Do you remember the reason for that commit?

adaszko:
bors try

adaszko:
bors try

adaszko:
bors try

woky:
@zsluedem You're the most involved in this. It seems OK to me but I'll wait for your review.

ArturGajowy (approved):
FWIW I was able to start integration tests without a hiccup using the updated readme and scripts. Yay! \o/ :heart:

sebtomba (approved):

zsluedem:
> @zsluedem I had to revert [38d1bf2](https://github.com/rchain/rchain/commit/38d1bf20cb49b0780748adf4f62a87b752df93bb) to make the tests work on macOS again. Do you remember the reason for that commit?

You don't have to revert it to make the test work in macOS.
There are two ways to solve the problem.

1. You can set up the environment variable `TMPDIR=/tmp` based on the [doc](https://docs.python.org/3.6/library/tempfile.html#tempfile.mkstemp) in your macOS(`TEMP` and `TMP` is not working in macOS).

2. `mkstemp` create the temporily files in `/var/folders` in macOS by default. You can add `/var/folders` directory in your [docker sharing folder](https://docs.docker.com/docker-for-mac/osxfs/#namespaces).

zsluedem (approved):

PR #2587, opened 2019-07-09
RCHAIN-3505: Error lock 0.1

JosephDenman:
## Overview
Preliminary changes for introducing error lock in interpreter.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3505


JosephDenman:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
```diff
@@ -53,10 +53,7 @@ object Tuplespace {
           case None => F.unit
         }
 
-      for {
-        res <- pureRSpace.produce(channel, data, persist = persistent, sequenceNumber)
-        _   <- go(res)
-      } yield ()
+      pureRSpace.produce(channel, data, persist = persistent, sequenceNumber) >>= (go(_))
```
odeac: I like the >>= operator but having more than one operation on a line is not good for readability.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
```diff
@@ -61,33 +61,30 @@ object Tuplespace {
...
-              case Some((continuation, dataList, updatedSequenceNumber)) =>
-                if (persistent)
-                  List(
-                    dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
-                    consume(binds, body, persistent, sequenceNumber)
-                  ).parSequence_
-                else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
-              case None => F.unit
-            }
+      def go(
```
odeac: Since you are cleaning this area you could also rename the `go` function to something like `dispatch`

odeac (approved):
I only have two minor comments.  The granularity of the commits is really nice for the reviewer.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
ArturGajowy: It is also bad for debugability, but in this case I can plug into the `go` function. Fine by me :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
```diff
@@ -61,33 +61,30 @@ object Tuplespace {
         body: ParWithRandom,
         persistent: Boolean,
         sequenceNumber: Int
-    ): F[Unit] =
-      binds match {
-        case Nil => F.raiseError(ReduceError("Error: empty binds"))
```
ArturGajowy, line 73: > The empty binds check is unnecessary since expressions of the form "for(){ P }" do not parse.

That's the point with defensive programming: you place your assertions where they need to hold, because you don't know who and how will call the method. Or if they make a bug that goes unnoticed because of our short-circuiting on an empty list here.

The alternative to defensive programming is: making invalid states unrepresentable. Since that would require an AST revamp, I say we stick with defensive programming. But I guess this class disappears?

EDIT: once we have it in the same class, and once `consume` is a private method, I'm more comfortable saying we can leave this as is.

rholang/src/test/scala/coop/rchain/rholang/interpreter/RholangOnlyDispatcher.scala
```diff
@@ -29,7 +28,7 @@ object RholangOnlyDispatcher {
     lazy val dispatcher: Dispatch[M, ListParWithRandom, TaggedContinuation] =
       new RholangOnlyDispatcher
 
-    lazy val tuplespaceAlg = Tuplespace.rspaceTuplespace(pureSpace, dispatcher)
+    lazy val tuplespaceAlg = new TuplespaceImpl(pureSpace, dispatcher)
```
ArturGajowy: Please call the class RspaceTuplespace. The only thing the `Impl` suffix contributes to a name is dread and confusion.

EDIT: NVM, the comment is only relevant for our future naming habits.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
```diff
@@ -39,19 +39,16 @@ class TuplespaceImpl[F[_], M[_]](
...
-        case Some((continuation, dataList, updatedSequenceNumber)) =>
-          if (persistent)
-            List(
-              dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
-              produce(channel, data, persistent, sequenceNumber)
-            ).parSequence_
-          else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
-        case None => F.unit
-      }
+    def go: Option[(TaggedContinuation, Seq[ListParWithRandom], Int)] => F[Unit] = {
```
ArturGajowy: I don't see any benefit here, and find this a downgrade in readability and editability in general (think: IDE support, adding a context bound, etc). Please drop this commit.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -65,28 +66,70 @@ class CostAccountingReducerTest extends FlatSpec with Matchers with TripleEquals
...
-    }
 
-    implicit val errorLog = new ErrorLog[Task]()
-    implicit val rand     = Blake2b512Random(128)
-    implicit val cost     = CostAccounting.initialCost[Task](Cost(1000)).runSyncUnsafe(1.second)
-    val reducer           = new DebruijnInterpreter[Task, Task.Par](tuplespaceAlg, Map.empty)
-    val send              = Send(Par(exprs = Seq(GString("x"))), Seq(Par()))
-    val test              = reducer.inj(send).attempt.runSyncUnsafe(1.second)
+    val iSpace =
+      new ISpace[Task, Par, BindPattern, ListParWithRandom, ListParWithRandom, TaggedContinuation] {
```
ArturGajowy: Holy moly! Could this have been the reason for having the Tuplespace trait? :thinking:

Please extract an `ISpaceStub` class that will leave all methods undefined, and allow for the test to be as readable as previously.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -71,12 +73,22 @@ class DebruijnInterpreter[M[_], F[_]](
...
+      case Some((continuation, dataList, updatedSequenceNumber)) =>
+        if (persistent)
+          List(
+            dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
+            produce(chan, data, persistent, sequenceNumber)
+          ).parSequence_
+        else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
+      case None => s.unit
+    }
+    pureRSpace.produce(chan, data, persist = persistent, sequenceNumber) >>= (go(_))
```
ArturGajowy, line 93: It's nice to have this logic in the only place it is used :+1:

Please restore the `go` method (vs lambda) though :)

ArturGajowy: If you're after concision, extracting a type alias/case class for the tuplespace return type could be a good way :)

ArturGajowy (changes requested):
One minor comment re: lambda vs method, + please consider adding the type alias/case class. :+1: on the commit log indeed, and on simplifying our code's structure :)

tymm (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
JosephDenman: How is it bad for debugability?

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
JosephDenman: Yes, `Tuplespace` is absorbed into `Reduce` and these methods become private.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/Tuplespace.scala
ArturGajowy: It's harder to put a breakpoint inside a chain of `>>=`

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
JosephDenman: What do you mean extract? In a separate file? At the top of the test? You basically want me to move it somewhere else and keep the implementation?

ArturGajowy: I want you to provide a:

```scala
class ISpaceStub extends ISpace {
  //all methods implemented using ??? - it's ok for a stub
}
```

so that the test can reduce its setup to what +/- it was previously
```scala
   val ispace = new ISPaceStub {
      override def produce(
          chan: Par,
          data: ListParWithRandom,
          persistent: Boolean,
          sequenceNumber: Int
      ): Task[Unit] =
        Task.raiseError(OutOfPhlogistonsError)
      override def consume(
          binds: Seq[(BindPattern, Par)],
          body: ParWithRandom,
          persistent: Boolean,
          sequenceNumber: Int
      ): Task[Unit] = Task.raiseError(OutOfPhlogistonsError)
    }
```

At a minimum, put the ISpaceStub at the bottom of the file, or in a separate file if you will.

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -66,28 +69,35 @@ class CostAccountingReducerTest extends FlatSpec with Matchers with TripleEquals
...
+
+    val iSpace = new ISpaceStub[
+      Task,
+      Par,
+      BindPattern,
+      ListParWithRandom,
+      ListParWithRandom,
+      TaggedContinuation
+    ] {
       override def produce(
```
ArturGajowy, line 81: are you sure you only wanted to override the produce method? Previously consume was stubbed as well... I guess, if the test passes - it's OK

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -66,28 +69,35 @@ class CostAccountingReducerTest extends FlatSpec with Matchers with TripleEquals
...
-          binds: Seq[(BindPattern, Par)],
-          body: ParWithRandom,
-          persistent: Boolean,
+          persist: Boolean,
           sequenceNumber: Int
-      ): Task[Unit] = Task.raiseError(OutOfPhlogistonsError)
+      )(
+          implicit m: Match[Task, BindPattern, ListParWithRandom, ListParWithRandom]
+      ): Task[
+        Option[(ContResult[Par, BindPattern, TaggedContinuation], Seq[Result[ListParWithRandom]])]
```
ArturGajowy, line 89: Maaan that's an ugly signature...

ArturGajowy (approved):

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
JosephDenman: Consume was previously unused. Only a send is evaluated, so only a produce is needed.

PR #2593, opened 2019-07-12
Implement toSet and toMap in Scala as list methods

odeac:
## Overview
Implement toSet and toMap used for converting lists.

This is needed in PoS implementation of withdraw, where we currently implemented these conversions in Rholang

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3465
https://rchain.atlassian.net/browse/RCHAIN-3466

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1402,6 +1402,37 @@ class DebruijnInterpreter[M[_], F[_]](
         result   <- toSet(baseExpr)
       } yield result
   }
+
+  private[this] val toMap: Method = new Method() {
+    def toMap(baseExpr: Expr): M[Par] =
+      baseExpr.exprInstance match {
+        case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
+          val keyPairs = basePs.map(RhoType.Tuple2.unapply)
+          if (keyPairs.exists(_.isEmpty))
+            MethodNotDefined("toMap", "types except List[(K,V)]").raiseError[M, Par]
+          else
+            (EMapBody(
+              ParMap(
+                keyPairs.flatMap(_.toList),
+                connectiveUsed,
+                locallyFree.get.pure[Coeval],
+                remainder
+              )
+            ): Par).pure[M]
+        case other =>
+          MethodNotDefined("toMap", other.typ).raiseError[M, Par]
+      }
+
+    override def apply(p: Par, args: Seq[Par])(implicit env: Env[Par]): M[Par] =
+      for {
+        _ <- if (args.nonEmpty)
+              MethodArgumentNumberMismatch("toMap", 0, args.length).raiseError[M, Unit]
+            else ().pure[M]
+        baseExpr <- evalSingleExpr(p)
+        result   <- toMap(baseExpr)
+      } yield result
+  }
+
   private val methodTable: Map[String, Method] =
     Map(
       "nth"         -> nth,
@@ -1420,7 +1451,9 @@ class DebruijnInterpreter[M[_], F[_]](
       "size"        -> size,
       "length"      -> length,
       "slice"       -> slice,
-      "toList"      -> toList
+      "toList"      -> toList,
+      "toSet"       -> toSet,
+      "toMap"       -> toMap
     )
 
   private def evalSingleExpr(p: Par)(implicit env: Env[Par]): M[Expr] =
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1380,6 +1380,15 @@ class DebruijnInterpreter[M[_], F[_]](
       baseExpr.exprInstance match {
         case e: ESetBody =>
           (e: Par).pure[M]
+        case EMapBody(ParMap(basePs, connectiveUsed, locallyFree, remainder)) =>
+          (ESetBody(
+            ParSet(
+              basePs.toSeq.map(t => ETupleBody(ETuple(Seq(t._1, t._2))): Par),
+              connectiveUsed,
+              locallyFree,
+              remainder
+            )
+          ): Par).pure[M]
         case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
           (ESetBody(
             ParSet(
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1404,23 +1404,28 @@ class DebruijnInterpreter[M[_], F[_]](
   }
 
   private[this] val toMap: Method = new Method() {
+    def makeMap(ps: Seq[Par], connectiveUsed : Boolean, locallyFree : Coeval[BitSet], remainder:Option[Var]) = {
+      val keyPairs = ps.map(RhoType.Tuple2.unapply)
+      if (keyPairs.exists(_.isEmpty))
+        MethodNotDefined("toMap", "types except List[(K,V)]").raiseError[M, Par]
+      else
+        (EMapBody(
+          ParMap(
+            keyPairs.flatMap(_.toList),
+            connectiveUsed,
+            locallyFree,
+            remainder
+          )
+        ): Par).pure[M]
+    }
     def toMap(baseExpr: Expr): M[Par] =
       baseExpr.exprInstance match {
         case e: EMapBody =>
           (e: Par).pure[M]
+        case ESetBody(ParSet(basePs, connectiveUsed, locallyFree, remainder)) =>
+          makeMap(basePs.toSeq, connectiveUsed, locallyFree, remainder)
         case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
-          val keyPairs = basePs.map(RhoType.Tuple2.unapply)
-          if (keyPairs.exists(_.isEmpty))
-            MethodNotDefined("toMap", "types except List[(K,V)]").raiseError[M, Par]
-          else
-            (EMapBody(
-              ParMap(
-                keyPairs.flatMap(_.toList),
-                connectiveUsed,
-                locallyFree.get.pure[Coeval],
-                remainder
-              )
-            ): Par).pure[M]
+          makeMap(basePs, connectiveUsed, locallyFree.get.pure[Coeval], remainder)
         case other =>
           MethodNotDefined("toMap", other.typ).raiseError[M, Par]
       }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1375,6 +1375,33 @@ class DebruijnInterpreter[M[_], F[_]](
       } yield result
   }
 
+  private[this] val toSet: Method = new Method() {
+    def toSet(baseExpr: Expr): M[Par] =
+      baseExpr.exprInstance match {
+        case e: ESetBody =>
+          (e: Par).pure[M]
+        case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
+          (ESetBody(
+            ParSet(
+              basePs,
+              connectiveUsed,
+              locallyFree.get.pure[Coeval],
+              remainder
+            )
+          ): Par).pure[M]
+        case other =>
+          MethodNotDefined("toSet", other.typ).raiseError[M, Par]
+      }
+
+    override def apply(p: Par, args: Seq[Par])(implicit env: Env[Par]): M[Par] =
+      for {
+        _ <- if (args.nonEmpty)
+              MethodArgumentNumberMismatch("toSet", 0, args.length).raiseError[M, Unit]
+            else ().pure[M]
+        baseExpr <- evalSingleExpr(p)
+        result   <- toSet(baseExpr)
+      } yield result
+  }
   private val methodTable: Map[String, Method] =
     Map(
       "nth"         -> nth,
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1404,7 +1404,12 @@ class DebruijnInterpreter[M[_], F[_]](
   }
 
   private[this] val toMap: Method = new Method() {
-    def makeMap(ps: Seq[Par], connectiveUsed : Boolean, locallyFree : Coeval[BitSet], remainder:Option[Var]) = {
+    def makeMap(
+        ps: Seq[Par],
+        connectiveUsed: Boolean,
+        locallyFree: Coeval[BitSet],
+        remainder: Option[Var]
+    ) = {
       val keyPairs = ps.map(RhoType.Tuple2.unapply)
       if (keyPairs.exists(_.isEmpty))
         MethodNotDefined("toMap", "types except List[(K,V)]").raiseError[M, Par]
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -1406,6 +1406,8 @@ class DebruijnInterpreter[M[_], F[_]](
   private[this] val toMap: Method = new Method() {
     def toMap(baseExpr: Expr): M[Par] =
       baseExpr.exprInstance match {
+        case e: EMapBody =>
+          (e: Par).pure[M]
         case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
           val keyPairs = basePs.map(RhoType.Tuple2.unapply)
           if (keyPairs.exists(_.isEmpty))
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
@@ -2346,19 +2346,10 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
 
   "Reducer" should "work correctly in succesful cases" in {
     forAll(successfulExamples) { (clue, input, output) =>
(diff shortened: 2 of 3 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2059,4 +2056,284 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+          EList(
+            List[Par](
+              GInt(1L),
+              GInt(2L),
+              GInt(3L)
+            )
+          )
+        ),
+        List[Par]()
+      )
```
ArturGajowy: We really should use the ParBuilder to build AST in this test.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2059,4 +2056,284 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+    val toSetCall: EMethod =
+      EMethod(
+        "toSet",
+        EListBody(
+          EList(
+            List[Par](
+            )
+          )
+        ),
+        List()
```
ArturGajowy: All those tests have the same structure. It feels as if we could reduce them all to:

`checkResult("@0!([1,2,3].toSet())", "@0!(Set(1, 2, 3))")`

Or, equally well, we could implement the test in RhoSpec.

Either way, there's duplication between the tests to be removed.

ArturGajowy: Ok, there are error cases - that's another test util method

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1334,6 +1334,64 @@ class DebruijnInterpreter[M[_], F[_]](
...
+        case EListBody(EList(basePs, locallyFree, connectiveUsed, remainder)) =>
+          (ESetBody(
+            ParSet(
+              basePs,
+              connectiveUsed,
+              locallyFree.get.pure[Coeval],
+              remainder
+            )
+          ): Par).pure[M]
+        case other =>
```
JosephDenman, line 1401: I'm curious what your reasoning is for not implementing `toSet` and `toMap` for all Rholang collections that are scala collections under-the-hood. You should just be able to delegate down to the scala collection converters.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
JosephDenman: On one hand, using `ParBuilder` in tests for `Reduce` introduces unnecessary dependencies and violates separation of concerns - testing both the normalizer and the evaluator. On the another, using `ParBuilder` increases readability by reducing preparation. I side with SOC.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1334,6 +1334,64 @@ class DebruijnInterpreter[M[_], F[_]](
...
+      for {
+        _ <- if (args.nonEmpty)
+              MethodArgumentNumberMismatch("toSet", 0, args.length).raiseError[M, Unit]
+            else ().pure[M]
+        baseExpr <- evalSingleExpr(p)
+        result   <- toSet(baseExpr)
+      } yield result
+  }
+
+  private[this] val toMap: Method = new Method() {
```
JosephDenman, line 1415: This PR would be significantly easier to review if this commit was split into two commits, one adding `toMap` and another adding `toSet`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2059,4 +2056,284 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+        Await.result(inspectTask.runToFuture, 3.seconds)
+    }
+    result should be(mutable.HashMap.empty)
+    errorLog.readAndClearErrorVector().unsafeRunSync should be(
+      Vector(MethodNotDefined("toSet", "Int"))
+    )
+  }
+
+
+  "toMap" should """transform [("a",1), ("b",2), ("c",3)] into {"a":1, "b":2, "c":3)""" in {
```
JosephDenman: What happens if `toMap` is called on lists like `["a", ("b", 2)]` or `[1, ("b", 2)]`?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2059,4 +2056,284 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
     result.exprs should be(Seq(Expr(resultList)))
     errorLog.readAndClearErrorVector().unsafeRunSync should be(Vector.empty[InterpreterError])
   }
+  "toSet" should "transform [1, 2, 3] into Set(1, 2, 3)" in {
```
JosephDenman: Please add a test that verifies duplicated elements are removed from the source collection when the source collection is transformed into a set.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
JosephDenman: Please add a test that shows that a list like `[("a", 1), ("a", 2)]` is transformed into a map `{ "a" : 1 }`.

JosephDenman (changes requested):

KentShikama (approved):
I would like to see tests for toSet/toMap with duplicate entries but otherwise looks good.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
ArturGajowy: Test setup is **not testing** the things used in test setup - it's setting up the stage for the test.

If a test fails in its setup - it's either due to a bug one has just introduced, or due to a change to a cross-cutting concern so drastic that the by-hand test setup would be obsolete and lead to the test no longer testing anything that can happen in production - making the test null in the worst case. It's like drawing a good conclusion from the wrong argument.

Heck, I bet some of the hand-crafted ASTs we have now in tests can't ever happen in production either (they're not sorted).

I don't think we understand separation of concerns the same way. If I need the input par to be normalized *and* sorted - does my calling `normalize` (which sorts underneath) violate SOC? (Hint: it doesn't.)

Hand-crafting complex input data for a test where using prod code in setup would suffice for the test's purpose is terribly misguided, detrimental to test reliability and test maintainability. I can't stress enough how wrong this is.

ArturGajowy (dismissed):
Please see comments above

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
ArturGajowy: > unnecessary dependencies 

quite the contrary, the fact that this test doesn't use ParBuilder for creating Reducer inputs (which is the only way a reducer input is produced in prod BTW) can lead to false positives in tests (par order can affect execution results / pattern matching; in general: ex falso quodlibet, or GIGO if you will).

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: What would be the expected behaviour for the other collections?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: It already is

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: indeed, there is some refactoring required for the ReduceSpec tests. I'll try to address the duplication.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: I side with Joe here. The ReduceSpec should test only the Reducer and the input of the reducer is the normalised AST, not the Rholang code.

If the unit tests are unreadable and unmaintainable then it's a problem in the API of the unit under test and the tests just point to that issue. Using tools to make the tests readable only hides the actual problem.

ArturGajowy: > The ReduceSpec **should test only** the Reducer and the input of the reducer is the normalised AST, not the Rholang code.

How can I stress more that **test setup is not testing**? Is that a controversial claim?

> the normalised AST, not the Rholang code.

My claim is, normalization is so complex, that the only **reliable** way to have normalized test input is to use the normalizer. Similarly for parsing. Otherwise we end up doing a lot of test fixing when a detail not relevant to them changes.

Have you read my argument for why pretending that we know how to create the complex test input data is wrong until one cannot write the test otherwise?

---

Consider SRP - each piece of code should only have one reason to change. A test can change for the following reasons:
 1. input data needs to change
 2. the way we call the tested code changes
 3. output data needs to change

When we compare 'by-hand inputs' with 'prod-code inputs', the latter reduces the number of times tests need to change -> we have better, more SRP-y tests.

---

Consider Separation of Concerns - Dijkstra mentions it in the context of `thinking intelligently`, which is to limit the scope of consideration so that each aspect is considered in isolation; at any time we occupy ourselves only with one of the aspects. 

Using 'by-hand inputs' forces test reader & writer to think about how a normalized AST with a peek looks like. Using 'prod-code inputs' allows one to focus on the semantics of peek execution. The rholang source code, **is**, in an ultimate sense, the input to the reducer - and the only form of that input a human being can reason about *intelligently* when we consider results of execution.

So, 'prod-code inputs' leads to better, more SoC-y tests.

---

Granted, there are cases where input data needs to be constructed by hand, like e.g.:
 - making sure the reducer terminates as expected for invalid rholang ASTs (not a requirement, but an example)
 - making sure that the sorter outputs a sorted AST when given an not-sorted one (one could still use a test data generator, although without a *guarantee* that the generated data would be not-sorted)

Testing how a peek reduces though should not depend on how exactly the input AST looks like, as long as it's a normalized AST for a peek. More often than not, hand-writen AST will not be normalized. Further, a change to peek normalization could happen that changes execution results (be it a bug or a purposeful feature) - and the test for peek will not notice until it uses prod normalizer.

odeac: if the test code is difficult to write then we have bad smell. Using tools to hide that smell is not helping at all. Instead we should refactor and eliminate the smell.

If we choose to have Rholang code for tests input then we are  not testing the reducer but instead we are testing the whole interpreter (parser/normalizer/reducer) so that test should not live in ReduceSpec but in some InterpreterSpec.

Now, thinking more about the issue, I would go in the opposite direction and  refactor each method (toSet, toMap etc) by moving them outside the reducer and then unit-test them separately by calling the method apply directly, not by running the reducer.

Then the reducer will only contain one test per method, which would test that the actual method implementation is plugged in correctly.

ArturGajowy: How is what I'm proposing difficult to write?

`checkResult("@0!([1,2,3].toSet())", "@0!(Set(1, 2, 3))")`

It's even easier than writing all of that in Rholang/RhoSpec.
It's also helping in at least 4 ways: readability, ease of writing, reliability (always uses what's **actually** normalized), and maintainability (less reasons for the test to change / SRP).

Maybe using prod code in setup is not using a tool? For sure it's using the ultimate thing that has a say what the test input should be: the code that produces that input in production.

---

> not testing the reducer but instead we are testing the whole interpreter (parser/normalizer/reducer) so that test should not live in ReduceSpec but in some InterpreterSpec.

I disagree that using prod code in test setup is a test for that prod code too. It does increase the fidelity level of that test though, and prevents testing impossible scenarios - and most tests aren't about those. 

Interpreter could be doing more things in `Interpreter.eval()` than just parsing and normalizing, and then reducing. So using parser + normalizer to test reduction is not testing the interperter. Using interpreter input (however obtained) and calling its method would be.

Moving the tests to InterpreterSpec would be :100: fine for me though, and I like it more than the other solution you proposed, for various reasons (incl. effort needed, test fidelity). Can we do that and close the discussion?

odeac: These tests are doing exactly the same as the other methods tests are doing. So let's keep them like this for now and then we see how to refactor all the ReduceSpec tests in a separate PR.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1361,6 +1361,37 @@ class DebruijnInterpreter[M[_], F[_]](
...
+          else
+            (EMapBody(
+              ParMap(
+                keyPairs.flatMap(_.toList),
+                connectiveUsed,
+                locallyFree.get.pure[Coeval],
+                remainder
+              )
+            ): Par).pure[M]
+        case other =>
```
JosephDenman, line 1443: Calling`toMap` on a map should be a no-op. This implementation throws an error.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1361,6 +1361,37 @@ class DebruijnInterpreter[M[_], F[_]](
         result   <- toSet(baseExpr)
       } yield result
   }
+
+  private[this] val toMap: Method = new Method() {
+    def toMap(baseExpr: Expr): M[Par] =
```
JosephDenman, line 1435: Just as `toMap` is only defined on lists when the list elements are tuples, `toMap` should be defined on sets when the set elements are tuples. This is specified in the scala collections library. Please use that as your reference to implement the other collection conversions.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -920,8 +932,9 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
     )
     val serializedProcess =
       com.google.protobuf.ByteString.copyFrom(Serialize[Par].encode(proc).toArray)
-    val toByteArrayCall           = EMethod("toByteArray", proc, List[Par]())
-    def wrapWithSend(p: Par): Par = Send(GString("result"), List[Par](p), persistent = false, BitSet())
+    val toByteArrayCall = EMethod("toByteArray", proc, List[Par]())
+    def wrapWithSend(p: Par): Par =
```
JosephDenman, line 1056: I see you duplicate this function in three separate tests. It should probably be extracted.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2260,133 +2166,130 @@ class ReduceSpec extends FlatSpec with Matchers with PersistentStoreTester {
...
+      MethodNotDefined("toSet", "Int")
+    )
+  )
 
-    val result = withTestSpace(errorLog) {
-      case TestFixture(space, reducer) =>
-        implicit val env = Env[Par]()
-        val inspectTask  = reducer.eval(toMapCall) >> space.toMap
-        Await.result(inspectTask.runToFuture, 3.seconds)
+  "Reducer" should """return report errors in failure cases""" in {
```
JosephDenman: Nit: you don't need the extract quotations around `"return report errors in failure cases"`.

JosephDenman (changes requested):
Hi. Second review. Are we in agreement that the `toMap` and `toSet` converters should be defined on a larger set of source collections? I think we should just delegate down to the Scala collection converters, throwing errors whenever Scala would throw an error. What do you think?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
odeac: this is not a change of mine. This is just automatic formatting.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: So to sum it up: basically toMap and toSet would be methods of the "Seq" protocol and any collection implementing Seq would implement these. This makes sense to me

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: ok. Do as Scala does makes sense

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
ArturGajowy: When will this PR come?
What's the outcome of this discussion?
What approach will this PR take?

I'd hate to restart this from scratch, so to me that's a no-no until we reach consensus on the above 3 questions.

To me, implementing the new tests in the/a InterpreterSpec would be an improvement over what's in the PR currently.

ArturGajowy: As mentioned in the stand up, I'm OK with this PR proceeding as it is now. We should have the above discussion settled though before we write the next test that utilizes rholang AST without a strict necessity.

ArturGajowy:
@odeac feel free to dismiss my review, I've got nothing more to add besides the test inputs approach discussion

JosephDenman (approved):

dckc:
please, please, PLEASE document all rholang changes (including new methods) in https://github.com/rchain/rchain/tree/dev/rholang/examples

ArturGajowy:
@dckc your repeating the please thrice and capsing it won't make us have more time (or willingness) to do this type of work. A pull request, OTOH, would be helpful.

dckc:
Willingness... Darn. I thought this was agreed and someone just forgot.

JosephDenman:
@dckc You have a point. I missed it in the review.

@odeac Would you mind opening a pull request that adds examples of the `toSet` and `toMap` methods?

PR #2596, opened 2019-07-15
Rholang execution metrics

JosephDenman:
## Overview
Introduce Span[M] to the Rholang evaluator.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3621

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -58,7 +58,7 @@ class DebruijnInterpreter[M[_], F[_]](
     urnMap: Map[String, Par]
 )(
     implicit parallel: cats.Parallel[M, F],
-    s: Sync[M],
+    syncM: Sync[M],
     fTell: FunctorTell[M, Throwable],
     cost: _cost[M]
 ) extends Reduce[M] {
@@ -88,7 +88,7 @@ class DebruijnInterpreter[M[_], F[_]](
               produce(chan, data, persistent, sequenceNumber)
             ).parSequence_
           else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
-        case None => s.unit
+        case None => syncM.unit
       }
     pureRSpace.produce(chan, data, persist = persistent, sequenceNumber) >>= (go(_))
   }
@@ -119,7 +119,7 @@ class DebruijnInterpreter[M[_], F[_]](
               consume(binds, body, persistent, peek, sequenceNumber)
             ).parSequence_
           else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
-        case None => s.unit
+        case None => syncM.unit
       }
     pureRSpace.consume(
       sources.toList,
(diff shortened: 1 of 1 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1448,14 +1448,13 @@ class DebruijnInterpreter[M[_], F[_]](
...
-        acc ++ newPar
-      }
+
+      for {
+        evaledExprs <- par.exprs.toList.traverse(evalExprToPar)
+        // Note: the locallyFree cache in par could now be invalid, but given
+        // that locallyFree is for use in the matcher, and the matcher uses
+        // substitution, it will resolve in that case. AlwaysEqual makes sure
+        // that this isn't an issue in the rest of cases.
+        result = evaledExprs.foldLeft(par.copy(exprs = Vector()))(_ ++ _)
```
KentShikama, line 1499: Nitpick: I preferred the older style which is more explicit

KentShikama (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -237,19 +255,21 @@ class DebruijnInterpreter[M[_], F[_]](
 
     val starts = jobs.map(_.size).scanLeft(0)(_ + _).toVector
 
-    jobs.zipWithIndex.parTraverse_ {
-      case (job, jobIdx) => {
-        def mkRand(termIdx: Int): Blake2b512Random =
-          split(starts.last, starts(jobIdx), rand)(termIdx)
-        job.run(mkRand)(env, sequenceNumber)
+    spanM.trace(parSpanLabel) {
```
ArturGajowy: I agree that the code above most likely executes fast, but I'd rather make the span wider than be surprised later. Options: 

1. Ideally, the span would cover the whole method body. This could be achieved by moving the nested methods out of the `eval` method and making them private.
 - Similar move could be made for `val(mat: Match)`, but feel free to skip it 

2. I'm fine if you just widen this span to cover `val filteredExprs` and `val jobs = List(` though.

ArturGajowy (approved):
One minor comment and we're good to go :)

The PR should probably be named 'Meter Rholang' or 'Rholang execution metrics'. I'm going to rename it just now :)

Reviewing this was a pleasure, praise the fine-grained commit logs! :raised_hands:

ArturGajowy (approved):
One minor comment and we're good to go :)

The PR should probably be named 'Meter Rholang' or 'Rholang execution metrics'. I'm going to rename it just now :)

Reviewing this was a pleasure, praise the fine-grained commit logs! :raised_hands:

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
lukasz-golebiewski: I prefer the newer

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
lukasz-golebiewski: It's enough to move the code from current line between lines 230 and 231

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -473,8 +501,10 @@ class DebruijnInterpreter[M[_], F[_]](
       }
     }
 
-    charge[M](newBindingsCost(neu.bindCount)) >>
-      alloc(neu.bindCount, neu.uri).flatMap(eval(neu.p)(_, rand, sequenceNumber))
+    spanM.trace(newSpanLabel) {
```
lukasz-golebiewski: Same here and in other places. Allocations of helper functions contribute to the measured operation's duration, so they should be included in the span

lukasz-golebiewski (approved):
One remark, otherwise looks good

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ArturGajowy: Guess that's another point for having them as private methods :)

PR #2606, opened 2019-07-22
Mergeable contract calls in volatile events

lukasz-golebiewski:
## Overview
See https://github.com/rchain/rchain/pull/2578

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -4,12 +4,12 @@ import cats.Monad
 import cats.implicits._
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.{BlockDagRepresentation, BlockStore}
-import coop.rchain.casper.protocol.Event.EventInstance.{Comm, Consume, Produce}
-import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.{DagOperations, ProtoUtil}
+import coop.rchain.casper.protocol.{Event => CasperEvent, _}
+import coop.rchain.casper.util.{DagOperations, EventConverter, ProtoUtil}
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.BlockMetadata
 import coop.rchain.shared.Log
+import coop.rchain.rspace.trace._
 
 import scala.collection.BitSet
 
@@ -75,22 +75,21 @@ object EstimatorHelper {
         } yield conflicts
     }
 
-  val extractProduce: Event => Option[ProduceEvent] = {
-    case Event(Produce(produce: ProduceEvent)) => Some(produce)
-    case _                                     => None
+  val extractProduce: Event => Option[Produce] = {
+    case produce: Produce => Some(produce)
+    case _                => None
   }
 
-  val extractConsume: Event => Option[ConsumeEvent] = {
-    case Event(Consume(consume: ConsumeEvent)) => Some(consume)
-    case _                                     => None
+  val extractConsume: Event => Option[Consume] = {
+    case consume: Consume => Some(consume)
+    case _                => None
   }
 
-  val extractComm: Event => Option[CommEvent] = {
-    case Event(Comm(commEvent: CommEvent)) => Some(commEvent)
-    case _                                 => None
+  val extractComm: Event => Option[COMM] = {
+    case commEvent: COMM => Some(commEvent)
+    case _               => None
   }
 
-  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
   private def buildBlockAncestorChannels[F[_]: Monad: BlockStore](
       blockAncestorsMeta: List[BlockMetadata]
   ): F[Set[ByteString]] =
@@ -99,15 +98,16 @@ object EstimatorHelper {
                          blockAncestorMeta => BlockStore[F].get(blockAncestorMeta.blockHash)
                        )
       ancestors = maybeAncestors.flatten
-      ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
-        ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))
+      ancestorEvents = (ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
+        ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog)))
+        .map(EventConverter.toRspaceEvent)
 
       produceEvents = ancestorEvents.map(extractProduce).flatten
       consumeEvents = ancestorEvents.map(extractConsume).flatten
       commEvents    = ancestorEvents.map(extractComm).flatten
 
       producesInCommEvents = commEvents.flatMap(_.produces)
-      consumeInCommEvents  = commEvents.flatMap(_.consume)
+      consumeInCommEvents  = commEvents.map(_.consume)
 
       ancestorProduceChannels = produceEvents
         .filterNot(producesInCommEvents.contains(_))
@@ -121,20 +121,18 @@ object EstimatorHelper {
 
       ancestorCommChannels = commEvents
         .filterNot {
-          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+          case COMM(consume: Consume, produces, _) =>
             !consume.persistent && consumeEvents.contains(consume) && produces.forall(
               produce => !produce.persistent && produceEvents.contains(produce)
             )
-          case _ => false
         }
         .flatMap {
-          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+          case COMM(consume: Consume, produces, _) =>
             consume.channelsHashes ++ produces.map(_.channelsHash)
-          case e @ CommEvent(None, _) =>
-            throw new RuntimeException(s"Unexpected comm event $e")
         }
         .toSet
 
       ancestorChannels = ancestorProduceChannels ++ ancestorConsumeChannels ++ ancestorCommChannels
-    } yield ancestorChannels
+      res              = ancestorChannels.map(_.toByteString)
+    } yield res
 }
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -126,9 +126,8 @@ object EstimatorHelper {
 
       ancestorCommChannels = commEvents
         .filterNot(isVolatile(_, consumeEvents, produceEvents))
-        .flatMap {
-          case COMM(consume: Consume, produces, _) =>
-            consume.channelsHashes ++ produces.map(_.channelsHash)
+        .flatMap { comm =>
+          comm.consume.channelsHashes ++ comm.produces.map(_.channelsHash)
         }
         .toSet
 
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -86,19 +86,29 @@ object EstimatorHelper {
       ancestors = maybeAncestors.flatten
       ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
         ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))
-      commEvents = ancestorEvents.filter {
-        case Event(Comm(CommEvent(Some(_: ConsumeEvent), _))) =>
-          true
-        case _ => false
+      (produceEvents, consumeEvents, commEvents) = ancestorEvents.foldLeft(
+        (List.empty[ProduceEvent], List.empty[ConsumeEvent], List.empty[CommEvent])
+      ) {
+        case ((produces, consumes, commEvents), ancestorEvent) =>
+          ancestorEvent match {
+            case Event(Produce(produce: ProduceEvent)) =>
+              (produces :+ produce, consumes, commEvents)
+            case Event(Consume(consume: ConsumeEvent)) =>
+              (produces, consumes :+ consume, commEvents)
+            case Event(Comm(commEvent: CommEvent)) =>
+              (produces, consumes, commEvents :+ commEvent)
+            case _ =>
+              throw new RuntimeException("Unexpected ancestor event")
+          }
       }
       producesInCommEvents = commEvents.flatMap {
-        case Event(Comm(CommEvent(Some(_: ConsumeEvent), produces))) =>
+        case CommEvent(Some(_: ConsumeEvent), produces) =>
           produces
         case _ =>
           throw new RuntimeException("Unexpected comm event")
       }
       consumeInCommEvents = commEvents.map {
-        case Event(Comm(CommEvent(Some(consume: ConsumeEvent), _))) =>
+        case CommEvent(Some(consume: ConsumeEvent), _) =>
           consume
         case _ =>
           throw new RuntimeException("Unexpected comm event")
@@ -117,16 +127,13 @@ object EstimatorHelper {
             consume.channelsHashes.toSet
           }
         case Event(Comm(CommEvent(Some(consume: ConsumeEvent), produces))) =>
-          val consumeChannelHashes = if (consumeInCommEvents.contains(consume)) {
-            Set.empty[ByteString] // Volatile consume
+          if (consumeEvents.contains(consume) && produces.forall(
+                produce => produceEvents.contains(produce)
+              )) {
+            Set.empty[ByteString] // Volatile consume & produces
           } else {
-            consume.channelsHashes.toSet
+            consume.channelsHashes.toSet ++ produces.map(_.channelsHash).toSet
           }
-          val produceChannelHashes = produces
-            .filterNot(produce => producesInCommEvents.contains(produce))
-            .map(_.channelsHash)
-            .toSet
-          consumeChannelHashes ++ produceChannelHashes
         case _ => throw new RuntimeException("incorrect ancestor events")
       }.toSet
     } yield ancestorChannels
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -75,6 +75,21 @@ object EstimatorHelper {
         } yield conflicts
     }
 
+  val extractProduce: Event => Option[ProduceEvent] = {
+    case Event(Produce(produce: ProduceEvent)) => Some(produce)
+    case _                                     => None
+  }
+
+  val extractConsume: Event => Option[ConsumeEvent] = {
+    case Event(Consume(consume: ConsumeEvent)) => Some(consume)
+    case _                                     => None
+  }
+
+  val extractComm: Event => Option[CommEvent] = {
+    case Event(Comm(commEvent: CommEvent)) => Some(commEvent)
+    case _                                 => None
+  }
+
   @SuppressWarnings(Array("org.wartremover.warts.Throw"))
   private def buildBlockAncestorChannels[F[_]: Monad: BlockStore](
       blockAncestorsMeta: List[BlockMetadata]
@@ -86,55 +101,40 @@ object EstimatorHelper {
       ancestors = maybeAncestors.flatten
       ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
         ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))
-      (produceEvents, consumeEvents, commEvents) = ancestorEvents.foldLeft(
-        (List.empty[ProduceEvent], List.empty[ConsumeEvent], List.empty[CommEvent])
-      ) {
-        case ((produces, consumes, commEvents), ancestorEvent) =>
-          ancestorEvent match {
-            case Event(Produce(produce: ProduceEvent)) =>
-              (produces :+ produce, consumes, commEvents)
-            case Event(Consume(consume: ConsumeEvent)) =>
-              (produces, consumes :+ consume, commEvents)
-            case Event(Comm(commEvent: CommEvent)) =>
-              (produces, consumes, commEvents :+ commEvent)
-            case _ =>
-              throw new RuntimeException("Unexpected ancestor event")
-          }
-      }
-      producesInCommEvents = commEvents.flatMap {
-        case CommEvent(Some(_: ConsumeEvent), produces) =>
-          produces
-        case _ =>
-          throw new RuntimeException("Unexpected comm event")
-      }
-      consumeInCommEvents = commEvents.map {
-        case CommEvent(Some(consume: ConsumeEvent), _) =>
-          consume
-        case _ =>
-          throw new RuntimeException("Unexpected comm event")
-      }
-      ancestorChannels = ancestorEvents.flatMap {
-        case Event(Produce(produce: ProduceEvent)) =>
-          if (producesInCommEvents.contains(produce)) {
-            Set.empty[ByteString] // Volatile produce
-          } else {
-            Set(produce.channelsHash)
-          }
-        case Event(Consume(consume: ConsumeEvent)) =>
-          if (consumeInCommEvents.contains(consume)) {
-            Set.empty[ByteString] // Volatile consume
-          } else {
-            consume.channelsHashes.toSet
-          }
-        case Event(Comm(CommEvent(Some(consume: ConsumeEvent), produces))) =>
-          if (consumeEvents.contains(consume) && produces.forall(
-                produce => produceEvents.contains(produce)
-              )) {
-            Set.empty[ByteString] // Volatile consume & produces
-          } else {
-            consume.channelsHashes.toSet ++ produces.map(_.channelsHash).toSet
-          }
-        case _ => throw new RuntimeException("incorrect ancestor events")
-      }.toSet
+
+      produceEvents = ancestorEvents.map(extractProduce).flatten
+      consumeEvents = ancestorEvents.map(extractConsume).flatten
+      commEvents    = ancestorEvents.map(extractComm).flatten
+
+      producesInCommEvents = commEvents.flatMap(_.produces)
+      consumeInCommEvents  = commEvents.flatMap(_.consume)
+
+      ancestorProduceChannels = produceEvents
+        .filterNot(producesInCommEvents.contains(_))
+        .map(p => p.channelsHash)
+        .toSet
+
+      ancestorConsumeChannels = consumeEvents
+        .filterNot(consumeInCommEvents.contains(_))
+        .flatMap(c => c.channelsHashes)
+        .toSet
+
+      ancestorCommChannels = commEvents
+        .filterNot {
+          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+            consumeEvents.contains(consume) && produces.forall(
+              produce => produceEvents.contains(produce)
+            )
+          case _ => false
+        }
+        .flatMap {
+          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+            consume.channelsHashes ++ produces.map(_.channelsHash)
+          case e @ CommEvent(None, _) =>
+            throw new RuntimeException(s"Unexpected comm event $e")
+        }
+        .toSet
+
+      ancestorChannels = ancestorProduceChannels ++ ancestorConsumeChannels ++ ancestorCommChannels
     } yield ancestorChannels
 }
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -116,12 +116,12 @@ object EstimatorHelper {
 
       ancestorProduceChannels = produceEvents
         .filterNot(producesInCommEvents.contains(_))
-        .map(p => p.channelsHash)
+        .map(_.channelsHash)
         .toSet
 
       ancestorConsumeChannels = consumeEvents
         .filterNot(consumeInCommEvents.contains(_))
-        .flatMap(c => c.channelsHashes)
+        .flatMap(_.channelsHashes)
         .toSet
 
       ancestorCommChannels = commEvents
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -123,7 +123,7 @@ object EstimatorHelper {
         .filterNot {
           case CommEvent(Some(consume: ConsumeEvent), produces) =>
             !consume.persistent && consumeEvents.contains(consume) && produces.forall(
-              produce => produceEvents.contains(produce)
+              produce => !produce.persistent && produceEvents.contains(produce)
             )
           case _ => false
         }
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -75,21 +75,6 @@ object EstimatorHelper {
         } yield conflicts
     }
 
-  val extractProduce: Event => Option[Produce] = {
-    case produce: Produce => Some(produce)
-    case _                => None
-  }
-
-  val extractConsume: Event => Option[Consume] = {
-    case consume: Consume => Some(consume)
-    case _                => None
-  }
-
-  val extractComm: Event => Option[COMM] = {
-    case commEvent: COMM => Some(commEvent)
-    case _               => None
-  }
-
   private def isVolatile(comm: COMM, consumes: Seq[Consume], produces: Seq[Produce]) =
     !comm.consume.persistent && consumes.contains(comm.consume) && comm.produces.forall(
       produce => !produce.persistent && produces.contains(produce)
@@ -107,9 +92,9 @@ object EstimatorHelper {
         ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog)))
         .map(EventConverter.toRspaceEvent)
 
-      produceEvents = ancestorEvents.map(extractProduce).flatten
-      consumeEvents = ancestorEvents.map(extractConsume).flatten
-      commEvents    = ancestorEvents.map(extractComm).flatten
+      produceEvents = ancestorEvents.collect { case p: Produce => p }
+      consumeEvents = ancestorEvents.collect { case c: Consume => c }
+      commEvents    = ancestorEvents.collect { case c: COMM    => c }
 
       producesInCommEvents = commEvents.flatMap(_.produces)
       consumeInCommEvents  = commEvents.map(_.consume)
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -86,13 +86,47 @@ object EstimatorHelper {
       ancestors = maybeAncestors.flatten
       ancestorEvents = ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
         ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog))
+      commEvents = ancestorEvents.filter {
+        case Event(Comm(CommEvent(Some(_: ConsumeEvent), _))) =>
+          true
+        case _ => false
+      }
+      producesInCommEvents = commEvents.flatMap {
+        case Event(Comm(CommEvent(Some(_: ConsumeEvent), produces))) =>
+          produces
+        case _ =>
+          throw new RuntimeException("Unexpected comm event")
+      }
+      consumeInCommEvents = commEvents.map {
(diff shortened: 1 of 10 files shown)


casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,33 +59,152 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+  object Nil extends Rho("Nil")
+
+  def conflictsForNow(b1: Rho, b2: Rho, base: Rho) = conflicts(b1, b2, base)
+
+  private def conflicts(b1: Rho, b2: Rho, base: Rho)(
+      implicit file: sourcecode.File,
+      line: sourcecode.Line
+  ) =
+    diamondConflictCheck(base, b1, b2, numberOfParentsForDiamondTip = 1) /*>>
+      diamondConflictCheck(base, b2, b1, numberOfParentsForDiamondTip = 1)*/
```
dzajkowski: what's the point of these comments?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,33 +59,152 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      diamondConflictCheck(base, b2, b1, numberOfParentsForDiamondTip = 2)*/
+
+  private def coveredBy(equivalent: String) = ().pure[Effect]
+
+  private def diamondConflictCheck(
+      base: Rho,
+      b1: Rho,
+      b2: Rho,
+      numberOfParentsForDiamondTip: Int
+  )(implicit file: sourcecode.File, line: sourcecode.Line): Effect[Unit] =
```
dzajkowski, line 248: 👍

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,33 +59,152 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "4! 4! diff !" -> conflictsForNow(F0, F1, S0 | S1),
+      // Skipping 4! 4!! merges, 4! !!X merges, 4! !!4 merges
+      "4! CX"        -> conflictsForNow(F_, C1, S0),
+      "4! C! same !" -> conflictsForNow(F_, C_, S0),
+      "4! C! diff !" -> conflictsForNow(F0, C1, S0 | S1),
+      "CX CX"        -> conflictsForNow(C_, C_, Nil),
+      "C! C! same !" -> conflicts(C_, C_, S0),
+      "C! C! diff !" -> conflictsForNow(C0, C1, S0 | S1)
+      // 4!! / !!4 row is similar to !4 / 4! and thus skipped
+      // C!! / !!C row is similar to !C / C! and thus skipped
```
dzajkowski, line 204: is there a reason to not test this? or marking it with `coveredBy("!X (4!!)")`?

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -86,14 +101,46 @@ object EstimatorHelper {
...
+            !consume.persistent && consumeEvents.contains(consume) && produces.forall(
+              produce => !produce.persistent && produceEvents.contains(produce)
+            )
+          case _ => false
+        }
+        .flatMap {
+          case CommEvent(Some(consume: ConsumeEvent), produces) =>
+            consume.channelsHashes ++ produces.map(_.channelsHash)
+          case e @ CommEvent(None, _) =>
+            throw new RuntimeException(s"Unexpected comm event $e")
```
dzajkowski: raiseError will not work?

dzajkowski (approved):
some small questions

ArturGajowy (changes requested):
My comments from Kent's PR still apply, I'll do the review of new changes (which seem to look good!) tomorrow. Please address the aforementioned comments though.

ArturGajowy:
Also, may I ask you to edit my `WIP pick symmetric variants of dag at random in mergability tests` commit to do what its title says (and no longer be WIP)? Basically it should switch places of the arguments b1 and b2 at random.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,33 +59,152 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    val F0 = Rho("for (@0 <- @0) { 0 }")
+    val F1 = Rho("for (@1 <- @0) { 0 }")
+    // Contracts (persistent receives)
+    val C_ = Rho("contract @0(id) = { 0 }")
+    val C0 = Rho("contract @0(@0) = { 0 }")
+    val C1 = Rho("contract @0(@1) = { 0 }")
+    //FIXME add missing cases for in-deploy COMM-s wherever there's a pair without X (a COMM)
+    //FIXME all `conflictsForNow` should eventually be replaced with `merges`
+    Map(
+      "!X !X"        -> conflictsForNow(S0, S0, Nil),
```
dzajkowski: it would be so cool to have an additional field with reference to the reason why it conflicts for now; ticket, page, discord discussion, anything that would allow someone not familiar with the (large) problem space to catch up and see what needs to be done to enable this.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,33 +59,152 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!X !4"        -> conflictsForNow(S0, S0, F_),
+      "!X (!4)"      -> merges(S0, S1 | F_, Nil),
```
dzajkowski: I don't follow this case. why consume `S1` and not `S0`?

dzajkowski: maybe to expand on my lack of understanding - it is confusing to see `"!X !4"        -> conflictsForNow(S0, S0, F_),` and `"!X (!4)"      -> merges(S0, S1 | F_, Nil),`. why the difference in data sent? does it matter?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,161 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!X !C"        -> conflictsForNow(S0, S1, C1),
```
dzajkowski: the document describing mergeability (TDDM?) comments here:
`If one produce got consumed and one didn't, they must be sending on different names`
this case sends data on the same name (`@0`)  so the result should be `!C !C`

is the assumption that this will be fixed somehow? I think that this isn't a possible state as `!X` will never happen with a `C` present on a common name. 

the next case I'm able to follow. there we introduce `C` on the right and it has to conflict.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,161 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!X (!C)"      -> conflicts(S0, S1 | C_, Nil),
+      "!X 4X"        -> conflicts(S0, F_, Nil),
```
dzajkowski: TDDM: `If we can prove they don't match, then they can merge.`

are there any plans to address this case? I guess this would mean running the SpatialMatcher.

lukasz-golebiewski: I don't think running the SM will be necessary. No match === no COMM event and this is all the info we need and have

lukasz-golebiewski: This one conflicts because `F_` and `S0` do match

dzajkowski: that's the point: IF we can show that `!X` would not match with `F_` it's still fine to merge.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,161 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!X (4!!)"     -> conflictsForNow(S0, F_ | R0, Nil),
+      "!X !!X"       -> conflictsForNow(S0, R0, Nil),
+      "!X !!4"       -> conflictsForNow(S0, R1, F1),
+      "!X (!!4)"     -> coveredBy("!X (4!!)"),
+      "!X CX"        -> conflicts(S0, C_, Nil),
+      "!X C!"        -> conflicts(S0, C_, S0),
+      "!X (C!)"      -> conflicts(S0, C_ | S0, Nil),
+      "!4 !4 same 4" -> conflicts(S0, S1, F_),
+      "!4 !4 diff 4" -> conflicts(S0, S1, F0 | F1),
+      "!4 (!4)"      -> merges(S0, S1 | F1, F0),
```
dzajkowski: what about: `"!4 (!4)"      -> conflicts(S0, S1 | F0, F0),`?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,161 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "(!4) (!4)"    -> merges(S0 | F_, S0 | F_, Nil),
+      "!4 !C"        -> conflictsForNow(S0, S1, F0 | C1),
```
dzajkowski: why is this marked as 'for now'? isn't this illegal?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,161 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!4 4X"        -> conflictsForNow(S0, F_, F_),
+      "!4 4!"        -> conflictsForNow(S0, F_, F0 | S1),
```
dzajkowski: are we fine with having `F0` produce on `@0` as a result?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: This is already fixed

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
lukasz-golebiewski: I changed the approach and this is no longer needed

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: This looks like an error, it should read `      "!X (!4)"      -> merges(S0, S0 | F_, Nil),`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Right, so we need to add `"!X 4X - no match"        -> conflictsForNow(S0, F1, Nil),`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: irrelevant

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: it does not introduce new cases (and it should not conflict)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: seems its legal

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: missed the pattern match in C1

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!4 4!!"       -> conflictsForNow(S0, F_, F0 | R1),
+      "!4 !!X"       -> conflictsForNow(S0, R1, F0),
+      "!4 !!4"       -> conflictsForNow(S0, R1, F_ | F1),
+      "!4 CX"        -> conflictsForNow(S0, C_, F_),
+      "!4 C!"        -> conflictsForNow(S0, C_, F0 | S1),
```
dzajkowski: shouldn't this `C_` be `C1`?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!C !C"        -> conflictsForNow(S0, S0, C_),
+      "!C 4X"        -> conflictsForNow(S0, F_, C_),
+      "!C 4!"        -> conflictsForNow(S0, F_, C0 | S1),
+      "!C 4!!"       -> conflictsForNow(S0, F_, C0 | R1),
```
dzajkowski: I don't agree with this usage of `_`. why not have it match exactly and remove any possibility for misunderstanding?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!C !!X"       -> conflictsForNow(S0, R1, C0),
+      "!C !!4"       -> conflictsForNow(S0, R1, C0 | F1),
```
dzajkowski: I think that it would be beneficial to have a `infinite` or `impossible` marker that could hold e.g. `4X C!!` for the sake of completeness.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "!C CX"        -> conflictsForNow(S0, C_, C_),
+      "!C C!"        -> conflictsForNow(S0, C_, C0 | S1),
+      "4X 4X"        -> conflictsForNow(F_, F_, Nil),
+      "4X 4!"        -> conflicts(F_, F_, S0),
+      // Skipping 4X 4!! merges, 4X !!X may merge or not, 4X !!4 may merge or not
+      "4X CX"        -> conflicts(F_, C_, Nil),
```
dzajkowski: shouldn't this be marked as "for now"?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: The JIRA ticket would be a better place than code

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -84,16 +88,35 @@ object EstimatorHelper {
...
-    } yield ancestorChannels
+      ancestorEvents = (ancestors.flatMap(_.getBody.deploys.flatMap(_.deployLog)) ++
+        ancestors.flatMap(_.getBody.deploys.flatMap(_.paymentLog)))
+        .map(EventConverter.toRspaceEvent)
+
+      produceEvents = ancestorEvents.collect { case p: Produce => p }
+      consumeEvents = ancestorEvents.collect { case c: Consume => c }
+      commEvents    = ancestorEvents.collect { case c: COMM    => c }
+
+      producesInCommEvents = commEvents.map(_.produces).flatten
```
ArturGajowy: flatMap

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Done

casper/src/test/scala/coop/rchain/casper/EstimatorHelperTest.scala
```diff
@@ -88,14 +88,14 @@ class EstimatorHelperTest
     }
   }
 
-  it should "conflict if their deploys contain same channel in deployLog" in withStorage {
+  ignore should "conflict if their deploys contain same channel in deployLog" in withStorage {
     implicit blockStore => implicit blockDagStorage =>
       testConflict[Task] { deploy =>
```
ArturGajowy, line 94: these two tests need to be fixed. Their names should be:

`detect conflicts within deployLog-s`
`detect conflicts within paymentLog-s`

there should also be a third one:

`detect conflicts between deployLog and paymentLog`

The bodies should be virtually the same, except `testConflict` will now take two deploy modifying functions: one for each deploy.

first deploy should do a produce on channel A
second deploy should do a consume on channel A
(this conflicts)

the only difference between those three tests is which logs contain the above events.

Ideally, we'd make the 3 tests say:

```scala
val deployLogSetter = deploy => log => deploy.copy(deployLog = log)
val paymentLogSetter = deploy => log => deploy.copy(paymnentLog = log)
testConflicts(deployLogSetter, deployLogSetter)
testConflicts(paymentLogSetter, paymentLogSetter)
testConflicts(deployLogSetter, paymentLogSetter)
```
//lenses ftw!

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    diamondConflictCheck(base, shuffledBlocks(0), shuffledBlocks(1), numberOfParentsForDiamondTip)
   }
 
+  private def diamondConflictCheck(
+      base: Rho,
+      b1: Rho,
+      b2: Rho,
+      numberOfParentsForDiamondTip: Int
+  )(implicit file: sourcecode.File, line: sourcecode.Line): Effect[Unit] =
+    Debug.print[Effect](base, b1, b2, numberOfParentsForDiamondTip) >>
```
ArturGajowy, line 249: please remove this, ideally by editing the commit that introduced it. Sorry about that.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
-        ConstructDeploy.sourceDeploy(echoContract(2), timestamp = time + 2)
-      )
-      for {
-        block0 <- nodes(0).addBlock(deploys(0))
-        block1 <- nodes(1).addBlock(deploys(1))
-        _      <- nodes(0).receive()
-        _      <- nodes(1).receive()
-        _      <- nodes(0).receive()
-        _      <- nodes(1).receive()
+    def echoContract(no: Int) =
```
ArturGajowy, line 64: maybe un-ignore the above test and use `conflictsForNow`?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: I agree, although the general answer would be 'until we implement the corresponding merging rule / until we make it true'. In most cases, there are no longer objective justifications for them to merge, other than development time.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Yes, you're right

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I agree, this is a good idea. It should make grokking the whole thing much easier

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: I'm not sure if we should ever use the spatial matcher for refining those types of conflicts. We should reserve `conflictsForNow` for cases where it's clear that it should merge, and we can implement that quite easily.

ArturGajowy: Having a `consider refining conflict detection by using spatial matcher` ticket is what we need here, I'd say

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Shouldn't this `conflictForNow` and eventually `merge`? The impossible case is going to run out of phlo and so the deploy is going to be reverted without touching the tuplespace. So there shouldn't be any conflicts. Correct me if I'm wrong.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: It could, but this makes the test more general. Notice there's no other send the C_ could match.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: I disagree :) This makes the tests more general and gives more chance for catching a bug when s/b modifies them wrongly / we break sth in code. We should only constrain the receives with concrete data when it's needed for the test to produce the events we need. Otherwise we should stick with the general `F_/C_` case.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: depends on what the interpreter will do :D

ArturGajowy: `infinite/impossible` are just `merges`. Maybe it'd make sense to add a 'cause' enum to merges, so e.g.

```scala
val INFINITE = "Infinite loop is an error. Mergeable with everything."
merges(S0, R0, C_, INFINITE)
```

I'd maybe leave it for a future PR though?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: The only reason is the amount of cases to test.
I originally asked for 36 cases, for everything in Mike's sheet except for pairs containing `!!` or peek. This though grows by the following factors:

 - whenever there's "match depending whether its same linear receive/send" we get one more case (same / diff)
 - whenever there's a comm event in the case, we need to test it locally to test the actual topic of this pr - volatile events. So eg. `4X !!4` spawns `4X (!!4)`, and `4! 4!` spawns `4! (4!)` and `(4!) (4!)`

BTW @lukasz-golebiewski I think we still need to add the "spawned" cases

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -123,7 +260,7 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
 
     val registryRho =
       """
-        |// Expected output
+        |"Expected output".
```
ArturGajowy: Why this change? Doesn't it break the test?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      "4! C! same !" -> conflictsForNow(F_, C_, S0),
+      "4! C! diff !" -> conflictsForNow(F0, C1, S0 | S1),
+      "CX CX"        -> conflictsForNow(C_, C_, Nil),
+      "C! C! same !" -> conflicts(C_, C_, S0),
+      "C! C! diff !" -> conflictsForNow(C0, C1, S0 | S1),
+      // 4!! / !!4 row is similar to !4 / 4! and thus skipped
+      // C!! / !!C row is similar to !C / C! and thus skipped
+      "CX !!X no match" -> conflictsForNow(R0, C1, Nil),
+      "CX !!X match"    -> conflicts(R0, C0, Nil)
+    ).values.toList.parSequence_
```
ArturGajowy, line 206: We should verify that we have cases generated by the following code,
which amounts to 69 cases (not including same/diff distinction).

We'd check that by:

having the Map defining the test cases named `tests`
asserting that all cases `c` defined in `withLocals` have a key in the `tests` map that starts with `c`

```

val all = List("!X", "!4", "!C", "4X", "4!", "4!!", "PX", "P!", "P!!", "!!X", "!!4", "!!C", "CX", "C!", "C!!")
val considered = all.filterNot(_.contains("P")).filterNot(_.contains("!!"))

val pairs = considered.combinations(2)
val diagonal = considered.map(x => List(x, x))
val cases = (pairs ++ diagonal).toList

def isComm(s: String) = !s.contains("X")
def makeLocal(s: String): List[String] = isComm(s) match {
  case false => List(s)
  case true => List(s, s"($s)")
}

def makeLocals(v: List[String]): List[List[String]] = for {
  a <- makeLocal(v(0))
  b <- makeLocal(v(1))
} yield List(a, b)

val withLocals = cases.flatMap(makeLocals)

val s = withLocals.map(_.mkString(" ")).mkString("\n")
```

output:

```
!X !4
!X (!4)
!X !C
!X (!C)
!X 4X
!X 4!
!X (4!)
!X CX
!X C!
!X (C!)
!4 !C
!4 (!C)
(!4) !C
(!4) (!C)
!4 4X
(!4) 4X
!4 4!
!4 (4!)
(!4) 4!
(!4) (4!)
!4 CX
(!4) CX
!4 C!
!4 (C!)
(!4) C!
(!4) (C!)
!C 4X
(!C) 4X
!C 4!
!C (4!)
(!C) 4!
(!C) (4!)
!C CX
(!C) CX
!C C!
!C (C!)
(!C) C!
(!C) (C!)
4X 4!
4X (4!)
4X CX
4X C!
4X (C!)
4! CX
(4!) CX
4! C!
4! (C!)
(4!) C!
(4!) (C!)
CX C!
CX (C!)
!X !X
!4 !4
!4 (!4)
(!4) !4
(!4) (!4)
!C !C
!C (!C)
(!C) !C
(!C) (!C)
4X 4X
4! 4!
4! (4!)
(4!) 4!
(4!) (4!)
CX CX
C! C!
C! (C!)
(C!) C!
(C!) (C!)
```

ArturGajowy (changes requested):
We should change the name of this PR to 'Volatile events mergable", because that's the scope of mergability that it provides. We should also link to the relevant jira issue (not the 'mergable contract calls' one - or we need to split that one)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -53,34 +61,163 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    val C_ = Rho("contract @0(id) = { 0 }")
+    val C0 = Rho("contract @0(@0) = { 0 }")
+    val C1 = Rho("contract @0(@1) = { 0 }")
+    //FIXME add missing cases for in-deploy COMM-s wherever there's a pair without X (a COMM)
+    //FIXME all `conflictsForNow` should eventually be replaced with `merges`
+    Map(
+      "!X !X"        -> conflictsForNow(S0, S0, Nil),
+      "!X !4"        -> conflictsForNow(S0, S0, F_),
+      "!X (!4)"      -> merges(S0, S0 | F_, Nil),
+      "!X !C"        -> conflictsForNow(S0, S1, C1),
```
ArturGajowy: This should say
```
"!X !C" -> conflictsForNow(S0, S0, C_)
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Notice also that using spatial matcher requires having actual data and continuations, not just hashes. So it'd require a 'report-type' replay.

ArturGajowy:
I think I've managed to classify TDDM's cases into 4 types, that will allow for clean implementation of the underlying merging rules.

It also occurs to me, that there are really just 2 merge criteria:
- the things merged cannot cause additional comms (can't match each other - we need to find them matches that prevent that in the original deploys)
- the things merged cannot be consuming the same linear event

The split that occured to me is:
- whether the incoming events have same or different polarity (send/receive).
  same polarity cant produce additional comms
  different polarity cant match the same linear event
- within same/diff polarity category - whether the considered case could lead to the wrong specific to its polarity category

See the description of cases:
SAME_POLARITY_MERGE
COULD_MATCH_SAME
HAD_ITS_MATCH
INCOMING_COULD_MATCH

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I added the `"!X 4X - no match" -> conflictsForNow(S0, F1, Nil),`. Making this and related cases merge is outside the scope of this PR

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Future PR sounds good

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: I was under the impression that we want to have a set of "rules" that describe how merging works.  the `!4 C!` has a very specific meaning. introducing generality in the implementation obfuscates the meaning. note that there is no generality in the term. it is a description of state.
IOW `!4 C!` is not the same as (S0, C_, F0 | S1). 
I understand that replacing the `_` will not yield equality but it will bring the implementation closer to what it should hold.

IF we had means to express the left hand in a better way than by discriminating with patterns - maybe it would be ok to try and write general laws. but we don't.

this is my 0.02$, do with it as you please.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: No idea, I didn't author it. Will revert though as it doesn't look sensible

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: exactly my point. link to a spot that describes what needs to change so the bus factor is >1.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Why?
Besides, `S0, S0, C_` will always conflict

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
dzajkowski: I was not arguing to solve it in this pr.
We have a description of how the merging rules should behave but we are ignoring half of the data there. who will remember to pick up that work? if it's part of this compiling documentation - we have a spot that tracks it.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I would argue that it's better to remove ambiguity during execution and split all cases containing `F_` into containing `F0` and `F1`. In any case let's not do that in this PR

casper/src/test/scala/coop/rchain/casper/EstimatorHelperTest.scala
lukasz-golebiewski:
```
first deploy should do a produce on channel A
second deploy should do a consume on channel A
(this conflicts)
```
This is a huge oversimplification as this isn't always going to conflict because the produce and consume might not match. Hence I would argue that this test should be removed given how fine grained the tests in `MultiParentCasperMergeSpec` are going to be

ArturGajowy: > the produce and consume might not match
remember we're in control of the code used for doing the deploy. The `testConfclits` funciton ensures they do match, by not specifying a pattern on the for.

MergeSpec can't test inclusion of payment logs until they're actually used, and is far heavier than this is. If it's such a pain - leave those ignored, I'll fix it when I come to it.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: yeah, sorry - I was wrong about this one. It's correct.
(`"!X !C"        -> conflictsForNow(S0, S1, C1),`)

casper/src/test/scala/coop/rchain/casper/EstimatorHelperTest.scala
lukasz-golebiewski: Nah, I'll unignore them

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Ok, in a follow up PR

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Notice we're talking about a test, not an implementation. The more general the tests, the more constrained the tested implementation.

Take for example `merges(S0, C_, F_)`.
Replacing it with `merges(S0, C0, F0)` poses the risk that the tested code will  only merge when the receives have the exact same matching pattern.

Granted, one could argue that `merges(S0, C_, F_)` will not spot cases where `merges(S0, C_, F0)` or `merges(S0, C0, F_)` would not hold. That would though require an odd explicit condition in the prod code, and can be fought by adding additional cases for specific data.

In general, I believe it's good to keep the test data minimal (C_ < C0) - which helps a better implementation, less over-specifying tests, and stays as close to the actual rules of the tested domain as possible.

ArturGajowy (approved):

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Just to be clear:

> I don't think running the SM will be necessary. No match === no COMM event and this is all the info we need and have

It will. Notice we're talking about merging two deploys (and their events) that weren't executing together. E.g.:

I deploy `@0!(0)`
You deploy `for (_ <- @0) { @1(1) }`

None of the deploys will have a COMM (or a match) in their log, but we can't merge them, as they'd leave TS in undefined state (more work is needed).

lukasz-golebiewski: True

PR #2611, opened 2019-07-23
Update Par evaluation to reduce extraneous thread creation

JosephDenman:
### Summary:
Previous to this PR, the Reducer would create a list of six EvalJobs, each of which would represent a sequence of processes to run in parallel. There were six because there are six different kinds of top-level processes. Then, it would spawn a thread for each EvalJob, so six threads. Then, each EvalJob would process its sequence of processes in parallel, spawning a thread for each process. These threads represent the execution of the actual processes. That means, in the worst case where you have a program with where all Pars use all six different kinds of top-level processes, six threads were needlessly introduced (one for each EvalJob) per Par. With these changes, the spawned threads map 1-to-1 to the top-level processes in the Par. Plus, we can get rid of the EvalJob craziness.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3699

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -149,6 +149,12 @@ class DebruijnInterpreter[M[_], F[_]](
     ) >>= (go(_))
   }
 
+  private def reportErrors(process: M[Unit]): M[Unit] =
+    process.handleErrorWith {
+      case e @ OutOfPhlogistonsError => e.raiseError[M, Unit]
+      case e                         => fTell.tell(e)
+    }
+
   private trait EvalJob {
     def run(mkRand: Int => Blake2b512Random)(
         env: Env[Par],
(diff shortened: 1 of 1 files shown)


ArturGajowy (approved):
That's a really nice refactoring, but one that should be split into several commits at least. THere's a lot of stuff happening and it's really not obvious that the changes are correct. And it should be obvious, especially when changing critical code like this.

Please split such changes into smaller commits, each commit doing one refactoring (e.g. extract 'handleError' method, remove evalExplicit, reindex the splits) and having a diff that does exactly what the title says.

JosephDenman:
Thanks! Will do.

JosephDenman:
bors try

tymm (approved):
LGTM but it's not obvious to me where the reduction in extraneous thread creation actually happens? Can you add a hint in the commit message of the commit where the reduction happens?

JosDenmark:
Previous to this PR, the Reducer would create a list of six `EvalJob`s, each of which would represent a sequence of processes to run in parallel. There were six because there are six different kinds of top-level processes. Then, it would spawn a thread for each `EvalJob`, so six threads. Then, each `EvalJob` would process its sequence of processes in parallel, spawning a thread for each process. These threads represent the execution of the actual processes. That means, in the worst case where you have a program with where all `Par`s use all six different kinds of top-level processes, six threads were needlessly introduced (one for each `EvalJob`) per `Par`. With these changes, the spawned threads map 1-to-1 to the top-level processes in the `Par`. Plus, we can get rid of the `EvalJob` craziness.

JosephDenman:
Tymm, I added an explanation in the summary section of the PR description.

PR #2620, opened 2019-07-25
[WIP] Use TreeMap in SortedParMap

JosephDenman:
## Overview
Uses a `TreeMap` in `SortedParMap` so that the entire map need not be re-sorted upon every modification. Three tests are failing, the big ugly commit needs to be separated into sensible commits, and if there's time the `ParMap` should be removed (that's what my intention was).


ArturGajowy:
What are the advantages of this approach over what Kent has in https://github.com/rchain/rchain/commit/b829ca2590bec4c1dfdbfbb2160a0203f86a6b25 ? That one looks more focused and easier to review.

JosephDenman:
The advantages of this are that it doesn't need a cache because the ordering information is preserved in the `TreeMap`. The `connective`, `locallyFree`, and `remainder` data from `ParMap` is included with `SortedParMap` because `ParMap` can be eliminated. Since it uses an underlying `HashMap`, lookups should be pretty speedy.

ArturGajowy:
> The advantages of this are that it doesn't need a cache because the ordering information is preserved in the TreeMap

That's super nice. What about sorting the inputs to get/getOrElse/contains? Kents cache approach handles that.

> connective, locallyFree, and remainder data from ParMap is included with SortedParMap because ParMap can be eliminated.

I'd wait with that till we see in the flamegraph that having this separation is an issue. Unless it's requried to have the former? Then a better commit split is needed.

> Since it uses an underlying HashMap, lookups should be pretty speedy.

We could apply that to Kent's change as well.

JosephDenman:
> That's super nice. What about sorting the inputs to get/getOrElse/contains? Kents cache approach handles that.

I think your intuition that sorting on inputs to `get`, `getOrElse`, and `contains` is correct. We should be using `SortedParMap` as a general collection, so if we assume that it's only utilized in `Reduce`, then we can narrow the cases where we'd (maybe) have to pre-sort the input to tests of `SortedParMap`. In Rholang, the methods on `SortedParMap` are only accessible through `EMethods` and some places in the registry, so if we know that `EMethod` sorts its arguments and we can verify that the registry calls do as well, then we're golden.

> I'd wait with that till we see in the flamegraph that having this separation is an issue. Unless it's requried to have the former? Then a better commit split is needed.

Agreed. WIP.

JosephDenman:
It actually looks like a `TreeSet` will suffice, since we only ever sort based on the keys and maps don't duplicate keys.

ArturGajowy:
TreeSet will suffice for what?

ArturGajowy:
>  if we assume that it's only utilized in Reduce

I know I proposed a change in that direction in #rholang, but I'd much rather have a correctness guarantee - even if that requires maintaining a cache.

JosephDenman:
We don't actually need the `TreeMap` because we don't actually need to do any lookups based on `ScoredTerm[Par]`. We can just keep a `TreeSet` around that keeps the keys sorted.

ArturGajowy:
I'm lost - you're using the TreeMap as a backed for the SortedParMap. Are you suggesting using a TreeSet of pairs?

Also, we most likely do need a separate hashmap for fast lookups stil.

JosephDenman:
Sorry. No. I use a `TreeMap` to keep the keys sorted, but never have to use the `TreeMap` for lookups because the `HashMap` is used. So, I can just keep the keys around in a `TreeSet` (sorted set).

ArturGajowy:
You probably shouldn't do that, or else the kv-pairs iterator will need to do lookups instead of leveraging a sorted kv-pairs collection

PR #2624, opened 2019-07-26
RCHAIN-3427: Add unforgeable name bound vaults

tymm:
## Overview
See title

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3427

### DIFF
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -39,11 +39,22 @@ in {
             DeployerIdOps!("pubKeyBytes", *deployerId, *deployerPubKeyBytesCh) |
             for (@deployerPubKeyBytes <- deployerPubKeyBytesCh) {
 
-              RevAddress!("fromPublicKey", deployerPubKeyBytes, *revAddrCh)|
+              RevAddress!("fromPublicKey", deployerPubKeyBytes, *revAddrCh) |
               for (@deployerRevAddress <- revAddrCh) {
 
                 @AuthKey!("make", (*_revVault, deployerRevAddress), *ret)
               }
+            }
+          }
+        } |
+
+        contract RevVault(@"unfAuthKey", unf, ret) = {
+
+          new RevAddress(`rho:rev:address`), revAddrCh in {
+            RevAddress!("fromUnfName", *unf, *revAddrCh) |
+            for (@unfRevAddress <- revAddrCh) {
+
+              @AuthKey!("make", (*_revVault, unfRevAddress), *ret)
             }
           }
         } |
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -48,10 +48,10 @@ in {
           }
         } |
 
-        contract RevVault(@"unfAuthKey", unf, ret) = {
+        contract RevVault(@"unforgeableAuthKey", unf, ret) = {
 
           new RevAddress(`rho:rev:address`), revAddrCh in {
-            RevAddress!("fromUnfName", *unf, *revAddrCh) |
+            RevAddress!("fromUnforgeable", *unf, *revAddrCh) |
             for (@unfRevAddress <- revAddrCh) {
 
               @AuthKey!("make", (*_revVault, unfRevAddress), *ret)
--- a/casper/src/test/resources/RevAddressTest.rho
+++ b/casper/src/test/resources/RevAddressTest.rho
@@ -5,7 +5,7 @@ new
   rl(`rho:registry:lookup`), RhoSpecCh,
   revAddress(`rho:rev:address`),
   stdlog(`rho:io:stdlog`),
-  test_valid_address, test_invalid_address, test_fromPublicKey, test_fromPublicKey_invalid, test_fromUnfName
+  test_valid_address, test_invalid_address, test_fromPublicKey, test_fromPublicKey_invalid, test_fromUnforgeable
 in {
   rl!(`rho:id:zphjgsfy13h1k85isc8rtwtgt3t9zzt5pjd5ihykfmyapfc4wt3x5h`, *RhoSpecCh) |
   for(@(_, RhoSpec) <- RhoSpecCh) {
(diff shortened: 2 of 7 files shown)


tymm:
@ArturGajowy suggested to add a prefix for RevAddresses based on unforgeable names. @odeac do you have an opinion on that?

casper/src/test/resources/RevAddressTest.rho
```diff
@@ -62,5 +63,13 @@ in {
       revAddress!("fromPublicKey", "F0F0".hexToBytes(), *retCh) |
       rhoSpec!("assert", (Nil, "== <-", *retCh), "correct RevAddress", *ackCh)
     }
+  } |
+
+  contract test_fromUnfName(rhoSpec, _, ackCh) = {
+    new unf1, unf2, r1, r2 in {
+      revAddress!("fromUnfName", *unf1, *r1) |
```
ArturGajowy: we should use the full name, `fromUnforgeableName`

ArturGajowy: Even better: `fromUnforgeable`. It's not that anyone can pass a name. We always send processes, and always receive names.

casper/src/test/resources/RevAddressTest.rho
```diff
@@ -62,5 +63,13 @@ in {
...
+      revAddress!("fromUnfName", *unf2, *r2) |
+      rhoSpec!("assert", (*r1, "!= <-", *r2), "expecting RevAddresses to be unequal", *ackCh)
```
ArturGajowy, line 73: we need a negative case too: pass a forgeable process (e.g. 42) and expect the result to be Nil

casper/src/main/resources/RevVault.rho
```diff
@@ -48,6 +48,17 @@ in {
           }
         } |
 
+        contract RevVault(@"unfAuthKey", unf, ret) = {
```
ArturGajowy: let's not use the abbreviation, let's call it `unforgeableNameAuthKey`. `unforgeableAuthKey` can be confusing w.r.t what's unforgeable.

casper/src/test/resources/RevAddressTest.rho
tymm: Can do but this would be inconsistent with how most (all?) system processes work. For example if you pass an unforgeable name to `RevAddress!fromPublicKey` it won't return `Nil` but will just stay in the tuple space.

ArturGajowy: I though I've seen a test saying otherwise:
https://github.com/rchain/rchain/blob/8bd29248267b484ad79e410207f7fe0c6694f89a/casper/src/test/resources/RevAddressTest.rho#L64

tymm: That is because `RevAddress!fromPublicKey` takes Rholang ByteArrays. Iff the ByteArray passed to `RevAddress!fromPublicKey` is not a valid PublicKey, then we return `Nil`. If you pass a Rholang String to `RevAddress!fromPublicKey` you won't get anything back though.

casper/src/test/resources/RevAddressTest.rho
tymm: Done

casper/src/main/resources/RevVault.rho
tymm: Done

casper/src/test/resources/RevAddressTest.rho
ArturGajowy: In both cases, not returning anything is the worst thing we can do. We should return Nil in both of those cases, if not an either with an error message.

casper/src/test/resources/RevAddressTest.rho
```diff
@@ -62,5 +64,20 @@ in {
...
+    new unf1, unf2, r1, r2 in {
+      revAddress!("fromUnforgeable", *unf1, *r1) |
+      revAddress!("fromUnforgeable", *unf2, *r2) |
+      rhoSpec!("assert", (*r1, "!= <-", *r2), "expecting RevAddresses to be unequal", *ackCh)
+    }
+  } |
+
+  contract test_fromUnforgeable_invalid(rhoSpec, _, ackCh) = {
+    new retCh in {
+      revAddress!("fromUnforgeable", 42, *retCh) |
```
ArturGajowy, line 79: We should add a similar test for revAddress!fromDeployerId that passes a non-deployerId (e.g. gprivate)

ArturGajowy (approved):
LGTM. Please consider adding the test I mentioned :)

casper/src/test/resources/RevAddressTest.rho
tymm: I thought about cleaning up all system processes in a separate PR but I am kind of blocked by not being sure on how to return an either from withing a system process. Any ideas?

ArturGajowy: `An Either is just a coproduct of two cases encoded as a tuple, what's the problem?` :stuck_out_tongue: 

I too would have to look up how to create `(false, "error msg")` or `(true, result)` in AST :)

ArturGajowy: Anyway, I think such a cleanup can cause an avalanche of changes. Let's then ticket it and for now leave as is.

tymm: Ouch! Was totally complicating it. Thanks...

tymm: Btw there is no `RevAddress!fromDeployerId` :)

ArturGajowy: Right! There should be! :joy:

ArturGajowy: I now see that we have deployerId -> pubKey and pubKey -> revAddress, so having deployerId -> revAddress would rather be a convenience than a necessity:)

odeac (approved):

PR #2631, opened 2019-07-31
Convert serializedSize calls to constants

KentShikama:
## Overview

Before on test against 300 vaults: 2m1s, 1m57s, and 1m59s
After on test against 300 vaults: 1m50s, 1m52s, and 1m50s

### JIRA ticket:

### Notes


ArturGajowy (changes requested):
Given in how many places the cost model relies on this, how this makes the cost model oblivious to data size, and how little performance we gain - I don't think this is a viable solution to any of our problems.

It made me realize though, that we don't charge for the genesis deploys anyway. What we should maybe do, is run the genesis without even counting the cost of execution. Ideally, we'd have all cost evaluation in a single tagless final algebra and - similarly to Span-s - just provide a noop implementation for genesis execution.

PR #2632, opened 2019-07-31
[RChain-2845] Test fault tolenrence

zsluedem:
## Overview
test fault tolerance

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-2845

### Notes

### DIFF
--- a/integration-tests/test/conftest.py
+++ b/integration-tests/test/conftest.py
@@ -86,15 +86,13 @@ def command_line_options(request: Any) -> Generator[CommandLineOptions, None, No
     yield command_line_options
 
 
-
 @contextlib.contextmanager
-def temporary_bonds_file(random_generator: Random, validator_keys: List[PrivateKey]) -> Generator[str, None, None]:
-    (fd, file) = tempfile.mkstemp(prefix="rchain-bonds-file-", suffix=".txt")
+def temporary_bonds_file(validator_bonds_dict: Dict[PrivateKey, int]) -> Generator[str, None, None]:
+    (fd, file) = tempfile.mkstemp(prefix="rchain-bonds-file-", suffix=".txt", dir="/tmp")
     try:
         with os.fdopen(fd, "w") as f:
-            for pair in validator_keys:
-                bond = random_generator.randint(1, 100)
-                f.write("{} {}\n".format(pair.get_public_key().to_hex(), bond))
+            for private_key, bond in validator_bonds_dict.items():
+                f.write("{} {}\n".format(private_key.get_public_key().to_hex(), bond))
         yield file
     finally:
         os.unlink(file)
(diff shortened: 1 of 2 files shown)


zsluedem:
bors try

zsluedem:
Sorry about taking this so long.

integration-tests/test/conftest.py
```diff
@@ -86,15 +86,13 @@ def command_line_options(request: Any) -> Generator[CommandLineOptions, None, No
     yield command_line_options
 
 
-
 @contextlib.contextmanager
-def temporary_bonds_file(random_generator: Random, validator_keys: List[PrivateKey]) -> Generator[str, None, None]:
-    (fd, file) = tempfile.mkstemp(prefix="rchain-bonds-file-", suffix=".txt")
+def temporary_bonds_file(validator_bonds_dict: Dict[PrivateKey, int]) -> Generator[str, None, None]:
```
woky, line 90: Are you sure that
```
validator_bonds_dict[PrivateKey.from_hex('foo')] = 123
validator_bonds_dict[PrivateKey.from_hex('foo')] == 123
```
? E.g. in Java you'd have to override `hashCode` so that two instances corresponding to a same key hash to same value. I don't know what's the equivalent in Python. I guess it's some `__foo__` method.

You only look up by unique instances per key so I guess it's fine, though it could mislead someone else.

integration-tests/test/test_dag_correctness.py
```diff
@@ -26,9 +27,85 @@
...
+    |                                                   b3=0.57894737
+    |                               b2=1
+    |           b1=1
+    |
+    |           Genesis=1
+    |
+    |           Bootstrap           Validator1          Validator2
+    | bonds         25                  20                  15
+    |
+    time
```
woky, line 48: Can you please point me to node source or some documentation so that I can learn how you came up with these values?

woky (approved):

integration-tests/test/conftest.py
zsluedem: Yes . You are right. It is `__hash__`

zsluedem: ~~I think it would be better to just use the hex to be the key here.~~

Let's hash the `byte` in the key.

zsluedem: https://github.com/rchain/pyrchain/pull/9 pr for that

KentShikama (approved):

integration-tests/test/test_dag_correctness.py
KentShikama: These values are based on the Safety Oracle: see https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala . A quick sanity check is to see that blocks lower in the dag are closer to finalized (near value of 1) and the blocks that are higher in the dag are less close to finalized (near value of -1). Also round robin-like and sandwiching patterns on top of a block should make it closer to finalized.

adaszko: > A quick sanity check is to see that blocks lower in the dag are closer to finalized (near value of 1) and the blocks that are higher in the dag are less close to finalized (near value of -1). Also round robin-like and sandwiching patterns on top of a block should make it closer to finalized.

In that case, I think the test shouldn't compare to exact numerical values but do something like you describe, i.e. `assert float(validator1.show_block_parsed(b3_hash)['faultTolerance']) < float(validator1.show_block_parsed(genesis_hash)['faultTolerance'])`, etc.

integration-tests/test/test_dag_correctness.py
```diff
@@ -26,9 +27,85 @@
...
+            wait_for_node_sees_block(context, validator2, b7_hash)
+
+            assert float(validator1.show_block_parsed(genesis_hash)['faultTolerance']) == pytest.approx(1.0)
+            assert float(validator1.show_block_parsed(b1_hash)['faultTolerance']) == pytest.approx(1.0)
+            assert float(validator1.show_block_parsed(b2_hash)['faultTolerance']) == pytest.approx(1.0)
+            assert float(validator1.show_block_parsed(b3_hash)['faultTolerance']) ==pytest.approx(0.57894737)
+            assert float(validator1.show_block_parsed(b4_hash)['faultTolerance']) ==pytest.approx(0.2631579)
+            assert float(validator1.show_block_parsed(b5_hash)['faultTolerance']) == pytest.approx(-1.0)
+            assert float(validator1.show_block_parsed(b6_hash)['faultTolerance']) == pytest.approx(-1.0)
+            assert float(validator1.show_block_parsed(b7_hash)['faultTolerance']) == pytest.approx(-1.0)
```
adaszko: See my other comment about using exact numerical values.

adaszko (changes requested):

adaszko (approved):

integration-tests/test/test_dag_correctness.py
```diff
@@ -26,9 +27,84 @@
 BONDED_VALIDATOR_KEY_4 = PrivateKey.from_hex("2a6018851984203e0983f0671e94fcf649ec04b614e5924f435081f7d1e3b44b")
 
 
-@pytest.mark.xfail
-def test_fault_tolerance(command_line_options: CommandLineOptions, docker_client: DockerClient) -> None:
-    assert False
+def test_fault_tolerance(command_line_options: CommandLineOptions, random_generator: Random, docker_client: DockerClient)-> None:
+    """
+    Below test would generate a dag like below. Every validator has the most latest block of other validators.
```
adaszko, line 32: Nitpick: s/most latest/latest/

PR #2633, opened 2019-07-31
Implement deploy indexing

itegulov:
## Overview
- Adds DAG storage index for deploys by their signatures to support efficient lookup
- Removes obsolete API for looking up blocks by their deploy creator and timestamp
- Adapts BlockAPI to use the new DAG storage index to lookup blocks by deploy signature

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3526

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
@@ -8,10 +8,10 @@ import cats.effect.concurrent.Semaphore
 import cats.effect.{Concurrent, Resource, Sync}
 import cats.implicits._
 import cats.mtl.MonadState
-
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.BlockDagFileStorage.{Checkpoint, CheckpointedDagInfo}
-import coop.rchain.blockstorage.util.BlockMessageUtil.{blockNumber, bonds, parentHashes}
+import coop.rchain.blockstorage.BlockDagStorage.DeployId
+import coop.rchain.blockstorage.util.BlockMessageUtil._
 import coop.rchain.blockstorage.util.byteOps._
 import coop.rchain.blockstorage.util.io.IOError.RaiseIOError
 import coop.rchain.blockstorage.util.io.{IOError, _}
@@ -24,14 +24,13 @@ import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
 import coop.rchain.shared.ByteStringOps._
 import coop.rchain.shared.Language.ignore
-import coop.rchain.shared.{AtomicMonadState, Log, LogSource}
-
+import coop.rchain.shared.{AtomicMonadState, Log, LogSource, StreamT}
 import monix.execution.atomic.AtomicAny
 import org.lmdbjava.DbiFlags.MDB_CREATE
 import org.lmdbjava.{Env, EnvFlags}
+
 import scala.ref.WeakReference
 import scala.util.matching.Regex
-
 import coop.rchain.metrics.{Metrics, MetricsSemaphore}
 import coop.rchain.metrics.Metrics.Source
 
@@ -40,6 +39,7 @@ private final case class BlockDagFileStorageState[F[_]: Sync](
     childMap: Map[BlockHash, Set[BlockHash]],
     dataLookup: Map[BlockHash, BlockMetadata],
     topoSort: Vector[Vector[BlockHash]],
+    blockHashesByDeploy: Map[DeployId, BlockHash],
     equivocationsTracker: Set[EquivocationRecord],
     invalidBlocks: Set[BlockMetadata],
     sortOffset: Long,
@@ -49,6 +49,8 @@ private final case class BlockDagFileStorageState[F[_]: Sync](
     latestMessagesCrc: Crc32[F],
     blockMetadataLogOutputStream: FileOutputStreamIO[F],
     blockMetadataCrc: Crc32[F],
+    blockHashesByDeployLogOutputStream: FileOutputStreamIO[F],
+    blockHashesByDeployCrc: Crc32[F],
     equivocationsTrackerLogOutputStream: FileOutputStreamIO[F],
     equivocationsTrackerCrc: Crc32[F],
     invalidBlocksLogOutputStream: FileOutputStreamIO[F],
@@ -67,6 +69,8 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
     equivocationTrackerCrcPath: Path,
     invalidBlocksLogPath: Path,
     invalidBlocksCrcPath: Path,
+    blockHashesByDeployLogPath: Path,
+    blockHashesByDeployCrcPath: Path,
     state: MonadState[F, BlockDagFileStorageState[F]]
 ) extends BlockDagStorage[F] {
   implicit private val logSource: LogSource = LogSource(BlockDagFileStorage.getClass)
@@ -79,6 +83,8 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
     state.get.map(_.dataLookup)
   private[this] def getTopoSort: F[Vector[Vector[BlockHash]]] =
     state.get.map(_.topoSort)
+  private[this] def getBlockHashesByDeploy: F[Map[DeployId, BlockHash]] =
+    state.get.map(_.blockHashesByDeploy)
   private[this] def getSortOffset: F[Long] =
     state.get.map(_.sortOffset)
   private[this] def getEquviocationsTracker: F[Set[EquivocationRecord]] =
@@ -97,6 +103,10 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
     state.get.map(_.blockMetadataLogOutputStream)
   private[this] def getBlockMetadataCrc: F[Crc32[F]] =
     state.get.map(_.blockMetadataCrc)
+  private[this] def getBlockHashesByDeployLogOutputStream: F[FileOutputStreamIO[F]] =
+    state.get.map(_.blockHashesByDeployLogOutputStream)
+  private[this] def getBlockHashesByDeployCrc: F[Crc32[F]] =
+    state.get.map(_.blockHashesByDeployCrc)
   private[this] def getEquivocationsTrackerLogOutputStream: F[FileOutputStreamIO[F]] =
     state.get.map(_.equivocationsTrackerLogOutputStream)
   private[this] def getEquivocationsTrackerCrc: F[Crc32[F]] =
@@ -145,6 +155,10 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
       f: Vector[Vector[BlockHash]] => Vector[Vector[BlockHash]]
   ): F[Unit] =
     state.modify(s => s.copy(topoSort = f(s.topoSort)))
+  private[this] def modifyBlockHashesByDeploy(
+      f: Map[DeployId, BlockHash] => Map[DeployId, BlockHash]
+  ): F[Unit] =
+    state.modify(s => s.copy(blockHashesByDeploy = f(s.blockHashesByDeploy)))
   private[this] def modifyEquivocationsTracker(
       f: Set[EquivocationRecord] => Set[EquivocationRecord]
   ): F[Unit] =
@@ -174,6 +188,7 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
       childMap: Map[BlockHash, Set[BlockHash]],
       dataLookup: Map[BlockHash, BlockMetadata],
       topoSortVector: Vector[Vector[BlockHash]],
+      blockHashesByDeploy: Map[DeployId, BlockHash],
       invalidBlocksSet: Set[BlockMetadata],
       sortOffset: Long
   ) extends BlockDagRepresentation[F] {
@@ -222,6 +237,8 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
       } else {
         false.pure[F]
       }
+    def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]] =
+      blockHashesByDeploy.get(deployId).pure[F]
     def topoSort(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
       if (startBlockNumber >= sortOffset) {
         val offset = startBlockNumber - sortOffset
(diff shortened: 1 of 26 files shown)


models/src/test/scala/coop/rchain/models/blockImplicits.scala
```diff
@@ -35,15 +38,60 @@ object blockImplicits {
 
   implicit val arbitraryJustification: Arbitrary[Justification] = Arbitrary(justificationGen)
 
+  val defaultSec = PrivateKey(
+    Base16.unsafeDecode("a68a6e6cca30f81bd24a719f3145d20e8424bd7b396309b0708a16c7d8000b76")
```
ArturGajowy: This should be replaced with ConstructDeploy.defaultSec, I think

models/src/test/scala/coop/rchain/models/blockImplicits.scala
```diff
@@ -35,15 +38,60 @@ object blockImplicits {
...
+        deployer = ByteString.copyFrom(Secp256k1.toPublic(defaultSec).bytes),
+        timestamp = timestamp,
+        term = term,
+        phloLimit = 90000,
+        phloPrice = 1L
+      )
+      signed = sign(defaultSec, deployData)
+    } yield ProcessedDeploy(
+      deploy = Some(signed)
+    )
```
ArturGajowy, line 60: quite a lot of this code seems to be duplicated from other modules. Rather than copying code to here, we should add any implicits needed for generating random BlockMessages in the module BlockMessage belongs - casper.

ArturGajowy: Also/alternatively: I don't think any of your tests rely on the deploys being signed or having proper phlo limit. Stick to totally random data maybe?

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -63,6 +67,8 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
     equivocationTrackerCrcPath: Path,
     invalidBlocksLogPath: Path,
     invalidBlocksCrcPath: Path,
+    blockHashesByDeployLogPath: Path,
+    blockHashesByDeployCrcPath: Path,
```
ArturGajowy, line 73: the amount of fields and places in code that need to change together each time a new 'aspect' is introduced in here is worrying. We should probably revisit this in the future.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -218,6 +235,8 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
       } else {
         false.pure[F]
       }
+    def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]] =
+      blockHashesByDeploy.get(deployId).pure[F]
     def topoSort(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
```
ArturGajowy, line 242: nit: please always leave whitespace between methods. It's hard to read them and scan for them otherwise.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+                                    readRec(
+                                      (ByteString.copyFrom(deployIdBytes) -> ByteString
+                                        .copyFrom(blockHashBytes)) :: result
+                                    )
+                                  case _ =>
+                                    Sync[F].raiseError[List[(DeployId, BlockHash)]](
+                                      BlockHashesByDeployLogIsCorrupted
+                                    )
+                                }
+                     } yield result
```
ArturGajowy: Looks as if this whole method (and all its siblings) should be replaced by scodec, or a similar tool (In a separate JIRA/PR ofc). WDYT?

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+      result <- Monad[F].ifM(fullCalculatedCrc.value.map(_ == readBlockHashesByDeployCrc))(
+                 (blockHashesByDeploy, fullCalculatedCrc).pure[F],
+                 blockHashesByDeploy match {
+                   case Nil =>
+                     Sync[F].raiseError[(List[(DeployId, BlockHash)], Crc32[F])](
+                       BlockHashesByDeployLogIsCorrupted
+                     )
+                   case _ :: _ =>
+                     // Trying to delete the last log entry
+                     for {
```
ArturGajowy: is that method deleting anything? If so, why is it called 'validate'?

Or maybe we allow some inconsistency for the last log entry? If so, why?

ArturGajowy (commented):
I can offer some comments, but I won't be able to offer a thorough review this time. Please find another reviewer, I'm snowed under :(

models/src/test/scala/coop/rchain/models/blockImplicits.scala
itegulov: > we should add any implicits needed for generating random BlockMessages in the module BlockMessage belongs - casper.

Moving the implicits for generating random BlockMessages to casper module would mean that we won't be able to use them from blockstorage module (or we will have to make it depend on casper which we can't do because it will be a cyclic dependency).

> Stick to totally random data maybe?

You are right, using random data is probably better than copy-pasting code.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
itegulov: Agreed. I heavily underestimated the amount of indices we will eventually have when I was initially writing the storage...

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
itegulov: The node could have crashed while writing the last log entry. If it did, then the node didn't update the CRC value located in a separate file. Which means we can delete the last entry and check if it matches the written CRC value.

itegulov:
@ArturGajowy thanks for your review. I needed your opinion mainly on the last two commits to check if we still agree on deleting the API for finding blocks by deploy creator and timestamp as well as other changes around that part of code.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
ArturGajowy: what happens if I fail here, and then restart the node again?

Also, this looks like a generic concern for the indexing store - all logs should be the same in that matter, and it should be handled by the same code :)

ArturGajowy:
> I needed your opinion mainly on the last two commits to check if we still agree on deleting the API for finding blocks by deploy creator and timestamp as well as other changes around that part of code.

Those look good, maybe just the below commit could be split in two :)

```
Make BlockAPI use blockstorage deploy index  …
Also make all casper code to use `ByteString` as `DeployId` instead
of `Array[Byte]`.
```

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
itegulov: Nothing will really change. The last entry garbage which was there previously will potentially be overwritten by some new garbage. After restarting the node will realise that the CRC does not match and will try to delete the last entry once again.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -357,12 +358,10 @@ object BlockAPI {
       _.withCasper[ApiErr[LightBlockQueryResponse]](
         implicit casper =>
           for {
-            dag               <- casper.blockDag
```
KentShikama, line 360: I know you didn't write this method name but we should probably call this function `findByDeploy`

KentShikama (approved):

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+    def readRec(
+        result: List[(DeployId, BlockHash)]
+    ): F[List[(DeployId, BlockHash)]] =
+      for {
+        deployIdSizeOpt <- randomAccessIO.readInt
+        result <- deployIdSizeOpt match {
+                   case Some(deployIdSize) =>
+                     val blockHashBytes = Array.ofDim[Byte](BlockHash.Length)
+                     val deployIdBytes  = Array.ofDim[Byte](deployIdSize)
+                     for {
```
odeac: This can be replaced by:
```
randomAccessIO.readFully(deployIdBytes)
.product(randomAccessIO.readFully(blockHashBytes)))
.map{case (readDeployId, readBlockHash) => ...}
.getOrElse(Sync[F].raiseError(...))
```

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+                                  case _ =>
+                                    Sync[F].raiseError[List[(DeployId, BlockHash)]](
+                                      BlockHashesByDeployLogIsCorrupted
+                                    )
+                                }
+                     } yield result
+                   case None =>
+                     result.reverse.pure[F]
+                 }
+      } yield result
```
odeac, line 1098: you have too many "result" names and it becomes confusing.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
             }
     } yield LMDBStore[F](env, dbi)
 
+  private def readBlockHashesByDeploy[F[_]: Sync](
+      randomAccessIO: RandomAccessIO[F]
+  ): F[List[(DeployId, BlockHash)]] = {
+    def readRec(
+        result: List[(DeployId, BlockHash)]
+    ): F[List[(DeployId, BlockHash)]] =
+      for {
```
odeac, line 1093: It looks like the outer for can be implemented in terms of traverse

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+      randomAccessIO: RandomAccessIO[F],
+      readBlockHashesByDeployCrc: Long,
+      blockHashesByDeploy: List[(DeployId, BlockHash)]
+  ): F[(List[(DeployId, BlockHash)], Crc32[F])] =
+    for {
+      fullCalculatedCrc <- calculateBlockHashesByDeployCrc[F](blockHashesByDeploy)
+      result <- Monad[F].ifM(fullCalculatedCrc.value.map(_ == readBlockHashesByDeployCrc))(
+                 (blockHashesByDeploy, fullCalculatedCrc).pure[F],
+                 blockHashesByDeploy match {
+                   case Nil =>
```
odeac: It reads better to have ```if (blockHashesByDeploy.isEmpty) ... else ...``` instead of ```case blockHashesByDeploy match Nil => ... case _ : _ => ..```

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
```diff
@@ -1034,6 +1082,121 @@ object BlockDagFileStorage {
...
+                       result <- Monad[F].ifM(
+                                  withoutLastCalculatedCrc.value
+                                    .map(_ == readBlockHashesByDeployCrc)
+                                )(
+                                  for {
+                                    _ <- truncateBlockHashesByDeployLog(
+                                          randomAccessIO,
+                                          blockHashesByDeploy
+                                        )
+                                  } yield (blockHashesByDeploy.init, withoutLastCalculatedCrc),
```
odeac: How big is the list of blockHashesByDeploy? The init method is linear and you are calling it twice so it might be worth to cache the result

odeac (changes requested):

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
itegulov: Good point

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
itegulov: I don't think traverse is applicable here but I simplified (i.e. got rid of "for") it by using flatMap

block-storage/src/main/scala/coop/rchain/blockstorage/BlockDagFileStorage.scala
odeac: I was wrong here. My suggestion didn't improve much.

Anyway. I still have a problem with this `readBlockHashesByDeploy` function because I find it quite unreadable.

So I played a little bit with the code and I got to this:

```scala
  private def readBlockHashesByDeploy[F[_]: Sync](
      randomAccessIO: RandomAccessIO[F]
  ): F[List[(DeployId, BlockHash)]] = {
    def readDeployBytes(deployIdSize : Int) : F[(DeployId, BlockHash)] = {
      val blockHashBytes = Array.ofDim[Byte](BlockHash.Length)
      val deployIdBytes = Array.ofDim[Byte](deployIdSize)

      for {
        deployIdResult <- randomAccessIO.readFully(deployIdBytes)
        blockHashBytesResult <- randomAccessIO.readFully(blockHashBytes)
      } yield
        deployIdResult.product(blockHashBytesResult)
          .getOrElse(Sync[F].raiseError[(DeployId, BlockHash)](
            BlockHashesByDeployLogIsCorrupted
          ))
    }

    Stream.continually(randomAccessIO.readInt)
      .sequence
      .flatMap( sizes =>
        sizes.takeWhile(_.isDefined)
          .traverse(readDeployBytes(_))
      )
      .map(_.toList)
  }
```

Note: This compiles but I haven't actually tested it.

itegulov: After playing some more with the code I have arrived at this:
```scala
  private def readBlockHashesByDeploy[F[_]: Sync](
      randomAccessIO: RandomAccessIO[F]
  ): F[List[(DeployId, BlockHash)]] = {
    def readDeployBytes(deployIdSize: Int): F[(DeployId, BlockHash)] =
      for {
        deployIdResult       <- randomAccessIO.readByteString(deployIdSize)
        blockHashBytesResult <- randomAccessIO.readByteString(BlockHash.Length)
        readResult           = deployIdResult.product(blockHashBytesResult)
        result               <- readResult.liftTo[F](BlockHashesByDeployLogIsCorrupted)
      } yield result

    StreamT
      .continually(randomAccessIO.readInt)
      .takeWhile(_.isDefined)
      .mapF {
        case Some(deployIdSize) => readDeployBytes(deployIdSize)
        case None               => Sync[F].raiseError[(DeployId, BlockHash)](new IllegalStateException())
      }
      .toList
  }
```

I have implemented `RandomAccessIO.readByteString` to simplify the logic here and also had to replace `Stream` with `StreamT` since calling `Stream.continually(randomAccessIO.readInt).sequence` hangs the runtime (seems like cats eagerly evaluates infinite stream while sequencing for some reason). Hence I also had to extend `StreamT` with `StreamT.continually` and `StreamT.mapF` - check them out.

odeac (approved):

PR #2637, opened 2019-08-02
No spatial matcher costs

ArturGajowy:
## Overview

As discussed on last standup. Please see the jira ticket for more details.

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3722

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
@@ -45,24 +45,20 @@ object SpatialMatcher extends SpatialMatcherInstances {
   //Also, the following workaround is needed b/c of precisely this reason:
   type Alternative[F[_]] = Alternative_[F]
 
-  def spatialMatchAndCharge[M[_]: Sync: _error](target: Par, pattern: Par)(
-      implicit
-      cost: _cost[M]
+  def spatialMatchResult[M[_]: Sync: _error](
+      target: Par,
+      pattern: Par
   ): M[Option[(FreeMap, Unit)]] = {
     type R[A] = MatcherMonadT[M, A]
 
-    implicit val _                            = matcherMonadCostLog[M]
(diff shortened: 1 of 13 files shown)


lukasz-golebiewski (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
```diff
@@ -450,43 +446,33 @@ trait SpatialMatcherInstances {
...
-      else
-        for {
-          _ <- listMatchSingle(target.binds, pattern.binds)
-          _ <- spatialMatch(target.body, pattern.body)
-        } yield ()
+      for {
+        _ <- Alternative_[F].guard(target.persistent == pattern.persistent)
+        _ <- listMatchSingle(target.binds, pattern.binds)
+        _ <- spatialMatch(target.body, pattern.body)
+      } yield ()
```
KentShikama, line 467: So `listMatchSingle` doesn't get executed if `target.persistent != pattern.persistent`?

KentShikama (approved):
I find the Alternative_[F] to be less readable without knowing what Alternative_[F] does. But maybe this pattern is common enough to be worth being less readable.

rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/SpatialMatcher.scala
ArturGajowy: Yes. Alternative.guard(false) == MonoidK.empty, and that short-circuits a monadic bind (flatMap).

(I think the laws-based explanation would be sth. like "whatever ap empty == empty due to Alternative laws, and because Monad must be consistent with Applicative (which Alternative is), empty flatMap whatever == empty" - but can't tell for sure)

Notice Altierntive_ is just a wrapper over cats.Alternative and the `guard` method comes from there. (In general whenever we have Foo_, this is just a wrapper over cats.Foo, just to avoid 'ambiguous implicit' errors. Other examples: ApplicativeError_, MonadError_).

PR #2638, opened 2019-08-02
Extract static 'template' methods for RevVault's instance methods

ArturGajowy:
As discussed during the ongoing standup call:

This should reduce the collective size of persistent contracts we store
in the tuplespace per a RevVault instance.

## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -21,7 +21,9 @@ new
   _makeVault,
   _newVault,
   _getOrCreate,
-  _revVault
+  _revVault,
+  _transferTemplate,
+  _depositTemplate
 in {
   rl!(`rho:rchain:makeMint`, *MakeMintCh) |
   rl!(`rho:rchain:authKey`, *AuthKeyCh) |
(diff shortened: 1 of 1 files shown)


ArturGajowy:
@pyrocto I was unable to assign you as a reviewer, will work on that with guys - please share any comments you might have in the meanwhile :)

JosephDenman (commented):
I don't think this will actually reduce the number of COMM events in genesis. Remember that the only consumes that are inserted in the tuplespace are the top-level methods for the vault. More specifically, 
```
contract revVault(@"balance", ret)
```
```
contract revVault(@"transfer", @revAddress, @amount, authKey, ret)
```
and
```
contract @{ownRevAddress | bundle0{*_revVault}}(@"_deposit", depositPurse, retCh)
```

None of the consumes inside the continuations of these consumes (which are the consumes you are reducing) are executed until they are triggered by a corresponding send, and there are no such sends in the genesis block.

ArturGajowy:
> I don't think this will actually reduce the number of COMM events in genesis

I never said it would. It's not the point of this change. It reduces the size of persistent (!) continuations that land in the hotstore and (then) in LMDB.

So, your point is a good argument in favor of this change: it doesn't change the number of COMM events in genesis, and it should reduce the size of hot store (and tuplespace bloat in general).

ArturGajowy:
IOW this is targeting the OOME error @rabbitonweb described today.

JosephDenman:
We are only discussing OOME's because of the genesis block.

> I never said it would. It's not the point of this change. It reduces the size of persistent (!) continuations that land in the hotstore and (then) in LMDB.

The change does not reduce the number of consumes that land in the hot-store during the processing of the genesis block. While it will reduce the number of persistent consumes in non-genesis blocks, the trade-off is that all `"transfer"` and `"deplosit"` are now funneled through `_depositTemplate` and `_transferTemplate`. That means the continuations at `_depositTemplate` and `_transferTemplate` will be _initialized_ in message-arrival-order (though still concurrently). It's a bottle-neck.

ArturGajowy:
> We are only discussing OOME's because of the genesis block.

Yes. I thought that's an issue we want to solve, and the contracts persisted for the vault instances contribute to the hotstore size, which in turn was determined to be at least part of the cause of the OOME?

> While it will reduce the number of persistent consumes in non-genesis blocks

It won't, each vault instance still has its persistent contracts - just shorter, more shallow, and thus taking up less space.

> That means the continuations at _depositTemplate and _transferTemplate will be initialized in message-arrival-order (though still concurrently). It's a bottle-neck.

WDYM by 'continuations being initialized'? the `_*Template` contracts are going to be stored just **once**, during the genesis block. They are going to be matched, and their continuations are going to be evaluated, each time the instance methods will be called. Their evaluation will proceed concurrently.

> It's a bottle-neck.

I wouldn't be so conclusive about that, for sure not until we have benchmarks.

If you're worried by a lock being taken to send to the persistent continuation on the same channel - each of the template continuations makes many other reads and writes to TS, and the lock is only needed when reading the continuation after the match (at the beginning of contract call). Those other sends/receives, and rholang evaluation in the continuations' bodies are going to dwarf any contention that's going to happen on the template contracts - especially that we're executing using hotstore.

We could store the continuations on a channel and peek them, and then drop them - that would though mean a more complex refactoring (one we've not tried so far), and would only be beneficial if we use a reader-lock for peeks (do we?)

So I say let's try if this improves things, and improve from there.

JosephDenman:
> WDYM by 'continuations being initialized'? the _*Template contracts are going to be stored just once, during the genesis block. They are going to be matched, and their continuations are going to be evaluated, each time the instance methods will be called. Their evaluation will proceed concurrently.

I mean that, when multiple messages are received on a channel where a persistent continuation exists, the messages are picked off in arrival order. The contents of the message are bound in the continuation, and that _continuation_ is executed in parallel with another _consume_. Hence, bottleneck.

ArturGajowy:
I'm having a hard time making sense of what you're saying:

> the messages are picked off in arrival order.

You mean 'the produce lock forms a FIFO queue'? How's that relevant?

`continuation is executed in parallel with another consume. Hence, bottleneck.`

How is parallel execution a bottleneck!?

Once again: a lock, or a sync point is not always an **actual** bottleneck - just a potential one. It might very well be it's never going to have enough contention to become an actual one.

BTW: I've just checked and we don't do a reader lock for peeks.
So the tradeoff is between a potential contention point (be it with peek or with a contract) or storing copious amounts of instance contract bodies.

ArturGajowy:
Incorporated in #2640

PR #2640, opened 2019-08-05
User vaults at genesis

ArturGajowy:
## Overview

This allows us to actually attempt a genesis ceremony with vaults based on a wallets.txt file. It also contains some refactorings in rholang code allowing for the genesis vaults to be created without validation logic, and for it to run faster.

Running:

`%rnode java invocation% run -s --wallets-file wallets.txt`

**Make sure to `rm -rf ~/.rnode` before each run**, or else you won't be running as if in an actual genesis ceremony.

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -108,7 +108,7 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
         ((false, _), _)               => return!(a)
         (_, (false, _))               => return!(b)
         ((true, va), (true, vb)) => {
-          new ret, stdout(`rho:io:stdout`) in {
+          new ret in {
             match *f {
               { for (x, y, r <- fName) { _ } } => {
                 *f |
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -108,10 +108,15 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
         ((false, _), _)               => return!(a)
         (_, (false, _))               => return!(b)
         ((true, va), (true, vb)) => {
-          new ret in {
-            f!(va, vb, *ret) |
-            for (@v <- ret) {
-              return!((true, v))
+          new ret, stdout(`rho:io:stdout`) in {
+            match *f {
+              { for (x, y, r <- fName) { _ } } => {
+                *f |
+                fName!(va, vb, *ret) |
+                for (@v <- ret) {
+                  return!((true, v))
+                }
+              }
             }
           }
         }
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
@@ -21,23 +21,12 @@ final case class RevGenerator(
     s""" new rl(`rho:registry:lookup`), revVaultCh in {
        #   rl!(`rho:rchain:revVault`, *revVaultCh) |
        #   for (@(_, RevVault) <- revVaultCh) {
-       #
-       #     new genesisVaultCh in {
-       #       @RevVault!(
-       #         "findOrCreateGenesisVault",
-       #         "${genesisAddress.toBase58}",
-       #         $supply,
-       #         *genesisVaultCh
-       #       )
-       #     } |
-       #
        #     new ret in {
        #       @RevVault!("init", *ret) |
        #       for (vaultMapStore, initVault <- ret) {
        #         ${vaultInitCode()}
        #       }
        #     }
-       #
        #   }
        # }
      """.stripMargin('#')
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
@@ -19,8 +19,9 @@ final case class RevGenerator(
 
   val code: String =
     s""" new rl(`rho:registry:lookup`), revVaultCh in {
-       #   rl!(`rho:rchain:revVault`, *revVaultCh)
-       #   | for (@(_, RevVault) <- revVaultCh) {
+       #   rl!(`rho:rchain:revVault`, *revVaultCh) |
+       #   for (@(_, RevVault) <- revVaultCh) {
+       #
        #     new genesisVaultCh in {
        #       @RevVault!(
        #         "findOrCreateGenesisVault",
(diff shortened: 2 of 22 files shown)


ArturGajowy:
For 800 vaults, this [runs in ~4:30](https://gist.github.com/ArturGajowy/695825babce0679cfe804a0d45517632) on my laptop, which includes a replay of the original vaults execution.

[Current dev took ~8:00 when running for **400** vaults](https://gist.github.com/ArturGajowy/ee06c12c68965cc13d3c9c13111e3c5c)

It so far hasn't finished its run for 800 vaults.
It reached the end of the first (non-replay) execution (`Finished execution of ApprovedBlockProtocol`), in ~13:30 which took the code in this PR just ~3:00 mins.
EDIT: [800 vaults took current dev ~24:30](https://gist.github.com/ArturGajowy/e03657786131b92f0a9ee7212ca41a95)

EDIT: [this PR - 1600 vaults - ~10:30](https://gist.github.com/ArturGajowy/7282ed14a4c2c1767302d5d77580658f)

EDIT: The last commit (`Extract static 'template' methods`) could probably be dropped, as it doesn't improve (or degrade) performance as measured on [no templates - 800 vaults - ~4:30](https://gist.github.com/ArturGajowy/4c8769a954f7816cb3bde9dc67ff87e9)

I've also done a ['inline all vault instance methods' experiment](https://github.com/ArturGajowy/rchain/commit/4ac290955b7bc289da4f087233bed666f548a3f1) similar to what @JosDenmark has in his branch, and it degrades the performance: [inline all - 800 vaults - ~5:30](https://gist.github.com/ArturGajowy/21886537936088f30eaccd203519e8b8)

ArturGajowy:
> this runs in ~4:30 on my laptop

I mean the time from start to `Requested fork tip from peers`

JosephDenman (changes requested):
The last commit should not be included in this PR. You have already opened a PR for this change and the discussion should continue there: <https://github.com/rchain/rchain/pull/2638>.

ArturGajowy:
[Benchmark results summary](https://docs.google.com/spreadsheets/d/11WdCneacNV4yWD14TXz2WKL-PxGYQbFGiAiXPNF0V-8/edit?usp=sharing)

JosephDenman:
TLDR; which commit are you saying yields performance benefits? <https://github.com/rchain/rchain/pull/2640/commits/ebeedf4623a88211938ec41d52fd7f860ad18121> ?

ArturGajowy:
First 3 commits are refactorings
ebeedf4 brings the most performance improvement

the last commit - 6f40bbe - improves memory usage, tuplespace size, and seems to improve performance. As you can see in the spreadsheet, the last commit resulted in: +4%, +9%, -3% vaults/s changes in 3 various tests.

I included the commit in the benchmarks, because I believe it is an improvement. Benchmarks so far don't deny that.

JosephDenman:
Have you verified empirically that, for a randomly chosen vault, the `"balance"` and `"transfer"` methods complete successfully? Something like this <https://github.com/rchain/rchain/pull/2622/files#diff-5d1a606e74c23e98205ad0443160cde9> perhaps.

EncodePanda:
@JosDenmark haven't we just discussed that a 3 min ago?

JosephDenman:
@rabbitonweb We discussed benchmarking transfers and balance checks, but a test that transfers and balance checks work should be included in this PR.

ArturGajowy:
> a test that transfers and balance checks work should be included in this PR.

It's already in the codebase, in RevVaultSpec

EncodePanda:
@JosDenmark given that there were no benchmarks in the past and you think this change will slow things down, I truly believe you should step up and add those benchmarks to this PR if it really concerns you. I guess that's what I would do.

JosephDenman:
@ArturGajowy @rabbitonweb Maybe I'm not being clear. I'm suggesting tests for `"transfer"` and `"balance"` should be included to make sure that the generated code works correctly. On Friday, Artur found a bug on my branch in the generated code that wasn't apparent in the performance tests <https://github.com/rchain/rchain/blob/testing-RevVault-methods/casper/src/main/scala/coop/rchain/casper/genesis/contracts/VaultGenerator.scala#L29>.Testing `"balance"` and `"transfer"` would have found said bug. 

Then we can do benchmarks with certainty 👌.

ArturGajowy:
@JosDenmark I've added tests in `Add tests ensuring genesis-created vaults are intact`

ArturGajowy:
bors try

ArturGajowy:
bors try

ArturGajowy:
bors try

tymm (approved):

ArturGajowy:
bors try

casper/src/test/scala/coop/rchain/casper/engine/BlockApproverProtocolTest.scala
```diff
@@ -87,6 +87,7 @@ object BlockApproverProtocolTest {
       (new BlockApproverProtocol(
         node.validatorId,
         genesisParams.timestamp,
+        Seq(),
```
JosephDenman: It seems like validator vaults should be passed explicitly here. Previously, they were constructed in the call

```
vaults = Traverse[List]	
          .traverse(posParams.validators.map(_.pk).toList)(RevAddress.fromPublicKey)	
          .get	
          .map(Vault(_, 0L))
```

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -62,6 +62,7 @@ object Setup {
     val bap = new BlockApproverProtocol(
       validatorId,
       deployTimestamp,
+      Seq(),
```
JosephDenman: Here, also, it looks like vaults for validators need to be passed explicitly.

casper/src/main/resources/Either.rho
```diff
@@ -108,7 +108,7 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
         ((false, _), _)               => return!(a)
         (_, (false, _))               => return!(b)
         ((true, va), (true, vb)) => {
-          new ret, stdout(`rho:io:stdout`) in {
```
JosephDenman: Nit: this reversion should've been squashed into your second commit.

casper/src/main/resources/Either.rho
```diff
@@ -108,10 +108,15 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
         ((false, _), _)               => return!(a)
         (_, (false, _))               => return!(b)
         ((true, va), (true, vb)) => {
-          new ret in {
-            f!(va, vb, *ret) |
-            for (@v <- ret) {
-              return!((true, v))
+          new ret, stdout(`rho:io:stdout`) in {
+            match *f {
```
JosephDenman, line 112: IMO this reads better when `@f` is passed and `*f` occurrences are `f` occurrences.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -19,57 +19,63 @@ final case class RevGenerator(
 
   val code: String =
     s""" new rl(`rho:registry:lookup`), revVaultCh in {
-       #   rl!(`rho:rchain:revVault`, *revVaultCh)
-       #   | for (@(_, RevVault) <- revVaultCh) {
+       #   rl!(`rho:rchain:revVault`, *revVaultCh) |
+       #   for (@(_, RevVault) <- revVaultCh) {
+       #
        #     new genesisVaultCh in {
        #       @RevVault!(
```
JosephDenman: Now that vault construction is inlined, we can eliminate the genesis vault. Please remove it.

JosephDenman (changes requested):
LGTM generally. I do have a couple of points. 

First, please remove the genesis vault and associated methods. It was only ever necessary to instantiate the other genesis vaults and would make a particularly ugly artifact. 

Second, please be sure to double check that the changes in your first commit, in particular to the ones in `BlockApproverProtocolTest` and `Setup` where you pass an empty sequence of genesis vaults, are not failing to construct the validator vaults that were constructed before your change. Failing to produce validator vaults could produce a bug in future tests that rely on `BlockApproverProtocolTest` and `Setup` and assume that those vaults are constructed.

Third, do you have an intuition for why the way you inlined vault creation performs better than the way I inlined vault creation? The only fundamental difference I've identified is that your constructions are funneled through `initVault`. It could be that the reduction in parallelism caused by funneling all of the constructions through `initVault` actually improves performance.

If you need support on these tasks, just let me know.

casper/src/test/scala/coop/rchain/casper/engine/BlockApproverProtocolTest.scala
ArturGajowy: Id skip that part of setup if the tests dont rely on that, but have no big preference

casper/src/main/resources/Either.rho
ArturGajowy: I tend to pass non-value stuff without the @ pattern. Otoh the receive is used as a value here, so maybe it makes sense? The * reminds that this is being evaluated, at least to me. No big preference : )

ArturGajowy:
Re intuition on performance: this version needa to process (parse, interpret, hash, serialize and store) much less rholang terms. The funnel-related contention is shadowed by its surroundings.

ArturGajowy:
I think we also should compensate for the validation we've just removed:
each vault created at genesis must have positive balance, and its revadress must be valid. So we should have a .scala tool to validate a wallets.txt before using it for genesis.

We also should point out during next standup, that there no longer is a single origin of all rev - there is 16k origins. CC @cboscolo

casper/src/main/resources/Either.rho
```diff
@@ -108,10 +108,15 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
...
+              { for (x, y, r <- fName) { _ } } => {
+                *f |
+                fName!(va, vb, *ret) |
+                for (@v <- ret) {
+                  return!((true, v))
+                }
+              }
```
KentShikama, line 119: Doesn't this change the behavior of map2 whenever you're not passing in a `for`?

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
```diff
@@ -16,6 +16,9 @@ object RevAddress {
 
   private val tools = new AddressTools(prefix, keyLength = Validator.Length, checksumLength = 4)
 
+  def fromDeployerId(deployerId: Array[Byte]): Option[RevAddress] =
```
KentShikama, line 19: Shouldn't we call this "fromPublicKey"?

KentShikama (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/util/RevAddress.scala
ArturGajowy: No, there's a `fromPublicKey` just next to it

casper/src/main/resources/Either.rho
ArturGajowy: yes, it does. There's only one usage though.

JosephDenman:
bors try

JosephDenman:
@tymm @KentShikama would you take a second look?

JosephDenman:
@ArturGajowy Still working on getting an integration test to pass.

ArturGajowy:
Interesting. @JosDenmark can you :heavy_check_mark: this? :)

JosephDenman (approved):
Great work Joe and Artur 🎉

PR #2646, opened 2019-08-08
Bump version to v0.9.12

woky:
This PR contains also previous bump to v0.9.11 which wasn't merged yet. I've
closed the previous PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -242,7 +242,7 @@ lazy val node = (project in file("node"))
   .settings(commonSettings: _*)
   .enablePlugins(RpmPlugin, DebianPlugin, JavaAppPackaging, BuildInfoPlugin)
   .settings(
-    version := "0.9.11" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
+    version := "0.9.12" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
     name := "rnode",
     maintainer := "RChain Cooperative https://www.rchain.coop/",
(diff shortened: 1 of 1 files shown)


kayvank (approved):

kayvank:
bors try

lukasz-golebiewski (approved):

build.sbt
```diff
@@ -242,7 +242,7 @@ lazy val node = (project in file("node"))
   .settings(commonSettings: _*)
   .enablePlugins(RpmPlugin, DebianPlugin, JavaAppPackaging, BuildInfoPlugin)
   .settings(
-    version := "0.9.10" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
+    version := "0.9.12" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
```
lukasz-golebiewski, line 245: This information should be taken from a git tag. Otherwise in the time between this PR being submitted and it making it into dev some features which should not make it into the release will be included

woky: @lukasz-golebiewski You're right, but this won't work when you distribute sources as tarball. For official packages in some Linux distributions, source tarballs are required, for example for RedHat/Fedora (https://fedoraproject.org/wiki/Packaging:SourceURL#Git_Hosting_Services). Generally, when building specific release, source tarballs are preferred over Git clone (speaking from inspecting ArchLinux AUR `PKGBUILD`s). I'd be happy to switch to Git tags, but this should be considered.

PR #2660, opened 2019-08-12
Refactor casper to use BlockMetadata in BFS traverses and EstimatorHelper

itegulov:
## Overview 
Title says it all

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3748

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -14,6 +14,7 @@ import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.{PrivateKey, PublicKey}
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.models.BlockMetadata
 import coop.rchain.models.Validator.Validator
 import coop.rchain.rholang.interpreter.Runtime.BlockData
 import coop.rchain.rholang.interpreter.accounting
@@ -146,17 +147,20 @@ object BlockCreator {
       validDeploys = deploys.filter(
         d => notFutureDeploy(currentBlockNumber, d) && notExpiredDeploy(earliestBlockNumber, d)
       )
-      deploysInCurrentChain <- DagOperations
-                                .bfTraverseF[F, BlockMessage](parents.toList)(
-                                  b =>
-                                    ProtoUtil
-                                      .unsafeGetParentsAboveBlockNumber[F](b, earliestBlockNumber)
-                                )
-                                .map { b =>
-                                  ProtoUtil.deploys(b).flatMap(_.deploy)
-                                }
-                                .toList
-    } yield (validDeploys -- deploysInCurrentChain.flatten).toSeq
+      parentsMetadata = parents.map(p => BlockMetadata.fromBlock(p, invalid = false))
+      result <- DagOperations
+                 .bfTraverseF[F, BlockMetadata](parentsMetadata.toList)(
+                   b =>
+                     ProtoUtil
+                       .getParentMetadatasAboveBlockNumber[F](b, earliestBlockNumber, dag)
+                 )
+                 .foldLeftF(validDeploys) { (deploys, blockMetadata) =>
+                   for {
+                     block        <- ProtoUtil.unsafeGetBlock[F](blockMetadata.blockHash)
+                     blockDeploys = ProtoUtil.deploys(block).flatMap(_.deploy)
+                   } yield deploys -- blockDeploys
+                 }
+    } yield result.toSeq
 
   private def notExpiredDeploy(earliestBlockNumber: Long, d: DeployData): Boolean =
     d.validAfterBlockNumber > earliestBlockNumber
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -54,8 +54,8 @@ object BlockCreator {
       for {
         tipHashes             <- Estimator.tips[F](dag, genesis)
         _                     <- spanF.mark("after-estimator")
-        parents               <- EstimatorHelper.chooseNonConflicting[F](tipHashes, dag)
-        maxBlockNumber        = ProtoUtil.maxBlockNumber(parents)
+        parentMetadatas       <- EstimatorHelper.chooseNonConflicting[F](tipHashes, dag)
+        maxBlockNumber        = ProtoUtil.maxBlockNumberMetadata(parentMetadatas)
         invalidLatestMessages <- ProtoUtil.invalidLatestMessages[F](dag)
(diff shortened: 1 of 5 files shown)


KentShikama:
```scala
[error] /drone/src/github.com/rchain/step.run-unit-tests/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala:103:11: [wartremover:Throw] throw is disabled
[error]           throw new Exception(s"DAG storage is missing hash ${PrettyPrinter.buildString(hash)}")
[error]           ^
[error] one error found
[info] Done compiling.
```

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -21,14 +21,14 @@ object EstimatorHelper {
...
       dag: BlockDagRepresentation[F]
-  ): F[Seq[BlockMessage]] = {
-    def nonConflicting(b: BlockMessage): BlockMessage => F[Boolean] =
+  ): F[Seq[BlockMetadata]] = {
+    def nonConflicting(b: BlockMetadata): BlockMetadata => F[Boolean] =
       conflicts[F](_, b, dag).map(b => !b)
 
     for {
-      blocks <- blockHashes.toList.traverse(hash => ProtoUtil.unsafeGetBlock[F](hash))
+      blocks <- blockHashes.toList.traverse(hash => ProtoUtil.unsafeGetBlockMetadata[F](hash, dag))
```
ArturGajowy: why are ` ProtoUtil.unsafeGetBlockMetadata][F]` and ` ProtoUtil.unsafeGetBlock[F]` unsafe?

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -146,17 +148,19 @@ object BlockCreator {
...
-    } yield (validDeploys -- deploysInCurrentChain.flatten).toSeq
+      result <- DagOperations
+                 .bfTraverseF[F, BlockMetadata](parents.toList)(
+                   b =>
+                     ProtoUtil
+                       .unsafeGetParentMetadatasAboveBlockNumber[F](b, earliestBlockNumber, dag)
+                 )
+                 .foldLeftF(validDeploys) { (deploys, blockMetadata) =>
+                   for {
+                     block        <- ProtoUtil.unsafeGetBlock[F](blockMetadata.blockHash)
```
ArturGajowy, line 159: I was going to say that maybe we're actually avoiding some `getBlock`s in here - but no, it seems every BlockMedatada usage added is followed by BlockMessage fetching.

Not saying that it's wrong - it's a change for the better (we restrict the scope where BlockMessage seems to be needed). It's just that it won't bring a performance imrpovement.

FYI @sebtomba

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -91,6 +91,19 @@ object ProtoUtil {
...
+  def unsafeGetBlockMetadata[F[_]: Monad](
+      hash: BlockHash,
+      dag: BlockDagRepresentation[F]
+  ): F[BlockMetadata] =
+    for {
+      maybeBlockMetadata <- dag.lookup(hash)
+      blockMetadata = maybeBlockMetadata match {
+        case Some(b) => b
+        case None =>
+          throw new Exception(s"DAG storage is missing hash ${PrettyPrinter.buildString(hash)}")
```
ArturGajowy: throw is a no-go. Please replace it with `.raise`. I'm fine if you use a `:Sync` constraint in here - still a better love story than throws. Ideally it'd be `FunctorRaise[CasperError` though

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
itegulov: > it seems every BlockMedatada usage added is followed by BlockMessage fetching

This is not entirely true. Previous implementation could go to BlockStore for fetching a block which was an ancestor to several initial blocks (those in `parents.toList`) multiple times while the new implementation will do this only once. LRU cache would help with this problem but since adding it is controversial (CC @sebtomba), this is an optimisation.

> It's just that it won't bring a performance imrpovement.

Performance improvement is not the only goal here. We should restrict the amount of BlockMessages simultaneously loaded into memory since they can be of unlimited size. This concern was raised by @sebtomba last Friday.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
itegulov: They are unsafe as in "will crash if the requested block is not actually in the block store/dag storage".

ArturGajowy: then they should be just `getBlock` or `tryGetBlock` if we need anything more to tell that than just the type. `unsafe` should be reserved for dirty stuff like impure effects underneath, or throwing exceptions which those methods should not be doing.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
sebtomba: @ArturGajowy @itegulov Yes, we are killing here two birds with one stone. Very nice!

However, I would like to go even one step further and optimize `DagOperations.bfTraverseF`. The current signature is
```scala
def bfTraverseF[F[_]: Monad, A](start: List[A])(neighbours: A => F[List[A]]): StreamT[F, A] 
```
I would like to pass the already visited neighbors hash set to the function which calculates next neighbors
```scala
def bfTraverseF[F[_]: Monad, A](start: List[A])(neighbours: (A, HashSet[A]) => F[List[A]]): StreamT[F, A]
```

In the current implementation `DagOperations.bfTraverseF` filters the neighbors **after** they have been fetched (again) from the store:
```scala
for {
  ns      <- neighbours(curr)
  visited = prevVisited + curr
  newQ    = rest.enqueue[A](ns.filterNot(visited))
} yield StreamT.cons(curr, Eval.always(build(newQ, visited)))
```

We can postpone this change for later because it requires many modifications.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
itegulov: I renamed `unsafeGetBlockMetadata` added in this PR to `getBlockMetadata` and changed its signature from `F[_]: Monad` to `F[_]: Sync`. Doing the same with `unsafeGetBlock` (and other "unsafe" methods in ProtoUtil) would make this relatively focused PR to touch the entire casper code base.

ArturGajowy: Can you do that in a follow-up PR then?

itegulov: Sure

sebtomba (approved):

KentShikama (approved):

PR #2661, opened 2019-08-12
[RCHAIN-3751] build.sbt: Get version from git describe

woky:
Build will fail when run outside of Git. Furthermore, the Git repository
from which the build is run needs to have tags fetched.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -242,7 +242,14 @@ lazy val node = (project in file("node"))
   .settings(commonSettings: _*)
   .enablePlugins(RpmPlugin, DebianPlugin, JavaAppPackaging, BuildInfoPlugin)
   .settings(
-    version := "0.9.12" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
+    /*
+     * Components in Git describe string are separated by dashes, but RPM version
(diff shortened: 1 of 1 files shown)


woky:
Waiting for previous Bump PR to merge so that dev is on top of latest tag.

build.sbt
```diff
@@ -242,7 +242,7 @@ lazy val node = (project in file("node"))
   .settings(commonSettings: _*)
   .enablePlugins(RpmPlugin, DebianPlugin, JavaAppPackaging, BuildInfoPlugin)
   .settings(
-    version := "0.9.10" + git.gitHeadCommit.value.map(".git" + _.take(8)).getOrElse(""),
```
kayvank: This will work and I think its a very good idea.
 However, it looks as if we are changing how releases and builds are version. Should we create a JIRA ticket for this work so that its documented outside of git?

woky: @kayvank Here it is: https://rchain.atlassian.net/browse/RCHAIN-3751

ArturGajowy (approved):
What's not to like? :heavy_check_mark:

kayvank (approved):

lukasz-golebiewski (approved):
Very nice!

PR #2664, opened 2019-08-13
Add asynchronous map.

ghost:
## Overview
Artur suggested that a map with linear performance rather than quadratic would speed up the genesis vault creation.  These are two implementations of such a map.  Adding to the map works by creating cell contracts at each node along a path in a binary tree and storing whether the left and right child exist.  Adding an element to a map containing n items should only change around log(n) existing cells.

The PR has two versions, one using a pattern that "cleans up after itself":
```
new loop in {
  match {
    for (<args>, @P <- loop) {
      ...
      // when recursing
      P | loop!(<new args>, P)
    }
  } {
    // base case
    P => { P | loop!(<initial args>, P) }
  }
}
```
the other using a pattern where the loop is extracted into a helper contract.

The latter seems to be about five times faster, presumably because the linear consume in the former is not removed and added each time through the loop.

I'm working on modifying the latter to become a 16-ary tree instead of a binary one; this should reduce the time by another factor of four.

### JIRA ticket:
N/A

### Notes
I can make the hash length configurable; 256 bits is overkill if we're only putting 16K items in.  64 bits is plenty unless we're expecting to put more than 4 billion items in.  That would reduce the time by another factor of at least four.

I didn't do any of the registry magic and would appreciate some help with that when the time comes.


ghost:
A more complicated data structure can put off the cost of traversing the whole hash by waiting until there are more entries in the data structure and rebalancing when necessary, but if you have an estimate for how many items will be in the map, the total cost ends up being about the same.

ghost:
None of these are threadsafe.  Need to add an update method to atomically read, set a bit, and write the value of the cell.

ghost:
Fixed in the fast version.

ghost:
The cell implementation uses a linear for, which makes querying it in two different blocks conflict.  Thinking about whether it's possible to use a contract instead.

ArturGajowy:
Can it use peek for gets? We should have merging of peek pretty soon

ArturGajowy:
Can you share the code/setup you use for benchmarking this?

casper/src/main/resources/AsyncMap4.rho
```diff
@@ -0,0 +1,213 @@
...
+
+new AsyncMap, o(`rho:io:stdout`) in {
+  new MakeCell, ByteArrayToNybbleList, AsyncMapSetter, AsyncMapGetter, powersCh in {
+    powersCh!([1,2,4,8,16,32,64,128,256,512,1024,2048,4096,8192,16384,32768, 65536]) |
+    for (@powers <- powersCh) {
+      contract MakeCell(@val, cell) = {
+        for (@cmd, @args <- cell) {
+          match cmd {
+            "get" => {
+              match args {
```
ArturGajowy: you could save one match by doing:

```
        for (@args <- cell) {
          match args {
            [@"get", ret] => {
```
I'd also use a tuple for that purpose

casper/src/main/resources/AsyncMap4.rho
```diff
@@ -0,0 +1,213 @@
...
+            x => { o!(["ERROR: MATCH FAIL. MakeCell line 45", x]) }
+          }
+        }
+      } |
+
+      contract ByteArrayToNybbleList(@ba, @n, @len, @acc, ret) = {
+        if (n == len) {
+          ret!(acc)
+        } else {
+          ByteArrayToNybbleList!(ba, n+1, len, acc ++ [ ba.nth(n) % 16, ba.nth(n) / 16 ], *ret)
```
ArturGajowy: this most likely deserves a system process

ArturGajowy:
> Artur suggested that a map with linear performance rather than quadratic would speed up the genesis vault creation.

That would be true prior to our removing of modifying the vaultsMap in a loop during genesis. Currently only one modification is done.

Efficiency of updates is still an issue though for subsequent updates to the vaultsMap. With the current 'grab the huge 16k-elements map term' approach, even lookup is terribly slow and costly (mostly for the validator...).

casper/src/main/resources/AsyncMap4.rho
```diff
@@ -0,0 +1,213 @@
...
+
+      contract AsyncMap(@"get", @map, @key, ret) = {
+        new hashCh, nybListCh in {
+          // Hash the key to get a 256-bit array
+          @"keccak256Hash"!(key.toByteArray(), *hashCh) |
+          for (@hash <- hashCh) {
+            // Get the bit list
+            ByteArrayToNybbleList!(hash, 0, hash.length(), [], *nybListCh) |
+            for (@nybList <- nybListCh) {
+              AsyncMapGetter!(map, nybList.slice(0, 16), 0, 16, *ret)
```
ArturGajowy: Seems like we're using just (16 nybbles * 4 bits/nybble) = 64 bits of the hash. This is OK for me, but we should probably slice the hash before converting to nybbles, or we waste cycles (and COMMs!)

casper/src/main/resources/AsyncMap4.rho
ghost: MakeCell rewritten entirely.

casper/src/main/resources/AsyncMap4.rho
ghost: I agree.  The bitwise operators also deserve system processes.

casper/src/main/resources/AsyncMap4.rho
ghost: done

ghost:
The new implementation of MakeCell (once peek is available) will resolve the conflict issues.

ghost:
For benchmarking, I just ran `time java -Dlogback.configurationFile=logback-rholang_cli.xml -jar rholang-cli/target/scala-2.12/rholangCLI-assembly-0.1.0-SNAPSHOT.jar -qu casper/src/main/resources/AsyncMap.rho`

ghost:
In this implementation I'm trading security for speed, using only 64 bits instead of 256.  If other parties are able to specify keys (e.g. if the keys in the map are public keys chosen by users), they can arrange for a collision by trying only around 2^32 keys due to the birthday attack.  If that's part of the threat model, we need to increase the number of bits used.

One possibility is to store a rholang map at the end of the 64-bit path and use the remaining bits of the hash as keys into the rholang map.  That way we only have the possibility of 256-bit collisions.  It's still easy to cause collisions of prefixes, but at worst it slows down the map slightly; there are no significant security problems.

ArturGajowy:
> For benchmarking, I just ran time java -Dlogback.configurationFile=logback-rholang_cli.xml -jar rholang-cli/target/scala-2.12/rholangCLI-assembly-0.1.0-SNAPSHOT.jar -qu casper/src/main/resources/AsyncMap.rho

I think we need a better benchmark. A benchmark this short could be affected by Chrome deciding it's going to eat the CPU along with RAM for a change. Unless you did many repetitions of a single run and had low variance?

lukasz-golebiewski:
I'd say it's time for a first rholang `jmh` benchmark, one which would bypass the CLI. Can be done in a separate PR

ghost:
In the first version, the time to add something to the map and look it up again was on the order of three minutes, so `time` was a reasonable approach.  It's now around four seconds, which is comparable to the startup time for the CLI, so a better benchmark is called for.  I agree it should be done in a separate PR.

ghost:
Since adding a rholang map at the end of the path eliminates duplicates, we can speed it up more by using fewer bits of the hash as a prefix.  16K vaults means we need around 15 bits of hash at the start; a million vaults would only need 20 bits of hash.  24 bits is the next number of full bytes, which means we wouldn't expect any rholang map to store more than one value until around 16M entries.

ghost:
Rather than hardcode the number of bits of the hash to use, do you think it should be a parameter?

ghost:
If there are no more comments on the code, I'll remove the logging statements.  I'll need help getting the registry magic done.

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+          cellStore!(initVal) |
+          contract cell(@"get", ret) = {
+            // TODO: convert to peek
+            for (@val <- cellStore) {
+              cellStore!(val) |
+              ret!(val)
+            }
+          } |
+          
+          contract cell(@"set", @newVal, ack) = {
```
ArturGajowy: This is unused

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+// }
+
+new AsyncMap, o(`rho:io:stdout`) in {
+  new MakeCell, ByteArrayToNybbleList, AsyncMapSetter, AsyncMapGetter, powersCh in {
+    powersCh!([1,2,4,8,16,32,64,128,256,512,1024,2048,4096,8192,16384,32768,65536]) |
+    for (@powers <- powersCh) {
+      contract MakeCell(@initVal, cell) = {
+        new cellStore in {
+          cellStore!(initVal) |
+          contract cell(@"get", ret) = {
```
ArturGajowy: We should probably parameterize the cell-methods with a 'this' pointer and only create them once to avoid TS bloat.

Also, can we name it a Node? I think that's what it is :) So we'd have:

```
contract MakeNode(@val, prefix, store)
contract nodeGet(store, ret)
contract nodeSetBit(store, @bit, ack)
...
```

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+//         for (@val1 <- ret1; @val2 <- ret2; @val3 <- ret3) {
+//           // Should print ["Got these from the map: ", ["some val"], ["some other val"], []]
+//           o!(["Got these from the map: ", val1, val2, val3])
+//         }
+//       }
+//     }
+//   }
+// }
+
+new AsyncMap, o(`rho:io:stdout`) in {
```
ArturGajowy: Why AsyncMap? Everythin in rholang is async :) Maybe HashTrieMap?

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+      } |
+
+      contract AsyncMapGetter(@map, @nybList, @n, @len, @suffix, ret) = {
+        // Look up the value of the cell at [map, nybList.slice(0, n + 1)]
+        new valCh, restCh in {
+          @(map, nybList.slice(0, n))!("get", *valCh) |
+          for (@val <- valCh) {
+            if (n == len) {
+              restCh!(val.get(suffix)) |
+              for (@rest <- restCh) {
```
ArturGajowy: we could use `match` for binding a variable without incurring a COMM.

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+                if (rest == Nil) { ret!([]) }
+                else { ret!(rest) }
```
ArturGajowy: Let's use `value \/ Nil` as the return type, just like the built-in map does.

casper/src/main/resources/AsyncMap.rho
```diff
@@ -0,0 +1,208 @@
...
+        }
+      } |
+
+      contract AsyncMapGetter(@map, @nybList, @n, @len, @suffix, ret) = {
+        // Look up the value of the cell at [map, nybList.slice(0, n + 1)]
+        new valCh, restCh in {
+          @(map, nybList.slice(0, n))!("get", *valCh) |
+          for (@val <- valCh) {
+            if (n == len) {
+              restCh!(val.get(suffix)) |
```
ArturGajowy: I think we should store the entries under their keys rather than hash suffixes. This will make the map behave correctly in spite of collisions.

ArturGajowy (changes requested):
I really like how this is shaping up! Pointer-programming with rholang is strangely appealing... :astonished:

The next steps would be to register this under the registry, and add thorough tests - could be using RhoSpec, but I have another approach in mind if you find the need for it.

casper/src/main/resources/AsyncMap.rho
ArturGajowy: Actualy the if beneath could be absorbed by the match?

With 'storing entries under keys' and other comments applied we should just be able to `ret!(val.get(key))` though

casper/src/main/resources/AsyncMap.rho
ghost: removed

casper/src/main/resources/AsyncMap.rho
ghost: Done

casper/src/main/resources/AsyncMap.rho
ghost: Done

casper/src/main/resources/AsyncMap.rho
ghost: Removed comm, merged match and if.  Can't just `ret!(val.get(key))` without changing the API to return `Nil` instead of `[]`.  Rholang's built-in maps can't distinguish between storing Nil at a key and having nothing stored at the key.  I chose to make the distinction in this map implementation, but if you prefer consistency, I can use the same API as the built-in map instead.

casper/src/main/resources/AsyncMap.rho
ghost: You'd really rather not be able to tell whether someone stored Nil at the key?  If you say so...

casper/src/main/resources/AsyncMap.rho
ghost: Every key going into the same rholang map already matches on the prefix, so that wouldn't help.

ghost:
What do you have in mind for testing?  I've just been running the "Usage" example to test it.

casper/src/main/resources/AsyncMap.rho
ArturGajowy: Yeah, that's unfortunate. I pick consistency though, partly because List is confusing when used as a poor man's Option.

casper/src/main/resources/AsyncMap.rho
ArturGajowy: I'm not sure I understand? I'm not after reducing performance impact, but after not having the entry overwritten on key hash collision - which I believe happens now?

casper/src/main/resources/AsyncMap.rho
ArturGajowy: we could add a `containsKey` method if needed, couldn't we? That's the way Java and Scala handle this. Not that they're my favourites, but we're far from consistency in regard of Nil or a type system that would enforce it...

ArturGajowy:
You can have a look at https://github.com/rchain/rchain/blob/04e0f9f972ff247fb38372c46c4495157b111143/casper/src/test/scala/coop/rchain/casper/genesis/contracts/RevVaultSpec.scala#L11 and the corresponding https://github.com/rchain/rchain/blob/d1ea8b423ff7f3b048e4c8edaf1b0339c8884d88/casper/src/test/resources/RevVaultTest.rho, and other `*Test.rho` in https://github.com/rchain/rchain/blob/d1ea8b423ff7f3b048e4c8edaf1b0339c8884d88/casper/src/test/resources/

Alternatively, one could think of a Scala-level "test framework" that would run two snippets of rholang code, extract their results from an arranged output channel, and compare the results / do the assertions in Scala. Could be useful if we want to employ generative testing to model the behaviour of our map after an existing implementation (pick a method, set of parameters, execute on the model - e.g. scala.immutable.Map - execute on the tested map, compare results, repeat 1000 tiems)

casper/src/main/resources/TreeHashMap.rho
```diff
@@ -0,0 +1,220 @@
...
+            }
+          }
+        }
+      } |
+
+      contract TreeHashMap(@"get", @map, @key, ret) = {
+        new hashCh, nybListCh in {
+          // Hash the key to get a 256-bit array
+          @"keccak256Hash"!(key.toByteArray(), *hashCh) |
+          for (@hash <- hashCh; @depth <- @(map, "depth")) {
```
ArturGajowy: In the sad world we live in, every join usage is conflicting between all the channels involved. We should split that into two receives

casper/src/main/resources/TreeHashMap.rho
```diff
@@ -0,0 +1,220 @@
...
+            // Get the bit list
+            ByteArrayToNybbleList!(hash, 0, depth, [], *nybListCh) |
+            for (@nybList <- nybListCh) {
+              new restCh, valCh in {
+                nodeGet!((map, nybList), *restCh) |
+                for (@rest <- restCh) {
+                  valCh!(rest.get(hash.slice(depth, 32))) |
+                  for (@val <- valCh) {
+                    if (val == Nil) { ret!(Nil) }
+                    else { ret!(val) }
```
ArturGajowy: I think this is equivalent to `ret!(rest.get(hash.slice(depth, 32)))`?

ghost: Oops, yes.  Thx

casper/src/main/resources/TreeHashMap.rho
ghost: Done

casper/src/main/resources/AsyncMap.rho
ghost: Done

casper/src/main/resources/AsyncMap.rho
ghost: We could.  JavaScript does the same thing with their map-like "object literals".  To tell if a key is in the map, you use the 'in' keyword.

casper/src/main/resources/AsyncMap.rho
ghost: Sorry, I misunderstood what you meant by key---you meant the externally provided key, not the internally used hash of the key.  I could store the item under the key, but that would mean keeping a copy of the entire key.  If keys serialize to much less than (32-depth) bytes, that's a win.  For the use case of a public key, the serialized form will be larger than that.

You're right that storing it under the key will completely prevent collisions, but the current implementation uses a 256-bit cryptographically secure hash function.  If you can cause a collision, keccak and every system that depends on it is broken.

casper/src/main/resources/TreeHashMap.rho
pyrocto: Done.

casper/src/test/resources/TreeHashMapTest.rho
```diff
@@ -0,0 +1,145 @@
...
+  new TreeHashMapCh in {
+    rl!(`rho:lang:treeHashMap`, *TreeHashMapCh) |
+    for(@(_, *TreeHashMap) <- TreeHashMapCh) {
+      contract test_get_before_set(rhoSpec, _, ackCh) = {
+        new ret, ch in {
+          TreeHashMap!("init", 3, *ret) |
+          for (@thm <- ret) {
+            TreeHashMap!("get", thm, "no such key", *ch) |
+            for (@x <- ch) {
+              ch!(x) |
```
ArturGajowy: You can skip the above receive and send. That's the purpose of `<-` in `== <-`.

casper/src/test/resources/TreeHashMapTest.rho
```diff
@@ -0,0 +1,145 @@
...
+        }
+      } |
+    
+      contract test_get_after_set(rhoSpec, _, ackCh) = {
+        new ret, ch1, ch2 in {
+          TreeHashMap!("init", 3, *ret) |
+          for (@thm <- ret) {
+            TreeHashMap!("set", thm, "some key", "some val", *ch1) |
+            for (_ <- ch1) {
+              TreeHashMap!("get", thm, "some key", *ch1) |
```
ArturGajowy, line 48: please don't reuse return channels, it gets confusing very easily

casper/src/main/resources/TreeHashMap.rho
```diff
@@ -0,0 +1,257 @@
...
+          for (@val <- valCh) {
+            if (n == len) {
+              ret!(val.get(suffix))
+            } else {
+              // Otherwise check if the rest of the path exists.
+              // Bit k set means node k exists.
+              // nybList.nth(n) is the node number
+              // val & powers.nth(nybList.nth(n)) is nonzero if the node exists
+              // (val/powers.nth(nybList.nth(n))) % 2 is 1 if the node exists
+              if ((val/powers.nth(nybList.nth(n))) % 2 == 0) {
```
ArturGajowy: Please add spaces around binary operators (`/`)

casper/src/main/resources/TreeHashMap.rho
```diff
@@ -0,0 +1,257 @@
...
+                TreeHashMapGetter!(map, nybList, 0, 2 * depth, hash.slice(depth, 32), *ret)
+              }
+            }
+          }
+        }
+      } |
+
+      // Doesn't walk the path, just tries to fetch it directly.
+      // Will hang if there's no key with that 64-bit prefix.
+      // Returns [] like "get" does if there is some other key with
```
ArturGajowy: Returns `Nil` like `"get"?

ArturGajowy (approved):
Just minor comments, LGTM after addressing them :+1:

casper/src/test/resources/TreeHashMapTest.rho
ghost: done

casper/src/test/resources/TreeHashMapTest.rho
ghost: done

casper/src/main/resources/TreeHashMap.rho
ghost: done

casper/src/main/resources/TreeHashMap.rho
ghost: thx, done

ghost:
bors try

ghost:
bors try

ghost:
bors try

lukasz-golebiewski (approved):

PR #2642, opened 2019-08-05
Implement merging rules

lukasz-golebiewski:
## Overview
Implements most of the merging rules from the spreadsheet

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3638

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2666, opened 2019-08-14
Mergeability rules self validation

lukasz-golebiewski:
## Overview
Follow-up for https://github.com/rchain/rchain/pull/2642

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3638

### Notes


casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -99,149 +172,236 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
-    "CX (C!)"   -> PERSISTENT_COULD_NOT_MATCH -> merges(C_, C_ | S0, Nil),
-    "(!4) 4!"   -> VOLATILE_EVENT             -> merges(S1 | F1, F_, S0),
-    "(!4) !C"   -> VOLATILE_EVENT             -> merges(S1 | F1, S0, C0),
-    "(!4) C!"   -> VOLATILE_EVENT             -> merges(S0 | F0, C_, S1),
-    "(4!) CX"   -> VOLATILE_EVENT             -> merges(F_ | S0, C_, Nil),
-    "(C!) (C!)" -> PERSISTENT_COULD_NOT_MATCH -> merges(C_ | S0, C_ | S0, Nil)
+    "!X !X"     -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, S0, Nil),
+    "!X !4"     -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, S1, F1),
+    "!X (!4)"   -> VolatileEvent.desc           -> VolatileEvent.mergeable(S0)(S0, F_)(Nil),
+    "!X !C"     -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, S1, C1),
```
ArturGajowy: We should avoid the $FOO.desc -> $FOO.mergable/conflicting duplication on each row by having the values in the map be instances of the follwing:

```scala
trait MergeabilityTestCase {
  val description: String
  def preconditions: F[Unit]
  def assertion: F[Unit]
}
```

then:
```scala
type Name = String // for lack of a better name, e.g. !4 !4

val mergeabilityCases = Map[Tag, MergeabilityTestCase](
  "!X !X"     -> SamePolarityMerge(S0, S0, Nil)
  ...
)
```
and the test would be sth. like:

```scala
  mergeabilityCases.toList.parTraverse { case (name, c) =>
    val result = (for {
      _ <- c.precondition
      _ <- c.assertion
    } yield ()).attempt
	result.map(name -> _)
  } >>= (results => /* summarize results into a nice assertion */)
```

To handle the MERGES/CONFLICTS subcases of COULD_MATCH_SAME_*, we can either:
 - have `if`-s in the implementation cases
 - have separate trait instances for the subcases (I prefer this)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -99,149 +172,236 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    "!X 4!!"    -> HadItsMatch.desc             -> HadItsMatch.mergeable(S0, F_, R0),
+    "!X (4!!)"  -> PersistentNoMatch.desc       -> PersistentNoMatch.mergeable(S0)(F_, R0)(Nil),
+    "!X !!X"    -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, R0, Nil),
+    "!X !!4"    -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, R1, F1),
+    "!X (!!4)"  -> PersistentNoMatch.desc       -> coveredBy("!X (4!!)"),
+    "!X CX"     -> IncomingCouldMatch.desc      -> IncomingCouldMatch.conflicting(S0, C_, Nil),
+    "!X C!"     -> IncomingCouldMatch.desc      -> IncomingCouldMatch.conflicting(S0, C_, S0),
+    "!X (C!)"   -> PersistentCouldMatch.desc    -> PersistentCouldMatch.conflicting(S0)(C_, S0)(Nil),
+    "!4 !4"     -> CouldMatchSame.descConflicts -> CouldMatchSame.conflicting(S0, S1, F_),
+    "!4 !4"     -> CouldMatchSame.descMerges    -> CouldMatchSame.mergeable(S0, S1)(F0, F1),
```
ArturGajowy: Notice that in:

`"!4 !4"     -> CouldMatchSame.descMerges    -> CouldMatchSame.mergeable(S0, S1)(F0, F1),`

There's at least 5 kinds of errors possible:
1) mismatch of mergeability category between 1st and 2nd column
2) mismatch of mergeability category between 2nd and 3rd column (due to $FOO.{desc, merges} duplication)
3) mismatch of descMerges/descConflicts with 1st column
4) mismatch of mergable/conflicts with 2nd column
5) mismatch of actual test case with 1st column

The approach I suggest in previous comment elliminates 2) and 4) and will provide way of extension to elliminate the others too.

I'm pointing this out, because I find this crucial: whenever we pass two values separately (vs as a single value) we need to make sure the values are valid w.r.t each other, which requires separate logic outside the place we pass the values in.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: We can skip (get rid of) the `coveredBy` cases, or provide actual examples for them too.

ArturGajowy (changes requested):
Sharing what I have now, and proceeding to add more comments :)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: What you propose isn't going to work because a `Map` cannot have duplicate keys and so the last entry for `!X !X` is going to win and all previous values are going to be ignored. Originally the description was used to disambiguate between the keys/tags/names. Another way to mitigate this would be to get rid of the `Map` and store tuples in a `List`. The second element of the tuple would be a sequence of cases (of size one for most). WDYT?

lukasz-golebiewski: I am not in favor of introducing a single trait for all test cases. Implementing what I proposed in the comment above feels enough. I agree that it makes sense to split the `CouldMatch*` case into two.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -99,149 +172,236 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    "!4 (!4)"   -> VolatileEvent.desc           -> VolatileEvent.mergeable(S0)(S1, F_)(F0),
+    "(!4) (!4)" -> VolatileEvent.desc           -> VolatileEvent.mergeable(S0, F_)(S0, F_)(Nil),
+    "!4 !C"     -> CouldMatchSame.descMerges    -> CouldMatchSame.mergeable(S0, S1)(F0, C1),
```
ArturGajowy: The number of variants of arities is confusing and inconsistent across the cases.
E.g. here it should say `CouldMatchSame(S0)(S1)(F0, C1)` to be consistent
 
Let's have all MergeabilityCases take 3 curried parameters: left, right, base.
The last one can be vararg.

For 'volatile/local' cases we can have all of them be wildcards.

```scala
def samePolarityMerge(left: Rho)(right: Rho)(base: Rho*): MergrabilityTestCase = new ...

def volatileEvent(lefts: Rho*)(rights: Rho*)(base: Rho*): MergrabilityTestCase = new ...
```

This works with an `apply` method on a companion object as well.

Assertions underneath then become
e.g. 
```
//all are linear
assert((lefts ++ rights).forall(_.cardinality == Linear))
//same polarity
assert(left.polarity == right.polarity) 
//different polarities within one side 
assert(noDuplicatesBy(lefts)(_.polarity))
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Good point! Using a map will cause 'duplicate key silent errors' without means of validating it. I'm against nested lists though. 

We should use a (flat) list, and enforce the names to be unique / make them unique by combining with the list index.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Good idea, I like it

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: > I am not in favor of introducing a single trait for all test cases.

It's more about the duplication of
`!X !4 -> $FOO.desc -> $FOO.{merges/fails}`
instead of having a common interface. The latter is just a means.

I think you'll find the trait needed to obtain the description for the assertion message. Notice also: once you change the `merges/conflicts` to `assertion` (which is needed, as the assertion should be same for all examples of a given case), the common interface is already there. Adding a trait atop is a matter of a couple `extends`.

lukasz-golebiewski: This has a downside of unnecessarily complicating the test which checks whether all cases from the spreadsheet are covered. Hence I would rather go with a list. Unless there is a better argument for what you are proposing

ArturGajowy: I'm fine if we fuse `preconditions` and `assertion` into just `assertion`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,31 +66,103 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      (b1.maybeCardinality === Some(NonLinear)) || (b2.maybeCardinality === Some(NonLinear)),
+      "expected one of the events to be persistent"
+    )
+
+  object SamePolarityMerge {
+    val desc =
+      """
         |Two incoming sends/receives, at most one had a matching dual in TS.
         |The incoming events won't cause more COMMs together (same polarity).
         |They couldn't be competing for the same linear receive/send (at most one had a match).
```
ArturGajowy: consider this:

```
def matches(x: Rho, y: Rho) = {
  val (a, b) = if (x.polarity == Send) (x, y) else (y, x)
  val wildcards = List(F_, C_)
  val knownMatches = 
    List(S0, R0).flatMap(s => (F0 :: C0 :: wildcards).map(r => s -> r)) ++
    List(S1, R1).flatMap(s => (F1 :: C1 :: wildcards).map(r => s -> r))
  
 knownMatches.contains(a -> b)
}
```

Now we can incorporate the preconditions based on what matches what

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: > This has a downside of unnecessarily complicating the test which checks whether all cases from the spreadsheet are covered.

OTOH having the a list for each case instead of just a single value massively complicates the main test, and puts a lot of strain on the compiler - just to have 4-ish names have two tests instead of one.

The complication you mention is simply:
add the index before doing any set operations (e.g. all cases are one of those expected),
drop the index before doing any exhaustivity checks (e.g. each expected case has at least one representant)

ArturGajowy: Frankly, you'd have coded it up faster than the time we've spent discussing this.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -99,149 +172,236 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    "(!4) (!4)" -> VolatileEvent.desc           -> VolatileEvent.mergeable(S0, F_)(S0, F_)(Nil),
+    "!4 !C"     -> CouldMatchSame.descMerges    -> CouldMatchSame.mergeable(S0, S1)(F0, C1),
+    "!4 4X"     -> HadItsMatch.desc             -> HadItsMatch.mergeable(S0, F_, F_),
+    "!4 4!"     -> HadItsMatch.desc             -> HadItsMatch.mergeable(S0, F_)(F0, S1),
+    "!4 (4!)"   -> VolatileEvent.desc           -> VolatileEvent.mergeable(S0)(S1, F_)(F0),
+    "!4 4!!"    -> HadItsMatch.desc             -> HadItsMatch.mergeable(S0, F_)(F0, R1),
+    "!4 !!X"    -> SamePolarityMerge.desc       -> SamePolarityMerge.mergeable(S0, R1, F0),
+    "!4 !!4"    -> CouldMatchSame.descConflicts -> CouldMatchSame.conflicting(S0, R1, F_),
+    "!4 !!4"    -> CouldMatchSame.descMerges    -> CouldMatchSame.mergeable(S0, R1)(F0, F1),
+    "!4 CX"     -> HadItsMatch.desc             -> HadItsMatch.mergeable(S0, C_, F_),
```
ArturGajowy: I know you consider this overkill, but validating the name (`!4 CX`) matches the mergeability case (HadItsMatch) is what was missing for spotting quite a few wrong tests during the review automatically, and who knows how many currently and in the future.

At the same time, parsing the:
```
case class Event(p: Polarity, c: Cardinality) 
case class Example(leftIn: Event, rightIn: Event, leftBase: Option[Event], rightBase: Option[Event])
```
from the Name-s should really be straightforward.

Just as the assertions to be done based on them:
```scala
//same/different polarity:
ex.leftIn.polarity ==/!= ex.rightIn.polarity

//number of matches
(leftBase, rightBase).exists(_.isEmpty) //at most one match
```

I implore you to reconsider this :)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: > massively complicates the main test, and puts a lot of strain on the compiler 

This is greately exaggerated

> Frankly, you'd have coded it up faster than the time we've spent discussing this.

The comment above could be spared, it doesn't bring anything. We're talking about sensibility of the approaches

Why are you suggesting combining with an index at all? Entries in a `List` do not have to be unique

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Ok, I'm going to reconsider

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: > This is greately exaggerated

Maybe w.r.t. to the compiler part. But usage-wise, reading and editing
`"4! 4!" -> List(CouldMatchSameMerges(X, Y, Z))`
on each line, is going to be a pain. It also makes the actual validations and the main test harder, b/c now you're handling a list of lists.

You're right that combining with an index might not even be needed. I suggested it as a means of regaining uniqueness, if it's needed in validations (e.g. we build a set somewhere). Most likely it's not needed though.

> The comment above could be spared, it doesn't bring anything. We're talking about sensibility of the approaches

I though really believe that the raised problem of `unnecessarily complicating the (spreadsheet coverage) test` would've been cheaper to handle in that test without this discussion. So while I'm all for considering sensibility of approaches, I hate to discuss such minutia.
I don't know, maybe I'm overlooking sth big in there?

lukasz-golebiewski:
```
Maybe w.r.t. to the compiler part. But usage-wise, reading and editing
"4! 4!" -> List(CouldMatchSameMerges(X, Y, Z))
on each line, is going to be a pain. It also makes the actual validations and the main test harder, b/c now you're handling a list of lists.
```

Maybe the reason I don't find this harder is that I'm used to handling nested lists and flattening doesn't give me headaches. I find grouping related cases under a single key more natural (and less error prone as it is easier to spot a missing case). Since this is largely opinion based I think we need another opinion to decide the way forward.

> I don't know, maybe I'm overlooking sth big in there?

Yes, the comment is inappropriate because perceived speed of developing a review suggestion doesn't say anything about it's sensibility

ArturGajowy: I was asking about the complications in `unnecessarily complicating the (spreadsheet coverage) test`.

Re: the meta issue - sorry, I should've phrased that differently.

lukasz-golebiewski: Looks like the `description` field is unnecessary. It's contents are going to the scaladoc

lukasz-golebiewski: And since the precondition can also go away, so can the whole `MergeabilityTestCase` trait. `Effect[Unit]` being the second element of the tuple is enough

ArturGajowy: I thought you'd say that :) Though to validate that the Name matches the MergeabilityCase we need a common interface to provide the name to - assuming we want to retain alignment and readability. To elaborate - we can either do:

`"!4 !4 -> HadItsMatch(X, Y, Z)` and in the test:
`(mergeabilityCase.checkMatchesName(name) >> mergeabilityCase.assertion).attempt`

or

`hadItsMatch("!4 !4", X, Y, Z)` and in the test:
`value.attempt`

I prefer the former, as the latter reads worse and aligns worse

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: This will have to wait until after rules for peek are implemented

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Makes sense, although it's going to need to wait

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: I can live with that. Can you though? :) I mean, that's additional safety net that would help both of us when implementing peek. I'm fine either way though :)

lukasz-golebiewski: I treat this self validation as a means to spare the future maintainers some stress. I've been looking at those examples for so long that I know them by heart already, so yes, I can live with that

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,191 +66,414 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+
+  def onePersistent(rho: Seq[Rho]): Boolean =
+    rho.forall(_.maybeCardinality.isDefined) &&
+      rho.exists(_.maybeCardinality.get == NonLinear)
+
+  def matches(x: Rho, y: Rho) = {
+    val (a, b)    = if (x.maybePolarity == Some(Send)) (x, y) else (y, x)
+    val wildcards = List(F_, C_)
+    val knownMatches =
+      List(S0, R0).flatMap(s => (F0 :: C0 :: wildcards).map(r => s   -> r)) ++
```
dzajkowski: `.map(r => s   -> r)) ++` formatting?

dzajkowski (approved):

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -204,23 +204,19 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
     "!C (C!)"   -> PERSISTENT_COULD_NOT_MATCH -> coveredBy("!C (!C)"),
     "4X 4X"     -> SAME_POLARITY_MERGE        -> merges(F_, F_, Nil),
     "4X 4!"     -> SAME_POLARITY_MERGE        -> merges(F0, F_, S1),
-    // Skipping 4X 4!! merges, 4X !!X may merge or not, 4X !!4 may merge or not
```
ArturGajowy, line 198: We should add those missing cases at some point though. Maybe in this PR?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -131,6 +131,13 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
         |These could not spawn more work.
         |""".stripMargin
 
+  object SamePolarityMerge {
+
+    def apply(left: Rho*)(right: Rho*)(base: Rho*): Effect[Unit] =
+      merges(left.reduce(_ | _), right.reduce(_ | _), base.reduce(_ | _))
```
ArturGajowy: Love the ASCII art :heart:

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -184,103 +184,102 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
   )
 
   it should "respect mergeability rules when merging blocks" in effectTest {
     mergeabilityCases
       .map {
-        case ((name, desc), y) =>
-          val key = s"$name $desc"
-          y.attempt.map(_.fold(_ => s"$key FAILED", _ => key))
+        case (key, test) =>
+          test.attempt.map(_.fold(_ => s"$key FAILED", _ => key))
```
ArturGajowy: How do we discern same-named cases when a test fails?
One way out would be to force distinct names :)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      leftPolarity  <- left.flatMap(_.maybePolarity)
+      rightPolarity <- right.flatMap(_.maybePolarity)
+    } yield {
+      assert(leftPolarity == rightPolarity)
+    }
+
+  def differentPolarities(rho: Rho*)(
+      implicit
+      pos: source.Position
+  ): Any =
```
ArturGajowy: please add a better type :)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+        base: Rho*
+    )(implicit file: sourcecode.File, line: sourcecode.Line, pos: source.Position): Effect[Unit] =
+      precondition(left: _*)(right: _*)(base: _*) >>
+        conflicts(left.reduce(_ | _), right.reduce(_ | _), base.reduce(_ | _))
+  }
+
+  def samePolarities(left: Seq[Rho])(right: Seq[Rho])(
+      implicit
+      pos: source.Position
+  ) =
```
ArturGajowy: please add a return type :)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      rightPolarity <- right.flatMap(_.maybePolarity)
+    } yield {
+      assert(leftPolarity == rightPolarity)
+    }
+
+  def differentPolarities(rho: Rho*)(
+      implicit
+      pos: source.Position
+  ): Any =
+    if (rho.size == 2) {
```
ArturGajowy: While technically it works, because there are only two polarities (or 3, if we count None, which doesn't affect the assertion) - it's quite confusing and raises doubt immediately.

I'd much prefer a solution that does `assert(noDuplicates(rho.flatMap(_.maybePolarity))`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -110,35 +227,86 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
     */
-  object HadItsMatch extends MergeableCase
+  object HadItsMatch extends MergeableCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] = {
+      differentPolarities(left)(right)
+      crossedPolarity(left)(base)
+      crossedPolarity(right)(base)
```
ArturGajowy: I'm not sure about the relevance of the `crossedPolarity` check. Its occurrence isn't used in the reasoning in the scaladoc; it is only checked for cases where there is two terms in base deploy. I also think whatever it checks is going to be coverd by `someMatch` case in the latter commits (a match implies existence of different polarities). I'd remove it, but maybe there's a benefit I'm missing?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+      ((rho.maybePolarity === base1.maybePolarity) && (rho.maybePolarity !== base2.maybePolarity)) || ((rho.maybePolarity !== base1.maybePolarity) && (rho.maybePolarity === base2.maybePolarity))
+    )
+  }
+
+  def allLinear(rho: Seq[Rho])(
+      implicit
+      pos: source.Position
+  ) = {
+    assert(rho.forall(_.maybeCardinality.isDefined))
+    assert(rho.forall(_.maybeCardinality.get == Linear))
```
ArturGajowy: people and tools justly frown at the usage of `Option.get`. Why not just `assert(rho.forall(_.maybeCardinality == Some(Linear))`?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -110,35 +227,86 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
     */
-  object VolatileEvent extends MergeableCase
+  object VolatileEvent extends MergeableCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] = {
+      if (left.size == 1 && right.size == 1) fail("No volatile COMM event")
+      differentPolarities(left: _*)
+      allLinearWhenTwo(left)
```
ArturGajowy: I think this too is going to be covered by the `matches` cases added later. Can we remove it?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    assert(rho.forall(_.maybeCardinality.isDefined))
+    if (rho.size == 2)
+      assert(rho.forall(_.maybeCardinality.get == Linear))
+  }
+
+  def onePersistent(rho: Seq[Rho])(
+      implicit
+      pos: source.Position
+  ) = {
+    assert(rho.forall(_.maybeCardinality.isDefined))
```
ArturGajowy: this is more than the function name says. It's also not relevant to the preconditions this method's used in. I'd remove this line.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,14 +66,100 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    assert(rho.exists(_.maybeCardinality.get == NonLinear))
```
ArturGajowy: We can avoid the `.get` by `rho.flatMap(_.maybeCardinality).exists(_ == NonLinear)`

ArturGajowy: Also, this (and what's currently in code) would be `atLeastOnePersistent`. We should probably just make it `rho.filter(_.maybeCardinality == Some(NonLinear)).size == 1`

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -234,14 +402,14 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
-    "(!C) (4!)" -> VolatileEvent(S0, C_)(F_, S0)(Nil),
+    "(!C) (4!)" -> PersistentNoMatch(S0, C_)(F_, S0)(Nil),
     "(!C) (C!)" -> PersistentNoMatch(S0, C_)(C_, S0)(Nil),
     "(!C) 4!"   -> PersistentCouldMatch(S0, C_)(F_)(S0),
     "(!C) 4X"   -> PersistentNoMatch(S0, C_)(F_)(Nil),
     "(!C) C!"   -> PersistentCouldMatch(S0, C_)(C_)(S0),
     "(!C) CX"   -> PersistentNoMatch(S0, C_)(C_)(Nil),
     "(4!) (4!)" -> VolatileEvent(F_, S0)(F_, S0)(Nil),
-    "(4!) (C!)" -> VolatileEvent(F_, S0)(C_, S0)(Nil),
+    "(4!) (C!)" -> PersistentNoMatch(F_, S0)(C_, S0)(Nil),
```
ArturGajowy: yay! :joy:

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -154,14 +174,38 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+        List(S1, R1).flatMap(s => (F1 :: C1 :: wildcards).map(r => s -> r))
+
+    knownMatches.contains(a -> b)
   }
 
+  def allMatch(left: Seq[Rho], right: Seq[Rho]) =
+    (for {
+      l <- left
+      r <- right
+    } yield matches(l, r)).fold(true)(_ && _)
```
ArturGajowy: We could make that:
```
    Applicative[List]
      .product(left.toList, right.toList)
      .forall(Function.tupled(matches))
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -154,14 +174,38 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+
+  def someMatch(left: Seq[Rho], right: Seq[Rho]) =
+    (for {
+      l <- left
+      r <- right
+    } yield matches(l, r)).fold(false)(_ || _)
```
ArturGajowy: Similarly, this could be:
```
    Applicative[List]
      .product(left.toList, right.toList)
      .exists(Function.tupled(matches))
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -154,14 +174,38 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+
+  def ifCommEvent(rhos: Seq[Rho])(f: (Rho, Rho) => Assertion) =
+    if (rhos.size == 2) {
```
ArturGajowy: this doesn't check that the args form a comm event. Maybe this should be `ifTwo`?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -176,6 +220,7 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
       samePolarities(left)(right)
       differentPolarities(left)(base)
       differentPolarities(right)(base)
+      assert(!allMatch(left ++ right, base))
```
ArturGajowy: This would also pass if there were 2 matches, as long as there is one non-matching pair.
At this point, operating on lists for non-volatile cases really gets in the way.
We could say `assert(someMatch(left, base) != someMatch(right, base))`.
What we would really like to say, is:
`assert(hasMatch(left, base) != hasMatch(right, base))`, where left and right are singular Rho (not lists).

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Also, its definition is very special-cased

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,191 +66,414 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+   There was a COMM within one of the deploys, with one side non-linear.
+   The other deploy had an event without a match in TS, of same polarity to the non-linear.
+   These could not spawn more work.
+    */
+  object PersistentNoMatch extends VolatileCase with MergeableCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] =
+      super.precondition(left: _*)(right: _*)(base: _*) >> {
```
ArturGajowy: Can we replace the inheritance with calling a common method with a telling name?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,191 +66,414 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    */
+  object CouldMatchSameMerges extends MergeableCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] = {
+      samePolarities(left)(right)
+      differentPolarities(left)(base)
+      differentPolarities(right)(base)
+      assert(!allMatch(left ++ right, base) || onePersistent(base))
```
ArturGajowy: This should be `allMatch(left, base) && allMatch(right, base`), + an assertion that they've either matched different events, or the matched-by-both event is non-linear.

I can't make a connection between the assertion in code ane the mergability rules/cases we're testing.

if we had signular `left` and `right`, this could be:

```
val m1 = findMatch(left, base) //throws asseriton error if not found
val m2 = findMatch(right, base) // we could also return Option-s and do the needed assertions
assert(m1 != m2 || m1.cardinality == NonLinear || m2.cardinality == NonLinear)
```

Notice the `Conflicts` case is exactly the same, just with the last assertion negated.
(`m1 == m2 && m1.cardinality == Linear && m2.cardinality == Linear`)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,191 +66,414 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
+    */
+  object HadItsMatch extends MergeableCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] = {
+      differentPolarities(left)(right)
+      crossedPolarity(left)(base)
+      crossedPolarity(right)(base)
+      assert(someMatch(left, base) || someMatch(right, base))
```
ArturGajowy: it'd be great to encode the
```
 either:
   - both were linear
   - one was non-linear, the other had a match
```
criteria for the incoming events.

With singular left/right:
```
val (a, b) = if (left.cardinality == Linear) (left, right) else (right, left)
assert(a.cardinbality == Lineear && (b.cardinality == Linear || hasMatch(a, base)))
```

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -65,191 +66,414 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
-        |""".stripMargin
-
-  //deploy-local, merges
-  val PERSISTENT_COULD_NOT_MATCH =
-    """
-        |There's been a COMM within one of the deploys, with one side non-linear.
-        |The other deploy had an event without a match in TS, of same polarity to the non-linear.
-        |These could not spawn more work.
-        |""".stripMargin
+  trait TestCase {
```
ArturGajowy: Would be great to move this ant its descendants ot a separate file/package, probably with Rho and the util assertions

ArturGajowy (changes requested):
Great work! Had some questions/remarks

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: No, this one is big enough already

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I would rather add (very) short descriptions to all test cases

ArturGajowy: How about threading the source position?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Like what?

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: How about using the source position and displaying it in the assertion?

ArturGajowy: (are we doing that already? If so - case closed for me :))

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Like Seq[Assertion] f.e. Change the body if needed. See my latter comment on how you can change this.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Ok, I'll write something similar

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: `.get` is acceptable in tests, but you're right, the error message is going to be friendlier when written like you proposed

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Yes, this was added before the "match" assertions and is now superfluous. It can go away

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: This was auto-formatted, sorry

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Yes :-)

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I'd leave this as it's an additional safety measure for not letting persistent operations into COMM events for this case

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: What you're suggesting isn't going to work with `"!X !X"     -> SamePolarityMerge(S0)(S0)(Nil),`

ArturGajowy: The problem I raised still holds though. You get the idea. We should be asserting that at most one incoming event had a match.

lukasz-golebiewski: Yeah, I'll fix this

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: If you insist

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: It's enough to revert back to:
```
it should "respect mergeability rules when merging blocks" in effectTest {
    mergeabilityCases.map(_._2).parSequence
  }
```

but then it will only display the first failure. I think this is good enough

lukasz-golebiewski: This is already the case with precondition failures

ArturGajowy (approved):

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Ok, done

PR #2667, opened 2019-08-14
Delay effects properly in transformAndStore for actual parallelism

ArturGajowy:
Previously the whole `transformAndStore` stage was executed on a single
core, most likely due to the computation taking place during its
definition (vs during evaluation, as is the case when delayed properly).

Updated benchmarks:
https://docs.google.com/spreadsheets/d/11WdCneacNV4yWD14TXz2WKL-PxGYQbFGiAiXPNF0V-8/edit#gid=0

## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
@@ -149,6 +149,12 @@ final case class HistoryRepositoryImpl[F[_]: Sync: Par, C, P, A, K](
         s"${key.toHex};delete-join;0"
     }
 
+  private def transformAndStore(hotStoreActions: List[HotStoreAction]): F[List[HistoryAction]] =
+    for {
+      transformedActions <- Sync[F].delay(hotStoreActions.map(transform))
+      historyActions     <- storeLeaves(transformedActions)
+    } yield historyActions
+
   private def transform(action: HotStoreAction): Result =
     action match {
       case i: InsertData[C, A] =>
@@ -188,9 +194,6 @@ final case class HistoryRepositoryImpl[F[_]: Sync: Par, C, P, A, K](
     val toBeStored = leafs.collect { case (key, Some(data), _) => (key, data) }
     leafStore.put(toBeStored).map(_ => leafs.map(_._3))
   }
-
-  val transformAndStore: List[HotStoreAction] => F[List[HistoryAction]] =
-    ((l: List[HotStoreAction]) => l.map(transform)).andThen(storeLeaves)
 
   override def checkpoint(actions: List[HotStoreAction]): F[HistoryRepository[F, C, P, A, K]] = {
     implicit val P = Par[F].parallel


dzajkowski (approved):

odeac (approved):

lukasz-golebiewski (approved):

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
```diff
@@ -149,6 +149,12 @@ final case class HistoryRepositoryImpl[F[_]: Sync: Par, C, P, A, K](
         s"${key.toHex};delete-join;0"
     }
 
+  private def transformAndStore(hotStoreActions: List[HotStoreAction]): F[List[HistoryAction]] =
+    for {
+      transformedActions <- Sync[F].delay(hotStoreActions.map(transform))
```
lukasz-golebiewski, line 154: Can't this be parallellised using `hotStoreActions.par.map(transform)`?

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
```diff
@@ -149,6 +149,12 @@ final case class HistoryRepositoryImpl[F[_]: Sync: Par, C, P, A, K](
...
+      historyActions     <- storeLeaves(transformedActions)
```
lukasz-golebiewski, line 155: `hotStoreActions.map(transform)` is pure, so it would be enough to delay only `storeLeaves` here.

lukasz-golebiewski:
I find your benchmarks' results peculiar. How does `Sync[F].delay` improve performance?

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
ArturGajowy: I've considered that, but both me and Dom think it's better to not mix paradigms and use the same thread pool underneath :)

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
ArturGajowy: that's what's been here previously, and it ended up being executed during computation definition (becuase it's pure). Turns out parallelism is effectful :D

ArturGajowy:
> I find your benchmarks' results peculiar. How does Sync[F].delay improve performance?

Feel free to repeat the benchmarks, see the spreadsheet for HSCTN harness code.

I've double checked just now, switching .delay to .pure causes single-threaded execution in that area. The best explanation I can offer is in the commit message.

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
lukasz-golebiewski: Interesting find!

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
lukasz-golebiewski: Because of LMDB?

ArturGajowy: Because we can control/tune/override the one used by cats, and we can't the global one underneath list.par.map(...)

PR #2657, opened 2019-08-09
RCHAIN-3732: Verify vault properties

JosephDenman:
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3732

PR #2669, opened 2019-08-16
Remove prefix from eth addresses in wallets.txt

KentShikama:
## Overview

The title says it all

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
@@ -144,13 +144,13 @@ object Genesis {
   private def fromLine(line: String): Either[String, Vault] = line.split(",") match {
     case Array(ethAddressString, initRevBalanceStr, _) =>
       Try(initRevBalanceStr.toLong) match {
-        case Success(initRevBalance) =>
+        case Success(initRevBalance) if initRevBalance >= 0 =>
           RevAddress
             .fromEthAddress(ethAddressString)
             .map(Vault(_, initRevBalance))
             .toRight(s"Ethereum address $ethAddressString is invalid.")
         case Failure(_) =>
-          Left(s"Failed to parse given initial balance $initRevBalanceStr as long.")
+          Left(s"Failed to parse given initial balance $initRevBalanceStr as positive long.")
       }
 
     case _ => Left(s"Invalid vault specification:\n$line")
--- a/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
@@ -145,8 +145,7 @@ object Genesis {
     case Array(ethAddress, initRevBalanceStr, _) =>
       Try(initRevBalanceStr.toLong) match {
         case Success(initRevBalance) =>
-          val ethAddressWithoutPrefix = ethAddress.drop(2)
-          Right(Vault(RevAddress.fromEthAddress(ethAddressWithoutPrefix), initRevBalance))
+          Right(Vault(RevAddress.fromEthAddress(ethAddress), initRevBalance))
         case Failure(_) =>
           Left(s"Failed to parse given initial balance $initRevBalanceStr as long.")
       }
--- a/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
@@ -145,7 +145,8 @@ object Genesis {
     case Array(ethAddress, initRevBalanceStr, _) =>
       Try(initRevBalanceStr.toLong) match {
         case Success(initRevBalance) =>
-          Right(Vault(RevAddress.fromEthAddress(ethAddress), initRevBalance))
+          val ethAddressWithoutPrefix = ethAddress.drop(2)
+          Right(Vault(RevAddress.fromEthAddress(ethAddressWithoutPrefix), initRevBalance))
         case Failure(_) =>
           Left(s"Failed to parse given initial balance $initRevBalanceStr as long.")
       }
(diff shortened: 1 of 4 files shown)


JosephDenman (approved):

itegulov (approved):

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -145,7 +145,8 @@ object Genesis {
     case Array(ethAddress, initRevBalanceStr, _) =>
       Try(initRevBalanceStr.toLong) match {
         case Success(initRevBalance) =>
-          Right(Vault(RevAddress.fromEthAddress(ethAddress), initRevBalance))
+          val ethAddressWithoutPrefix = ethAddress.drop(2)
+          Right(Vault(RevAddress.fromEthAddress(ethAddressWithoutPrefix), initRevBalance))
```
ArturGajowy: This should be done in AddressTools.fromEthAddress, which in turn should assert that the prefix is there, and is of correct form. I see you mean to strip `0x` to be stripped - I agree we could make it optional. In any case (prefix optional or mandatory) the prefix needs to be checked (so that we don't strip anything different than 0x).

Also, AddressTools should get corresponding tests for that.

ArturGajowy (changes requested):

ArturGajowy:
Thank God the integration tests failed.

Also: how did you find the need for that change? This looks like a serious bug making the wallets.txt file useless. I'm asking b/c there's likely a scenario that we should be testing, and are not/do it wrongly.

KentShikama:
> how did you find the need for that change? This looks like a serious bug making the wallets.txt file useless. I'm asking b/c there's likely a scenario that we should be testing, and are not/do so wrongly.

The wallets.txt that is generated by the https://github.com/rchain/rhoc-balance-reporter script has 0x prefixes. Our integration tests don't have 0x prefixes.

ArturGajowy:
What does that mean for all the benchmarks that have been run with a 0x-prefixed wallets file?

KentShikama:
> What does that mean for all the benchmarks that have been run with a 0x-prefixed wallets file?

It doesn't matter. As you suggest, I'm changing this to be handled with AddressTools.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
ArturGajowy: Stripping the `0x` prefix should probably even be the responsibility of `Base16` class, so that it works for all addresses we use. I won't insist on that though.

ArturGajowy:
I see, there's no validation in th fromEthAddress method. with, or without prefix, the address would be keccak-hashed and used as payload for the RevAddress.

Please add address length validation in there fromEthAddress, along with unit tests.

KentShikama:
Isn't validation what https://github.com/rchain/rchain/pull/2657 is trying to accomplish?

ArturGajowy:
That one is attempting to use existing validation logic in a different place. That logic being inexistent is a separate bug. Also it makes sense to not wait with the fix b/c of timezones :)

KentShikama:
@ArturGajowy PTAL

KentShikama:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/util/RevAddressSpec.scala
```diff
@@ -17,4 +17,17 @@ class RevAddressSpec extends FlatSpec with Matchers {
...
+      Some("1111Gzo7ywxbcXVumSS9Lzd8JBAqnF1zniNszvMLHQ2APa3dzs2rG")
+    )
+  }
+
+  "fromEthAddress" should "work correctly with a prefix" in {
+    val ethAddress = "0x06a441c277bf454c5d159b0e5bdafca69b296733"
+    RevAddress.fromEthAddress(ethAddress).map(_.toBase58) should be(
+      Some("1111Gzo7ywxbcXVumSS9Lzd8JBAqnF1zniNszvMLHQ2APa3dzs2rG")
+    )
+  }
```
ArturGajowy, line 32: Please add negative cases: wrong alphabet, wrong length

ArturGajowy: We should also probably implement the address checksum checking according to https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md. I'm fine with a follow-up PR for that.

KentShikama: Done PTAL

ArturGajowy (approved):
Are you going to work on eth checksum validation right away? If not, please ticket it.

KentShikama:
See https://rchain.atlassian.net/browse/RCHAIN-3759

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -142,10 +142,16 @@ object Genesis {
     )
 
   private def fromLine(line: String): Either[String, Vault] = line.split(",") match {
-    case Array(ethAddress, initRevBalanceStr, _) =>
+    case Array(ethAddressString, initRevBalanceStr, _) =>
       Try(initRevBalanceStr.toLong) match {
         case Success(initRevBalance) =>
-          Right(Vault(RevAddress.fromEthAddress(ethAddress), initRevBalance))
+          RevAddress.fromEthAddress(ethAddressString) match {
```
odeac: This can be replaced by 
```scala
RevAddress.fromEthAddress(ethAddressString)
.map(Vault(_, initRevBalance))
.asRight(s"Ethereum address $ethAddressString is invalid.")
```

JosephDenman (approved):

JosephDenman:
While you're at it, you should go ahead and disallow negative initial vault values. Then, we could go ahead and close <https://github.com/rchain/rchain/pull/2657>.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
KentShikama: Done

KentShikama:
@JosDenmark Done

PR #2671, opened 2019-08-16
RCHAIN-3697: Create rholang-based registry

tymm:
## Overview
See title.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3697


ArturGajowy:
@tymm please rebase, there's quite a bunch of conlicts

tymm:
@ArturGajowy done

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -429,40 +404,19 @@ object Runtime {
...
-      implicit F: Concurrent[F]
-  ): F[Unit] = {
-    // This random value stays dead in the tuplespace, so we can have some fun.
-    // This is from Jeremy Bentham's "Defence of Usury"
-    val rand = Blake2b512Random(
-      ("there can be no such thing as usury: " +
-        "for what rate of interest is there that can naturally be more proper than another?")
+  def bootstrapRegistry[F[_]: FlatMap](runtime: Runtime[F]): F[Unit] = {
+    // This is from Nassim Taleb's "Skin in the Game"
+    implicit val rand = Blake2b512Random(
```
tymm, line 409: @ArturGajowy and I were chatting about having a new quote for the rholang-based registry.
This is my suggestion. Happy to take other suggestions or we can take a vote or something :)

tymm:
bors try

rholang/src/test/scala/coop/rchain/rholang/interpreter/CryptoChannelsSpec.scala
```diff
@@ -83,11 +83,10 @@ class CryptoChannelsSpec
   ): Any = {
     implicit val (reduce, store) = fixture
 
-    val serializeAndHash: (Array[Byte] => Array[Byte]) => Par => Array[Byte] =
```
tymm: Honestly I didn't really understand this test. Why did we serialize here?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
cboscolo: How about:
> I only take advice from a utilitarian when it maximizes happiness and well-being for the majority of a population.  -Chris Boscolo

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -103,6 +103,13 @@ class RhoSpec(
           3,
           105L,
           ctx => DeployerIdContract.get(ctx)(_, _)
+        ),
+        SystemProcess.Definition[F](
+          "rho:test:crypto:secp256k1Sign",
```
tymm, line 108: I wonder if we should make this available outside of RhoSpec?

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RegistryBootstrap.scala
```diff
@@ -0,0 +1,42 @@
...
+
+object RegistryBootstrap {
+  val AST: Par = Par(
+    news = Seq(
+      bootstrap(FixedChannels.REG_LOOKUP2),
+      bootstrap(FixedChannels.REG_INSERT_RANDOM2),
+      bootstrap(FixedChannels.REG_INSERT_SIGNED2)
+    )
+  )
+  private def bootstrap(channel: Par): New =
```
ArturGajowy: Please add a comment describing the equivalent rholang, and newlines around methods/values for readability

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
ArturGajowy: Quote taken aside, I love the act itself. It's a great way to stay true to the original implementation's spirit :joy:

rholang/src/main/scala/coop/rchain/rholang/interpreter/registry/RegistryBootstrap.scala
```diff
@@ -1,4 +1,4 @@
-package coop.rchain.casper.genesis.contracts
+package coop.rchain.rholang.interpreter.registry
```
ArturGajowy, line 1: can you fix the commit log to include the moved version in the first commit?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -376,12 +374,12 @@ object RuntimeManager {
       _                <- active.space.clear()
       _                <- active.replaySpace.clear()
       _                <- Runtime.injectEmptyRegistryRoot(active.space, active.replaySpace)
+      _                <- Runtime.bootstrapRegistry(active)
```
ArturGajowy: can you rename `active` to `runtime` on the occasion? :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -209,6 +210,23 @@ object SystemProcesses {
...
+            case RhoType.Name(gprivate) =>
+              val hashKeyBytes = Blake2b256.hash(gprivate.id.toByteArray)
+              RhoType.Uri(Registry.buildURI(hashKeyBytes))
+            case RhoType.ByteArray(ba) =>
+              val hashKeyBytes = Blake2b256.hash(ba)
+              RhoType.Uri(Registry.buildURI(hashKeyBytes))
+            case _ => Par()
+          }
+          produce(Seq(response), ack)
+      }
```
ArturGajowy, line 226: Please add RhoSpec for this contract, including one showing that

`buildUri!(name) == buildUri!(name.toByteArray)`

If the above is not true, let's just have one overload of `buildUri`, taking a byteArray.

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
ArturGajowy: It'd be kind-of useless, if not dangerous: any usage on-chain would mean the private key used for signing is no longer private. Even if one passes an unforgrable name as private key, one can extract its hash off-chain and pass the private key as a byte array.

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -103,6 +103,13 @@ class RhoSpec(
...
+          Runtime.byteName(106),
```
ArturGajowy, line 109: please double check there's no conflict on that magic number. Ideally, we'd have all those magic numbers extracted as constants in some object, so it's less easy to shoot oneself in the foot.

casper/src/main/resources/Registry.rho
```diff
@@ -0,0 +1,80 @@
...
+    }
+  } |
+
+  bootstrapInsertArbitrary!(*insertArbitraryCh) | // this will work only once
+  for (insertArbitrary <- insertArbitraryCh) {
+    contract insertArbitrary(data, ret) = {
+
+      new seed, uriCh in {
+        ops!("buildUri", *seed, *uriCh) |
+        for (@uri <- uriCh; @map <- _registryStore) {
```
ArturGajowy: please don't use joins, they don't merge till Venus most likely

cc @odeac, I think I've seen some joins in PoS code

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
```diff
@@ -105,7 +105,7 @@ object CostAccountingPropertyTest {
     mkRuntime[Task](prefix, 1024 * 1024).use { runtime =>
       for {
         _    <- runtime.cost.set(Cost.UNSAFE_MAX)
-        _    <- Runtime.injectEmptyRegistryRoot[Task](runtime.space, runtime.replaySpace)
+        _    <- Runtime.bootstrapRegistry[Task](runtime)
```
ArturGajowy: I've got a hunch that registry is not needed for this test. Can you check this? (just delete this line :))

casper/src/test/resources/RegistryTest.rho
```diff
@@ -0,0 +1,77 @@
...
+      (pk, sk) => {
+        new signCh, verifyCh, retCh, hashCh in {
+          blake2b256!("foo", *hashCh) |
+          for (@hash <- hashCh) {
+            sign!(hash, sk, *signCh) |
+            for (@sig <- signCh) {
+              secpVerify!(hash, sig, pk, *verifyCh) |
+              for (@b <- verifyCh) {
+                retCh!(b) |
+                rhoSpec!("assert", (true, "== <-", *retCh), "Expecting true", *ackCh)
```
ArturGajowy: You can skip the verifyCh -> retCh sending, and pass verifyCh to the assertion direclty

casper/src/test/resources/RegistryTest.rho
```diff
@@ -0,0 +1,77 @@
...
+  } |
+
+  contract test_insertSignedAndLookup(rhoSpec, _, ackCh) = {
+    match (1,
+           "foo",
+           "04cc94ab15247b0db1b8219388218eb7461ac74b3d88d4da8c1816fac8e258f5e0f5db9db6bb82b0cc066589dfe77a0d7449db295dab248fb93855ba91813154a9".hexToBytes(),
+           "4d24bf7b37a4b3dc417b8e0e03a0f232b31a535f4880a73575ac855acae49640".hexToBytes()) {
+      (nonce, data, pk, sk) => {
+        new uriCh, signCh, valueCh, hashCh in {
+          blake2b256!((nonce, data), *hashCh) |
```
ArturGajowy: we need additional tests:
 - inserted nonce is bigger than the present one on re-insertion
 - make sure we don't integer-overflow the nonce (nonce = MAX_LONG means no update of the entry is possible)
 - lookup of inexistent uri returns Nil (or whatever the old registry does)
 - insertSigned with wrong signature returns Nil (or whatever the old registry does)

casper/src/test/resources/RegistryTest.rho
```diff
@@ -0,0 +1,77 @@
+//requires RhoSpec
+new
+  RhoSpecCh,
+  rl(`rho:registry:lookup`),
+  rs(`rho:registry:insertSigned:secp256k1`),
+  rr(`rho:registry:insertArbitrary`),
```
ArturGajowy: please name `rl, rs, rr` with full names (`lookup`, `insertSigned` `insertArbitrary`)

rholang/src/main/scala/coop/rchain/rholang/interpreter/registry/Registry.scala
```diff
@@ -1,4 +1,4 @@
-package coop.rchain.rholang.interpreter
```
ArturGajowy: The commit `Move Registry.scala` is not needed, as this PR deletes Registry.scala

casper/src/main/resources/Registry.rho
```diff
@@ -64,8 +64,10 @@ in {
                   true => {
                     ops!("buildUri", *pubKeyBytes, *uriCh) |
                     for (@uri <- uriCh; @map <- _registryStore) {
-                      _registryStore!(map.set(uri, (nonce, data))) |
-                      ret!(uri)
+                      match map.get(uri) {
+                        Nil => { _registryStore!(map.set(uri, (nonce, data))) | ret!(uri) }
+                        (oldNonce, _) => { if (nonce > oldNonce) { _registryStore!(map.set(uri, (nonce, data))) | ret!(uri) } else ret!(Nil) }
+                      }
```
ArturGajowy, line 82: Please squash this into the original `registry.rho` commit

casper/src/test/resources/RegistryTest.rho
```diff
@@ -1,77 +1,118 @@
...
+  in {
+    rl!(`rho:id:zphjgsfy13h1k85isc8rtwtgt3t9zzt5pjd5ihykfmyapfc4wt3x5h`, *RhoSpecCh) |
+    for(@(_, RhoSpec) <- RhoSpecCh) {
+      @RhoSpec!("testSuite",
+        [
+          ("Verifying a signature should work", *test_secp),
+          ("InsertSigned and lookup should work", *test_insertSignedAndLookup),
+          ("InsertArbitrary and lookup should work", *test_insertArbitraryAndLookup),
+          ("InsertArbitrary should work with increased nonce", *test_insertSignedValidNonce),
+          ("InsertArbitrary should return Nil for invalid nonce", *test_insertSignedInvalidNonce),
```
ArturGajowy, line 44: Likewise with the added tests, those should be part of the original commit

ArturGajowy (changes requested):
Good job! Comments below :)

Please `git rebase -i -x 'sleep 1.1'` to fix the ordering of commits on github :)

casper/src/main/resources/Registry.rho
```diff
@@ -0,0 +1,82 @@
...
+            ret!(map.get(shorthands.getOrElse(uriOrShorthand, uriOrShorthand)))
+          }
+        }
+      }
+    }
+  } |
+
+  bootstrapInsertArbitrary!(*insertArbitraryCh) | // this will work only once
+  for (insertArbitrary <- insertArbitraryCh) {
+    contract insertArbitrary(data, ret) = {
```
ghost: Please make all the contract signatures match what you did in the first one, where you put @ on those parameters used as values.

casper/src/test/resources/RegistryTest.rho
```diff
@@ -0,0 +1,118 @@
...
+    } |
+
+    contract test_secp(rhoSpec, _, ackCh) = {
+      new signCh, verifyCh, retCh, hashCh in {
+        blake2b256!("foo", *hashCh) |
+        for (@hash <- hashCh) {
+          sign!(hash, sk1, *signCh) |
+          for (@sig <- signCh) {
+            secpVerify!(hash, sig, pk1, *verifyCh) |
+            for (@b <- verifyCh) {
```
ghost: You can replace
```
for (@b <- verifyCh) {
  retCh!(b) |
  rhoSpec!("assert", (true, "== <-", *retCh), "Expecting true", *ackCh)
}
```
with
```
rhoSpec!("assert", (true, "== <-", *verifyCh), "Expecting true", *ackCh)
```

casper/src/test/resources/RegistryTest.rho
```diff
@@ -0,0 +1,118 @@
...
+              retCh!(b) |
+              rhoSpec!("assert", (true, "== <-", *retCh), "Expecting true", *ackCh)
+            }
+          }
+        }
+      }
+    } |
+
+    contract test_insertSignedAndLookup(rhoSpec, _, ackCh) = {
+      match (1, "foo") {
```
ghost: You can just inline these values and avoid the match like you do in the next test.

casper/src/main/resources/Registry.rho
```diff
@@ -0,0 +1,82 @@
...
+        `rho:lang:listOps` : `rho:id:6fzorimqngeedepkrizgiqms6zjt76zjeciktt1eifequy4osz35ks`,
+        `rho:lang:nonNegativeNumber` : `rho:id:hxyadh1ffypra47ry9mk6b8r1i33ar1w9wjsez4khfe9huzrfcytx9`,
+        `rho:rchain:authKey` : `rho:id:1qw5ehmq1x49dey4eadr1h4ncm361w3536asho7dr38iyookwcsp6i`,
+        `rho:rchain:lockbox` : `rho:id:w5zfefaasr5euzu1x551wwbgnt8o1c4345r3cgcwa1guyfre36fihu`,
+        `rho:rchain:makeMint` : `rho:id:asysrwfgzf8bf7sxkiowp4b3tcsy4f8ombi3w96ysox4u3qdmn1wbc`,
+        `rho:rchain:pos` : `rho:id:m3xk7h8r54dtqtwsrnxqzhe81baswey66nzw6m533nyd45ptyoybqr`,
+        `rho:rchain:revVault` : `rho:id:6zcfqnwnaqcwpeyuysx1rm48ndr6sgsbbgjuwf45i5nor3io7dr76j`
+      } {
+        shorthands => {
+          for (@map <<- _registryStore) {
```
ghost, line 31: Reading and writing to _registryStore scales quadratically with the number of contracts stored in the registry.  When bors is done with https://github.com/rchain/rchain/pull/2664, I'll submit that.  `rho:lang:treeHashMap` will give constant time for read/write from _registryStore and O(log(n)) lookup time given the map.

casper/src/main/resources/Registry.rho
```diff
@@ -0,0 +1,82 @@
...
+
+      match *value {
+        (nonce, data) => {
+          new uriCh, hashCh, verifyCh in {
+            blake2b256!((nonce, data), *hashCh) |
+            for (@hash <- hashCh) {
+              secpVerify!(hash, *sig, *pubKeyBytes, *verifyCh) |
+              for (@verified <- verifyCh) {
+                match verified {
+                  true => {
```
ArturGajowy: Match on boolean looks a bit strange. How about an `if`? Up to you though :)

casper/src/main/resources/Registry.rho
```diff
@@ -0,0 +1,82 @@
...
+                    ops!("buildUri", *pubKeyBytes, *uriCh) |
+                    for (@uri <- uriCh; @map <- _registryStore) {
+                      match map.get(uri) {
+                        Nil => { _registryStore!(map.set(uri, (nonce, data))) | ret!(uri) }
```
ArturGajowy: please add newlines after `{` and before `}` for each of the cases, and otherwise indendt the code (`if` below) in a standard mabber

rholang/src/main/scala/coop/rchain/rholang/interpreter/registry/Registry.scala
tymm: It deletes everything in Registry.scala except `buildURI` and a helper object for `buildURI`.

ArturGajowy: :+1: Sorry, I've missed that :)

casper/src/main/resources/Registry.rho
ghost: @odeac I took care of them here: https://github.com/rchain/rchain/pull/2689

casper/src/test/resources/RegistryTest.rho
tymm: I'm not sure if the first two tests make sense. Are you assuming here that insertion increases the nonce? It's the user who has to do this.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
tymm: Yep, you were right.

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
tymm: What about contracts that for some reason want to use `insertSigned` on-chain? E.g. a contract that wants to update itself. Sounds like a typical DAO usecase (?).

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RegistryBootstrap.scala
tymm: Done

rholang/src/main/scala/coop/rchain/rholang/interpreter/registry/RegistryBootstrap.scala
tymm: Done

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
tymm: Done

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
tymm: Done

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
tymm: Done; I had this discussion already with @odeac and thought I created a ticket for this already. But can't find it anymore and therefore creating a new one: https://rchain.atlassian.net/browse/RCHAIN-3801

casper/src/main/resources/Registry.rho
tymm: Done

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingPropertyTest.scala
tymm: Done

casper/src/test/resources/RegistryTest.rho
tymm: Done

casper/src/test/resources/RegistryTest.rho
tymm: Done

casper/src/main/resources/Registry.rho
tymm: Done

casper/src/test/resources/RegistryTest.rho
tymm: Done

casper/src/main/resources/Registry.rho
tymm: Done

casper/src/test/resources/RegistryTest.rho
tymm: Done

casper/src/test/resources/RegistryTest.rho
tymm: Done

casper/src/main/resources/Registry.rho
tymm: Done

casper/src/main/resources/Registry.rho
tymm: Done

casper/src/test/resources/RegistryTest.rho
ArturGajowy: How did the old registry behave?

I think we rely on the first two points for the security of system contracts. Their private keys used for deploy are hardcoded in our sources after all...

ArturGajowy (commented):
Almost there :) We just need to resolve the "nonce controversy". I'd also like to have a look at the final commit log due to the amount of fixups (I know we agreed to publish amendments this way, but I really never benefited from that and in this case it's getting unwieldy :)) - so please apply/squash the fixups :)

casper/src/main/resources/Registry.rho
ArturGajowy: Maybe it'd make sense as a follow-up PR? No preference from my side.

casper/src/test/resources/RegistryTest.rho
tymm: If you try to insert with a nonce bigger than MAX_LONG you'll get a `java.lang.NumberFormatException: For input string: "9223372036854775808"` from Rholang. Same should be true for the old registry.

ArturGajowy: ~~We should rather return Nil in the new registry.~~

Anyway, the more important question is: does the old registry enforces monotonic nonce?

If not, what are the consequences to system process security?

casper/src/main/resources/Registry.rho
tymm: I would prefer it as a follow-up PR.

casper/src/test/resources/RegistryTest.rho
tymm: > We should rather return Nil in the new registry.

Sounds good.
> does the old registry enforces monotonic nonce?

No, it enforces a strictly monotonic nonce. Trying to insert with a nonce thats `<=` the old nonce failed and fails with `Nil`.

ArturGajowy: I've read through the above and the code again - I've mistaken the properties I want guaranteed with the tests. You're right, those two lines make no sense :)

ArturGajowy (approved):
Love to see this happen! :heart:

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
ArturGajowy: I think it's insufficient for the DAO usecase because of the singature being public (and thus: not secure)

casper/src/main/resources/Registry.rho
ghost: ok

ghost (approved):

PR #2676, opened 2019-08-19
RCHAIN-3505: Error handling rholang

JosephDenman:
## Overview
A synchronous lock isn't used, but this is the ticket this approach grew from: <https://rchain.atlassian.net/browse/RCHAIN-3505?atlOrigin=eyJpIjoiZjE2YmNmNmU0NjY5NDllM2EzNWI1MzdmNjM0ZTBhNmYiLCJwIjoiaiJ9>

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -68,10 +68,12 @@ class DebruijnInterpreter[M[_], F[_]](
 )(
     implicit parallel: cats.Parallel[M, F],
     syncM: Sync[M],
-    fTell: FunctorTell[M, Throwable],
+    error: _error[M],
     cost: _cost[M],
     spanM: Span[M]
 ) extends Reduce[M] {
+
+  def dummy = error
 
   private[this] val injectSpanLabel                 = "inject"
   private[this] val parSpanLabel                    = "par"
@@ -185,7 +187,7 @@ class DebruijnInterpreter[M[_], F[_]](
       indexedTerms.parTraverse_ {
         case (term, index) =>
           val random = split(index)
-          reportErrors(eval(term)(env, random, sequenceNumber))
+          eval(term)(env, random, sequenceNumber)
       }
     } else {
       //TODO: Investigate if we can avoid the shuffling and manual parallelism limiting by tweaking:
@@ -202,7 +204,7 @@ class DebruijnInterpreter[M[_], F[_]](
         _.traverse_ {
           case (term, index) =>
             val random = split(index)
-            reportErrors(eval(term)(env, random, sequenceNumber))
+            eval(term)(env, random, sequenceNumber)
         }
       }
     }
@@ -228,12 +230,6 @@ class DebruijnInterpreter[M[_], F[_]](
           case other          => BugFoundError(s"Undefined term: \n $other").raiseError[M, Unit]
         }
       case other => BugFoundError(s"Undefined term: \n $other").raiseError[M, Unit]
-    }
-
-  private def reportErrors(process: M[Unit]): M[Unit] =
-    process.handleErrorWith {
-      case e @ OutOfPhlogistonsError => e.raiseError[M, Unit]
-      case e                         => fTell.tell(e)
     }
 
   override def inj(
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -31,10 +31,6 @@ import coop.rchain.rholang.interpreter.error_handling._
 import scala.collection.immutable.BitSet
 import scala.util.{Random, Try}
 
-object Reduce {
-  val parallelism = lang.Runtime.getRuntime.availableProcessors() * 2
-}
-
 /** Reduce is the interface for evaluating Rholang expressions.
   *
   * @tparam M The kind of Monad used for evaluation.
@@ -184,8 +180,12 @@ class DebruijnInterpreter[M[_], F[_]](
       else if (terms.size > 256) rand.splitShort(id.toShort)
       else rand.splitByte(id.toByte)
 
+    // The number of processors available to the JVM may change over the lifetime of an application, so the number of
+    // available processors needs to be calculated upon each evaluation of Par.
+    val parallelism = lang.Runtime.getRuntime.availableProcessors() * 2
+
     val indexedTerms = terms.zipWithIndex.toList
-    if (terms.size <= Reduce.parallelism) {
+    if (terms.size <= parallelism) {
       runAndReportErrors[M] {
         indexedTerms.parTraverse_ {
           case (term, index) =>
@@ -203,7 +203,7 @@ class DebruijnInterpreter[M[_], F[_]](
 
       // we index terms before shuffling to have deterministic Blake2b512Random seeds for each term
       val indexedTermsShuffled = Random.shuffle(indexedTerms)
-      val groups               = indexedTermsShuffled.groupBy(_._2 % Reduce.parallelism).values.toList
+      val groups               = indexedTermsShuffled.groupBy(_._2 % parallelism).values.toList
       runAndReportErrors[M] {
         groups.parTraverse_ {
           _.traverse_ {
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -4,7 +4,6 @@ import java.lang
 
 import cats.effect.Sync
 import cats.implicits._
-import cats.mtl.FunctorTell
 import cats.{Eval => _}
 import com.google.protobuf.ByteString
 import coop.rchain.crypto.codec.Base16
@@ -27,6 +26,7 @@ import coop.rchain.rspace.Serialize
 import coop.rchain.rspace.util._
 import monix.eval.Coeval
 import scalapb.GeneratedMessage
+import coop.rchain.rholang.interpreter.error_handling._
 
 import scala.collection.immutable.BitSet
 import scala.util.{Random, Try}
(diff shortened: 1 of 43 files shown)


JosephDenman:
bors try

JosephDenman:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/error_handling/package.scala
```diff
@@ -0,0 +1,25 @@
...
+  type _error[F[_]] = Ref[F, Option[Throwable]]
+
+  def runAndReportErrors[F[_]: MonadError[?[_], Throwable]](
+      process: F[Unit]
+  )(implicit error: _error[F]): F[Unit] =
+    error.get >>= {
+      case Some(_) => ().pure[F]
+      case None =>
+        process.handleErrorWith { currentError =>
+          error.update {
```
odeac, line 18: this is easier to read and more compact: ```error.update(_.getOrElse(currentError))```

rholang/src/main/scala/coop/rchain/rholang/interpreter/error_handling/package.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package object error_handling {
+
+  type _error[F[_]] = Ref[F, Option[Throwable]]
+
+  def runAndReportErrors[F[_]: MonadError[?[_], Throwable]](
+      process: F[Unit]
+  )(implicit error: _error[F]): F[Unit] =
+    error.get >>= {
+      case Some(_) => ().pure[F]
```
odeac, line 15: Instead of the match an if would be more readable:

```scala
error.get >>= (e =>
 if(e.isEmpty) ().pure[F]
else process.handleErrorWith {...} )
```

rholang/src/main/scala/coop/rchain/rholang/interpreter/error_handling/package.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.rholang.interpreter
+
+import cats.implicits._
+import cats.MonadError
+import cats.effect.concurrent.Ref
+
+package object error_handling {
+
+  type _error[F[_]] = Ref[F, Option[Throwable]]
```
odeac, line 9: To me it feels that error is not a good name. An "error" should be a value, not a mutable storage.

I would go with _errorHandler instead of _error. Or maybe _errorStore

ArturGajowy: The current name is fine to me. It won't matter though for reasons pointed out in my future comment.

ArturGajowy (changes requested):
Nice work! We need one more push (pun intended) or two before this is merged though.

1. The error ref and the `runAndReportErrors` method must be private to the DebriujinInterpreter, with the error created per call to `eval(Par)` and other entry points, and passed through the private methods of the reducer (implicit is fine for that). Otherwise it's just the global-state-ErrorLog in disguise, with less slots.
2. Reduction errors must be signalled by the Reducer via its effect type or its return type
3. We need to wait for the produce/consume-s in progress to finish before returning.
4. EvaluateResult should carry Option[Throwable] instead of a Vector

In effect:
 - the ```Replace `ErrorLog` dependency with `_error` dependency.``` will become ```Remove `ErrorLog```.
 - error_handling.ErrorHandling, and the error_handling packagbe object introduced in `Introduce error reference alias and combinators.` are going to disappear
 - the ```Move `errors` into `error_handling` package.``` will be redundant, and should be dropped

To make the above easier, it might make sense to:  
- Restrict the ChargingReducer interface to just `eval` and `inj`. `evalExpr` and `evalExprToPar` are only used in tests, and should be exposed to them via `private[rholang]`
- Restrict the Reduce interface to just `eval` and `inj`
- define the implementation of `inj` in the trait itself, or inline it whatsoever.

Make sure to remove the `ErrorLog` class, and any other code you're making unused.

The commits:
```
Have `fixture` take an optional error instead of `_error[Task]`. JosDenmark* 19.08.2019, 19:27
Eliminate `withTestSpace` from `PersistentStoreTester`. JosDenmark* 19.08.2019, 19:04
Use `fixture` instead of `withRegistryAndTestSpace` in `RegistrySpec`. JosDenmark* 19.08.2019, 19:03
Use `fixture` instead of `withTestSpace` in `ReduceSpec`. JosDenmark* 19.08.2019, 18:38
```
are a great improvement! They though should be squashed, as they edit the same places over and over again. Further, some the edited places still have a lot of duplicated structure, so it might make sense to push it a little more in there - and extract the whole repeated pattern at once instead of stripping it away bit by bit.  

~~```Make `parallelism` a local value.``` should be dropped, as there's no point in re-evaluating the `Reduce.parallelism` value over and over again. Feel free to make it private[rholang], or sth like that.~~

The `Fix bug in integration tests.` commit should be part of the commit that broke the tests.

ArturGajowy:
If possible, please prioritize the 'charge for system processes' work before coming back to this.

rholang/src/main/scala/coop/rchain/rholang/interpreter/error_handling/package.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.rholang.interpreter
+
+import cats.implicits._
+import cats.MonadError
+import cats.effect.concurrent.Ref
+
+package object error_handling {
```
lukasz-golebiewski, line 7: Why not introduce and nest the `handling` package under `error` and avoid the ugly underscore in the name?
I would even opt for getting rid of this package object completely and putting it's contents into the `error` package object

lukasz-golebiewski: Just took a peek at the next commit. I find the name `error_handling.errors` confusing

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -80,8 +81,8 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
     )
 
   "evalExpr" should "handle simple addition" in {
-    implicit val errorLog = new ErrorLog[Task]()
-    val result = withTestSpace(errorLog) {
+    implicit val error = ErrorHandling.emptyError[Task].runSyncUnsafe()
```
lukasz-golebiewski: Adding more `runSyncUnsafe`'s decreases readability. Is it a lot of effort to get rid of them?

Does the `val error` need to be implicit if it's passed in explicitly on the line below?

rholang/src/test/scala/coop/rchain/rholang/interpreter/CostAccountingReducerTest.scala
```diff
@@ -97,11 +97,12 @@ class CostAccountingReducerTest extends FlatSpec with Matchers with TripleEquals
...
     val send                    = Send(Par(exprs = Seq(GString("x"))), Seq(Par()))
-    val test                    = chargingReducer.inj(send).attempt.runSyncUnsafe(1.second)
-    assert(test === Left(OutOfPhlogistonsError))
+    chargingReducer.inj(send).attempt.runSyncUnsafe(1.second)
+    val runtimeError = error.get.runSyncUnsafe(1.second)
+    assert(runtimeError.contains(OutOfPhlogistonsError))
   }
 
-  it should "stop interpreter threads as soon as deploy runs out of phlo" ignore {
+  it should "stop interpreter threads as soon as deploy runs out of phlo" in {
```
lukasz-golebiewski, line 105: Yay!

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -184,8 +180,12 @@ class DebruijnInterpreter[M[_], F[_]](
       else if (terms.size > 256) rand.splitShort(id.toShort)
       else rand.splitByte(id.toByte)
 
+    // The number of processors available to the JVM may change over the lifetime of an application, so the number of
```
lukasz-golebiewski, line 183: Good catch!

lukasz-golebiewski:
> Make `parallelism` a local value. should be dropped, as there's no point in re-evaluating the Reduce.parallelism value over and over again.

@ArturGajowy It turns out that the value of available processors for the JVM may change over time so Joe's change is justified

ArturGajowy:
Yup, I missed the comment :+1: That's the best "well, actually" I've seen in a while!

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
JosephDenman: No. It does not need to be implicit. But the next few commits eliminate `error` in the tests, so the correction is unnecessary.

lukasz-golebiewski: What about the `runSyncUsafe`s?

JosephDenman: I see what you mean. I'll remove them.

rholang/src/main/scala/coop/rchain/rholang/interpreter/error_handling/package.scala
JosephDenman: It seems reasonable to keep the error classes and the error handler separate, no?

JosephDenman:
@ArturGajowy 

1. Basically, `error` must be in scope in `eval(par: Par)(...)` so that the error reference can be checked before the sub-processes are evaluated. This also prevents the duplication of `runAndReportErrors` in each of the evaluation methods. There are a few options:

- rewrite the `Reduce` interface so that `eval(par: Par)(...)` takes `error`.
- make a `create` method for `DebruijnInterpreter` that returns `F[DebruijnInterpreter[...]]`.
- keep the global dependency.

2. Fixed

3. This is enforced by the behavior of `parSequence_`. Previously, the evaluator used `raiseError` to cancel concurrent tasks - an inherently non-deterministic action. That is what was causing the produce/consume leakage before. We're no longer using `raiseError` to cancel concurrent tasks, so we sidestep those issues. This is easy to verify locally.

4.  Fixed

PR #2680, opened 2019-08-20
RCHAIN-3769: Refund for system processes

JosephDenman:
## Overview
Updates system processes so that they use `RhoPureSpace` (charging RSpace) instead of `RhoISpace`. One result is that programs like this:

```
new stdoutAck(`rho:io:stdoutAck`) in {
  stdoutAck!("DUPA!", 0) |
  for (_ <- @0) { 0 }
}
```
cost _less_ because now refunds are given for COMM events involving system processes.

Also contains some refactorings to RSpace and the Rholang evaluator.

### JIRA ticket:
<https://rchain.atlassian.net/browse/RCHAIN-3769>


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -43,7 +43,7 @@ trait Reduce[M[_]] {
   def eval(par: Par)(
       implicit env: Env[Par],
       rand: Blake2b512Random,
-      sequenceNumber: Int
+      sequenceNumber: Int = 0
```
ArturGajowy: Default arguments in production code lead to terrible bugs. Please drop this commit.

rholang/src/main/scala/coop/rchain/rholang/interpreter/ChargingReducer.scala
```diff
@@ -1,56 +0,0 @@
-package coop.rchain.rholang.interpreter
-import coop.rchain.crypto.hash.Blake2b512Random
-import coop.rchain.models.{Expr, Par}
-import coop.rchain.rholang.interpreter.accounting.{_cost, Cost, CostAccounting}
-
-// TODO: After refactoring Reduce to not include implicits in methods,
-//       make ChargingReducer extend Reduce.
-
-trait ChargingReducer[F[_]] {
```
ArturGajowy, line 9: Thank you! :clap:

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -380,7 +381,7 @@ object Runtime {
       (reducer, replayReducer) = {
         lazy val replayDispatchTable: RhoDispatchMap[F] =
           dispatchTableCreator(
-            replaySpace,
+            ChargingRSpace.pureRSpace(replaySpace),
```
ArturGajowy: Rather than creating the pureRspace ourselves, we should inject it from the outside. There should only be one call to `ChargingRSpace.pureRSpace` in prod code, and its output should find its way here.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/implicits.scala
```diff
@@ -27,7 +27,7 @@ object implicits {
       }
     }
 
-  def matchListPar[F[_]: Sync: Span]: StorageMatch[F, BindPattern, ListParWithRandom] =
+  implicit def matchListPar[F[_]: Sync: Span]: StorageMatch[F, BindPattern, ListParWithRandom] =
```
ArturGajowy: We should be instantiating `matchListPar` explicitly. Otherwise good luck with finding where it's being wired in, and how many instances there are. So please drop the `implicit`

ArturGajowy (changes requested):
~~Two~~ ~~three~~ It's four! Four things, superb otherwise! :ok_hand:

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
```diff
@@ -32,7 +27,7 @@ import coop.rchain.rholang.interpreter.accounting.{Cost, CostAccounting}
   * @param dispatcher the dispatcher
   */
 class ContractCall[F[_]: Concurrent: Span](
-    space: RhoISpace[F],
+    space: RhoPureSpace[F],
```
ArturGajowy: One more thing:
`Charge for storage in system processes.`
we'll need a test for that, probably in CostAccountingSpec.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosephDenman: There was an analogous default argument in `ChargingReducer`. Moving it here keeps the refactoring manageable. I can remove it, but a lot of useless zeros will pop up.

ArturGajowy: I see... Those zeros would pop up b/c of duplication in test code. Could you try and introduce test-util methods that don't require the sequence number, and still make the sequenceNumber non-default? Any usage of the default value in prod code we should investigate with high suspicion...

rholang/src/main/scala/coop/rchain/rholang/interpreter/ContractCall.scala
JosephDenman: Which ones? It seems redundant to test for each of them since it's basically the same pattern. One could also argue that testing this at all is redundant since the cases reduce to standard COMM events (which are already tested).

ArturGajowy: A single test that guard against regressions will suffice. So a test that exemplifies that produces done by system processes are charged for. 

>One could also argue that testing this at all is redundant since the cases reduce to standard COMM events (which are already tested).

One would be wrong to do so. The point is to make sure system processes are charged-for. They weren't after all.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosephDenman: I don't know what you mean by duplication in code. Zeros pop up whenever `eval` is called. This is an example:

```
val inspectTaskSendFirst = for {
          _   <- reducer.eval(send)(env, splitRand0)
          _   <- reducer.eval(receive)(env, splitRand1)
          res <- space.toMap
        } yield res
```

usage. There are a total of 208 errors when this default parameter is removed. Given that the default parameter being in `Reduce` leaves the code base in no worse shape than it was when the default parameter was in `ChargingReducer`, I think we should save that fight for another day. I'll leave a TODO.

ArturGajowy: 208 errors, of which 200 are in test code - which admittedly shouldn't be passing that parameter and use the default value, contrary to all prod usages. I call 'duplication' the duplicated call to the eval methods, most likely with the same type of assertion, done dozens of times.

A TODO is an improvement :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -61,7 +62,7 @@ trait Reduce[M[_]] {
 }
 
 class DebruijnInterpreter[M[_], F[_]](
-    pureRSpace: RhoPureSpace[M],
+    pureRSpace: RhoISpace[M],
```
ArturGajowy: please change the field name as well (to `tuplespace`)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -119,5 +118,29 @@ object ChargingRSpace {
       override def createCheckpoint(): F[Checkpoint]    = space.createCheckpoint()
       override def reset(hash: Blake2b256Hash): F[Unit] = space.reset(hash)
       override def close(): F[Unit]                     = space.close()
+
+      implicit override val m: StorageMatch[F, BindPattern, ListParWithRandom] = space.m
+
+      override def getData(channel: Par): F[Seq[internal.Datum[ListParWithRandom]]] =
```
ArturGajowy: ah, yes - those methods were properly hidden from rholang by the PureRSpace. I guess thought the indirection and the confusion it caused are not worth it, assuming we all agree those should not be used.

ArturGajowy (approved):
Thanks! :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
JosephDenman: Yeah. This could be ameliorated if `ISpace` was broken into a number of smaller traits.

lukasz-golebiewski: Instead of agreeing that this should not be used why not make sure this is the case by failing fast with `???` instead of delegating?

rspace/src/main/scala/coop/rchain/rspace/pure/PureRSpace.scala
```diff
@@ -1,72 +0,0 @@
-package coop.rchain.rspace.pure
-
-import cats.Id
-import cats.effect.Sync
-import coop.rchain.rspace.ISpace.IdISpace
-import coop.rchain.rspace._
-
-import scala.collection.SortedSet
-
-trait PureRSpace[F[_], C, P, A, K] {
```
lukasz-golebiewski, line 10: I love to see this go!

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: I've just checked, that the only methods rholang needs are: `install`, `consume`, `produce`, `close`. @lukasz-golebiewski @dzajkowski would you be OK with us introducing a `Tuplespace` trait containing them, that the ISpace trait would extend?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -41,6 +41,7 @@ object Reduce {
   */
 trait Reduce[M[_]] {
 
+  // TODO: Remove `sequenceNumber` default argument.
```
lukasz-golebiewski, line 44: This should be part of the first commit

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
lukasz-golebiewski: Sure thing

lukasz-golebiewski (approved):
Damn good stuff, sir!

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: @JosDenmark do you want to take care of it?

JosephDenman: Happy to.

PR #1501, opened 2018-09-19
RHOL-11: Short circuit spatial matcher when deployment runs out of phlo.

goral09:
## Overview
Short circuit spatial matcher when deployment runs out of phlo. Also, we now subtract phlos used from the initial available phlos.

### Which JIRA issue does this PR relate to? If there is not a JIRA issue addressing this work, please create one now and add the link here.
Add link to corresponding JIRA issue.

### Complete this checklist before you submit the PR
- [ ] This PR contains no more than 200 lines of code, excluding test code.
- [x] This PR meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards).
- [x] You assigned one person to review this PR

### If you are not a member of the core development team, please confirm:
- [ ] You signed the commit. Merging requires a signature. Please see the [developer wiki](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain) for instructions.
- [ ] Your GitHub account is also an account with [Travis CI](https://travis-ci.org). Unit tests will not run on your PR unless you have an account with Travis. Merge requires passed unit tests.

### Notes
Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else.

PR #2682, opened 2019-08-21
RCHAIN-3770 Make cost not rely on produce/consume order

ArturGajowy:
## Overview

### JIRA ticket:

https://rchain.atlassian.net/browse/RCHAIN-3770

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -12,6 +12,11 @@ import coop.rchain.rspace.{ContResult, Result, Match => StorageMatch}
 import scala.collection.SortedSet
 
 object ChargingRSpace {
+
+  private sealed trait TriggeredBy
+  private object Consume                                     extends TriggeredBy
+  private final case class Produce(datum: ListParWithRandom) extends TriggeredBy
+
   def storageCostConsume(
       channels: Seq[Par],
       patterns: Seq[BindPattern],
@@ -44,7 +49,10 @@ object ChargingRSpace {
         Option[(ContResult[Par, BindPattern, TaggedContinuation], Seq[Result[ListParWithRandom]])]
       ] =
         for {
-          _ <- charge[F](storageCostConsume(channels, patterns, continuation))
+          _ <- charge[F](
+                storageCostConsume(channels, patterns, continuation)
+                  .copy(operation = "consume storage")
+              )
           consRes <- space.consume(
                       channels,
                       patterns,
@@ -53,7 +61,7 @@ object ChargingRSpace {
                       sequenceNumber,
                       peeks
                     )
-          _ <- handleResult(consRes)
+          _ <- handleResult(consRes, Consume)
         } yield consRes
 
       override def install(
@@ -72,15 +80,16 @@ object ChargingRSpace {
         Option[(ContResult[Par, BindPattern, TaggedContinuation], Seq[Result[ListParWithRandom]])]
       ] =
         for {
-          _       <- charge[F](storageCostProduce(channel, data))
+          _       <- charge[F](storageCostProduce(channel, data).copy(operation = "produces storage"))
           prodRes <- space.produce(channel, data, persist, sequenceNumber)
-          _       <- handleResult(prodRes)
+          _       <- handleResult(prodRes, Produce(data))
         } yield prodRes
 
       private def handleResult(
           result: Option[
             (ContResult[Par, BindPattern, TaggedContinuation], Seq[Result[ListParWithRandom]])
-          ]
+          ],
+          triggeredBy: TriggeredBy
       ): F[Unit] =
         result match {
 
@@ -88,36 +97,39 @@ object ChargingRSpace {
 
           case Some((cont, dataList)) =>
             val refundForConsume =
-              if (cont.persistent) Cost(0)
+              if (cont.persistent && triggeredBy != Consume) Cost(0)
               else
                 storageCostConsume(cont.channels, cont.patterns, cont.value)
 
             val refundForProduces = refundForRemovingProduces(
               dataList,
-              cont.channels
+              cont.channels,
+              triggeredBy
             )
 
-            val refundValue = refundForConsume + refundForProduces
-
-            for {
-              _ <- if (refundValue == Cost(0))
-                    Sync[F].unit
-                  else charge[F](Cost(-refundValue.value, "storage refund"))
-            } yield ()
+            charge[F](Cost(-refundForConsume.value, "consume storage refund")) >>
+              charge[F](Cost(-refundForProduces.value, "produces storage refund"))
         }
 
       private def refundForRemovingProduces(
           dataList: Seq[Result[ListParWithRandom]],
-          channels: Seq[Par]
-      ): Cost =
-        dataList
+          channels: Seq[Par],
+          triggeredBy: TriggeredBy
+      ): Cost = {
+        val removedData = dataList
           .zip(channels)
-          .filterNot { case (data, _) => data.persistent }
+          // A persistent produce is charged for upfront before reaching the TS, and needs to be refunded
+          // after each iteration it matches an existing consume. We treat it as 'removed' on each such iteration.
+          .filterNot {
+            case (data, _) => data.persistent && triggeredBy != Produce(data.removedDatum)
+          }
+        removedData
           .map {
             case (data, channel) =>
-              storageCostProduce(channel, data.matchedDatum)
+              storageCostProduce(channel, data.removedDatum)
           }
           .foldLeft(Cost(0))(_ + _)
+      }
 
       override def close(): F[Unit] = space.close()
     }
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -40,20 +40,6 @@ object ChargingRSpace {
       case Empty => BugFoundError("Damn you pROTOBUF").raiseError[F, Blake2b512Random]
     }
 
-  def storageCostConsume(
-      channels: Seq[Par],
-      patterns: Seq[BindPattern],
-      continuation: TaggedContinuation
-  ): Cost = {
-    val bodyCost = Some(continuation).collect {
-      case TaggedContinuation(ParBody(ParWithRandom(body, _))) => body.storageCost
-    }
-    channels.storageCost + patterns.storageCost + bodyCost.getOrElse(Cost(0))
-  }
-
-  def storageCostProduce(channel: Par, data: ListParWithRandom): Cost =
-    channel.storageCost + data.pars.storageCost
-
   def chargingRSpace[F[_]: Sync: Span](
       space: RhoTuplespace[F]
   )(implicit cost: _cost[F]): RhoTuplespace[F] =
@@ -73,8 +59,9 @@ object ChargingRSpace {
       ] =
         for {
           _ <- charge[F](
-                storageCostConsume(channels, patterns, continuation)
-                  .copy(operation = "consume storage")
+                storageCostConsume(channels, patterns, continuation).copy(
+                  operation = "consume storage"
+                )
               )
           consRes <- space.consume(
                       channels,
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -5,17 +5,9 @@ import cats.implicits._
 import coop.rchain.metrics.Span
 import coop.rchain.models.TaggedContinuation.TaggedCont.ParBody
 import coop.rchain.models._
-import coop.rchain.rholang.interpreter.Runtime.{RhoISpace, RhoTuplespace}
+import coop.rchain.rholang.interpreter.Runtime.RhoTuplespace
 import coop.rchain.rholang.interpreter.accounting._
-import coop.rchain.rspace.{
-  Blake2b256Hash,
-  Checkpoint,
-  ContResult,
-  Result,
-  SoftCheckpoint,
-  internal,
-  Match => StorageMatch
-}
+import coop.rchain.rspace.{ContResult, Result, Match => StorageMatch}
 
 import scala.collection.SortedSet
 
@@ -123,7 +115,7 @@ object ChargingRSpace {
           .filterNot { case (data, _) => data.persistent }
           .map {
             case (data, channel) =>
-              storageCostProduce(channel, data.value)
+              storageCostProduce(channel, data.matchedDatum)
           }
           .foldLeft(Cost(0))(_ + _)
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -1,21 +1,44 @@
 package coop.rchain.rholang.interpreter.storage
 
+import java.nio.ByteBuffer
+
 import cats.effect.Sync
 import cats.implicits._
+import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.metrics.Span
-import coop.rchain.models.TaggedContinuation.TaggedCont.ParBody
+import coop.rchain.models.TaggedContinuation.TaggedCont.{Empty, ParBody, ScalaBodyRef}
 import coop.rchain.models._
 import coop.rchain.rholang.interpreter.Runtime.RhoTuplespace
 import coop.rchain.rholang.interpreter.accounting._
+import coop.rchain.rholang.interpreter.errors.BugFoundError
+import coop.rchain.rholang.interpreter.storage.ChargingRSpace.consumeId
 import coop.rchain.rspace.{ContResult, Result, Match => StorageMatch}
 
 import scala.collection.SortedSet
 
 object ChargingRSpace {
 
-  private sealed trait TriggeredBy
-  private object Consume                                     extends TriggeredBy
-  private final case class Produce(datum: ListParWithRandom) extends TriggeredBy
+  private sealed trait TriggeredBy {
+    val id: Blake2b512Random
+    val persistent: Boolean
+    val channelsCount: Int
+  }
+
+  private final case class Consume(id: Blake2b512Random, persistent: Boolean, channelsCount: Int)
+      extends TriggeredBy
+
+  private final case class Produce(id: Blake2b512Random, persistent: Boolean) extends TriggeredBy {
+    override val channelsCount = 1
+  }
+
+  private def consumeId[F[_]: Sync](continuation: TaggedContinuation): F[Blake2b512Random] =
+    //TODO: Make ScalaBodyRef-s have their own random state and merge it during its COMMs
+    continuation.taggedCont match {
+      case ParBody(value) => value.randomState.pure[F]
+      case ScalaBodyRef(value) =>
+        Blake2b512Random(ByteBuffer.allocate(8).putLong(value).array()).pure[F]
+      case Empty => BugFoundError("Damn you pROTOBUF").raiseError[F, Blake2b512Random]
+    }
 
   def storageCostConsume(
       channels: Seq[Par],
@@ -61,7 +84,8 @@ object ChargingRSpace {
                       sequenceNumber,
                       peeks
                     )
-          _ <- handleResult(consRes, Consume)
+          id <- consumeId(continuation)
+          _  <- handleResult(consRes, Consume(id, persist, channels.size))
         } yield consRes
 
       override def install(
@@ -82,7 +106,7 @@ object ChargingRSpace {
         for {
           _       <- charge[F](storageCostProduce(channel, data).copy(operation = "produces storage"))
           prodRes <- space.produce(channel, data, persist, sequenceNumber)
-          _       <- handleResult(prodRes, Produce(data))
+          _       <- handleResult(prodRes, Produce(data.randomState, persist))
         } yield prodRes
 
       private def handleResult(
@@ -93,22 +117,27 @@ object ChargingRSpace {
       ): F[Unit] =
         result match {
 
-          case None => Sync[F].unit
+          case None => charge[F](eventStorageCost(triggeredBy.channelsCount))
 
           case Some((cont, dataList)) =>
-            val refundForConsume =
-              if (cont.persistent && triggeredBy != Consume) Cost(0)
-              else
+            for {
+              id <- consumeId(cont.value)
+
+              // We refund for non-persistent continuations, and for the persistent continuation triggering the comm.
+              // That persistent continuation is going to be charged for (without refund) once it has no matches in TS.
+              refundForConsume = if (!cont.persistent || id == triggeredBy.id) {
                 storageCostConsume(cont.channels, cont.patterns, cont.value)
+              } else {
+                Cost(0)
+              }
+              refundForProduces = refundForRemovingProduces(dataList, cont.channels, triggeredBy)
 
-            val refundForProduces = refundForRemovingProduces(
-              dataList,
-              cont.channels,
-              triggeredBy
-            )
-
-            charge[F](Cost(-refundForConsume.value, "consume storage refund")) >>
-              charge[F](Cost(-refundForProduces.value, "produces storage refund"))
+              _ <- charge[F](Cost(-refundForConsume.value, "consume storage refund"))
+              _ <- charge[F](Cost(-refundForProduces.value, "produces storage refund"))
+              _ <- charge[F](eventStorageCost(triggeredBy.channelsCount))
+                    .unlessA(triggeredBy.persistent)
+              _ <- charge[F](commEventStorageCost(cont.channels.size))
+            } yield ()
         }
 
       private def refundForRemovingProduces(
@@ -120,14 +149,12 @@ object ChargingRSpace {
           .zip(channels)
           // A persistent produce is charged for upfront before reaching the TS, and needs to be refunded
           // after each iteration it matches an existing consume. We treat it as 'removed' on each such iteration.
-          .filterNot {
-            case (data, _) => data.persistent && triggeredBy != Produce(data.removedDatum)
+          // It is going to be 'not removed' and charged for on the last iteration, where it doesn't match anything.
+          .filter {
+            case (data, _) => !data.persistent || data.removedDatum.randomState == triggeredBy.id
           }
         removedData
-          .map {
-            case (data, channel) =>
-              storageCostProduce(channel, data.removedDatum)
-          }
+          .map { case (data, channel) => storageCostProduce(channel, data.removedDatum) }
           .foldLeft(Cost(0))(_ + _)
       }
 
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
@@ -108,12 +108,12 @@ object ChargingRSpace {
 
           case Some((cont, dataList)) =>
             for {
-              id <- consumeId(cont.value)
+              id <- consumeId(cont.continuation)
 
               // We refund for non-persistent continuations, and for the persistent continuation triggering the comm.
               // That persistent continuation is going to be charged for (without refund) once it has no matches in TS.
               refundForConsume = if (!cont.persistent || id == triggeredBy.id) {
-                storageCostConsume(cont.channels, cont.patterns, cont.value)
+                storageCostConsume(cont.channels, cont.patterns, cont.continuation)
               } else {
                 Cost(0)
               }
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
@@ -52,11 +52,10 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
 
   val contracts = Table(
     ("contract", "expectedTotalCost"),
-    ("""@0!(2)""", 33L),
-    ("""@0!(2) | @1!(1)""", 69L),
-    ("""for(x <- @0){ Nil }""", 64L),
-    ("""for(x <- @0){ Nil } | @0!(2)""", 73L),
-    /*
+    ("""@0!(2)""", 97L),
+    ("""@0!(2) | @1!(1)""", 197L),
+    ("""for(x <- @0){ Nil }""", 128L),
+    ("""for(x <- @0){ Nil } | @0!(2)""", 329L),
     ("""new loop in {
          contract loop(@n) = {
            match n {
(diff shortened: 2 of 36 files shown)


ArturGajowy:
The casper tests are spinning endlessly, because we test that `C!! C!!` doesn't conflict. And it doesn't, technically... It's just that the cost model now makes `for (_ <= @0) { 0 } | @0!!(0)` loop forever, because there's no other cost to COMM than the refund for its event pair. I'll change that.

(fixed the example in the first line as pointed out by Łukasz. Thx!)

lukasz-golebiewski:
`CX !!X` does conflict because `C` and `!!` could match each other

lukasz-golebiewski:
`(C !!)` does not (in the same block) because such deploy will be marked as invalid and it's changes on the TS are going to be reverted

rspace/src/main/scala/coop/rchain/rspace/ISpace.scala
```diff
@@ -5,7 +5,7 @@ import coop.rchain.rspace.internal._
 
 import scala.collection.SortedSet
 
-final case class Result[A](value: A, persistent: Boolean)
+final case class Result[A](matchedDatum: A, persistent: Boolean)
```
JosephDenman: While you're at it, please rename `value` in `ContResult` to `continuation` and make the signature `ContResult[C, P, K]`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -89,6 +105,31 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+        assert(result._1.errors.isEmpty)
+        result
+      }
+    }
+
+  def checkRepeatableCost(block: => (EvaluateResult, Chain[Cost])): Unit = {
+    val repetitions = 20
+    val first       = block
+    // execute in parallel to trigger different interleaves
+    val subsequents = (1 to repetitions).par.map(_ => block).toList
```
JosephDenman, line 123: What's the point of parallelizing this? This creates 20 different runtimes and never writes to disk. Removing the par also prevents the weird errors you're getting in the tests from materializing.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -89,6 +105,31 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    // check assertions sequentially to avoid "suppressed exceptions" output on assertion failure
+    subsequents.foreach { subsequent =>
+      val expected = subsequent._1.cost.value
```
JosephDenman: Nit: I'd reverse the names `expected` and `actual`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -56,7 +56,24 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    ("@0!!(0) | for (@0 <- @0) { 0 }", 80L),
+    ("@0!(0) | for (_ <- @0) { 0 }", 77L),
+    ("@0!(0) | for (x <- @0) { 0 }", 77L),
+    ("@0!(0) | for (@0 <- @0) { 0 }", 71L),
+    ("@0!(0) | for (_ <= @0) { 0 }", 98L),
+    ("@0!(0) | for (x <= @0) { 0 }", 100L),
+    ("@0!(0) | for (@0 <= @0) { 0 }", 85L),
+    ("@0!(0) | for (@0 <- @0) { 0 } | @0!(0) | for (_ <- @0) { 0 }", 151L),
+    ("@0!(0) | for (@0 <- @0) { 0 } | @0!(0) | for (@1 <- @0) { 0 }", 167L),
+    /* TODO This should work once we start charging for system processes
```
JosephDenman: This first TODO can be removed.

JosephDenman (changes requested):
I like your solution to produce/consume non-determinism. A few corrections, then LGTM.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: To make it run faster. Also, it's easier to observe cost non-determinism that way (the 'keccak' test didn't fail for me even with 100 repetitions without the `.par`)

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: So would I :D

rspace/src/main/scala/coop/rchain/rspace/internal.scala
```diff
@@ -23,7 +23,12 @@ object internal {
       Datum(a, persist, Produce.create(channel, a, persist, sequenceNumber))
   }
 
-  final case class DataCandidate[C, A](channel: C, datum: Datum[A], datumIndex: Int)
+  final case class DataCandidate[C, A](
+      channel: C,
+      datum: Datum[A],
+      removedDatum: A,
```
lukasz-golebiewski, line 29: A hunch tells me that this field is redundant. I cannot check this because tests are currently failing. Will get back to reviewing once this is sorted out

ArturGajowy: You can believe me that the distinction between `removed` and `matched` is part of the non-symmetric costs problem.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
JosephDenman: I get the performance argument, but not the non-determinism argument. Each run uses a different hot store, so why is it easier to observe non-determinism?

ArturGajowy: my observation was the test doesn't catch cost nondeterminism otherwise. my intution is having more tasks than cores is needed for non-as-if-sequential interleaves to happen.

rspace/src/main/scala/coop/rchain/rspace/ISpace.scala
ArturGajowy: Done

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: Done

rspace/src/main/scala/coop/rchain/rspace/internal.scala
ArturGajowy: @lukasz-golebiewski the tests are now passing, please have another stab at it

ArturGajowy: You can verify the need for removed datum by using the test I've added in CostAccountingSpec in https://github.com/rchain/rchain/pull/2682/commits/763bfd35cabd1293a11467879a1ae22fb1649435

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: Done

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -52,11 +52,22 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
 
   val contracts = Table(
     ("contract", "expectedTotalCost"),
-    ("""@0!(2)""", 33L),
-    ("""@0!(2) | @1!(1)""", 69L),
-    ("""for(x <- @0){ Nil }""", 64L),
-    ("""for(x <- @0){ Nil } | @0!(2)""", 73L),
-    /*
+    ("""@0!(2)""", 97L),
```
lukasz-golebiewski, line 55: I also got this

```
[info] - should be repeatable *** FAILED *** (25 seconds, 323 milliseconds)               
[info]   TimeoutException was thrown during property evaluation. (CostAccountingSpec.scala:110)   
[info]     Message: Task.runSyncUnsafe(25 seconds)                                             
[info]     Occurred at table row 0 (zero based, not counting headings), which had values (                                                                                                                 
[info]       contract = @0!(2),                                               
[info]       expectedTotalCost = 97                                                                                                                                                                        
[info]     )       
```
:/
Not sure if this is related

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -52,11 +52,22 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    ("""@0!(2) | @1!(1)""", 197L),
+    ("""for(x <- @0){ Nil }""", 128L),
+    ("""for(x <- @0){ Nil } | @0!(2)""", 329L),
+    ("@0!!(0) | for (_ <- @0) { 0 }", 342L),
+    ("@0!!(0) | for (x <- @0) { 0 }", 342L),
+    ("@0!!(0) | for (@0 <- @0) { 0 }", 336L),
```
lukasz-golebiewski, line 61: I added `("@0!!(0) | @1!!(1) | for (_ <- @0;_ <- @1) { 0 }", 586L)` and it looks to me like it's failing non-deterministically. Sometimes it expects 586, other times 594. Shouldn't this also be solved here?

```Cost was not repeatable, expected 594, got 586.```
```Cost was not repeatable, expected 586, got 594```

**Review by lukasz-golebiewski — CHANGES_REQUESTED** [review]
Found some errors which I think should be corrected before this gets in

File: rspace/src/main/scala/coop/rchain/rspace/internal.scala (hunk 5 above)
Comment (lukasz-golebiewski) [reply rspace/src/main/scala/coop/rchain/rspace/internal.scala:29]: Did it, thanks!

File: rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala (hunk 7 above)
Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: How do you run your tests? sbt? Please paste the exact command so that there's less variables when I'm trying to reproduce that.

Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: As you've seen, the CI passed - and so did my local runs. Thx for catching those!

**Comment by ArturGajowy** [comment]
bors try

File: rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala (hunk 7 above)
Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: Can you paste the full assertion output? Or even better a link to gist.github.com?

Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: Oh, that's a new test case! How did you arrive at it? Maybe you've got some intuitions what could be wrong that you used when creating it?

Comment (lukasz-golebiewski) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: I ran `sbt:rholang> testOnly *CostAccounting*`. Going to add a link to the gist in a moment

Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: Got it locally, thx. Any ideas how to test that better?

Comment (lukasz-golebiewski) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: https://gist.github.com/lukasz-golebiewski/151c65891b7ec3e8f9c14698e013cbe0

Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:61]: @lukasz-golebiewski please have another look. Unless you can find any more bugs, I'd like to split additional testing (possibly generative) for cost repeatability into a separate PR.

**Comment by ArturGajowy** [comment]
I've re-done the follwoing commits:
`Make the cost not rely on produce/consume order`
`Add cost accounting for event storage`

File: rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala (hunk 6 above)
Comment (ArturGajowy) [reply rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala:55]: This seems to be the general stability problem with these tests...

**Comment by ArturGajowy** [comment]
Fixed the wart remover errors. My bloop-based setup doesn't run wart remover, hence the errors.

**Comment by ArturGajowy** [comment]
bors try

File: rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala [hunk 8]
```diff
@@ -1,38 +1,44 @@
...
   }
 
-  def storageCostProduce(channel: Par, data: ListParWithRandom): Cost =
-    channel.storageCost + data.pars.storageCost
+  private def consumeId[F[_]: Sync](continuation: TaggedContinuation): F[Blake2b512Random] =
+    //TODO: Make ScalaBodyRef-s have their own random state and merge it during its COMMs
+    continuation.taggedCont match {
+      case ParBody(value) => value.randomState.pure[F]
+      case ScalaBodyRef(value) =>
+        Blake2b512Random(ByteBuffer.allocate(8).putLong(value).array()).pure[F]
```
ArturGajowy, line 39: @pyrocto It occured to me that Scala continuations (so: all system processes) don't have their own random state, and the dispatcher doesn't do a merge when COMM-ing with them. Do you see an issue with that?

My default reaction would be to make it uniform between rholang- and scala- continuations, and make both of them have a "randomState", and have that merged on each COMM (so that COMM always means a random state merge). Do you see reasons not to do that?

ArturGajowy: Relevant part of dispatcher code:
https://github.com/rchain/rchain/blob/9676e49488da094428e0fdf8c312f8d8f94db221/rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala#L29-L50

ArturGajowy:
bors try

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ghost: > make it uniform between rholang- and scala- continuations, and make both of them have a "randomState", and have that merged on each COMM (so that COMM always means a random state merge).

I agree.

> Do you see reasons not to do that?

On the contrary, I think uniformity is the only way you'll get consistency across validators.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -52,11 +52,23 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+    ("@0!!(0) | @1!!(1) | for (_ <- @0;_ <- @1) { 0 }", 594L),
+    ("@0!(0) | for (_ <- @0) { 0 }", 333L),
+    ("@0!(0) | for (x <- @0) { 0 }", 333L),
+    ("@0!(0) | for (@0 <- @0) { 0 }", 327L),
+    ("@0!(0) | for (_ <= @0) { 0 }", 354L),
+    ("@0!(0) | for (x <= @0) { 0 }", 356L),
+    ("@0!(0) | for (@0 <= @0) { 0 }", 341L),
+    ("@0!(0) | @0!(0) | for (_ <= @0) { 0 }", 574L),
+    ("@0!(0) | for (@0 <- @0) { 0 } | @0!(0) | for (_ <- @0) { 0 }", 663L),
+    ("@0!(0) | for (@0 <- @0) { 0 } | @0!(0) | for (@1 <- @0) { 0 }", 551L),
```
lukasz-golebiewski, line 72: I would add a few peek tests just to make sure those are also deterministic

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Sounds good!

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Just noticed that this was a timeout, most probably due to a larger amount of tests executed in parallel. If this fails from time to time on my machine then it definitely is going to fail in CI sooner or later. Better increase the timeout just in case

ArturGajowy: I tried locally with both bigger and smaller timeouts, without visible difference... I'm slightly against increasing the timeout, because most likely 25s per test case is far more than needed, and anything that hangs for 10s will most likely hang for 50s too - so it'd only make local hiccups more cumbersome. Unless you insist? :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: Check out the cost log for `("@0!(0) | for (@0 <<- @0) { 0 }", 328L)`:
```
List(Cost(30,parsing), Cost(11,send eval), Cost(11,receive eval), Cost(4,substitution), Cost(4,substitution), Cost(4,substitution), Cost(4,substitution), Cost(8,produces storage), Cost(4,substitution), Cost(14,consume storage), Cost(64,event storage cost), Cost(-14,consume storage refund), Cost(-8,produces storage refund), Cost(64,event storage cost), Cost(128,comm event storage cost))
```

The produce should not have been refunded for because peek isn't going to remove it from the tuplespace

ArturGajowy: Great catch! :+1:

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
lukasz-golebiewski: I don't insist, I just think this is a wise thing to do. `checkRepeatableCost` executes many tasks in parallel so some of them might get starved for quite a long while. I don't mind deferring the timeout increase until (if at all) it has become a nuisance in CI

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -93,22 +117,27 @@ object ChargingRSpace {
...
+              refundForConsume = if (!cont.persistent || id == triggeredBy.id) {
+                storageCostConsume(cont.channels, cont.patterns, cont.value)
+              } else {
+                Cost(0)
+              }
+              refundForProduces = refundForRemovingProduces(dataList, cont.channels, triggeredBy)
+
+              _ <- charge[F](Cost(-refundForConsume.value, "consume storage refund"))
+              _ <- charge[F](Cost(-refundForProduces.value, "produces storage refund"))
+              _ <- charge[F](eventStorageCost(triggeredBy.channelsCount))
```
JosephDenman, line 124: As I was saying in standup, the event log is emptied upon each call to `createCheckpoint`, which is done after each batch of deploys is processed. Why are you charging for this?

JosephDenman (dismissed):
I'm not convinced that charging for RSpace event storage is a sensible thing to do. For one thing, the event log is emptied each time `createCheckpoint` is called, and `createCheckpoint` is called after each set of deploys is evaluated. It doesn't make sense to charge a non-refundable fee for data that will certainly be removed. Doing so also pollutes the semantic mapping from source code to the cost of the program the source code defines, making it more difficult for users to see why their program costs what it does.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ghost: What happens with multiple persistent sends or mixed sends?  E.g. `@0!!(0) | @0!!(0) | for (_ <- @0) { 0 }` or `@0!!(0) | @0!(0) | for (_ <- @0) { 0 }`.

Are you going to refund storage fees for stuff that can be garbage collected?

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: > It doesn't make sense to charge a non-refundable fee for data that will certainly be removed

The event log is going to be stored in the BlockMessage, and will never be removed. At least not until we implement finalization and dag pruning based on it.

> difficult for users to see why their program costs what it does.

It's actually easier to estimate than with tuplespace storage: you pay 128 phlo per COMM, and 64 per produce/consume (assuming single-channel consumes).

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ghost: Have any major changes been made to the cost accounting in the spatial matcher since https://github.com/rchain/rchain/pull/1501 ?  If so, it's worth testing that cost accounting there is also deterministic.

ArturGajowy: I'll add the cases you proposed and make sure they're deterministic.

Garbage collection is out of scope for Mercury, so refunds based on it(s possibility) should be too, I think.

We're no longer charging for matching since the merge of #2637, as pointed out in one of the commit messages.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
JosephDenman: > The event log is going to be stored in the BlockMessage, and will never be removed.

If you're in favor of charging for the space occupied by deploy information, then you must also be in favor of charging for space occupied by the source code and all other information stored in `ProcessedDeploy`.

I see the problem you're trying to solve, but I think a better solution is to attach an execution cost to produces and consumes. Firstly, programs like `x!!(Q) | for(_ <= x){ P }` would actually terminate. Secondly, since there would always be a cost associated with produces and consumes (which are the only ways to add entries to the event logs), agents would not be able to create arbitrary event log entries without recourse.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: `@0!!(0) | @0!!(0) | for (_ <- @0) { 0 }` passes, I've added it to the test suite

`@0!!(0) | @0!(0) | for (_ <- @0) { 0 }` can have different costs due to the execution result itself being non-deterministic. We need a different test for that, one that uses replay and an event log. I'd like to have it part of the next PR.

ArturGajowy: @pyrocto please see the above :)

@lukasz-golebiewski I attempted fixing this, the fix is blocked by the bug I mentioned in https://discordapp.com/channels/391435113415573504/406508474261045250/615482010747666443. Let's pull the peek tests into the next PR, or the one coming after the bug fix.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
```diff
@@ -80,52 +91,59 @@ object ChargingRSpace {
...
 
       private def refundForRemovingProduces(
           dataList: Seq[Result[ListParWithRandom]],
-          channels: Seq[Par]
-      ): Cost =
-        dataList
+          channels: Seq[Par],
+          triggeredBy: TriggeredBy
+      ): Cost = {
+        val removedData = dataList
```
ArturGajowy, line 135: Refund for produces should be 0 when it's triggered by a peek.
Deterministic peek cost can't be achieved and tested though unit the bug I've mentioned in the other comment is fixed (due to event storage costs being wrong).

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: Created https://rchain.atlassian.net/browse/RCHAIN-3791

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: > then you must also be in favor of charging for space occupied by the source code and all other information stored in ProcessedDeploy.

At least the part that has size dependent on the deploy. I think we're doing so, but feel free to inspect that and create relevant JIRA tickets.

> a better solution is to attach an execution cost to produces and consumes

Any storage cost also doubles as execution cost. One has to process the data to store it.

> programs like x!!(Q) | for(_ <= x){ P } would actually terminate

They do under the current approach, so this is not a differentiator.

> agents would not be able to create arbitrary event log entries without recourse.

This is also true under the current approach, as adding arbitrary log entries will cause replay failures.

The current approach has a benefit of being more directly bound to a physical measure of cost (byte size) vs using an arbitrary iteration cost. The latter would miss the fact that iteration cost is different for a multi-channel receive than for a single-channel one. So: event-storage-size is more accurate.

ghost: I agree with Artur that it makes sense to charge for it, for reasons discussed here:
https://discordapp.com/channels/391435113415573504/420661762828926978/614683476410761237

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ghost: It's interesting how many of the equations in pi calculus do not hold in rholang.  In pi calculus, `new x in { new x in { P } }` is equal, as a process, to `new x in { P }`,  `new x in { new y in { P } }` is equal to `new y in { new x in { P } }`, `new x in { Nil }` equals `Nil`, `{ new x in { P } } | Q` equals `new x in { P | Q }` when x is not free in `Q`, and most relevant to this point, `!P` (replication) is equal to `!P | P`, which is why I suggested it as a test case.  If those are not equal, then yes, they will have different accounting.

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
JosephDenman: @pyrocto This is the statement from the discord discussions I disagree with:

> Otherwise, our cost model (which assumes symmetric charge and refund for TS storage, and only now is actually going to implement that correctly) will never terminate an infinite loop.

There should be a time-cost associated with ephemeral produces and consumes, and the time-cost associated with persistent produces and consumes should be that time-cost iterated. As it stands, the time-cost of doing persistent produces and consumes is unaccounted for. If it were (and it certainly should be if we are at all attempting to have cost reflect complexity), then infinite loops would be disallowed.

> Any storage cost also doubles as execution cost. One has to process the data to store it.

It sounds like you're proposing that we do not differentiate space complexity and time complexity, and that is absolutely wrong. The time-complexity of produces and consumes is not equivalent to their space-complexity. It takes the same amount of time to add a large amount of data to the hot store as it does to add a small amount of data. You said this yourself during Friday standup. So, this argument is invalid.

> The current approach has a benefit of being more directly bound to a physical measure of cost (byte size) vs using an arbitrary iteration cost.

Are you calling time-costs arbitrary? You must be because, again, the time-cost of a persistent produce/consume will be the iterated time-cost of an ephemeral produce/consume. So, this argument is invalid.

JosephDenman: Regarding using cost accounting to compensate validators for system storage, blocks are stored by all validators, and that would have to be factored into the cost calculation. If you're saying that the cost of a byte is one phlo, then the true price of a byte should be the number of currently bonded validators.

JosephDenman: In general, your proposal to use cost accounting to compensate validators for resources consumed as a result of our protocol opens the door to a large investigation into places where the protocol relies on accumulated state. For example, `CasperState` has a field `deployHistory` that is a sequence of previous deploys. ATM, it only shrinks when a block is finalized. Why would you not also charge users for the memory consumed by adding their deploy to `deployHistory` and refund upon finalization?

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
ArturGajowy: I'm not sure what you mean by "`!P` replication". Having the same process twice?

If yes, notice the `@0!!(0)`-s below are different processes
`@0!!(0) | @0!!(0) | for (_ <- @0) { 0 }`, `@0!!(0)` is always different than `@0!(0)`, and in all cases the `!0` events are different b/c they come from different namespaces. I feel I still don't get your point though? :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/storage/ChargingRSpace.scala
ArturGajowy: I'll reiterate:

Any storage cost also doubles as execution cost. One has to process the data to store it. 

The size-based (execution/storage)-cost is more accurate than adding a constant on each iteration, because it's less detached from actual work that needs to be done.

> not differentiate space complexity and time complexity, and that is absolutely wrong

It is, and that's what we have currently with TS storage costs. Changing that is out of this PR's scope. Any changes to that would need to acknowledge both 'storage cost as CPU cost' and 'storage cost per se', so it makes sense to use event storage cost for now.

> The time-complexity of produces and consumes is not equivalent to their space-complexity

:+1: At the same time, any space-cost not refunded for during the deploy also implies a time-cost (serialization).

> the time-cost of a persistent produce/consume will be the iterated time-cost of an ephemeral produce/consume

The time-cost of (whatever) produce/consume is lower-bounded by its event storage cost. I've not seen a more precise proposal from you, and anything correct you could propose would need to account for that. So feel free to file a JIRA describing the intended improvement.

> Regarding using cost accounting to compensate validators for system storage (...)

This whole section applies equally well to the TS storage costs we have now, and improving that is also out of scope of this PR.

> In general, your proposal to use cost accounting to compensate validators for resources consumed as a result of our protocol opens the door to a large investigation (...)

which is out of scope of this PR.

I consider this resolved. You seem to not acknowledge this, but the event storage costs are doing what you've asked for, and more.

lukasz-golebiewski (approved):

ghost (approved):

ArturGajowy:
Bors r+

PR #2684, opened 2019-08-23
RCHAIN-3505: Error handling rholang

JosephDenman:
## Overview
A synchronous lock isn't used, but this is the ticket this approach grew from: <https://rchain.atlassian.net/browse/RCHAIN-3505?atlOrigin=eyJpIjoiZjE2YmNmNmU0NjY5NDllM2EzNWI1MzdmNjM0ZTBhNmYiLCJwIjoiaiJ9>


odeac (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -82,6 +78,15 @@ class DebruijnInterpreter[M[_], F[_]](
 
   type Application = Option[(TaggedContinuation, Seq[ListParWithRandom], Int, Boolean)]
 
+  private def runAndReportError(process: M[Unit]): M[Unit] =
+    errorRef.get >>= {
```
lukasz-golebiewski, line 82: Since `Ref` doesn't give means of synchronization I can imagine cases where identical paths of Rholang execution raise different errors. Using `MVar` for synchronization on the other hand would most probably slow reduction down dramatically. I assume we can live in a world where errors are not necessarily raised deterministically?

ArturGajowy: I'm stil not convinced that this is enough to prevent tuplespace pollution. I also disagree with the claim that performance would suffer - this is easy to benchmark, and should be.

I'll have a look at the PR as soon as I can.

JosephDenman: This error model will not work because it’s not replay-able. If there are two processes in parallel composition, both of which throw errors, the order in which they’re run will determine which error is reported. That’s why the reducer didn’t cancel all threads when one erred. This is true regardless of whether we synchronize around the error reference.

JosephDenman:
I'm closing this ticket because this approach does not solve replay-ability (see this comment <https://github.com/rchain/rchain/pull/2684#discussion_r317301144>). I have a feeling that there is a pre-existing problem with replay-ability and error-induced process cancellation. Specifically, the only error that results in the cancellation of a concurrent process is the `OutOfPhlogistonError`. Cancellation is triggered by `raiseError`, and this function is inherently non-deterministic. There's no way to guarantee that a program will be cancelled the same way twice. We probably haven't witnessed that behavior because there's no test for it. I'll write a test and report my findings.

PR #2690, opened 2019-08-28
Move some contracts around so only one copy is ever made.  

ghost:
## Overview
Some contracts were declared inside others, so a new copy of the contract was made every time the containing one was called.  Since there's no garbage collection yet, that's a memory leak.  Also, there was a self-regenerating loop that I turned into a contract to minimize interaction with rspace.  I didn't get all of them, because there were lots of places where combinatorAndCondition closes over variables.

ListOps should probably be modified to include a partialFold where instead of passing in a combinatorAndCondition that closes over state, you pass in a generic one and explicitly pass the state that used to be closed over.  We need some benchmarks to tell how urgent such a refactoring is.

### DIFF
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -164,19 +164,20 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
-    // Composes a list of functions returning Either into a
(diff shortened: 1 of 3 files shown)


ArturGajowy (approved):
Thx for spotting and fixing those! :+1:

lukasz-golebiewski (approved):

PR #2691, opened 2019-08-28
Rchain 3755: Fix genesis block approval protocol

odeac:
## Overview
The bootstrap node waits for a minimum block-approval-duration and a minimum number of signatures before it sends the ApprovedBlock to the validators. 

On the validator side each validator verifies that ApprovedBlock has the required number of signatures AND that his own signature is present in the ApprovedBlock. 

If some validator cannot validate the genesis block and send his block approval in time before the block-approval-duration expires then his signature won’t appear in the block signatures and he will consider the block to be invalid.

The fix that I discussed with @woky is to remove from the validator condition for valid block the part that he requires his own signature to be present because this condition contradicts the idea that we have a valid genesis block with just the minimum required number of valid signatures.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3755

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
@@ -117,10 +117,23 @@ object ApproveBlockProtocol {
                            (newSigs, (sigs, newSigs))
                          })
           (before, after) = modifyResult
+
           _ <- if (after > before)
-                Metrics[F].incrementCounter("genesis")
-              else ().pure[F]
+                Log[F].info("APPROVAL: New signature received") >>
+                  Metrics[F].incrementCounter("genesis")
+              else
+                Log[F].info("APPROVAL: No new sigs received")
+
           _ <- Log[F].info(s"APPROVAL: received block approval from $sender")
+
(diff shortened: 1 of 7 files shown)


EncodePanda (approved):
Good catch

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
```diff
@@ -112,13 +112,28 @@ object ApproveBlockProtocol {
...
+                Log[F].info("APPROVAL: New signature received") >>
+                  Metrics[F].incrementCounter("genesis")
+              else
+                Log[F].info("APPROVAL: No new sigs received")
+
           _ <- Log[F].info(s"APPROVAL: received block approval from $sender")
+
+          _ <- Log[F].info(
+                s"APPROVAL: ${after.size} approvals received: ${after
+                  .map(s => "<" + Base16.encode(s.publicKey.toByteArray).substring(0, 10) + "...>")
```
KentShikama: You could use the PrettyPrinter (.buildString I think) which I believe already does this truncating

KentShikama (approved):
Good catch!

PR #2679, opened 2019-08-20
Store peeked channels' indexes in casper events proto

lukasz-golebiewski:
## Overview
Fixes a bug with COMM events with peek being not replayable

### JIRA ticket:
Related to: https://rchain.atlassian.net/browse/RCHAIN-3639

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2692, opened 2019-08-28
Peek merging rules

lukasz-golebiewski:
## Overview
Merging rules for peek events

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3639

### Notes


casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -20,31 +20,38 @@ case object Receive extends Polarity
...
 
   private[this] def toOperation(produce: Produce): TuplespaceOperation =
     TuplespaceOperation(Send, if (produce.persistent) NonLinear else Linear, produce.hash)
 
-  private[this] def toOperation(consume: Consume): TuplespaceOperation =
-    TuplespaceOperation(Receive, if (consume.persistent) NonLinear else Linear, consume.hash)
+  private[this] def toOperation(consume: Consume, peek: Boolean): TuplespaceOperation =
+    TuplespaceOperation(
+      Receive,
+      if (consume.persistent) NonLinear else if (peek) Peek else Linear,
```
dzajkowski: my OCD says "this needs to be extracted to a `match`" but I guess it's fine. ifs are faster, right?

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -113,6 +113,9 @@ trait MergeabilityRules {
   def atLeastOnePersistent(rho: Seq[Rho]): Boolean =
     rho.flatMap(_.maybeCardinality).exists(_ == NonLinear)
 
+  def atLeastOnePeek(rho: Seq[Rho]): Boolean =
+    rho.flatMap(_.maybeCardinality).exists(_ == Peek)
```
dzajkowski: random thought: `==` vs `===` shouldn't we use the latter?

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -328,6 +331,39 @@ trait MergeabilityRules {
...
   }
 
+  /***
+   There was a COMM within one of the deploys caused on one side by a peek.
+   The other deploy had an event without a match in TS, of same polarity to the data matched by peek.
+   These could not spawn more work.
+    */
+  object PeekedNoMatch extends MergeableCase {
+
+    override def precondition(
```
dzajkowski, line 403: code duplication alert

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -198,7 +198,9 @@ trait MergeabilityRules {
       (expectOne(left) { findMatch(_, base) }, expectOne(right) { findMatch(_, base) }) match {
         case (Some(m1), Some(m2)) =>
           assert(
-            m1 == m2 && m1.maybeCardinality == Some(Linear) || m2.maybeCardinality == Some(Linear)
+            m1 == m2 && m1.maybeCardinality != Some(NonLinear) || m2.maybeCardinality != Some(
```
dzajkowski: wouldn't it make sense to introduce a few helpers?

eg. `def isNonLinear(m) = m1.maybeCardinality == Some(NonLinear)`

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -224,14 +224,20 @@ trait MergeabilityRules {
...
-          )
-        case (m1, m2) => fail(s"Expected two matches but got $m1 and $m2")
+      expectOne(left) { left =>
+        expectOne(right) { right =>
+          (findMatch(left, base), findMatch(right, base)) match {
+            case (Some(m1), Some(m2)) =>
+              assert(
+                (m1 != m2 || m1.maybeCardinality == Some(NonLinear) || m2.maybeCardinality == Some(
+                  NonLinear
+                )) || (m1 == m2 && left.maybeCardinality == Some(Peek) && right.maybeCardinality == Some(
```
dzajkowski: is `left` & `right` right here?

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -263,7 +263,9 @@ trait MergeabilityRules {
         expectOne(right) { right =>
           val (a, b) = if (left.maybeCardinality == Some(Linear)) (left, right) else (right, left)
           assert(
-            a.maybeCardinality == Some(Linear) && (b.maybeCardinality == Some(Linear) || hasMatch(
+            a.maybeCardinality == Some(Linear) && (b.maybeCardinality == Some(Linear) || b.maybeCardinality == Some(
+              Peek
+            ) || hasMatch(
```
dzajkowski, line 320: this reads `(a linear and b linear) or b peek`.
a doesn't matter it the or'ed case?

dzajkowski (approved):

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Thanks for noticing this, I need to make preconditions a bit more sophisticated

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: Right :-)

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -20,31 +20,38 @@ case object Receive extends Polarity
...
+      consume.hash
+    )
 
   def from(produce: Produce): (Blake2b256Hash, TuplespaceEvent) =
     produce.channelsHash -> TuplespaceEvent(toOperation(produce), None)
 
   def from(consume: Consume): Option[(Blake2b256Hash, TuplespaceEvent)] = consume match {
     case Consume(singleChannelHash :: Nil, _, _, _) =>
-      Some(singleChannelHash -> TuplespaceEvent(toOperation(consume), None))
+      Some(singleChannelHash -> TuplespaceEvent(toOperation(consume, false), None))
```
ArturGajowy, line 42: Everytime I look at `Consume` not having a `peek` field I spend 10 minutes tracing back why that is. I also find the hardcoded `false` in here highly disturbing. It's like saying 'peeks never happen until there's a COMM, at which point the consume magically turns into a peek'.

If this can be deemed correct - it's only due to the fact that 4X has same rules as PX. But this is not expressed by this code, nor this class is the place to express that.

We really should be passing the `peek`-ness through Consume and using it here - for the sake of future maintainers.

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -328,6 +341,39 @@ trait MergeabilityRules {
...
+   These could spawn more work.
+   Mergeable if we use spatial matcher to prove they don't match.
+    */
+  object PeekedCouldMatch extends ConflictingCase {
+
+    override def precondition(
+        left: Rho*
+    )(right: Rho*)(base: Rho*)(implicit pos: source.Position): Effect[_] =
+      volatileEventPrecondition(left: _*)(right: _*)(base: _*) >> {
+        assert(atLeastOnePeek(left ++ right))
```
ArturGajowy: We should differentiate the precondtions between the `Peeked*` cases. Or, ideally, we'd incorporate peek cases into the existing ones.

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -446,19 +496,67 @@ trait MergeabilityRules {
...
   ) =
-    randomDiamondConflictCheck(base, b1, b2, numberOfParentsForDiamondTip = 1)
+    checkBothWays(base, b1, b2, numberOfParentsForDiamondTip = 1)
 
   def merges(b1: Rho, b2: Rho, base: Rho)(
       implicit file: sourcecode.File,
       line: sourcecode.Line
   ) =
-    randomDiamondConflictCheck(base, b1, b2, numberOfParentsForDiamondTip = 2)
+    checkBothWays(base, b1, b2, numberOfParentsForDiamondTip = 2)
```
ArturGajowy, line 636: can we revert to `randomDiamondConflictCheck`? It's a matter of 2x test time decrease. I'm fine if 'check both ways' stays as a util for future debugging

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -68,7 +68,11 @@ class MultiParentCasperMergeSpec
   }
 
   it should "respect mergeability rules when merging blocks" in effectTest {
-    mergeabilityCases.map(_._2).parSequence
+    baseMergeabilityCases.map(_._2).parSequence
+  }
+
+  it should "respect mergeability rules when merging blocks containing events with peek" in effectTest {
+    peekMergeabilityCases.map(_._2).parSequence
```
ArturGajowy, line 75: Let's add it to the previous test, maybe? Peek is not that special.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -151,7 +152,7 @@ object EstimatorHelper {
   ): Map[Blake2b256Hash, Set[TuplespaceEvent]] = {
     val nonPersistentProducesInComms = b.comms.flatMap(_.produces).filterNot(_.persistent)
     val nonPersistentConsumesInComms = b.comms.map(_.consume).filterNot(_.persistent)
-    val freeProduces                 = b.produces.diff(nonPersistentProducesInComms)
+    val freeProduces                 = b.produces.diff(nonPersistentProducesInComms).filter(_.sequenceNumber == 0)
```
ArturGajowy: 1. why did that have to change? How does addition of peeks affects `nonPersistentProducesInComms`? Is that name still correct? Or did we have a bug?
2. we shouldn't be relying on sequenceNumber, we intend to remove it after all.

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -20,31 +20,38 @@ case object Receive extends Polarity
...
   }
 
   def from(comm: COMM, produces: Set[Produce]): Option[(Blake2b256Hash, TuplespaceEvent)] =
     comm match {
-      case COMM(consume, produce :: Nil, _) => {
+      case COMM(consume, produce :: Nil, peek) => {
         val incoming: TuplespaceOperation =
-          if (produces.contains(produce)) toOperation(produce) else toOperation(consume)
+          if (produces.contains(produce)) toOperation(produce)
+          else toOperation(consume, peek.nonEmpty)
```
ArturGajowy: `s/peek/peeks`?

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -20,31 +20,38 @@ case object Receive extends Polarity
...
         val matched: Option[TuplespaceOperation] = Some(
-          if (incoming == toOperation(produce)) toOperation(consume) else toOperation(produce)
+          if (incoming == toOperation(produce)) toOperation(consume, peek.nonEmpty)
+          else toOperation(produce)
```
ArturGajowy: let's extract `val produceOperation = toOperation(produce)`

ArturGajowy: and `val consumeOperation = toOperation(consume, peek.nonEmpty)` (see usages above too)

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -56,20 +63,36 @@ object TuplespaceEvent {
...
       if (ev.incoming.polarity == other.incoming.polarity) {
 
+        val bothPeeks = (ev.incoming.cardinality == Peek) && (other.incoming.cardinality ==
+          Peek)
+
         val bothMatchedSameLinearEvent = for {
           thisMatched  <- ev.matched
           otherMatched <- other.matched
-        } yield thisMatched == otherMatched && otherMatched.cardinality == Linear
+        } yield (thisMatched == otherMatched) && (otherMatched.cardinality != NonLinear)
```
ArturGajowy: please drop the redundant parens

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ArturGajowy: `def isNonLinear(m) = m1.maybeCardinality.contains(NonLinear)`

ArturGajowy (changes requested):
Have some comments already, will go through particular cases in tests and implementation now

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: No, I'd rather have tests take a few seconds longer than need to revert something that made it into `dev` only because tests didn't run i a particular order

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
lukasz-golebiewski: 1) No, it's still correct. It affects `freeProduces`. Those with a `sequenceNumber > 0` were reintroduced into the tuplespace in the Reducer, but technically they were peeked, so they shouldn't count as "free"
2) There's no other way while we still have them. Once sequenceNumbers in produces are replaced with counters these will provide the necessary information

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: As you wish

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: This is only loosely related to this PR

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: The latter is not possible. I already promised Dom I'm going to improve the preconditions https://github.com/rchain/rchain/pull/2692#discussion_r323189222

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: Let's leave them separated

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: This won't make much difference in case of `case objects` but I'll change that because it's a good habit

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -446,19 +496,67 @@ trait MergeabilityRules {
...
+    "(P!) (P!)" -> VolatileEvent(P0, S0)(P0, S0)(Nil),
+    "(P!) P!"   -> PeekedNoMatch(P0, S0)(P_)(S1),
+    "4! (P!)"   -> PeekedNoMatch(F0)(P1, S1)(S0),
+    // The case below can merge iff peek block is executed as first - conflicts for now
+    "4! P!"     -> CouldMatchSameConflicts(F0)(P_)(S0),
+    "4! P!"     -> CouldMatchSameMerges(F0)(P1)(S0, S1),
+    "4! PX"     -> SamePolarityMerge(F0)(P1)(S0),
+    "4X PX"     -> SamePolarityMerge(F_)(P_)(Nil),
+    "P! (C!)"   -> PersistentNoMatch(P_)(C0, S0)(S1),
+    "P! (P!)"   -> HadItsMatch(S0)(P_)(F0),
```
ArturGajowy: I think this is more of `VolatileEvent`. `HadItsMatch` is not defined in terms of volatile events, and the right side had no 'awaiting' side of a comm (both events in parens happened during the right deploy, not during base)

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -446,19 +496,67 @@ trait MergeabilityRules {
...
+    "(P!) (C!)" -> PeekedCouldMatch(P0, S0)(C_, S0)(Nil),
+    "(!C) (P!)" -> PeekedCouldMatch(S0, C_)(P_, S0)(Nil),
+    "(P!) C!"   -> PeekedCouldMatch(P0, S0)(C0)(S1),
+    "(P!) CX"   -> PeekedCouldMatch(P_, S0)(C_)(Nil),
+    "4X (P!)"   -> PeekedCouldMatch(F0)(P_, S0)(Nil),
+    "P! CX"     -> SamePolarityMerge(P0)(C1)(S0),
+    "4X P!"     -> SamePolarityMerge(F0)(P_)(S1),
+    "PX P!"     -> SamePolarityMerge(P0)(P_)(S1),
+    // TODO:
+    "P! P!" -> CouldMatchSameConflicts(P_)(P_)(S0) // conflicts until replay errors are fixed
```
ArturGajowy: During our discussion of https://github.com/rchain/rchain/pull/2712 you mentioned (in the wiki):

> This shouldn't be a problem for now, as P! P! isn't going to conflict, and this is the pattern that will be used to access the PoS contract via the registry

What are the errors preventing this from being merged?
Do we have an approach for fixing them?

This is the main use case for peek (think: registry lookup). Fixing this should have immediate priority after this PR.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: Why? I offered rationale for merging. What's yours?

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ArturGajowy: It's not a few seconds (60-ish, in my case), and you'd run the tests for the code you're changing locally anyway - but fair enough.

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ArturGajowy: I've seen that comment after submitting my review

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
ArturGajowy: Not so sure about that. A hardcoded value is an immediate red flag, and the hardcoding happened in this PR. Also, when we've [last discussed this issue](https://github.com/rchain/rchain/pull/2679#discussion_r315866034), it's effectively been dismissed. I'd like to hear rationale against having `peek` flag in `Consume` other than the so far offered:

> They [4x px] behave identically in the mergeability matrix

as said above, this is a very implicit assumption and leads to confusion.

> it would make the Casper events model different than that of RSpace

the event models are different anyway (they use quite different ADTs, and non-trivial transform is done between them)

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -328,6 +341,39 @@ trait MergeabilityRules {
...
+        assert(atLeastOnePeek(left ++ right))
+      }.pure[Effect]
+
+  }
+
+  /***
+   There was a COMM within one of the deploys, with a peek on one side.
+   The other deploy had an event without a match in TS, dual to the non-linear.
+   These could spawn more work.
+   Mergeable if we use spatial matcher to prove they don't match.
```
ArturGajowy: We should change occurrences of this line to sth like:

```Not mergeable. Determining if the send matches the produce would require re-running the deploys, defeating the purpose of merging rules (knowing if deploys can be merged without replaying them).```

File: casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala [hunk 18]
```diff
@@ -328,6 +341,39 @@ trait MergeabilityRules {
 
   }
 
+  /***
+   There was a COMM within one of the deploys caused on one side by a peek.
+   The other deploy had an event without a match in TS, of same polarity to the data matched by peek.
```
ArturGajowy: > had an event without a match in TS

this can be (wrongly) interpreted as "no comms in the other deploy"

What would help me, is having this say "without a match *waiting* in TS"

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Yes

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Ok

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -56,20 +63,36 @@ object TuplespaceEvent {
...
+
+        if (bothPeeks) {
+          // TODO - should always return false
+          bothMatchedSameLinearEvent.getOrElse(false)
+        } else bothMatchedSameLinearEvent.getOrElse(false)
 
-        bothMatchedSameLinearEvent.getOrElse(false)
       } else {
+        ev.peeked.map(peeked => TuplespaceEvent(peeked, None).conflicts(other)).getOrElse(false) ||
+        other.peeked.map(peeked => TuplespaceEvent(peeked, None).conflicts(ev)).getOrElse(false) ||
```
ArturGajowy: The above two lines are only needed b/c of test cases involving a 'local' peek `(P!)`.
I've deleted the above, commented out the local-peek cases, and the peek tests passed.

This means our test cases for local-peek are in some way not consistent with the general merging rules. This should not be the case to my understanding.

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
ArturGajowy: `bothMatchedSameLinearEvent` is no longer a valid name - the condition includes peeks as well.
`bothMatchedSameOneTimeEvent`? `bothMatchedSameNonpersistentEvent`?

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
ArturGajowy: What is going to be the sequence number of the send `@0!(1)` below?

```
@0!(0) |
for (_ <- @0) {
  @0!(1)
}
```

I think it's going to be non-zero, so `sequenceNumber > 0` is not good enough for selecting peek-introduced sends.

See https://github.com/rchain/rchain/blob/425d76806470d744b44fadc18891b9364883fb3a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala#L286

ArturGajowy (changes requested):
Some more comments. Will go through test data for peek cases now (S0 vs S_, and stuff like that)

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
lukasz-golebiewski: No, it should be 0

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: The list of cases is already very long and hard to grasp so "dividing and conquering" is the way to go. Let's start with peek. "Peek is not that special" is just an opinion

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: I find random failures a nuisance during development. If this becomes too heavy a burden and if there haven't been any new changes to peek code for a while we might go back to selective testing

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: `bothMatchedSameNonNonLinearEvent` :trollface:

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Good catch!

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Replay errors. Getting rid of sequence numbers should sort this out. I don't want to keep this PR unmerged until this happens

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: We don't care whether a receive is going to potentially peek or consume a datum if there was no COMM event. Only after a COMM event happened this information becomes needed. Hence there's no objective rationale for storing this information with each and every consume

ArturGajowy:
I believe the following changes should be applied to the mergeability cases tests:
https://github.com/ArturGajowy/rchain/commit/f91766d2cdcb32c7066e7adcfc96b59df1cbe91b

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
ArturGajowy: I think I've found (part of?) the issue:
The spreadsheet (and our tests) only considers P! / P!! cases.
We should also be considering !P and !!P. This way, I think, the `4X (P!)` case will be covered by 4X !P (Incoming could match).

So, the `unsatisfied` condition will also be true for linear sends that matched a peek.

ArturGajowy: I've just confirmed that the tests pass with the above 2 lines deleted and `unsatisfied` defined as follows:

```scala
    private[casper] def unsatisfied: Boolean =
      ev.incoming.cardinality match {
        case Peek      => ev.matched.isEmpty
        case Linear    => ev.matched.forall(_.cardinality == Peek)
        case NonLinear => ev.matched.forall(_.cardinality != NonLinear)
      }
```

lukasz-golebiewski: Another good catch!

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: ok

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
ArturGajowy: Have you checked that?

lukasz-golebiewski: You're right, I'll tweak this part of code

lukasz-golebiewski:
@ArturGajowy please resolve outdated conversations

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
ghost: What sequence number is this?  It's not the one that shows up in discord discussions where the sequence number is the total order on blocks produced by any given validator.

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
ghost: I would also have assumed that a consume would have a flag for being a peek.  That said, peek itself is an optimization to allow merging blocks, and optimized code is not necessarily very maintainable.  If there is a way to implement this with similar performance and having the flag on the consume rather than the comm, I'd prefer it.

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ghost: @dzajkowski I think your parens are wrong.

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ghost: At Google, there were three levels of tests.  All the short tests taken together for a project would complete in a few minutes.  These would get run regularly during development.  Medium tests would take much longer, up to half an hour.  Medium tests would have to be run once before requesting a code review and again before submitting code.  Long tests could take many hours and involve many disparate systems.  They would be run periodically and would alert anyone who submtted code during that interval when they failed that their change might have caused the failure.  Later they were able to do a binary search (like git bisect) on the relevant PRs to find better information about whose code broke the tests.

Consider agreeing as a team on what constitutes a short test and how long running all the short tests should take.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ghost: I agree with Lukasz: so long as peek is not merely syntactic sugar for consume-then-produce, it merits being tested separately.

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
lukasz-golebiewski: Consume has this flag on the tuplespace level, it doesn't in the replay log because this flag is only relevant in replay when a COMM event happened. Since changing this is not "a must" for the peek mechanism to work, let's consider this idea and it's impact in a separate ticket

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
lukasz-golebiewski: No, it's a different mechanism, for details see https://github.com/rchain/rchain/pull/1817

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
```diff
@@ -213,16 +217,16 @@ class MultiParentCasperMergeSpec
...
+        //"4!!",
         "PX",
         "P!",
         "P!!",
-        "!!X",
-        "!!4",
-        "!!C",
+        //"!!X",
+        //"!!4",
+        //"!!C",
```
ArturGajowy, line 226: what's up with this?

ArturGajowy: Ah, I see - they were excluded by the filter :+1:

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ArturGajowy: What we're discussing is in the range of short tests. Still, a test longer than 10s affects productivity, because each iteration on the tested code results in a distraction.

In the future, we could replace the 'backend' of this test to just involve the 'merge rules' instead of the whole (test) casper code. We'd then generate an event log by running rholang interpereter on the input rho, feed log to the merge rules (hopefully extracted from EstimatorHelper by that time), and check the 'conflicts' result directly from there, rather than by checking how many parents the 'merge' block has.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
lukasz-golebiewski: I would add those tests in a separate PR as these are not directly connected with peek (and hence this PR)

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: Sounds good

ghost (approved):
Approved but please address Artur's comments if you haven't already.

lukasz-golebiewski:
Thanks @pyrocto . I think I have, but it might be hard to notice with so many unresolved conversations referring to outdated code

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
ArturGajowy: It's also relevant for saying whether two event logs are conflicting. We're essentially abusing the fact that PX has the same merging rules as 4X, without expressing it in code. Let's then have a separate PR addressing this.

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
ArturGajowy: > There's no other way while we still have them.

:stuck_out_tongue_winking_eye:

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
ArturGajowy: Please ping me so that we can discuss the problem and fix approach - this will likely make the PR and its review easier :). So far, I'm not sure why removing sequence numbers will fix that.

casper/src/test/scala/coop/rchain/casper/MultiParentCasperMergeSpec.scala
ArturGajowy: I think we should also consider making our tests for merge rules smaller. Maybe it'd make sense to do some of:

1) remove the notion of volatile events from tests, as they're secondary to the deploy-boundary-agnostic, original merge rules

2) add test cases for !P and !!P (not covered by the spreadsheet)

The two above would probably reduce the number of `MergeabilityCase`-s to 4

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
lukasz-golebiewski: I see that you are still "requesting changes". What exactly are the changes that are absolutely required before this PR can be merged?

casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala
```diff
@@ -56,20 +65,32 @@ object TuplespaceEvent {
...
-        ev.unsatisfied && other.unsatisfied
-      }
+        if (bothPeeks) {
+          // TODO - should always return false
+          bothMatchedSameNonPersistentEvent.getOrElse(false)
+        } else bothMatchedSameNonPersistentEvent.getOrElse(false)
+
+      } else ev.unsatisfied && other.unsatisfied
+
+    private[casper] def peeked: Option[TuplespaceOperation] =
```
ArturGajowy: I think this is unused

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -446,19 +544,98 @@ trait MergeabilityRules {
...
+  val peekMergeabilityCases = List(
+    "!4 P!"     -> HadItsMatch(S0)(P_)(F0, S1),
+    "!4 (P!)"   -> PeekedNoMatch(S0)(P_, S1)(F0),
+    "!X PX"     -> IncomingCouldMatch(S0)(P_)(Nil),
+    "!4 PX"     -> HadItsMatch(S0)(P_)(F_),
+    "!C PX"     -> HadItsMatch(S0)(P_)(C_),
+    "!X (P!)"   -> PeekedNoMatch(S0)(P_, S0)(Nil),
+    "!X P!"     -> HadItsMatch(S0)(P_)(S0),
+    "(!4) P!"   -> VolatileEvent(S0, F0)(P_)(S1),
+    "(!4) PX"   -> VolatileEvent(S0, F0)(P_)(Nil),
```
ArturGajowy, line 556: I think it'd be helpful to prefix all mergability case names with `Merge / `Conflict`, e.g.:

```
MergeHadItsMatch
MergePeekedNoMatch
ConflictIncomingCouldMatch
MergeVolatileEvent
MergeSamePolarityMerge
MergeCouldMatchSame[Didnt]
ConflictCouldMatchSame[Did]
```

ArturGajowy (approved):
Last two comments, LGTM!

lukasz-golebiewski:
bors try

PR #2695, opened 2019-08-28
Create methods for Either and ListOps that do not leak memory.

ghost:
## Overview
Either and ListOps had memory leaks where contracts would be created during a method call, closing over some temporary state, and then never be used again.  This change creates methods that allow passing state to a single contract rather than creating a new one each time.  Even when garbage collection is implemented, these methods will be more efficient.

Changing call sites will come in another PR.

To see the leak, compare `map` and `mapClean`.  In `map`, each call creates a new contract `combinator` that closes over the `function` argument.  This combinator is only ever used during the one invocation, but because it's persistent, there's no way to clean it up.  (There is a way using match to create a fixed point of a linear for that recreates itself, but that kind of interaction with rspace is much much slower.)

Contrast `mapClean`, where `function` is passed explicitly and the combinator is only ever created once.  It also allows the function being passed into `mapClean` to be of the same form, a process with no free variables and some state.

```
  contract ListOps(@"map", @list, function, return) = {
    new combinator in {
      contract combinator(@head, @accumulatedValue, return) = {
        new mappedHead in {
          function!(head, *mappedHead) |
          for(@x <- mappedHead) {
            ListOps!("append", accumulatedValue, x, *return)
          }
        }
      } |
      ListOps!("fold", list, [], *combinator, *return)
    }
  } |

  new combinator in {
    contract combinator(@(function, state), @head, @accumulatedValue, return) = {
      new mappedHead in {
        @function!(state, head, *mappedHead) |
        for(@x <- mappedHead) {
          ListOps!("append", accumulatedValue, x, *return)
        }
      }
    } |
    contract ListOps(@"mapClean", @list, function, @state, return) = {
      ListOps!("foldClean", list, [], *combinator, (*function, state), *return)
    }
  }
```

### DIFF
--- a/casper/src/main/resources/Either.rho
+++ b/casper/src/main/resources/Either.rho
@@ -89,7 +89,7 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
-    // Combines two eithers, discarding the first value and keeping the second
+    // Combines two eithers, discarding the first Right value and keeping the second.  Propagates Lefts.
     contract Either(@"productR", @a, @b, return) = {
       match (a, b) {
         ((false, _), _) => return!(a)
@@ -98,25 +98,55 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
     } |
 
     // Like map2, except it takes `aCh` and `bCh` and reads them to obtain the `a` and `b` parameters
-    contract Either(@"map2 <-", aCh, bCh, f, return) = {
+    contract Either(@"map2 <-", aCh, bCh, @f, return) = {
       for (@a <- aCh) {
         for (@b <- bCh) {
-          Either!("map2", a, b, *f, *return)
+          Either!("map2", a, b, f, *return)
+        }
+      }
+    } |
+
+    // Like map2Clean, except it takes `aCh` and `bCh` and reads them to obtain the `a` and `b` parameters
+    contract Either(@"map2Clean <-", aCh, bCh, @f, @state, return) = {
+      for (@a <- aCh) {
+        for (@b <- bCh) {
+          Either!("map2Clean", a, b, f, state, *return)
         }
       }
     } |
 
     // Composes two eithers `a: A` and `b: B` using the function `f`. See cats.Apply.map2 for details.`
-    contract Either(@"map2", @a, @b, f, return) = {
+    contract Either(@"map2", @a, @b, @f, return) = {
       match (a, b) {
         ((false, _), _)               => return!(a)
         (_, (false, _))               => return!(b)
         ((true, va), (true, vb)) => {
           new ret in {
-            match *f {
+            match f {
               { for (x, y, r <- fName) { _ } } => {
-                *f |
+                f |
                 fName!(va, vb, *ret) |
+                for (@v <- ret) {
+                  return!((true, v))
+                }
+              }
+            }
+          }
+        }
+      }
+    } |
+
+    // Like map2, but allows passing state to f instead of making f close over it.
+    contract Either(@"map2Clean", @a, @b, @f, @state, return) = {
+      match (a, b) {
+        ((false, _), _)               => return!(a)
+        (_, (false, _))               => return!(b)
+        ((true, va), (true, vb)) => {
+          new ret in {
+            match f {
+              { for (@state, x, y, r <- fName) { _ } } => {
+                f |
+                fName!(state, va, vb, *ret) |
                 for (@v <- ret) {
                   return!((true, v))
                 }
@@ -134,6 +164,13 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
+    // Like flatMapClean, except it takes `eitherCh` and reads it to obtain the `either` parameter
+    contract Either(@"flatMapClean <-", eitherCh, f, @state, return) = {
+      for (@either <- eitherCh) {
+        Either!("flatMapClean", either, *f, state, *return)
+      }
+    } |
+
     // Right-biased flatMap; f must also return an Either
     contract Either(@"flatMap", @either, f, return) = {
       match either {
@@ -142,10 +179,25 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
+    // Like flatMap, except it allows passing state to f instead of making f close over it.
+    contract Either(@"flatMapClean", @either, f, @state, return) = {
+      match either {
+        (true, value) => { f!(state, value, *return) }
+        (false, _)      => { return!(either) }
+      }
+    } |
+
     // Like map, except it takes `eitherCh` and reads it to obtain the `either` parameter
     contract Either(@"map <-", eitherCh, f, return) = {
       for (@either <- eitherCh) {
         Either!("map", either, *f, *return)
+      }
+    } |
+
+    // Like mapClean, except it takes `eitherCh` and reads it to obtain the `either` parameter
+    contract Either(@"mapClean <-", eitherCh, f, @state, return) = {
+      for (@either <- eitherCh) {
+        Either!("mapClean", either, *f, state, *return)
       }
     } |
 
@@ -164,8 +216,23 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
+    // Like map, except it allows passing state to f instead of making f close over it.
+    contract Either(@"mapClean", @either, f, @state, return) = {
+      match either {
+        (true, value) => {
+          new mappedResultCh in {
+            f!(state, value, *mappedResultCh) |
+            for(@mappedResult <- mappedResultCh) {
+              return!((true, mappedResult))
+            }
+          }
+        }
+        (false, _) => return!(either)
+      }
+    } |
+
     new combinator in {
-      contract combinator(@head, @accumulatedValue, return) = {
+      contract combinator(_, @head, @accumulatedValue, return) = {
         Either!("flatMap", accumulatedValue, head, *return)
       } |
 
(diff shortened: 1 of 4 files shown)


JosephDenman:
Just so we’re clear, would you provide an example of the problem you’re solving in the summary?

ghost:
@JosDenmark Done.

casper/src/main/resources/ListOps.rho
```diff
@@ -87,11 +128,25 @@ new  ListOps, rs(`rho:registry:insertSigned:secp256k1`), uriOut in {
...
+      ListOps!("foldClean", list, [], *combinator, (*function, state), *return)
+    }
+  } |
+
   new getOutput in {
     contract getOutput(name, ret)  = { for(@result <- name){ ret!(result) } } |
     contract ListOps(@"parMap", @list, function, return) = {
       new namesListCh, startProcess in {
-        contract startProcess(@x, ret) = { new outCh in { function!(x, *outCh) | ret!(*outCh) } } |
+        contract startProcess(@x, ret) = { function!(x, *ret) } |
```
JosephDenman: Without `outCh`, `"parMap"` is not a parallel map. In the previous version, `outCh` is immediately returned on `ret`, so the result of `startProcess` is immediately known. So, when the `combinator` contract inside `mapClean` calls `function!(...)`, it immediately gets back `outCh` and immediately appends that channel to the `accumulatedValue`. Without `outCh`, the call `function!(...)` has to completely finish before the next process can be started. Obviously, the same is true for `"parMapClean"`.

casper/src/main/resources/ListOps.rho
```diff
@@ -87,11 +128,25 @@ new  ListOps, rs(`rho:registry:insertSigned:secp256k1`), uriOut in {
     }
   } |
 
+  new combinator in {
+    contract combinator(@(function, state), @head, @accumulatedValue, return) = {
+      new mappedHead in {
+        @function!(state, head, *mappedHead) |
+        for(@x <- mappedHead) {
+          ListOps!("append", accumulatedValue, x, *return)
```
JosephDenman: The more efficient `accumulatedValue ++ [x]` should be in-lined here.

casper/src/main/resources/ListOps.rho
```diff
@@ -141,6 +213,45 @@ new  ListOps, rs(`rho:registry:insertSigned:secp256k1`), uriOut in {
...
+        _  => { startCount!(count) }
+      }
+    } |
+
+    contract ListOps(@"unorderedParMapClean", @list, function, @state, return) = {
+      new collect, startCount, completeCount, accCh in {
+        //Spawn processes, each computing one element
+        //also processes listening for the result
+        //add a result to our accumulating list
+        contract collect(@result) = {
```
JosephDenman, line 232: It looks like `collect` can be extracted in a manner similar to what you've done for the others.

JosephDenman (changes requested):
The refactorings that excise the temporary contracts are really nice. A couple of points:

- This PR can be improved by adding tests for each of the methods you "purified", and at the very least a couple of tests for `partialFoldClean`.

- At least in the places that you use it (though preferably in all occurrences), list concatenation should be used in place of `ListOps!("append")/ ListOps!("prepend")`. This includes the use in `reverse`, where `"prepend"` is used instead of another `prepend`.

I have mixed feelings about the `state` parameter. I see that it increases reusability of two-argument processes like `greaterThan(@bound, @x, ret) = { ret!(x < bound) }`, but it seems like the generalization to two arguments is somewhat arbitrary. It's not useful, for example, if I have a process that takes three arguments. I think that a more general (probably post-Mercury) solution to the code reusability issues we see with `contract` will solve the issue that the state parameter attempts to solve and more. In the mean time, leaving out `state` would make the methods more comprehensible. I'm curious what others think.

I've only looked at `ListOps` and `ListOpsTest`. I'll finish the rest tomorrow.

casper/src/main/resources/ListOps.rho
ghost: You're right, but you're looking at old code.  It was fixed in the last two commits yesterday.

ghost:
> I have mixed feelings about the state parameter. I see that it increases reusability of two-argument processes like greaterThan(@bound, @x, ret) = { ret!(x < bound) }, but it seems like the generalization to two arguments is somewhat arbitrary. It's not useful, for example, if I have a process that takes three arguments.

In general, the state is a tuple containing all the variables that would otherwise be closed over, though I don't bother wrapping singletons.  See, for instance, `foldClean` where `combinatorAndCondition` needs both the combinator and the state for the combinator when delegating to `partialFoldClean`:

```
contract combinatorAndCondition(@(combinator, state), @head, @accumulatedValue, return) = { ... } |
ListOps!("partialFoldClean", list, start, *combinatorAndCondition, (*combinator, state), *result)
```
> I think that a more general (probably post-Mercury) solution to the code reusability issues we see with contract will solve the issue that the state parameter attempts to solve and more.

Yes, garbage collection would solve all these issues; in the meantime, this prevents the leaks.  Once garbage collection is in place, these changes could conceivably be reverted; the old code will still be in the repository.

> In the mean time, leaving out state would make the methods more comprehensible.

Unfortunately, without garbage collection that comes at the cost of leaks due to closing over the state.

casper/src/main/resources/ListOps.rho
ghost: No, it can't: `collect` closes over `startCount`, `completeCount`, and `accCh`, but I can't pass those in without changing the signature of `function`.  That kind of refactoring would have to alter the call site.

ghost:
> At least in the places that you use it (though preferably in all occurrences), list concatenation should be used in place of ListOps!("append")/ ListOps!("prepend"). This includes the use in reverse, where "prepend" is used instead of another prepend.

Done.

casper/src/main/resources/ListOps.rho
ghost: Done.

casper/src/main/resources/Either.rho
```diff
@@ -127,13 +136,41 @@ new  Either, rs(`rho:registry:insertSigned:secp256k1`), uriOut, rl(`rho:registry
       }
     } |
 
+    // Like map2, but allows passing state to f instead of making f close over it.
+    contract Either(@"map2Clean", @a, @b, f, @state, return) = {
```
odeac: Having both `method` and `methodClean` is confusing. It would be nice to keep only one set of methods

ghost: It would, but because the signatures changed, it would involve changing all the call sites.  That will come gradually over a few PRs, after which we can remove the leaky versions.

ghost:
@JosDenmark 
> This PR can be improved by adding tests for each of the methods you "purified", and at the very least a couple of tests for partialFoldClean.

Done.

casper/src/test/resources/EitherTest.rho
```diff
@@ -33,70 +52,186 @@ in {
...
+        @Either!("fromSingletonList <-", *ch3, "If empty", *ch3ret) |
         rhoSpec!("assertMany",
           [
-            (((true, 1), "== <-", *ch1), "empty function list behaves like pure"),
-            (((true, 2), "== <-", *ch2), "Composing one function behaves like fmap . pure"),
-            (((true, 4), "== <-", *ch3), "Composing two functions behaves like flatMap . pure"),
-            (((false, "Div by zero!"), "== <-", *ch4), "Left values are propagated through compose")
+            (((false, "If empty"), "== <-", *ch1ret), "Uses default value when empty"),
+            (((true, "Not empty"), "== <-", *ch2ret), "Uses given value when singleton"),
+            (((false, "Invalid argument, expected a list of at most 1 element"), "== <-", *ch3ret), "Returns error when too long"),
```
JosephDenman, line 91: It surprises me that passing an empty list returns the default vault but passing a two-element list does not. Unless singleton lists are more invalid than two-element lists, these should probably be uniform. It wouldn't be difficult to change it in this PR, but that's up to you. At least a TODO would be helpful.

JosephDenman (approved):
LGTM!

casper/src/test/resources/EitherTest.rho
ghost: I kept all the APIs the same with the exception of adding `state` to the `*Clean` methods.  API changes should be ticketed and in a separate PR.

casper/src/test/resources/EitherTest.rho
```diff
@@ -25,31 +48,174 @@ in {
...
+        @Either!("map2Clean <-", *ch1a, *ch2a, for (@state, @x, @y, r <- f) { r!(x + y + state) }, 7, *reta) |
+        @Either!("map2Clean <-", *ch1b, *ch2b, for (@state, @x, @y, r <- f) { r!(x + y + state) }, 7, *retb) |
+        @Either!("map2Clean <-", *ch1c, *ch2c, for (@state, @x, @y, r <- f) { r!(x + y + state) }, 7, *retc) |
         rhoSpec!("assertMany",
           [
-            (((false, "message"), "== <-", *ch1), "Left values are not affected by map"),
-            (((true, 6), "== <-", *ch2), "Right values are changed by map")
+            (((false, 1), "== <-", *reta), "If a is Left, return a"),
+            (((false, 4), "== <-", *retb), "If a is Right and b is Left, return b"),
+            (((true, 18), "== <-", *retc), "If a and b are Right, apply f")
```
lukasz-golebiewski: I was wondering where 18 came from. Maybe extracting `7` to `chInitialState!!(7)` and then using that channel in test call would help with noticing that?

lukasz-golebiewski (changes requested):
This could be easily split into two PRs to reduce the burden on the reviewers. Let's not do this now, just a remark for the future.
There are a few commits with "fix" or "undo" in their names which I think should be melded into other commits.
Aside from that only one test readability improvement suggestion from me.
Otherwise looks good!

casper/src/test/resources/EitherTest.rho
ghost: Changed 18 to 5+6+7.

ghost (commented):
> This could be easily split into two PRs to reduce the burden on the reviewers. Let's not do this now, just a remark for the future.

Since it was an almost automatic transformation of the existing code, I didn't think it would be too onerous.

lukasz-golebiewski (approved):

odeac (approved):

PR #2697, opened 2019-08-29
Add new metrics

sebtomba:
## Overview
New metrics:
- Propose counter
- Propose failed counter
- Propose time (only if the lock is free)
- Create block time
- Add block time
- ~~RSpace comm consume/produce/install time~~
- ~~LMDB-Store read/write time~~
- Rholang cost accounting metrics

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3739


shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
```diff
@@ -28,14 +34,18 @@ final case class LMDBStore[F[_]: Sync](env: Env[ByteBuffer], dbi: Dbi[ByteBuffer
...
 
   def withReadTxnF[R](f: Txn[ByteBuffer] => R): F[R] =
-    Sync[F].delay {
-      withTxn(env.txnRead)(f)
-    }
+    Sync[F]
+      .delay {
+        withTxn(env.txnRead)(f)
+      }
+      .timer("lmdb.read")
```
dzajkowski: we need to educate about this metric and we're not allowed to just draw out conclusions

dzajkowski (approved):

dzajkowski:
bors try

rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
```diff
@@ -40,7 +47,9 @@ object StoreInstances {
                 .setMaxReaders(config.maxReaders)
                 .open(config.path.toFile, config.flags: _*)
             }
-      dbi   <- Sync[F].delay { env.openDbi("db", MDB_CREATE) }
+      dbi <- Sync[F].delay {
```
lukasz-golebiewski: This change is unnecessary

shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
lukasz-golebiewski: Upon each invocation of this method the underlying `TrieMap` is going to be accessed to search whether the timer already exists. I'm worried about the performance impact. This should be only a temporary change for a single debugging session, I really don't think this should go into dev

rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
sebtomba: I don't recall to make this change. Wondering if the Scalafmt did this?

shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
sebtomba: removed all LMDB-Store metrics

KentShikama (approved):
Casper metrics looks good and have been helpful so far on devnet under this branch

PR #2709, opened 2019-09-03
Rchain 3811: raise error if requiredSigs is not smaller than the number of bonded validators

odeac:
## Overview
see title & Jira

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3811

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
@@ -9,7 +9,10 @@ import cats.implicits._
 import coop.rchain.blockstorage.util.io.IOError.RaiseIOError
 import coop.rchain.casper.LastApprovedBlock.LastApprovedBlock
 import coop.rchain.casper.genesis.Genesis
+import coop.rchain.casper.genesis.Genesis.createGenesisBlock
+import coop.rchain.casper.genesis.contracts.{ProofOfStake, Validator}
 import coop.rchain.casper.protocol._
+import coop.rchain.casper.util.{BondsParser, VaultParser}
 import coop.rchain.casper.util.comm.CommUtil
 import coop.rchain.casper.util.rholang.RuntimeManager
 import coop.rchain.casper.{LastApprovedBlock, PrettyPrinter, Validate, _}
@@ -69,17 +72,32 @@ object ApproveBlockProtocol {
       interval: FiniteDuration
   ): F[ApproveBlockProtocol[F]] =
     for {
-      genesisBlock <- Genesis.fromInputFiles[F](
-                       maybeBondsPath,
-                       numValidators,
-                       genesisPath,
-                       maybeVaultsPath,
-                       minimumBond,
-                       maximumBond,
-                       shardId,
-                       deployTimestamp
+      now       <- Time[F].currentMillis
+      timestamp = deployTimestamp.getOrElse(now)
+
+      vaults <- VaultParser.parse[F](maybeVaultsPath, genesisPath.resolve("wallets.txt"))
+      bonds <- BondsParser.parse[F](
+                maybeBondsPath,
+                genesisPath.resolve("bonds.txt"),
+                numValidators,
+                genesisPath
+              )
+      validators = bonds.toSeq.map(Validator.tupled)
+      genesisBlock <- createGenesisBlock(
+                       implicitly[RuntimeManager[F]],
+                       Genesis(
+                         shardId = shardId,
+                         timestamp = timestamp,
+                         proofOfStake = ProofOfStake(
+                           minimumBond = minimumBond,
+                           maximumBond = maximumBond,
+                           validators = validators
+                         ),
+                         vaults = vaults,
+                         supply = Long.MaxValue
+                       )
                      )
-      now   <- Time[F].currentMillis
+
       sigsF <- Ref.of[F, Set[Signature]](Set.empty)
     } yield new ApproveBlockProtocolImpl[F](
       genesisBlock,
--- a/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
@@ -1,12 +1,17 @@
 package coop.rchain.casper.engine
 
+import java.nio.file.Path
+
 import cats.FlatMap
-import cats.effect.Sync
+import cats.effect.{Concurrent, Sync}
 import cats.effect.concurrent.Ref
 import cats.implicits._
+import coop.rchain.blockstorage.util.io.IOError.RaiseIOError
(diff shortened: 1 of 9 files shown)


casper/src/test/scala/coop/rchain/casper/engine/BlockApproverProtocolTest.scala
```diff
@@ -43,7 +43,7 @@ class BlockApproverProtocolTest extends FlatSpec with Matchers {
   }
 
   it should "log a warning for invalid ApprovedBlockCandidates" in effectTest {
-    val n = 8
+    val n = 3
```
itegulov: I assume this number was changed due to its "implicit" dependency on number of validators generated in GenesisParameters so maybe make it explicitly depend on it. In other words, change it to something like `genesisParams.genesis.bonds.size - 1`

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
```diff
@@ -69,17 +72,32 @@ object ApproveBlockProtocol {
...
+                         shardId = shardId,
+                         timestamp = timestamp,
+                         proofOfStake = ProofOfStake(
+                           minimumBond = minimumBond,
+                           maximumBond = maximumBond,
+                           validators = validators
+                         ),
+                         vaults = vaults,
+                         supply = Long.MaxValue
+                       )
```
itegulov: Why have you decided to inline it here? If there is a good reason for that, shouldn't you delete `Genesis.fromInputFiles` since it isn't used anywhere else? BTW, you have also lost the test coverage by inlining it here since `GenesisTest` no longer tests this piece of code.

itegulov (changes requested):
The `Genesis.fromInputFiles` inlining is a concern for me. Please address.

ArturGajowy (commented):
I concur with @itegulov. The validity check should be part of `Genesis.createGenesisBlock`, ideally. `Genesis.fromInputFiles` at the very least.

Also, the validity checks for vaults in https://github.com/rchain/rchain/blob/3ffc4fee6e86ec813d1805920a7a75ed441b8ccf/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala#L145-L150 would ideally be part of `Genesis.createGenesisBlock`

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
odeac: That function was doing too  much and I needed to split it into parsing the bonds file and using the bonds map. This way I had access to the bonds map so I can check that requiredSigs is smaller than the number of bonds.

The other problem is that the function was used in other places too so I had to inline it here and keep the other usages as they were.

ArturGajowy: Leaving this inlined is a step back. Instead, you should separate the parsing logic and validation logic *within* `fromInputFiles` (ideally moving validation logic into `createGenesisBlcok`), and add the check in the "logic" part (it's valid for all use cases of Genesis.createGenesisBlock)

odeac: That's what I did.I separated the parsing logic and moved bonds parser to util/BondsParser.scala and the vaults parser to util/VaultsParser.scala and then I called them instead of calling Genesis.fromInputFiles.

As for the validation logic I don't think it's a part of createGenesisBlock because the genesisBlock should not be aware of the requiredSigs parameter. That is specific to ApproveBlockProtocol so the validation should be done in its constructor i.e. the function ApproveBlockProtocol.of

itegulov: @odeac it is possible to access genesis bonds without actually having access to the parsed data structure from `BondsParser`. They are available through `ProtoUtil.bonds(genesis)` which internally just calls `genesis.body.get.state.get.bonds`. This way you don't have to touch `Genesis.fromInputFiles` at all.

odeac: That seems to be a good solution. I'll do that

odeac: Correction: I've opened a discussion in #blockchain to see if we keep that field or if it should be removed. I'll continue with this PR after we clarify this aspect

ArturGajowy: > As for the validation logic I don't think it's a part of createGenesisBlock because the genesisBlock should not be aware of the requiredSigs parameter. That is specific to ApproveBlockProtocol so the validation should be done in its constructor i.e. the function ApproveBlockProtocol.of

It is an integrity check for our config values, regardless of where it's used. `bonds.size == validators.size`, and `validators` is part of `Genesis`. Hence, the smart constructor for Genesis should validate that constraint. All the parties in the genesis (including ceremony master) should have same config values for requiredSigs and initial bonds.

odeac: @itegulov Following the discussion sin #blockchain discord channel, the decision is that we should remove BlockMessage.body.state.bonds so this is no longer a valid solution

odeac: @ArturGajowy So you want to add a parameter named requiredSigs in createGenesisBlock such that I can verify the condition that requiredSigs < bonds.size and fail if that's not the case?

ArturGajowy: I thought `requiredSigs` belongs to the `Genesis` case class, so I'd add a field there if it's not there already.

ArturGajowy: > I'd add a field there if it's not there already.

No, that's wrong. You can actually validate the `requiredSigs >= bondedValidators` by having `val bondedValidators = genesis.proofOfStake.validators.size`

odeac: The function Genesis.fromInputFiles doesn't return an instance of the Genesis case class. It returns a BlockMessage representing the genesis block. That's the problem.

I don't understand why do we care about the Genesis.fromInputFiles function which is only used in one place in the tests and one place in production. I've inlined both its uses and we don't have any problems.

ArturGajowy: > I've inlined both its uses and we don't have any problems.

I haven't noticed that - it changed during the works, as you've mentioned. What you've done with moving out the config parsing functions makes sense :+1:

itegulov (approved):

ArturGajowy (approved):

PR #2711, opened 2019-09-05
Add a rnode CLI command for checking whether given block has been finalized

itegulov:
## Overview
Title says it all

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3806

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -16,7 +16,7 @@ import coop.rchain.casper._
 import coop.rchain.casper.DeployError._
 import coop.rchain.casper.MultiParentCasper.ignoreDoppelgangerCheck
 import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.{EventConverter, ProtoUtil}
+import coop.rchain.casper.util.{DagOperations, EventConverter, ProtoUtil}
 import coop.rchain.casper.util.rholang.RuntimeManager
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b512Random
@@ -24,7 +24,7 @@ import coop.rchain.graphz._
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.metrics.implicits._
 import coop.rchain.models.BlockHash.BlockHash
-import coop.rchain.models.Par
+import coop.rchain.models.{BlockMetadata, Par}
 import coop.rchain.models.rholang.sorter.Sortable._
 import coop.rchain.models.serialization.implicits.mkProtobufInstance
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
@@ -609,6 +609,48 @@ object BlockAPI {
             lastFinalizedBlock <- casper.lastFinalizedBlock
             blockInfo          <- getFullBlockInfo[F](lastFinalizedBlock)
           } yield LastFinalizedBlockResponse(blockInfo = Some(blockInfo)).asRight,
+        Log[F].warn(errorMessage).as(s"Error: $errorMessage".asLeft)
+      )
+    )
+  }
+
+  def isFinalized[F[_]: Monad: EngineCell: SafetyOracle: BlockStore: Log](
+      request: IsFinalizedQuery
+  ): Effect[F, IsFinalizedResponse] = {
+    val errorMessage =
+      "Could not check if block is finalized, casper instance was not available yet."
+    EngineCell[F].read >>= (
+      _.withCasper[ApiErr[IsFinalizedResponse]](
+        implicit casper =>
+          for {
+            lastFinalizedBlock <- casper.lastFinalizedBlock
+            lastFinalizedBlockMetadata = BlockMetadata
+              .fromBlock(lastFinalizedBlock, invalid = false)
+            dag                   <- casper.blockDag
+            givenBlockHash        = ProtoUtil.stringToByteString(request.hash)
+            givenBlockMetadataOpt <- dag.lookup(givenBlockHash)
+            result <- givenBlockMetadataOpt match {
+                       case None =>
+                         s"Could not find block with hash ${request.hash}"
+                           .asLeft[IsFinalizedResponse]
+                           .pure[F]
+                       case Some(givenBlockMetadata) =>
+                         DagOperations
+                           .bfTraverseF(List(lastFinalizedBlockMetadata)) { b =>
+                             b.parents.traverse(dag.lookup).map { parentOpts =>
+                               parentOpts.flatten.distinct
+                                 .filter(_.blockNum >= givenBlockMetadata.blockNum)
+                             }
+                           }
+                           .contains(givenBlockMetadata)
+                           .map(
+                             isContained =>
+                               IsFinalizedResponse(
+                                 isFinalized = isContained
+                               ).asRight[Error]
+                           )
+                     }
+          } yield result,
         Log[F].warn(errorMessage).as(s"Error: $errorMessage".asLeft)
       )
     )
--- a/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
@@ -110,6 +110,7 @@ object DagOperations {
 
         val ln = l.blockNum
         val rn = r.blockNum
+        // Notice the inverted order of compared items, which makes the ordering descending
         rn.compare(ln) match {
           case 0 => compareByteString(l.blockHash, r.blockHash)
           case v => v
--- a/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
@@ -101,11 +101,12 @@ object DagOperations {
       b2: BlockMetadata,
       dag: BlockDagRepresentation[F]
   ): F[BlockMetadata] = {
+    val iterableByteOrdering = Ordering.Iterable[Byte]
 
     implicit val blockMetadataByNumDecreasing: Ordering[BlockMetadata] =
       (l: BlockMetadata, r: BlockMetadata) => {
         def compareByteString(l: ByteString, r: ByteString): Int =
-          l.hashCode().compareTo(r.hashCode())
+          iterableByteOrdering.compare(l.toByteArray, r.toByteArray)
 
         val ln = l.blockNum
         val rn = r.blockNum
--- a/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
@@ -92,10 +92,11 @@ object DagOperations {
   }
 
   /**
-    * Conceptually, the LCA is the lowest point at which the histories of b1 and b2 diverge.
-    * We compute by finding the first block that is the "lowest" (has highest blocknum) common block.
+    * Conceptually, the LUCA is the lowest point at which the histories of b1 and b2 diverge.
+    * We compute by finding the first block that is the "lowest" (has highest blocknum) block common
+    * for both blocks' ancestors.
     */
-  def lowestCommonAncestorF[F[_]: Monad](
+  def lowestUniversalCommonAncestorF[F[_]: Monad](
       b1: BlockMetadata,
       b2: BlockMetadata,
       dag: BlockDagRepresentation[F]
--- a/casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
@@ -1,0 +1,141 @@
+package coop.rchain.casper.api
+
+import cats.effect.Sync
+import cats.implicits._
+import com.google.protobuf.ByteString
+import coop.rchain.blockstorage.BlockStore
+import coop.rchain.casper.engine.Engine
+import coop.rchain.casper.{EngineWithCasper, SafetyOracle}
+import coop.rchain.casper.helper._
+import coop.rchain.casper.helper.BlockGenerator._
+import coop.rchain.casper.protocol._
+import coop.rchain.casper.util.GenesisBuilder._
+import coop.rchain.casper.util.ProtoUtil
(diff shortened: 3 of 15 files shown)


casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,263 @@
...
+          import node.{blockStore, casperState, cliqueOracleEffect, logEff}
+          val engine = new EngineWithCasper[Task](node.casperEff)
+          for {
+            b2 <- createBlock[Task](
+                   Seq(genesis.blockHash),
+                   genesis,
+                   v3,
+                   bonds,
+                   Map(v1 -> genesis.blockHash, v2 -> genesis.blockHash, v3 -> genesis.blockHash)
+                 )
```
ArturGajowy: Why are you steering block creation by hand instead of just simulating real node behaviour by using 3 nodes? Those tests would be like 10x shorter with that approach.

See https://github.com/rchain/rchain/blob/dc1f0ef633618ff0671975ee88c6709e851286e6/casper/src/test/scala/coop/rchain/casper/EstimatorHelperTest.scala#L47-L77 for an example of creating a complex graph (yous seems simpler :))

itegulov: Good point. I rewrote the PR tests by using three real nodes instead. Unfortunately, many older tests in `coop.rchain.casper.api` follow the "manual" block creation pattern.

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,161 @@
...
+    HashSetCasperTestNode.networkEff(genesisContext, networkSize = 3).use {
+      case n1 +: n2 +: n3 +: _ =>
+        import n1.{blockStore, cliqueOracleEffect, logEff}
+        val engine = new EngineWithCasper[Task](n1.casperEff)
+        for {
+          produceDeploys <- (0 until 5).toList.traverse(i => basicDeployData[Task](i))
+
+          b1 <- n1.addBlock(produceDeploys(0))
+          _  <- n2.receive()
+          _  <- n3.receive()
```
ArturGajowy: Thx for rephrasing the tests! :)

Could you add a `.publishBlock(deployDatums: DeployData*)(nodes: HashSetCasperTestNode: _*)` method to HashSetCasperTestNode, that would do `nodes.filter(_ != this).traverse_(_.receive())`? This way we could

`b1 <- n1.publishBlock((produceDeploys(0))(nodes)`, where `val nodes = List(n1, n2, n3)`

KentShikama (approved):
Perhaps add a test for when it is the secondary parent of the finalized block?

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,161 @@
...
+
+          lastFinalizedBlock <- n1.casperEff.lastFinalizedBlock
+          _                  = lastFinalizedBlock shouldBe b2
+
+          engineCell <- Cell.mvarCell[Task, Engine[Task]](engine)
+          // Checking if last finalized block is finalized
+          isB2Finalized <- BlockAPI.isFinalized[Task](
+                            IsFinalizedQuery(hash = ProtoUtil.hashString(b2))
+                          )(Sync[Task], engineCell, SafetyOracle[Task], blockStore, logEff)
+          _ = isB2Finalized.right.value shouldBe IsFinalizedResponse(isFinalized = true)
```
ArturGajowy: The above 4 lines are duplicated 4 times in the added tests. I think they deserve a util method in that test class, which will also make the tests clearer (hopefully blockStore and logEff can be easily made implicit, so we'd only pass `engineCel` or `engine` to that method)

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -606,4 +606,45 @@ object BlockAPI {
...
+                             lastFinalizedBlockMetadata,
+                             givenBlockMetadata,
+                             dag
+                           )
+                           .map { lca =>
+                             if (lca == givenBlockMetadata) {
+                               IsFinalizedResponse(isFinalized = true).asRight
+                             } else {
+                               IsFinalizedResponse(isFinalized = false).asRight
+                             }
```
ArturGajowy, line 643: This could just be `IsFinalizedResponse(isFinalized = lca == givenBlockMetadata).asRight`

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
itegulov: Done

itegulov:
~~While adding new tests as per @KentShikama's request, I realized that our current LCA implementation is buggy. I have ticketed my observations in https://rchain.atlassian.net/browse/RCHAIN-3828 and, since what is being added in this PR is higly dependent on LCA algorithm, I am temporarily switching my focus on RCHAIN-3828.~~

See https://github.com/rchain/rchain/pull/2719#issuecomment-529865613

itegulov:
See: https://github.com/rchain/rchain/pull/2719

itegulov:
@ArturGajowy PTAL. I have addressed your comments and also had to do a few more changes:
1. Renamed **current** lowestCommonAncestor function to lowestCommonForParentsAncestor
2. Implemented classical lowestCommonAncestor function which follows the classical LCA algorithm for DAGs
3. Used the newly introduced classical LCA for is-finalized call

Sorry for bringing a lot of confusion to this PR.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -91,6 +91,62 @@ object DagOperations {
...
+  private def getParents[F[_]: Monad](
+      dag: BlockDagRepresentation[F],
+      p: BlockMetadata
+  ): F[Set[BlockMetadata]] =
+    p.parents.traverse(dag.lookup).map(_.toSet.flatten)
+
+  /**
+    * LCA is the "lowest" (has highest blocknum) block which is an ancestor to both blocks.
+    */
+  def lowestCommonAncestorF[F[_]: Monad](
```
ArturGajowy: I think that rather than implementing this algorithm twice, we should express `isFinalized(given)` in terms of  
`.bfTraverse(lastFinalized)(n => n.parents.fliter(_.blockNum >= given.blockNum)).contains(given)`

Notice that the `.filter` above is just for early cutoff when `given` is above/sideways to `lastFinalized`

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -91,37 +91,80 @@ object DagOperations {
...
+        .map(_._1.head)
+    }
+  }
+
+  /**
+    * Conceptually, the LCfPA is the lowest point at which the histories of b1 and b2 diverge.
+    * We compute by finding the first block that is the "lowest" (has highest blocknum) block common
+    * for both blocks' parents.
+    */
+  def lowestCommonForParentsAncestorF[F[_]: Monad](
```
ArturGajowy: I think `for parents` is not precise here, as for the following:

```
A   B
 \ / \
  C   D
   \ / \ 
    E   F
     \ /
      G
```
LCFPA(A, B) == G, while - if we take parents of A and B, their LCA(A, B) == E

I think this should be called 'lowest universal common ancestor', which reflects the fact we require the ancestor to be the "Adam" for A and B, or "the nearest common ancestor of A, B, and all their ancestors"

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -91,37 +91,80 @@ object DagOperations {
     })
   }
 
+  private val blockMetadataByNumDecreasing: Ordering[BlockMetadata] =
+    (l: BlockMetadata, r: BlockMetadata) => {
+      def compareByteString(l: ByteString, r: ByteString): Int =
+        l.hashCode().compareTo(r.hashCode())
```
ArturGajowy: I know you're not the author, but I find this a terrible hack:

1. given two byte strings, it's very unclear which one is going to be the 'smaller' one, as their hash is used for comparison,
2. this isn't always a tie breaker, as hashes can collide as well.

We should compare the byte strings lexicographically.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -91,37 +91,80 @@ object DagOperations {
...
+
+      val ln = l.blockNum
+      val rn = r.blockNum
+      rn.compare(ln) match {
```
ArturGajowy: I've spent 15 minutes looking for why this thing sorts in descending order. A comment saying 'notice the inverted order of compared items, which makes the ordering descending' would hopefully help me.

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,145 @@
...
+  val validatorKeyPairs = (1 to 3).map(_ => Secp256k1.newKeyPair)
+  val (_, validatorPks) = validatorKeyPairs.unzip
+  val validators        = validatorPks.map(pk => ByteString.copyFrom(pk.bytes))
+  val v1                = validators(0)
+  val v2                = validators(1)
+  val v3                = validators(2)
+  val genesisParameters = buildGenesisParameters(
+    validatorKeyPairs,
+    Map(validatorPks(0) -> 10, validatorPks(1) -> 10, validatorPks(2) -> 5)
+  )
```
ArturGajowy: We can replace all the above setup with:
`  val genesisParameters = buildGenesisParameters(bondsFunction = _.zip(List(10L, 10L, 5L)).toMap)`

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,145 @@
...
+   *        |  |
+   *       b2  b3
+   *         \ |
+   *           b1
+   *           |
+   *         genesis
+   */
+  "isFinalized" should "return true for ancestors of last finalized block" in effectTest {
+    HashSetCasperTestNode.networkEff(genesisContext, networkSize = 3).use {
+      case n1 +: n2 +: n3 +: _ =>
```
ArturGajowy: this can be

`case nodes @ n1 +: n2 +: n3 +: Seq()`

and then we don't need to declare `val nodes  = List(n1, n2, n3)` below

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
```diff
@@ -0,0 +1,145 @@
...
+   *       b2  b3 <- last finalized block
+   *         \ |
+   *           b1
+   *           |
+   *         genesis
+   */
+  it should "return false for children and siblings of last finalized block" in effectTest {
+    HashSetCasperTestNode.networkEff(genesisContext, networkSize = 3).use {
+      case n1 +: n2 +: n3 +: _ =>
+        import n3.{blockStore, cliqueOracleEffect, logEff}
```
ArturGajowy: why do we differ the 'focused' node between those tests?

I have a hard time understanding which node sees what based on the picture. Maybe each node should have its own swim lane?

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
ArturGajowy: Couldn't we somehow express the test with unitary weights? (all nodes have same weight?)

ArturGajowy (changes requested):
A couple comments :)

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
itegulov: You are right. The initial plan was to reuse LCA but, since our current implementation is not what I thought it was, it is better to just BFS here.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
itegulov: Renamed it to lowestUniversalCommonAncestor and clarified the comment a bit further

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
itegulov: Replaced this with basically `Ordering.Iterable[Byte]`

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
itegulov: The weights are now unitary

casper/src/test/scala/coop/rchain/casper/api/LastFinalizedAPITest.scala
itegulov: No particular reason. I changed it so that we always look from the n1 perspective.

Not sure what exactly do you mean by "swim lane" here but describing the whole state from all nodes POV would be overly verbose. Note that there is also "time" axis involved here: a node can get a block later than the other nodes (e.g. by missing it initially but fetching it as a dependency for a new one) which can affect node's decision on finalization. This is what happens in the second test (`it should "return false for children and siblings of last finalized block"`).

ArturGajowy: I meant sth similar to how our dag pngs look like: each validator has a swim-lane, time is reflected by by the happens-before relationship encoded by justifications. No big deal :)

ArturGajowy (approved):
Good job! Thx for the fixups :)

I've added a commit with some more test cases for LUCA, I felt that test kinda looked through fingers on the fact that there's U in LUCA.

itegulov:
@ArturGajowy Thanks! I just wanted to say that I really appreciate your thoroughness and amount of attention you pay when it comes to reviewing PRs.

ArturGajowy:
@itegulov oh stop it you

Thx! :heart: This means a lot to me :)

PR #2712, opened 2019-09-05
Fix peek replay

lukasz-golebiewski:
## Overview
Fixes non-determinism in replay when the same channel is peeked and consumed in the same deploy

### JIRA ticket:
Part of: https://rchain.atlassian.net/browse/RCHAIN-3816

### Notes

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -66,7 +66,11 @@ class DebruijnInterpreter[M[_], F[_]](
     fTell: FunctorTell[M, Throwable],
     cost: _cost[M]
 ) extends Reduce[M] {
-  type Application = Option[(TaggedContinuation, Seq[ListParWithRandom], Int, Boolean)]
+
+  type Application =
+    Option[
+      (TaggedContinuation, Seq[(Par, ListParWithRandom, ListParWithRandom, Boolean)], Int, Boolean)
+    ]
 
   /**
     * Materialize a send in the store, optionally returning the matched continuation.
@@ -81,20 +85,12 @@ class DebruijnInterpreter[M[_], F[_]](
       data: ListParWithRandom,
       persistent: Boolean,
       sequenceNumber: Int
-  ): M[Unit] = {
-    def go(res: Application): M[Unit] =
-      res match {
-        case Some((continuation, dataList, updatedSequenceNumber, peek)) =>
-          if (persistent || peek)
-            List(
-              dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
-              produce(chan, data, persistent, sequenceNumber)
-            ).parSequence_
-          else dispatcher.dispatch(continuation, dataList, updatedSequenceNumber)
-        case None => syncM.unit
-      }
-    space.produce(chan, data, persist = persistent, sequenceNumber) >>= (go(_))
-  }
+  ): M[Unit] =
+    space.produce(chan, data, persist = persistent, sequenceNumber) >>= (continue(
+      _,
+      produce(chan, data, persistent, sequenceNumber),
+      persistent
+    ))
 
   /**
     * Materialize a send in the store, optionally returning the matched continuation.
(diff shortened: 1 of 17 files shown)


lukasz-golebiewski:
bors try

lukasz-golebiewski:
bors retry

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -85,12 +89,15 @@ class DebruijnInterpreter[M[_], F[_]](
     def go(res: Application): M[Unit] =
       res match {
         case Some((continuation, dataList, updatedSequenceNumber, peek)) =>
-          if (persistent || peek)
-            List(
-              dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
+          if (persistent)
+            runInParallel(dispatch(continuation, dataList, updatedSequenceNumber))(
```
ghost: Factoring out `dispatch(continuation, dataList, updatedSequenceNumber)` would make future maintenance easier: just one expression to change instead of three.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -117,11 +124,14 @@ class DebruijnInterpreter[M[_], F[_]](
       res match {
         case Some((continuation, dataList, updatedSequenceNumber, _)) =>
           if (persistent)
-            List(
-              dispatcher.dispatch(continuation, dataList, updatedSequenceNumber),
+            runInParallel(dispatch(continuation, dataList, updatedSequenceNumber))(
```
ghost: Ditto.  These two code blocks are so similar that I'm tempted to have you pull them both out into a separate method, but they're so small that I don't think it's worth it.

ghost (approved):

dzajkowski (approved):

lukasz-golebiewski:
bors retry

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
lukasz-golebiewski: Done

lukasz-golebiewski:
bors retry

ArturGajowy (dismissed):
I think there's a hole in the approach implemented, please see https://rchain.atlassian.net/wiki/spaces/CORE/pages/753303627/Replay+issues?focusedCommentId=753860611#comment-753860611

lukasz-golebiewski:
Without this PR peek replay may fail randomly even for single parent blocks so I'm not sure why you are rejecting it. I'm going to reply to your concern in the wiki

ArturGajowy:
I rejected it, because while this PR prevents a class of 'unused comm events', it introduces another one. More importantly, it seemed as if a drastically approach was needed. You found a good way forward though, namely:

```
Łukasz Gołębiewski
This means that until we are able to merge tuplespaces peeks and (non-)linear operations on the same channels in different blocks will have to conflict.

This shouldn't be a problem for now, as P! P! isn't going to conflict, and this is the pattern that will be used to access the PoS contract via the registry

Artur Gajowy
Good point! So we revert the rule for `4! P!/!!` to "X Could have matched same produce. Mergeable if different linear produces or peek's produce is nonlinear." for now.
```

I'd though argue that since we know a flaw in this PRs approach, and know how to fix it, and its fixable separately, we should either incorporate the fix (modify the mering rule) or issue a separate PR the merging rule that is merged before this one.

In any case, we should add a test against that flaw in this PR
(two nodes deploy on top of s0, D1 = f0, D2 = p0, a block with justifications = D1, D2 is created and processed successfully (there are no 'umused comm events' errors, and - because of the changed merging rule - the top block has just one parent))

lukasz-golebiewski:
We do not have peek merging rules in dev yet

ArturGajowy:
This PR has also the nice property of invalidating send/receive sequence numbers, I think (now that the scala registry is gone). Fantastic! :clinking_glasses:

lukasz-golebiewski:
bors retry

PR #2713, opened 2019-09-05
Alter call sites to use non-leaky List and Either APIs

ghost:
## Overview
This is a followup to #2695, altering the call sites to use the new APIs that don't leak memory.  I recommend using `?w=1` when viewing the diff.  Also, there are no changes in functionality; I just moved code around and added a parameter for passing in variables that used to be closed over.

### DIFF
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -112,31 +112,34 @@ in {
               }
             } |
 
-            contract _getOrCreate(@revAddress, constructor, retCh) = {
-              new revAddressValidCh, revAddressValidEitherCh, getOrCreate in {
-                RevAddress!("validate", revAddress, *revAddressValidCh) |
-                @Either!("fromNillableError <-", *revAddressValidCh, *revAddressValidEitherCh) |
-                @Either!("flatMap <-", *revAddressValidEitherCh, *getOrCreate, *retCh) |
-                for (_, ret <- getOrCreate) {
-                  for (@vaultMap <- vaultMapStore) {
-                    match vaultMap.get(revAddress) {
-                      Nil => {
-                        new resCh in {
-                          constructor!(*resCh) |
-                          for (@eitherVault <- resCh) {
-                            match eitherVault {
-                              (true, vault) => vaultMapStore!(vaultMap.set(revAddress, vault))
-                              (false,  _    ) => vaultMapStore!(vaultMap)
-                            } |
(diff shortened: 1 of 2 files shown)


casper/src/main/resources/RevVault.rho
```diff
@@ -112,32 +112,35 @@ in {
...
                     }
                   }
                 }
+              } |
+
+              contract _getOrCreate(@revAddress, constructor, retCh) = {
+                new revAddressValidCh, revAddressValidEitherCh in {
+                  RevAddress!("validate", revAddress, *revAddressValidCh) |
+                  @Either!("fromNillableError <-", *revAddressValidCh, *revAddressValidEitherCh) |
+                  @Either!("flatMapClean <-", *revAddressValidEitherCh, *_getOrCreateCombinator, (revAddress, *constructor), *retCh)
```
ArturGajowy, line 142: I have big issues with how the end-user API changed in the previous PR (sorry for missing its review).  This method lost resemblence to flatMap to the point where it's really confusing what does what. Why does the combinator ignore its arguments?

I really don't feel comfortable with us merging this PR without having the benchmarks showing that self-respawning linear receive is indeed slower.

casper/src/main/resources/RevVault.rho
```diff
@@ -173,39 +176,43 @@ in {
...
-                  new amountPurseCh in {
-                    purse!("split", amount, *amountPurseCh) |
-                    @Either!("fromSingletonList <-", *amountPurseCh, "Insufficient funds", *retCh)
-                  }
-                } |
-
-                @Either!("flatMap <-", *eitherPurseCh, *doDeposit, *ret) |
-                for (@p, retCh <- doDeposit) {
-                  @{revAddress | bundle0{*_revVault}}!("_deposit", p, *retCh)
+            new split, doDeposit in {
```
ArturGajowy, line 179: please declare the combinators in the topmost scope possible to needless scope nesting and  indentation changes. This will also make the change clearer to reviewers

casper/src/main/resources/RevVault.rho
ArturGajowy: Since all of the combinators are supposed to be stateless (are they?), we should be able to put them in the top-level scope and never declare another copy of them past genesis.

9rb:
bors try

9rb:
@pyrocto will create a new PR with necessary changes.

PR #2714, opened 2019-09-06
Improve Casper code with EitherT

sebtomba:
## Overview
Improvements:
- Use of `EitherT` instead of `joinRight.traverse`
- Make the `BlockStatus` _is in dag_ logic explicit instead of using inheritance
- Make the `BlockStatus` _Slashable_ logic explicit instead of using inheritance
- Unify Casper functions return types to `ValidBlockProcessing`
- Improve code readability in Casper
- Change Casper `foldM` to `tailRecM` in `InterpreterUtil` and `RuntimeManager`

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3616

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
@@ -1,81 +1,33 @@
 package coop.rchain.casper
 
 sealed trait BlockStatus
-
-sealed trait BlockError extends BlockStatus
-
-sealed trait InvalidBlock extends BlockError
-
-sealed trait Slashable
-
-sealed trait ValidBlock extends BlockStatus
-
-case object Valid extends ValidBlock
-
-case object Processing extends BlockStatus
-
-final case class BlockException(ex: Throwable) extends BlockError
-
-// AdmissibleEquivocation are blocks that would create an equivocation but are
-// pulled in through a justification of another block
-case object AdmissibleEquivocation extends InvalidBlock with Slashable
-// TODO: Make IgnorableEquivocation slashable again and remember to add an entry to the equivocation record.
-// For now we won't eagerly slash equivocations that we can just ignore,
-// as we aren't forced to add it to our view as a dependency.
-// TODO: The above will become a DOS vector if we don't fix.
-case object IgnorableEquivocation extends InvalidBlock
-case object MissingBlocks         extends InvalidBlock
-
-sealed trait InvalidUnslashableBlock extends InvalidBlock
-case object InvalidFormat            extends InvalidUnslashableBlock
-case object InvalidSignature         extends InvalidUnslashableBlock
-case object InvalidSender            extends InvalidUnslashableBlock
-case object InvalidVersion           extends InvalidUnslashableBlock
-case object InvalidTimestamp         extends InvalidUnslashableBlock
-
-case object InvalidBlockNumber      extends InvalidBlock with Slashable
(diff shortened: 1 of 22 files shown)


casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
```diff
@@ -1,61 +1,107 @@
 package coop.rchain.casper
 
-sealed trait BlockStatus {
-  val inDag: Boolean
-}
+sealed trait BlockStatus
+object BlockStatus {
+  def valid: ValidBlock                    = ValidBlock.Valid
+  def processing: BlockError               = BlockError.Processing
+  def exception(ex: Throwable): BlockError = BlockError.BlockException(ex)
```
ArturGajowy, line 7: can we get rid of this constructor? I think the only usage is caused by this line:
https://github.com/rchain/rchain/blob/6e0e557de1eb4eaf03be2826502270b326265023/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala#L240-L242

So we replace this with `InvalidBlock.ReplayFailed` and we can get rid of this, I think

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -290,11 +290,12 @@ object Genesis {
     for {
       _       <- skFiles
       printer <- Sync[F].delay { new PrintWriter(genBondsFile) }
-      _ <- Foldable[List].foldM[F, (PublicKey, Long), Unit](bonds.toList, ()) {
-            case (_, (pub, stake)) =>
+      _ <- bonds.toList.traverse {
```
ArturGajowy: I think `traverse_` will be a bit more typesafe

ArturGajowy (approved):
Neat refactorings! Make code much more readable. Especially the 'replace joinRight-traverse with EtiherT` part.

One point to consider in the future: ValidBlock carries no values and is isomorphic to Unit. We should either make it carry the 'valid block value' (or the partial results used for constructing it), or making BlockProcessing an Either[BlockError, Unit]

Some minor remarks otherwise

casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
sebtomba: There is, unfortunately, a second usage in `InterpreterUtil.processPossiblePreStateHash`.
Even worse there are 17 matches on `Sync[F].raiseError` in Casper. I created a ticket to deal with this _unworthiness_ [RCHAIN-3818](https://rchain.atlassian.net/browse/RCHAIN-3818).

sebtomba:
@ArturGajowy Yes, `ValidBlock` is isomorphic to `Unit`. But there is some logic attached to this `trait` in `BlockStatus.isInDag`. It's not easy to come up with a good replacement for this behavior.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
sebtomba: Good catch! :)

KentShikama (approved):
Ideally we could also remove the BlockError trait

sebtomba:
@KentShikama I don't think this is possible since all `InvalidBlock`s are in the DAG. We need the `BlockError` trait for all errors where the block is not (yet) in the DAG.

PR #2723, opened 2019-09-11
Make DeployGrpcService.visualizeDag stack safe

sebtomba:
## Overview
Make DeployGrpcService.visualizeDag stack safe

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3236

### DIFF
--- a/graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
+++ b/graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
@@ -2,22 +2,21 @@ package coop.rchain.graphz
 
 import java.io.FileOutputStream
 
-import cats._, cats.data._, cats.implicits._
+import cats._
 import cats.effect.Sync
+import cats.implicits._
 import cats.mtl._
 
 trait GraphSerializer[F[_]] {
   def push(str: String, suffix: String = "\n"): F[Unit]
 }
 
-class StringSerializer[F[_]: Applicative: MonadState[?[_], StringBuffer]]
-    extends GraphSerializer[F] {
+class StringSerializer[F[_]: MonadState[?[_], StringBuffer]] extends GraphSerializer[F] {
   def push(str: String, suffix: String): F[Unit] =
     MonadState[F, StringBuffer].modify(sb => sb.append(str + suffix))
 }
 
-class ListSerializer[F[_]: Applicative: MonadState[?[_], Vector[String]]]
-    extends GraphSerializer[F] {
+class ListSerializer[F[_]: MonadState[?[_], Vector[String]]] extends GraphSerializer[F] {
   def push(str: String, suffix: String): F[Unit] =
     MonadState[F, Vector[String]].modify(_ :+ (str + suffix))
 }
@@ -184,7 +183,7 @@ class Graphz[F[_]: Monad](gtype: GraphType, t: String)(implicit ser: GraphSerial
       arrowHead: Option[GraphArrowType] = None,
       constraint: Option[Boolean] = None
   ): F[Unit] = {
-    import Graphz.{showArrowType, showShape, showStyle}
+    import Graphz.{showArrowType, showStyle}
     val attrStyle: Map[String, String] = style.map(s => Map("style" -> s.show)).getOrElse(Map.empty)
     val attrArrowHead: Map[String, String] =
       arrowHead.map(s => Map("arrowhead" -> s.show)).getOrElse(Map.empty)
--- a/graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
+++ b/graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
@@ -1,13 +1,15 @@
 package coop.rchain.graphz
 
-import cats._, cats.data._, cats.implicits._
-import cats.mtl._
+import cats._
+import cats.data._
+import cats.implicits._
 import cats.mtl.implicits._
+
 import org.scalatest._
 
 class GraphzSpec extends FunSpec with Matchers with BeforeAndAfterEach with AppendedClues {
 
-  type Effect[A] = StateT[Id, StringBuffer, A]
+  type Effect[A] = State[StringBuffer, A]
 
   implicit val ser = new StringSerializer[Effect]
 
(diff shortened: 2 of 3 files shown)


graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -10,16 +10,14 @@ trait GraphSerializer[F[_]] {
...
-    MonadState[F, StringBuffer].modify(sb => sb.append(str + suffix))
+    Defer[F].defer(MonadState[F, StringBuffer].modify(sb => sb.append(str + suffix)))
 }
 
-class ListSerializer[F[_]: Applicative: MonadState[?[_], Vector[String]]]
-    extends GraphSerializer[F] {
+class ListSerializer[F[_]: Defer: MonadState[?[_], Vector[String]]] extends GraphSerializer[F] {
   def push(str: String, suffix: String): F[Unit] =
-    MonadState[F, Vector[String]].modify(_ :+ (str + suffix))
+    Defer[F].defer(MonadState[F, Vector[String]].modify(_ :+ (str + suffix)))
```
ArturGajowy: I don't think the `defer`-s are needed here. Unless you've seen this fail with Eval and without the `defer`-s?

graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
```diff
@@ -1,13 +1,14 @@
...
+import cats._
+import cats.data._
 import cats.mtl.implicits._
+
 import org.scalatest._
 
 class GraphzSpec extends FunSpec with Matchers with BeforeAndAfterEach with AppendedClues {
 
-  type Effect[A] = StateT[Id, StringBuffer, A]
+  type Effect[A] = StateT[Eval, StringBuffer, A]
```
ArturGajowy: Why not Coeval? Monix docs have:

```
at the moment of writing this, the Monix Coeval takes care of error handling, while the Cats Eval does not, providing operations for recovery, thus also working well as a replacement for Scala’s Try type.
```

How does Eval handle errors in the wrapped code? Will `.value` simply throw an exception?

ArturGajowy (approved):

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
sebtomba: @ArturGajowy You are right.

graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
sebtomba: @ArturGajowy I see your point. But let's stick to `cats` for now. All of our current `GraphSerializer` instances are pure and won't throw any (non-fatal) exceptions.

marcin-rzeznicki: Irrelevant side note: `StateT[Eval, A, B]` is just `State[A, B]`

ArturGajowy: @marcin-rzeznicki are you sure that State[A, B] is stacksafe?

marcin-rzeznicki: Yeah: 
https://github.com/typelevel/cats/blob/8c031e5a4ed63a8efc34c196f95ccfcfea7eeb77/core/src/main/scala/cats/data/package.scala#L58

https://github.com/typelevel/cats/blob/ed99e170ee65dfe9d03995dd72b33ec90ae9f926/tests/src/test/scala/cats/tests/IndexedStateTSuite.scala#L40

marcin-rzeznicki: Had it used `Id` then obviously it wouldn't have been stack safe. It strongly relies on stack-safety of the underlying `F` ie.
https://github.com/typelevel/cats/blob/ed99e170ee65dfe9d03995dd72b33ec90ae9f926/core/src/main/scala/cats/data/IndexedStateT.scala#L25

ArturGajowy: Oh, cool! Wasn't aware of the Eval in there :)

graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
```diff
@@ -334,14 +336,21 @@ class GraphzSpec extends FunSpec with Matchers with BeforeAndAfterEach with Appe
...
     }
 
+    it("Huge graph") { // test for a stack overflow
+      val graph = for {
+        g <- Graphz[Effect]("G", DiGraph)
+        _ <- (1 to 1000).toList.traverse(i => g.edge(s"e$i" -> s"e${i + 1}"))
+        _ <- g.close
+      } yield g
+      graph.show // ignore
+    }
```
marcin-rzeznicki, line 346: :+1:

node/src/main/scala/coop/rchain/node/api/DeployGrpcService.scala
```diff
@@ -57,10 +57,10 @@ object DeployGrpcService {
         defer(BlockAPI.getBlock[F](q))
 
       override def visualizeDag(q: VisualizeDagQuery): Observable[GrpcEither] = {
-        type Effect[A] = StateT[Id, Vector[String], A]
+        type Effect[A] = StateT[Eval, Vector[String], A]
```
marcin-rzeznicki: Right - it's the same as `type Effect[A] = State[Vector[String], A]` so maybe you could simplify it

graphz/src/test/scala/coop/rchain/graphz/GraphzSpec.scala
marcin-rzeznicki: Maybe simplify the type to `State[Sth, A]`

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -2,22 +2,21 @@ package coop.rchain.graphz
...
-class StringSerializer[F[_]: Applicative: MonadState[?[_], StringBuffer]]
-    extends GraphSerializer[F] {
+class StringSerializer[F[_]: MonadState[?[_], StringBuffer]] extends GraphSerializer[F] {
   def push(str: String, suffix: String): F[Unit] =
     MonadState[F, StringBuffer].modify(sb => sb.append(str + suffix))
 }
 
-class ListSerializer[F[_]: Applicative: MonadState[?[_], Vector[String]]]
-    extends GraphSerializer[F] {
+class ListSerializer[F[_]: MonadState[?[_], Vector[String]]] extends GraphSerializer[F] {
```
marcin-rzeznicki, line 19: Why do we actually need it here? :thinking: And why does a serializer return `Unit`? :thinking:

marcin-rzeznicki (approved):

PR #2724, opened 2019-09-11
RCHAIN-3365: Use per-validator vaults in PoS

tymm:
## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -429,13 +429,16 @@ new  PoS,
       } |
 
       contract pickActiveValidators(@allBonds, @withdrawers, returnCh) = {
-        new availableValidatorsCh, isAvailable in {
+        new availableValidatorsCh, isAvailable, pksCh in {
           @ListOps!("filter", allBonds.toList(), *isAvailable, *availableValidatorsCh) |
           contract isAvailable(@(pk, _), resultCh) = { resultCh!(not withdrawers.contains(pk))} |
 
           for (@availableValidators <- availableValidatorsCh) {
-            //TODO choose only a limited number of validators
-            @ListOps!("map", availableValidators, *fst, *returnCh)
+            @ListOps!("map", availableValidators, *fst, *pksCh) |
+            for (@pks <- pksCh) {
+              // TODO: Randomly select 100 active validators once we have on-chain randomness
+              returnCh!(pks.take($$numberOfActiveValidators$$))
+            }
           }
         }
       } |
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -176,8 +176,7 @@ new  PoS,
                     resultCh!(
                       state.set("withdrawers",
                                 state.get("withdrawers")
-                                     .set(userPk,
-                                       $$quarantineLength$$ + $$epochLength$$ * (1 + blockNumber / $$epochLength$$))),
+                                     .set(userPk, $$quarantineLength$$ + $$epochLength$$ + blockNumber)),
                       (true, Nil))
                   } else {
                     resultCh!(state, (false, "User is not bonded"))
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -33,6 +33,7 @@ new  PoS,
     runMVar,
     fst,
     deposit, distributeRewards,
+    perActiveValidatorPosVaultsCh,
     rs(`rho:registry:insertSigned:secp256k1`),
     uriOut in {
   registryLookup!(`rho:rchain:revVault`, *revVaultCh) |
@@ -44,6 +45,25 @@ new  PoS,
 
       @RevVault!("deployerAuthKey", *posAuthKeyCh) |
       getCurrentUserAddress!(*posRevAddressCh) |
+
+      new rangeCh, makeVault in {
+        contract makeVault(_, returnCh) = {
+          new unf, revAddressCh, vaultCh in {
+            revAddressOps!("fromUnforgeable", *unf, *revAddressCh) |
+            for (@revAddress <- revAddressCh) {
+              @RevVault!("findOrCreate", revAddress, *vaultCh) |
+              for (@(true, vault) <- vaultCh) {
+                returnCh!((*unf, revAddress, vault))
+              }
+            }
+          }
+        } |
+
+        @ListOps!("range", 0, $$numberOfActiveValidators$$, *rangeCh) |
+        for (@range <- rangeCh) {
+          @ListOps!("parMap", range, *makeVault, *perActiveValidatorPosVaultsCh)
+        }
+      } |
 
       for(@posRevAddress <- posRevAddressCh;
           @posAuthKey <- posAuthKeyCh) {
@@ -77,6 +97,15 @@ new  PoS,
               }
             } |
 
+            contract PoS (@"getActiveValidators", returnCh) = {
+              new tmpCh in {
+                getMVar!(*stateCh, *tmpCh) |
+                for (@state <- tmpCh) {
+                  returnCh!(state.get("activeValidators"))
+                }
+              }
+            } |
+
             /**
              * Returns a Map[PublidKey, Int] containing the rewards accumulated for each validator.
              * The returned map contains only the committed rewards after the last closeBlock
@@ -104,7 +133,7 @@ new  PoS,
                   } else if (amount > $$maximumBond$$) {
                     resultCh!(state, (false, "Bond is greater than maximum!"))
                   } else {
-                    deposit!(userPk, amount, *depositCh) |
+                    deposit!(userPk, amount, state.get("activeValidators"), *depositCh) |
                     for(@depositResult <- depositCh) {
                       match depositResult {
                         (true, _) => {
@@ -132,7 +161,6 @@ new  PoS,
                   blockDataCh
               in {
                 runMVar!(*stateCh, *processCh, *returnCh) |
-
                 getBlockData!(*blockDataCh) |
                 getUser!(*userCh) |
                 for(@state, resultCh <- processCh;
@@ -161,7 +189,7 @@ new  PoS,
                 runMVar!(*stateCh, *processPayCh, *returnCh) |
                 for(@state, payResultCh <- processPayCh;
                     @userPk <- userCh) {
-                  deposit!(userPk, amount, *depositCh) |
+                  deposit!(userPk, amount, state.get("activeValidators"), *depositCh) |
 
                   distributeRewards!(
                     amount,
@@ -207,13 +235,15 @@ new  PoS,
             } |
 
             contract PoS(@"closeBlock", ackCh) = {
-              new commitRewards,
+              new accumulateDeposits,
+                  commitRewards,
                   changeEpoch,
                   removeQuarantinedWithdrawers,
                   payValidators
               in {
                 new blockDataCh, mvarProcessCh,
                     getBlockData(`rho:block:data`),
+                    accDepositsDoneCh,
                     commitRewardsCh,
                     newValidatorsCh, removeQuarantinedCh, paymentDoneCh
                 in {
@@ -222,6 +252,9 @@ new  PoS,
                   runMVar!(*stateCh, *mvarProcessCh, *ackCh) |
                   for(_, @blockNumber, _ <- blockDataCh;
                       @state, mvarResultCh <- mvarProcessCh) {
+                    // Before choosing new active validators and paying out any withdrawers,
+                    // deposits have to be accumulated into one vault for now.
+                    accumulateDeposits!(posRevAddress, *accDepositsDoneCh) |
 
                     changeEpoch!(
                       blockNumber,
@@ -261,6 +294,33 @@ new  PoS,
                           },
                           Nil)
                       }
+                    }
+                  }
+                } |
+
+                new moveFunds in {
+                  contract moveFunds(@(unf, _, vault), returnCh) = {
+                    new balanceCh, unfAuthKeyCh, transferDoneCh in {
+                      @vault!("balance", *balanceCh) |
+                      for (@balance <- balanceCh) {
+                        if (balance > 0) {
+                          @RevVault!("unforgeableAuthKey", unf, *unfAuthKeyCh) |
+                          for (@authKey <- unfAuthKeyCh) {
+                            @vault!("transfer", posRevAddress, balance, authKey, *transferDoneCh) |
+                            for (@(true, Nil) <- transferDoneCh) {
+                              returnCh!(true)
+                            }
+                          }
+                        } else {
+                          returnCh!(true)
+                        }
+                      }
+                    }
+                  } |
+
+                  contract accumulateDeposits(destRevAddress, ackCh) = {
+                    for(@vaults <<- perActiveValidatorPosVaultsCh) {
+                      @ListOps!("foreach", vaults, *moveFunds, *ackCh)
                     }
                   }
                 } |
@@ -344,17 +404,27 @@ new  PoS,
           }
         } |
 
-        contract deposit (@fromUserPk, @amount, returnCh) = {
+        contract deposit (@fromUserPk, @amount, @activeValidators, returnCh) = {
           new vaultCh,
               revAddressCh,
+              indexSender,
+              getBlockData(`rho:block:data`), blockDataCh,
               authKeyCh in {
             @RevVault!("deployerAuthKey", *authKeyCh) |
-            for (@authKey <- authKeyCh) {
-              revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
-              for (@fromRevAddress <- revAddressCh) {
-                @RevVault!("findOrCreate", fromRevAddress, *vaultCh) |
-                for (@(true, fromVault) <- vaultCh) {
-                  @fromVault!("transfer", posRevAddress, amount, authKey, *returnCh)
+            revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
+            for (@authKey <- authKeyCh; @fromRevAddress <- revAddressCh) {
+              @RevVault!("findOrCreate", fromRevAddress, *vaultCh) |
+              getBlockData!(*blockDataCh) |
+              for (@(true, fromVault) <- vaultCh; _, _, @sender <- blockDataCh) {
+                @ListOps!("indexOf", activeValidators, sender, *indexSender) |
+                for (@idx <- indexSender) {
+                  for (@vaults <<- perActiveValidatorPosVaultsCh) {
+                    match vaults.nth(idx) {
+                      (_, activeValidatorVault, _) => {
+                        @fromVault!("transfer", activeValidatorVault, amount, authKey, *returnCh)
+                      }
+                    }
+                  }
                 }
               }
             }
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -42,7 +42,6 @@ new  PoS,
   for(@(_, RevVault) <- revVaultCh;
       @(_, ListOps) <- listOpsCh) {
     new posRevAddressCh, posAuthKeyCh, posVaultCh, initialActiveCh in {
-
       @RevVault!("deployerAuthKey", *posAuthKeyCh) |
       getCurrentUserAddress!(*posRevAddressCh) |
 
@@ -103,6 +102,12 @@ new  PoS,
                 for (@state <- tmpCh) {
                   returnCh!(state.get("activeValidators"))
                 }
+              }
+            } |
+
+            contract PoS (@"getActiveValidatorVaults", returnCh) = {
+              for(@vaults <<- perActiveValidatorPosVaultsCh) {
+                returnCh!(vaults)
               }
             } |
 
@@ -279,7 +284,8 @@ new  PoS,
                     } |
 
                     for(@newValidators <- newValidatorsCh;
-                        @(payments, newBonds, newWithdrawers, newRewards) <- removeQuarantinedCh) {
+                        @(payments, newBonds, newWithdrawers, newRewards) <- removeQuarantinedCh;
+                        _ <- accDepositsDoneCh) {
 
                       payValidators!(payments, *paymentDoneCh) |
 
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -137,7 +137,7 @@ new  PoS,
                 getUser!(*userCh) |
                 for(@state, resultCh <- processCh;
                     @userPk <- userCh;
-                    _, @blockNumber <- blockDataCh) {
+                    _, @blockNumber, _ <- blockDataCh) {
                   if (state.get("allBonds").contains(userPk)) {
                     // the withdwal is in effect starting from the next epoch
                     resultCh!(
@@ -220,7 +220,7 @@ new  PoS,
                   getBlockData!(*blockDataCh) |
 
                   runMVar!(*stateCh, *mvarProcessCh, *ackCh) |
-                  for(_, @blockNumber <- blockDataCh;
+                  for(_, @blockNumber, _ <- blockDataCh;
                       @state, mvarResultCh <- mvarProcessCh) {
 
                     changeEpoch!(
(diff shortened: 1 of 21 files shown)


casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -193,6 +194,7 @@ object BlockCreator {
       deploys: Seq[DeployData],
       justifications: Seq[Justification],
       maxBlockNumber: Long,
+      sender: PublicKey,
```
odeac, line 204: I feel that this should be named "creator" instead of sender?

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1405,6 +1428,7 @@ class DebruijnInterpreter[M[_], F[_]](
       "size"        -> size,
       "length"      -> length,
       "slice"       -> slice,
+      "take"        -> take,
```
odeac, line 1430: I'm not sure this should be a method or a contract in ListOps.rho. My feeling is that, since map is a contract in ListOps.rho then take should be too

casper/src/test/resources/RevAddressTest.rho
```diff
@@ -79,7 +79,9 @@ in {
     new unf1, unf2, r1, r2 in {
       revAddress!("fromUnforgeable", *unf1, *r1) |
       revAddress!("fromUnforgeable", *unf2, *r2) |
-      rhoSpec!("assert", (*r1, "!= <-", *r2), "expecting RevAddresses to be unequal", *ackCh)
+      for (@fromUnf1 <- r1) {
+        rhoSpec!("assert", (fromUnf1, "!= <-", *r2), "expecting RevAddresses to be unequal", *ackCh)
```
odeac, line 87: I would also add a positive test like deriving twice from the same unforgeable name returns the same address

casper/src/test/resources/PoSTest.rho
```diff
@@ -97,15 +97,23 @@ match (
...
-                  [
-                    (true, "all good here")
-                  ],
-                  *ackCh
-                )
+                // After the first "closeBlock" is done, there are active validators
+                // For testing we need one of the active validators to be the block sender
+                @PoS!("getActiveValidators", *activeValidatorsCh) |
+                for (@activeValidators <- activeValidatorsCh) {
+                  setBlockData!("sender", activeValidators.nth(0), *blockDataSet) |
```
odeac: why do you add this assertion to "test_close_block_finishes"? Are they related?

I think a better place would be to add this assert to the test of getBlockData

casper/src/main/resources/PoS.rhox
```diff
@@ -429,13 +429,15 @@ new  PoS,
...
+        new availableValidatorsCh, isAvailable, pksCh in {
           @ListOps!("filter", allBonds.toList(), *isAvailable, *availableValidatorsCh) |
           contract isAvailable(@(pk, _), resultCh) = { resultCh!(not withdrawers.contains(pk))} |
 
           for (@availableValidators <- availableValidatorsCh) {
-            //TODO choose only a limited number of validators
-            @ListOps!("map", availableValidators, *fst, *returnCh)
+            @ListOps!("map", availableValidators, *fst, *pksCh) |
+            for (@pks <- pksCh) {
+              returnCh!(pks.take($$numberOfActiveValidators$$))
```
odeac, line 515: This is not a fair random selection. I think we should allow all the validators be active in the first epoch and then, in the second epoch, when we have a random number produced from all the validators, we can fairly select 100 of them

casper/src/main/resources/PoS.rhox
```diff
@@ -45,6 +46,25 @@ new  PoS,
...
+              for (@(true, vault) <- vaultCh) {
+                returnCh!((*unf, revAddress, vault))
+              }
+            }
+          }
+        } |
+
+        @ListOps!("range", 0, $$numberOfActiveValidators$$, *rangeCh) |
+        for (@range <- rangeCh) {
+          @ListOps!("map", range, *makeVault, *perActiveValidatorPosVaultsCh)
```
odeac: you can use "parMap" here

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,25 @@ new  PoS,
...
               authKeyCh in {
             @RevVault!("deployerAuthKey", *authKeyCh) |
+            getBlockData!(*blockDataCh) |
             for (@authKey <- authKeyCh) {
               revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
               for (@fromRevAddress <- revAddressCh) {
                 @RevVault!("findOrCreate", fromRevAddress, *vaultCh) |
-                for (@(true, fromVault) <- vaultCh) {
-                  @fromVault!("transfer", posRevAddress, amount, authKey, *returnCh)
+                for (@(true, fromVault) <- vaultCh; _, _, @sender <- blockDataCh) {
```
odeac: I would put the receive from blockDataCh closer to the getBlockData!(*blockDataCh) for better readability. One next to the other is the easier to match visually

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -13,17 +13,25 @@ import coop.rchain.casper.util.Sorting.byteArrayOrdering
...
 import coop.rchain.shared.{Log, LogSource, Time}
 
 final case class Genesis(
     shardId: String,
     timestamp: Long,
     proofOfStake: ProofOfStake,
     vaults: Seq[Vault],
-    supply: Long
+    supply: Long,
+    publicKey: PublicKey
```
JosephDenman: It looks like this field is only accessed to do the assignment `publicKey = Genesis.pk`. It also looks like the only reason we need this field in the first place is so that a valid key is returned in `getBlockData`. Since this data is going to be over-written upon the first block, and (if I'm not mistaken) we don't need it anywhere outside of Rholang. So, could we just `Secp256k1.newKeyPair._2` in `computeGenesis` in `RuntimeManager`?

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,25 @@ new  PoS,
...
+                  @ListOps!("indexOf", activeValidators, sender, *indexSender) |
+                  for (@idx <- indexSender) {
+                    for (@vaults <<- perActiveValidatorPosVaultsCh) {
+                      @fromVault!("transfer", vaults.nth(idx).nth(1), amount, authKey, *returnCh)
```
odeac: Can you explain why is this needed "vaults.nth(idx).nth(1)"?

odeac (changes requested):

casper/src/test/resources/PoSTest.rho
```diff
@@ -97,15 +97,23 @@ match (
...
               for ( _ <- resultCh) {
-                rhoSpec!("assertMany",
-                  [
-                    (true, "all good here")
-                  ],
-                  *ackCh
-                )
+                // After the first "closeBlock" is done, there are active validators
+                // For testing we need one of the active validators to be the block sender
+                @PoS!("getActiveValidators", *activeValidatorsCh) |
```
JosephDenman: This method isn't added until `8199c63e`. This should probably be introduced after or with that commit.

JosephDenman (changes requested):
Still trying to grok the updated PoS logic. Will continue the review tomorrow.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
tymm: Since this directly relates to the `sender` field in `BlockMessage`, I feel like `sender` is the better name. Happy to change it if @JosDenmark also prefers "creator" :)

casper/src/test/resources/PoSTest.rho
tymm: For the other tests to work, we have to set the block sender to one of the active validators.
Since we potentially pick a new set of active validators in `PoS!closeBlock`, we have to set the sender after running `PoS!closeBlock`.

I just realized that we should do that after each call to `PoS!closeBlock`. So I guess I should create a helper contract that does that and replace every occurrence of `@PoS!("closeBlock", *closeBlockCh)` in `PoSTest.rho` with the helper contract.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
JosephDenman: I'm indifferent.

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,25 @@ new  PoS,
...
 
-        contract deposit (@fromUserPk, @amount, returnCh) = {
+        contract deposit (@fromUserPk, @amount, @activeValidators, returnCh) = {
           new vaultCh,
               revAddressCh,
+              indexSender,
+              getBlockData(`rho:block:data`), blockDataCh,
               authKeyCh in {
             @RevVault!("deployerAuthKey", *authKeyCh) |
+            getBlockData!(*blockDataCh) |
```
JosephDenman: This code would be more readable if the send on `getBlockData` was in the same lexical scope as the receive on `getBlockData`.

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,25 @@ new  PoS,
...
         } |
 
-        contract deposit (@fromUserPk, @amount, returnCh) = {
+        contract deposit (@fromUserPk, @amount, @activeValidators, returnCh) = {
           new vaultCh,
               revAddressCh,
+              indexSender,
+              getBlockData(`rho:block:data`), blockDataCh,
               authKeyCh in {
             @RevVault!("deployerAuthKey", *authKeyCh) |
```
JosephDenman, line 418: I think this should be run concurrently with `revAddressOps!("fromPublicKey", ...)`. Not only does it increase concurrency, it also makes data dependencies syntactically evident. You should also review the Rholang you added to ensure that this is generally true. Same with the lexical scope comment below.

JosephDenman (commented):
I know this isn't part of this PR, but while you're at it, would you change `$$quarantineLength$$ + $$epochLength$$ * (1 + blockNumber / $$epochLength$$)))` in `PoS` to `$$quarantineLength$$ + $$epochLength$$  + blockNumber`? It's also not immediately obvious to me why this is the block number at which a validator should be able to cash out. @odeac Please shed some light so the description can be included in a comment.

casper/src/main/resources/PoS.rhox
tymm: I added a `match` to make it a little easier to understand.

`vaults.nth(idx)` is a triple consisting of an unforgeable name, a rev address derived from that unforgeable name and a vault instance derived from the rev address.
`vaults.nth(idx).nth(1)` is therefore needed to get the rev address of the vault that is associated to the current block sender.

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +411,25 @@ new  PoS,
...
+            getBlockData!(*blockDataCh) |
             for (@authKey <- authKeyCh) {
               revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
               for (@fromRevAddress <- revAddressCh) {
                 @RevVault!("findOrCreate", fromRevAddress, *vaultCh) |
-                for (@(true, fromVault) <- vaultCh) {
-                  @fromVault!("transfer", posRevAddress, amount, authKey, *returnCh)
+                for (@(true, fromVault) <- vaultCh; _, _, @sender <- blockDataCh) {
```
lukasz-golebiewski: Joined channels are more prone to cause conflicts because the mergeability rules are not applied to them. 
I'm not sure if it makes sense to get rid of this particular join, just double-checking.

tymm: Hmm, why "more prone"? In which cases are they not conflicting?

lukasz-golebiewski: When disparate channels are used in both blocks

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
odeac: Sender makes sense in the context of message routing. When discussing blocks in the DAG, for me "sender" doesn't make sense anymore. Anyway, I don't have strong opinions about this

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
tymm: `Secp256k1.newKeyPair._2` doesn't work out since during replay we will get another value. And in general we want to keep things referentially transparent. Instead I changed it to `PublicKey(Array[Byte]())`.

casper/src/main/resources/PoS.rhox
tymm: Then this join should not be conflicting since new channels get created with every contract call.

tymm:
bors try

tymm:
@odeac and @JosDenmark please have another look. If you prefer the squashed commit log before approving, let me know.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tymm: It looks like the performance hit of using a ListOps version of take is so big that integration tests fail due to timeouts. Also I had to increase the timeout for the `PoSSpec`.

JosephDenman:
@tymm The squashed commit log would be much easier to review. Thanks.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: Does this mean that the Rholang implementation of ListOps is not really usable?

Do we have a rule about putting list operations in the list methods or in the ListOps contract?

tymm:
@JosDenmark done.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tymm: Let me take that back. Actually integration tests failing because of timeouts is probably also due to the overall performance decrease that comes with the new PoS logic.
However `PoSSpec` passed on my laptop with a 120s timeout before the change to `ListOps!take`. Which wasn't the case afterwards.

There was a discussion between Joseph, Artur and Mike in #rholang and I think everyone agreed that methods are not the way to go.

JosephDenman: @tymm That was a slightly different conversation. It was a debate to determine whether the cryptographic functions should be methods or system processes. I would go with adding `take` as a method alongside `slice` and `nth`. It's bound to be more performant and won't introduce unnecessary bloat in the tuplespace. It's also syntactically cleaner.

casper/src/main/resources/PoS.rhox
odeac: do you really need the take here? Can't you just use all the pks as active validators?

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,29 @@ new  PoS,
...
+                  @ListOps!("indexOf", activeValidators, sender, *indexSender) |
+                  for (@idx <- indexSender) {
+                    for (@vaults <<- perActiveValidatorPosVaultsCh) {
+                      match vaults.nth(idx).nth(1) {
+                        activeValidatorVault => {
+                          @fromVault!("transfer", activeValidatorVault, amount, authKey, *returnCh)
+                        }
+                      }
```
odeac, line 430: this looks better as it makes it obvious that the second nth is applied on a tuple

```scala
                     match vaults.nth(idx) {
                        (activeValidatorVault, _) => {
                          @fromVault!("transfer", activeValidatorVault, amount, authKey, *returnCh)
                        }
                      }
```

casper/src/test/resources/PoSTest.rho
```diff
@@ -463,6 +465,42 @@ match (
...
+              }
+            }
+          } |
+
+          contract closeBlock(ackCh) = {
+            new ackCh0, activeValidatorsCh, setBlockData(`rho:test:block:data:set`), blockDataSet in {
+              @PoS!("closeBlock", *ackCh0) |
+              @PoS!("getActiveValidators", *activeValidatorsCh) |
+              for (@ack <- ackCh0; @activeValidators <- activeValidatorsCh) {
+                // After the first "closeBlock" is done, there are active validators
```
odeac: I don't understand this comment. There should be active validators all the time, not only after the first closeBlock.

casper/src/main/resources/PoS.rhox
```diff
@@ -77,6 +96,21 @@ new  PoS,
...
+            contract PoS (@"getActiveValidators", returnCh) = {
+              new tmpCh in {
+                getMVar!(*stateCh, *tmpCh) |
+                for (@state <- tmpCh) {
+                  returnCh!(state.get("activeValidators"))
+                }
+              }
+            } |
+
+            contract PoS (@"getActiveValidatorVaults", returnCh) = {
```
odeac, line 108: I don't think this should be part of the PoS API. The fact that we have a vault per validator should be only an implementation detail

casper/src/test/resources/PoSTest.rho
```diff
@@ -467,6 +470,42 @@ match (
...
+
+              @posVault!("balance", *posVaultBalanceCh) |
+              for (@posVaultBalance <- posVaultBalanceCh) {
+                @PoS!("getActiveValidatorVaults", *vaultsCh) |
+                for (@vaults <- vaultsCh) {
+                  @ListOps!("fold", vaults, posVaultBalance, *accBalances, *retCh)
+                }
+              }
+            }
+          } |
```
odeac, line 494: I think that the logic of the getPoSBalance contract should be part of the PoS API

odeac (changes requested):
I still have a problem with the fact that the vault-per-validator implementation is visible in the API. Do we really need that?

casper/src/test/scala/coop/rchain/casper/helper/BlockDataContract.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  def set[F[_]: Concurrent: Span](
+      ctx: SystemProcess.Context[F]
+  )(message: (Seq[ListParWithRandom], Int)): F[Unit] = {
+
+    val isContractCall = new ContractCall(ctx.space, ctx.dispatcher)
+    message match {
+      case isContractCall(
+          produce,
+          Seq(RhoType.String("sender"), (RhoType.ByteArray(pk)), ackCh)
```
JosephDenman: Nit: extra parenthesis

casper/src/test/resources/PoSTest.rho
```diff
@@ -463,6 +465,42 @@ match (
                 }
               }
             }
+          } |
+
+          contract getPosBalance(retCh) = {
```
JosephDenman, line 475: This helper contract should be introduced in the next commit where it's used.

casper/src/main/resources/PoS.rhox
```diff
@@ -344,17 +405,29 @@ new  PoS,
...
+        contract deposit (@fromUserPk, @amount, @activeValidators, returnCh) = {
           new vaultCh,
               revAddressCh,
+              indexSender,
+              getBlockData(`rho:block:data`), blockDataCh,
               authKeyCh in {
             @RevVault!("deployerAuthKey", *authKeyCh) |
             for (@authKey <- authKeyCh) {
               revAddressOps!("fromPublicKey", fromUserPk, *revAddressCh) |
+              getBlockData!(*blockDataCh) |
```
JosephDenman, line 422: Again, this would be more readable if the produce on `getBlockData` was in the same scope as the consume.

casper/src/main/resources/PoS.rhox
JosephDenman: This is still the case after your changes. I can't bear Rholang being harder to read than it already is.

casper/src/test/resources/PoSTest.rho
JosephDenman: @odeac The PoS vault will always appear empty from the outside. It's only used as a temporary source vault for transfers during `closeBlock`, so I don't think it should make it to the API.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/PoSSpec.scala
```diff
@@ -15,7 +15,7 @@ class PoSSpec
     extends RhoSpec(
       CompiledRholangSource("PoSTest.rho", NormalizerEnv.Empty),
       Seq.empty,
-      120.seconds,
+      240.seconds,
```
JosephDenman: What did you determine was the source of this increase?

casper/src/main/resources/PoS.rhox
JosephDenman: I agree, but I don't see  another way to access the vaults in Rholang. One hack would be to interpolate unforgeable names into the Rholang source code (for the necessary methods) and call those names through system processes in the tests. We might have to do that with some of the current public methods in `PoS` that will need to be private in production. If I'm not mistaken, `closeBlock` is an example.

JosephDenman:
@tymm It's getting there. The main issue for me is the `getValidatorVaults` method that is introduced solely for testing. As I mentioned, interpolating an unforgeable name into `PoS.rhox` to represent a `getValidatorVaults` method might be a feasible approach to getting a handle on the vaults for testing. We'll have to solve the issue of protocol private methods (like `closeBlock`) in Rholang system contracts regardless.

Also, I think that `take` should be introduced as a method. It's not a higher-order function and implementing it as a built-in is likely to be far more performant that implementing it in `ListOps`.

casper/src/main/resources/PoS.rhox
tymm: There can be more available validators than active ones, right? If that's the case the answer is "no".

casper/src/main/resources/PoS.rhox
tymm: Sorry, I was sure that I had changed that already.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/PoSSpec.scala
tymm: `ListOps!take`.

casper/src/test/resources/PoSTest.rho
tymm: It's confusing and I removed it.
I guess what I meant is that once `closeBlock` is done, we are good to set the block sender to one of the active validators.

casper/src/test/resources/PoSTest.rho
tymm: @odeac we can make it part of the PoS API once we need it there.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/PoSSpec.scala
JosephDenman: I guess that's another reason to make `take` a built-in.

casper/src/main/resources/PoS.rhox
odeac: but the take is not fair so it's not correct

tymm: Yeah, I put a comment about that.
Once we have on-chain randomness we can shuffle before we take.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
odeac: @JosDenmark 's argument that "take" is not a higher-order function so it should be a method sounds fair to me

casper/src/test/resources/PoSTest.rho
odeac: ok

casper/src/main/resources/PoS.rhox
odeac: We could extract the "vault per validator" logic into a separate vault-like contract that would have an address and a `balance` method, the same as all the "normal" vaults. The PoS contract will not really care about the actual implementation of its vault and the vault-per-validator logic can be tested outside of PoS tests.

9rb:
replaced by 2795. Some of the reviewer conversation may still be relevant.

PR #2731, opened 2019-09-17
Convert CasperLaunch to tagless final

ArturGajowy:
## Overview

I wanted to make BlockApi a TF-algebra, and somehow arrived at this as a needed first step. I think @marcin-rzeznicki will take over from here according to our talks.

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
@@ -2,7 +2,6 @@ package coop.rchain.casper.engine
 
 import java.nio.file.Paths
 
-import cats.Monad
 import cats.effect.{Concurrent, Sync}
 import cats.implicits._
 import coop.rchain.blockstorage.BlockStore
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -20,121 +19,114 @@ import coop.rchain.comm.transport.TransportLayer
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.shared._
 
-object CasperLaunch {
+trait CasperLaunch[F[_]] {
+  def launch(): F[Unit]
+}
+
+class DefaultCasperLaunch[F[_]: LastApprovedBlock: Metrics: Span: BlockStore: ConnectionsCell: TransportLayer: RPConfAsk: SafetyOracle: LastFinalizedBlockCalculator: Sync: Concurrent: Time: Log: EventLog: BlockDagStorage: EngineCell: EnvVars: RaiseIOError: RuntimeManager: Running.RequestedBlocks: EventPublisher: SynchronyConstraintChecker](
```
marcin-rzeznicki: Concurrent < Async < Sync

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -20,121 +19,114 @@ import coop.rchain.comm.transport.TransportLayer
...
           "ApprovedBlock not found in storage, taking part in ceremony as ceremony master"
-        val action = initBootstrap[F](init)
+        val action = initBootstrap()
         (msg, action)
       case None =>
         val msg    = "ApprovedBlock not found in storage, connecting to existing network"
-        val action = connectAndQueryApprovedBlock[F](init)
+        val action = connectAndQueryApprovedBlock()
         (msg, action)
     } >>= {
```
marcin-rzeznicki: `flatTap` might be a bit cleaner.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -701,7 +699,7 @@ object NodeRuntime {
         implicit val ra = rpConfAsk
         implicit val eb = eventPublisher
         implicit val sc = synchronyConstraintChecker
-        CasperLaunch[F](casperInit)
+        new DefaultCasperLaunch[F](conf.casper).launch()
```
marcin-rzeznicki: `F` is still abstract at this point - why committing to a concrete implementation at this point?

ArturGajowy: WDYM? We're at the edge of the world, hence the concrete instance.

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
ArturGajowy: The best I could arrive at was
```
      .flatTap(msgAction => Log[F].info(msgAction._1))
      .map(_._2)
```
I think I like the current better
```
    } >>= {
      case (msg, action) => Log[F].info(msg) >> action
    }
```

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -20,121 +19,114 @@ import coop.rchain.comm.transport.TransportLayer
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.shared._
 
-object CasperLaunch {
+trait CasperLaunch[F[_]] {
```
sebtomba, line 22: This new `trait` isn't used anywhere. It's not really a TF algebra, just some launcher. I don't understand the context of this change.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
marcin-rzeznicki: You're not actually at the edge of the world which is in `main` - you're in `setupNodeProgramF` which is abstract in `F` . So IMO the cleaner version would be dependent on algebra you created eg `def setupNodeProgramF[F: CasperLaunch] = ...` or, explicitly `def setupNodeProgramF[F[_]](launcher: CasperLauch[F]) = ...` if you treat it more as an utitlity. This concern is also mirrored in @sebtomba review (and if you coded it this way, it'd solve both problems)

marcin-rzeznicki: IMO there is another valid approach here - if we actually wanted to treat `NodeRuntime` as the end of the known world, then `DefaultCasperLaunch` could be ditched in favor of `implicit val casperLaunchInstance: CasperLaunch[TaskEnv] = ...` . I am not sure which one is better tbh

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
marcin-rzeznicki: Well, for one thing, it eliminated a few sets of long context bounds. Please take a look at my comments - I think we share the same opinion

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
sebtomba: I agree. We should treat `CasperLaunch` as any other dependency.

ArturGajowy:
bors try

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
ArturGajowy: Good point on the trait not being used.

It is a TF algebra, no matter how trivial and unused :)

The context is: I wanted TF further down in casper, and figured it's easiest to introduce top-down. CasperLaunch was closest to `main` that wasn't TF, so I started from there.

Ignoring the above, the change makes a couple methods private, removes the unneeded CasperInit wrapper, and deduplicates the deps lists as pointed out by @marcin-rzeznicki.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
ArturGajowy: setupNodeProgramF is the producer of casper launch, not the consumer. I've pulled the actual `.launch()` call to the top level (to `main`) to make the launch more explicit. This also makes the trait used.

If you guys mean it should say 

```
CasperLaunch[F]
```

instead of

```
new DefaultCasperLaunch(casperConf)
```

then I beg to differ: we should not use the `apply` method on a TF's companion object for anything else than the implicit summoner (`def apply[F[_](implicit ev: Foo[F]) = ev`), or we risk ad-lib instantiation of TF algebra instances down in the program. This is, it would seem, is in opposition to what we currently have in some places, e.g.:

https://github.com/rchain/rchain/blob/f802f1f4bacd2b7e2492f7cfa6a8eb2e562bcb48/node/src/main/scala/coop/rchain/node/NodeRuntime.scala#L711

(notice the above spawns a new CasperPacketHandler wherever its dependencies are available)

https://github.com/rchain/rchain/blob/a0f0c1bad01c272f5385e442992dc9d1094c3730/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala#L13-L16

marcin-rzeznicki: >  we should not use the apply method on a TF's companion object for anything else than the implicit summoner (def apply[F[_](implicit ev: Foo[F]) = ev), or we risk ad-lib instantiation of TF

I wasn't proposing that - I agree with you here. What lost me here is that in the previous version `setupNodeProgramF` was actually creating **launch** effect (which made a lot of sense given the name, barring the fact that it was hidden under `apply`). In your design what we have is effectively a lawless algebra wich advertises that it can be interpreted under any `F` to produce `launch` side-effect - whatever this means is left unspecified (and I agree, semantics here are hard to capture - just like, say, context shift). So my previous comment about passing it as a bound is, in fact, wrong because there is no `setup` needed. Now, this method is not, in the strict sense of the word, a producer of algebra. It is just its, hmm, by-passer :-) So how about:

```scala
object CasperLaunch {
  def of[F: WhatYouGot: AndMore](conf: CaperConf) = //the real deal
}
// default is a bit misleading as well because there is no "alternative"
// if you wanted to eg. turn it off for tests, you could simply do
// new CasperLaunch[F] { def launch() = Applicative[F].unit }
// but it's not like you would be offering this as an alternative to the default
```

If what I said doesn't make sense then just ignore me :-) overall, the PR is fine

ArturGajowy: If you insist, I can alias the `new DefaultCasperLaunch()` constructor as `CasperLaunc.of`, but that would mean duplicating the bounds list. Used as a convention, that would mean we need to declare (and change when needed) each TF bounds list twice.

Maybe a better suffix/prefix than `Default*` would do? `*Impl` is out of question for me :) FWIW, in general, if there's just one choice I consider it a good `Default` choice :D

> setupNodeProgramF was actually creating launch effect

it was executing the very same launch effect under the covers of innocuous `CasperLaunch[F]` :)

> algebra wich advertises that it can be interpreted under any F

This is true of every TF trait :) Hopefully, at least, or else the TF flavor at hand is not composable (can't stub without providing the deps; deps propagate, need to express them twice (in trait and impl)).

> lawless algebra

We're not yet at the stage to have any lawful algebras, and most TF-'algebras' aren't. Should we be calling them TF-traits? Or are we fine with 'lawless' being implied by 'algebra' used near 'TF'? (I'm fine with both :)) 

> overall, the PR is fine

May I have a :heavy_check_mark: ? :)

marcin-rzeznicki: > If you insist, I can alias the new DefaultCasperLaunch() constructor as CasperLaunc.of, but that would mean duplicating the bounds list

I was thinking about anonymous class - no duplication needed and no name needs to be invented :-) Anyways, if you think `Default` is ok then I believe in your judgement.

> or else the TF flavor at hand is not composable

Here we differ a bit. Constraints on derived operations do not hurt composability. Contravariance does :-)
```scala
trait Lauch[F[_]] { self =>
  def launch(finalizer: F[Unit]): F[LaunchResult]

  final def mapk[G[_]](fk: F ~> G): Alg1[G] = new Alg1[G] {
    override def launch(finalizer: G[Unit]): G[LaunchResult] = sys.error("impossible")
  }
}
```

In this case it's fine to be lawless (I can't see any useful law attached to an algebra with a single operation returning `F[Unit]` :-)), just as it is fine for a `ContextShift` to be lawless. I am not sure about the "most aren't" part, but I am fine with "algebra" terminology :-).

My dubious nitpicking aside - a change for the better!

ArturGajowy:
@marcin-rzeznicki

> I was thinking about anonymous class - no duplication needed and no name needs to be invented :-)

:heart: 

Forgot about that option, fixed :)

sebtomba (approved):
LGTM

marcin-rzeznicki (approved):

PR #2734, opened 2019-09-18
RCHAIN-3677 protobufs only on the outside

EncodePanda:
## Overview
It's here. OMG it's finally here.

Before this change:

Casper protobufs (`CasperMessage`) were used everywhere in the code base. Problem with those are following:

- pattern matching agains generated classes (especially those with `oneof` field) is a true pain
- each field is optional which requires checking if that value exists each time the field is being used. In practice that would be a matter of running everything in `OptionT` monad, in practice it's `value.get` or `value.getOrElse(some_random_default)`

After this change

Casper protobuds are renamed to `CasperMessageProto`. Those messages are being used for serialization purposes only: sending over the wire, storing on the disk. In the casper code a validated entites (called `CasperMessage`) are being used. Those are simple `case classes` whose all fields were previously validated for existance (however validated once).

Before the review please see the `CasperMessage.scala` first - it will give good understanding of that change that was being made.

Good example of it is `CasperPacketHandler`

```
def handlePacket(peer: PeerNode, packet: Packet): F[Unit] =
  toCasperMessage(toCasperMessageProto(packet), peer: PeerNode).fold(...)
```

where

```
def toCasperMessage(
    maybeCMP: Option[CasperMessageProto],
    peer: PeerNode
): Either[String, CasperMessage] = maybeCMP match {
  case None      => Left(s"Could not extract casper message from packet sent by $peer")
  case Some(cmp) => CasperMessage.from(cmp)
}
```

and

```
object CasperMessage {
  def from(cm: CasperMessageProto): Either[String, CasperMessage] = ...
```

`CaspePacketHandler` receives a `Packet` that itself holds a protobuf (`toCasperMessageProto`) and that protobuf is transformed to `CasperMessage`. At each stage an error is handled, thus finally we receive `Either[String, CasperMessage]` - either `CasperMessage` or and error while parsing/deserializing.

A lot of code in casper is simplified because of this, thus the 80+ files change.

Some other notes:

- all funcitonality on casper module should work on `CasperMessage` instances
- other modules (lower in dependency hierarchy) like blockStore or comm, will work internally on protobufs (yet they my expose `CaspeerMessage` instances)
- you will notice a lot of changes in tests like

```
- BlockMessage().withHeader(Header().withVersion(v).withTimestamp(ts))
```

```
+   BlockMessage
+      .withHeader(Header().withVersion(v).withTimestamp(ts))	      .from(
+        BlockMessageProto()
+          .withBlockHash(bh)
+          .withHeader(
+            HeaderProto()
+              .withVersion(v)
+              .withTimestamp(ts)
+          )
+          .withBody(BodyProto().withState(RChainStateProto()))
+      )
+      .right
+      .get
+
```

This is due to the fact that in tests a lot of objects that were created were not strictly 'legal' objects -they were missing fields. With `CasperMessage` that construction is no longer possible, thus the diffs like the one above.
Please note that most of that boiler plate was aggregated in one place `object CasperMessageFactory`.

- please not that `Seq`s were replaced with `List`s

### JIRA ticket
RCHAIN-3677


block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -131,7 +136,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
         result <- filteredIndex.flatTraverse {
                    case (blockHash, indexEntry) =>
                      readBlockMessage(indexEntry)
-                       .map(block => List(blockHash -> block))
+                       .map(block => List(blockHash -> BlockMessage.from(block).right.get)) // TODO FIX-ME
```
EncodePanda, line 139: BlockStore works internally on protobufs so each message it contains will be in proper format (this code will only break if it wsa broken before). Nevertheless this should be address in the future (this PR exposed the problem that wsa not visible before - what if protobufs stored in the past are no longer valid objects in the future)

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockStore.scala
```diff
@@ -25,20 +25,22 @@ class InMemBlockStore[F[_]] private ()(
...
-    } yield state.get(blockHash)
+    } yield state.get(blockHash) >>= (bmp => BlockMessage.from(bmp).toOption)
 
   override def find(p: BlockHash => Boolean): F[Seq[(BlockHash, BlockMessage)]] =
     for {
       _     <- metricsF.incrementCounter("find")
       state <- refF.get
-    } yield state.filterKeys(p(_)).toSeq
+    } yield state.filterKeys(p(_)).toSeq.map {
+      case (bh, bmp) => (bh, BlockMessage.from(bmp).right.get) //TODO FIX-ME
```
EncodePanda, line 35: same comment as with `FileLMDBIndexBlockStore.scala` above

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -194,11 +190,6 @@ object BlockApproverProtocol {
     } yield ()).value
   }
 
-  def packetToUnapprovedBlock(msg: Packet): Option[UnapprovedBlock] =
```
EncodePanda, line 197: This method already exists in `package object protocol {` in casper

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -361,16 +336,23 @@ object ProtoUtil {
   ): BlockMessage = {
     val hash = hashUnsignedBlock(header, justifications)
 
-    BlockMessage()
-      .withBlockHash(hash)
-      .withHeader(header)
-      .withBody(body)
-      .withJustifications(justifications)
-      .withShardId(shardId)
+    // TODO FIX-ME fields that can be empty SHOULD be optional
```
EncodePanda, line 339: This is not part of this PR, but should be a seprate concern (needs a ticket)

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
dzajkowski: please open a ticket for this TODO

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockStore.scala
dzajkowski: please open a ticket for this TODO

casper/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -13,23 +13,20 @@ object PrettyPrinter {
...
 
-  def buildString(t: GeneratedMessage): String =
+  def buildString(t: CasperMessage): String =
     t match {
       case b: BlockMessage => buildString(b)
-      case d: DeployData   => buildString(d)
       case _               => "Unknown consensus protocol message"
     }
 
+  // TODO shouldn header.parentsHashList be nonempty list?
```
dzajkowski, line 22: please open a ticket for this TODO

EncodePanda: 👍

block-storage/src/main/scala/coop/rchain/blockstorage/util/BlockMessageUtil.scala
```diff
@@ -6,22 +6,14 @@ import coop.rchain.casper.protocol.{BlockMessage, Bond, DeployData}
...
 
   def deployData(b: BlockMessage): Seq[DeployData] =
-    (for {
-      bs <- b.body
-    } yield bs.deploys.flatMap(_.deploy)).getOrElse(List.empty)
+    b.body.deploys.map(_.deploy)
 
   def parentHashes(b: BlockMessage): Seq[ByteString] =
-    b.header.fold(Seq.empty[ByteString])(_.parentsHashList)
+    b.header.parentsHashList
```
ArturGajowy, line 18: Changes in this file yield a good promise of redeeming the overly huge diffstat :joy: :ok_hand:

casper/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
itegulov: Isn't it empty for genesis? Can we make genesis an exception to this rule somehow?

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -66,14 +71,14 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
   private[this] def modifyCurrentIndex(f: Int => Int): F[Unit] =
     state.modify(s => s.copy(currentIndex = f(s.currentIndex)))
 
-  private def readBlockMessage(indexEntry: IndexEntry): F[BlockMessage] = {
-    def readBlockMessageFromFile(storageFile: RandomAccessIO[F]): F[BlockMessage] =
+  private def readBlockMessage(indexEntry: IndexEntry): F[BlockMessageProto] = {
```
ArturGajowy, line 74: We should make the block-storage API use the new, non-proto case classes too.

Maybe doing it in this PR would even decrease the total changed lines count, b/c all the Foo -> FooProto renames vanish?
(just pointing this out, not asking for that :))

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -112,10 +117,10 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
...
     lock.withPermit(
       for {
-        indexEntryBytesOpt <- index.get(blockHash.toDirectByteBuffer)
-        indexEntryOpt      = indexEntryBytesOpt.map(IndexEntry.load)
-        result             <- indexEntryOpt.traverse(readBlockMessage)
-      } yield result
+        indexEntryBytesOpt     <- index.get(blockHash.toDirectByteBuffer)
+        indexEntryOpt          = indexEntryBytesOpt.map(IndexEntry.load)
+        maybeBlockMessageProto <- indexEntryOpt.traverse(readBlockMessage)
+      } yield maybeBlockMessageProto >>= (bmp => BlockMessage.from(bmp).toOption)
```
ArturGajowy, line 123: a `BlockMessage.from == Left(_)` should be raised as a `MalformedStorageError`, or similar.
I see you've pointed that out later in the file :+1:

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
```diff
@@ -38,8 +38,19 @@ trait BlockStoreTest
...
+          .withBlockHash(bh)
+          .withHeader(
+            HeaderProto()
+              .withVersion(v)
+              .withTimestamp(ts)
+          )
+          .withBody(BodyProto().withState(RChainStateProto()))
+      )
+      .right
+      .get
```
ArturGajowy, line 53: It'd be useful to have `Dummies.blockMessage`, which would allow:

```scala
val dummy = Dummmies.blockMessage
val bmp = dummy.copy(
  header = dummy.header.copy(
    version = v,
    timestamp = ts
  )
)
```
or sth like that.

Also, using lenses we should be able to get rid of the `dummy` val above. (I know we don't have them for new ADT yet)

I think I'm going to insist on at least a JIRA for having a Dummies helper object, returning "default" values for each of the new ADTs used in tests. We should not be using protos in tests where it's possible to not do so.

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
```diff
@@ -82,17 +93,21 @@ trait BlockStoreTest
...
                 case (k, v1, _) => store.get(k).map(_ shouldBe Some(v1))
               }
+          /*
           _ <- items.traverse_[Task, Unit] { case (k, _, v2) => store.put(k, v2) }
           _ <- items.traverse_[Task, Assertion] {
                 case (k, _, v2) => store.get(k).map(_ shouldBe Some(v2))
               }
           result <- store.find(_ => true).map(_.size shouldEqual items.size)
         } yield result
+         */
```
ArturGajowy: why is that commented out?

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
```diff
@@ -196,10 +211,24 @@ class FileLMDBIndexBlockStoreTest extends BlockStoreTest {
...
+                      .withHeader(HeaderProto())
+                  ),
+                  1
+                )
+              ),
+              List(Signature(ByteString.EMPTY, "", ByteString.EMPTY))
+            )
+          )
+          .right
+          .get
```
ArturGajowy, line 229: this would be `Dummies.approvedBlock` + a line or two of overrides

casper/src/main/scala/coop/rchain/casper/engine/ApproveBlockProtocol.scala
```diff
@@ -129,29 +129,24 @@ object ApproveBlockProtocol {
...
-    private val sigData                   = Blake2b256.hash(candidate.toByteArray)
+    private val sigData                   = Blake2b256.hash(candidate.toProto.toByteArray)
 
     def addApproval(a: BlockApproval): F[Unit] = {
-      val validSig = for {
-        c   <- a.candidate if c == this.candidate
-        sig <- a.sig if Validate.signature(sigData, sig)
-      } yield sig
+      val validSig =
+        (a.candidate == this.candidate) && Validate.signature(sigData, a.sig)
```
ArturGajowy, line 141: [insert THANK YOU gif] :)

EncodePanda:
bors try

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
```diff
@@ -14,11 +14,13 @@ object StandardDeploys {
...
-      deployer = stringToByteString(user),
-      timestamp = timestamp,
-      term = compiledSource.code,
-      phloLimit = accounting.MAX_VALUE
+    val deployData = DeployData.from(
+      DeployDataProto(
+        deployer = stringToByteString(user),
+        timestamp = timestamp,
+        term = compiledSource.code,
+        phloLimit = accounting.MAX_VALUE
```
ArturGajowy, line 22: We should get rid of default values usage in production code. Please open a JIRA for this case - we shouldn't be using dummy values for system deploys.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -74,22 +74,23 @@ object Genesis {
     import genesis._
 
     val state = RChainState(
-      bonds = bondsProto(proofOfStake),
       preStateHash = startHash,
-      postStateHash = stateHash
+      postStateHash = stateHash,
+      blockNumber = 0,
```
ArturGajowy, line 79: This is fine, b/c this is genesis block. Having a `GENESIS_BLOCK_NUMBER` constant would prevent all doubt.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -94,17 +94,19 @@ object DeployRuntime {
             timestamp <- Sync[F].delay(System.currentTimeMillis())
 
             //TODO: allow user to specify their public key
```
ArturGajowy, line 96: Hopefully this is outdated? Can you confirm?

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -94,17 +94,19 @@ object DeployRuntime {
...
+              term = code,
+              timestamp = timestamp,
+              sig = ByteString.EMPTY,
+              sigAlgorithm = "",
+              phloPrice = phloPrice,
+              phloLimit = phloLimit,
+              validAfterBlockNumber = validAfterBlock
+            )
 
             signedData = maybePrivateKey.fold(d)(SignDeployment.sign(_, d))
```
ArturGajowy, line 108: shouldn't we require a private key being used for each deploy by now? Do we still allow unsigned deploys? Please raise an urgent JIRA for that.

casper/src/main/scala/coop/rchain/casper/util/rholang/InternalProcessedDeploy.scala
```diff
@@ -15,23 +15,23 @@ final case class InternalProcessedDeploy(
...
-      s = if (pd.errored) UnknownFailure else Succeeded
-    } yield InternalProcessedDeploy(d, c, l, p, s)
+  def fromProcessedDeploy(pd: ProcessedDeploy): InternalProcessedDeploy =
+    InternalProcessedDeploy(
+      pd.deploy,
+      pd.cost,
+      pd.deployLog.map(EventConverter.toRspaceEvent),
+      pd.paymentLog.map(EventConverter.toRspaceEvent),
+      if (pd.errored) UnknownFailure else Succeeded
+    )
```
ArturGajowy, line 36: yummy!

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -451,7 +421,7 @@ object ProtoUtil {
     * only those fields. This allows users to more readily pre-generate names for signing.
     */
   def stripDeployData(d: DeployData): DeployData =
-    DeployData().withDeployer(d.deployer).withTimestamp(d.timestamp)
+    DeployData.from(DeployDataProto().withDeployer(d.deployer).withTimestamp(d.timestamp))
```
ArturGajowy, line 424: I'm really grateful that we've undertaken this effort, as it's already revealed at least one (if not more) critical security flaws. E.g. here we need to include the deployed code in the signature. Nothing needs to be done, assuming we'll have a JIRA as requested in https://discordapp.com/channels/391435113415573504/420661762828926978/623825225250766888.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -211,7 +212,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: ConnectionsCell: TransportLa
       _.asLeft[DeployId].pure[F],
       kp(
         InterpreterUtil
-          .mkTerm(d.term, NormalizerEnv(d))
+          .mkTerm(d.term, NormalizerEnv(DeployData.toProto(d)))
```
ArturGajowy, line 215: We should be using the new ADT in InterpreterUtil / NormalizerEnv, please JIRA. Anybody on rholang team can handle that.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -125,19 +122,13 @@ object Validate {
       } yield result
     }
 
+  /** TODO move to CasperPacketHandler, we should never work with empty fields
```
ArturGajowy: To qoute a classic:

> JIRA this shit

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
```diff
@@ -0,0 +1,438 @@
...
+      .withBlockHash(bm.blockHash)
+      .withHeader(Header.toProto(bm.header))
+      .withBody(Body.toProto(bm.body))
+      .withJustifications(bm.justifications.map(Justification.toProto))
+      .withSender(bm.sender)
+      .withSeqNum(bm.seqNum)
+      .withSig(bm.sig)
+      .withSigAlgorithm(bm.sigAlgorithm)
+      .withShardId(bm.shardId)
+      .withExtraBytes(bm.extraBytes)
```
ArturGajowy, line 189: sadly, there's no way to use the type system that no fields are omitted here.

Please add a generative roundTripTest similar to https://github.com/rchain/rchain/blob/d8f1e7af888a5245d077b5efae5a5f0505537d19/models/src/test/scala/coop/rchain/models/RhoTypesTest.scala#L19-L27

Please disregard the stacksafety tests that are piggy-backed in there :sweat_smile:

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -125,19 +122,13 @@ object Validate {
...
+    */
```
dzajkowski: please open a ticket for this TODO

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
ArturGajowy: It could be part of CasperMessageFactory too

dzajkowski (approved):
took a long stare at the code. looks mechanically correct. in tests we trust.

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
ArturGajowy: A JIRA for having block-storage API use the new ADTs will be needed at least.

ArturGajowy (changes requested):
A couple comments so far, mostly requests for JIRA issues and praise :heart: 

That's been *WAAAAY* better than anticipated, given the diffstat and the scope. Huge props! I nodded my head all the way throughout the PR :)

The biggest code change request is adding a generative test preventing ommissions in to/fromProto conversions.

Great job, hope to have this merged by tomorrow evening!

ArturGajowy (changes requested):
A couple comments so far, mostly requests for JIRA issues and praise :heart: 

That's been *WAAAAY* better than anticipated, given the diffstat and the scope. Huge props! I nodded my head all the way throughout the PR :)

The biggest code change request is adding a generative test preventing ommissions in to/fromProto conversions.

Great job, hope to have this merged by tomorrow evening!

ArturGajowy:
I've skipped casper tests in the first pass, will proceed to review them now :)

casper/src/test/scala/coop/rchain/casper/engine/GenesisCeremonyMasterSpec.scala
```diff
@@ -73,10 +73,19 @@ class GenesisCeremonyMasterSpec extends WordSpec {
         lastApprovedBlock <- LastApprovedBlock[Task].get
         _                 = assert(lastApprovedBlock.isDefined)
         _                 <- EngineCell[Task].read >>= (_.handle(local, blockApproval))
-        head              = transportLayer.requests.head
+        head              = transportLayer.requests(1)
+        _                 = println("lastApprovedBlock.get.sigs: " + lastApprovedBlock.get.sigs)
+        proto             = ApprovedBlockProto.parseFrom(head.msg.message.packet.get.content.toByteArray)
+        _                 = println("proto.sigs: " + proto.sigs)
```
ArturGajowy: please remove the printlns :)

casper/src/test/scala/coop/rchain/casper/MergeabilityRules.scala
```diff
@@ -494,7 +494,6 @@ trait MergeabilityRules {
             multiParentBlock <- nodes(0).addBlock(deploys(3))
 
             _ = nodes(0).logEff.warns.isEmpty shouldBe true
-            _ = multiParentBlock.header.get.parentsHashList.size shouldBe numberOfParentsForDiamondTip
```
ArturGajowy, line 681: I think we'll need this assertion back :sweat_smile:

ArturGajowy (changes requested):
Two more comments for the casper tests. Please address the comments (do paste the JIRA links for reference!) and it's good to merge :)

ArturGajowy (changes requested):
Two more comments for the casper tests. Please address the comments (do paste the JIRA links for reference!) and it's good to merge :)

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -60,44 +61,44 @@ message Signature {
...
-  Header                 header         = 2;
-  Body                   body           = 3;
-  repeated Justification justifications = 4; // map of all validators to latest blocks based on current view
-  bytes                  sender         = 5; // public key of the validator that created the block
-  int32                  seqNum         = 6; // number of blocks created by the validator
-  bytes                  sig            = 7; // signature generated by signing `hash(hash(justification) concat blockHash)`.
-  string                 sigAlgorithm   = 8; // name of the algorithm used to sign
-  string                 shardId        = 9; // identifier of the shard where the block was created
-  bytes extraBytes = 10;
+message BlockMessageProto {
```
zsluedem, line 90: ~~This might break integration tests.~~

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
EncodePanda: Done https://rchain.atlassian.net/browse/RCHAIN-3844

block-storage/src/main/scala/coop/rchain/blockstorage/InMemBlockStore.scala
EncodePanda: Done https://rchain.atlassian.net/browse/RCHAIN-3845

casper/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
EncodePanda: Ticketed https://rchain.atlassian.net/browse/RCHAIN-3846

casper/src/main/scala/coop/rchain/casper/Validate.scala
EncodePanda: This comment was later removed

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
EncodePanda: DONE https://rchain.atlassian.net/browse/RCHAIN-3844

zsluedem:
This is a good one. It makes much easier for a non-scala developer like me to read the codes. 👍

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
EncodePanda: I've moved `CasperMessageFactory` to `models/test` and renamed it to `Dummies`

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
EncodePanda: Good question :)

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
EncodePanda: 👍

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
EncodePanda: Ticketed https://rchain.atlassian.net/browse/RCHAIN-3847

casper/src/main/scala/coop/rchain/casper/Validate.scala
EncodePanda: This comment was later removed

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
EncodePanda: Can you elaborate on the above?

ArturGajowy: I was going to say:

> If I add a field to BlockMessage, there's nothing that will prevent me from not adding it to BlockMessageProto, and the related to/from methods (not the tests, not the compiler

it is though true that I'll need to provide a value out of thin air in `fromProto` to make stuff compile, at which point I'll remember about the Proto.

Still, nothing prevents me from making a typo/copy-paste error and binding two fields of BlockMessage to the same field of BlockMessageProto (making the newly-added field in the Proto unused).

A test that does:
```scala
forall(bm: BlockMessage) {
  BlockMessage.fromProto(bm.toProto) == Right(bm)
}
```
would prevent that, and similar errors (e.g. hardcoding a value in fromProto).

I'm fine if this is ticketed as a separate PR, as I later realized it requires making shapeless spit out the Arbitrary instances for the new ADT, and that might mandate a migration to scalacheck-magnolia from (presumably slower) scalacheck-shapeless.

casper/src/main/scala/coop/rchain/casper/Validate.scala
ArturGajowy: Can you double-check? It still shows in github: https://github.com/rchain/rchain/pull/2734/files

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
EncodePanda: Test is brought back

EncodePanda:
@ArturGajowy this is now ready for your second review

EncodePanda:
bors try

block-storage/src/test/scala/coop/rchain/blockstorage/BlockStoreTest.scala
ArturGajowy: My bigger point here was to use the Dummies.blockMessage in here and other places in tests that create default-valued ADTs via protos for test setup. Won't push on that though, this can be done in the future :)

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
ArturGajowy: ping :)

casper/src/main/scala/coop/rchain/casper/Validate.scala
ArturGajowy: It's OK now

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
EncodePanda: Oh, ok. Yeah, I was thinking about something like that. Have it on my TODO

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
EncodePanda: https://rchain.atlassian.net/browse/RCHAIN-3854

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
EncodePanda: https://rchain.atlassian.net/browse/RCHAIN-3855

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
EncodePanda: Have fun :) https://rchain.atlassian.net/browse/RCHAIN-3856

ArturGajowy (approved):

PR #2736, opened 2019-09-18
Rchain 3467: add PoS methods commitRandomImage and commitRandom

odeac:
## Overview
The two added methods will be needed further on to produce a random number

Details of the protocol can be found here: https://rchain.atlassian.net/wiki/spaces/CORE/pages/750288944/Proof+of+Stake+contract+specification

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3467

### Notes

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -65,6 +65,8 @@ new  PoS,
               "committedRewards":{},
               "activeValidators":initialActive,
               "withdrawers":{},
+              "randomImages":{},
+              "randomNumbers":{},
               "allBonds":$$initialBonds$$
             }) |
 
@@ -257,7 +259,9 @@ new  PoS,
                             "committedRewards":newRewards,
                             "activeValidators":newValidators,
                             "withdrawers":newWithdrawers,
-                            "allBonds":newBonds
+                            "allBonds":newBonds,
+                            "randomImages":{},
+                            "randomNumbers":{},
                           },
                           Nil)
                       }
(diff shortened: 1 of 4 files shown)


ghost:
This is for validator selection, right?  Please add some commentary about how these methods will be used and what the threat model is.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -1387,6 +1387,27 @@ class DebruijnInterpreter[M[_], F[_]](
       } yield result
   }
 
+  private[this] val hash: Method = new Method() {
```
tymm: We already have a blake2b256 system process: https://github.com/rchain/rchain/blob/d0a3dd737828eca8ec9320b7d590d5421567879f/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala#L240

tymm (changes requested):

odeac:
> This is for validator selection, right? Please add some commentary about how these methods will be used and what the threat model is.

@pyrocto I've updated the PR description with a link to wiki

casper/src/main/resources/PoS.rhox
```diff
@@ -340,6 +344,52 @@ new  PoS,
...
               }
+            } |
+
+            contract PoS(@"commitRandomImage", @hash, ackCh) = {
+              new userCh, mvarCh in {
+                getUser!(*userCh) |
+
+                runMVar!(*stateCh, *mvarCh, *ackCh) |
+                for (@validator <- userCh;
+                     @state, resultCh <- mvarCh) {
```
ghost, line 355: Don't use a join here: they always cause merge conflicts and it's not needed here.

casper/src/main/resources/PoS.rhox
```diff
@@ -340,6 +344,52 @@ new  PoS,
...
+               }
+             }
+            } |
+
+            contract PoS(@"revealRandom", @random, ackCh) = {
+              new userCh, mvarCh in {
+                getUser!(*userCh) |
+
+                runMVar!(*stateCh, *mvarCh, *ackCh) |
+                for (@validator <- userCh;
```
ghost, line 375: Split up join.

casper/src/main/resources/PoS.rhox
```diff
@@ -340,6 +344,52 @@ new  PoS,
...
+                     @state, resultCh <- mvarCh) {
+                  if (state.get("randomImages").contains(validator)) {
+                    if (random.hash() == state.get("randomImages").get(validator))
```
ghost: Use `@"keccak256hash"!(random, ret)` (assuming `random` is a byte array).

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
ghost: Prefer `@"keccak256Hash"` or similar.

casper/src/main/resources/PoS.rhox
odeac: The conflicts will come from reading the state from the stateCh. Does the system process getUser affect the mergeability?

ghost: Sorry for the delayed response; I was traveling yesterday.  Reading from the state will cause contention with other blocks trying to read from the state, but a join will cause contention with *every* other block.  

I don't know how system processes affect mergeability.  Is the definition of getUser injected somehow?  I only see sends to the name.

tymm: According to @lukasz-golebiewski this join here should be fine. See: https://github.com/rchain/rchain/pull/2724#discussion_r326368279

ghost: If joins only cause conflicts on the names being used, then this is fine.  I thought that joins conflict with every block because you can set up weird chains where only certain of them are allowed to happen at the same time.

ghost (approved):

tymm (approved):

PR #2740, opened 2019-09-20
Respond with full block hash in block propose

sebtomba:
## Overview
Respond with full block hash in block propose

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3788

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -16,17 +16,17 @@ import coop.rchain.casper._
 import coop.rchain.casper.DeployError._
 import coop.rchain.casper.MultiParentCasper.ignoreDoppelgangerCheck
 import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.{DagOperations, EventConverter, ProtoUtil}
+import coop.rchain.casper.util._
 import coop.rchain.casper.util.rholang.RuntimeManager
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Blake2b512Random
 import coop.rchain.graphz._
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.metrics.implicits._
-import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.{BlockMetadata, Par}
 import coop.rchain.models.rholang.sorter.Sortable._
 import coop.rchain.models.serialization.implicits.mkProtobufInstance
+import coop.rchain.models.BlockHash.{BlockHash, _}
 import coop.rchain.rholang.interpreter.storage.StoragePrinter
 import coop.rchain.rspace.StableHashProvider
 import coop.rchain.rspace.trace._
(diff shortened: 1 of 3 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -87,9 +87,11 @@ object BlockAPI {
                 result <- maybeBlock match {
                            case err: NoBlock =>
                              Metrics[F]
-                               .incrementCounter("propose-failed") >> s"Error while creating block: $err"
-                               .asLeft[DeployServiceResponse]
-                               .pure[F]
+                               .incrementCounter("propose-failed") >>
+                               Log[F].error(s"Error while creating block: $err") >>
```
itegulov: I don't think this should be logged at the error level. I usually interpret error-level log entries as "something went terribly wrong". The errors we are talking about here are not fatal (no new deploys, synchrony constraint, read only mode etc) so I think warning suits better here.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -101,8 +103,9 @@ object BlockAPI {
                          }
               } yield result).timer("propose-total-time").attempt.flatMap {
                 case Left(e) =>
-                  Metrics[F].incrementCounter("propose-failed") >> Sync[F]
-                    .raiseError[ApiErr[DeployServiceResponse]](e)
+                  Metrics[F].incrementCounter("propose-failed") >>
+                    Log[F].error(s"Error while creating block: ${e.getMessage}") >>
```
itegulov: Are you sure we need this line? Won't it be logged twice because of `ProposeGrpcService.defer` (it calls `attemptAndLog` internally)?

itegulov (changes requested):

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
sebtomba: yes, I know. But `attemptAndLog` actually doesn't log anything
```scala
case Left(ex)      => Task.delay(ex.printStackTrace()) >> Task.raiseError[A](ex)
```
We can remove this line after we change `attemptAndLog`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
sebtomba: You are maybe right. I'll change it.

sebtomba: Now we won't raise here any exceptions anymore.

itegulov (approved):

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -555,7 +561,7 @@ object BlockAPI {
   ): Effect[F, DeployServiceResponse] =
     status
       .map { _ =>
-        val hash    = PrettyPrinter.buildString(block.blockHash)
+        val hash    = Base16.encode(block.blockHash.toByteArray)
```
dzajkowski: you can use the existing helper: `BlockHash.base16String`

dzajkowski (approved):

PR #2742, opened 2019-09-23
Pre rchain 3819

marcin-rzeznicki:
## Overview
Groundwork for RCHAIN-3819. I decided to split refactor stuff into seperate commit for easier reviewing

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3819

### Notes

-  make Packet transport layer much more type-safe
- eliminates needles `orElse` chaining when converting packets to Casper messages, also makes it more structured
- no more writing `F` everywhere (only for code I touched) due to more selective implicits
- many functional idioms are used in place of ad-hoc coding eg. `whenA`, `unit` instead of `().pure`
- tried to eliinate for-comprehension in favor of more expressive constructs where it makes sense


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -451,42 +450,29 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: ConnectionsCell: TransportLa
...
                                     )
+                                  )
                               status <- attemptAdd(b, updatedDag)
                             } yield ((b, status) :: attempts, status._2)
                         }
       (attempts, updatedDag) = attemptsWithDag
-      _ <- if (attempts.isEmpty) {
-            ().pure[F]
-          } else {
+      _ <- F.unlessA(attempts.isEmpty) {
```
KentShikama: Nitpick: I find unlessA to be harder to read

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -178,59 +162,51 @@ object Running {
     casperContains(b.blockHash)
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
-        for {
-          _      <- Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.")
-          status <- casperAdd(b)
-          _ <- if (BlockStatus.isInDag(status.merge)) RequestedBlocks[F].modify(_ - b.blockHash)
-              else noop[F]
-        } yield ()
+        Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") *> casperAdd(b) >>= (
```
KentShikama: Didn't we decide to stick with >> ?

KentShikama (approved):

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
marcin-rzeznicki: Thanks, well, IMO it's definitely better than the  `().pure[F]` ceremony (even `F.unit` would've been better) by hand. But I'm open to suggestions. What would you do instead?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
marcin-rzeznicki: Dunno - but sure, will change, thanks

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
KentShikama: F.unit is fine. `if` statements naturally parse for me; `unlessA` takes a bit more effort.

marcin-rzeznicki: I think I like the fact that there is no need to specify what happens if the condition isn't met, plus you can pass it anything (it _voidifies_ any F[A]). Would you mind if I kept it?

sebtomba: @marcin-rzeznicki for better readability, maybe we can use here `F.whenA(attempts.nonEmpty)`?

KentShikama: Ok

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
sebtomba: yes, we did. `*>` is banned.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
marcin-rzeznicki: :+1:

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
marcin-rzeznicki: Well, I can imagine why it's frowned upon (no evaluation order) but downright _banned_ seems harsh. I do not care actually what is the evaluation order of the log statement wrt to "action"

marcin-rzeznicki: Replaced in aa9d7aa394d9ee9e6a01bd3e62b89fb417e5cc86

marcin-rzeznicki:
Ping. Reabsed on top of @rabbitonweb changes. Please resume reviewing. His changes enabled me to tackle the problem of "string-based" programming when serializing/deserializing packets - so I included the change during rebase (if you're interested, please see the aa9d7aa394d9ee9e6a01bd3e62b89fb417e5cc86)

casper/src/main/scala/coop/rchain/casper/protocol/package.scala
```diff
@@ -1,63 +1,29 @@
...
-      transport.ApprovedBlockRequest,
-      ApprovedBlockRequestProto.parseFrom
-    )
-  def packetToBlockRequest(msg: Packet): Option[BlockRequestProto] =
-    convert[BlockRequestProto](msg, transport.BlockRequest, BlockRequestProto.parseFrom)
-  def packetToHasBlockRequest(msg: Packet): Option[HasBlockRequestProto] =
-    convert[HasBlockRequestProto](msg, transport.HasBlockRequest, HasBlockRequestProto.parseFrom)
-  def packetToHasBlock(msg: Packet): Option[HasBlockProto] =
-    convert[HasBlockProto](msg, transport.HasBlock, HasBlockProto.parseFrom)
+  @inline def convert[Tag <: PacketTypeTag](
```
lukasz-golebiewski, line 24: Is the explicit `@inline` needed here? Isn't this going to be inlined automatically by the compiler?

models/src/main/scala/coop/rchain/casper/protocol/PacketTypeTag.scala
```diff
@@ -0,0 +1,99 @@
...
+import java.io.IOException
+
+import com.google.protobuf.ByteString
+import coop.rchain.comm.protocol.routing.Packet
+import enumeratum._
+import scalapb.{GeneratedMessage, GeneratedMessageCompanion, Message}
+
+import scala.util.{Failure, Try}
+
+sealed abstract class PacketTypeTag extends EnumEntry
```
lukasz-golebiewski, line 12: Why not `sealed trait`?

models/src/main/scala/coop/rchain/casper/protocol/PacketTypeTag.scala
```diff
@@ -0,0 +1,99 @@
...
+  case object ForkChoiceTipRequest     extends PacketTypeTag
+  case object ApprovedBlock            extends PacketTypeTag
+  case object ApprovedBlockRequest     extends PacketTypeTag
+  case object BlockApproval            extends PacketTypeTag
+  case object UnapprovedBlock          extends PacketTypeTag
+  case object NoApprovedBlockAvailable extends PacketTypeTag
+
+  override val values = findValues
+
+  sealed abstract class ValueOf[Of <: PacketTypeTag] {
```
lukasz-golebiewski, line 28: This is also a good candidate for `sealed trait`

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -64,6 +62,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: ConnectionsCell: TransportLa
   import MultiParentCasper.MetricsSource
 
   implicit private val logSource: LogSource = LogSource(this.getClass)
+  private[this] val F                       = Sync[F]
```
lukasz-golebiewski: `syncF` for consistency?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
lukasz-golebiewski: I was also about to ask this in some other line, but I noticed this "resolved conversation". Better do a fixup to avoid confusion

casper/src/main/scala/coop/rchain/casper/protocol/package.scala
marcin-rzeznicki: Long story short: no. We should configure inliner and optimizer as described here: https://www.lightbend.com/blog/scala-inliner-optimizer - only then it will take effect. I'm not sure if we do that in prod, we definitely should (I benchmarked this stuff and monomorphisation is a huge win, especially with tagless-final and/or transformers)

models/src/main/scala/coop/rchain/casper/protocol/PacketTypeTag.scala
marcin-rzeznicki: I developed a habit of using `trait` only when specifically needed. Abstract classes used to compile to a better bytecode (no copies of methods or forwarders), you have no problem with passing parameters, using non-final `val`s etc. But I agree that in this case it might not really matter. For more details please see: https://www.scala-lang.org/blog/2016/07/08/trait-method-performance.html

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
marcin-rzeznicki: Yeah - as you wish :smile:

marcin-rzeznicki: Fixed in 142b3c680bdb5116c371380a004ad6004c81d6eb

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
marcin-rzeznicki: Fixed in 142b3c680bdb5116c371380a004ad6004c81d6eb

lukasz-golebiewski (approved):
LGTM

marcin-rzeznicki:
bors try

PR #2744, opened 2019-09-24
[WIP] Use deployer ID in PoS.pay

JosephDenman:
## Overview
This PR consists primarily of refactorings to the proof-of-stake tests. Notice that, since `DeployerId` is determined statically, we can no longer use system processes like `setDeployParameters` to modify the deployer identity dynamically. That means that we have to evaluate the sections of the tests that assume different identities independently. That's why `PoSTest` has to be split into pieces. 

I spent a significant chunk of time exploring ways to rewrite the testing infrastructure to be less verbose + more powerful before I came to the conclusion that Rholang simply isn't powerful enough to succinctly describe one. The original interface by @odeac was particularly resourceful.

### JIRA ticket:
<https://rchain.atlassian.net/browse/RCHAIN-3709>


casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -119,38 +126,35 @@ class RhoSpec(
...
-                .eval(testObject.code, runtime, testObject.normalizerEnv)(
-                  implicitly,
-                  rand.splitShort(1)
-                )
-                .timeout(executionTimeout)
-
-          result <- testResultCollector.getResult
-        } yield result
-      }
+        evaluateDeploy(rhoSpecDeploy, runtime) >> tests.toList
```
odeac, line 129: minor: the code indentation is confusing here

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -119,38 +126,35 @@ class RhoSpec(
...
+    runtime.deployParametersRef.set(ProtoUtil.getRholangDeployParams(deploy)) >> Interpreter[Task]
+      .injAttempt(
+        runtime.reducer,
+        runtime.errorLog,
+        deploy.term,
+        Cost.UNSAFE_MAX,
+        NormalizerEnv(deploy.toProto)
+      ) >>= {
+      case EvaluateResult(_, errors) =>
+        if (errors.nonEmpty) errors.head.raiseError[Task, Unit]
```
odeac, line 155: only the first error is raised and the others are hidden. You can concatenate the error messages and raise an error containing all the failure information

casper/src/test/resources/PoSTest/test_make_pos_succeeds.rho
```diff
@@ -0,0 +1,25 @@
...
+      for(@bonds <- bondsCh; @rewards <- rewardsCh) {
+        new rewardsZeroPred, rewardsZeroCh in {
+          @ListOps!("forall", rewards.toList(), *rewardsZeroPred, *rewardsZeroCh) |
+          contract rewardsZeroPred(@(_, reward), resultCh) = {
+            resultCh!(reward == 0)
+          } |
+          for(@rewardsZero <- rewardsZeroCh){
+            new assert(`rho:test:assertAck`), testSuiteCompleted(`rho:test:testSuiteCompleted`), ack in {
+              testSuiteCompleted!(true) |
+              assert!("have a bonds map", 1, bonds.size() >= 0, "", *ack) |
```
odeac, line 17: We need a tool like assertMany that would sequentially run the assertions. Otherwise two consecutive runs of the tests might fail differently and that would confuse the developer

casper/src/test/resources/PoSTest.rho
```diff
@@ -71,20 +69,6 @@ match (
             }
           } |
 
-          contract test_close_block_finishes(rhoSpec, _, ackCh) = {
```
odeac, line 99: I don't agree with this removal. This is a smoke test for closeBlock. If it fails you know where to look for a class of problems. Without this test you have a much bigger area to search for problems when the tests hang

casper/src/test/resources/PoSTest/test_bonding_succeeds.rho
```diff
@@ -0,0 +1,70 @@
...
+                                  @rewards <- rewardsCh;
+                                  @deployerPk <- deployerPkCh){
+                                match bonds.get(deployerPk) {
+                                  bondAmount => {
+                                    new assert(`rho:test:assertAck`), testSuiteCompleted(`rho:test:testSuiteCompleted`), ack in {
+                                      testSuiteCompleted!(true) |
+                                      assert!("add the new bond to the bonds map", 1, bondAmount == 100, "", *ack) |
+                                      assert!("decrease deployer vault balance by 100", 1, initialDeployerBalance - finalDeployerBalance == 100, "", *ack) |
+                                      assert!("increase PoS vault balance by 100", 1, finalPoSBalance - initialPoSBalance == 100, "", *ack) |
+                                      assert!("record no rewards for new validator", 1, rewards.get(deployerPk) == Nil, "", *ack)
```
odeac, line 51: Since there are multiple messages sent to the ack channel. The result collector doesn't know when the assertions have finished. 

The original idea was that testSuiteCompleted !(true) happens at the end of the assertions in order to notify the test result collector that this test is finished

odeac (changes requested):
I like the direction of the PR. My main concern is about the order of the assertions. I'm pretty sure we should make them sequential.

casper/src/test/resources/PoSTest/test_bonding_succeeds.rho
JosephDenman: Yeah. Using `testSuiteCompleted` here is definitely a hack. I'm weighing that against the syntactic overhead of using `RhoSpec` to run these assertions. I don't see a middle-ground between removing the "has finished" assertion in `RhoSpec` and using `testSuiteCompleted` here, but I'm certainly open to ideas.

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
JosephDenman: My thinking is, if we're running a test that consists of multiple programs, the test should terminate the first time an error is thrown. If a program fails, the programs after it are likely to fail, or at least behave unexpectedly.

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
```diff
@@ -67,43 +65,43 @@ class RhoSpec(
 
   def hasFailures(assertions: List[RhoTestAssertion]) = assertions.find(_.isSuccess).isDefined
 
-  private def testFrameworkContracts[F[_]: Log: Concurrent: Span](
-      testResultCollector: TestResultCollector[F]
-  ): Seq[SystemProcess.Definition[F]] = {
+  private def testFrameworkContracts(
```
tymm, line 72: I don't understand these changes to `Task`. Why would you want to do that?

JosephDenman: The invocations already used Task. The call to runSync at the bottom of RhoSpec requires a concrete effect type.

tymm: Ah, thanks. Now it makes more sense.
I think we could use `SyncIO` though. Then we don't have to commit to `Task`.

tymm (approved):
Looks good to me besides what @odeac already mentioned.

However you could have just used `rho:test:deployerId:get` to construct deployerIds in tests.

casper/src/test/resources/PoSTest/test_bonding_succeeds.rho
odeac: I think we should have an assertMany construct either as a system process or a rholang wrapper on top of the assert system process

casper/src/test/scala/coop/rchain/casper/helper/RhoSpec.scala
odeac: I agree with your reasoning but I'm thinking that the errors field is a List for a reason. 

Or do you think it should be changed (in a future PR)?

PR #2746, opened 2019-09-26
Use DeployerID in PoS

JosephDenman:
## JIRA
<https://rchain.atlassian.net/browse/RCHAIN-3709>

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -346,9 +346,9 @@ new  PoS,
               }
             } |
 
-            contract PoS(@"commitRandomImage", @hash, ackCh) = {
+            contract PoS(@"commitRandomImage", @deployerId, @hash, ackCh) = {
               new userCh, mvarCh in {
-                getUser!(*userCh) |
+                getUser!(deployerId, *userCh) |
 
                 runMVar!(*stateCh, *mvarCh, *ackCh) |
                 for (@validator <- userCh;
@@ -366,9 +366,9 @@ new  PoS,
              }
             } |
 
-            contract PoS(@"revealRandom", @random, ackCh) = {
+            contract PoS(@"revealRandom", @deployerId, @random, ackCh) = {
               new userCh, mvarCh, computeHash(`rho:crypto:keccak256Hash`), hashCh in {
-                getUser!(*userCh) |
+                getUser!(deployerId, *userCh) |
 
                 runMVar!(*stateCh, *mvarCh, *ackCh) |
                 computeHash!(random, *hashCh) |
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -27,13 +27,13 @@ new  PoS,
     revVaultCh, listOpsCh,
     getUser,
     getCurrentUserAddress,
-    getCurrentUserVault,
     pickActiveValidators,
     getMVar,
     runMVar,
     fst,
     deposit, distributeRewards,
     rs(`rho:registry:insertSigned:secp256k1`),
+    posId(`rho:rchain:deployerId`),
     uriOut in {
   registryLookup!(`rho:rchain:revVault`, *revVaultCh) |
   registryLookup!(`rho:lang:listOps`, *listOpsCh) |
@@ -42,11 +42,12 @@ new  PoS,
       @(_, ListOps) <- listOpsCh) {
     new posRevAddressCh, posAuthKeyCh, posVaultCh, initialActiveCh in {
 
-      @RevVault!("deployerAuthKey", *posAuthKeyCh) |
-      getCurrentUserAddress!(*posRevAddressCh) |
+      @RevVault!("deployerAuthKey", *posId, *posAuthKeyCh) |
+      getCurrentUserAddress!(*posId, *posRevAddressCh) |
 
       for(@posRevAddress <- posRevAddressCh;
           @posAuthKey <- posAuthKeyCh) {
+
         @RevVault!("findOrCreate", posRevAddress, *posVaultCh) |
         pickActiveValidators!($$initialBonds$$, {}, *initialActiveCh) |
 
@@ -54,7 +55,7 @@ new  PoS,
              @initialActive <- initialActiveCh) {
           new stateCh in {
             // State structure:
-            // pendingRewards : Map[PublickKey, Int] - are accummulated by calling "pay"
+            // pendingRewards : Map[PublickKey, Int] - are accummulated by calling "chargeDeploy"
             // committedRewards : Map[PublickKey, Int] - are moved from pendingRewards at each closeBlock
             // activeValidators : List[PublicKey] - the active validators
             // withdrawers : Map[PublicKey, Int] - the validators willing to withdraw and the blocknumber when their quarantine ends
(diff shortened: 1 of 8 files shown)


casper/src/main/resources/PoS.rhox
```diff
@@ -90,10 +90,9 @@ new  PoS,
               }
             } |
 
-            contract PoS (@"bond", @amount, returnCh) = {
+            contract PoS (@"bond", deployerId, @amount, returnCh) = {
```
odeac: what exactly is the deployerId? Is there some wiki page that describes the architecture?

tymm: Remember the drain-vault attack (https://gist.github.com/ArturGajowy/79b29ea837b744f16122fa81eac8a165)? The deployerId is a statically determined value that prevents drain-vault attacks.
`DeployerIdTest.scala` might also be helpful.

tymm: Please use `@deployerId` for consistency here and further in the file.

casper/src/test/resources/PoSTest.rho
```diff
@@ -61,13 +61,21 @@ match (
...
+                deployerIdCh,
+                revAddrCh,
+                vaultCh in {
               setDeployData!("userId", pk.hexToBytes(), *identitySet) |
               revAddressOps!("fromPublicKey", pk.hexToBytes(), *revAddrCh) |
-              for (_ <- identitySet; @revAddress <- revAddrCh) {
+              makeDeployerId!("deployerId", pk.hexToBytes(), *deployerIdCh) |
+              for (_ <- identitySet;
+                  @revAddress <- revAddrCh;
+                  deployerId <- deployerIdCh) {
```
tymm: Please also write `@deployerId` in receives here and further in the file.

casper/src/test/resources/PoSTest.rho
```diff
@@ -256,19 +264,19 @@ match (
                 initialUser1BalanceCh, finalUser1BalanceCh,
                 initialUser2BalanceCh, finalUser2BalanceCh in {
               prepareUser!("2222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222222", *setupCh) |
-              for (@user1PubKey, _, @user1Vault <- setupCh) {
+              for (@user1PubKey, _, @user1Vault, deployerId0 <- setupCh) {
```
tymm: Nitpick: Alignment of the indices here and further down the file.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -63,12 +63,13 @@ object BlockCreator {
...
                               Base16.encode(invalidBlockHash.toByteArray)
+                            // TODO: Do something useful with the result of "slash".
                             ConstructDeploy.sourceDeployNowF(
                               s"""
-                               #new rl(`rho:registry:lookup`), posCh in {
+                               #new rl(`rho:registry:lookup`), deployerId(`rho:rchain:deployerId`), posCh in {
                                #  rl!(`rho:rchain:pos`, *posCh) |
                                #  for(@(_, PoS) <- posCh) {
-                               #    @PoS!("slash", "$encodedInvalidBlockHash".hexToBytes(), "IGNOREFORNOW")
+                               #    @PoS!("slash", *deployerId, "$encodedInvalidBlockHash".hexToBytes(), "IGNOREFORNOW")
```
tymm, line 72: Hmm, in `PoS!slash` we want to transfer from the validator who gets slashed, right? Therefore there shouldn't be a need for the deployerId of the block creator.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -14,7 +15,16 @@ final case class ProofOfStake(
     quarantineLength: Int = 50000
 ) extends CompiledRholangTemplate(
       "PoS.rhox",
-      NormalizerEnv.Empty,
+      NormalizerEnv(
```
tymm, line 19: This is used outside of tests as well and the public key you use here should not leak into the wild. Otherwise someone with the corresponding private key can access the PoS vault.

tymm (changes requested):
Looks very good beside some minor remarks.

Also the second commit message doesn't make much sense to me.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
tymm: But maybe that's out of scope for this PR.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
JosephDenman: There is a pre-existing ticket to remove generated data (`RegistrySigGen` stuff) from each of the Rholang libraries, but is out of scope for this PR. That includes `PoS` spec. That would effectively introduce `PublicKey` as a parameter here instead of a constant.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
JosephDenman: The validator who gets slashed is determined in the `invalidBlocks.getOrElse(blockHash, userPk)` expression. It's either the validator in `invalidBlocks`, or the validator who called slash if `invalidBlocks` is empty.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
tymm: Okay. Then we should add a todo comment above the `PublicKey` that links to the ticket.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
tymm: The problem is that `slashingDeploys` take the `privateKey` from the block creator (line 77 in dev). From this we derive the public key which will end up as the `deployerId` that we pass to `PoS!slash` here.

JosephDenman: You're saying we should just pass the public key of the block creator?

JosephDenman: I don't see an issue with the situation you described above (where deployerId is passed).

tymm: Sorry, my bad. I wasn't looking closely enough. All good...

casper/src/main/resources/PoS.rhox
JosephDenman: If there's no further clarification needed, I'm going to resolve this conversation.

JosephDenman:
@tymm This is ready for another look.

casper/src/test/resources/PoSTest.rho
tymm: Looks like you forgot about this.

casper/src/main/resources/PoS.rhox
odeac: Ah! OK, we haven't discussed that in a while. It's clear now. Still, we should move that gist to someplace in our wiki so it's easy to be found

JosephDenman:
@tymm @odeac This is ready for another look.

casper/src/test/resources/PoSTest.rho
```diff
@@ -443,12 +451,12 @@ match (
                 finalRewardsCh
             in {
               prepareUser!("9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999", *setupCh) |
-              for (@validatorPubKey, _, _ <- setupCh) {
-                @PoS!("bond", 100, *setupCh) |
+              for (@validatorPubKey, _, _, @deployerId0 <- setupCh) {
```
odeac: I would keep the names of the pubKey in sync with the names of the deployerId.

I'm thinking validatorDeployerId, user1Deployer1 and so on

odeac (approved):

KentShikama (approved):

tymm (approved):

PR #2750, opened 2019-10-01
Transport layer optimisations

sebtomba:
## Overview
1. When Casper wants to stream a block to a remote peer and the outgoing message queue is full then comm should drop the block. Currently, Casper waits (forever) until it can enqueue the block.

2. When streaming of a block to a remote peer fails then don't retry. Currently, we have 3 retries with a 1-second delay.

3. Make the outgoing and the incoming message queues much shorter (f.e. 100). Currently, we allow a node to fall 1000 messages (blocks) behind before we start dropping messages.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3858

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
@@ -2,14 +2,13 @@ package coop.rchain.comm.transport
 
 import java.nio.file._
 
-import scala.concurrent.duration._
-
 import cats.implicits._
 
 import PacketOps._
 import coop.rchain.catscontrib.Catscontrib._
 import coop.rchain.comm.PeerNode
 import coop.rchain.shared.Log
+import coop.rchain.shared.PathOps._
 
 import monix.eval.Task
 import monix.execution.{Cancelable, Scheduler}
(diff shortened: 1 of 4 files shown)


comm/src/main/scala/coop/rchain/comm/transport/StreamObservable.scala
```diff
@@ -40,23 +36,17 @@ class StreamObservable(bufferSize: Int, folder: Path)(implicit log: Log[Task], s
...
-      if (retryCount % LogRetryEvery == 0)
-        log.warn(s"Client stream message queue is full. Retrying push. Retry $retryCount")
-      else Task.unit
-
-    logStreamInformation >> storeBlob >>= (_.fold(Task.unit)(propose(_, 1)))
+    def propose(file: Path): Task[Unit] =
+      push(file) >>= (
+        _.fold(
+          log.debug(s"Enqueued for streaming packet $file"),
+          log.warn(
```
KentShikama, line 43: I think info is fine for this

KentShikama (approved):
When you say casper waits forever to enqueue do you mean it continues to retry every second as of now?

itegulov (approved):

sebtomba:
@KentShikama yes, currently when the outgoing message queue is full then the transport layer retries every second to enqueue the blob message. Casper is the only client of the stream API and gets blocked (asynchronously) by the transport layer in that case. This PR removes the retries. When the outgoing message queue is full then the message will get dropped. However, this situation is an edge case for a node with a slow network. With the recent changes where we don't broadcast full blocks anymore this situation shouldn't occur very often.

PR #2751, opened 2019-10-01
Add "refundDeploy" method to PoS

JosephDenman:
### JIRA ticket:
<https://rchain.atlassian.net/browse/RCHAIN-3710?atlOrigin=eyJpIjoiOTVkMDc1ZThhYjEzNGJjMzgxNWJlM2EyYjlkMTBhN2MiLCJwIjoiaiJ9>


casper/src/main/resources/PoS.rhox
```diff
@@ -155,26 +154,61 @@ new  PoS,
...
+                }
+              } |
+
+              contract PoS (@"refundDeploy", @refundAmount, return) = {
+                for(@(deployerId, initialPaymentAmount) <- currentDeployerData){
+                  new revAddressCh in {
+                    revAddressOps!("fromDeployerId", deployerId, *revAddressCh) |
+                    for(@deployerRevAddress <- revAddressCh){
+                      new transferCh in {
+                        @posVault!("transfer", deployerRevAddress, refundAmount, posAuthKey, *transferCh) |
```
KentShikama, line 184: What prevents a normal user from just calling `PoS!("refundDeploy", 99999999, ...)`?

JosephDenman: There are a number of PoS methods that will have to be made private to the protocol somehow. This interface assumes we'll get around to that. In that case, refund amounts, deployer identity, and ensuring that each `chargeDeploy` is followed by a `refundDeploy` is left to the scala-level protocol.

KentShikama: Please add a comment of that nature then.

JosephDenman:
@KentShikama I've made the change you requested.

KentShikama (approved):

casper/src/main/resources/PoS.rhox
```diff
@@ -155,26 +154,64 @@ new  PoS,
...
+
+              // Note: under no circumstances should this method be merged as is.
+              //       this method, along with "chargeDeploy", "closeBlock", and "slash"
+              //       must be made private, as they're only used in the protocol.
+              contract PoS (@"refundDeploy", @refundAmount, return) = {
+                for(@(deployerId, initialPaymentAmount) <- currentDeployerData){
+                  new revAddressCh in {
+                    revAddressOps!("fromDeployerId", deployerId, *revAddressCh) |
+                    for(@deployerRevAddress <- revAddressCh){
+                      new transferCh in {
```
tymm, line 183: I think we established the rule to group all `new` together at the beginning of a contract.

casper/src/main/resources/PoS.rhox
```diff
@@ -155,26 +154,64 @@ new  PoS,
...
-                    state.get("pendingRewards"),
-                    *newPendingCh) |
-
-                  for(@depositResult <- depositCh;
-                      @newPending <- newPendingCh) {
-                    payResultCh!(state.set("pendingRewards", newPending), depositResult)
+                  for(@depositResult <- depositCh){
+                    match depositResult {
+                      (true, _) => {
+                        currentDeployerData!((deployerId, amount)) |
```
tymm, line 164: Since the `currentDeployerData` channel would be shared by all users, we won't be able to process deploys concurrently.
Are we not able to do that anyway or what are your thoughts here?

casper/src/main/resources/PoS.rhox
```diff
@@ -155,26 +154,64 @@ new  PoS,
...
+                        return!((false, "Deploy payment failed: " ++ errorMessage))
+                      }
+                    }
+                  }
+                }
+              } |
+
+              // Note: under no circumstances should this method be merged as is.
+              //       this method, along with "chargeDeploy", "closeBlock", and "slash"
+              //       must be made private, as they're only used in the protocol.
```
odeac, line 177: is there a JIRA for this?

odeac (approved):

casper/src/main/resources/PoS.rhox
odeac: So you are complaining about a mergeability issue. I think this has a lower priority at this point but we should probably add a Jira item to address this issue

casper/src/main/resources/PoS.rhox
odeac: I don't know if it's a rule but I agree it would look better because it reduces the indentation level.

casper/src/main/resources/PoS.rhox
JosephDenman: I don't believe so. I'll make one.

casper/src/main/resources/PoS.rhox
tymm: Mergeability is about blocks getting merged. My question here is more about concurrent deploy processing in one block.

casper/src/main/resources/PoS.rhox
JosephDenman: It feels wrong to move the name to the top of the file because names should be scoped according to their use. Moving all names to the top of the file is akin to declaring all values at the top-level in Scala or something. It also makes it harder to find out where names are actually used. But, if both of you insist, I'll move the channel to the top of the file.

casper/src/main/resources/PoS.rhox
JosephDenman: > Are we not able to do that anyway or what are your thoughts here?

That's right. There are fundamental problems preventing us from processing deploys in parallel in the current architecture, so I don't really think that's an issue.

casper/src/main/resources/PoS.rhox
tymm: I don't insist. I just wanted to mention that we have been doing it this way so far :)

casper/src/main/resources/PoS.rhox
tymm: Out of curiosity: what are the other issues?

JosephDenman: Say I have `n` deploys and one throws an error. Right now, we have an opportunity to revert state before the next deploy is processed. If all deploys were processed in parallel, we'd have to cancel the entire computation, revert the state, extract the erring deploy, and run the sequence again without the erring process.

tymm (approved):

casper/src/main/resources/PoS.rhox
ghost: https://rchain.atlassian.net/browse/RCHAIN-3888

PR #2759, opened 2019-10-08
Casper refactors

ArturGajowy:
## Overview

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -299,9 +299,17 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
   ): F[(ValidBlockProcessing, BlockDagRepresentation[F])] = {
     val validationStatus: EitherT[F, BlockError, ValidBlock] =
       for {
-        _      <- EitherT(Validate.blockSummary(b, genesis, dag, shardId, expirationThreshold))
-        _      <- EitherT.liftF(Span[F].mark("post-validation-block-summary"))
-        _      <- EitherT(Validate.transactions(b, dag, runtimeManager))
+        _ <- EitherT(Validate.blockSummary(b, genesis, dag, shardId, expirationThreshold))
+        _ <- EitherT.liftF(Span[F].mark("post-validation-block-summary"))
+        _ <- EitherT(
+              InterpreterUtil
+                .validateBlockCheckpoint(b, dag, runtimeManager)
+                .map {
+                  case Left(ex)       => Left(ex)
+                  case Right(Some(_)) => Right(BlockStatus.valid)
+                  case Right(None)    => Left(BlockStatus.invalidTransaction)
+                }
+            )
         _      <- EitherT.liftF(Span[F].mark("transactions-validated"))
         _      <- EitherT(Validate.bondsCache(b, runtimeManager))
         _      <- EitherT.liftF(Span[F].mark("bonds-cache-validated"))
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -387,17 +387,21 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
            * will build off this side of the equivocation, we will get another attempt to add this block
            * through the admissible equivocations.
            */
-          Log[F].info(
-            s"Did not add block ${PrettyPrinter.buildString(block.blockHash)} as that would add an equivocation to the BlockDAG"
-          ) >> dag.pure
+          Log[F]
+            .info(
+              s"Did not add block ${PrettyPrinter.buildString(block.blockHash)} as that would add an equivocation to the BlockDAG"
+            )
+            .as(dag)
 
         case ib: InvalidBlock if InvalidBlock.isSlashable(ib) =>
           handleInvalidBlockEffect(ib, block)
 
         case ib: InvalidBlock =>
-          Log[F].warn(
-            s"Recording invalid block ${PrettyPrinter.buildString(block.blockHash)} for $ib."
-          ) >> dag.pure
+          Log[F]
+            .warn(
+              s"Recording invalid block ${PrettyPrinter.buildString(block.blockHash)} for $ib."
+            )
+            .as(dag)
 
         case BlockError.Processing =>
           syncF.raiseError[BlockDagRepresentation[F]](
@@ -405,8 +409,10 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
           )
 
         case BlockError.BlockException(ex) =>
-          Log[F].error(s"Encountered exception in while processing block ${PrettyPrinter
-            .buildString(block.blockHash)}: ${ex.getMessage}") >> dag.pure
+          Log[F]
+            .error(s"Encountered exception in while processing block ${PrettyPrinter
+              .buildString(block.blockHash)}: ${ex.getMessage}")
+            .as(dag)
 
       }
       .merge
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
@@ -33,48 +33,23 @@ object InterpreterUtil {
   //Returns (None, checkpoints) if the block's tuplespace hash
   //does not match the computed hash based on the deploys
   def validateBlockCheckpoint[F[_]: Sync: Log: BlockStore: Span](
-      b: BlockMessage,
+      block: BlockMessage,
       dag: BlockDagRepresentation[F],
       runtimeManager: RuntimeManager[F]
   ): F[BlockProcessing[Option[StateHash]]] = {
-    val incomingPreStateHash = ProtoUtil.preStateHash(b)
-    val tsHash               = ProtoUtil.tuplespace(b)
-    val deploys              = ProtoUtil.deploys(b)
-    val internalDeploys      = deploys.map(InternalProcessedDeploy.fromProcessedDeploy)
-    val timestamp            = b.header.timestamp
-    val blockNumber          = b.body.state.blockNumber
+    val incomingPreStateHash = ProtoUtil.preStateHash(block)
     for {
       _                  <- Span[F].mark("before-unsafe-get-parents")
-      parents            <- ProtoUtil.getParents(b)
+      parents            <- ProtoUtil.getParents(block)
       _                  <- Span[F].mark("before-compute-parents-post-state")
       preStateHashEither <- computeParentsPostState(parents, dag, runtimeManager).attempt
-      _                  <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(b)}.")
+      _                  <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(block)}.")
       result <- preStateHashEither match {
                  case Left(ex) =>
                    BlockStatus.exception(ex).asLeft[Option[StateHash]].pure
                  case Right(computedPreStateHash) =>
                    if (incomingPreStateHash == computedPreStateHash) {
-                     for {
-                       invalidBlocksSet <- dag.invalidBlocks
-                       unseenBlocksSet  <- ProtoUtil.unseenBlockHashes(dag, b)
-                       seenInvalidBlocksSet = invalidBlocksSet.filterNot(
-                         block => unseenBlocksSet.contains(block.blockHash)
-                       ) // TODO: Write test in which switching this to .filter makes it fail
-                       invalidBlocks = seenInvalidBlocksSet
-                         .map(block => (block.blockHash, block.sender))
-                         .toMap
-                       _         <- Span[F].mark("before-process-pre-state-hash")
-                       isGenesis = b.header.parentsHashList.isEmpty
-                       result <- replayBlockDeploys(
-                                  runtimeManager,
-                                  incomingPreStateHash,
-                                  tsHash,
-                                  internalDeploys,
-                                  BlockData(timestamp, blockNumber),
-                                  invalidBlocks,
-                                  isGenesis
-                                )
-                     } yield result
+                     replayBlockDeploys(block, dag, runtimeManager)
                    } else {
                      //TODO at this point we may just as well terminate the replay, there's no way it will succeed.
                      Log[F].warn(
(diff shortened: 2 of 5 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -37,72 +37,52 @@ object InterpreterUtil {
...
-                 runtimeManager,
-                 preStateHash,
-                 tsHash,
-                 internalDeploys,
-                 possiblePreStateHash,
-                 BlockData(timestamp, blockNumber),
-                 invalidBlocks,
-                 isGenesis = b.header.parentsHashList.isEmpty
-               )
+      isGenesis     = b.header.parentsHashList.isEmpty
```
marcin-rzeznicki: I am not sure if I'm reading it right - but my sense is that we're doing unnecessary work if `preStateHashEither` is `Left` here

marcin-rzeznicki: So looks like the whole thing might be rewritten as (I just rearranged it, perhaps it can be written in a nicer way)

```scala
    for {
      _                  <- Span[F].mark("before-unsafe-get-parents")
      parents            <- ProtoUtil.getParents(b)
      _                  <- Span[F].mark("before-compute-parents-post-state")
      preStateHashEither <- computeParentsPostState(parents, dag, runtimeManager).attempt
      _                  <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(b)}.")
      result <- preStateHashEither match {
                 case Left(ex) =>
                   BlockStatus.exception(ex).asLeft[Option[StateHash]].pure
                 case Right(computedPreStateHash)
                     if (incomingPreStateHash == computedPreStateHash) =>
                   for {
                     invalidBlocksSet <- dag.invalidBlocks
                     unseenBlocksSet  <- ProtoUtil.unseenBlockHashes(dag, b)
                     seenInvalidBlocksSet = invalidBlocksSet.filterNot(
                       block => unseenBlocksSet.contains(block.blockHash)
                     ) // TODO: Write test in which switching this to .filter makes it fail
                     invalidBlocks = seenInvalidBlocksSet
                       .map(block => (block.blockHash, block.sender))
                       .toMap
                     _         <- Span[F].mark("before-process-pre-state-hash")
                     isGenesis = b.header.parentsHashList.isEmpty
                     result <- replayBlockDeploys(
                                runtimeManager,
                                incomingPreStateHash,
                                tsHash,
                                internalDeploys,
                                BlockData(timestamp, blockNumber),
                                invalidBlocks,
                                isGenesis
                              )
                   } yield result
                 case Right(computedPreStateHash) => //TODO at this point we may just as well terminate the replay, there's no way it will succeed.
                   Log[F].warn(
                     s"Computed pre-state hash ${PrettyPrinter.buildString(computedPreStateHash)} does not equal block's pre-state hash ${PrettyPrinter
                       .buildString(incomingPreStateHash)}"
                   ) >> none[StateHash].asRight[BlockError].pure
               }
    } yield result
```

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -188,77 +169,67 @@ object InterpreterUtil {
...
-                           case Left((_, status)) =>
-                             val parentHashes =
-                               parents.map(p => Base16.encode(p.blockHash.toByteArray).take(8))
-                             new Exception(
-                               s"Failed status while computing post state of $parentHashes: $status"
-                             ).asInstanceOf[Throwable]
-                               .asLeft[StateHash]
-                               .asRight
-                         }
+      replayResult <- blocksToApply.foldM(initStateHash) {
```
marcin-rzeznicki, line 226: Genius!

marcin-rzeznicki (commented):
Let me know what you think of my remark - if I'm wrong then I'll approve :-)

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: > we're doing unnecessary work if preStateHashEither is Left here

Indeed we are. Since I find nested for-blocks ugly, I went ahead and extracted a method for the nested one (or rather: pushed that logic into the existing `replayBlockDeploys`). Thanks for the nudge, I think it's been fruitful! :)

Process remark: normally, when some deficiency was already there before the PR, and the PR merely moves it (does not introduce it) - we'd require a fix for the deficiency only if its critical and approve the PR if it's good otherwise. Even when it's critical, we'd sometimes consider a separate follow-up PR for fixing it.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -47,7 +47,7 @@ object InterpreterUtil {
       _                    <- Span[F].mark("before-unsafe-get-parents")
       parents              <- ProtoUtil.getParents[F](b)
       _                    <- Span[F].mark("before-compute-parents-post-state")
-      possiblePreStateHash <- computeParentsPostState[F](parents, dag, runtimeManager)
+      possiblePreStateHash <- computeParentsPostState[F](parents, dag, runtimeManager).attempt
```
JosephDenman: It's not clear to me why this is better. You're eliminating the `Either` return type from `computeParentsPostState`, but in exchange you have to re-throw the errors from `replayIntoMergeBlock`. Maybe it'd be worthwhile if you also eliminated the `Either` return type of `replayIntoMergeBlock`.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -37,12 +37,12 @@ object InterpreterUtil {
       dag: BlockDagRepresentation[F],
       runtimeManager: RuntimeManager[F]
   ): F[BlockProcessing[Option[StateHash]]] = {
-    val preStateHash    = ProtoUtil.preStateHash(b)
-    val tsHash          = ProtoUtil.tuplespace(b)
-    val deploys         = ProtoUtil.deploys(b)
-    val internalDeploys = deploys.map(InternalProcessedDeploy.fromProcessedDeploy)
-    val timestamp       = b.header.timestamp
-    val blockNumber     = b.body.state.blockNumber
+    val incomingPreStateHash = ProtoUtil.preStateHash(b)
```
JosephDenman: Since `replayBlockDeploys` is only called in `validateBlockCheckpoint`, it makes sense to rename the `preStateHash` parameter in `replayBlockDeploys` to `incomingPreStateHash`.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -257,7 +250,14 @@ object InterpreterUtil {
                        invalidBlocks,
                        isGenesis //should always be false
                      )
-    } yield replayResult
+    } yield replayResult.leftMap[Throwable] {
```
JosephDenman, line 259: I think `raiseError` would be preferable to `rethrow` here.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -49,28 +49,32 @@ object InterpreterUtil {
...
+                                  incomingPreStateHash,
+                                  tsHash,
+                                  internalDeploys,
+                                  BlockData(timestamp, blockNumber),
+                                  invalidBlocks,
+                                  isGenesis
+                                )
+                     } yield result
                    } else {
                      //TODO at this point we may just as well terminate the replay, there's no way it will succeed.
```
JosephDenman, line 54: Does this mean that this todo can be removed?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -86,67 +61,91 @@ object InterpreterUtil {
...
-            case UnknownFailure =>
-              Log[F].warn(s"Found unknown failure") >>
-                none[StateHash].asRight[BlockError].pure
-            case UnusedCommEvent(_) =>
-              Sync[F].raiseError(new RuntimeException("found UnusedCommEvent"))
-          }
-        case Left((None, status)) =>
-          status match {
-            case UnusedCommEvent(_: ReplayException) =>
+    result.pure.flatMap {
```
JosephDenman, line 103: This can just be `result match`.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -86,67 +61,91 @@ object InterpreterUtil {
...
+      case Left((Some(deploy), status)) =>
+        status match {
+          case InternalErrors(exs) =>
```
JosephDenman, line 106: To me, it looks like `InternalErrors` is just used as a catch-all for non-Interpreter errors found during the evaluation of a deployment, and that feels wrong.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -86,67 +61,91 @@ object InterpreterUtil {
...
             ) >>
               none[StateHash].asRight[BlockError].pure
-
-          }
-      }
+          case UnknownFailure =>
+            Log[F].warn(s"Found unknown failure") >>
+              none[StateHash].asRight[BlockError].pure
+          case UnusedCommEvent(_) =>
+            Sync[F].raiseError(new RuntimeException("found UnusedCommEvent"))
```
JosephDenman: While you're in the area, it's a good idea to shorten this as well.

JosDenmark: It looks like receiving a `(Some, UnusedCommEvent)` pair is supposed to be indicative of an error a level above the validation protocol. It looks like it's supposed to represent an error. If so, `BugFoundError` should be used here.

JosephDenman (dismissed):
These re-factorings are really nice 👌A couple of questions and comments. Otherwise, LGTM.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -203,7 +203,7 @@ object InterpreterUtil {
       _                  <- Span[F].mark("before-compute-parents-post-state-find-multi-parents")
       blockHashesToApply <- findMultiParentsBlockHashesForReplay(parents, dag)
       _ <- Log[F].info(
-            s"computeMultiParentsPostState computed number of parents: ${blockHashesToApply.length}"
+            s"replayIntoMergeBlock computed number of parents: ${blockHashesToApply.length}"
```
tymm: While you are at it: I think it should say "computed number of uncommon ancestors".
"parents" is misleading as we were able to see while trying to understand the LPT issue.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: I was rather intending to terminate the replay immediately, by raising an error, after figuring out the error handling scheme. I guess @KentShikama should also confirm our understanding. @KentShikama are there any reasons to continue with the replay if the replay of parant states diverged already? (see the TODO in question)

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: not really, there are `Log` effects used down below

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: Thanks for sharing this! I'd investigate and address that in a subsequent PR though.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: I do exactly what you suggest in the latter commits :)

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: That parameter is replaced with the `block` itself in one of the last commits.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: I don't see how, and am having a hard time writing sth similarly concise. I guess it'd be most efficient if you changed that later if you have a better solution.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
KentShikama: I'm confused where this TODO came from. We currently are just returning after logging the warning right?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -299,9 +299,17 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+        _ <- EitherT.liftF(Span[F].mark("post-validation-block-summary"))
+        _ <- EitherT(
+              InterpreterUtil
+                .validateBlockCheckpoint(b, dag, runtimeManager)
+                .map {
+                  case Left(ex)       => Left(ex)
+                  case Right(Some(_)) => Right(BlockStatus.valid)
+                  case Right(None)    => Left(BlockStatus.invalidTransaction)
+                }
+            )
```
KentShikama, line 312: Nitpick: I preferred it extracted

KentShikama (approved):

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: yes. The point of this TODO is: whatever happens after this point is doomed to lead to a faulty replay. In fact, we could proclaim a replay failure right away and slash the block, as clearly the proposing validator confabulates w.r.t. the pre-state hash resulting from merging the parents.

ArturGajowy: That's why I think it'd make sense to raise an error here. The only reasons why I'm not doing that right away are in my first comment.

KentShikama: What is happening right now though?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: Done :)

tymm (approved):

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: Done :)

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
ArturGajowy: After the warning? As mentioned, I think the replay continues but inevitably fails.

KentShikama: Yeah. How does the replay continue? If it is, then I agree it shouldn't be.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
JosephDenman: You can still return F in the body of the case statements :)

JosephDenman: You can still return F in the body of the case statements :)

PR #2763, opened 2019-10-10
RCHAIN-3878: Generalize NormalizerEnv

tymm:
## Overview

### JIRA ticket:

### Notes


JosephDenman (approved):
This is great. There should probably be a test that ensures we can extract data from the channels we pass in through `NormalizerEnv`, but aside from that LGTM.

models/src/main/protobuf/RhoTypes.proto
```diff
@@ -153,9 +153,8 @@ message New {
     Par p = 2 [(scalapb.field).no_box = true];
     // For normalization, uri-referenced variables come at the end, and in lexicographical order.
     repeated string uri = 3;
-    DeployId deployId = 4;
-    DeployerId deployerId = 5;
-    bytes locallyFree = 6 [(scalapb.field).type = "coop.rchain.models.AlwaysEqual[scala.collection.immutable.BitSet]"];
+    map<string, GUnforgeable> injections = 4;
```
ArturGajowy: This should probably be just `map<string, Par>` though, as otherwise we're unable to treat `4` or `"woot"` as inputs. A bit of a bummer that GUnforgable aren't a variant of Expr - we could then just have `Expr` in here.

models/src/main/scala/coop/rchain/models/EqualsM.scala
```diff
@@ -75,6 +75,17 @@ object EqualM extends EqualMDerivation {
 
   }
 
+  implicit def mapEqual[A: EqualM, B: EqualM]: EqualM[Map[A, B]] = new EqualM[Map[A, B]] {
+
+    override def equal[F[_]: Sync](self: Map[A, B], other: Map[A, B]): F[Boolean] = {
+      val pairsA = self.keys.toStream.zip(other.keys)
+      val pairsB = self.values.toStream.zip(other.values)
+      Sync[F].delay(self.size == other.size) &&^ pairsA
+        .forallM(tupled(EqualM[A].equal[F])) &&^ pairsB.forallM(tupled(EqualM[B].equal[F]))
```
ArturGajowy, line 84: :heart_eyes:

models/src/main/protobuf/RhoTypes.proto
tymm: Isn't the idea that we want to inject values of types that can only be created by the system? Under the assumption that `4` is a `GInt`, why would you want to have it as an input?

ArturGajowy: > Isn't the idea that we want to inject values of types that can only be created by the system?

It's the other way around: we need to be able to inject values of types that can only be created by the system, among others - like e.g. `prechargeAmount(`sys:capser:prechargeAmount`)`, which would be a GInt.

tymm: I see.

The disconnect was that I was thinking the idea is to create unforgeable ground types, e.g. `GUnforgeableInt`, and by that protecting us from users calling for example PoS methods.

Since this seems to not be the case: is there already a plan how we will prohibit user from calling the PoS contract?

ArturGajowy: I think we should add a `validatorId: GDeployerId` parameter to each of the methods we only want the bonded validators to call. Based on that, we'd look if the corresponding publicKey is currently bonded, and short-circuit if not.

Notice the `bond` method should be available to user deploys, because it's merely a request to join the bonded validators group. The validator should probably provide its GDeployerId, so that we're sure the request comes from the interested party.

models/src/main/scala/coop/rchain/models/EqualsM.scala
marcin-rzeznicki: I'm probably nitpicking but use of `Sync[F].delay` is unnecesssary as it ties us to `Sync` for no good reason. `Applicative`  should be all that is needed here right?

```scala
((self.size == other.size).pure, pairsA.forallM(_), pairsB.forallM(_)).mapN(_ && _ && _)
```

or, if you want the optimization you're after `if (self.size == other.size) ... else false.pure` (Pelase note that this `&&^` is not even lazy in its arguments)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -399,28 +401,32 @@ class DebruijnInterpreter[M[_], F[_]](
...
-          neu.deployId
-            .map { case DeployId(sig) => newEnv.put(GDeployId(sig)).asRight[InterpreterError] }
-            .getOrElse(normalizerBugFound("DeployId", deployIdUrn).asLeft[Env[Par]])
-        else if (urn == deployerIdUrn)
-          neu.deployerId
-            .map { case DeployerId(pk) => newEnv.put(GDeployerId(pk)).asRight[InterpreterError] }
-            .getOrElse(normalizerBugFound("DeployerId", deployerIdUrn).asLeft[Env[Par]])
+        if (urnMap.get(urn).isEmpty)
+          // If `urn` can't be found in `urnMap`, it must be referencing an injection
+          neu.injections
```
marcin-rzeznicki, line 405: :de: :smile:

models/src/main/scala/coop/rchain/models/EqualsM.scala
tymm: We need `Sync` for `EqualM[A].equal[F]` being stack-safe as far as I know.

marcin-rzeznicki: Hmm, the only place where it might matter is typeclass derivation, right? But I am not sure why would it matter? Anyways, if this is the case then fine

ArturGajowy: > Pelase note that this &&^ is not even lazy in its arguments

I think it's enough if the underlying F is delaying computations.
OTOH it's a good point, we should probably make `%%^` lazy in its second argument.

marcin-rzeznicki: Well - you can't know that :smile: More seriously, you still pay the price for constructing the "tree" But my question is: is `Sync` really needed here?

models/src/main/protobuf/RhoTypes.proto
JosephDenman: My hesitation for guarding PoS methods with the deployerId is that it adds a requirement that each of the PoS methods check that the deployerId is valid, which is extra complexity where we want none. Using system deploy interface to pass in names for the private PoS methods OTOH prevents them from ever being called. But, I guess since computations are metered, we don't have to worry about users spamming the PoS contract. Either is fine.

models/src/main/scala/coop/rchain/models/EqualsM.scala
ArturGajowy: Sync/Defer are a must if the computation can be costly, otherwise we'll incur the cost during computation construction. This is not the case in this particular EqualM instance, but can be an issue in others.

More importantly, we need Sync/Defer to delay the recursion, or else stackoverflow.

Technically, we could just require `Monad` and do 
```scala
def delayIfTrampolined(f: F[A]) = Applicative[F].unit >> f
```
but this would be a hack, and would only yield stacksafety for trampolined monads.

marcin-rzeznicki: Hmm - I am sorry, I still can't see it. Is there any place where you construct the computation for `equals` and don't want to use it right away? In cases like above you could've omitted the construction of some computations (as per the discussion above), yet this seemingly has never been a problem. As for delaying the recursion - 

```scala
import cats.{Applicative, Monad}

trait EqualM[F[_], A] {
  def eq(self: A, other: A): F[Boolean]
}

case class Rec(down: Option[Rec]) {
  override def toString: String = s"Rec(${down.fold("")(_ => "...")})"
}

object Rec {
  def nest(n: Int) = Rec {
    (0 to n).iterator.foldLeft[Option[Rec]](None)((down, _) => Some(Rec(down)))
  }
  implicit def eqRec[F[_]: Monad]: EqualM[F, Rec] = new EqualM[F, Rec] {
    override def eq(self: Rec, other: Rec): F[Boolean] =
      Monad[F].tailRecM((self.down, other.down)) {
        case (Some(r1), Some(r2)) => Monad[F].pure(Left((r1.down, r2.down)))
        case (r1, r2)             => Monad[F].pure(Right(r1 == r2))
      }
  }
}

val x1 = Rec.nest(100000)
val x2 = Rec.nest(100000)

import cats.instances.option._
Rec.eqRec[Option].eq(x1, x2)
```
yields

```
x1: Rec = Rec(...)
x2: Rec = Rec(...)

import cats.instances.`package`.option._
res0: Option[Boolean] = Some(true)
```

ArturGajowy: > In cases like above you could've omitted the construction of some computations (as per the discussion above), yet this seemingly has never been a problem. 

As mentioned, `a.size == b.size` would not be a problem. I think we had a similar *actual* problem with `hugeList.hash` in HashM.

It's a good observation that recursion could be handled in the derivation using `tailrecM`. In fact, `forallM` is used anyway, so maybe it's just a matter of dropping the Sync constraint in the typeclass definition and seeing that the tests pass?

Still, EqualsM would only be stacksafe for stacksafe (trampolining) F-s, and this is mostly because tailrecM would cause a suspension. (BTW the tailrecM suspension is merely a convention not guarded by cats-laws, as exemplified by this [PR of mine to ZIO](https://github.com/zio/zio/pull/795/files)). Sync was used to make the `suspending monad` requirement explicit. 

https://github.com/rchain/rchain/blob/6dfde3f46ed67db35d303cde638784799d337aba/models/src/main/scala/coop/rchain/models/EqualsM.scala#L130-L138

Since delaying computations is needed at times, I'd lean to staying with sync and delaying all the computations rather than risking bugs due to inconsistency. Given the hack I mentioned above though, both approaches (Sync/tailrecM) can be made to work in all our use cases, and we have some tests for EqualsM stacksafety, hopefully. So we could change that.

Whatever you guys decide though, should most likely by implemented in a separate PR.

models/src/main/scala/coop/rchain/models/HashM.scala
```diff
@@ -62,7 +62,9 @@ object HashM extends HashMDerivation {
   implicit val BitSetHash: HashM[BitSet]                     = opaqueHash
   implicit val ByteStringHash: HashM[ByteString]             = opaqueHash
   implicit val Blake2b512RandomHash: HashM[Blake2b512Random] = opaqueHash
-  implicit def alwaysEqualHash[A]: HashM[AlwaysEqual[A]]     = opaqueHash
+  // This should be okay since injections are not recursive
+  implicit val InjectionsMapHash: HashM[Map[String, Par]] = opaqueHash
```
tymm: @ArturGajowy would be nice if you could confirm that this is actually okay.

ArturGajowy: Well... that implicit is going to be used whenever you use a `Map[String, Par]` in any of our datatypes we use HashM for.

It'd be safer to derive a HashM instance:
`implicit def mapHash[A: HashM, B: HashM]: HashM[Map[A, B]]`

tymm: Do you have an idea on how to do that in a stack-safe manner? I'm a little lost here.

tymm: I guess alternatively I could move `NormalizerEnv` to the models module and do `implicit val InjectionsMapHash: HashM[NormalizerEnv] = opaqueHash`.

ArturGajowy: Yeah, that could work. I'd place the implicit on NormalizerEnv companion object

ArturGajowy: Frankly, it'll likely work just as it is know - we're simply trying to reduce surface area for bugs to sneak in

tymm:
bors try

tymm:
bors try

models/src/main/scala/coop/rchain/models/EqualsM.scala
marcin-rzeznicki: > BTW the tailrecM suspension is merely a convention not guarded by cats-laws

Not necessarily:
https://github.com/typelevel/cats/blob/eb52c0bf90f81d60cd4120f3d10c62dd7dbbbd66/laws/src/main/scala/cats/laws/MonadLaws.scala#L39

But, I won't be obsessing over it. I think this requirement is "too much" (after all, `Sync` doesn't have much to say about stack-safety) but it's all in all harmless (minus slight performance penalty for `defer`\`delay`) and fixing it does not fit this PR. Thanks @ArturGajowy for the inspiring discussion

ArturGajowy: Notice the test you linked to doesn't enforce stacksafety for running tailrecM with tailrecM - unlike the one I contributed to ZIO does:
https://github.com/zio/zio/pull/795/files#diff-b2ebf0e7fbc67d716ccbed5ecc008736R11-R17

I guess it doesn't fit the stacksafety requirements for tailrecM though, as no non-suspending monad can pass that test. Notice that very bug was caught by our stacksafety tests when I was taking ZIO for a ride in rholang tests.

> Sync doesn't have much to say about stack-safety

Stacksafety is achieved via suspension, which is the purpose of Sync.
Defer mentions suspending the computation too, but itself is not enough for our needs in the EqualsM/HashM (need Monad). Maybe we should just go with Monad + Defer?

marcin-rzeznicki: > Notice the test you linked to doesn't enforce stacksafety for running tailrecM with tailrecM

Ah, right. They sweep it under the carpet by saying: `constant stack space relative to f`.

> Stacksafety is achieved via suspension, which is the purpose of Sync.

Given that: `its implementation of flatMap is stack safe, meaning that you can describe tailRecM in terms of it ` then you can conclude that (it's the law I believe): `Sync.suspend(f) <-> Sync(f).flatten`. So, by the same reasoning, you can write everything you'd write with `suspend` using `flatMap` for any monad that's stack-safe (or maybe I am missing something here)

But, ok, I'm buying your arguments - with "normal" `Monad.tailRecM` you have to code carefully. I do not have objections then

models/src/main/scala/coop/rchain/models/NormalizerEnv.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.models
+
+import com.google.protobuf.ByteString
+import coop.rchain.casper.protocol.DeployDataProto
+import coop.rchain.crypto.PublicKey
+import coop.rchain.models.rholang.implicits._
+
+object NormalizerEnv {
+  type UriString     = String
+  type NormalizerEnv = Map[UriString, Par]
```
marcin-rzeznicki, line 10: Now, I actually have an objection. I am not sure if this is the right type to model the environment. Why?
1. there is no control over what's in it
1. I have to always check if I can get the value I **know** I put in (or must be in).

I am not sure if this is overkill but can we possibly model it like shapeless record (with literal Strings as keys) and have the precise type of the Env for System Deploys?
I am taking a `no` for an answer BTW :smile: @JosDenmark @ArturGajowy @tymm ?

tymm: It could actually be `Map[UriString, Expr | GUnforgeable]`.
However it would make the Protobuf and possibly the `HashM` and `EqualM` instances more complex.

If no one has any objections, I would prefer creating a ticket for that and come back to that in the future.

JosephDenman: @marcin-rzeznicki I relate, but this PR is blocking the rest of the cost-accounting work.

marcin-rzeznicki: I hear you gents - rest of it is looking great. Approved

marcin-rzeznicki (approved):
LGTM. (Maybe some time later we can model it in a "type-safe" way and that would be beneficial for reasoning about what kind of Env a deploy needs.  Just let's keep it mind)

PR #2767, opened 2019-10-11
Compare Evaluation Costs in Replay

JosephDenman:


casper/src/test/scala/coop/rchain/casper/EstimatorHelperTest.scala
```diff
@@ -110,24 +110,6 @@ class EstimatorHelperTest
       }
   }
 
-  it should "conflict if their deploys conflict in the paymentLog" in withStorage {
```
ArturGajowy, line 113: a thought: once we introduce system deploys, we need to remember to consider them when determining conflicts. I'll add a test for that in the gist.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -307,24 +307,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+               case internalErrors: InternalErrors =>
+                 (deploy.some, internalErrors: Failed).some.pure[F]
                case replayStatus =>
+                 /* Regardless of success or failure, verify that deploy status' match. */
                  if (status.isFailed != replayStatus.isFailed)
-                   (deploy.some, ReplayStatusMismatch(replayStatus, status): Failed).some
-                     .pure[F]
+                   (deploy.some, ReplayStatusMismatch(replayStatus, status): Failed).some.pure[F]
+                 /* Error-throwing Rholang programs do not have deterministic evaluation costs
+                    and event logs, so they cannot be reliably compared. */
```
ArturGajowy: :ok_hand:

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -324,20 +324,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
-                         Log[F].error(s"Failed during processing of deploy: ${processedDeploy}") >>
-                           (none[DeployData], UnusedCommEvent(ex): Failed).some
-                             .pure[F]
-                       }
-                       case Left(ex) =>
-                         (none[DeployData], UserErrors(Vector(ex)): Failed).some
-                           .pure[F]
-                     }
-                 }
+                     .revertToSoftCheckpoint(softCheckpoint)
```
ArturGajowy: This looks like an unrelated change for `Simplify syntax in `replayComputeState`.`, so I'm not sure if it's correct.

It also seems that we should revert to soft-checkpoint on cost discrepancy and event log discrepancy as well - currently we do so only for rholang execution errors.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -288,19 +287,19 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
-              if (ts.isEmpty)
-                ().asRight[ReplayFailure].asRight[Seq[InternalProcessedDeploy]].pure[F]
-              else replayDeploy(ts.head).map(_.map(_.asLeft[Unit]).toRight(ts.tail))
-            }
-          )
-      _   <- EitherT.right(Span[F].mark("before-replay-deploys-create-checkpoint"))
-      res <- EitherT.right[ReplayFailure](runtime.replaySpace.createCheckpoint())
-    } yield res.root.toByteString).value
+  ): F[Either[ReplayFailure, StateHash]] = {
+    import cats.implicits.{catsStdInstancesForList, catsStdInstancesForOption}
```
ArturGajowy: I'd go with the less scary `cats.instances.{list._, option._}` or however scala wants the syntax :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -288,19 +287,19 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+    runtime.replaySpace.reset(Blake2b256Hash.fromByteString(startHash)) >> terms.toList.traverse(
+      replayDeploy
+    ) >>= { replayedDeploys =>
+      replayedDeploys.foldK match {
+        case Some(replayError) => replayError.asLeft[StateHash].pure[F]
+        case None =>
+          Span[F].mark("before-replay-deploys-create-checkpoint") >>
+            runtime.replaySpace.createCheckpoint().map(_.root.toByteString.asRight[ReplayFailure])
+      }
```
ArturGajowy: This is going to replay all deploys even if the first fails.

I think we should make/convert the result type of replayDeploy to be Either[ReplayError, Unit] and do a `terms.toList.foldM`
(edit: we want to return via F, so we'd need to raiseError whenever the F contains a Left to short-circuit the foldM) 

Also, I think the changes you did would be more readable if the for-comprehension stayed despite the EitherT going away :)

ArturGajowy: See for example https://github.com/rchain/rchain/blob/2485709556f9d1d22db6aedd34e2ebcca6ab9e57/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala#L214-L222

ArturGajowy: Actually, isn't this method a third copy of 'replay block deploys' code?
See the comment at https://github.com/rchain/rchain/blob/2485709556f9d1d22db6aedd34e2ebcca6ab9e57/casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala#L66-L67

ArturGajowy (commented):
A couple comments for your consideration :)

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosephDenman: This was caused by a bad merge. Good catch!

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -307,24 +307,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
-      _                    <- Span[F].mark("before-replay-deploy-compute-effect")
-      replayEvaluateResult <- computeEffect(runtime, runtime.replayReducer)(processedDeploy.deploy)
-      //TODO: compare replay deploy cost to given deploy cost
-      EvaluateResult(_, errors) = replayEvaluateResult
-      _                         <- Span[F].mark("before-replay-deploy-status")
+      _                                           <- runtime.replaySpace.rig(processedDeploy.deployLog)
+      softCheckpoint                              <- runtime.replaySpace.createSoftCheckpoint()
+      _                                           <- Span[F].mark("before-replay-deploy-compute-effect")
+      replayEvaluateResult                        <- computeEffect(runtime, runtime.replayReducer)(processedDeploy.deploy)
+      EvaluateResult(Cost(replayCost, _), errors) = replayEvaluateResult
```
tymm: Looks like you actually don't use `replayCost` anywhere.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -307,24 +307,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
                case replayStatus =>
+                 /* Regardless of success or failure, verify that deploy status' match. */
                  if (status.isFailed != replayStatus.isFailed)
-                   (deploy.some, ReplayStatusMismatch(replayStatus, status): Failed).some
-                     .pure[F]
+                   (deploy.some, ReplayStatusMismatch(replayStatus, status): Failed).some.pure[F]
+                 /* Error-throwing Rholang programs do not have deterministic evaluation costs
+                    and event logs, so they cannot be reliably compared. */
                  else if (errors.nonEmpty)
-                   runtime.replaySpace.revertToSoftCheckpoint(softCheckpoint) >> none[ReplayFailure]
```
tymm, line 319: Why is it okay to not revert here?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -307,24 +307,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
     for {
-      _                    <- runtime.replaySpace.rig(processedDeploy.deployLog)
-      softCheckpoint       <- runtime.replaySpace.createSoftCheckpoint()
-      _                    <- Span[F].mark("before-replay-deploy-compute-effect")
-      replayEvaluateResult <- computeEffect(runtime, runtime.replayReducer)(processedDeploy.deploy)
-      //TODO: compare replay deploy cost to given deploy cost
-      EvaluateResult(_, errors) = replayEvaluateResult
-      _                         <- Span[F].mark("before-replay-deploy-status")
+      _                                           <- runtime.replaySpace.rig(processedDeploy.deployLog)
+      softCheckpoint                              <- runtime.replaySpace.createSoftCheckpoint()
```
tymm: Can we get rid of this line if the revert further down is not needed anymore?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
tymm: Also where is the check actually happening?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -324,20 +324,22 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
-                       }
-                       case Left(ex) =>
-                         (none[DeployData], UserErrors(Vector(ex)): Failed).some
-                           .pure[F]
-                     }
-                 }
+                     .revertToSoftCheckpoint(softCheckpoint)
+                     .as(none[ReplayFailure])
+                 /* Since there are no errors, verify evaluation costs and COMM events match. */
+                 else if (replayCost != cost.cost)
```
tymm: Oh, here the check happens.
This should go into the previous commit.

tymm (approved):
Nice!
Please fix commits before merging.

JosephDenman:
This is ready for another review!

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -324,14 +324,12 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                      .attempt
                      .flatMap {
                        case Right(_) => none[ReplayFailure].pure[F]
-                       case Left(ex: ReplayException) => {
-                         Log[F].error(s"Failed during processing of deploy: ${processedDeploy}") >>
-                           (none[DeployData], UnusedCommEvent(ex): Failed).some
-                             .pure[F]
-                       }
+                       case Left(ex: ReplayException) =>
+                         runtime.replaySpace.revertToSoftCheckpoint(softCheckpoint) >> Log[F].error(
```
ArturGajowy: I'm pretty sure reverting should be done after any exception in replay, not just ReplayException. Especially that the other `Left` case mentions `UserErrors`.

ArturGajowy: Either that, or we should not do the revert in here at all.

In fact, I can see only two options:

1) we have a bug currently, and this change needs a test confirming that bug
2) there's no bug, so revert is done elsewhere. In that case we should either:
  - a) move all the revert logic in replay here
  - b) refrain from this change, e.g. if a) is not possible

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -295,43 +291,61 @@ object ReportingCasper {
...
+                                        ReplayFailure
+                                          .internalError(deploy, throwable)
+                                          .asLeft[Seq[ReportingEvent]]
+                                          .pure[F]
+                                    }
+                                  }
+                              }
+                          case Left(throwable) =>
+                            runtime.reportingSpace
+                              .revertToSoftCheckpoint(softCheckpoint)
```
ArturGajowy: It would seem that we `revertToSoftCheckpoint` iff we're returning a `ReplayFailure(...).asLeft`. I think it'd make sense to de-duplicate the revert code and make the correlation explicit, so we don't forget about it should there be more error ReplayFailure scenarios.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -303,47 +299,66 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+                              .as(
+                                ReplayFailure
+                                  .replayCostMismatch(deploy, cost.cost, replayCost.value)
+                                  .some
+                              )
+                          else
+                            runtime.replaySpace.checkReplayData().attempt >>= {
+                              case Right(_) => none[ReplayFailure].pure[F]
+                              case Left(throwable) =>
+                                runtime.replaySpace.revertToSoftCheckpoint(softCheckpoint) >> {
```
ArturGajowy: my comment in ReportingCasper applies here as well - we should do `revertToSoftCheckpoint` once, iff we're returning a `ReplayFailure`

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -303,47 +299,66 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+                                  .replayStatusMismatch(isFailed, replayUserErrors.nonEmpty)
+                                  .some
+                              )
+                          /* Error-throwing Rholang programs do not have deterministic evaluation costs
+                             and event logs, so they cannot be reliably compared. */
+                          else if (replayUserErrors.nonEmpty)
+                            runtime.replaySpace
+                              .revertToSoftCheckpoint(softCheckpoint)
+                              .as(none[ReplayFailure])
+                          /* Since there are no errors, verify evaluation costs and COMM events match. */
```
ArturGajowy: Those are one of the most helpful comments I've ever seen in code, thank you! :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -68,26 +76,38 @@ object Interpreter {
...
+                                                          ).pure[F]
+                                                      }
+                                   } yield evaluateResult
+                                 case Left(interpreterError: InterpreterError) =>
+                                   EvaluateResult(parsingCost, Vector(interpreterError)).pure[F]
+                                 case Left(throwable) => throwable.raiseError[F, EvaluateResult]
+                               }
+                             case Left(interpreterError: InterpreterError) =>
+                               EvaluateResult(parsingCost, Vector(interpreterError)).pure[F]
+                             case Left(throwable) => throwable.raiseError[F, EvaluateResult]
```
ArturGajowy, line 111: I'm not sure what happened here, but I think I prefer the previous version... :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -205,29 +205,29 @@ class DebruijnInterpreter[M[_], F[_]](
...
+      }
     }
 
   private def reportErrors(process: M[Unit]): M[Unit] =
     process.handleErrorWith {
-      case e @ OutOfPhlogistonsError => e.raiseError[M, Unit]
-      case e                         => fTell.tell(e)
+      case error @ OutOfPhlogistonsError => error.raiseError[M, Unit]
+      case error: InterpreterError       => fTell.tell(error)
+      case error                         => error.raiseError[M, Unit]
```
ArturGajowy, line 232: Why don't we just `.raiseError` in all cases? Did we try that approach in the 'error handling vs tuplespace corruption' investigation?

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -35,13 +36,20 @@ object BoolNormalizeMatcher {
...
+        M.fromTry(
+            Try(gi.longliteral_.toLong).adaptError {
+              case e: NumberFormatException => NormalizerError(e.getMessage)
+            }
+          )
+          .map { long =>
+            Expr(GInt(long))
+          }
+      case gs: GroundString => Expr(GString(stripString(gs.stringliteral_))).pure[M]
+      case gu: GroundUri    => Expr(GUri(stripUri(gu.uriliteral_))).pure[M]
```
ArturGajowy, line 53: A thought: @marcin-rzeznicki's combo of:
```
1. remove `import cats.implicits._`
2. add `import cats.syntax.all._`
3. add the needed `import cats.instances.$datatype._`
   in the nearest enclosing scope, creating one if needed
```
can usually be done in less than 5 mins per file. Maybe it makes sense to do it the moment we see a type annotation like those in `.pure[M]`?

ArturGajowy (commented):
Good job! I don't think I'll have much more comments than those below.
I think it'd make most sense if s/b else did a review though, to a point where they're comfortable giving a :heavy_check_mark:. It's most likely not me who's going to build upon that after all... @marcin-rzeznicki you seem to be the natural candidate :)

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
JosephDenman: I'm catching all of the `InterpreterErrors` and returning them in `EvaluateResult` and bubbling all other throwables up to the next level. I don't really see a less verbose way of doing this, and I'm not confident that all errors coming from the normalizer have been caught and converted to `InterpreterError`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
JosephDenman: My best guess (and it must be a guess because there's no documentation) is that `Vector[InterpreterError]` is used to report errors so that errors can be reliably compared upon replay, despite the fact that errors are not compared upon replay 🥇. We can probably remove the error-log and just use the `raiseError` channel, but I think that should be delayed until after deploy charging is in.

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
JosephDenman: I've taken a few minutes to try this in the normalizer. It's not yielding the same reductions as this method has in other parts of the code base. I'm not convinced it merits investigating why right now.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosephDenman: I agree and a subsequent commit does just that 👍

tymm: Why was it possible to not revert before?

rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -35,13 +36,20 @@ object BoolNormalizeMatcher {
...
       case gi: GroundInt =>
-        GInt(gi.longliteral_.toLong) //TODO raise NumberFormatException in a pure way
-      case gs: GroundString => GString(stripString(gs.stringliteral_))
-      case gu: GroundUri    => GUri(stripUri(gu.uriliteral_))
+        M.fromTry(
+            Try(gi.longliteral_.toLong).adaptError {
+              case e: NumberFormatException => NormalizerError(e.getMessage)
+            }
+          )
+          .map { long =>
```
tymm, line 49: Nitpick: `()` instead of `{}` but who cares :)

rholang/src/test/scala/coop/rchain/rholang/interpreter/NormalizeTest.scala
```diff
@@ -18,6 +18,7 @@ import coop.rchain.models.Expr.ExprInstance._
 import coop.rchain.models.Var.VarInstance._
 import coop.rchain.models.Var.WildcardMsg
 import coop.rchain.models._
+import coop.rchain.models.rholang.implicits
```
tymm, line 22: Looks like this is not needed.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -50,11 +52,17 @@ object Interpreter {
...
+          injAttempt(runtime.reducer, runtime.errorLog, term, initialPhlo, normalizerEnv).attempt >>= {
+            case Right(evaluateResult) =>
+              if (evaluateResult.errors.nonEmpty)
+                runtime.space.revertToSoftCheckpoint(checkpoint).as(evaluateResult)
+              else evaluateResult.pure[F]
+            case Left(throwable) =>
+              runtime.space.revertToSoftCheckpoint(checkpoint) >> throwable
+                .raiseError[F, EvaluateResult]
+          }
+        }
```
tymm, line 68: Nitpick: I would prefer a for-comprehension here.

tymm (approved):
Looks great as far as I can tell :)
I'll give it an approval but please take it with a grain of salt: I would need a lot more time to be very confident about my approval since I am not familiar with some parts of the code here and the PR is rather big.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosephDenman: That's a good question. Now that I'm looking at it, I think reverting in case of an invalid replay status is overkill, since these eventually halt processing of the block. The state is reset on every call to `replayComputeState` anyway, so the only case that needs reversion are cases where the block is still valid and we don't want to keep the effects of the deploy. Like `computeState`, that's only if there's been a user error in the deploy. Good catch!

ArturGajowy: If I may advise:

> reverting in case of an invalid replay status is overkill,
> (...)
> the only case that needs reversion are cases where the block is still valid and we don't want to keep the effects of the deploy.

That's one way to think about it. Another would be:
 - we reset the state before propose/replay starts, b/c we need to start from a given preState
 - each `RuntimeManager.{process,replay}[System]Deploy` call leaves the system in a well-defined state.

One could argue that if the replay is to be interrupted, it doesn't matter anymore what state we leave the TS in. I'd counter-argue that it's better to have consistent error handling, with simple rules. E.g. "error => revert".

That's also what I suggested to @marcin-rzeznicki.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -295,43 +289,46 @@ object ReportingCasper {
...
+                                    .unusedCOMMEvent(deploy, replayException)
+                                    .asLeft[Seq[ReportingEvent]]
+                                    .pure[F]
+                                case Left(throwable) =>
+                                  ReplayFailure
+                                    .internalError(deploy, throwable)
+                                    .asLeft[Seq[ReportingEvent]]
+                                    .pure[F]
+                              }
+                          case Left(throwable) =>
```
marcin-rzeznicki: How about:

```scala
        replayEvaluateResult <- computeEffect(runtime, runtime.replayReducer)(processedDeploy.deploy)
        failureEither <- EitherT.rightT[F, ReplayFailure](replayEvaluateResult)
                          .ensureOr(
                            result =>  ReplayFailure.replayStatusMismatch(isFailed, result.executionFailed)
                          )(result => isFailed != result.executionFailed)
                          .ensureOr(
                            result => ReplayFailure.replayCostMismatch(deploy, cost.cost, result.cost.value)
                          )(result => cost.cost != result.cost.value)
                          .semiflatMap(
                            result =>
                              if (result.executionFailed)
                                runtime.reportingSpace.revertToSoftCheckpoint(softCheckpoint)
                              else runtime.reportingSpace.checkReplayData()
                          )
                          .semiflatMap(_ => runtime.reportingSpace.getReport)
                          .value
                          .recover {
                            case (replayException: ReplayException) =>
                              Left(ReplayFailure.unusedCOMMEvent(deploy, replayException))
                            case throwable => Left(ReplayFailure.internalError(deploy, throwable))
                          }
```

marcin-rzeznicki: Just please make sure it does everything the original one did :smile: 

EDIT: Oops, sorry, the final `recover` must actually be tied to the whole computation (rather than to a single branch) to catch errors from `computeEffect`.

marcin-rzeznicki (changes requested):

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
marcin-rzeznicki: IMO it has the following benefits:
1. eliminates duplication in error handling
1. has much cleaner "happy-path"
1. you can actually read what it does (well, minus our formatting style that was apparently chosen by a person programming on teletype (maxColumn = 100, really?)

YMMV

JosephDenman: Thanks! One important difference between your code and mine is that yours compares costs before checking for user errors and mine does the opposite. We can only compare costs if there are no errors since costs become non-deterministic in the event of errors. At any rate, this is very helpful.

marcin-rzeznicki: Ah, I see it. Interesting. Do you want me to attempt (hehe, pun intended) to fix the example?

marcin-rzeznicki: So I have another suggestion. If costs are non-det in presence of failures than maybe it should be reflected in the type. Looks like simmple ADT will do (`data EvaluateResult = Correct cost | CatastropheOfBiblicalProportions errors` )

JosephDenman: > So I have another suggestion. If costs are non-det in presence of failures than maybe it should be reflected in the type. Looks like simmple ADT will do (data EvaluateResult = Correct cost | CatastropheOfBiblicalProportions errors )

I'll repeat what I wrote to you in Discord: 

Also, I've tried your ADT suggestion. I think it's unnecessary. Eventually (possibly/hopefully before main-net), we'll be able to compare program errors reliably. In that cost, it will make sense to return an ADT that includes cost and errors. Right now, it introduces unnecessary branching in RuntimeManager, since I would have to duplicate the check that failure status' match. Whichever ADT I get back during replay (one of two), I have to verify that I got the same one last time. I can appreciate the motivation though :thumbsup:

marcin-rzeznicki: Agreed on that

marcin-rzeznicki: > One important difference between your code and mine is that yours compares costs before checking for user errors and mine does the opposite. We can only compare costs if there are no errors since costs become non-deterministic in the event of errors. 

So it seems to me that this simple change will make it right:
```
+ result => !result.executionFailed && cost.cost != result.cost.value)
- result => cost.cost != result.cost.value)
```

Costs are not compared, subsequent `semiflatMap` calls `revertToCheckpoint` and they lived happily everafter

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -295,43 +289,42 @@ object ReportingCasper {
...
+                          )
+                          .ensureOr(
+                            result => ReplayFailure.replayStatusMismatch(isFailed, result.isFailed)
+                          )(result => isFailed == result.isFailed)
+                          .ensureOr(
+                            result =>
+                              ReplayFailure.replayCostMismatch(deploy, cost.cost, result.cost.value)
+                          )(
+                            result =>
+                              if (!result.isFailed) cost.cost == result.cost.value
```
marcin-rzeznicki: Really minor nitpick: it is `result.isFailed || cost.cost == ...` :smiley:

marcin-rzeznicki (approved):
Terrific work!

PR #2768, opened 2019-10-11
RCHAIN-3881: Prohibit terms wider than Short.MAX_VALUE splits/terms

tgrospic:
## Overview
Bulletproof unforgeable-name namespace uniqueness by prohibiting terms wider than Short.MAX_VALUE splits/terms

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3881

### Notes
~Test of eval with max split size takes almost 30 seconds, with just empty sends, so it's not included.~
Test of eval with max split size is included and it takes about 10 seconds.

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -178,7 +178,14 @@ class DebruijnInterpreter[M[_], F[_]](
       else rand.splitByte(id.toByte)
 
     val indexedTerms = terms.zipWithIndex.toList
-    if (terms.size <= Reduce.parallelism) {
+    // Term split size is limited to half of Int16 because other half is for injecting
+    // things to system deploys through NormalizerEnv
+    val termSplitLimit = Short.MaxValue
+    if (terms.size > termSplitLimit) {
+      ReduceError(
+        s"The number of Pars in the term is ${terms.size}, which exceeds the limit of ${termSplitLimit}."
+      ).raiseError[M, Unit]
+    } else if (terms.size <= Reduce.parallelism) {
       indexedTerms.parTraverse_ {
         case (term, index) =>
           val random = split(index)
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
@@ -2511,4 +2511,38 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
 
     task.runSyncUnsafe(timeout)
   }
+
+  "term split size max" should "be evaluated (Int16/2 - 1)" in {
+    implicit val errorLog = new ErrorLog[Task]()
+    val p                 = New(bindCount = 1)
+    val news              = Seq.fill(Short.MaxValue)(p)
+    val proc              = Par(news = news)
+
+    val result = withTestSpace(errorLog) {
+      case TestFixture(_, reducer) =>
+        implicit val env = Env[Par]()
+        val task         = reducer.eval(proc)
+        Await.result(task.runToFuture, 30.seconds)
+    }
+
+    result should be(())
+  }
+
+  "term split size" should "be limited to Int16/2" in {
+    implicit val errorLog = new ErrorLog[Task]()
+    val p                 = Send(Par(), Seq(Par()))
+    val sends             = Seq.fill(Short.MaxValue + 1)(p)
+    val proc              = Par(sends)
+
+    val result = withTestSpace(errorLog) {
+      case TestFixture(_, reducer) =>
+        implicit val env = Env[Par]()
+        val task         = reducer.eval(proc)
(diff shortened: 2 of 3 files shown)


tgrospic:
bors try

tgrospic:
bors delegate+

zsluedem:
bors delegate+

zsluedem:
@woky Help @tgrospic with the permission

tgrospic:
bors try

tgrospic:
I can run bors now, thanks @zsluedem and @woky.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -182,8 +182,13 @@ class DebruijnInterpreter[M[_], F[_]](
       else if (terms.size > 256) rand.splitShort(id.toShort)
       else rand.splitByte(id.toByte)
 
-    val indexedTerms = terms.zipWithIndex.toList
-    if (terms.size <= Reduce.parallelism) {
+    val indexedTerms   = terms.zipWithIndex.toList
+    val termSplitLimit = Short.MaxValue + 1
```
ArturGajowy: this should say `val termSplitLimit = Short.MaxValue`

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2511,4 +2511,21 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
 
     task.runSyncUnsafe(timeout)
   }
+
+  "term split size" should "be limited to Short.MaxValue" in {
+    implicit val errorLog = new ErrorLog[Task]()
+    val sends             = Seq.fill(Short.MaxValue + 2)(Send())
```
ArturGajowy: The minimum failing case should be `Short.MaxValue + 1`.

We should also have a positive test showing that `Short.MaxValue` pars is OK

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2511,4 +2511,21 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
...
+    val proc              = Par(sends)
+
+    val result = withTestSpace(errorLog) {
+      case TestFixture(_, reducer) =>
+        implicit val env = Env[Par]()
+        val task         = reducer.eval(proc)
+        Await.result(task.failed.runToFuture, 0.seconds)
```
ArturGajowy: The timeout here gives me chills. If it's really that snappy, let's at least have 1.second

ArturGajowy (dismissed):
Please see comments

JosephDenman:
After Artur's comments are addressed, LGTM.

JosephDenman (approved):

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tgrospic: Should this be the same as for bytes, to detect overflow? Limit is the count and not the index. Am I missing something?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
tgrospic: Do you think on test I mentioned in the description? Can we make it faster?

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
tgrospic: This is on purpose. It's really like sync operation, if terms size is to large, it should fail immediately.

But for sure it depends on internal implementation if 0 sec. can sometimes fail. Can `10.milis` be more than enough for a method call?

tgrospic:
`Short.MaxValue` gives signed value which is incorrect. I fixed that and changed to hex literal to be more clear.

ArturGajowy:
> Short.MaxValue gives signed value which is incorrect

It is not. We reserve half of the Short domain for injecting things to system deploys through NormalizerEnv.

tgrospic:
I fixed limit and test duration, it's ready for next round of reviews.

sebtomba (approved):

sebtomba (approved):

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -2511,4 +2511,38 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
 
     task.runSyncUnsafe(timeout)
   }
+
+  "term split size max" should "be evaluated (Int16/2 - 1)" in {
+    implicit val errorLog = new ErrorLog[Task]()
+    val p                 = New(bindCount=1)
```
tgrospic: FYI @ArturGajowy with `New` positive test is cca. 3 times faster then with sends or receives.

PR #2774, opened 2019-10-18
Add multisig vault

ghost:
## Overview
Adds a multisig vault implementation.

### JIRA ticket:
RCHAIN-3196

### Notes
Needs tests.


casper/src/main/resources/RevVault.rho
```diff
@@ -27,178 +30,347 @@ in {
...
-                  @Either!("fromBoolean <-", *depositSuccessCh, "BUG FOUND: purse deposit failed", *retCh)
+              purse!("deposit", amount, *depositPurse, *depositSuccessCh) |
+              @Either!("fromBoolean <-", *depositSuccessCh, "BUG FOUND: purse deposit failed", *retCh)
+            }
+          }
+        } |
+
+        // MultiSig
+        // Returns channel on which to invoke methods
+        contract RevVault(@"createMultiSig", @publicKeys, @authKeys, @quorumSize, ret) = {
```
ArturGajowy: the way I imagined it was for MultiSig to be a separate contract, only wrapping (delegating to) RevVault, without knowing its internal details / sharing any names. Is there a reason why this is not possible?

ghost: This code is the culprit: `@ListOps!("map", publicKeys, *_fromPublicKey, *authPubKeysCh)`

It can be an entirely separate thing if you don't expect a deployer authkey from RevVault to work in a multisig vault.

ArturGajowy: Yes, totally. There's no need for auth-key reuse, and I'd claim it's safer and more intuitive for each authKey to only have one purpose. And it reduces coupling between RevVault and Multisig.

ghost:
I copied the RevVault test to make the multisig one because at some point we have to transfer money between normal vaults and multisig vaults.  Most of the existing tests in that file can go, but they're useful for reference to see how to create and work with standard vaults.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
```diff
@@ -71,6 +71,13 @@ object StandardDeploys {
       1559156183943L
     )
 
+  def multiSigRevVault: DeployData =
+    toDeploy(
+      CompiledRholangSource("MultiSigRevVault.rho", NormalizerEnv.Empty),
+      "04fe2eb1e0e7462b1a8f64600389e1e76727f8b2d38804eaa4b48f7a7d6715130fc24d3c4dac2d8bdc19e0b49879dbaf07c30773cd9740a9d14a092ef76339207a",
```
zsluedem, line 73: Actually I wonder how this hash is generated? Is there a rule?

casper/src/test/resources/MultiSigRevVaultTest.rho
```diff
@@ -0,0 +1,156 @@
...
+        }
+      } |
+
+      contract testTransferOut(rhoSpec, @(RevVault, MultiSigRevVault), ackCh) = {
+        new genesisVaultCh, carolUnf, carolAuthCh, carolVaultCh, ret1, ret2, retOk in {
+          withVaultAndIdentityOf!(genesisPubKey, *genesisVaultCh) |
+          @MultiSigRevVault!("unforgeableAuthKey", *carolUnf, *carolAuthCh) |
+          for (@carolAuth <- carolAuthCh) {
+            // Single sig as a special case of multisig.
+            @MultiSigRevVault!("create", [], [carolAuth], 1, *carolVaultCh) |
```
zsluedem: We need test for more than one auth key

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
ghost: It's from the header of MultiSigRevVault.rho, which was generated by RegistrySigGen.scala.

casper/src/test/resources/MultiSigRevVaultTest.rho
ghost: Yes.  Feel free to add daveUnf, daveAuthCh, eveUnf, eveAuthCh, change the quorum size to 2, check that the create method returns (true, (true, 0)), then use the confirm method with dave's authKey to give the second confirmation and check it returns (true, (false, "done")).

ghost: We also need to derive authKeys from public keys and use the developerAuthKey method to check that they work.

casper/src/main/resources/RevVault.rho
ghost: Done.

casper/src/test/resources/MultiSigRevVaultTest.rho
ghost: Rao says the more extensive tests should come in a separate PR.

zsluedem:
bors try

zsluedem (approved):

casper/src/test/resources/MultiSigRevVaultTest.rho
```diff
@@ -0,0 +1,156 @@
+// Match a list of known valid publicKey -> RevAddress pairs to bind them to veriables
```
tgrospic: It's just a typo in the last word.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,215 @@
...
+    ListOpsCh
+in {
+  rl!(`rho:lang:listOps`, *ListOpsCh) |
+  rl!(`rho:rchain:authKey`, *AuthKeyCh) |
+  rl!(`rho:rchain:revVault`, *RevVaultCh) |
+  for(@(_, ListOps) <- ListOpsCh;
+      @(_, AuthKey) <- AuthKeyCh;
+      @(_, RevVault) <- RevVaultCh) {
+
+    contract MultiSigRevVault(@"foo", ret) = { ret!("foo") } |
```
tymm: Leftover?

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,215 @@
...
+                      @revVault!("balance", *ret)
+                    } |
+
+                    // Creates transfer and "signs" with first account.
+                    // returns (true, (true, nonce)) when more sigs required
+                    //         (true, (false, "done")) when only one sig required
+                    //         (false, errorMsg) on failure
+                    contract vault(@"transfer", @targetRevAddress, @amount, @authKey, ret) = {
+                      new idxAuthKeysCh, idxAuthPubKeysCh in {
+                        // Check that authKeys is in the list of authKeys or authPubKeys
```
tymm: Typo: first *authKeys* should be *authKey*.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,215 @@
...
+      if (quorumSize > publicKeys.length() + authKeys.length() or
+          quorumSize <= 0) {
+        ret!((false, "invalid quorum size"))
+      } else {
+        new unf, authKeyCh, authPubKeysCh, RevAddress(`rho:rev:address`), revAddrCh in {
+          RevAddress!("fromUnforgeable", *unf, *revAddrCh) |
+          @RevVault!("unforgeableAuthKey", *unf, *authKeyCh) |
+          @ListOps!("map", publicKeys, *_fromPublicKey, *authPubKeysCh) |
+          for (@vaultAuthKey <- authKeyCh; @authPubKeys <- authPubKeysCh; @revAddr <- revAddrCh) {
+            new vault, nonceStore, nonceMapStore, revVaultCh in {
```
tymm: I would find it is easier to read when we rename `vault` to `multisig`.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,215 @@
...
+        RevAddress!("fromPublicKey", publicKey, *revAddrCh)|
+        for (@revAddress <- revAddrCh) {
+          @AuthKey!("make", (*_multiSigRevVault, revAddress), *ret)
+        }
+      }
+    } |
+    
+    // MultiSig
+    // Returns (false, errorMsg) or (true, (multiSigVault, revAddress, revVault))
+    contract MultiSigRevVault(@"create", @publicKeys, @authKeys, @quorumSize, ret) = {
```
tymm: Isn't it problematic that all `authKeys` have to be in one place in order to create the multisig? Especially since authKeys are the same between multiple multisig instances.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,215 @@
...
+  rl!(`rho:rchain:authKey`, *AuthKeyCh) |
+  rl!(`rho:rchain:revVault`, *RevVaultCh) |
+  for(@(_, ListOps) <- ListOpsCh;
+      @(_, AuthKey) <- AuthKeyCh;
+      @(_, RevVault) <- RevVaultCh) {
+
+    contract MultiSigRevVault(@"foo", ret) = { ret!("foo") } |
+
+    contract MultiSigRevVault(@"deployerAuthKey", deployerId, ret) = {
+      new RevAddress(`rho:rev:address`),
```
tymm: `RevAddress` seems to be not needed here.

casper/src/main/resources/MultiSigRevVault.rho
tymm: Having two parameters (`publicKeys` and `authKeys`), where one is always unused, might be confusing to users.
I agree though that the alternatives are suboptimal as well (using an Either or creating two methods: `createFromPks` and `createFromAuthKeys`).
I would probably slightly prefer the alternative where we create two methods.

tymm (commented):
Looks good!
This is part 1 of my review. I'll come back for a review on the details of the transfer and confirm method.

casper/src/test/resources/MultiSigRevVaultTest.rho
ghost: Added test for 2 of 3.

casper/src/test/resources/MultiSigRevVaultTest.rho
ghost: Done

casper/src/main/resources/MultiSigRevVault.rho
ghost: Done.

casper/src/main/resources/MultiSigRevVault.rho
ghost: Done.

casper/src/main/resources/MultiSigRevVault.rho
ghost: Done

casper/src/main/resources/MultiSigRevVault.rho
ghost: So you're worried that if there are two contracts that want to form a multisig, one of them has to trust the other?  That's reasonable.  At the moment that threat can be mitigated to some extent by doing a code analysis, but that's notoriously unreliable without formal verification.

If I were designing it, I would have used an unsealer (the ocaps version of a public key) instead of an authKey.

casper/src/main/resources/MultiSigRevVault.rho
ghost: Done.

casper/src/main/resources/MultiSigRevVault.rho
ghost: Either or both can be used.

ghost:
bors try

casper/src/main/resources/MultiSigRevVault.rho
tymm: Yes, that's what I'm worried about. I don't really see a point in creating a multisig with `authKeys` when the caller of `Multisig!create` has access to all `authKeys`.

We could provide a `create` method that takes `size`, `publicKeys` and `quorumSize` and returns a contract that then can be used to safely collect `authKeys`.

ghost: It doesn't necessarily have access to all authkeys.  If public keys are used, there's no way to derive the authkeys from them.  Authkeys derived from unforgeable names are only for use by contracts.

An alternative would be for contracts to provide a name of the form `bundle+{P}`, where only the contract can read from it, and the create call creates an unforgeable name, derives an authkey from it, and sends that authkey on the supplied channel.  That would eliminate the need to trust the caller of `create`, but it deviates somewhat from the design doc.

tymm: > Authkeys derived from unforgeable names are only for use by contracts.

Not sure what you mean here? If two contracts derive an `authKey` from an unforgeable name, there still is the problem of bringing the authKeys together in a safe way to then call `Multisig!create`, right?

I agree that everything is fine if only `publicKeys` are provided.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,212 @@
...
+          for (@revVaultAuthKey <- authKeyCh; @authPubKeys <- authPubKeysCh; @revAddr <- revAddrCh) {
+            new multiSig, nonceStore, nonceMapStore, revVaultCh in {
+              @RevVault!("findOrCreate", revAddr, *revVaultCh) |
+              for (@maybeRevVault <- revVaultCh) {
+                match maybeRevVault {
+                  (false, _) => { ret!(maybeRevVault) }
+                  (true, revVault) => {
+                    ret!((true, (*multiSig, revAddr, revVault))) |
+                    // Nonce starts at zero
+                    @(*multiSig, *nonceStore)!(0) |
```
tymm, line 98: I wish I would remember why @ArturGajowy bundle0-ed the `_revVault` name: https://github.com/rchain/rchain/blob/81e96f69c60323079066075b001339e7b9fee067/casper/src/main/resources/RevVault.rho#L141

Maybe it's relevant here.

ghost:
bors try

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,212 @@
...
+                    // returns (true, (true, nonce)) when more sigs required
+                    //         (true, (false, "done")) when only one sig required
+                    //         (false, errorMsg) on failure
+                    contract multiSig(@"confirm", @targetRevAddress, @amount, @authKey, @nonce, ret) = {
+                      // Check that authKeys is in the list of authKeys or authPubKeys
+                      new idxAuthKeysCh, idxAuthPubKeysCh in {
+                        @ListOps!("indexOf", authKeys, authKey, *idxAuthKeysCh) |
+                        @ListOps!("indexOf", authPubKeys, authKey, *idxAuthPubKeysCh) |
+                        for (@idxAuthKeys <- idxAuthKeysCh; @idxAuthPubKeys <- idxAuthPubKeysCh) {
+                          if (idxAuthKeys >= 0 or idxAuthPubKeys >= 0) {
```
tymm: Might be worth it to create a helper method for the check here and in `multiSig!transfer`.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,212 @@
...
+                                      amount != nonceAmount) {
+                                    @(*multiSig, *nonceMapStore)!(nonceMap) |
+                                    ret!((false, "given data does not match stored data"))
+                                  } else {
+                                    if (nonceConfirmers.contains(authKey)) {
+                                      @(*multiSig, *nonceMapStore)!(nonceMap) |
+                                      ret!((false, "already confirmed"))
+                                    } else {
+                                      // If this is the last confirmation
+                                      if (nonceConfirmers.size() == quorumSize - 1) {
```
tymm, line 202: What if two valid confirms get executed in parallel when we are in the state of `nonceConfirmers.size() == quorumSize - 1`? Are you sure that there is no chance that the RevVault transfer won't be executed twice?

tymm (commented):
Second and last part of my review.
I like the confirm logic a lot. Very clean and neat.

Once my questions are answered, I'll approve.

casper/src/main/resources/MultiSigRevVault.rho
ghost: I can alter the design to use sealer/unsealer pairs instead of public/private keys.  Instead of providing the `create` call with an authkey from an unforgeable name, a contract can provide an unsealer.  The unsealer is the ocap version of a signature verification key.

When confirming a transaction, the contract would provide a sealed tuple `(multisig, nonce)` instead of the authkey.  Since that tuple is only ever used once, there's no possibility of a replay attack.

casper/src/main/resources/MultiSigRevVault.rho
ghost: The two confirm calls would race for acquisition of the nonceMap, which behaves like a mutex.  One would acquire it first and update the set of confirmers to Nil.  The other would acquire it second and would receive the "inactive transaction nonce" message.

casper/src/main/resources/MultiSigRevVault.rho
ghost: Yeah, I can't see why.  If the name `@{ownRevAddress | bundle0{*_revVault}}` leaks, a contract can recover `bundle0{*_revVault}` and use it to break the security.

ghost:
bors try

casper/src/main/resources/MultiSigRevVault.rho
ghost: I updated the code to use unsealers for contracts instead of authKeys.  When calling transfer or confirm, the contract needs to seal all the parameters for the call just like a person signs the content when constructing a deploy.  Instead of exposing an `unforgeableAuthKey` method, it now exposes a `makeSealerUnsealer` method.

ghost:
bors try

casper/src/main/resources/MultiSigRevVault.rho
tymm: But currently the following seems possible:
First confirm sets nonceMap
Second confirm can acquire nonceMap
First confirm calls the RevVault transfer
Second confirm calls the RevVault transfer again

I think what needs to be done is to set the nonceMap after the RevVault transfer executed.

ghost:
bors try

ghost:
bors try

casper/src/main/resources/MultiSigRevVault.rho
tymm: Sorry @pyrocto, I had to switch over to reviewing cost accounting. Once I'm done there I'll review your changes.

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,263 @@
...
+
+                    contract verify(@auth, idxAuthPubKeysCh, unsealedCh) = {
+                      new unsealerTest in {
+                        // Check whether auth is in the list of authPubKeys
+                        @ListOps!("indexOf", authPubKeys, auth, *idxAuthPubKeysCh) |
+                        // Check whether auth is a sealed tuple
+                        @ListOps!("partialFold", unsealers, (-1, Nil), *unsealerTest, *unsealedCh) |
+                        contract unsealerTest(head, @(prevPos, _), result) = {
+                          new retH in {
+                            head!(auth, *retH) |
```
tymm, line 117: It's difficult to read because `auth` is not an `auth` at all in this case but a sealed tuple.
Maybe add a comment?

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -0,0 +1,263 @@
...
+                    contract multiSig(@"balance", ret) = {
+                      @revVault!("balance", *ret)
+                    } |
+
+                    contract verify(@auth, idxAuthPubKeysCh, unsealedCh) = {
+                      new unsealerTest in {
+                        // Check whether auth is in the list of authPubKeys
+                        @ListOps!("indexOf", authPubKeys, auth, *idxAuthPubKeysCh) |
+                        // Check whether auth is a sealed tuple
+                        @ListOps!("partialFold", unsealers, (-1, Nil), *unsealerTest, *unsealedCh) |
```
tymm, line 114: Why not make it `(false, Nil)`?

tymm (approved):
LGTM. Great work.
Please see my two comments though.

casper/src/main/resources/MultiSigRevVault.rho
ghost: The first element in the pair is the index of the unsealer in the list.  The value -1 is standard for "not in the list". https://www.scala-lang.org/api/2.12.3/scala/collection/immutable/StringOps.html#indexOf(elem:Char,from:Int):Int

casper/src/main/resources/MultiSigRevVault.rho
ghost: It is a value demonstrating that the request is authorized.  It is not necessarily an auth*Key*, which is why I changed the name.  The contract verify is for internal use only; there are comments on the public-facing API where auth is provided on lines 127 and 174.

PR #2775, opened 2019-10-18
Add new termination condition for isInMainChain

itegulov:
## Overview
While looking at the [flamegraph](https://storage.googleapis.com/heapdumps.bucket.rchain-dev.tk/testnet.rchain-dev.tk/2019-10-06T20-04-12.testnet2-1.node0/ap/finalization.svg), I noticed that `isInMainChain` is responsible for 50% of the CPU time. Although there were some talks about DAG storage being the bottleneck here (and ways how to improve it), I don't think that `isInMainChain` should make as many calls to DAG storage anyway.

My conclusion is that `isInMainChain` was lacking a very important termination condition. It should not try to continue looking for candidate block once target block's number becomes lower or equal than the candidate block's number.

Imagine a block A with two children B and C. If we call `isInMainChain(B, C)`, the current implementation will start looking through C's ancestors trying to find block B but never actually finding it. Eventually it will reach genesis and stop concluding that C is not in the main chain of B. This is critically inefficient when we are talking about 100k ancestors.

### JIRA ticket:
<https://rchain.atlassian.net/browse/RCHAIN-3890>

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -36,13 +36,18 @@ object ProtoUtil {
     import coop.rchain.catscontrib.Catscontrib.ToBooleanF
     import cats.instances.option._
 
-    (candidateMetadata.blockHash == targetBlockHash).pure[F] ||^ {
-      for {
-        targetBlockOpt       <- dag.lookup(targetBlockHash)
-        mainParentOpt        = targetBlockOpt >>= (_.parents.headOption)
-        inMainParentChainOpt <- mainParentOpt.traverse(isInMainChain(dag, candidateMetadata, _))
-      } yield inMainParentChainOpt.getOrElse(false)
-    }
+    (candidateMetadata.blockHash == targetBlockHash).pure[F] ||^
+      dag.lookup(targetBlockHash).flatMap {
+        case Some(targetBlock) =>
+          if (targetBlock.blockNum <= candidateMetadata.blockNum) {
+            false.pure[F]
+          } else {
+            val mainParentOpt = targetBlock.parents.headOption
+            mainParentOpt.traverse(isInMainChain(dag, candidateMetadata, _)).map(_.getOrElse(false))
+          }
+        case None =>
+          false.pure[F]
+      }
   }
 
   def getMainChainUntilDepth[F[_]: Sync: BlockStore](
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -28,20 +28,19 @@ object ProtoUtil {
   /*
    * c is in the blockchain of b iff c == b or c is in the blockchain of the main parent of b
    */
-  // TODO: Move into BlockDAG and remove corresponding param once that is moved over from simulator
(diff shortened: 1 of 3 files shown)


nzpr:
Iirc this flamegraf was calculated when finalizing DAG built with synchrony threshold 0.99, so all blocks should have a single parent. What’s the expected behavior of IsInMainChain in this case?

itegulov:
@nzpr Good point about the flamegraph being from a scenario with synchrony constraint threshold 0.99. However, even though all blocks have a single parent, we still call `isInMainChain` on latest messages of other validators which could have happenned before the block we are trying to test against (i.e. we will be trying to find the candidate block among target block's ancestors when it is in fact its descendant).

Another point I want to make here is that invoking finalization on a regular basis should speed it up overall. We won't have to traverse the entire DAG from the latest messages back to the root of the DAG because we haven't run finalization for a few days. I believe that finalization should be fairly cheap when latest messages are not to far away from the current last finalized block. Especially after this change makes it to the dev branch.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -36,13 +36,18 @@ object ProtoUtil {
...
+        case Some(targetBlock) =>
+          if (targetBlock.blockNum <= candidateMetadata.blockNum) {
+            false.pure[F]
+          } else {
+            val mainParentOpt = targetBlock.parents.headOption
+            mainParentOpt.traverse(isInMainChain(dag, candidateMetadata, _)).map(_.getOrElse(false))
+          }
+        case None =>
+          false.pure[F]
+      }
```
ArturGajowy, line 50:
```suggestion
    (candidateMetadata.blockNum <= targetBlock.blockNum).pure[F] &&^ (
      (candidateMetadata.blockHash == targetBlockHash).pure[F] ||^ {
        for {
          targetBlockOpt       <- dag.lookup(targetBlockHash)
          mainParentOpt        = targetBlockOpt >>= (_.parents.headOption)
          inMainParentChainOpt <- mainParentOpt.traverse(isInMainChain(dag, candidateMetadata, _))
        } yield inMainParentChainOpt.getOrElse(false)
      }
    )
```

ArturGajowy: ~~Alternatively, we could even make the check only once `candidateMetadata.blockNum <= targetBlock.blockNum`, and then delegate to a recursive helper that omits the check.~~

itegulov: We have to look up `targetBlock` first (line 53) so I can't move the check out of that `for` unfortunately. It is also impossible to do this check only once (`targetBlock` is dynamic here and changes the further we go) unless you mean we can deduce how far we should go from the difference between block numbers of target and candidate blocks. If so, I would argue that it brings unnecessary complexity into the method.

ArturGajowy: You're right, my comment about doing the check once is wrong :)

Still we should be able to avoid pattern matching and manual control flow using combinators. We could for example:

```scala
        targetBlockOpt <- dag.lookup(targetBlockHash)
          .map(_.filter(_.blockNum >= candidateMetadata.blockNum))

```

It's a matter of taste though :)

sebtomba (approved):

zsluedem (approved):

PR #2776, opened 2019-10-18
Fair round robin packet dispatcher

sebtomba:
## Overview
Implement a fair round-robin packet dispatcher. The dispatcher strategy for a peer P1 and packet is:
1. Check if the peer P1 is known to the dispatcher and if not add him to the queue (and internal state)
2. Add the packet to the peer's P1 message queue.
3. Pick the next peer P2 in turn from the queue. (This can be a different peer than peer P1)
4. If the peer P2 has a packet in its message queue 
    - Dispatch the packet
    - Remove the packet from the message queue
    - Move peer P2 to the end of the queue
    - Goto 3.
5. If the peer's P2 message queue is empty
    - Skip dispatching
    - If already max number of skips is reached then give up (move peer P2 to the end of the queue)
    - If the max number of retries for peer P2 is reached then remove peer P2 from the queue.
    - Goto 3. (only once and only if P2 == P1)

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3879

### Notes
If max number of skips = 0 and max number of retries = 0 then the dispatcher passes effectively all packets thru.


sebtomba:
bors try

sebtomba:
bors try

casper/src/test/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcherSpec.scala
```diff
@@ -0,0 +1,589 @@
...
+        new TestEnv(10, 3, 10) {
+          validate(
+            for {
+              _ <- dispatcher.ensureSourceExists("A")
+              _ <- dispatcher.ensureSourceExists("B")
+              _ <- dispatcher.ensureSourceExists("C")
+              _ <- dispatcher.dispatch("B", 1)
+              _ <- dispatcher.dispatch("B", 2)
+              _ <- dispatcher.dispatch("C", 3)
+            } yield ()
```
tgrospic, line 483: Very nice work with this PR. :)

One side question. Is there a reason why not use just an applicative in these places? It seems less noisy to me.

```scala
validate(
  dispatcher.ensureSourceExists("A") *>
    dispatcher.ensureSourceExists("B") *>
    dispatcher.ensureSourceExists("C") *>
    dispatcher.dispatch("B", 1) *>
    dispatcher.dispatch("B", 2) *>
    dispatcher.dispatch("C", 3)
)
```

sebtomba: @tgrospic It's a matter of taste. I thought that the code woulde be more readable that way. However, in general, an applicative `*>` gives no guarantee of the order of execution. Potentially both values could get evaluated in parallel. Here it doesn't matter since I'm using `SyncIO`. But to get an equivalent code we would have to use the monadic `>>`. Actually we would safe one `map` (for the `yield`) at the end.

sebtomba:
bors try

sebtomba:
bors try

casper/src/test/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcherSpec.scala
tgrospic: Thanks, this is exactly the answer I was looking for.

sebtomba:
bors try

kayvank (approved):

zsluedem (approved):

itegulov (approved):

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
```diff
@@ -0,0 +1,144 @@
...
+import FairRoundRobinDispatcher._
+import coop.rchain.shared.Log
+
+class FairRoundRobinDispatcher[F[_]: Sync: Log, S: Show, M](
+    filter: M => F[Dispatch],
+    handle: (S, M) => F[Unit],
+    queue: Ref[F, Queue[S]],
+    messages: Ref[F, Map[S, Queue[M]]],
+    retries: Ref[F, Map[S, Int]],
+    skipped: Ref[F, Int],
```
marcin-rzeznicki, line 19: Maybe it'd be easier to have a single structure holding all these. I'm thinking `state: Ref[F, State]` `case class State(q: Queue[S], m: Map[S. Queue[M]], retries: ...)`. I even wouldn't be afraid of higher contention. Many methods in the dispatcher operate on more than one piece of data

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
```diff
@@ -0,0 +1,144 @@
...
+    maxSourceQueueSize: Int,
+    giveUpAfterSkipped: Int,
+    dropSourceAfterRetries: Int
+) {
+  assert(maxSourceQueueSize > 0)
+  assert(giveUpAfterSkipped >= 0)
+  assert(dropSourceAfterRetries >= 0)
+
+  def dispatch(source: S, message: M): F[Unit] =
```
marcin-rzeznicki, line 28: Is this called synchronously? If the answer is no then I'd be wary of potential races eg. you're in the middle of dropping a source (say just before line 85) while the message arrives from the same source and proceeds to line 51. Then you execute the line 85 and boom

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
```diff
@@ -0,0 +1,144 @@
...
+        messages.update(ps => ps.updated(source, ps(source).enqueue(message))) *>
+          retries.update(_.updated(source, 0)),
+        Log[F].debug(s"Dropped message from ${source.show}")
+      )
+
+  private[comm] def handleMessages(source: S): F[Unit] =
+    queue.get.flatMap { q =>
+      if (q.head == source) ().pure
+      else failure(q.head)
+    } >> handleNextMessage // because maybe gave up on failure
```
marcin-rzeznicki: Let's say we have three sources: A, B, C; giveUpAfterSkipped = 5. It's `A` turn but `B` has sent 4 messages cranking the `skipped` counter to 4. Then C sends a single message - it ends up in `failure(C)` and `giveUp(C)` but `C` is a bit innocent here. Am I reading it right? Is it possible?

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  private[comm] val handleNextMessage: F[Unit] =
+    queue.get.map(_.head) >>= { s =>
+      handleMessage(s).ifM(
+        success(s) >> handleNextMessage,
```
marcin-rzeznicki: It doesn't feel right - if `A` is at the head of the queue and enqueued 1k messages and a message from `B` arrives then it is going to wait till these 1k messages get handled. Or maybe there is something else at play?

marcin-rzeznicki (commented):
I left some questions. Please respond

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
sebtomba: As you said, it's `A`s turn. If `C` sent the last message (to skip) then `q.head` is still `A`.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
sebtomba: Those `Ref`s are just classic variables. However, decorated with some pure FP. I thought that keeping them apart would keep the code smaller and more readable (and more performant). However, I'll check and compare it with the compound `State` solution.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
sebtomba: Yes, this code is called synchronously. There is a smart constructor that uses a `Semaphore` for concurrent use cases.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
sebtomba: No, each `success` rotates the queue. So after the first message dispatched for `A` it's `B`'s turn.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
marcin-rzeznicki: I guess it's actually _less_ performant as each `update`, `modify` etc is doing a CAS plus you need to glue it with more `map` `flatMap` etc. Please consider

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
marcin-rzeznicki: Ah, ok, now I see that this part is the key. Thanks for explaining.

PR #2778, opened 2019-10-21
Cost accounting I

marcin-rzeznicki:
## Overview
Cost accounting
The groundwork for the cost accounting has been laid out by @ArturGajowy  in [this gist](https://gist.github.com/ArturGajowy/73e481767548b084a8e5d7c19bb1f70c/4766ea917285cb6909231ec25977993ad6d1ab1a). This PR is joint work between @JosDenmark and me

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3767
https://rchain.atlassian.net/browse/RCHAIN-3624
https://rchain.atlassian.net/browse/RCHAIN-3625

### Notes

Currently it contains the following:
- adds notion of SystemDeploy
- implements PreCharge
- implements Refund
- implements Slash
- implements CheckBalance
- adds type-safe inputs (NormalizerEnv)
- integrates with existing RuntimeManager flow to _play_ and _replay_ system deploys
- contains tests for basic scenarios (they're not completed yet but @JosDenmark has made sure that basic scenarios work)

In subsequent PRs we'll be adding:
- CloseBlock
- more tests
- charging logic

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
@@ -1,63 +1,58 @@
 package coop.rchain.casper.genesis.contracts
 
-import coop.rchain.models.NormalizerEnv.NormalizerEnv
-import coop.rchain.models.{NormalizerEnv, Par}
+import coop.rchain.models.NormalizerEnv
 import coop.rchain.rholang.build.CompiledRholangSource
-import coop.rchain.rholang.interpreter.ParBuilder
-import monix.eval.Coeval
 
-final case class RevGenerator(userVaults: Seq[Vault], supply: Long) extends CompiledRholangSource {
-
+final class RevGenerator private (supply: Long, code: String)
+    extends CompiledRholangSource(code, NormalizerEnv.Empty) {
   val path: String = "<synthetic in Rev.scala>"
+}
 
-  val code: String =
-    s""" new rl(`rho:registry:lookup`), revVaultCh in {
-       #   rl!(`rho:rchain:revVault`, *revVaultCh) |
-       #   for (@(_, RevVault) <- revVaultCh) {
-       #     new ret in {
-       #       @RevVault!("init", *ret) |
-       #       for (vaultMapStore, initVault <- ret) {
-       #         ${vaultInitCode()}
-       #       }
-       #     }
-       #   }
-       # }
-     """.stripMargin('#')
-
-  val normalizerEnv: NormalizerEnv = NormalizerEnv.Empty
-
-  val term: Par =
-    ParBuilder[Coeval]
-      .buildNormalizedTerm(code, normalizerEnv)
-      .value()
-
-  private def vaultInitCode(): String =
-    if (userVaults.isEmpty) {
-      "vaultMapStore!({})"
-    } else {
-      s"""
-        #new ${userVaults.indices.map("x" + _).mkString(", ")} in {
-        #  vaultMapStore!(${vaultsMap()}) |
-        #  ${concatenate(initVault)}
-        #}
-        #""".stripMargin('#')
-    }
-
-  private def vaultsMap(): String =
-    s"""
-       #{
-       #${concatenate(mapEntry, separator = ",\n")}
-       #}
-       #""".stripMargin('#')
-
-  private def mapEntry(userVault: Vault, index: Int): String =
-    s"""  "${userVault.revAddress.toBase58}" : *x$index"""
-
-  private def initVault(userVault: Vault, index: Int): String =
-    s"""initVault!(*x$index, "${userVault.revAddress.toBase58}", ${userVault.initialBalance})"""
-
-  private def concatenate(f: (Vault, Int) => String, separator: String = " |\n\n"): String =
+object RevGenerator {
+  private def concatenate(
+      userVaults: Seq[Vault]
+  )(f: (Vault, Int) => String, separator: String = " |\n\n"): String =
     if (userVaults.nonEmpty) userVaults.zipWithIndex.map(Function.tupled(f)).mkString(separator)
     else "Nil"
 
+  private def vaultInitCode(userVaults: Seq[Vault]): String = {
+    def initVault(userVault: Vault, index: Int): String =
+      s"""initVault!(*x$index, "${userVault.revAddress.toBase58}", ${userVault.initialBalance})"""
+    def mapEntry(userVault: Vault, index: Int): String =
+      s"""  "${userVault.revAddress.toBase58}" : *x$index"""
+
+    if (userVaults.isEmpty) {
+      "vaultMapStore!({})"
+    } else {
+      val vaultsMap: String =
+        s"""
+           #{
+           #${concatenate(userVaults)(mapEntry, separator = ",\n")}
+           #}
+           #""".stripMargin('#')
+      s"""
+         #new ${userVaults.indices.map("x" + _).mkString(", ")} in {
+         #  vaultMapStore!($vaultsMap) |
+         #  ${concatenate(userVaults)(initVault)}
+         #}
+         #""".stripMargin('#')
+    }
+  }
+
+  def apply(userVaults: Seq[Vault], supply: Long): RevGenerator = {
+    val code: String =
+      s""" new rl(`rho:registry:lookup`), revVaultCh in {
+         #   rl!(`rho:rchain:revVault`, *revVaultCh) |
+         #   for (@(_, RevVault) <- revVaultCh) {
+         #     new ret in {
+         #       @RevVault!("init", *ret) |
+         #       for (vaultMapStore, initVault <- ret) {
+         #         ${vaultInitCode(userVaults)}
+         #       }
+         #     }
+         #   }
+         # }
+     """.stripMargin('#')
+    new RevGenerator(supply, code)
+  }
 }
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -81,11 +81,10 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
     BindPattern(List(toPar(Expr(EVarBody(EVar(Var(FreeVar(0))))))), freeCount = 1)
   private[this] val emptyContinuation = TaggedContinuation()
 
-  def playSystemDeploy[S <: SystemDeploy, Env0](runtime: Runtime[F], systemDeploy: S)(
-      normalizerEnv: NormalizerEnv[Env0]
+  def playSystemDeploy[S <: SystemDeploy](runtime: Runtime[F], systemDeploy: S)(
+      normalizerEnv: NormalizerEnv[systemDeploy.Env]
   )(
-      implicit provided: NormalizerEnv.Provides[Env0, systemDeploy.Env],
-      ev: ToEnvMap[systemDeploy.Env]
+      implicit ev: ToEnvMap[systemDeploy.Env]
   ): F[Either[SystemDeployFailure, SystemDeployResult[systemDeploy.Result]]] = {
     implicit val c: _cost[F] = runtime.cost
     for {
@@ -93,7 +92,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       result <- EitherT
                  .liftF(
                    Interpreter[F]
-                     .evaluate(runtime, systemDeploy.source, systemDeploy.env(normalizerEnv))
+                     .evaluate(runtime, systemDeploy.source, normalizerEnv.toEnv)
                  )
                  .ensureOr(evalResult => UnexpectedErrors(evalResult.errors))(
                    !_.isFailed
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -81,19 +81,26 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
     BindPattern(List(toPar(Expr(EVarBody(EVar(Var(FreeVar(0))))))), freeCount = 1)
   private[this] val emptyContinuation = TaggedContinuation()
 
-  def playSystemDeploy[S <: SystemDeploy](runtime: Runtime[F], systemDeploy: S)(
-      normalizerEnv: NormalizerEnv[systemDeploy.Env]
-  )(
+  def playSystemDeploy[S <: SystemDeploy](
+      runtime: Runtime[F],
+      systemDeploy: S,
+      rand: Blake2b512Random
+  )(normalizerEnv: NormalizerEnv[systemDeploy.Env])(
       implicit ev: ToEnvMap[systemDeploy.Env]
-  ): F[Either[SystemDeployFailure, SystemDeployResult[systemDeploy.Result]]] = {
-    implicit val c: _cost[F] = runtime.cost
+  ): F[Either[SystemDeployFailure, SystemDeployResult[systemDeploy.Result]]] =
     for {
       softCheckpoint <- runtime.space.createSoftCheckpoint()
       result <- EitherT
-                 .liftF(
-                   Interpreter[F]
-                     .evaluate(runtime, systemDeploy.source, normalizerEnv.toEnv)
-                 )
+                 .liftF {
+                   implicit val c: _cost[F] = runtime.cost
+                   Interpreter[F].injAttempt(
+                     runtime.reducer,
+                     runtime.errorLog,
+                     systemDeploy.source,
+                     Cost.UNSAFE_MAX,
+                     normalizerEnv.toEnv
+                   )(rand)
+                 }
                  .ensureOr(evalResult => UnexpectedErrors(evalResult.errors))(
                    !_.isFailed
                  )
@@ -139,7 +146,6 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                  .value
 
     } yield result
-  }
 
   def captureResults(
       start: StateHash,
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -376,16 +376,16 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                         )
                         .ensureOr(
                           /* Regardless of success or failure, verify that deploy status' match. */
-                          result => ReplayFailure.replayStatusMismatch(isFailed, result.isFailed)
-                        )(result => isFailed == result.isFailed)
+                          result => ReplayFailure.replayStatusMismatch(isFailed, result.failed)
+                        )(result => isFailed == result.failed)
                         .ensureOr(
                           result =>
                             /* Since there are no errors, verify evaluation costs and COMM events match. */
                             ReplayFailure.replayCostMismatch(deploy, cost.cost, result.cost.value)
-                        )(result => result.isFailed || cost.cost == result.cost.value)
+                        )(result => result.failed || cost.cost == result.cost.value)
                         .semiflatMap(
                           result =>
-                            if (result.isFailed)
+                            if (result.failed)
                               /* Since the state of `replaySpace` is reset on each invocation of `replayComputeState`,
                                  and `ReplayFailure`s mean that block processing is cancelled upstream, we only need to
                                  reset state if the replay effects of valid deploys need to be discarded. */
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -357,6 +357,7 @@ object RuntimeManager {
       costState: _cost[F],
       errorLog: ErrorLog[F]
   )(deploy: DeployData): F[EvaluateResult] = {
+    import rholang.implicits._
     implicit val rand: Blake2b512Random = Blake2b512Random(
       ProtoUtil.stripDeployData(deploy).toProto.toByteArray
     )
@@ -366,7 +367,7 @@ object RuntimeManager {
       errorLog,
       deploy.term,
       Cost(deploy.phloLimit),
-      NormalizerEnv(deploy.toProto)
+      NormalizerEnv(deploy).toEnv
     )
   }
 }
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -127,7 +127,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
     withResetRuntimeLock(stateHash) { runtime =>
       runtime.space.createSoftCheckpoint() >>= { preDeploySoftCheckpoint =>
         evaluateSystemSource(runtime)(systemDeploy, replay = false) >>= {
-          case EvaluateResult(cost, errors) =>
+          case EvaluateResult(_, errors) =>
             if (errors.nonEmpty) UnexpectedSystemErrors(errors).raiseError
             else
               consumeResult(runtime)(systemDeploy, replay = false) >>= {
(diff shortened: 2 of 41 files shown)


casper/src/test/scala/coop/rchain/casper/util/rholang/Interactive.scala
```diff
@@ -54,7 +50,7 @@ class Interactive private (runtime: Runtime[Task])(implicit scheduler: Scheduler
   def checkpointNames: List[String] = checkpoints.keys.toList
 
   def eval(code: String): Unit = {
-    TestUtil.eval(code, runtime, NormalizerEnv.Empty).runSyncUnsafe(Duration.Inf)
+    TestUtil.eval(code, runtime, Map.empty).runSyncUnsafe(Duration.Inf)
```
tymm, line 53: Why do you prefer `Map.empty` over `NormalizerEnv.Empty`?

models/src/main/scala/coop/rchain/models/HashM.scala
```diff
@@ -88,14 +86,15 @@ object HashM extends HashMDerivation {
...
 
-  implicit val ParHash: HashM[Par] = gen[Par]
-  implicit val ExprHash            = gen[Expr]
-  implicit val VarHash             = gen[Var]
-  implicit val SendHash            = gen[Send]
-  implicit val ReceiveHash         = gen[Receive]
-  implicit val ReceiveBindHash     = gen[ReceiveBind]
-  implicit val NewHash             = gen[New]
-  implicit val MatchHash           = gen[Match]
+  implicit val Env: HashM[Map[String, Par]] = opaqueHash
```
tymm, line 89: Please see this discussion: https://github.com/rchain/rchain/pull/2763#discussion_r335401269
If possible we should stick to `HashM[NormalizerEnv]`.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -1,63 +1,58 @@
...
-import coop.rchain.models.{NormalizerEnv, Par}
+import coop.rchain.models.NormalizerEnv
 import coop.rchain.rholang.build.CompiledRholangSource
-import coop.rchain.rholang.interpreter.ParBuilder
-import monix.eval.Coeval
-
-final case class RevGenerator(userVaults: Seq[Vault], supply: Long) extends CompiledRholangSource {
 
+final class RevGenerator private (supply: Long, code: String)
+    extends CompiledRholangSource(code, NormalizerEnv.Empty) {
```
tymm, line 7: This and the changes related to this should be in a separate commit.

marcin-rzeznicki: Definitely :+1:

casper/src/test/scala/coop/rchain/casper/util/rholang/Interactive.scala
marcin-rzeznicki: I'm not sure if my reasoning holds water, please evaluate: Since `NormallizerEnv` is now type-safe - meaning it tracks the environment it contains, and, some pieces of code can't make use of it - I decided to split it. When you have to work with "just a map" you don't need to depend on `NormalizerEnv`. Only when you can take advantage of a precise `Env` type(like SD do), you should depend on `NornalizerEnv`. OTOH, it is possible to pass `NornalizerEnv.Empty` etc - but then you just type more (type var declaration, additional conversion to `.toEnv`) for no gain

models/src/main/scala/coop/rchain/models/HashM.scala
marcin-rzeznicki: I read that - but I don't understand still. Please explain - `NornalizerEnv` was just a type alias before

tymm: It's just a little bit safer in terms of stack-safety because if someone one day creates a `map: Map[String, Par]` with some really big `Par` inside, Scala will pick up the `opaqueHash` instance and will blow up the stack.
With `HashM[NormalizerEnv]` that someone won't be able to do that. At least not without writing the non stack-safe `HashM` instance himself.

casper/src/test/scala/coop/rchain/casper/util/rholang/Interactive.scala
tymm: I guess that makes sense. Thank you for the explanation.

models/src/main/scala/coop/rchain/models/HashM.scala
marcin-rzeznicki: Agreed

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -73,6 +77,70 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+  )(
+      implicit provided: NormalizerEnv.Provides[Env0, systemDeploy.Env],
+      ev: ToEnvMap[systemDeploy.Env]
+  ): F[Either[SystemDeployFailure, SystemDeployResult[systemDeploy.Result]]] = {
+    implicit val c: _cost[F] = runtime.cost
+    for {
+      softCheckpoint <- runtime.space.createSoftCheckpoint()
+      result <- EitherT
+                 .liftF(
+                   Interpreter[F]
```
JosephDenman: This needs to be a `Interpreter[F].injAttempt` call, and we should pull `c: _cost[F]` into the `liftF` scope. That's also where we'll have to provide the `rand` implicit.

marcin-rzeznicki: `into the liftF scope` ?

JosephDenman: The minimal scope. 

```
EitherT.liftF {
   implicit c: _cost[F] = runtime.cost
   Interpreter[F].injAttempt(...)
}
```

marcin-rzeznicki: Fixed in 7ca418a

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
```diff
@@ -23,6 +26,11 @@ abstract class SystemDeploy(val rand: Blake2b512Random) {
   protected def toEnvMap: ToEnvMap[Env]
   final def env: Map[String, Par] = normalizerEnv.toEnv(toEnvMap)
 
+  implicit protected val envsReturnChannel: Contains.Aux[Env, `sys:casper:return`, GUnforgeable]
+  protected def mkReturnChannel =
```
JosephDenman, line 31: It's interesting that it is critical that this is a definition.

marcin-rzeznicki: You might get NPE otherwise (though I didn't check). When a class is initialized on JVm it's superclasses are initialized first. Had this been `val`, `envsReturnChannel` would still be `null`. Boom.

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
```diff
@@ -31,14 +29,12 @@ final class PreChargeDeploy(chargeAmount: Long, pk: PublicKey, rand: Blake2b512R
       (`sys:casper:return` ->> GUnforgeable) :: HNil
 
   import toPar._
-  protected override val toEnvMap = ToEnvMap[Env]
+  protected override val envsReturnChannel = Contains[Env, `sys:casper:return`]
```
JosephDenman, line 28: Can we de-duplicate `envsReturnChannel` from `PreChargeDeploy` and `RefundDeploy` by defining it in `SystemDeploy` using the `Env` existential?

marcin-rzeznicki: Unfortunately, I don't think it would work. In `SystemDeploy` `Env` is, as you noted, existential which means that it's structure is not known. When you take a look at the implicit expansion at, say, `PreChargeDeploy` you will notice that the expression `Contains[Env, returnChannel]` expands to `NormalizerEnv.Contains.Aux[(sys:casper:deployerId ->> GDeployerId) :: (sys:casper:invalidBlockHash ->> Expr.ExprInstance.GString) :: (sys:casper:return ->> GUnforgeable) :: HNil,sys:casper:return,ev.KeyType]` - this is precisely the info we're missing in `SystemDeploy` . I don't see how it would work :disappointed:

marcin-rzeznicki: PS. I guess my point is that even though the expression looks the same, it actually expands to different things (using knowledge present at the point of expansion).

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
```diff
@@ -0,0 +1,43 @@
...
+  import NormalizerEnv._
+
+  type Output
+  type Result
+  type Env
+
+  final val `sys:casper:initialDeployerId` = Witness("sys:casper:initialDeployerId")
+  final val `sys:casper:return`            = Witness("sys:casper:return")
+
+  type `sys:casper:initialDeployerId` = `sys:casper:initialDeployerId`.T
```
tymm: What exactly is this for? I assume it's the user's deployer id? I don't think every SD will need this.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
```diff
@@ -0,0 +1,43 @@
...
+  type `sys:casper:return`            = `sys:casper:return`.T
+
+  def source: String
```
tymm: I thought the idea was to use a `templateId` and not have code in the SDs?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/package.scala
```diff
@@ -0,0 +1,7 @@
+package coop.rchain.casper.util.rholang
+
+import shapeless.labelled.FieldType
+
+package object costacc {
+  type ->>[A, B] = FieldType[A, B]
```
tymm, line 6: This is something different than the `->>` from `shapeless.syntax.singleton`?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
```diff
@@ -0,0 +1,61 @@
...
+  import syntax.singleton._
+  import record._
+
+  type Output = (RhoBoolean, RhoString)
+  type Result = Unit
+
+  val `sys:casper:chargeAmount` = Witness("sys:casper:chargeAmount")
+  type `sys:casper:chargeAmount` = `sys:casper:chargeAmount`.T
+
+  type Env =
```
tymm, line 24: Seeing it in action I'm no longer sure that all this additional shapeless magic is worth the benefit 😬

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
```diff
@@ -0,0 +1,61 @@
...
+        ("sys:casper:chargeAmount" ->> GInt(chargeAmount)) ::
+        ("sys:casper:return" ->> GUnforgeable(
+          GPrivateBody(GPrivate(ByteString.copyFrom(rand.next())))
+        )) :: HNil
+    )
+
+  override val source: String =
+    """
+                                |new rl(`rho:registry:lookup`),
+                                |poSCh,
```
tymm: Please align like that:
```
new rl(`rho:registry:lookup`),
    poSCh,
    initialDeployerId
    ...
in {
```
and use `#` instead of `|`.

tymm (commented):
First couple of comments. More will follow.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
marcin-rzeznicki: Hmm, this is outdated. I am not sure why this thing shows up. "Fixed" in 4cb5d212502306738871a89fce590dec3c913e17

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
marcin-rzeznicki: Well, the `templateId` is the type (of SystemDeploy) and `code` is a "mapping" between SD and corresponding code. Is this interpretation OK?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/package.scala
marcin-rzeznicki: I promoted shapeless' `->>` to a type for easier writing of `Env` types. So, just as value level `->>` creates a `FieldType` with appropriate singleton type - this type constructor create a type that value level `->>` would create. It is used just for easy writing of SD specs

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
marcin-rzeznicki: Yeah, maybe - but:
1. we castrated this part a bit (we had more fun stuff like checking at compile type if you can combine environments etc)
1. CA correctness is important so it's nice to know that your "strings" match
1. I'd love to expand it virally through the system - I was actually astonished that string-based programming prevails here :-)
1. we'll have more use cases later (eg. asserting at compile time you're using only `sys` namespace etc)
1. people who'll come after us will either cry and make their SDs literally verified by the compiler or tear it down - but no sloppiness will creep out here :-)
1. In Scala 2.13 this is first-class so you can probably throw away shapeless

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
```diff
@@ -35,27 +34,25 @@ object PreChargeDeploy extends SystemDeploy {
...
-                                |chargeAmount(`sys:casper:chargeAmount`),
-                                |return(`sys:casper:return`) in
-                                |{
-                                |  rl!(`rho:rchain:pos`, *poSCh) |
-                                |  for(@(_, PoS) <- poSCh) {
-                                |    @PoS!("chargeDeploy", *initialDeployerId, *chargeAmount, *return)
-                                |  }
-                                |
-                                |}""".stripMargin
+    """|new rl(`rho:registry:lookup`),
```
tymm: The same from my previous comment about alignment applies here as well.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
tymm: Oh, makes sense. What we end up sending over the wire are `ProcessedSystemDeploy`, right? And they won't contain the `source`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -74,6 +86,171 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+            consumeResult(runtime)(systemDeploy, replay = false) >>= {
+              case Some((_, Seq(ListParWithRandom(Seq(par), _)))) =>
+                systemDeploy.extractResult(par) match {
+                  case Right(result) =>
+                    runtime.space
+                      .createSoftCheckpoint()
+                      .map(
+                        postDeploySoftCheckpoint => (postDeploySoftCheckpoint.log, result.asRight)
+                      )
+                  case Left(systemDeployError: SystemDeployError) =>
```
tymm, line 237: An example for this case would be a `chargeDeploy` SD that failed because of insufficient funds, correct?
Why can we revert to `preDeploySoftCheckpoint`? Couldn't this cause problems with replayability?

marcin-rzeznicki: Yes  like `insufficient funds`. Hmm, we're facing some errors in replays. Maybe your comment is spot on. @JosDenmark could you please take a look?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
marcin-rzeznicki: Fixed in 31424de3

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/PreChargeDeploy.scala
marcin-rzeznicki: Also, fixed in 31424de3

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/RefundDeploy.scala
```diff
@@ -0,0 +1,54 @@
...
+  def createNormalizerEnv(refundAmount: Long, rand: Blake2b512Random) =
+    new NormalizerEnv(
+      ("sys:casper:refundAmount" ->> GInt(refundAmount)) ::
+        ("sys:casper:return" ->> GUnforgeable(
+          GPrivateBody(GPrivate(ByteString.copyFrom(rand.next())))
+        )) :: HNil
+    )
+
+  override val source: String =
+    """|new rl(`rho:registry:lookup`),
```
tymm: Please see previous comment about alignment.

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -284,6 +284,9 @@ object implicits {
         None
       }
 
+    def isNil(): Boolean =
+      p.sends.isEmpty && p.receives.isEmpty && p.news.isEmpty && p.matches.isEmpty && p.bundles.isEmpty && p.exprs.isEmpty
```
tymm: Shouldn't we add `&& p.unforgeables.isEmpty && p.connectives.isEmpty`?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -31,10 +31,17 @@ import coop.rchain.rholang.interpreter.{
...
-import coop.rchain.rspace.{Blake2b256Hash, ReplayException, Result}
+import coop.rchain.rspace.{Blake2b256Hash, ReplayException}
 import coop.rchain.shared.Log
 
 trait RuntimeManager[F[_]] {
+  def playSystemDeploy[S <: SystemDeploy](startHash: StateHash)(
+      systemDeploy: S
+  ): F[SystemDeployResult[systemDeploy.Result]]
+  def replaySystemDeploy[S <: SystemDeploy](startHash: StateHash)(
+      systemDeploy: S,
```
tymm, line 43: Just want to be sure that I get it right: every node has to generate system deploys from user deploys and then checks these against `ProcessedSystemDeploy`s, correct?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/SlashDeploy.scala
```diff
@@ -0,0 +1,61 @@
...
+    new NormalizerEnv(
+      ("sys:casper:deployerId" ->> GDeployerId(ByteString.copyFrom(pk.bytes))) :: ("sys:casper:invalidBlockHash" ->> GString(
+        invalidBlockHash.base16String
+      )) :: mkReturnChannel :: HNil
+    )
+
+  override val source: String =
+    """|new rl(`rho:registry:lookup`),
+       |poSCh,
+       |deployerId(`sys:casper:deployerId`),
```
tymm: Before I had seen `initialDeployerId`. Could we make it `deployerId` in all SDs?

tymm: Or maybe here it would even make sense to name it `validatorDeployerId` for more clarity.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -201,11 +282,60 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
     } yield ()
   }
 
+  "computeState" should "be replayed by replayComputeState" in effectTest {
+    runtimeManagerResource.use { runtimeManager =>
+      for {
+        deploy <- ConstructDeploy.sourceDeployNowF(
+                   """
+                                                            # new deployerId(`rho:rchain:deployerId`),
```
tymm, line 290: Alignment seems a bit off here.

tymm (approved):
Great work! Thanks!

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/SlashDeploy.scala
tymm: And in the others `userDeployerId` would maybe make sense. But up to you guys.

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/RefundDeploy.scala
marcin-rzeznicki: Yep, fixed in 31424de361c2e2ba1351182e102a76e2dbcac1b4

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
marcin-rzeznicki: Correct!

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/SlashDeploy.scala
marcin-rzeznicki: Yeah, as I commented: you'd seen the outdated version where we'd had the names mixed up. Now it's `deployerId` everywhere. Makes sense?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
marcin-rzeznicki: While waiting for your assessment I removed these doubtful reverts in 0b62efd50073afc1c958d6a5642b3042a7cc92be

models/src/main/scala/coop/rchain/models/rholang/implicits.scala
marcin-rzeznicki: Ah, correct, thanks!

marcin-rzeznicki: Fixed in b4829c4c8f6b0c8cc2a55d0a7dc2735aff2e97b1

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -74,6 +86,171 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+    runtime.replaySpace.rig(processedSystemDeploy.eventList.map(EventConverter.toRspaceEvent)) >> {
+      runtime.replaySpace.createSoftCheckpoint() >>= { softCheckpoint =>
+        evaluateSystemSource(runtime)(systemDeploy, replay = true) >>= {
+          case EvaluateResult(_, interpreterErrors) =>
+            if (interpreterErrors.nonEmpty) UnexpectedSystemErrors(interpreterErrors).raiseError
+            else
+              consumeResult(runtime)(systemDeploy, replay = true) >>= { consumeResultOpt =>
+                runtime.replaySpace.checkReplayData().attempt.map(_.swap.toOption) >>= {
+                  case Some(replayException: ReplayException) =>
+                    ReplayFailure.unusedCOMMEvent(replayException).asLeft.pure
```
zsluedem, line 173: Do we need `revertToSoftCheckpoint ` here?

tymm: I think we should add some comments to the code about when and why we revert or not. It's hard to understand from code.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosephDenman: Timm is right. In a forthcoming PR we fix these issues, reverting only when an user error (`InterpreterError`) is thrown during `replayDeploy` or `processDeploy`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
JosephDenman: @zsluedem https://github.com/rchain/rchain/pull/2778#discussion_r341574398

JosephDenman (approved):

PR #2779, opened 2019-10-21
RCHAIN-3895: Make `NormalizerEnv` more type safe

tymm:
## Overview
See discussion here: https://github.com/rchain/rchain/pull/2763#discussion_r336118087

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
@@ -957,12 +957,17 @@ object ProcNormalizeMatcher {
         val newCount           = newEnv.count - input.env.count
         val requiresDeployId   = uris.contains(deployIdUri)
         val requiresDeployerId = uris.contains(deployerIdUri)
+        val containsDeployId =
+          env.get(deployIdUri).exists(_.injInstance.unfBody.exists(_.unfInstance.isGDeployIdBody))
+        val containsDeployerId = env
+          .get(deployerIdUri)
+          .exists(_.injInstance.unfBody.exists(_.unfInstance.isGDeployerIdBody))
 
         def missingEnvElement(name: String, uri: String) =
           NormalizerError(s"`$uri` was used in rholang usage context where $name is not available.")
-        if (requiresDeployId && env.get(deployIdUri).forall(_.singleDeployId().isEmpty))
+        if (requiresDeployId && !containsDeployId)
           missingEnvElement("DeployId", deployIdUri).raiseError[M, ProcVisitOutputs]
-        else if (requiresDeployerId && env.get(deployerIdUri).forall(_.singleDeployerId().isEmpty))
+        else if (requiresDeployerId && !containsDeployerId)
           missingEnvElement("DeployerId", deployerIdUri).raiseError[M, ProcVisitOutputs]
         else {
           normalizeMatch[M](p.proc_, ProcVisitInputs(VectorPar(), newEnv, input.knownFree)).map {
--- a/models/src/main/protobuf/RhoTypes.proto
+++ b/models/src/main/protobuf/RhoTypes.proto
@@ -153,8 +153,15 @@ message New {
     Par p = 2 [(scalapb.field).no_box = true];
     // For normalization, uri-referenced variables come at the end, and in lexicographical order.
     repeated string uri = 3;
-    map<string, Par> injections = 4;
+    map<string, Injection> injections = 4;
     bytes locallyFree = 5 [(scalapb.field).type = "coop.rchain.models.AlwaysEqual[scala.collection.immutable.BitSet]"];
+}
+
+message Injection {
+    oneof inj_instance {
+        Expr expr_body = 1;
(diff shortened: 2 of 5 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/normalize.scala
```diff
@@ -957,12 +957,17 @@ object ProcNormalizeMatcher {
...
+          .get(deployerIdUri)
+          .exists(_.injInstance.unfBody.exists(_.unfInstance.isGDeployerIdBody))
 
         def missingEnvElement(name: String, uri: String) =
           NormalizerError(s"`$uri` was used in rholang usage context where $name is not available.")
-        if (requiresDeployId && env.get(deployIdUri).forall(_.singleDeployId().isEmpty))
+        if (requiresDeployId && !containsDeployId)
           missingEnvElement("DeployId", deployIdUri).raiseError[M, ProcVisitOutputs]
-        else if (requiresDeployerId && env.get(deployerIdUri).forall(_.singleDeployerId().isEmpty))
+        else if (requiresDeployerId && !containsDeployerId)
```
ArturGajowy, line 970: I would consider handling all missing injections with the same logic, without differentiating for deploy[er]Id :)

tymm: I'm not sure how this would work without differentiation since you have to know what you are checking for.

marcin-rzeznicki (commented):
Can `Injection` be parametrized over the type of expression it holds? It would be helpful given my changes to `NormalizerEnv[Env]`.

tymm:
@marcin-rzeznicki I don't think Protobuf supports that unfortunately.

marcin-rzeznicki:
> @marcin-rzeznicki I don't think Protobuf supports that unfortunately.

Ah right - Protobuf - weren't we supposed to introduce a model between protobuf and the rest of the system? I guess it's been half-done, as usual. OK, I won't be blocking the PR

tymm:
@marcin-rzeznicki Hmm, I haven't heard of such a thing?

tymm:
@marcin-rzeznicki I haven't heard of anything like that. Maybe I missed that conversation.

marcin-rzeznicki:
Ok, nvm then, maybe I'm mistaken. Approved

marcin-rzeznicki (approved):

JosephDenman:
Is consensus that `Par` is too general to be included as a binding in the normalizer environment? If so, why? In the case that we have to add bindings for non-unforgeable and non-expressions, we'd have to expand `Injection` to wrap more types. In the limit, it looks like `Injection` wraps a subset of-Rholang values that approaches the set of values wrapped in Par.

tymm:
@JosDenmark I thought there was consensus on this. See https://github.com/rchain/rchain/pull/2763#pullrequestreview-303403867 and https://github.com/rchain/rchain/pull/2763#discussion_r333846923.
I have a hard time thinking about cases where something else than `Expr` or `GUnforgeable` makes sense.

JosephDenman:
@tymm I'm recanting my positive vote. Just using `Expr`s as bindings is insufficient to prevent runtime errors, since I could pass a `BoundVar(?)` into the normalizer environment, the evaluation of which would result in a runtime error. More generally, my argument is: `Par`s are values, we want to bind values, so we should bind `Par`s.

tymm:
> Just using Exprs as bindings is insufficient to prevent runtime errors, since I could pass a BoundVar(?) into the normalizer environment, the evaluation of which would result in a runtime error.

Constraining the types is nevertheless a win.

However happy to close this PR if others agree with @JosDenmark and/or this makes @marcin-rzeznicki life more complicated in regards to his recent work on making the `NormalizerEnv` more type-safe.

marcin-rzeznicki:
> However happy to close this PR if others agree with @JosDenmark and/or this makes @marcin-rzeznicki life more complicated in regards to his recent work on making the `NormalizerEnv` more type-safe.

I think my life is unaffected :smile: In general I track the environment as a (rather complicated) type used to prove that `NormalizerEnv` you want to use for a system deploy contains everything (a superset of) the system deploy declared it needs. While interacting with "lower level" parts of code that happened to use a `Map[String, Par]` I carry an additional proof that his conversion is possible and sound. In theory these checks that @ArturGajowy mentioned could've been completely eliminated, had this design been persistent. But given that it isn't the case and things are modeled the way they are (everything is a protobuf-based type, there isn't even a sound hierarchy to compute LUBs you could work with) - I conclude this is the lost cause. So, I am fine with `Map[String, Whatever]` as long as I can convert to it - which I believe I can.

tymm:
If we come up with a stack-safe `HashM` instance for maps, then we could constrain the types just in `NormalizerEnv.scala` which would be more flexible compared to introducing a `Injection` protobuf.
@JosDenmark would you prefer that solution?

JosephDenman:
@tymm My issue isn't really with `Injection`, it's with the type constraining generally. It seems like an arbitrary refinement to me 🤷‍♂.

PR #2781, opened 2019-10-23
RCHAIN-3888: Secure access to protocol-private PoS methods

tymm:
## Overview
See: https://gist.github.com/tymm/f1b89ef2d3308ccaec858b16937e6807 and https://rchain.atlassian.net/browse/RCHAIN-3888

I would appreciate if @marcin-rzeznicki and @JosDenmark could take a closer look on what I did here with replaying system deploys from `BlockMessage`.
Also please have a look on how I use the `invalidBlockHash` as a seed for the rand value for `SlashDeploy`s.

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -36,6 +36,7 @@ new  PoS,
     rs(`rho:registry:insertSigned:secp256k1`),
     posId(`rho:rchain:deployerId`),
     PendingRewards, MergeIntMaps, treeHashMapCh,
+    sysAuthTokenOps(`sys:authToken:ops`),
     uriOut
 in {
   registryLookup!(`rho:rchain:revVault`, *revVaultCh) |
@@ -193,20 +194,29 @@ in {
             } |
 
             new currentDeployerData in {
-              contract PoS (@"chargeDeploy", @deployerId, @amount, return) = {
-                new depositCh in {
-                  for (@state <<- stateCh) {
-                    deposit!(deployerId, amount, state.get("activeValidators"), *depositCh) |
-                    for(@depositResult <- depositCh){
-                      match depositResult {
-                        (true, _) => {
-                          currentDeployerData!((deployerId, amount)) |
-                          return!(depositResult)
-                        }
-                        (_, errorMessage) => {
-                          return!((false, "Deploy payment failed: " ++ errorMessage))
+              contract PoS (@"chargeDeploy", @deployerId, @amount, @sysAuthToken, return) = {
+                new isValidTokenCh in {
+                  sysAuthTokenOps!("check", sysAuthToken, *isValidTokenCh) |
+                  for (@isValid <- isValidTokenCh) {
+                    if (isValid) {
+                      new depositCh in {
+                        for (@state <<- stateCh) {
+                          deposit!(deployerId, amount, state.get("activeValidators"), *depositCh) |
+                          for(@depositResult <- depositCh){
+                            match depositResult {
+                              (true, _) => {
+                                currentDeployerData!((deployerId, amount)) |
+                                return!(depositResult)
+                              }
+                              (_, errorMessage) => {
+                                return!((false, "Deploy payment failed: " ++ errorMessage))
+                              }
+                            }
+                          }
                         }
                       }
+                    } else {
+                      return!((false, "Invalid system auth token"))
                     }
                   }
                 }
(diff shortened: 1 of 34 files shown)


tymm:
This PR is blocked until I can replace the following code with a system deploy.
https://github.com/rchain/rchain/blob/179571510a48fe6d5338f8d98a0e232ab0917192/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala#L65-L77

marcin-rzeznicki:
Hey @tymm I'll finish slashing system deploy today to unblock you, thx

marcin-rzeznicki:
> This PR is blocked until I can replace the following code with a system deploy.
> 
> https://github.com/rchain/rchain/blob/179571510a48fe6d5338f8d98a0e232ab0917192/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala#L65-L77

This is fixed in 9e562ab (once it's merged)

marcin-rzeznicki (approved):
:+1:

tymm:
bors try

tymm:
bors try

woky:
bors try

tymm:
bors try

JosephDenman:
@tymm Would you explain why this is something that a system `DeployerID` can't address? Or send me a link to our previous discussion with Artur?

tymm:
@JosDenmark the discussion can be found in this gist: https://gist.github.com/tymm/f1b89ef2d3308ccaec858b16937e6807

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -166,19 +166,21 @@ object BlockApproverProtocol {
             Long.MaxValue
           )
           .toSet
-        blockDeploys = block.body.deploys
+        blockDeploys       = block.body.deploys
+        blockSystemDeploys = block.body.systemDeploys
```
JosephDenman: If there are no expected system deploys in the genesis block, we should verify that the block contains none.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -166,19 +166,21 @@ object BlockApproverProtocol {
...
     (for {
-      result                    <- EitherT(validate.pure[F])
-      (blockDeploys, postState) = result
+      result                                        <- EitherT(validate.pure[F])
+      (blockDeploys, blockSystemDeploys, postState) = result
       stateHash <- EitherT(
                     runtimeManager
                       .replayComputeState(runtimeManager.emptyStateHash)(
                         blockDeploys,
+                        blockSystemDeploys,
```
JosephDenman: Why add a parameter instead of using the pre-existing `replaySystemDeploy` function?

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/SlashDeploy.scala
```diff
@@ -1,16 +1,21 @@
...
-
-final class SlashDeploy(invalidBlockHash: BlockHash, pk: PublicKey, rand: Blake2b512Random)
-    extends SystemDeploy(rand) {
+import coop.rchain.shared.Serialize
+import scodec.bits.ByteVector
+
+final case class SlashDeploy(
+    invalidBlockHash: BlockHash,
+    pk: PublicKey,
+    override val rand: Blake2b512Random
```
JosephDenman: Isn't the `override` redundant since `rand` is passed to `SystemDeploy`?

JosephDenman: And the `val` modifier? I don't see it accessed anywhere outside of `SystemDeploy`.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
JosephDenman: Like you did for the `computeDeploysCheckpoint` function.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -158,20 +159,36 @@ object InterpreterUtil {
...
                           )
-      preStateHash <- computeParentsPostState(nonEmptyParents, dag, runtimeManager)
-      result       <- runtimeManager.computeState(preStateHash)(deploys, blockData, invalidBlocks)
-    } yield preStateHash +: result
+      preStateHash                  <- computeParentsPostState(nonEmptyParents, dag, runtimeManager)
+      resultDeploys                 <- runtimeManager.computeState(preStateHash)(deploys, blockData, invalidBlocks)
+      (startHash, processedDeploys) = resultDeploys
+      resultSystemDeploys <- {
+        import cats.instances.list._
+        systemDeploys.toList.foldM((startHash, List.empty[ProcessedSystemDeploy])) {
```
JosephDenman: Appending to a list is an anti-pattern. You should use `Vector`.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -158,20 +159,36 @@ object InterpreterUtil {
...
+          case ((startHash, processedSystemDeploys), sd) =>
+            runtimeManager.playSystemDeploy(startHash)(sd) >>= {
+              case PlaySucceeded(stateHash, processedSystemDeploy, _) =>
+                (stateHash, processedSystemDeploys :+ processedSystemDeploy).pure[F]
+              case PlayFailed(_) =>
```
JosephDenman: The error message should be extracted and added to the `Exception` message.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -88,6 +88,7 @@ object InterpreterUtil {
       isGenesis = block.header.parentsHashList.isEmpty
       replayResult <- runtimeManager.replayComputeState(preStateHash)(
                        internalDeploys,
+                       internalSystemDeploys,
```
JosephDenman, line 92: This is an attack vector, since we're blindly running system deploys that a potentially malicious validator has sent us. We need to recalculate the slashing deploys on the replay side and verify that they match the system deploys sent in the block message. Then we replay them.

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
```diff
@@ -356,23 +358,60 @@ object ProcessedDeploy {
   }
 }
 
+final case class SlashSystemDeployData(invalidBlockHash: BlockHash, issuerPublicKey: PublicKey)
+
+final case class SystemDeployData(maybeSlashData: Option[SlashSystemDeployData])
```
JosephDenman: Why not just have `SlashSystemDeployData` extend `SystemDeployData`? Then `playSucceeded` is subtype polymorphic over the `SystemDeployData`. More extensible.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
```diff
@@ -19,7 +19,7 @@ abstract class SystemDeploy(initialRand: Blake2b512Random) {
   type Result
   type Env
 
-  final val rand = initialRand.copy()
+  val rand = initialRand.copy()
```
JosephDenman: Again, I'm unsure why you would want to use a different RNG than the one passed to `SystemDeploy`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -419,7 +439,9 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       runtime: Runtime[F],
       startHash: StateHash,
       terms: Seq[ProcessedDeploy],
-      replayDeploy: ProcessedDeploy => F[Option[ReplayFailure]]
+      systemDeploys: Seq[ProcessedSystemDeploy],
+      replayDeploy: ProcessedDeploy => F[Option[ReplayFailure]],
+      replaySystemDeploy: ProcessedSystemDeploy => F[Option[ReplayFailure]]
```
JosephDenman, line 444: I don't think either of the `replayDeploy` or `replaySystemDeploy` parameters are necessary, since we use the `isGenesis` flag in `replayComputeState`. The first existed before this PR, so whether you want to remove it is up to you. `replaySystemDeploy` should probably be removed though.

JosephDenman (dismissed):
Good work overall. I'm only up to the fourth commit (which I think should've been its own PR), but I want you to respond to some of my comments before I continue.

casper/src/main/resources/PoS.rhox
```diff
@@ -244,242 +254,269 @@ new  PoS,
               // Note: under no circumstances should this method be merged as is.
               //       this method, along with "chargeDeploy", "closeBlock", and "slash"
               //       must be made private, as they're only used in the protocol.
```
tgrospic: Is this comment still valid?

casper/src/main/resources/PoS.rhox
```diff
@@ -191,21 +192,30 @@ new  PoS,
...
+                              }
+                              (_, errorMessage) => {
+                                return!((false, "Deploy payment failed: " ++ errorMessage))
+                              }
+                            }
+                          }
                         }
                       }
+                    } else {
+                      return!(Nil)
```
tgrospic: Shouldn't this return the same Either type e.g. `return!((false, "Invalid auth token"))`? In other methods also.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
tymm: Good call!

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
tymm: Because the pre-existing `replaySystemDeploy(startHash: StateHash)(...)` function does not fit well into `RuntimeManager.replayComputeState` which is unaware of the the start state hashes of each deploy or system deploy.

casper/src/main/scala/coop/rchain/casper/util/rholang/costacc/SlashDeploy.scala
tymm: The `override val` was necessary since `SystemDeploy` has a val called "rand".
But now that you said that, I realized that I don't need `override val` when I use an identifier different from "rand". Thanks.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
tymm: The malicious validator would slash him or herself in that case. This is what ```getInvalidBlocks(`rho:casper:invalidBlocks`)``` in `PoS!slash` ensures.
I didn't know this myself before consulting Kent about it.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
tymm: This was related to what I described here: https://github.com/rchain/rchain/pull/2781#discussion_r365625061

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
tymm: Are you saying the parameters can be removed because `replayComputeState` is only used for genesis?

casper/src/main/resources/PoS.rhox
tymm: Good catch!

casper/src/main/resources/PoS.rhox
tymm: Done.

casper/src/main/resources/PoS.rhox
```diff
@@ -241,245 +251,269 @@ new  PoS,
...
+                                        state.get("committedRewards").set(validator, 0))
+                                   .set("allBonds",
+                                        state.get("allBonds").set(validator, 0)),
+                              true)
+                          }
+                        }
+                      }
                     }
+                  } else {
+                    returnCh!((false, "Invalid system auth token"))
```
tgrospic, line 344: It seems that `slash` is returning just Boolean (line 330).

casper/src/main/resources/PoS.rhox
```diff
@@ -241,245 +251,269 @@ new  PoS,
...
                           }
                         }
                       }
-                    } |
-
-                    contract isQuarantineFinished(@(_, blockNumber), resultCh) = {
-                      resultCh!(blockNumber >= currentBlockNumber)
                     }
+                  } else {
+                    ackCh!((false, "Invalid system auth token"))
```
tgrospic, line 534: `closeBlock` is returning `Nil` for success. Maybe it makes sense to change `slash` and `closeBlock` to return Either as other methods.

tymm:
bors try

tymm:
bors try

tgrospic (approved):

tymm:
> bors r+

@9rb please don't just try to merge PRs without explicit approval from the creator.

tymm:
I think this PR would really need further input from Joe but Rao wants this to be merged.
bors r+

PR #2785, opened 2019-10-26
Rchain 3896 backward compatible

zsluedem:
## Overview
This pr does two things.

1. create a script to generate rnode data.
2. add test backward compatible which uses the history rnode data generated above

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3896
https://rchain.atlassian.net/browse/RCHAIN-3901

### Notes


zsluedem:
bors try

integration-tests/test/test_backward_compatible.py
```diff
@@ -0,0 +1,84 @@
...
+
+        wait_for_approved_block_received_handler_state(context, bootstrap_node)
+        wait_for_approved_block_received_handler_state(context, validator_a)
+        wait_for_approved_block_received_handler_state(context, validator_b)
+        wait_for_peers_count_at_least(context, bootstrap_node, 2)
+        wait_for_peers_count_at_least(context, validator_a, 2)
+        wait_for_peers_count_at_least(context, validator_b, 2)
+
+
+        # why is here 28?
```
nzpr: should be 22 here?

integration-tests/generate_rnode_data.py
```diff
@@ -0,0 +1,159 @@
...
+        VALIDATOR_B_PRIVATE: 2000
+    }
+    logging.info("start the genesis ceremony")
+
+    with generate_cmd_opts() as command_line_options ,\
+            docker_client_context() as  docker_cli, \
+            testing_context(command_line_options, random_seed, docker_cli, bootstrap_key=CEREMONY_MASTER_PRIVATE, peers_keys=peers_keypairs, validator_bonds_dict=bonding_map) as context, \
+            temporary_wallets_file(context.random_generator, wallet_balance_from_private_key=wallet_map) as wallets, \
+            ready_bootstrap(context=context, cli_options=bootstrap_cli_options, wallets_file=wallets) as ceremony_master, \
+            started_peer(context=context, network=ceremony_master.network, bootstrap=ceremony_master, name='validator-a', private_key=VALIDATOR_A_PRIVATE, wallets_file=wallets, cli_flags=peers_cli_flags, cli_options=peers_cli_options, synchrony_constraint_threshold=0.33) as validator_a, \
```
nzpr, line 87: Generated DAG consists of 1 genesis, then 1 each node propose a single block, then 3 REV transfers on ceremony master, then again each node propose a single block. So should be 22 blocks total, if there won't be blocks proposed at the same height. Are you sure this won't be the case? Would it be more reliable to bump sync constraint threshold  to > 2/3?

nzpr (approved):
IMO this tests does more than checking if RNode is able to start using existing data. So it can potentially fail not only when genesis tuplespace validation failed. But more is better than less I guess.

integration-tests/test/test_backward_compatible.py
zsluedem: You are right. I forgot to update the comments.

zsluedem:
> IMO this tests does more than checking if RNode is able to start using existing data. So it can potentially fail not only when genesis tuplespace validation failed. But more is better than less I guess.

Yes. The ultimate purpose of the test is never `genesis tuplespace validation failed`. The ` genesis tuplespace validation failed` is not error in the genesis but a backward-compatible problem.

You are right that. It can fail on other reasons. That's why we should test it.

kayvank (approved):

woky (approved):

PR #2786, opened 2019-10-29
Use block creator for the fairness constraint

sebtomba:
## Overview
Use the block creator instead of an intermediate peer for the fairness constraint.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3904

### DIFF
--- a/models/src/main/protobuf/CasperMessage.proto
+++ b/models/src/main/protobuf/CasperMessage.proto
@@ -100,7 +100,8 @@ message BlockMessageProto {
 
 message BlockHashMessageProto {
   option (scalapb.message).extends = "CasperMessageProto";
-  bytes  hash       = 1;
+  bytes  hash         = 1;
+  bytes  blockCreator = 2;
 }
 
 message BlockMetadataInternal {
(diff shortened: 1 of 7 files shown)


sebtomba:
bors try

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -100,7 +100,8 @@ message BlockMessageProto {
 
 message BlockHashMessageProto {
   option (scalapb.message).extends = "CasperMessageProto";
-  bytes  hash       = 1;
+  bytes  hash         = 1;
+  bytes  blockCreator = 2;
```
zsluedem, line 104: Is `blockCreator` the same as `sender`?
Maybe we should use the same name or add a comment here to let others know this is the same as sender.

sebtomba: @zsluedem IMO we should start fixing the misleading field name `sender`. I'll add a comment

zsluedem (approved):

itegulov (approved):

PR #2788, opened 2019-10-31
Add generated tests

ghost:
## Overview
Adds a `fromLong` function that generates a class of rholang programs and a fuzzer test that those programs give the same cost when replayed. 

### JIRA ticket
N/A

### Notes
This test uncovered a bug in the replay logic.  Several programs have triggered replay errors.  One of them that does so is 
`fromLong(1716417707L) = @2!!(0) | @0!!(0) | for (_ <<- @2) { 0 } | @2!(0)`.
Therefore, tests do not pass yet.

The function `fromLong` returns the empty string if it produces both a persistent send and a persistent receive, even if they're on different channels.  That's something that can be improved later.  Also, my original intent was to iterate over all programs, so I kept the total number of programs fairly small, 0x144000000 ≈ 5 billion.  I then decided to randomly select from the set instead, so if the reviewers want me to, I can expand the set of programs I'm checking.


ghost:
bors try

marcin-rzeznicki (approved):
Approved: a comment that search space of the fuzzer does not have holes (which @pyrocto brilliantly assured) would be nice (also, maybe recursive version would be easier to track?). But all in all - great work

JosephDenman:
This is actually my fault, but the test `"be repeatable when generated"` never actually executes because it's never run. You'll want to add `.runSyncUnsafe(75.seconds)` at the end of `evaluateAndReplay` in the same spirit as `evaluateWithCostLog`. That'll actually run the tests. You'll also have to change the return type from `Task[(EvaluateResult, EvaluateResult)] ` to `(EvaluateResult, EvaluateResult)`.

ghost:
bors try

ghost:
Well, it appears that the tests worked and discovered an actual bug.  The test fails whenever there's a join on the same name and enough sends to trigger it.  For example:
```
fromLong(510661906) = @1!(0) | @1!(0) | for (_ <= @1; _ <= @1) { 0 }
fromLong(5302000922) = @1!!(0) | @1!(0) | for (_ <<- @1; _ <<- @1) { 0 }
fromLong(1189080727) = for (_ <- @3) { 0 } | @2!(0) | @3!!(0) | for (_ <<- @3; _ <<- @3) { 0 }
```

JosephDenman:
I'm not really sure what to do with this PR, since it introduces non-deterministically failing tests. I don't think it'd be a problem to keep the draft around until we can fix the bugs since Rholang isn't being actively developed. @pyrocto @marcin-rzeznicki What do you think?

JosephDenman (approved):

ghost:
I filed bug 3917 and added an `ignore` to the particular test.  Whoever fixes 3917 can remove the `ignore`.

PR #2789, opened 2019-11-03
RCHAIN-3866: Add support for Ethereum personal signature format

tgrospic:
## Overview
Allowing for deploys to signed this way would mean we could leverage all the wallet tools that have been built for Ethereum (such as Metamask) to sign RChain deploys.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3866
https://rchain.atlassian.net/browse/RCHAIN-3867

### Notes
The same parameter is used in deploy `sigAlgorithm` but with additional label `eth` to mark additional verification support. `secp256k1:eth` means the same signature algorithm as with `secp256k1` but message is constructed with prefix and different hashing algorithm.

### DIFF
--- a/crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
+++ b/crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
@@ -8,7 +8,6 @@ import java.security.interfaces.{ECPrivateKey, ECPublicKey}
 import java.security.spec._
 import java.util.Base64
 
-import cats.syntax.all._
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Keccak256
 import org.bouncycastle.asn1.{
@@ -21,7 +20,7 @@ import org.bouncycastle.asn1.{
 import org.bouncycastle.util.BigIntegers
 
 import scala.io.Source
-import scala.util.Try
+import scala.util.{Failure, Try}
 
 object CertificateHelper {
 
@@ -128,7 +127,7 @@ object CertificateHelper {
     cert2
   }
 
-  def encodeSignatureRStoDER(signatureRS: Array[Byte]): Either[Throwable, Array[Byte]] = {
+  def encodeSignatureRStoDER(signatureRS: Array[Byte]): Try[Array[Byte]] = {
     def toASN1Int(bytes: Array[Byte]) = new ASN1Integer(BigIntegers.fromUnsignedByteArray(bytes))
 
     def convert = {
@@ -149,12 +148,12 @@ object CertificateHelper {
     }
 
     if (signatureRS.isEmpty)
-      new IllegalArgumentException("Input array must not be empty").asLeft
+      Failure(new IllegalArgumentException("Input array must not be empty"))
     else
-      Try(convert).toEither
+      Try(convert)
   }
 
-  def decodeSignatureDERtoRS(signatureDER: Array[Byte]): Either[Throwable, Array[Byte]] = {
+  def decodeSignatureDERtoRS(signatureDER: Array[Byte]): Try[Array[Byte]] = {
     def toBytes(x: ASN1Encodable) = {
       val asn1 = x.toASN1Primitive.asInstanceOf[ASN1Integer]
       BigIntegers.asUnsignedByteArray(asn1.getValue)
@@ -168,18 +167,22 @@ object CertificateHelper {
         val Array(r, s, _*) = asnSeq.toArray
         toBytes(r) ++ toBytes(s)
       } finally {
-        asn.close
+        // Ensure `close` does not throw an exception
+        Try(asn.close).getOrElse(())
         // > Closing a ByteArrayInputStream has no effect.
         // https://docs.oracle.com/javase/10/docs/api/java/io/ByteArrayInputStream.html
         bis.close
       }
     }
 
+    import coop.rchain.shared.TrySyntax._
+
     if (signatureDER.isEmpty)
-      new IllegalArgumentException("Input array must not be empty").asLeft
+      Failure(new IllegalArgumentException("Input array must not be empty"))
     else
-      Try(convert).toEither
-        .leftMap(new IllegalArgumentException("Input array is not valid DER message format", _))
+      Try(convert).mapFailure(
+        new IllegalArgumentException("Input array is not valid DER message format", _)
+      )
   }
 
 }
--- a/crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
+++ b/crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
@@ -8,6 +8,7 @@ import java.security.interfaces.{ECPrivateKey, ECPublicKey}
 import java.security.spec._
 import java.util.Base64
 
+import cats.syntax.all._
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.hash.Keccak256
 import org.bouncycastle.asn1.{
@@ -20,6 +21,7 @@ import org.bouncycastle.asn1.{
 import org.bouncycastle.util.BigIntegers
 
 import scala.io.Source
+import scala.util.Try
 
 object CertificateHelper {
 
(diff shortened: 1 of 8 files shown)


tgrospic:
With changes from #2794 and introduction of `Signed` type this PR must be refactored.

tgrospic:
Ping @marcin-rzeznicki, @tymm, @JosDenmark.

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +118,49 @@ object CertificateHelper {
...
+
+    val (r, s) = signatureRS.take(64).splitAt(32)
+    val bos    = new ByteArrayOutputStream(72)
+    val seq    = new DERSequenceGenerator(bos)
+    try {
+      seq.addObject(toASN1Int(r))
+      seq.addObject(toASN1Int(s))
+      seq.close
+      bos.toByteArray
+    } catch { case _: Throwable => Array[Byte]() } finally {
```
marcin-rzeznicki: More nuanced exception handling please!

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +118,49 @@ object CertificateHelper {
...
+      if (seq != null) seq.close()
```
marcin-rzeznicki: This can throw an exception I think, which means that `bos` will never get closed

marcin-rzeznicki (dismissed):

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
tgrospic: It seems that closing `bos` is not necessary, but exception to close `seq` must be handled.
> Closing a ByteArrayOutputStream has no effect. The methods in this class can be called after the stream has been closed without generating an IOException.
https://docs.oracle.com/javase/7/docs/api/java/io/ByteArrayOutputStream.html

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
tgrospic: @marcin-rzeznicki I'm not sure what you mean. Is there a way to use `using`?

ghost: Since it's possible for this to throw an exception, use an error type (option or either is fine) instead of a zero-length byte array.  Ditto for the method below.

tgrospic: I've added error handling to these functions.

And also found that some functions on SignatureAlg implementation can throw exceptions. I've created a new ticket for that. E.g. if DeployData#sig contains array greater than 520 bytes.
https://rchain.atlassian.net/browse/RCHAIN-3959

9rb (approved):
lgtm

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +120,65 @@ object CertificateHelper {
...
+      val (r, s) = signatureRS.take(64).splitAt(32)
+      val bos    = new ByteArrayOutputStream(72)
+      val seq    = new DERSequenceGenerator(bos)
+      try {
+        seq.addObject(toASN1Int(r))
+        seq.addObject(toASN1Int(s))
+        seq.close
+        bos.toByteArray
+      } finally {
+        if (seq != null) seq.close()
```
marcin-rzeznicki: So is `seq` now closed twice? It doesn't seem right

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +120,65 @@ object CertificateHelper {
...
+        // > Closing a ByteArrayOutputStream has no effect.
+        // https://docs.oracle.com/javase/10/docs/api/java/io/ByteArrayOutputStream.html
+      }
+    }
+
+    if (signatureRS.isEmpty)
+      new Exception("Input array must not be empty.").asLeft
```
marcin-rzeznicki: `Exception` is rather too-generic. You could use `sys.error` (RuntimeError) or `IllegalArgumentException` etc

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +120,65 @@ object CertificateHelper {
...
+    else
+      Try(convert).toEither
```
marcin-rzeznicki: `Try[A]` is isomorphic to `Either[Throwable, A]` (well, there are edge cases but for practical purposes, it is). I am not sure why we cannot just stick to `Try[A]` then

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +120,65 @@ object CertificateHelper {
...
+      val bis = new ByteArrayInputStream(signatureDER)
+      val asn = new ASN1InputStream(bis)
+      try {
+        val asnSeq          = asn.readObject.asInstanceOf[ASN1Sequence]
+        val Array(r, s, _*) = asnSeq.toArray
+        toBytes(r) ++ toBytes(s)
+      } finally {
+        if (asn != null) asn.close()
+        // > Closing a ByteArrayInputStream has no effect.
+        // https://docs.oracle.com/javase/10/docs/api/java/io/ByteArrayInputStream.html
```
marcin-rzeznicki, line 173: This is true, but, being paranoid, what if someone changes the type of `InputStream` to something that needs to be closed?

crypto/src/test/scala/coop/rchain/crypto/util/DERConverterSpec.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.scalacheck.Prop.propBoolean
+
+class DERConverterSpec extends PropSpec with GeneratorDrivenPropertyChecks with Matchers {
+  implicit val arbBytes: Arbitrary[Array[Byte]] = arbContainer[Array, Byte]
+
+  property("DER converter should work with any non-empty byte array") {
+    forAll { (bytes: Array[Byte]) =>
+      !bytes.isEmpty ==> !CertificateHelper.encodeSignatureRStoDER(bytes).right.get.isEmpty
+
+      !bytes.isEmpty ==> !CertificateHelper.decodeSignatureDERtoRS(bytes).right.get.isEmpty
```
marcin-rzeznicki: Should we check that encode/decode roundtrip gives us the same value?

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
tgrospic: @marcin-rzeznicki is there an implicit converter between these two types? Without explicit conversion `toEither` compiler complains for the wrong type.

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
```diff
@@ -110,14 +120,68 @@ object CertificateHelper {
...
+
+    def convert = {
+      val (r, s) = signatureRS.take(64).splitAt(32)
+      val bos    = new ByteArrayOutputStream(72)
+      val seq    = new DERSequenceGenerator(bos)
+      try {
+        seq.addObject(toASN1Int(r))
+        seq.addObject(toASN1Int(s))
+        // Close to write the sequence
+        seq.close
```
tgrospic, line 141: @marcin-rzeznicki it seems that closing a sequence is only for writing the data and not for release any resource so it's not necessary in the finally block.

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
tgrospic: @marcin-rzeznicki maybe also to wrap `Try(asn.close)`?

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
marcin-rzeznicki: No, they're isomorphic by "nature". You could switch to `Try` simply - I think it's less hassle than maintaining `Either`. I don't think this is critical though

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
marcin-rzeznicki: Makes sense, :+1:

crypto/src/main/scala/coop/rchain/crypto/util/CertificateHelper.scala
tgrospic: You mean to change return type to `def encodeSignatureRStoDER(signatureRS: Array[Byte]): Try[Array[Byte]]`?

tgrospic: I like the `Try` (except mapping failure :smiley:).

tgrospic:
bors try

nzpr (approved):
As all comments were addressed, with permission of Marcin I approve this as he cannot do that so this PR won't turn stale.

PR #2784, opened 2019-10-24
Rchain 3902 - change DeployData to Signed[DeployData] with maximum expirationThreshold

odeac:
## Overview
See title

Expiration threshold is set to maximum for now in order to avoid this PR from getting too big

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3880
https://rchain.atlassian.net/browse/RCHAIN-3902

### Notes
<sup>_Optional. Add any notes on caveats, approaches you tried that didn't work, or anything else._</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #2794, opened 2019-11-12
RCHAIN-3902: change DeployData to Signed[DeployData] (ref: PR#2784)

tgrospic:
## Overview
This PR is created from @odeac's branch https://github.com/odeac/rchain/tree/rchain-3910 with expiration threshold set back to 50 and with a fix for deploy error response.
It is related to [PR#2784](https://github.com/rchain/rchain/pull/2784) which contains dirty fix for DeployData deployer, which is in this PR removed.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3880
https://rchain.atlassian.net/browse/RCHAIN-3902


tgrospic:
bors try

tgrospic:
@zsluedem can you please take a look why integration tests are failing here?

tgrospic:
bors try

zsluedem:
@tgrospic 
1. your pr made the rnode not backward compatible which we are allowed to do that. In order to fix the test ,you have to put `@pytest.mark.skip` in front of https://github.com/rchain/rchain/blob/dev/integration-tests/test/test_backward_compatible.py#L37

2. https://github.com/rchain/rchain/blob/dev/integration-tests/test/test_network_topology.py#L77 this test failed. It seemed that the data is not stored in the chain in your pr. I need to investigate it.

3. The slashing test is failed. It is probably that you change the signature which affects the block protocol which the slashing test is using. I suggest you also skip the test.

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -38,12 +39,14 @@ object PrettyPrinter {
     }
 
   def buildString(d: ProcessedDeploy): String =
-    s"User: ${buildStringNoLimit(d.deploy.deployer)}, Cost: ${d.cost.toString} " +
+    s"User: ${buildStringNoLimit(d.deploy.pk.bytes)}, Cost: ${d.cost.toString} " +
```
zsluedem, line 42: you changed the output of the `show-block` command which is one of reason fail the integration tests. I suggest you add the sig in the end instead of replacing the original content.

tgrospic: I don't think that output should be changed because of this. Now are strict rules for signed deploys so maybe now deployer can be different (correct).
Can you post an example? I see signature printed on the end in this function, I'm not sure that I understand what you mean.

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -38,12 +39,14 @@ object PrettyPrinter {
...
       s"${buildString(d.deploy)}"
 
   def buildString(b: ByteString): String =
     limit(Base16.encode(b.toByteArray), 10)
 
+  def buildString(sd: Signed[DeployData]): String =
+    s"Sig: ${buildString(sd.sig)}, SigAlgorithm: ${sd.sigAlgorithm.name}"
   def buildString(d: DeployData): String =
```
tgrospic, line 50: @zsluedem I think this is not called anymore and the whole `DeployData ...` is not printed. Is this what you see?

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -38,12 +39,14 @@ object PrettyPrinter {
...
   def buildString(d: ProcessedDeploy): String =
-    s"User: ${buildStringNoLimit(d.deploy.deployer)}, Cost: ${d.cost.toString} " +
+    s"User: ${buildStringNoLimit(d.deploy.pk.bytes)}, Cost: ${d.cost.toString} " +
       s"${buildString(d.deploy)}"
 
   def buildString(b: ByteString): String =
     limit(Base16.encode(b.toByteArray), 10)
 
+  def buildString(sd: Signed[DeployData]): String =
+    s"${buildString(sd.data)}, SigAlgorithm: ${sd.sigAlgorithm.name}, Sig: ${buildString(sd.sig)}"
```
tgrospic, line 49: @zsluedem signature is now added at the end. Also I put signature algorithm in front because it's much shorter.

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
zsluedem: You are right~! This is the right place to change.

tgrospic:
bors try

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -185,28 +186,20 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
-    case _ =>
-      val maybeVerified = SignDeployment.verify(deployData)
-      maybeVerified.fold[Either[DeployError, Unit]](
-        unknownSignatureAlgorithm(deployData.sigAlgorithm).asLeft
-      ) {
-        case false => signatureVerificationFailed.asLeft
-        case true  => ().asRight
-      }
-  }
+  private def validateDeploy(deployData: DeployData): Either[DeployError, Unit] =
```
marcin-rzeznicki: `Either.cond` is a bit less verbose

models/src/main/scala/coop/rchain/models/EqualsM.scala
```diff
@@ -98,6 +99,17 @@ object EqualM extends EqualMDerivation {
...
 
   implicit val ConnectiveEqual = gen[Connective]
+  implicit def SignedEqual[A: EqualM] = new EqualM[Signed[A]] {
+    override def equal[F[_]: Sync](self: Signed[A], other: Signed[A]): F[Boolean] =
+      EqualM[A]
+        .equal(self.data, other.data)
+        .map(
+          dataEquals =>
+            dataEquals &&
+              self.sigAlgorithm == other.sigAlgorithm &&
```
marcin-rzeznicki: I think it is easier to check signatures first (less expensive check) and check data after (more expensive).

marcin-rzeznicki (approved):
Only minor comments. Good job!

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: This code is removed in later commits, when deployer is removed from DeployData.

tgrospic:
bors try

models/src/main/scala/coop/rchain/models/EqualsM.scala
```diff
@@ -98,6 +99,12 @@ object EqualM extends EqualMDerivation {
   implicit val MatchEqual            = gen[Match]
 
   implicit val ConnectiveEqual = gen[Connective]
+  implicit def SignedEqual[A: EqualM] = new EqualM[Signed[A]] {
+    override def equal[F[_]: Sync](self: Signed[A], other: Signed[A]): F[Boolean] =
+      EqualM[A]
+        .ensuring((self.sigAlgorithm, self.sig) == (other.sigAlgorithm, other.sig))
```
tgrospic, line 105: @marcin-rzeznicki is `ensuring` function OK to use here. For false cases it will throw assertion exception.
https://github.com/scala-native/scala-native/blob/master/scalalib/overrides-2.11/scala/Predef.scala#L169

tgrospic:
Integration tests passed.
https://drone.rchain-dev.tk/rchain/rchain/6255/7

With these tests skipped:
https://github.com/rchain/rchain/pull/2794/commits/11a31a778cec32c4b2f1b9b10d0e6a30f3521bd0
- test_backward_compatible
- test_slash_invalid_validator_approve_evil_block
- test_slash_GHOST_disobeyed

My assumption is that changes in signatures in standard deploys caused these tests to break.
https://github.com/rchain/rchain/pull/2794/files#diff-de0ffe6861ab4a88452dbad0d3a0db76

zsluedem (approved):

models/src/main/scala/coop/rchain/models/EqualsM.scala
marcin-rzeznicki: Why would you want to throw an exception here?

tgrospic: @marcin-rzeznicki I don't want exception here. I was not sure if this exception is handled differently by the compiler.
Any idea what can be used instead of `EqualM[A].ensuring`?

tgrospic: Fix in #2801.

PR #2795, opened 2019-11-12
RCHAIN-3365: Use per-validator vaults in PoS

ghost:
## Overview
This is an attempt to make Timm's old PR https://github.com/rchain/rchain/pull/2724 compile and pass tests

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3365

### Notes
Currently compiles but doesn't pass tests.


ghost:
In RuntimeManagerTest.scala, "PreChargeDeploy" should "reduce user account balance by the correct amount" fails because on line 131, it uses the genesis block, which has no sender, so the PoS deposit contract says "Hey, this block wasn't sent by an active validator!" on line 522.  I could simply return "true" without actually making a deposit in that case, but it feels like a kludge.

This is the only failing test.

casper/src/main/resources/PoS.rhox
```diff
@@ -306,6 +347,33 @@ new  PoS,
...
+                        } else {
+                          returnCh!(true)
+                        }
+                      }
+                    }
+                  } |
+
+                  contract accumulateDeposits(destRevAddress, ackCh) = {
+                    for(@vaults <<- perActiveValidatorPosVaultsCh) {
+                      @ListOps!("foreach", vaults, *moveFunds, *ackCh)
```
JosephDenman: This can be a `parMap`.

casper/src/main/resources/PoS.rhox
```diff
@@ -433,17 +501,27 @@ new  PoS,
           }
         } |
 
-        contract deposit (@deployerId, @amount, returnCh) = {
+        contract deposit (@fromUserPk, @amount, @activeValidators, returnCh) = {
```
JosephDenman: Why are you changing `deployerId` to `fromUserPk`? I see that you use RevVault to compute the `authKey`, but the `bond` and `chargeDeploy` methods still use deployer IDs.

JosephDenman: Ah. Outdated. Nvm.

ghost: Yeah, that first change was part of Timm's code and I didn't notice it was still using the old stuff until later.

JosephDenman:
The problem was that the reference that's read when `getBlockData` is called was empty because we weren't setting the block data during the system deploy logic. Now that I've fixed this, the error being thrown is `Unexpected user error during play: (Bug found) Deploy refund failed: Insufficient funds`. This is happening because `chargeDeploy` transfers REV from the user vault to the validator vault in the PoS contract while `refundDeploy` transfers REV from the PoS vault to the user vault. To fix this, the refund transfer has to be from the PoS validator vault to the user vault.

casper/src/main/resources/PoS.rhox
ghost: Done

ghost:
Timm made https://github.com/rchain/rchain/pull/2808 based on this and merged that, so no more need for this.

PR #2801, opened 2019-11-16
RCHAIN-3902: fix signed type equality

tgrospic:
## Overview
Fix for the change introduced in #2794.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3880
https://rchain.atlassian.net/browse/RCHAIN-3902

### DIFF
--- a/models/src/main/scala/coop/rchain/models/EqualsM.scala
+++ b/models/src/main/scala/coop/rchain/models/EqualsM.scala
@@ -1,7 +1,8 @@
 package coop.rchain.models
 
 import cats.effect.Sync
-import cats.implicits._
+import cats.instances.stream._
+import cats.syntax.all._
 import com.google.protobuf.ByteString
 import coop.rchain.casper.protocol._
 import coop.rchain.catscontrib.Catscontrib._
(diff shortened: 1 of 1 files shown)


kayvank (approved):

models/src/main/scala/coop/rchain/models/EqualsM.scala
```diff
@@ -101,9 +102,8 @@ object EqualM extends EqualMDerivation {
   implicit val ConnectiveEqual = gen[Connective]
   implicit def SignedEqual[A: EqualM] = new EqualM[Signed[A]] {
     override def equal[F[_]: Sync](self: Signed[A], other: Signed[A]): F[Boolean] =
-      EqualM[A]
-        .ensuring((self.sigAlgorithm, self.sig) == (other.sigAlgorithm, other.sig))
-        .equal(self.data, other.data)
+      ((self.sigAlgorithm, self.sig) == (other.sigAlgorithm, other.sig))
```
marcin-rzeznicki: Hye, this is still not optimal because if the first part is `false` the other does not have too be executed. Also allocation of two tuples for a method like this is wasteful. IMO this should be sth like:
```scala
if (self.sigAlgorithm == other.sigAlgorithm && self.sig == other.sig) EqualM[A].equal(self.data, other.data) else Sync[F].pure(false)
```
```

marcin-rzeznicki (changes requested):
I think the `equalM` method should be rewritten

tymm (approved):

marcin-rzeznicki (approved):
I think it's OK!

PR #2809, opened 2019-11-26
Persistent deploys

itegulov:
## Overview
Implements persistent storage for deploys. I made a few decisions while working on this:

- SQLite is used as a lightweight embeded RDBMS
- Doobie is used as a functional JDBC layer
- Flyway is used as a database migration control system

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3840

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
@@ -4,13 +4,14 @@ import cats.effect.{Concurrent, Sync}
 import cats.implicits._
 import coop.rchain.blockstorage.dag.{BlockDagRepresentation, BlockDagStorage}
 import coop.rchain.blockstorage.BlockStore
+import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.casper.CasperState.CasperStateCell
 import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.catscontrib.ListContrib
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.shared.Log
 
-final class LastFinalizedBlockCalculator[F[_]: Sync: Log: Concurrent: BlockStore: BlockDagStorage: SafetyOracle](
+final class LastFinalizedBlockCalculator[F[_]: Sync: Log: Concurrent: BlockStore: BlockDagStorage: SafetyOracle: DeployStorage](
     faultToleranceThreshold: Float
 ) {
   def run(dag: BlockDagRepresentation[F], lastFinalizedBlockHash: BlockHash)(
@@ -32,17 +33,11 @@ final class LastFinalizedBlockCalculator[F[_]: Sync: Log: Concurrent: BlockStore
 
   private def removeDeploysInFinalizedBlock(
       finalizedChildHash: BlockHash
-  )(implicit state: CasperStateCell[F]): F[Unit] =
+  ): F[Unit] =
     for {
-      block              <- ProtoUtil.getBlock[F](finalizedChildHash)
-      deploys            = block.body.deploys.map(_.deploy)
-      stateBefore        <- state.read
-      initialHistorySize = stateBefore.deployHistory.size
-      _ <- state.modify { s =>
-            s.copy(deployHistory = s.deployHistory -- deploys)
-          }
-      stateAfter     <- state.read
-      deploysRemoved = initialHistorySize - stateAfter.deployHistory.size
+      block          <- ProtoUtil.getBlock[F](finalizedChildHash)
+      deploys        = block.body.deploys.map(_.deploy)
+      deploysRemoved <- DeployStorage[F].remove(deploys)
       _ <- Log[F].info(
             s"Removed $deploysRemoved deploys from deploy history as we finalized block ${PrettyPrinter
               .buildString(finalizedChildHash)}."
@@ -73,7 +68,7 @@ object LastFinalizedBlockCalculator {
   def apply[F[_]](implicit ev: LastFinalizedBlockCalculator[F]): LastFinalizedBlockCalculator[F] =
     ev
 
-  def apply[F[_]: Sync: Log: Concurrent: BlockStore: BlockDagStorage: SafetyOracle](
+  def apply[F[_]: Sync: Log: Concurrent: BlockStore: BlockDagStorage: SafetyOracle: DeployStorage](
       faultToleranceThreshold: Float
   ): LastFinalizedBlockCalculator[F] =
     new LastFinalizedBlockCalculator[F](faultToleranceThreshold)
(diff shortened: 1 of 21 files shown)


casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
```diff
@@ -32,17 +33,14 @@ final class LastFinalizedBlockCalculator[F[_]: Sync: Log: Concurrent: BlockStore
...
-      stateBefore        <- state.read
-      initialHistorySize = stateBefore.deployHistory.size
-      _ <- state.modify { s =>
-            s.copy(deployHistory = s.deployHistory -- deploys)
-          }
-      stateAfter     <- state.read
-      deploysRemoved = initialHistorySize - stateAfter.deployHistory.size
+      initialHistorySize <- DeployStorage[F].getUnfinalized.map(_.size)
+      _                  <- DeployStorage[F].remove(deploys)
+      finalHistorySize   <- DeployStorage[F].getUnfinalized.map(_.size)
```
marcin-rzeznicki: Shouldn't this be happening in a single transaction?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -577,6 +580,11 @@ object NodeRuntime {
   val taskToEnv: Task ~> TaskEnv = λ[Task ~> TaskEnv](ReaderT.liftF(_))
   val envToTask: TaskEnv ~> Task = λ[TaskEnv ~> Task](_.run(NodeCallCtx.init))
 
+  private[this] val dbConnScheduler =
+    Scheduler.cached("db-conn", 1, 64)
```
marcin-rzeznicki: Oh, most certainly you do not need that many. https://github.com/brettwooldridge/HikariCP/wiki/About-Pool-Sizing

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -798,4 +822,18 @@ object NodeRuntime {
     val propose               = ProposeGrpcServiceV1.instance(blockApiLock)
     APIServers(repl, propose, deploy)
   }
+
+  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
+  private def runRdmbsMigrations[F[_]: Sync](serverDataDir: Path): F[Unit] =
```
marcin-rzeznicki, line 827: I am philosophically opposed to running db migrations as part of a normal program run. Not only this makes things harder to test in isolation, but also there is a risk of things going wrong on production. IMO it should be a conscious decision

marcin-rzeznicki (commented):
Bravo for using `doobie` ! Nice work BTW - I left few comments

tgrospic:
I have few concerns and questions about persistent storage in RNode.

Adding SQLite would be the third method to store data alongside LMDB and file. Is it possible and convenient to use LMDB to persist deploys? Also having `Signed[DeployData]` type definition in clear text on two places makes additional burden to maintain the code.
I share Marcin concern when to run migration but also I'm not sure that we want to have this logic in SQL if we can have it in Scala only.

Is there a strategy how to persist data in RNode? What is needed to use Rholang and tuple space to save and query data?

casper/src/main/scala/coop/rchain/casper/LastFinalizedBlockCalculator.scala
itegulov: Good point. Fixed

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
itegulov: How do you feel about keeping this as a CLI option? In other words, you will have to run the node with `--migrate-db` to run migrations automatically.

itegulov:
@tgrospic Originally, we were using LMDB for all of our storage, but quickly realized that LMDB is not really well suited for holding several tens of thousand of heavy blocks. Instead of switching to something more conservative like SQL we decided to implement our custom file-based storage to achieve better flexibility and performance (especially for DAG storage). I think it was a wrong decision as file systems are very complicated and implementing our own persistence storage correctly has proven proven to be a consuming task (see https://danluu.com/deconstruct-files/ for some of the shenanigans we have to deal in our file-based storages).

So, to answer your questions:
> Is it possible and convenient to use LMDB to persist deploys?

It is possible, but I would prefer avoid relying on LMDB (for the reasons stated above). Time permitting (which it probably won't), I would rewrite block storage to use SQL as well.

> Also having Signed[DeployData] type definition in clear text on two places makes additional burden to maintain the code.

`readDeployData` and `writeDeployData` are not really type definitions but rather serializations. I would have to implement those (albeit in a slightly different form) even if I were to write a file based storage (see https://github.com/rchain/rchain/blob/dev/block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala).

> I share Marcin concern when to run migration but also I'm not sure that we want to have this logic in SQL if we can have it in Scala only.

See my reply to Marcin on running migrations and feel free to participate in the discussion there. Having to write SQL migrations is a downside, but this is approach is more maintainable than having to deal with migrations in hand-written file storage IMHO.

> Is there a strategy how to persist data in RNode? What is needed to use Rholang and tuple space to save and query data?

I don't understand the first question; there is no general "guidelines" to persistence we have agreed upon previously if that is what you are asking. RSpace is definitely using a least LMDB, but I didn't delve into RSpace deep enough to fully answer that question.

tgrospic:
Thanks Daniyar for very elaborate answer.

I agree that managing custom file-based storage is much harder then with LMDB. And Sqlite claims that can be faster than file system for small blobs.
https://www.sqlite.org/fasterthanfs.html
But it's also single user access. I'm not sure how well will Sqlite behave with concurrent reads/writes in comparison to LMDB. Also can we reuse our code if we need to switch to another DB? I'm little suspicious that SQL can solve our problems.
For me a good example is Datomic which doesn't depend on backend storage because saves blobs of 64kB.
https://docs.datomic.com/on-prem/storage.html#storage-services

For migrations I think is easier to maintain if they can be type checked in Scala.
This is why I don't like these two places where `Signed[DeployData]` is defined/used in clear text [Create table](https://github.com/rchain/rchain/blob/e5b42f54/block-storage/src/main/resources/db/migration/V1__Initial_Setup.sql) definition and [Insert](https://github.com/rchain/rchain/blob/e5b42f54/block-storage/src/main/scala/coop/rchain/blockstorage/deploy/SQLiteDeployStorage.scala#L40) statement.

For strategy I think about more abstract definition of storage where we can have different implementations and not depend of some specific DB.
[RSpace tutorial](https://github.com/rchain/rchain/blob/dev/docs/rspace/Tutorial.md) shows how to work with custom data. Maybe we can sore deploys that way? It would be cool at least. :smile:

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
marcin-rzeznicki: I think it's better. For what it's worth it might as well be a separate program

tgrospic:
We have, from recently, persistence for last finalized block with file-based storage. It has introduced a lot of boilerplate code #2780. What was the reason not to choose LMDB?

itegulov:
Closing in favor of https://github.com/rchain/rchain/pull/2820

PR #2812, opened 2019-11-27
Rewrite ReplayRSpace matching

marcin-rzeznicki:
Flamegraphs shown that it is very slow. Reference: https://cdn.discordapp.com/attachments/476150978555674644/646102858651009027/cpu_test8.svg

This PR is for now in Draft mode till performance tests we're conducting show that it is useful

## Overview
Based on the flamegraph above we can conclude three things:
- Amount of time spent on preparing strings is quite significant. These come from calls to `Log[F]` (but these are lazy) and calls to `assertF` - fixed by making string argument passed to `assertF` lazy
- quite a lot of time spending on looping through sequences, mostly from `zipWithIndex` which is used extensively everywhere!! (this is a sign of really bad design tbh). It made sense to introduce iterated version instead (`seq.iterator.zipWithIndex...toSeq` instead of `seq.zipWithIndex...`) to at least eliminate creation of intermediate collections
- the biggest offender seemed to be `match` call. More specifically it spent a lot of time checking if elements are contained within sequence of `Produce`. Rewritten to make use of `Map` for faster containment check.

### JIRA ticket:

Don't think there is any

### Notes

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
@@ -98,7 +98,11 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
             )
     } yield r
 
-// TODO: refactor this monster
+  private def filterMatching(comm: COMM)(data: Seq[internal.Datum[A]]) =
+    data.iterator.zipWithIndex.filter {
+      case (data, _) => comm.matches(data, produceCounter.get(data.source))
+    }.toSeq
+
   def getCommAndConsumeCandidates(
       channels: Seq[C],
       patterns: Seq[P],
@@ -115,7 +119,7 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
       channelToIndexedDataList <- channels.traverse { c: C =>
                                    store
                                      .getData(c)
-                                     .map(_.zipWithIndex.filter(matches(comm)))
+                                     .map(filterMatching(comm))
                                      .map(c -> _)
                                  }
       result <- extractDataCandidates(channels.zip(patterns), channelToIndexedDataList.toMap, Nil)
@@ -179,22 +183,21 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
       channels =>
         store
           .getContinuations(channels)
-          .map(_.zipWithIndex.filter {
+          .map(_.iterator.zipWithIndex.filter {
             case (WaitingContinuation(_, _, _, _, source), _) =>
               comm.consume == source
-          }),
+          }.toSeq),
       c =>
         store
           .getData(c)
-          .map(_.zipWithIndex)
-          .map(
-            as =>
-              c -> {
-                if (c == channel)
-                  Seq((Datum(data, persist, produceRef), -1))
-                else as
-              }.filter { matches(comm) }
-          )
+          .map { as =>
+            val d = Datum(data, persist, produceRef)
+            val matching = if (c == channel) {
+              if (comm.matches(d, produceCounter.get(produceRef))) Seq(d -> -1) //WTF?
+              else Seq.empty
+            } else filterMatching(comm)(as)
+            c -> matching
+          }
     )
 
   private[this] def handleMatch(
@@ -247,15 +250,6 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
   }
 
   override def clear(): F[Unit] = syncF.delay { replayData.clear() } >> super.clear()
-
-  private[this] def matches(comm: COMM)(datumWithIndex: (Datum[A], _)): Boolean = {
-    val datum: Datum[A] = datumWithIndex._1
-    def wasRepeatedEnoughTimes: Boolean =
-      if (!datum.persist) {
-        comm.timesRepeated(datum.source) === produceCounter.get(datum.source)
-      } else true
-    comm.produces.contains(datum.source) && wasRepeatedEnoughTimes
-  }
 
   protected override def logComm(
       dataCandidates: Seq[ConsumeCandidate[C, A]],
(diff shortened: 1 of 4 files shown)


rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
```diff
@@ -45,7 +45,10 @@ final case class COMM(
     produces: Seq[Produce],
     peeks: SortedSet[Int],
     timesRepeated: Map[Produce, Int]
-) extends Event
+) extends Event {
+  def matches(datum: Datum[_], produceCounter: => Int): Boolean =
+    timesRepeated.get(datum.source).exists(cnt => datum.persist || cnt == produceCounter)
```
JosephDenman: Nitpick: based on what I've heard about abbreviated value names in the Haskell world, I prefer `count` to `cnt`.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -98,7 +98,11 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
             )
     } yield r
 
-// TODO: refactor this monster
+  private def filterMatching(comm: COMM)(data: Seq[internal.Datum[A]]) =
+    data.zipWithIndex.filter {
```
JosephDenman: While we're disparaging `zipWithIndex`, should this not also be `seq.iterator.zipWithIndex...toSeq`?

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -98,7 +98,11 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
...
+      case (d, _) => comm.matches(d, produceCounter.get(d.source))
```
JosephDenman: My nitpick about naming applies here as well. You might consider it redundant, but naming a value based on its type reduces the burden on the programmer to determine the type of the value.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -179,22 +183,21 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
       channels =>
         store
           .getContinuations(channels)
-          .map(_.zipWithIndex.filter {
+          .map(_.iterator.zipWithIndex.filter {
```
JosephDenman, line 186: Why is this more performance than just `zipWithIndex`? Whatever the answer is, i think it'd be a good idea to leave a comment about why you're using `iterator` + `toSeq` instead of `zipWithIndex`.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
marcin-rzeznicki: OMG, it slipped through the cracks. Probably when I was merging my branches. Thanks for noticing!

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
marcin-rzeznicki: When you do `seq.zipWithIndex.filter` you're allocating two extra collections. One that is the result of `seq.zipWithIndex` and the other that is the first one filtered. With `.iterator` the intermediate collection is eliminated, and allocation is constrained to only one (possibly much smaller) collection

marcin-rzeznicki: So, to be really descriptive, what is happening without the iterator is:
```
temp1 = new Builder
for every element at i { temp1 += (element, i) }
afterZipWithIndex = temp1.makeSeq
temp2 = new Builder
for every element of afterZipWithIndex { if (p) temp2 += element }
res = temp2.makeSeq
```

wheras with the iterator:
```
orig = elements.iterator
temp1 = new Iterator { override next = (orig.next, i) }
temp2 = new Iterator { override next = if p(temp1.next) .. }
temp2.toSeq
```

rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
marcin-rzeznicki: Fixed in fb0526681397ec4e067344109bb88797ea6c568d

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
marcin-rzeznicki: Fxied in fb0526681397ec4e067344109bb88797ea6c568d

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
marcin-rzeznicki: Fixed in fb0526681397ec4e067344109bb88797ea6c568d

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -179,22 +183,21 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
...
-              c -> {
-                if (c == channel)
-                  Seq((Datum(data, persist, produceRef), -1))
-                else as
-              }.filter { matches(comm) }
-          )
+          .map { as =>
+            val d = Datum(data, persist, produceRef)
+            val matching = if (c == channel) {
+              if (comm.matches(d, produceCounter.get(produceRef))) Seq(d -> -1) //WTF?
```
tymm, line 196: Why the "wtf"?

marcin-rzeznicki: Oh, sorry, probably a left over from my workings. I'll remove that. Well, this `-1` is just a clever hack_ for `store.removeDatum(candidateChannel, dataIndex).whenA(dataIndex >= 0 && !persistData)` to work. I don't  feel confidence in such hacks (that is, passing "special" values as kind of "wink" to other parts of system that "wink back" because they happen to understand the hack)

marcin-rzeznicki:
Ping @tymm @tgrospic @JosDenmark

tgrospic (approved):
LGTM

9rb (approved):
lgtm

PR #2814, opened 2019-11-28
Remove ad-hoc parallelism bound on Pars.

JosephDenman:
## Overview
The existing `parallelism` value assumes a 2-to-1 mapping from sub-terms 
of a Par to processors. This assumption is far too strong, since the Par 
evaluation strategy depends entirely on the instance of `Parallel` 
passed to `DebruijnInterpreter`.

Monix discourages users from making assumptions about the evaluation order 
that `parTraverse` chooses, so shuffling terms to increase fairness is
superfluous.

This one slipped by me, but this change includes a comment, "Verify if using 
traverse-per-batch inside parTraverse means that less tasks are spawned
at once". The performance benefits of this change should have been verified
before this PR was merged.

### JIRA ticket:
Unticketed

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -33,10 +33,6 @@ import scala.collection.SortedSet
 import scala.collection.immutable.BitSet
 import scala.util.{Random, Try}
 
-object Reduce {
-  val parallelism = lang.Runtime.getRuntime.availableProcessors() * 2
-}
-
 /** Reduce is the interface for evaluating Rholang expressions.
   *
   * @tparam M The kind of Monad used for evaluation.
@@ -151,6 +147,11 @@ class DebruijnInterpreter[M[_], F[_]](
           produce(chan, removedData, false)
       }
 
+  /**
+    * The evaluation of a Par is at the mercy of `Parallel` instance passed to `DebruijnInterpreter`. Therefore, if a
+    * particular concurrency semantics is desired, it should be achieved by modifying the `Parallel` instance, not by
+    * modifying this method.
+    */
   override def eval(par: Par)(
       implicit env: Env[Par],
       rand: Blake2b512Random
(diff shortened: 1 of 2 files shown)


tgrospic (approved):
LGTM

marcin-rzeznicki (approved):
Exactly what I had in mind! 

> Verify if using
traverse-per-batch inside parTraverse means that less tasks are spawned
at once

`parTraverse` is probably implemented (via `Parallel` instance) by spawning fibers which are then scheduled to run on real threads (via starting the event loop) - the actual execution tree will resemble sth like this: `join(fork(b1), join(fork(b2), join(...), )` where `fork` spawns a fiber and `join` *represents a compound fiber* that combines the results. So:`surmise: the task for the next element of a batch is not spawned until the previous one is processed` is, of course, false. And, as you pointed out in this PR, the answer to:
> TODO: Investigate if we can avoid the shuffling and manual parallelism limiting by tweaking [...]

is, of course, **yes**.

Really great!

PS. The actual implementation of this case is in Monix 
```scala
def mapBoth[A1, A2, R](fa1: Task[A1], fa2: Task[A2])(f: (A1, A2) => R): Task[R] =
    TaskMapBoth(fa1, fa2)(f)
```

We need to check how the interpreter deals with `TaskMapBoth`

marcin-rzeznicki:
> term split size should be limited to Int16/2 

This test has a bit non-sensical assertion based on the thrown exception's message

tgrospic:
Test with [maximum number of terms in a Par](https://github.com/tgrospic/rchain/blob/59a8dcf8a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala#L2515) now lasts about 4 times longer.
With `New` constructor it was ~10s now ~40s, and with `Send` or `Receive` was ~30s now ~120s.

Also just using sequential traverse runs faster than `parTraverse_`, for more than ~10k terms it's much slower. This is tested with just empty `Send()`.
```scala
terms.zipWithIndex.toList.traverse_ {
  case (term, index) =>
    eval(term)(env, split(index))
}
```

marcin-rzeznicki:
@tgrospic Yes, beause the tests use the `global` Monix scheduler which tends to submit every async boundary to the global execution context This is likely more costly then actually executing the task!

marcin-rzeznicki:
IOW @JosDenmark @tgrospic it doesn't matter if this test executes longer or shorter. In this particular case the  cost of thread pool management is likely much higher than the cost of evaluating terms. The test is misguided in this sense. The actual parallelism level is function of term complexity which is very low here, so it does not matter if it executes longer or not. If it bothers anyone, please change the scheduler used in tests from `global` to `TestScheduler` and see the difference. (Actually this test is evil and should be removed). Joe's change is very positive in a sense that it eliminates the, also misguided, manual "batching" that certainly was **not** helpful and increases parallelism (with complex terms where term complexity makes up for communication costs).

marcin-rzeznicki:
Ping @JosDenmark @tgrospic

tgrospic:
@marcin-rzeznicki I'm ok to merge this changes and to remove test with max number of terms.

PR #2816, opened 2019-12-09
Limit maximum parent depth

itegulov:
## Overview
Title says it all

### JIRA ticket:
<https://rchain.atlassian.net/browse/RCHAIN-3951>

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/CasperConf.scala
+++ b/casper/src/main/scala/coop/rchain/casper/CasperConf.scala
@@ -38,7 +38,8 @@ final case class CasperConf(
     approveGenesisDuration: FiniteDuration,
     deployTimestamp: Option[Long],
     finalizationRate: Int,
-    maxNumberOfParents: Int
+    maxNumberOfParents: Int,
+    maxParentDepthOpt: Option[Int]
 )
 
 object CasperConf {
--- a/casper/src/main/scala/coop/rchain/casper/Estimator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Estimator.scala
@@ -214,6 +214,9 @@ object Estimator {
   def apply[F[_]](implicit ev: Estimator[F]): Estimator[F] =
     ev
 
-  def apply[F[_]: Sync: Log: Metrics: Span](maxNumberOfParents: Int): Estimator[F] =
-    new Estimator(maxNumberOfParents, None)
+  def apply[F[_]: Sync: Log: Metrics: Span](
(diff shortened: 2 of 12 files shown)


tymm:
@itegulov can you say a few words on what the problem is that this PR is trying to solve? The title does not says it all for me ;)

itegulov:
@tymm sorry, I was planning to add a description once I am finished with the PR. I don't understand why @9rb requested your review even when this PR is a draft.

We are adopting Ethereum's GHOST liveness heuristics. One of them is a restriction to only choose parents which go up to 7 levels deep (i.e. their block number should not be less than `max_block_num - 7`). This should limit the growth of uncommon ancestors (which eventually results in the need to replay hundreds of blocks each time we try to merge two very different branches). This can hopefully mitigate LPT problem.

nzpr:
bors try

zsluedem:
bors try

zsluedem (approved):
LGTM

nzpr:
I've run a test with `--max-parent-depth 7` and this this constraint did not hold. Block `cdf0f84` originated from node4 with `blockNumber: 161` used `1722a11` with `blockNumber: 149` as a parent
Screenshot 2019-12-17 at 15 56 45
However, you can see that node2 does not use deep blocks as parents, so looks like in that case constraint took action.
Here are DAGs for all nodes https://transfer.sh/V7ZOb/dags.png

itegulov:
@nzpr the constraint I have implemented here is somewhat counter-intuitive. It actually limits the depth of secondary parents in regards to the depth of the main parent. In other words, if a block with number 150 is chosen to be a main parent and a block with number 160 is chosen to be a secondary parent, then, even though the depth limit is set to 7, this situation will be deemed to be valid and the block creation will succeed. It won't work vice-versa though (main parent with 160 and secondary parent with number 150).

The behaviour is certainly debatable and I am open to a conversation justifying one approach over another.

nzpr:
@itegulov I believe we should do more empirical tests, should not be hard to try several configurations.  But imo first we need to eliminate the problem with high number of `block retrieve failed` to be able to get valid results. I can see proposal time increasing when some nodes go far ahead as its shown on the DAG above. Might be this discrepancy between local DAGs is caused by nodes failing to propagate blocks created.

nzpr (approved):
As investigation on `block retrieve failed` did not bring much yet, let's merge this PR. It enables one more variable in the equation and noes not change the default behaviour.

PR #2817, opened 2019-12-10
Fix RCHAIN-3775

ghost:
## Overview
Makes it so you have to import crypto primitives from `rho:crypto` instead of using a forgeable name, i.e.
```
new keccak256Hash(`rho:crypto:keccak256Hash`) in {
  keccak256Hash!(...)
}
```
instead of
```
@"keccak256Hash"!(...)
```

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3775


marcin-rzeznicki (approved):

tgrospic (approved):
LGTM

PR #2820, opened 2019-12-16
Persistent deploys via LMDB

itegulov:
## Overview
Implements persistent storage for deploys via LMDB

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3840

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/deploy/DeployStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/deploy/DeployStorage.scala
@@ -1,0 +1,14 @@
+package coop.rchain.blockstorage.deploy
+
+import coop.rchain.casper.protocol.DeployData
+import coop.rchain.crypto.signatures.Signed
+
+trait DeployStorage[F[_]] {
+  def add(deploys: List[Signed[DeployData]]): F[Unit]
+  def remove(deploys: List[Signed[DeployData]]): F[Int]
+  def getUnfinalized: F[Set[Signed[DeployData]]]
+}
+
+object DeployStorage {
+  def apply[F[_]](implicit ev: DeployStorage[F]): DeployStorage[F] = ev
+}
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
@@ -1,0 +1,112 @@
+package coop.rchain.blockstorage.deploy
+
+import java.nio.ByteBuffer
+import java.nio.file.Path
+
+import cats.syntax.traverse._
+import cats.syntax.applicative._
+import cats.syntax.functor._
+import cats.syntax.flatMap._
+import cats.instances.list._
+import cats.effect.{Resource, Sync}
+import com.google.protobuf.ByteString
+import coop.rchain.blockstorage.dag.codecs.codecSignedDeployData
+import coop.rchain.blockstorage.util.io.IOError.RaiseIOError
+import coop.rchain.blockstorage.util.io.{makeDirectory, notExists, IOError}
+import coop.rchain.casper.protocol.DeployData
+import coop.rchain.crypto.signatures.Signed
+import coop.rchain.lmdb.LMDBStore
+import coop.rchain.shared.ByteStringOps._
+import org.lmdbjava.DbiFlags.MDB_CREATE
+import org.lmdbjava.{Env, EnvFlags}
+import scodec.Attempt.{Failure, Successful}
+import scodec.DecodeResult
+import scodec.bits.BitVector
+
+import scala.collection.mutable.ArrayBuffer
+
+case object DeployEncodeError extends Exception
+case object DeployDecodeError extends Exception
+
+class LMDBDeployStorage[F[_]: Sync] private (storage: LMDBStore[F]) extends DeployStorage[F] {
+  def add(deploys: List[Signed[DeployData]]): F[Unit] =
+    deploys
+      .traverse(
+        d =>
+          codecSignedDeployData.encode(d) match {
+            case Successful(value) =>
+              val signature        = d.sig.toDirectByteBuffer
+              val serializedDeploy = ByteString.copyFrom(value.toByteArray).toDirectByteBuffer
+              (signature -> serializedDeploy).pure[F]
+            case Failure(_) =>
+              Sync[F].raiseError[(ByteBuffer, ByteBuffer)](DeployEncodeError)
+          }
+      )
+      .flatMap { serializedDeploys =>
+        storage.put(serializedDeploys)
+      }
+
+  def remove(deploys: List[Signed[DeployData]]): F[Int] =
+    storage.delete(deploys.map(_.sig.toDirectByteBuffer))
+
+  def getUnfinalized: F[Set[Signed[DeployData]]] =
+    for {
+      serializedDeployDataList <- storage.iterate { iterator =>
+                                   val arrayBuffer = ArrayBuffer.empty[BitVector]
+                                   while (iterator.hasNext) {
+                                     val item = iterator.next()
+                                     arrayBuffer += BitVector(item.`val`())
+                                   }
+                                   arrayBuffer.toList
+                                 }
+      deployDatas <- serializedDeployDataList.traverse { serializedDeployData =>
+                      codecSignedDeployData.decode(serializedDeployData) match {
+                        case Successful(DecodeResult(value, _)) =>
+                          value.pure[F]
+                        case Failure(_) =>
+                          Sync[F].raiseError[Signed[DeployData]](DeployDecodeError)
+                      }
+                    }
+    } yield deployDatas.toSet
+
+  def close: F[Unit] =
+    storage.close()
+}
+
+object LMDBDeployStorage {
+  final case class Config(
+      storagePath: Path,
+      mapSize: Long,
+      maxDbs: Int = 1,
+      maxReaders: Int = 126,
+      noTls: Boolean = true
+  )
+
+  def make[F[_]: Sync](config: Config): Resource[F, DeployStorage[F]] = {
+    implicit val raiseIOError: RaiseIOError[F] = IOError.raiseIOErrorThroughSync[F]
+
+    Resource
+      .make[F, LMDBDeployStorage[F]](
+        for {
+          _ <- notExists[F](config.storagePath).ifM(
+                makeDirectory[F](config.storagePath).void,
+                ().pure[F]
+              )
+          env <- Sync[F].delay {
+                  val flags = if (config.noTls) List(EnvFlags.MDB_NOTLS) else List.empty
+                  Env
+                    .create()
+                    .setMapSize(config.mapSize)
+                    .setMaxDbs(config.maxDbs)
+                    .setMaxReaders(config.maxReaders)
+                    .open(config.storagePath.toFile, flags: _*)
+                }
+          dbi <- Sync[F].delay {
+                  env.openDbi("deploy_storage", MDB_CREATE)
+                }
+          lmdbStore = new LMDBStore[F](env, dbi)
+        } yield new LMDBDeployStorage[F](lmdbStore)
+      )(_.close)
+      .widen
+  }
+}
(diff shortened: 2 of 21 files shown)


nzpr:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/DeployStorage.scala
```diff
@@ -0,0 +1,14 @@
+package coop.rchain.blockstorage.deploy
+
+import coop.rchain.casper.protocol.DeployData
+import coop.rchain.crypto.signatures.Signed
+
+trait DeployStorage[F[_]] {
+  def put(deploys: List[Signed[DeployData]]): F[Unit]
```
tgrospic: Maybe _add_ or _append_ would be more appropriate name in this case? My intuition for _put_ is to replace state.

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
```diff
@@ -0,0 +1,112 @@
...
+                                 }
+      deployDatas <- serializedDeployDataList.traverse { serializedDeployData =>
+                      codecSignedDeployData.decode(serializedDeployData) match {
+                        case Successful(DecodeResult(value, _)) =>
+                          value.pure[F]
+                        case Failure(_) =>
+                          Sync[F].raiseError[Signed[DeployData]](DeployDecodeError)
+                      }
+                    }
+    } yield deployDatas.toSet
```
tgrospic, line 70: Do we care about the order of deploys? Do we loose ordering if we use `Set`?

block-storage/src/test/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorageTest.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  implicit val raiseIOError: RaiseIOError[Task] = IOError.raiseIOErrorThroughSync[Task]
+
+  def withStorageLocation[R](f: Path => Task[R]): R = {
+    val testProgram = Sync[Task].bracket {
+      createTemporaryDirectory[Task]("lmdb-deploy-storage")
+    } { storagePath =>
+      f(storagePath)
+    } { storagePath =>
+      deleteRecursively[Task](storagePath).void
```
tgrospic, line 35: Why not just delete the whole folder `storagePath`?

models/src/test/scala/coop/rchain/models/blockImplicits.scala
```diff
@@ -38,12 +38,11 @@ object blockImplicits {
...
   implicit val arbitraryJustification: Arbitrary[Justification] = Arbitrary(justificationGen)
 
-  val processedDeployGen: Gen[ProcessedDeploy] =
+  val signedDeployDataGen: Gen[Signed[DeployData]] =
     for {
-      term        <- listOfN(32, Gen.alphaNumChar).map(_.mkString)
-      timestamp   <- arbitrary[Long]
-      (sec, pub)  = Secp256k1.newKeyPair
-      bytesLength <- Gen.chooseNum(128, 256)
+      term      <- listOfN(32, Gen.alphaNumChar).map(_.mkString)
```
tgrospic: Is there a reason not to use `arbitrary[String]`?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -597,14 +604,20 @@ object NodeRuntime {
...
+  ): Cleanup[F] =
     new Cleanup[F] {
       override def close(): F[Unit] =
         for {
           _ <- Log[F].info("Shutting down interpreter runtime ...")
           _ <- runtime.close()
           _ <- Log[F].info("Shutting down Casper runtime ...")
           _ <- casperRuntime.close()
+          _ <- Log[F].info("Shutting down deploy storage ...")
+          _ <- deployStorageCleanup
```
tgrospic, line 639: Is it makes sense to transform this to an applicative? Maybe closing can be done in parallel.

tgrospic (commented):
Very nice and clean. :) Just a few minor comments.

woky:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
itegulov: We don't really care about their order here. The only thing which matters is that they are unfinalized.

block-storage/src/test/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorageTest.scala
itegulov: Annoyingly, there is no easy way to delete the whole folder without getting [DirectoryNotEmptyException](https://docs.oracle.com/javase/10/docs/api/java/nio/file/DirectoryNotEmptyException.html) using Java's standard library.

models/src/test/scala/coop/rchain/models/blockImplicits.scala
itegulov: It will generate non-alphanum characters which I believe causes some problems. The length though could be arbitrary choosen instead. I will fix it.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
itegulov: In general, this piece of code should only operate on [Resources](https://typelevel.org/cats-effect/datatypes/resource.html). I have been planning to refactor this place for ages.

nzpr:
bors try

block-storage/src/test/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorageTest.scala
marcin-rzeznicki: There are utilities in `apache-commons` - a minor convenience

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/DeployStorage.scala
```diff
@@ -0,0 +1,14 @@
+package coop.rchain.blockstorage.deploy
+
+import coop.rchain.casper.protocol.DeployData
+import coop.rchain.crypto.signatures.Signed
+
+trait DeployStorage[F[_]] {
+  def add(deploys: List[Signed[DeployData]]): F[Unit]
+  def remove(deploys: List[Signed[DeployData]]): F[Int]
```
marcin-rzeznicki, line 8: This is all fine I guess, but I can't help but think that we're loosing the ability to convey logical errors by returning `Unit` in `add`. Also `int` does not tell me much apart from - are all removed? Of course, we may not need this detailed information (or treat any error as "fatal")

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
```diff
@@ -0,0 +1,112 @@
...
+    deploys
+      .traverse(
+        d =>
+          codecSignedDeployData.encode(d) match {
+            case Successful(value) =>
+              val signature        = d.sig.toDirectByteBuffer
+              val serializedDeploy = ByteString.copyFrom(value.toByteArray).toDirectByteBuffer
+              (signature -> serializedDeploy).pure[F]
+            case Failure(_) =>
+              Sync[F].raiseError[(ByteBuffer, ByteBuffer)](DeployEncodeError)
```
marcin-rzeznicki, line 42: This will short-circuit, right? Is it the best thing to do?

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
```diff
@@ -0,0 +1,112 @@
...
+      noTls: Boolean = true
+  )
+
+  def make[F[_]: Sync](config: Config): Resource[F, DeployStorage[F]] = {
+    implicit val raiseIOError: RaiseIOError[F] = IOError.raiseIOErrorThroughSync[F]
+
+    Resource
+      .make[F, LMDBDeployStorage[F]](
+        for {
+          _ <- notExists[F](config.storagePath).ifM(
```
marcin-rzeznicki, line 91: Minor nitpick - `whenA` (or `unlessA`) seems better here (no need for `()` )

shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
```diff
@@ -50,6 +50,24 @@ final case class LMDBStore[F[_]: Sync](env: Env[ByteBuffer], dbi: Dbi[ByteBuffer
       }
     }
 
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def put(kvPairs: List[(ByteBuffer, ByteBuffer)]): F[Unit] =
+    withWriteTxnF { txn =>
+      kvPairs.foreach {
+        case (key, value) =>
+          if (!dbi.put(txn, key, value)) {
```
marcin-rzeznicki, line 58: This can block. I'd really make sure that you're not blocking on the main pool

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -117,6 +118,9 @@ object Setup {
     implicit val lastFinalizedStorage = LastFinalizedMemoryStorage
       .make[Task]
       .unsafeRunSync(monix.execution.Scheduler.Implicits.global)
+    implicit val deployStorage = InMemDeployStorage
+      .make[Task]
+      .unsafeRunSync(monix.execution.Scheduler.Implicits.global)
```
marcin-rzeznicki, line 123: `InMem` does not block, right?

marcin-rzeznicki (approved):
Minor comments - here and there, overall - very elegant

block-storage/src/test/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorageTest.scala
tgrospic: It would be great to have this information in a comment for easier understanding later.

tgrospic (approved):

block-storage/src/test/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorageTest.scala
itegulov: I don't think we have `apache-commons` in dependencies. The code for deleting a directory recursively only takes 7 lines so I couldn't justify bringing it in to just save us 7 lines of code.

I have added a comment as @tgrospic suggested.

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/DeployStorage.scala
itegulov: I don't think there is much to report apart from some errors I would consider fatal (e.g. our scodec codec can't handle some types of deploys as in your comment below).

As for returning `int` in `remove` signature, you are right in that it does not tell us much. I think it is okay as the only it is used in is logging so that you could see whether block finalization removed any old deploys in the logs.

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
itegulov: I can't think of anything you logically do if our scodec codec can't handle some types of deploys apart from crashing. I guess to make the system less brittle we could print to the node user "sorry, we couldn't serialize your deploy" without crashing.

block-storage/src/main/scala/coop/rchain/blockstorage/deploy/LMDBDeployStorage.scala
itegulov: I don't think `whenA` and `unlessA` are applicable here as `notExists[F]` is effectful. I need some function with a signature like `foo(cond: F[Boolean])(f: => F[A]): F[Unit]`, not `whenA(cond: Boolean)(f: => F[A]): F[Unit]`

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
itegulov: Nope

PR #2821, opened 2019-12-16
GitHub Actions CI

woky:
Add GitHub Actions CI
  
Not everything from Drone is ported yet. Uploading artifacts is missing.

### DIFF
--- a/integration-tests/test/rnode.py
+++ b/integration-tests/test/rnode.py
@@ -414,25 +414,6 @@ def make_node(
     return node
 
 
-def get_absolute_path_for_mounting(relative_path: str, mount_dir: Optional[str]=None)-> str:
-    """Drone runs each job in a new Docker container FOO. That Docker
-    container has a new filesystem. Anything in that container can read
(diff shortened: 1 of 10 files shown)


integration-tests/test/rnode.py
```diff
@@ -414,25 +414,6 @@ def make_node(
     return node
 
 
-def get_absolute_path_for_mounting(relative_path: str, mount_dir: Optional[str]=None)-> str:
```
nzpr, line 417: will this break drone once merged?

nzpr (commented):
This PR should be merged after it is complete and ready to replace drone, right?

woky:
> This PR should be merged after it is complete and ready to replace drone, right?

Well, technically it could run alongside Drone, but I've identified few issues since I opened the PR. I'll open another one.

integration-tests/test/rnode.py
woky: Nope. It's just a confusing function and mostly unused function.

PR #2837, opened 2020-01-01
Fix merge conflicts (REV vaultMap, PoS pending rewards state)

tgrospic:
## Overview
Resolve merge conflicts, multiparent history

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3969

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -174,7 +174,7 @@ new  PoS,
                 getUser!(deployerId, *userCh) |
                 for(@state, resultCh <- processCh;
                     @userPk <- userCh;
-                    _, @blockNumber, _ <- blockDataCh) {
+                    @blockNumber, _, _ <- blockDataCh) {
                   if (state.get("allBonds").contains(userPk)) {
                     // the withdwal is in effect starting from the next epoch
                     resultCh!(
@@ -330,7 +330,7 @@ new  PoS,
                   getBlockData!(*blockDataCh) |
 
                   runMVar!(*stateCh, *mvarProcessCh, *ackCh) |
-                  for(_, @blockNumber, _ <- blockDataCh;
+                  for(@blockNumber, _, _ <- blockDataCh;
                       @state, mvarResultCh <- mvarProcessCh) {
                     // Before choosing new active validators and paying out any withdrawers,
                     // deposits have to be accumulated into one vault for now.
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -586,17 +586,16 @@ in {
                 // Get pending rewards (maps) for list of validators (public keys)
                 @ListOps!("fold", validatorKeys, [], *collectMaps, *mapsListCh) |
                 contract collectMaps(@validatorKey, @acc, mergeRet) = {
-                  new currentRewardsCh in {
-                    // Get existing pending rewards (per validator)
-                    // - validator key is always initialized so here we can use unsafe get
-                    TreeHashMap!("fastUnsafeGet", pendingRewardsMap, validatorKey, *currentRewardsCh) |
-                    for (@currentRewards <- currentRewardsCh) {
-                      new stateResetCh in {
-                        // Reset per validator pending rewards state
-                        TreeHashMap!("set", pendingRewardsMap, validatorKey, {}, *stateResetCh) |
-                        for (_ <- stateResetCh) {
-                          mergeRet!(acc ++ [currentRewards])
-                        }
+                  new updateCh, updateAckCh in {
+                    // Update (per validator) pending rewards
+                    // - collect existing values and reset to empty map
+                    TreeHashMap!("update", pendingRewardsMap, validatorKey, *updateCh, *updateAckCh) |
+                    for (@rewardsMap, rewardsRet <- updateCh) {
+                      // Reset per validator pending rewards state
+                      rewardsRet!({}) |
+                      // Accumulate all reward maps
+                      for (_ <- updateAckCh) {
+                        mergeRet!(acc ++ [rewardsMap])
                       }
                     }
                   }
@@ -615,20 +614,18 @@ in {
                 // Get pending rewards (maps) for list of validators (public keys)
                 @ListOps!("fold", validatorKeys, 0, *collectRewards, return) |
                 contract collectRewards(@validatorKey, @acc, sumReturn) = {
-                  new currentRewardsCh in {
-                    // Get existing pending rewards (per validator)
-                    // - validator key is always initialized so here we can use unsafe get
-                    TreeHashMap!("fastUnsafeGet", pendingRewardsMap, validatorKey, *currentRewardsCh) |
-                    for (@rewardsMap <- currentRewardsCh) {
-                      // Reset slashed validator pending rewards
-                      new stateResetCh in {
-                        TreeHashMap!("set",
-                          pendingRewardsMap, validatorKey, rewardsMap.set(collectValidatorKey, 0), *stateResetCh) |
-                        for (_ <- stateResetCh) {
-                          match rewardsMap.get(collectValidatorKey) {
-                            Nil => sumReturn!(acc)
-                            val => sumReturn!(acc + rewardsMap.get(collectValidatorKey))
-                          }
+                  new updateCh, updateAckCh in {
+                    // Update validator pending rewards
+                    // - collect existing values and reset to zero
+                    TreeHashMap!("update", pendingRewardsMap, validatorKey, *updateCh, *updateAckCh) |
+                    for (@rewardsMap, rewardsRet <- updateCh) {
+                      // Reset validator pending rewards state
+                      rewardsRet!(rewardsMap.set(collectValidatorKey, 0)) |
+                      // Sum of validator pending rewards from all validators states
+                      for (_ <- updateAckCh) {
+                        match rewardsMap.get(collectValidatorKey) {
+                          Nil => sumReturn!(acc)
+                          val => sumReturn!(acc + rewardsMap.get(collectValidatorKey))
                         }
                       }
                     }
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -28,7 +28,6 @@ new  PoS,
     getUser,
     getCurrentUserAddress,
     pickActiveValidators,
-    getMVar,
     runMVar,
     fst,
     deposit, distributeRewards,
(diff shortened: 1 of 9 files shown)


casper/src/main/resources/PoS.rhox
```diff
@@ -254,22 +244,29 @@ new  PoS,
...
+                                    initialPaymentAmount - refundAmount,
+                                    state.get("activeValidators"),
+                                    state.get("allBonds"),
+                                    state.get("pendingRewards"),
+                                    *newPendingCh)
+                                } |
+                                for(@newPending <- newPendingCh) {
+                                  for(@state <<- stateCh) {
+                                    Nil
+                                    // if (state.get("pendingRewards") != newPending) {
```
ghost: Please add a reference to the atlassian issue for this bug.  It looks like no validator gets paid until this bug is fixed; is that right?

casper/src/main/resources/RevVault.rho
```diff
@@ -84,27 +85,51 @@ in {
...
+                          match vaultMap.get(revAddress) {
+                            Nil => {
+                              new resCh in {
+                                constructor!(*resCh) |
+                                for (@eitherVault <- resCh) {
+                                  match eitherVault {
+                                    (true, vault) => {
+                                      vaultMapStore!(vaultMap.set(revAddress, vault))
+                                    }
+                                  } |
```
ghost: You still need the `false` case to put `vaultMap` back into `vaultMapStore`.  Otherwise `vaultMap` goes out of scope and is lost.  Please add a test for this case.

casper/src/main/resources/PoS.rhox
tgrospic: PoS is still not for review. This is the place where per validator state must be used. Comments are temporary.

casper/src/main/resources/PoS.rhox
```diff
@@ -299,11 +300,11 @@ new  PoS,
                   new toBeSlashed in {
                     toBeSlashed!(invalidBlocks.getOrElse(blockHash, userPk)) |
                     for (@validator <- toBeSlashed) {
+                      // TODO: remove pending rewards
```
tgrospic: This is still unfinished. @pyrocto I think is enough to map over all validators pending reward maps and collect (and set to zero) current values.

casper/src/main/resources/PoS.rhox
```diff
@@ -330,7 +331,7 @@ new  PoS,
                   getBlockData!(*blockDataCh) |
 
                   runMVar!(*stateCh, *mvarProcessCh, *ackCh) |
-                  for(_, @blockNumber, _ <- blockDataCh;
+                  for(@blockNumber, _, _ <- blockDataCh;
```
tgrospic, line 341: This is fix unrelated to this PR. Block number is the first parameter. [SystemProcesses.scala](https://github.com/tgrospic/rchain/blob/5407c0047/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala#L250)
Thanks @Isaac-DeFrain!

casper/src/main/resources/PoS.rhox
```diff
@@ -531,7 +547,86 @@ new  PoS,
...
+                  TreeHashMap!("set", pendingRewardsMap, key, {}, ackCh)
+                }
+              }
+            } |
+
+            // Sum new pending rewards to existing state
+            contract PendingRewards(@"append", @validatorKey, @newRewards, @return) = {
+              new currentRewardsMaybeCh, currentRewardsCh in {
+                // Get existing pending rewards (per validator)
+                TreeHashMap!("get", pendingRewardsMap, validatorKey, *currentRewardsMaybeCh) |
```
tgrospic: Because all keys must be initialized maybe we can use `fastUnsafeGet` to get validator state?

casper/src/main/resources/PoS.rhox
```diff
@@ -531,7 +547,86 @@ new  PoS,
...
+                // Merge pending rewards (maps) to one map of pending rewards
+                for (@mapsList <- mapsListCh) {
+                  new collector in {
+                    @ListOps!("fold", mapsList, {}, *collector, return) |
+                    contract collector(@map, @acc, @mergeRet) = {
+                      new mergeFunc in {
+                        MergeMaps!(map, acc, *mergeFunc, mergeRet) |
+                        contract mergeFunc(@v1, @v2, valMergeRet) = {
+                          valMergeRet!(v1 + v2)
+                        }
```
tgrospic: Merging of Int maps is the same as in "append", maybe to define contract for that?

casper/src/main/resources/PoS.rhox
```diff
@@ -531,7 +544,86 @@ new  PoS,
                   }
                }
              }
+            } |
+
+            // Initialize pending rewards state for each validator
+            contract PendingRewards(@"init", @validatorKeys, @return) = {
```
ghost, line 555: The argument `validatorKeys` is unused.  Maybe you meant to use it instead of `initialActive` below?

casper/src/main/resources/PoS.rhox
```diff
@@ -65,48 +68,50 @@ new  PoS,
...
+
       for(@posRevAddress <- posRevAddressCh;
           @posAuthKey <- posAuthKeyCh) {
 
         @RevVault!("findOrCreate", posRevAddress, *posVaultCh) |
         pickActiveValidators!($$initialBonds$$, {}, *initialActiveCh) |
 
+        // Initialize pending rewards Tree Hash map
+        // - each key in pending rewards state is Map[PublickKey, Int] accummulated by calling "refundDeploy"
+        TreeHashMap!("init", 3, *pendingRewardsMapCh) |
```
tgrospic, line 80: Maybe in this case _depth_ can be just 1?

casper/src/main/resources/PoS.rhox
ghost: Yes, if you know the key is in the treeHashMap, then you can use fastUnsafeGet to retrieve it quickly.  But also, if you know the key is in the map, then do you need the currentRewardsMaybe below?

casper/src/main/resources/PoS.rhox
```diff
@@ -531,7 +544,86 @@ new  PoS,
...
+                        TreeHashMap!("set", pendingRewardsMap, validatorKey, mergeResult, return)
+                      }
+                    }
+                  }
+                }
+              }
+            } |
+
+            // Collect rewards from all validators / reset state to zero
+            contract PendingRewards(@"collect", @validatorKeys, @return) = {
```
ghost, line 584: Will different invocations of this contract ever be running at the same time?  Do we have to worry about concurrent access to the TreeHashMap when thinking about getting and setting?

casper/src/main/resources/PoS.rhox
tgrospic: It seems that tests are to simple. This is a BUG.

casper/src/main/resources/PoS.rhox
tgrospic: No, I'll change it to use `fastUnsafeGet`.

casper/src/main/resources/PoS.rhox
tgrospic: I'm not sure. Maybe it would be better solution to have "modify" function on TreeHashMap to atomically swap value?

casper/src/test/scala/coop/rchain/casper/util/rholang/InterpreterUtilTest.scala
```diff
@@ -229,12 +228,12 @@ class InterpreterUtilTest
...
     for {
-      b1 <- buildBlock[Task](Seq(genesis.blockHash), deploys = b1DeploysWithCost, now = 1)
-      b2 <- buildBlock[Task](Seq(genesis.blockHash), deploys = b2DeploysWithCost, now = 1)
+      b1 <- buildBlock[Task](Seq(genesis.blockHash), deploys = b1DeploysWithCost, now = 100)
+      b2 <- buildBlock[Task](Seq(genesis.blockHash), deploys = b2DeploysWithCost, now = 200)
       b3 <- buildBlock[Task](
              Seq(b1.blockHash, b2.blockHash),
              deploys = b3DeploysWithCost,
-             now = 2
+             now = 300
```
tgrospic, line 236: @JosDenmark can you confirm increasing these values are valid for this test to pass. This test was failing with _Unable to consume_ error.

casper/src/main/resources/PoS.rhox
tgrospic: It's finished now.

tgrospic: It's finished now.

tgrospic: Here is the ticket to write tests for slashing.
https://rchain.atlassian.net/browse/RCHAIN-3984

tgrospic:
@pyrocto, @tymm, @JosDenmark complete PR is ready for review.

tgrospic:
bors try

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: We'll also need to create TreeHashMaps for bonds, withdraws, and slashing since these methods currently update the global state.

tgrospic:
bors try

tgrospic:
bors try

nzpr:
I've run a network on this branch and multi parent is there!
vdags

casper/src/main/resources/PoS.rhox
tgrospic: Replaced with `MergeIntMaps`, there is no need for more generalized version.

casper/src/main/resources/PoS.rhox
tgrospic: @Isaac-DeFrain maybe it's enough just to keep separate channels to store these values? I'm not sure if per validator state would be beneficial. Am I missing something, can you give some example?

ghost: > atomically swap value?

Yes, I think that supplying an `update` method on TreeHashMap would be a good idea.  It will atomically read the current value, apply a function to it, and write the new value.

Isaac-DeFrain: @tgrospic I think you're right. Just having separate channels for those updates should do the trick.

9rb: > 
> 
> @Isaac-DeFrain maybe it's enough just to keep separate channels to store these values? I'm not sure if per validator state would be beneficial. Am I missing something, can you give some example?

@tgrospic , wouldn't the validators themselves and the network in general be interested in validator state? 
I'm also wondering about the best structural conceptualization for the TreeHashMap .. perhaps validators as branches and the various kinds of rewards and penalties as sub-branches? The corresponding channels can be visualized as leaves interacting with the environment? Does this make sense? This way the Tree represents the structural relationships and the leaves interact with the dynamic ? @pyrocto  ?

tgrospic: @9rb I'm not completely sure what is your notion of _validator state_. E.g. [PoS.rho](https://github.com/rchain/rchain/blob/1aa0765/casper/src/main/resources/PoS.rhox#L95-L130) already exposes validators bonds, rewards and active. Is this answers your first question?

tgrospic: @9rb pending rewards are now implemented very similarly as you described. Every validator is one branch (first level) in the TreeHashMap and it stores Rholang Map with each validator as a key (you can say again - validators as branches (second level)). Leaves of this inner map are Int values for pending rewards.

9rb (commented):
lgtm

9rb (approved):

ghost (approved):

Isaac-DeFrain (approved):

PR #2838, opened 2020-01-03
RCHAIN-3972 make lock more efficient

zsluedem:
## Overview
see the ticket below

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3972

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
@@ -4,21 +4,20 @@ import cats._
 import cats.effect.Concurrent
 import cats.effect.concurrent.Semaphore
 import cats.syntax.all._
-
+import coop.rchain.metrics._
 import coop.rchain.casper.engine._
 import coop.rchain.casper.engine.EngineCell._
 import coop.rchain.casper.engine.Running.RequestedBlocks
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.PrettyPrinter
 import coop.rchain.comm.PeerNode
-import coop.rchain.comm.protocol.routing.Packet
 import coop.rchain.p2p.effects._
 import coop.rchain.shared.Log
-
 import com.google.protobuf.ByteString
+import coop.rchain.comm.protocol.routing.Packet
+import coop.rchain.metrics.Metrics.Source
 
 object CasperPacketHandler {
-
   def apply[F[_]: FlatMap: EngineCell: Log]: PacketHandler[F] =
     (peer: PeerNode, packet: Packet) =>
       toCasperMessageProto(packet).toEither
@@ -28,12 +27,16 @@ object CasperPacketHandler {
           message => EngineCell[F].read >>= (_.handle(peer, message))
         )
 
-  def fairDispatcher[F[_]: Concurrent: EngineCell: Running.RequestedBlocks: Log](
+  def fairDispatcher[F[_]: Concurrent: EngineCell: Running.RequestedBlocks: Log: Span: Metrics](
       maxPeerQueueSize: Int,
       giveUpAfterSkipped: Int,
       dropPeerAfterRetries: Int
-  ): F[PacketHandler[F]] = {
+  )(implicit spanF: Span[F]): F[PacketHandler[F]] = {
     import FairRoundRobinDispatcher._
+
+    implicit val fairDispatcherMetricsSource: Source =
+      Metrics.Source(Metrics.BaseSource, "fairDispatcher")
+
     implicit def showSourceHolder: Show[BlockCreator] = sh => s"[${sh.toString}]"
     implicit val showCasperMessage: Show[(PeerNode, CasperMessage)] = {
       case (_, BlockHashMessage(h, _)) =>
@@ -75,7 +78,17 @@ object CasperPacketHandler {
           .flatMap(proto => CasperMessage.from(proto))
           .fold(
             err => Log[F].warn(s"Could not extract casper message from packet sent by $peer: $err"),
-            msg => lock.withPermit(dispatcher.dispatch(BlockCreator(msg), (peer, msg)))
+            msg =>
+              spanF.trace(fairDispatcherMetricsSource) {
+                for {
+                  _ <- spanF.mark("Casper message received")
+                  _ <- Metrics[F].incrementGauge("fairDispatcher.lock")
+                  _ <- Log[F].debug(s"Received message ${msg.getClass.getSimpleName} from $peer")
+                  _ <- lock.withPermit(dispatcher.dispatch(BlockCreator(msg), (peer, msg)))
+                  _ <- Metrics[F].decrementGauge("fairDispatcher.lock")
+                  _ <- spanF.mark("Casper message handle done")
+                } yield ()
+              }
           )
       }
     }
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
@@ -2,7 +2,6 @@ package coop.rchain.casper.util.comm
 
 import cats._
 import cats.effect.Concurrent
-import cats.effect.concurrent.Semaphore
 import cats.syntax.all._
 import coop.rchain.metrics._
 import coop.rchain.casper.engine._
@@ -16,6 +15,7 @@ import coop.rchain.shared.Log
 import com.google.protobuf.ByteString
 import coop.rchain.comm.protocol.routing.Packet
 import coop.rchain.metrics.Metrics.Source
+import coop.rchain.metrics.MetricsSemaphore
 
 object CasperPacketHandler {
   def apply[F[_]: FlatMap: EngineCell: Log]: PacketHandler[F] =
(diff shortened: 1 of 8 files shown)


casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
```diff
@@ -28,11 +27,13 @@ class FairRoundRobinDispatcher[F[_]: Sync: Log, S: Show, M: Show: Eq](
...
-        ensureSourceExists(source) >>
-          isDuplicate(source, message).ifM(
-            Log[F].info(s"Dropped duplicate message ${message.show} from ${source.show}"),
-            enqueueMessage(source, message) >> handleMessages
-          )
+        lock.withPermit(
+          ensureSourceExists(source) >>
+            isDuplicate(source, message).ifM(
+              Log[F].info(s"Dropped duplicate message ${message.show} from ${source.show}"),
+              enqueueMessage(source, message) >> handleMessages
```
tgrospic, line 34: Am I missing something or `handleMessages` will still process all messages in the queue before lock is released?

shared/src/main/scala/coop/rchain/metrics/MetricsSemaphore.scala
```diff
@@ -23,7 +23,13 @@ class MetricsSemaphore[F[_]: Sync: Metrics](
...
   def tryAcquireN(n: Long): F[Boolean] = underlying.tryAcquireN(n)
   def releaseN(n: Long): F[Unit]       = underlying.releaseN(n)
-  def withPermit[A](t: F[A]): F[A]     = Sync[F].bracket(acquire)(kp(t))(kp(release))
+  def withPermit[A](t: F[A]): F[A] =
+    for {
+      _      <- Metrics[F].incrementGauge("lock.permit")
+      result <- Sync[F].bracket(acquire)(kp(t))(kp(release))
+      _      <- Metrics[F].decrementGauge("lock.permit")
+    } yield result
+//    Sync[F].bracket(acquire)(kp(t))(kp(release))
```
tgrospic: Leftover.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
zsluedem: we have a filter here https://github.com/rchain/rchain/pull/2838/files#diff-ac0666f5850632bde5717addd98fa3f7R28.
So now only handling `blockHashMessage` still is limited by the lock

tgrospic: I'm not sure that I see how filter helps. Is it call to `handleMessages` process all messages in the queue? If so, the lock will not be released until `handleMessages` returns so any new `Handle` that passes filter will block.

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
```diff
@@ -28,12 +27,17 @@ object CasperPacketHandler {
...
       giveUpAfterSkipped: Int,
       dropPeerAfterRetries: Int
-  ): F[PacketHandler[F]] = {
+  )(implicit spanF: Span[F]): F[PacketHandler[F]] = {
     import FairRoundRobinDispatcher._
+
+    implicit val fairDispatcherMetricsSource: Source =
+      Metrics.Source(Metrics.BaseSource, "CasperPacketHandler")
+
+    val fairLock                                      = MetricsSemaphore.single[F]
```
tymm, line 40: `fairLock` has nothing to do with Metrics, right? Therefore I would directly use `Semaphore` here and not use the constructors from `MetricsSemaphore`. Otherwise it's a bit confusing.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
zsluedem: Only `blockHashMessage` is put into the queue. Other messages like `BlockMessage`, `BlockRequest`, `HasBlockRequest`, `ForkChoiceTipRequest` and `HasBlock` are filtered to be case `Pass`(https://github.com/rchain/rchain/pull/2838/files#diff-ac0666f5850632bde5717addd98fa3f7L36) and handle.

That's my understanding of how `FairRoundRobinDispatcher` works. Does that make sense to you?

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
tymm: Also I don't see the difference to the previous solution? `MetricsSemaphore.single[F]` also has one permit. Wouldn't you want to just increase the number of permits?

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
```diff
@@ -63,21 +67,32 @@ object CasperPacketHandler {
...
-        dropPeerAfterRetries
-      ).map { dispatcher => (peer, packet) =>
-        toCasperMessageProto(packet).toEither
-          .flatMap(proto => CasperMessage.from(proto))
-          .fold(
-            err => Log[F].warn(s"Could not extract casper message from packet sent by $peer: $err"),
-            msg => lock.withPermit(dispatcher.dispatch(BlockCreator(msg), (peer, msg)))
-          )
-      }
+    val fairRoundbinDispatcherF = for {
```
tymm: I would prefer `fairLock >>= { lock => ... }` instead of having a `for` with just one line and creating the`fairRoundbinDispatcherF` val (there is also a typo in the name of the val I guess?).

tymm (commented):
Sorry, I don't have enough knowledge about the comm layer to give an approval or block the PR from being merged.

Also I'm afraid that there was a reason that we have the semaphore with one permit for now. I think you should consult Sebastian or Pawel about that if you haven't done that already.

A potential issue might be that this PR won't help since afaik casper currently can't process blocks in parallel (see: https://rchain.atlassian.net/browse/RCHAIN-2709). Making comm able to handle messages in parallel while casper can't, might cause problems under high load.

casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala
tgrospic: I see now what you mean, `BlockHashMessage` is now only message locked when it's processed. What I'm not sure is do we need to wait for `handleMessages` to finish to release the lock.
What you did already greatly reduces waiting on synchronization.

tgrospic (approved):

tgrospic:
@tymm maybe I'm missing something but I don't see that Casper is handling messages differently. Only Fair dispatcher is now not blocked completely to Pass or Drop other messages that are not capt in the queue.

tgrospic:
@zsluedem do you have intention to allow multiple permits?

tymm:
> maybe I'm missing something but I don't see that Casper is handling messages differently

@tgrospic not sure what you mean here.

tgrospic:
@tymm before the whole `dispatcher.dispatch(BlockCreator(msg), (peer, msg))` was locked on every Casper message and now this is only when `BlockHashMessage` is receiver which are of interest for Fair dispatcher.
But on the other hand, [EngineCell#handle](https://github.com/rchain/rchain/blob/2f9d8e92c1/casper/src/main/scala/coop/rchain/casper/engine/Engine.scala#L26) can now be called in parallel. Are you saying this can be a problem?

tymm:
@tgrospic `MultiParentCasperImpl` has a `RuntimeManager` instance which is used for example to process incoming `BlockMessage` (see `attemptAdd`). When we process blocks in casper, we do it by putting a global lock on the runtime that is inside of the `RuntimeManager` instance in casper. This means there is no parallel processing of blocks in casper currently.

If you accept block messages in parallel at the comm layer and only are able to process in sequence at the layers above while not having a fancy backpressure strategy, then you are asking for trouble.

Please double check everything I said though. This is just what came to my mind without having looked into the code enough. I just remember that there was always talk about how casper is not able to process blocks in parallel yet (and the `RCHAIN-2709` ticket seems to be proof of that).

nzpr:
~~Let me mention that @itegulov is working on block buffer where blocks should reside until they are added to a DAG. This work seems to be related and looks like, so maybe worth syncing here. I believe after Daniyar finish we won't see `failed to retrieve block` again. This does not disable work in this PR but I guess will have an impact.~~ Actually no, block buffer is already there and his work on making it persistent wont have an impact.

zsluedem:
> @zsluedem do you have intention to allow multiple permits?

@tgrospic Not right now. I think sebatian adding lock because the fairness dispatcher is not thread safe. I don't know how allowing multiple permit helps.

casper/src/main/scala/coop/rchain/casper/util/comm/CasperPacketHandler.scala
zsluedem: I move the lock inside to not block every message which helps a little bit but it would still block in the `Handle` case and it is fine in `Pass` and `Drop` case.

zsluedem: I mean the problem in the casper should be considered. We can split pr to solve them independently. I think this lock would become a huge problem when the network have many nodes.

9rb:
bors try

9rb (approved):

PR #2840, opened 2020-01-07
Make registry use TreeHashMap

ghost:
## Overview
The registry was storing all its data in a single rholang map.  This meant that every time anything got added to the registry, the whole thing would get read and then written back again.  It also caused conflicts between any two blocks in which something was added to the registry.  This PR modifies the registry to use TreeHashMap, which mitigates these problems.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3978

### Notes
TreeHashMap was originally in its own file and got added to the registry after the registry was initialized.  This PR moves TreeHashMap into Registry.rho, where it is both used and added to the registry.  Nothing changed in TreeHashMap.

Changing the underlying implementation changed the cost of reading/writing to the registry, which bumped BondedStatusAPITest over the 100000 rev limit, so I increased the phlo in ConstructDeploy.scala to a million.

A bunch of the code only changed indentation; I suggest adding w=1 to the query parameters in the review URL.


ghost:
bors try

ghost:
bors try

tgrospic (approved):
LGTM

casper/src/main/resources/Registry.rho
```diff
@@ -8,28 +8,276 @@ new _registryStore,
...
-        `rho:rchain:makeMint` : `rho:id:asysrwfgzf8bf7sxkiowp4b3tcsy4f8ombi3w96ysox4u3qdmn1wbc`,
-        `rho:rchain:pos` : `rho:id:m3xk7h8r54dtqtwsrnxqzhe81baswey66nzw6m533nyd45ptyoybqr`,
-        `rho:rchain:revVault` : `rho:id:6zcfqnwnaqcwpeyuysx1rm48ndr6sgsbbgjuwf45i5nor3io7dr76j`,
-        `rho:rchain:multiSigRevVault` : `rho:id:b9s6j3xeobgset4ndn64hje64grfcj7a43eekb3fh43yso5ujiecfn`
-      } {
-        shorthands => {
-          for (@map <<- _registryStore) {
-            ret!(map.get(shorthands.getOrElse(uriOrShorthand, uriOrShorthand)))
+
+  // TreeHashMap is defined here because the registry uses it internally.
```
tymm, line 15: You could possibly use macro substitution (`CompiledRholangTemplate`) here to separate the TreeHashMap implementation from the Registry.

Isaac-DeFrain (approved):
lgtm

casper/src/main/resources/Registry.rho
```diff
@@ -43,9 +291,13 @@ in {
       new seed, uriCh in {
         ops!("buildUri", *seed.toByteArray(), *uriCh) |
         for (@uri <- uriCh) {
-          for (@map <- _registryStore) {
-            _registryStore!(map.set(uri, data)) |
-            ret!(uri)
+          for (@map <<- _registryStore) {
```
tymm, line 294: Let's assume that there are two inserts that run in parallel where both inserts share the same unpopulated common prefix with a key that is already in the TreeHashMap.
In that case we could loose one insert due to using peek here.
The same applies to `insertSigned`.

tymm (changes requested):
+1 for the conflict analysis in the source code.

casper/src/main/resources/Registry.rho
ghost: No, the peek here just gets the root of the tree.  Once we get to a leaf, the code that updates the map at the leaf, `nodeSetMap`, acquires a lock on the map at the leaf before adding the key.

PR #2842, opened 2020-01-10
Add update method to TreeHashMap

ghost:
## Overview
While the existing set method has internal logic for getting, updating, and setting the internal maps encoding the tree, there was not a client API for similar logic for the keys of the map.  This PR adds an atomic "update" method that reads a value, applies a function to it, and stores back the result.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3986

### DIFF
--- a/casper/src/main/resources/Registry.rho
+++ b/casper/src/main/resources/Registry.rho
@@ -54,8 +54,9 @@ in {
   //   }
   // }
 
-  new MakeNode, ByteArrayToNybbleList, TreeHashMapSetter, TreeHashMapGetter, TreeHashMapContains,
-      powersCh, storeToken, nodeGet, nodeSetBit, nodeSetMap in {
+  new MakeNode, ByteArrayToNybbleList, 
+      TreeHashMapSetter, TreeHashMapGetter, TreeHashMapContains, TreeHashMapUpdater,
+      powersCh, storeToken, nodeGet, nodeSetBit, nodeSetMap, nodeUpdateMap in {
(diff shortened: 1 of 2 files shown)


casper/src/main/resources/Registry.rho
```diff
@@ -247,6 +265,53 @@ in {
               }
             }
           }
+        } |
+        
+        contract TreeHashMapUpdater(@map, @nybList, @n, @len, update, @suffix, ret) = {
+          // Look up the value of the node at [map, nybList.slice(0, n + 1)
```
tgrospic: It looks like missing closing bracket.

casper/src/main/resources/Registry.rho
```diff
@@ -247,6 +265,53 @@ in {
...
+          new valCh, restCh in {
+            nodeGet!((map, nybList.slice(0, n)), *valCh) |
+            for (@val <- valCh) {
+              if (n == len) {
+                // If we're at the end of the path, set the node to newVal.
```
tgrospic: Comment is not related to _update_ but _set_.

tgrospic (approved):
Very nice, just two comments about comments.

Isaac-DeFrain (commented):
lgtm

Isaac-DeFrain (approved):

casper/src/main/resources/Registry.rho
ghost: Done

casper/src/main/resources/Registry.rho
ghost: done

PR #2846, opened 2020-01-15
RCHAIN-3981 - introduction of exploratory deploy

tgrospic:
## Overview
Exploratory deploy allows clients to get data from the blockchain. Exploratory deploy is executed on top of the last finalized block. Future versions will support custom block hash and other nice features.
**Method can be executed only on read-only RNode**.

Also `prepare-deploy` now returns not only block number but also sequence number which is needed  for deploy as parameter _valid after block number_.

**TODO**: exploratory deploy is for now only exposed on Web API. gRPC endpoint and tests will be done in an additional ticket.
https://rchain.atlassian.net/browse/RCHAIN-4018

The sample JS code to make exploratory deploy through Web API:
**Return channel is the first created unforgeable name**.
```js
const url = 'http://localhost:40403/api/explore-deploy'
const req = {
  method: 'POST',
  body: `new return in { return!(42) }`
}
fetch(url, req).then(r => r.json()).then(console.log)
// Returns `42` as Rholang AST (Seq[Par])
```
### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3981

### Notes
Epic
https://rchain.atlassian.net/browse/RCHAIN-3979
Design document
https://rchain.atlassian.net/wiki/spaces/CORE/pages/492732472/Obtaining+Data

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -71,6 +71,8 @@ trait RuntimeManager[F[_]] {
   ): F[Seq[(Seq[BindPattern], Par)]]
   def emptyStateHash: StateHash
   def withRuntimeLock[A](f: Runtime[F] => F[A]): F[A]
+  // Executes deploy as user deploy with immediate rollback
+  def playExploratoryDeploy(term: String, startHash: StateHash): F[Seq[Par]]
 }
 
 class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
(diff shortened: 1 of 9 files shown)


nzpr:
https://github.com/rchain/rchain/blob/ca9a3bc5f7b4332899968e2d5320698c9b6e07db/node/src/test/scala/coop/rchain/node/web/WebApiRoutesTest.scala#L122
Compilation failed because of not sufficient args for `PrepareResponse`

tgrospic:
Thanks @nzpr. I didn't seen that immediately.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -705,6 +708,20 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
        # }
        """.stripMargin('#')
 
+  // Executes deploy as user deploy with immediate rollback
+  // - InterpreterError is rethrown
+  def playExploratoryDeploy(term: String, hash: StateHash): F[Seq[Par]] = {
```
zsluedem, line 713: We better have tests for this function

zsluedem (approved):
LGTM. It would be better to have tests for it.

9rb (approved):
Please add any additional tests as a separate PR by Tomislav or Will. Since this is a ReadOnly Node functionality, may benefit from early testing by the community also

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
tgrospic: I've created new ticket for tests.
https://rchain.atlassian.net/browse/RCHAIN-4018

Isaac-DeFrain (approved):
LGTM

fabcotech:
If a private name `records` is not exposed through the registry, meaning I can't access it inside rholang, let's say the unforgeable name id is `abcdabcd`, will I be able to get the value with exploratory deploy ??

tgrospic:
@fabcotech we will add this in the future versions where input can be Rholang AST, and also unforgeable name similar how you already doing with _listen for data at name_. This will be special part of exploratory deploy which can not be valid with regular deploys.
For now you want to use registry as a way to share global channel on the blockchain that can be accessed with exploratory deploy.

PR #2848, opened 2020-01-16
Remove `ErrorLog` and use `Ref` to flag errors instead of `raiseError`.

JosephDenman:
## Overview
As the title says, this PR replaces the `raiseError` channel of `F[_]` with an explicit error channel implemented as a `Ref[F, Option[Throwable]]`. The `raiseError` behavior in the reducer would previously cross concurrent boundaries, thus resulting in non-deterministic thread cancellation, thus resulting in pesky concurrency bugs. Basically, if the process is `P | Q`, and `P` throws an error, `raiseError` is used to short-circuit `P`. What's different now is that instead of letting `raiseError` try to cancel `Q`, we use a `Ref` to signal to `Q` that an error has been thrown. This is obviously not replayable; it assumes a deployment policy wherein any program that throws an error elects to forfeit all of the REV that the deployment is allocated.


rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -51,16 +51,15 @@ class RholangAndScalaDispatcher[M[_]] private (
 
 object RholangAndScalaDispatcher {
 
-  def create[M[_], F[_]](
+  def create[M[_]: _error, F[_]](
```
marcin-rzeznicki, line 54: I'm aware that the code base is riddled with bad examples like this - but I have to say it - this is not an appropriate use of type-classes, `Ref` is not a capability or constraint - it's data.

marcin-rzeznicki (approved):
Great idea! One minor comment

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
JosephDenman: I agree. I'm merely replacing a bad example `ErrorLog` with another bad example `_error`.

tgrospic:
I don't think this PR is solving the root cause of unhandled errors and adds more complexity to manage errors. The issues that I have:
- guarding from errors is done on multiple places and not inside _parTraverse_ where/when new Tasks are created/exited
- keeps errors in global state in the Runtime and adds more logic to that
- it's more error prone because developer needs to know where to put additional checks

My suggestion is simple fix applied when _parTraverse_ is called. If we handle errors at the source we can use _raiseError_ without problems everywhere in our code.

```scala// Collect errors from all parallel execution paths (pars)
terms.zipWithIndex.toVector
  .parTraverse {
    case (term, index) =>
      eval(term)(env, split(index))
        .map(_ => none[Throwable])
        // This will prevent for Task to finish with an error
        // so failed Task cannot occur as a result of parTraverse
        .handleError(_.some)
  }
  .map(_.flattenOption)
  .flatMap(aggregateEvaluatorErrors)
```
Here is the PR #2853 with the fix and with `ErrorLog` completely removed. :tada:

Also in your PR you still have one bug that should cause this test in [RuntimeManagerTest](https://github.com/rchain/rchain/blob/890ffbb20/casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala#L605) to fail. The error in user code is not causing all phlos to be consumed.

tgrospic:
I found why in case of an error all cost is not consumed in [RuntimeManagerTest](https://github.com/rchain/rchain/blob/890ffbb20/casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala#L605).
In [Interpreter](https://github.com/rchain/rchain/blob/344a88d/rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala#L104-L107), case with only interpreter errors does not affect cost calculation. Is this maybe intended behavior?

JosephDenman:
Why would a program that didn't thrown an out of gas error consume all of its gas? Gas is just a counter.

tgrospic:
My understanding was that in case of an execution error there should not be refund and pre-charge should take phloLimit. Greg confirmed that and also that for parsing errors only parsing cost should be charged. Am I missing something?

JosephDenman:
That's correct, but AFAIK that isn't implemented yet. That fell under Marcin's work. To be clear, the maximum REV allocation associated with the deploy should be forfeited if there is an error. This is independent of the value of the gas counter since the maximum REV allocation is specified in the deploy.

tgrospic:
OK I see, this should not be decided inside interpreter but later as part of processing deploy. So it's not a bug in RuntimeManagerTest.

tgrospic:
Do you know what is the reason to forfeit all REV's for execution error in deploy?

JosephDenman:
If a program throws an OutOfGasError, all REV must be forfeited anyway. We decided to forfeit in case of _any_ error because, otherwise, you have to compare the gas consumed by a program during play and replay even if it threw an error. To reliably compare costs, you have to know that the program consumes the same amount of gas during play and replay even if there's an error reported, and to know that, you have to be able to cancel threads deterministically. And to be able to cancel threads deterministically, you have to not be trying to release main net in two weeks 😄.

tgrospic:
Closing this PR as it's superseded by #2853 and #2874 which fixes the use of cats `Parallel` instance and removes `ErrorLog` completely.

The future work is to implement re-playable errors.
https://rchain.atlassian.net/browse/RCHAIN-3505

PR #2849, opened 2020-01-16
Add logging to purses and rev vaults

ghost:
## Overview
Allow purse owners and rev vault owners to log deposits in and transfers out.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3983

### DIFF
--- a/casper/src/main/resources/MakeMint.rho
+++ b/casper/src/main/resources/MakeMint.rho
@@ -47,6 +47,7 @@ in {
                 }
               }
             } |
+            // Set to Nil to disable logging
             contract thisPurse(@"setLog", logCh, ack) = {
               for (_ <- logStore) { logStore!(*logCh) | ack!(Nil) }
             } |
@@ -68,9 +69,9 @@ in {
             } |
             contract thisPurse(@"deposit", @amount, @src, success) = {
               new result in {
-                for (logCh <<- logStore) {
-                  @src!(bundle0{bundle0{src}|*decr}, amount, *result) | //call src decr function.
-                  for(@decrSuccess, @auth <- result) {
+                @src!(bundle0{bundle0{src}|*decr}, amount, *result) | //call src decr function.
+                for(@decrSuccess, @auth <- result) {
+                  for (logCh <<- logStore) {
                     if (decrSuccess and auth == *decr) {
                       if (Nil != *logCh) { logCh!(["deposit", amount, true]) } |
                       balance!("add", amount, *success) // add transferred amount to this purse
--- a/casper/src/main/resources/MakeMint.rho
+++ b/casper/src/main/resources/MakeMint.rho
@@ -31,17 +31,24 @@ in {
         } |
         contract internalMakePurse(balance, return) = {
           //balance must be a name that NonNegativeNumber contract is listening on
-          new thisPurse in {
+          new thisPurse, logStore in {
+            logStore!(Nil) |
             new letCh in { // let thisDecr = ...
               letCh!(bundle0{bundle0{*thisPurse}|*decr}) | for (thisDecr <- letCh) {
                 contract thisPurse(@=*thisDecr, @amount, return) = {
-                  new success in {
-                    balance!("sub", amount, *success) | for (@subOk <- success) {
-                      return!(subOk, *decr)
+                  for (logCh <<- logStore) {
+                    new success in {
+                      balance!("sub", amount, *success) | for (@subOk <- success) {
+                        if (Nil != *logCh) { logCh!(["*decr", amount, subOk]) } |
+                        return!(subOk, *decr)
+                      }
                     }
                   }
                 }
               }
+            } |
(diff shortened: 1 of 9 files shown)


casper/src/main/resources/MakeMint.rho
```diff
@@ -31,18 +31,25 @@ in {
...
             new letCh in { // let thisDecr = ...
               letCh!(bundle0{bundle0{*thisPurse}|*decr}) | for (thisDecr <- letCh) {
                 contract thisPurse(@=*thisDecr, @amount, return) = {
-                  new success in {
-                    balance!("sub", amount, *success) | for (@subOk <- success) {
-                      return!(subOk, *decr)
+                  for (logCh <<- logStore) {
+                    new success in {
+                      balance!("sub", amount, *success) | for (@subOk <- success) {
+                        if (Nil != *logCh) { logCh!(["*decr", amount, subOk]) } |
```
tgrospic: Is there a meaning for `*` in the log message?

casper/src/main/resources/MakeMint.rho
```diff
@@ -61,12 +68,16 @@ in {
...
-                for(@decrSuccess, @auth <- result) {
-                  if (decrSuccess and auth == *decr) {
-                    balance!("add", amount, *success) // add transferred amount to this purse
-                  } else {
-                    success!(false) 
+                for (logCh <<- logStore) {
+                  @src!(bundle0{bundle0{src}|*decr}, amount, *result) | //call src decr function.
+                  for(@decrSuccess, @auth <- result) {
+                    if (decrSuccess and auth == *decr) {
+                      if (Nil != *logCh) { logCh!(["deposit", amount, true]) } |
```
tgrospic: `logCh` scope can be inside the scope of `for`.

casper/src/main/resources/MakeMint.rho
```diff
@@ -31,18 +31,25 @@ in {
...
+                        if (Nil != *logCh) { logCh!(["*decr", amount, subOk]) } |
+                        return!(subOk, *decr)
+                      }
                     }
                   }
                 }
               }
             } |
+            contract thisPurse(@"setLog", logCh, ack) = {
+              for (_ <- logStore) { logStore!(*logCh) | ack!(Nil) }
```
tgrospic, line 63: What is correct way to disable logging? Maybe it would be useful as comment here.

Is it makes sense to have method to check if log exists?

tgrospic (approved):
LGTM, only few minor comments.

casper/src/main/resources/MakeMint.rho
ghost: That's the method being invoked.  The method name is not a string like "deposit", it's the process `*decr`.

casper/src/main/resources/MakeMint.rho
ghost: > What is correct way to disable logging? Maybe it would be useful as comment here.

Set it to Nil.  I'll add a comment.

> Is it makes sense to have method to check if log exists?

It always exists because I set it on line 35 and then only peek for it/set it to something new.

casper/src/main/resources/MakeMint.rho
ghost: will change.

JosephDenman:
It surprises me that the logging isn't implemented at the wallet level, so that the log captures the identity of the transactor. What is the use-case for a purse that maintains a history of decrements?

Addendum:
You can use it as an event publisher/consumer utility for reacting to balance changes, but that still seems to fall short of a more useful service that records and provides transactor identity.

casper/src/main/resources/MakeMint.rho
ghost: done

ghost:
@JosDenmark I added Vault-level logging and block data.  Does that more closely match your thinking?

ghost:
bors try

JosephDenman:
@pyrocto You have a couple of integration tests failing. After those are fixed, LGTM.

JosephDenman (approved):

casper/src/main/resources/MakeMint.rho
```diff
@@ -63,10 +82,33 @@ in {
...
+                        }
+                      }
+                    } else {
+                      success!(false) |
+                      
+                      // log
+                      if (Nil != *logCh) {
+                        new bd(`rho:block:data`), bdCh in {
+                          bd!(*bdCh) |
+                          for (@timestamp, @blockNumber, @sender <- bdCh) {
```
tgrospic, line 106: Block number is on the first place - [SystemProcesses.scala](https://github.com/rchain/rchain/blob/f6050234/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala#L263).

ghost: Thanks.  I've fixed it everywhere in my tests and in BlockDataContractTest.rho where I copied it from.

9rb:
bors try

PR #2851, opened 2020-01-17
RCHAIN-3993 Transfer slashed funds to Coop vault

Isaac-DeFrain:
### Overview
Add Coop multiSig Rev vault to PoS

Create PoS vault with initial balance = sum of initial bonds

Remove getCurrentUserAddress contract from PoS

Slash method returns Either (SlashDeploy)

Short-circuit closeBlock if no epoch change occurs

Slashed validator moved to withdrawers map with correct block number

Add comments to PoS

Add linear consume for PoS vault creation to RevVault

Add CasperInvalidBlocksContract to create invalid blocks for PoSTest

Add slashing test to PoSTest

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3993

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -20,13 +20,13 @@
  8.  | 7, 1,      | secp256k1          | sig = 3044022054ff4bae3984252b116e41e28d98bb5533eaa39aec2729228159166e2784f641022066a0fd99e7ea33df812fab095cbe61250f9548bce6da3ec4c6a90c741b94087f
  9.  | 4,         | registry           | uri = rho:id:m3xk7h8r54dtqtwsrnxqzhe81baswey66nzw6m533nyd45ptyoybqr
  ----+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------
- */
-new  PoS,
+*/
+
+new PoS,
     registryLookup(`rho:registry:lookup`),
     revAddressOps(`rho:rev:address`),
     revVaultCh, listOpsCh,
     getUser,
-    getCurrentUserAddress,
     pickActiveValidators,
     runMVar,
     fst,
@@ -37,20 +37,22 @@ new  PoS,
     posId(`rho:rchain:deployerId`),
     PendingRewards, MergeIntMaps, treeHashMapCh,
     sysAuthTokenOps(`sys:authToken:ops`),
-    uriOut
+    multiSigRevVaultCh, coopVaultCh,
+    uriOut, sumFromPair
 in {
+  // Simultaneously retrieves RevVault, ListOps, TreeHashMap, and MultiSigRevVault contracts from the registry.
   registryLookup!(`rho:rchain:revVault`, *revVaultCh) |
   registryLookup!(`rho:lang:listOps`, *listOpsCh) |
   registryLookup!(`rho:lang:treeHashMap`, *treeHashMapCh) |
-
-  for(@(_, RevVault) <- revVaultCh;
-      @(_, ListOps) <- listOpsCh;
-      TreeHashMap <- treeHashMapCh) {
-    new posRevAddressCh, posAuthKeyCh, posVaultCh, initialActiveCh, pendingRewardsMapCh in {
-
-      @RevVault!("deployerAuthKey", *posId, *posAuthKeyCh) |
-      getCurrentUserAddress!(*posId, *posRevAddressCh) |
-
+  registryLookup!(`rho:rchain:multiSigRevVault`, *multiSigRevVaultCh) |
+  for (@(_, RevVault)         <- revVaultCh;
+       @(_, ListOps)          <- listOpsCh;
+       TreeHashMap            <- treeHashMapCh;
+       @(_, MultiSigRevVault) <- multiSigRevVaultCh) {
+    new posRevAddressCh, posAuthKeyCh, posVaultCh,
+        initialActiveCh, pendingRewardsMapCh
+    in {
+      // Creates a Rev vault for each active validator, in parallel.
       new rangeCh, makeVault in {
         contract makeVault(_, returnCh) = {
           new unf, revAddressCh, vaultCh in {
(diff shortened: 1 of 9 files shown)


casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
+                                          // Collects pending rewards from each validator and resets to zero.
+                                          PendingRewards!("collect", state.get("activeValidators"), *pendingRewardsAllCh) |
+                                          for (@pendingRewards <- pendingRewardsAllCh) {
+                                            new ackInitCh in {
+                                              // Initialize empty pending rewards state with current set of validators.
+                                              PendingRewards!("init", state.get("activeValidators"), *ackInitCh) |
+                                              for (_ <- ackInitCh) {
+                                                // Add all pending rewards to global committed rewards.
+                                                commitRewards!(pendingRewards, state.get("committedRewards"), *commitRewardsCh)
+                                              }
```
tgrospic, line 390: Why is necessary to transfer pending rewards to committed rewards here? If pending rewards are collected and reset so zero.
`PendingRewards!("collectValue", validator, state.get("activeValidators"), *valPendingRewardsCh)` will take only value for collect pending value for one validator from all validators states.

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
                             }
-                          } |
-
-                          contract isQuarantineFinished(@(_, blockNumber), resultCh) = {
-                            resultCh!(blockNumber >= currentBlockNumber)
                           }
                         }
+                      } else {
+                        ackCh!((false, "Invalid system auth token"))
                       }
```
tgrospic, line 589: When successful `closeBlock` will return just `Nil` (line 477 send on `stateUpdateCh`). Maybe it makes sense to return the same type (Either). It's similar in "slash" which returns Boolean if successful.

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
+                                                coopRevAddress,
+                                                valRewards + valBond,
+                                                posAuthKey,
+                                                *transferDoneCh
+                                              ) |
+                                              for (_ <- transferDoneCh) {
+                                                // Sets slashed validator's committed rewards and bond to 0 in state.
+                                                slashResultCh!({
+                                                  "allBonds"        : state.get("allBonds").set(validator, 0),
+                                                  "activeValidators": state.get("activeValidators"),
```
tgrospic: When is slashed validator removed from the list of active validators?

tgrospic (commented):
Kudos for extra comments. Only few remarks and questions.

casper/src/main/resources/PoS.rhox
```diff
@@ -20,13 +20,14 @@
...
-new  PoS,
+*/
+
+new PoS,
     registryLookup(`rho:registry:lookup`),
     revAddressOps(`rho:rev:address`),
     revVaultCh, listOpsCh,
     getUser,
-    getCurrentUserAddress,
+    //getCurrentUserAddress,
```
ghost: Cruft

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
+      // Creates the auth key and Rev vault associated with the PoS contract.
+      new posPkCh in {
+        @RevVault!("deployerAuthKey", *posId, *posAuthKeyCh) |
+        // Not required... but can be restored after real wallet public keys have been created.
+        //getCurrentUserAddress!(*posId, *posRevAddressCh) |
+        getUser!(*posId, *posPkCh) |
+        for (@posPk <- posPkCh) {
+          revAddressOps!("fromPublicKey", posPk, *posRevAddressCh) |
+          // Creates Coop multisig vault for slashed funds.
+          // (false, errorMsg) xor (true, (multiSigVault, revAddress, revVault)) are returned on coopWalletCh.
```
ghost: Please say "either ... or".  It took me way too long to figure out how you intended to xor tuples.

casper/src/main/resources/PoS.rhox
```diff
@@ -37,20 +38,22 @@ new  PoS,
     posId(`rho:rchain:deployerId`),
     PendingRewards, MergeIntMaps, treeHashMapCh,
     sysAuthTokenOps(`sys:authToken:ops`),
+    multiSigRevVaultCh, coopWalletCh,
```
ghost: "Vault" instead of "Wallet" here and below.  In all other blockchains, "wallets" hold keys and decrypt them when you type in your password, so we're trying to move away from that terminology to describe on-chain contracts.

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
-                              (unf, _, activeValidatorVault) => {
-                                @RevVault!("unforgeableAuthKey", unf, *unfAuthKeyCh) |
-                                for (@authKey <- unfAuthKeyCh) {
-                                  @activeValidatorVault!("transfer", deployerRevAddress, refundAmount, authKey, *returnCh)
+                  getUser!(deployerId, *userCh) |
+                  for (@state, resultCh   <- processCh;
+                       @userPk            <- userCh;
+                       @blockNumber, _, _ <- blockDataCh) {
+                    if (state.get("allBonds").contains(userPk)) {
+                      // The withdwal goes into effect at the sart of the next epoch.
```
ghost: typo: "start" instead of "sart"

Isn't there a waiting period of multiple months to cash out in order to avoid long-range attacks?

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +66,646 @@ in {
...
+                  new getBlockData(`rho:block:data`),
+                      blockDataCh, indexSender,
+                      unfAuthKeyCh, tmpCh
+                  in {
+                    getBlockData!(*blockDataCh) |
+                    for (_, _, @sender <- blockDataCh) {
+                      for (@state <<- stateCh) {
+                        @ListOps!("indexOf", state.get("activeValidators"), sender, *indexSender) |
+                        for (@idx <- indexSender) {
+                          for (@vaults <<- perActiveValidatorPosVaultsCh) {
```
ghost: Can move this into the else clause

ghost (commented):
Some initial comments.  I'm starting to review the main logic now.

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: Thanks for the lesson! I changed the terminology.

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: That is my understanding too. This is how the logic was in PoS originally. I thought the quarantine period was for that purpose.

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: We require a transfer from all per validator vaults, an update of each per validator state, and an update of the global state no matter how we slice it. This seemed like the easiest way to achieve all of that and didn't introduce any new contracts/methods.

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: This is something that still needs to be decided. In this change, the slashed validator's bond is transferred to the Coop vault and set to 0 so they will not receive any rewards because those are stake-weighted. I think the right answer is either to immediately move them to the withdrawers map (not sure how this would be rectified if the slash was found to be a mistake) or to include a slashed map in the global state with some grace period before they are moved to withdrawers.

casper/src/main/resources/PoS.rhox
```diff
@@ -252,437 +316,397 @@ in {
...
+                                for (_ <- paymentDoneCh) {
+                                  stateUpdateCh!({
+                                    "allBonds"        : newBonds,
+                                    "activeValidators": newValidators,
+                                    "committedRewards": newRewards,
+                                    "withdrawers"     : newWithdrawers,
+                                    "randomImages"    : {},
+                                    "randomNumbers"   : {},
+                                  },
+                                  (true, Nil))
```
Isaac-DeFrain: @tgrospic better? Return (true, Nil) if successful.

zsluedem (dismissed):
I found there are no unit tests for slashSystemDeploy. Can you add that?

casper/src/main/resources/PoS.rhox
9rb: Slashed validator should be immediately removed from active validators list if fully slashed. We do need to decide however, after how many validators are slashed do we require an epoch change - i.e. is there a minimum guaranteed size to the number of active validators ?

9rb:
bors try

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -15,6 +15,7 @@ object RevGenerator {
     if (userVaults.nonEmpty) userVaults.zipWithIndex.map(Function.tupled(f)).mkString(separator)
     else "Nil"
 
+  // TODO: decrement stake from initial vault
   private def vaultInitCode(userVaults: Seq[Vault]): String = {
     def initVault(userVault: Vault, index: Int): String =
```
Isaac-DeFrain: How can the Rev vault balance be decremented by the validator's stake amount?

Isaac-DeFrain:
bors try

casper/src/main/resources/PoS.rhox
Isaac-DeFrain: Slashed validator is immediately placed in the withdrawers map.

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +65,682 @@ in {
...
+                            removeQuarantinedWithdrawers
+                        in {
+                          new blockDataCh, stateProcessCh,
+                              getBlockData(`rho:block:data`),
+                              accDepositsDoneCh, commitRewardsCh,
+                              newValidatorsCh, removeQuarantinedCh,
+                              paymentDoneCh
+                          in {
+                            getBlockData!(*blockDataCh) |
+                            runMVar!(*stateCh, *stateProcessCh, *ackCh) |
```
tgrospic: This `runMVar` will _change_ the state so each block is conflicting. We should _peek_ state here and only change state on epoch change.
@Isaac-DeFrain I think you already done this when we were resolving mergeability issues. Can you include it in this PR?

Isaac-DeFrain: I must've missed that while merging conflicts... I'll make that change. Thanks for catching it!

Isaac-DeFrain: Currently, the epochLength is set to 1. So we're changing epochs at every closeBlock.

tgrospic: Are you referring to test configuration for epochLength? Default in [config](https://github.com/rchain/rchain/blob/d8a8e7f/node/src/main/resources/reference.conf#L160) is 10000.

Isaac-DeFrain: Okay, that's great! I was referring to https://github.com/rchain/rchain/blob/dev/casper/src/test/scala/coop/rchain/casper/util/GenesisBuilder.scala#L54

casper/src/test/scala/coop/rchain/casper/util/GenesisBuilder.scala
```diff
@@ -51,15 +51,16 @@ object GenesisBuilder {
         proofOfStake = ProofOfStake(
           minimumBond = 0L,
           maximumBond = Long.MaxValue,
-          epochLength = 1,
+          epochLength = 1, // TODO: change to proper quantity
```
tgrospic: This is genesis created for testing so it should be fine with fixed value. Does PoS tests depend on _epochLength_ = 1?

tgrospic (approved):
Very nice work on PoS!

casper/src/test/scala/coop/rchain/casper/util/GenesisBuilder.scala
Isaac-DeFrain: Since I introduced the block number set capability, I've been making the PoS tests parametric in the epoch length and quarantine length to test the behavior in relation to these quantities, e.g. that a withdrawing validator is not paid before their quarantine period is finished and they are payed at the first epoch change after it's finished.

tgrospic: Epoch length of 1 was causing tests with merging block to fail. I've updated your PR to include close block system deploy and default epoch length large enough to prevent conflicts in tests.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -73,12 +73,10 @@ object BlockCreator {
...
-        systemDeploys = if (slashingDeploys.nonEmpty)
-          slashingDeploys ++ Seq(
-            CloseBlockDeploy(Tools.rng(parents.head.blockHash.toByteArray))
-          )
-        else List.empty[SystemDeploy]
-        unsignedBlock <- if (deploys.nonEmpty || systemDeploys.nonEmpty) {
+        systemDeploys = slashingDeploys :+ CloseBlockDeploy(
+          Tools.rng(parents.head.blockHash.toByteArray)
+        )
+        unsignedBlock <- if (deploys.nonEmpty || slashingDeploys.nonEmpty) {
```
zsluedem, line 79: There is something wrong with closeBlock when merging now. I think we should leave this the old way now.

tgrospic: Close block is always in the list of `systemDeploys` so if there is no user `deploys`, block is created with only close block deploy, and error is throw that block is empty. I'm not sure if slashing deploy can be alone in the block.

zsluedem: > Close block is always in the list of `systemDeploys` so if there is no user `deploys`, block is created with only close block deploy, and error is throw that block is empty. I'm not sure if slashing deploy can be alone in the block.

Ah. You are right. I mislooked.

> slashing deploy can be alone in the block

I am not sure what you mean about this sentence. 

BTW, we should check if the slashing deploy would cause merge conflicts also.

tgrospic: This condition `if (deploys.nonEmpty || slashingDeploys.nonEmpty)` allows block to be created without user deploys, slashing deploys are checked like user deploys.

casper/src/main/resources/PoS.rhox
```diff
@@ -63,626 +65,675 @@ in {
...
+                            match coopVault {
+                              (_, coopRevAddress, _) => {
+                                new toBeSlashed, valBondCh, valRewardCh in {
+                                  // Flow of logic:
+                                  // 0. check if blockHash is the hash of an invalid block
+                                  // 1. accumulateDeposits
+                                  // 2. commitRewards
+                                  // 3. transfer slashed funds to Coop
+                                  // 4. update state
+                                  // Retrieves either public key of sender of invalid block or the deployer public key.
```
zsluedem, line 369: very good comments!!

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
Isaac-DeFrain: Old way is fine with me. It seems to be working properly.

casper/src/main/resources/RevVault.rho
```diff
@@ -83,6 +83,15 @@ in {
                 }
               }
             } |
+            // Used during genesis to create PoS vault with initial balance = sum of initial bonds.
+            for (@"createWithBalance", @ownerRevAddress, @amount, ret <- RevVault) {
```
zsluedem, line 87: I don't understand the grammar here `ret <- RevVault`. Is that a way to make this method can only be called when it is in genesis block?

9rb (approved):

casper/src/main/resources/RevVault.rho
Isaac-DeFrain: That's what I was going for. It's a linear receive so it will only be used the first time it's called. `ret` is the return channel supplied by the corresponding produce in the PoS contract.

PR #2852, opened 2020-01-20
Rchain 4006 close block system deploy

zsluedem:
## Overview
add closeBlock system deploy and add the closeBlock systemDeploy into one of the process to create a block

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4006

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -11,7 +11,7 @@ import coop.rchain.casper.CasperState.CasperStateCell
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.util.ProtoUtil._
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
-import coop.rchain.casper.util.rholang.costacc.SlashDeploy
+import coop.rchain.casper.util.rholang.costacc.{CloseBlockDeploy, SlashDeploy}
 import coop.rchain.casper.util.rholang.{SystemDeploy, _}
 import coop.rchain.casper.util.{DagOperations, ProtoUtil}
 import coop.rchain.crypto.PublicKey
@@ -49,7 +49,6 @@ object BlockCreator {
   )(implicit spanF: Span[F]): F[CreateBlockStatus] =
     spanF.trace(CreateBlockMetricsSource) {
       import cats.instances.list._
-
       val validator = ByteString.copyFrom(validatorIdentity.publicKey.bytes)
       for {
         tipHashes             <- Estimator[F].tips(dag, genesis)
@@ -73,7 +72,13 @@ object BlockCreator {
         now              <- Time[F].currentMillis
         invalidBlocksSet <- dag.invalidBlocks
         invalidBlocks    = invalidBlocksSet.map(block => (block.blockHash, block.sender)).toMap
-        unsignedBlock <- if (deploys.nonEmpty || slashingDeploys.nonEmpty) {
+        // make sure closeBlock is the last system Deploy
+        systemDeploys = if (slashingDeploys.nonEmpty)
+          slashingDeploys ++ Seq(
+            CloseBlockDeploy(Tools.rng(parents.head.blockHash.toByteArray))
+          )
+        else List.empty[SystemDeploy]
+        unsignedBlock <- if (deploys.nonEmpty || systemDeploys.nonEmpty) {
                           SynchronyConstraintChecker[F]
                             .check(dag, runtimeManager, genesis, validator)
                             .ifM(
@@ -89,7 +94,7 @@ object BlockCreator {
                                     runtimeManager,
                                     parents,
                                     deploys,
-                                    slashingDeploys,
+                                    systemDeploys,
                                     justifications,
                                     maxBlockNumber,
                                     validatorIdentity.publicKey,
@@ -189,7 +194,7 @@ object BlockCreator {
       invalidBlocks: Map[BlockHash, Validator]
   ): F[CreateBlockStatus] =
     (for {
-      blockData <- BlockData(now, maxBlockNumber + 1, sender).pure
+      blockData <- BlockData(now, maxBlockNumber + 1, sender, parents.head.blockHash.toByteArray).pure
       result <- InterpreterUtil.computeDeploysCheckpoint(
                  parents,
                  deploys,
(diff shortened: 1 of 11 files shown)


tymm:
While this should work, it would be better to call `closeBlock` in the same fashion as we do it with `reundDeploy` and `chargeDeploy`. This is because then we can avoid putting close-block system deploys into blocks.

zsluedem:
> While this should work, it would be better to call `closeBlock` in the same fashion as we do it with `reundDeploy` and `chargeDeploy`. This is because then we can avoid putting close-block system deploys into blocks.

Your point sounds ok. I have a question for you on the system deploy. What `initialRand ` should I pick in the `closeBlock` system deploy? I think it is the key to make the replay with the same state. Am I right?

marcin-rzeznicki:
@zsluedem It doesn't really matter as long as you use it consistently between play and replay. So it must be something that is available and unchanged between these two. We've used `Tools.rng(deploy.sig.toByteArray)` for this.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -150,11 +150,21 @@ message ProcessedDeployProto {
...
+message CloseBlockSystemDeployDataProto{
+}
+
+message SystemDeployDataProto{
+  oneof systemDeploy{
+    SlashSystemDeployDataProto slashSystemDeploy = 1;
+    CloseBlockSystemDeployDataProto closeBlockSystemDeploy = 2;
+  }
+}
+
```
tgrospic, line 167: This is not needed anymore, right?

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -176,6 +187,19 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
...
+        new CloseBlockDeploy(
+          initialRand = Blake2b512Random(Array(0.toByte))
+        ),
+        new CloseBlockDeploy(
+          initialRand = Blake2b512Random(Array(0.toByte))
+        )
+      )(_ => true)
+    }
+  }
+
```
tgrospic, line 202: Can you please add a test which shows failed execution of close block. For now just a seed mismatch should be fine. It would be nice to have more complicated cases with other failures in PoS.

tgrospic (commented):
LGTM, just a leftover of proto changes and additional test for failure case.

models/src/main/protobuf/CasperMessage.proto
zsluedem: No. It is still needed. Otherwise where should I put the `evenLogs`?

zsluedem: If I used the old `SystemDeployDataProto` , it is actually `SlashSystemDeploy` which require `invalidBlockHash`

zsluedem:
@tgrospic Still after a second thought in the morning, the way you describe yesterday -> doing the updating process after some block height with protobuf does not require hard fork, too. Protobuf here is never a problem when we do an upgrade.

I think we are both turning around and around here. You are trying to persuade me not adding anything in protobuf and I am trying to tell you the upgrade without hard fork is nothing about protobuf. Anyway, I will take the way you propose. I just want to clear protobuf is not the problem. It is actually the code structure matter.

models/src/main/protobuf/CasperMessage.proto
tgrospic: OK, I see it's needed just as a marker in protobuf to distinguish between _slash_ and _close block_ types.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -464,63 +476,71 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
...
+        failureOption <- deployEvaluator
+                          .semiflatMap { evalResult =>
+                            runtime.replaySpace
+                              .createSoftCheckpoint()
+                              .whenA(evalResult.succeeded)
+                          }
+                          .flatMap { _ =>
+                            /* This deployment represents either correct program `Some(result)`,
+or we have a failed pre-charge (`None`) but we agree on that it failed.
+In both cases we want to check reply data and see if everything is in order */
```
tgrospic: Comments indentation is not lined up.

tgrospic:
Only changes to protobuf that are problematic are reordering and removing fields. We can overcome this by just adding new fields and not use the old ones.
So you are right, protobuf is not an issue, but the way how we interpret these new fields or just change meaning of existing ones is what can cause hard-fork. The code must be aware of the block height where the rules have changed.

E.g. if in version, let's called BTC, you have parent block hash as a random seed. And you have new version called BCH, which includes all parents as a seed. Storing this seed in a proto message doesn't make any difference. BTC clients cannot process (create/validate) new messages and BCH clients must know from which block height the change is valid/required.

9rb:
bors try

tgrospic (approved):
Very nice work!

Isaac-DeFrain (approved):
lgtm as long as it passes the RevVaultSpec test

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -73,7 +72,13 @@ object BlockCreator {
         now              <- Time[F].currentMillis
         invalidBlocksSet <- dag.invalidBlocks
         invalidBlocks    = invalidBlocksSet.map(block => (block.blockHash, block.sender)).toMap
-        unsignedBlock <- if (deploys.nonEmpty || slashingDeploys.nonEmpty) {
+        // make sure closeBlock is the last system Deploy
+        systemDeploys = if (slashingDeploys.nonEmpty)
+          slashingDeploys ++ Seq(
+            CloseBlockDeploy(Tools.rng(parents.head.blockHash.toByteArray))
```
nzpr, line 78: @zsluedemI if several blocks have the same main parent (e.g. the very first blocks that all have genesis as a parent) - CloseBlockDeploy deploy will be initialised with the same seed. Isn't this a problem?

zsluedem: I actually already had a fix for that. But Rao said we should be careful about it. So I didn't bring up.

PR #2855, opened 2020-01-22
Rchain 4006 close block system deploy v2

zsluedem:
draft one


zsluedem:
@tgrospic The `closeBlock` return None here. https://github.com/rchain/rchain/pull/2855/files#diff-8172c898e2d53ee886672b92b2e248c9R249

tgrospic:
Do you see this only on reply or also when in play?

zsluedem:
> Do you see this only on reply or also when in play?

only in replay.

forget to mentioin about the test ->`testOnly coop.rchain.casper.util.rholang.RuntimeManagerTest`

PR #2866, opened 2020-02-03
Move block creation constraints checker to BlockAPI

nzpr:
## Overview
<sup>_Currently node checks for constraints in BlockCreator, which means lots of computing happens before we check if node should create block at all. This PR moves checker closer to GRPC endpoint, inside BlockAPI._</sup>

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
@@ -136,16 +136,19 @@ class CreateBlockAPITest extends FlatSpec with Matchers with EitherValues {
     }
   }
 
+  val syncConstraintThreshold = 1d / 3d
+
   it should "not allow proposals without enough new blocks from other validators" in effectTest {
     TestNode
-      .networkEff(genesis, networkSize = 5, synchronyConstraintThreshold = 1d / 3d)
+      .networkEff(genesis, networkSize = 5, synchronyConstraintThreshold = syncConstraintThreshold)
       .use {
         case nodes @ n1 +: n2 +: _ +: _ +: _ +: Seq() =>
           import n1.{logEff, metricEff, span, timeEff}
-          val engine                              = new EngineWithCasper[Task](n1.casperEff)
-          implicit val blockStore                 = n1.blockStore
-          implicit val lastFinalizedStorage       = n1.lastFinalizedStorage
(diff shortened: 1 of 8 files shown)


tgrospic (approved):
Nice work!

I updated Casper to have validator field as `Option[PublicKey]`. Checking if validator key is defined is then much more clear than checking for empty array. E.g. in [exploratory deploys](https://github.com/rchain/rchain/pull/2846/files#diff-2c412bf643d2b8a4b11e208d483e8710R618).

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -124,17 +142,22 @@ class CreateBlockAPITest extends FlatSpec with Matchers with EitherValues {
...
-            _  <- n1.syncWith(n2)
+            b1                         <- n1.publishBlock(deploys(0))(nodes: _*)
+            b2                         <- n2.publishBlock(deploys(1))(nodes: _*)
+            _                          <- n1.syncWith(n2)
 
-            _        <- n1.casperEff.addDeploy(deploys(2))
+            _ <- n1.casperEff.addDeploy(deploys(2))
             b3Status <- createBlock(blockApiLock)(engineCell)
 
             _ = b3Status.left.value should include(
```
tgrospic: @nzpr this test is now failing. Is this expected?

nzpr: Thanks for fixing this! I guess I overlooked it.

tgrospic: I didn't look into test. What I meant is, with your changes should this test be adjusted or it's a newly created bug?

nzpr: test should be adjusted - `SynchronyConstraintChecker[Effect](0)` is 0 here, but it should be set to corresponding value https://github.com/rchain/rchain/blob/4997ae2fe829a7c7f199136542ea135e995be6f6/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala#L141 in all tests. After I fixed this, now only the last test fails, looking into it.

nzpr: I will add ignore this test as now we are checking for constraints before check for deploys.

zsluedem (approved):
LGTM now

PR #2871, opened 2020-02-06
RCHAIN-4020 input password when running node

zsluedem:
## Overview
ask user to input their password of the rnode key file

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4020

### Notes

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -7,6 +7,7 @@ import cats.effect.Resource
 import coop.rchain.casper.util.comm._
 import coop.rchain.catscontrib.TaskContrib._
 import coop.rchain.crypto.PrivateKey
+import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.signatures.{Secp256k1, SignaturesAlg}
 import coop.rchain.crypto.util.KeyUtil
 import coop.rchain.node.configuration._
@@ -24,7 +25,6 @@ import scala.tools.jline.console._
 import scala.tools.jline.console.completer.StringsCompleter
 
 object Main {
-  private val RNodeDeployerPasswordEnvVar = "RNODE_DEPLOYER_PASSWORD"
 
   implicit private val logSource: LogSource     = LogSource(this.getClass)
   implicit private val log: Log[Task]           = effects.log
@@ -56,6 +56,14 @@ object Main {
       .traverse(k => log.warn(s"Unknown configuration key $k"))
       .void
 
+  private def decryptKeyFromCon(
+      encryptedPrivateKeyPath: Path
+  )(implicit console: ConsoleIO[Task]): Task[PrivateKey] =
+    for {
+      password   <- ConsoleIO[Task].readPassword("Password for private key file: ")
+      privateKey <- Secp256k1.parsePemFile[Task](encryptedPrivateKeyPath, password)
+    } yield privateKey
+
   private def mainProgram(conf: Configuration)(implicit scheduler: Scheduler): Task[Unit] = {
     implicit val replService: GrpcReplClient =
       new GrpcReplClient(
@@ -78,8 +86,6 @@ object Main {
 
     implicit val time: Time[Task] = effects.time
 
-    implicit val envVars: EnvVars[Task] = EnvVars.envVars
-
     implicit val console: ConsoleIO[Task] = consoleIO
 
     val program = conf.command match {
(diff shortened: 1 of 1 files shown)


nzpr (approved):
LGTM

node/src/main/scala/coop/rchain/node/configuration/hocon/model.scala
```diff
@@ -323,12 +323,11 @@ object Casper {
     import Configuration.RichConfig
 
     val casper = config.getConfig(Key)
-    val pkPath = casper.getPathOpt(keys.ValidatorPrivateKeyPath).map(Right(_))
-    val pk     = casper.getStringOpt(keys.ValidatorPrivateKey).map(Left(_))
+    val pk     = casper.getStringOpt(keys.ValidatorPrivateKey)
```
Isaac-DeFrain: I know this is nit picking, but I feel like it's always better to call the private key "sk" and public key "pk"

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -80,9 +85,21 @@ object Configuration {
     val optionsConfig = ConfigMapper.fromOptions(options)
     val fileConfig    = if (command == Run) loadConfigFile(configFile) else ConfigFactory.empty()
 
+    // check if validator private key path arg is specific and decrypt
```
Isaac-DeFrain: *specified (not specific)

Isaac-DeFrain (approved):
I left a few minor comments. Nice work!

tgrospic:
Can you use the same logic as here?
https://github.com/rchain/rchain/blob/226ffa35/node/src/main/scala/coop/rchain/node/Main.scala#L108-L113

casper/src/main/scala/coop/rchain/casper/CasperConf.scala
```diff
@@ -17,7 +17,7 @@ import scala.util.{Failure, Success, Try}
 
 final case class CasperConf(
     publicKeyBase16: Option[String],
-    privateKey: Option[Either[String, Path]],
+    privateKey: Option[String],
```
tgrospic: Is this means that private key path can only be specified as CLI parameter but not in config file? Am I missing something?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -34,9 +34,11 @@ object Configuration {
   private val CertificateConfigKey = s"${hocon.Tls.Key}.${hocon.Tls.keys.Certificate}"
   private val KeyConfigKey         = s"${hocon.Tls.Key}.${hocon.Tls.keys.Key}"
   private val GenesisPathConfigKey = s"${hocon.Casper.Key}.${hocon.Casper.keys.GenesisPath}"
+  private val PrivateKeyPathConfigKey =
+    s"${hocon.Casper.Key}.${hocon.Casper.keys.ValidatorPrivateKey}"
```
tgrospic: Variable here says key **path**, but from casper config you are using  `ValidatorPrivateKey` and not `ValidatorPrivateKeyPath`.

node/src/main/scala/coop/rchain/node/configuration/hocon/model.scala
tgrospic: If _pk_ is alone I'm lost. :wink:

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -37,20 +39,29 @@ object Main {
...
-    val configuration = Configuration.build(args)
-    System.setProperty("rnode.data.dir", configuration.server.dataDir.toString) // NonUnitStatements
-
     implicit val scheduler: Scheduler = Scheduler.computation(
       Math.max(java.lang.Runtime.getRuntime.availableProcessors(), 2),
       "node-runner",
       reporter = UncaughtExceptionLogger
     )
 
+    val configuration = Configuration.build(args, decryptPrivateKey(console))
```
tgrospic: This is main entry to RNode. Is this going to ask for password even on commands that don't need password.

Maybe it's better to push this check and console input inside Run command.

casper/src/main/scala/coop/rchain/casper/CasperConf.scala
zsluedem: I have to check if the config file is ok. I tried the command line.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
zsluedem: The configuration is a `final case class` and the `privateKey`  is stored in conf. Once we build the configuration we can not change it.
I actually ask for the password while building the configuration. 
If the `ValidatorPrivateKeyPath ` is set, decrypt it and then set the `ValidatorPrivateKey` in the config. Otherwise, it would read straight from `ValidatorPrivateKey` option.
The `ValidatorPrivateKey` option is hidden which only should be used for the tests currently.
It is not an elegant way to implement it, I agree.

node/src/main/scala/coop/rchain/node/Main.scala
zsluedem: I was trying to avoid putting `ConsoleIO`  into other packages which ends up with  not very nice organized code. I'll see if there are other ways to do that.

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -78,7 +86,7 @@ object Main {
 
     implicit val time: Time[Task] = effects.time
 
-    implicit val envVars: EnvVars[Task] = EnvVars.envVars
+//    implicit val envVars: EnvVars[Task] = EnvVars.envVars
```
tgrospic: Leftover.

tgrospic (approved):
Excellent!

PR #2872, opened 2020-02-06
Fix UnusedCOMM caused by system deploys execution

nzpr:
## Overview
DeploysResult computed under runtimeLock (inside computeState method), and all system deploys are calculated separately. And if RSpace is modified between, e.g. if node receives block from a peer - this corrupts calculation.


nzpr:
Not tested yet, and looks like I overlooked how system deplys are being handled
https://github.com/rchain/rchain/blob/68e640047dce997787695ceb753076a51af58f7d/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala#L124

zsluedem:
I just want to say that https://github.com/rchain/rchain/pull/2838 this pr makes the node process more messages concurrently which helps pop the problem into the surface. I am concerned that when the network becomes large that the lock of runtime would block everything.

9rb (approved):
I approve subject to all the tests passing

tgrospic (approved):

PR #2875, opened 2020-02-06
RCHAIN-3989 Use TreeHashMap in RevVault contract

Isaac-DeFrain:
## Overview
Uses TreeHashMap in the RevVault contract instead of an ordinary Rholang map.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3989

### DIFF
--- a/casper/src/main/resources/RevVault.rho
+++ b/casper/src/main/resources/RevVault.rho
@@ -16,10 +16,11 @@
  */
 new
   RevVault, rs(`rho:registry:insertSigned:secp256k1`), uriOut,
-  rl(`rho:registry:lookup`), RevAddress(`rho:rev:address`), MakeMintCh, AuthKeyCh, EitherCh,
+  rl(`rho:registry:lookup`), RevAddress(`rho:rev:address`),
+  MakeMintCh, AuthKeyCh, EitherCh, TreeHashMapCh,
   _makeVault,
   _newVault,
-  _getOrCreate,
+  _create,
   _revVault,
   _transferTemplate,
   _depositTemplate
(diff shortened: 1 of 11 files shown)


casper/src/main/resources/RevVault.rho
```diff
@@ -27,255 +28,242 @@ in {
...
-                                for (@eitherVault <- resCh) {
-                                  ret!(eitherVault) |
-                                  match eitherVault {
-                                    (true, vault) => vaultMapStore!(vaultMap.set(revAddress, vault))
-                                    (false, _)    => vaultMapStore!(vaultMap)
-                                  }
+            }
+          } |
+
+          contract _getOrCreate(@revAddress, constructor, retCh, ackCh) = {
```
tgrospic: Why is needed this additional _ackCh_ when there is already _retCh_?

casper/src/main/resources/RevVault.rho
```diff
@@ -27,255 +28,242 @@ in {
...
+                      for (@vault <- vaultCh) {
+                        match vault {
+                          Nil => {
+                            // The vault was not found, create a new one and set it in the vaultMap.
+                            new resCh in {
+                              constructor!(*resCh) |
+                              for (@eitherVault <- resCh) {
+                                retCh!(eitherVault) |
+                                match eitherVault {
+                                  (true, vault) => { TreeHashMap!("set", vaultMap, revAddress, vault, *ackCh) }
```
tgrospic: Vault is returned before it's `set` to vaults map. Can we just wait on TreeMap result and then return vault on `retCh`?

Also `get` on TreeMap is not blocking. I'm not sure if we need more synchronization. Maybe `update` would be better here? Between get and set other thread can create new vault for the same address.

casper/src/main/resources/RevVault.rho
Isaac-DeFrain: You're right! It's better to create a new _ackCh_ inside of __getOrCreate_ and consume the acknowledgement there. Thanks!

casper/src/main/resources/RevVault.rho
Isaac-DeFrain: I changed the the vault return to occur after the acknowledgement from `set`.

tgrospic:
It's strange that deploy signature test failed. I've re-run the tests, on the first attempt this test passed.
```
[info] - should be valid signature algorithm to sign a deploy *** FAILED *** (58 milliseconds)
[info]   Left("Invalid signature") was not an instance of scala.util.Right, but an instance of scala.util.Left (DeployValidationSpec.scala:36)
```

Isaac-DeFrain:
I rebased on current dev. We'll see if the signature test passes now.

casper/src/main/resources/RevVault.rho
```diff
@@ -16,266 +16,262 @@
...
-                }
+  rl!(`rho:lang:treeHashMap`, *TreeHashMapCh) |
+  for (@(_, MakeMint) <- MakeMintCh;
+       @(_, AuthKey)  <- AuthKeyCh;
+       @(_, Either)   <- EitherCh;
+       TreeHashMap    <- TreeHashMapCh) {
+    // Each invocation of the RevVault contract uses a different mint.
+    new mintCh, vaultMapStore, initVault in {
+      // Initializes the vaultMapStore.
+      TreeHashMap!("init", 3, *vaultMapStore) |
```
ghost: Should probably use more than a depth-3 tree for the vault map, though I suppose it could be changed later with an RCIP if there's too much contention.

Isaac-DeFrain: What depth do you suggest using?

ghost: I'd make it at least 4.

9rb (approved):
Please merge after rebasing on current dev and all tests pass

casper/src/main/resources/RevVault.rho
Isaac-DeFrain: Would there be a significant performance hit if it was made to be depth-4 and this depth wasn't necessary? For example, compared to depth-3?

ghost: I was operating under the assumption (which is silly in hindsight) that the vaultMapStore would be updated whenever there was a payment. 3 is plenty; even 1 would suffice for foreseeable usage levels.

9rb: Based on the above, it looks like we should certainly make this a tunable parameter, with a reasonable default value that would give decent performance.

tgrospic (approved):
LGTM

casper/src/main/resources/RevVault.rho
tgrospic: Is PoS situation with pending rewards is even simpler than with REV vaults. @pyrocto should it be enough to set depth to 1 also?
https://github.com/rchain/rchain/blob/46b9acfdd3e2706a099a79ce30dd7149af3c0538/casper/src/main/resources/PoS.rhox#L106-L108

ghost: Sure.

zsluedem:
bors try

Isaac-DeFrain:
bors try

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
```diff
@@ -86,11 +86,16 @@ object StandardDeploys {
       1559156420651L
     )
 
-  def revGenerator(vaults: Seq[Vault], supply: Long): Signed[DeployData] =
+  def revGenerator(
+      vaults: Seq[Vault],
+      supply: Long,
```
tgrospic, line 91: What is the purpose of `supply`? It's not used anywhere.

Isaac-DeFrain: I'm not really sure TBH. It used set to `Long.MaxValue` everywhere I see it used.

Isaac-DeFrain: My assumption is that it was once used for something, but is now deprecated.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -9,50 +9,42 @@ final class RevGenerator private (supply: Long, code: String)
...
+         #       @RevVault!("init", *revVaultInitCh) |
+         #       for (TreeHashMap, @vaultMap, initVault, initContinue <- revVaultInitCh) {
+         #         match [$vaultBalanceList] {
+         #           vaults => {
+         #             new iter in {
+         #               contract iter(@[(addr, initialBalance) ... tail]) = {
+         #                  iter!(tail) |
+         #                  new vault in {
+         #                    initVault!(*vault, addr, initialBalance) |
+         #                    TreeHashMap!("set", vaultMap, addr, *vault, Nil)
```
ghost: RSpace doesn't discard messages sent on Nil.  It would be better to create a channel and consume the response.

ghost (approved):

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -26,21 +26,32 @@ object Genesis {
...
-      StandardDeploys.revGenerator(vaults, supply),
+    StandardDeploys.registry +:
+      StandardDeploys.listOps +:
+      StandardDeploys.either +:
+      StandardDeploys.nonNegativeNumber +:
+      StandardDeploys.makeMint +:
+      StandardDeploys.authKey +:
+      StandardDeploys.revVault +:
+      StandardDeploys.multiSigRevVault +:
+      vaultDeploys :+
```
Isaac-DeFrain, line 52: vaultDeploys is the new and improved RevGenerator!

tgrospic: It's really a list of REV generator deploys, each creates 100 vaults.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
```diff
@@ -9,50 +9,43 @@ final class RevGenerator private (supply: Long, code: String)
...
+         #       for (TreeHashMap, @vaultMap, initVault, initContinue <- revVaultInitCh) {
+         #         match [$vaultBalanceList] {
+         #           vaults => {
+         #             new iter in {
+         #               contract iter(@[(addr, initialBalance) ... tail]) = {
+         #                  iter!(tail) |
+         #                  new vault, setDoneCh in {
+         #                    initVault!(*vault, addr, initialBalance) |
+         #                    TreeHashMap!("set", vaultMap, addr, *vault, *setDoneCh) |
+         #                    for (_ <- setDoneCh) { Nil }
```
Isaac-DeFrain, line 36: Nice avoidance of the bloat

Isaac-DeFrain (commented):
I'm assuming the iteration over the list of vaults is more performant than any of the ListOps

casper/src/main/scala/coop/rchain/casper/genesis/contracts/RevGenerator.scala
tgrospic: Thanks to @pyrocto :)

tgrospic:
> I'm assuming the iteration over the list of vaults is more performant than any of the ListOps

It's not a big difference but I got better results.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
tgrospic: We will address this later.

PR #2877, opened 2020-02-09
RCHAIN-4026: fix RSpace replay - sorted produces in COMM event log, match joins on the same channels

tgrospic:
## Overview
- Fixes bug in replay with joins on the same channels. Order of produces in COMM event log is now sorted.
- Fixes bug with missing produce when getting data on matching joins.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4026

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
@@ -1,0 +1,140 @@
+package coop.rchain.rholang.interpreter
+
+import cats.effect.Resource
+import cats.syntax.all._
+import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.metrics
+import coop.rchain.metrics.{Metrics, NoopSpan, Span}
+import coop.rchain.rholang.Resources
+import coop.rchain.rholang.interpreter.accounting.utils.costLog
+import coop.rchain.rholang.interpreter.accounting.{_cost, Cost, CostAccounting}
+import coop.rchain.rspace.SoftCheckpoint
+import coop.rchain.shared.Log
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.duration._
+
+class ReplaySpec extends FlatSpec with Matchers {
+
+  // TODO: these tests are temporary and specific to bugs found in replay.
+  // Testing execution for many iteration doesn't make sense.
+  // In tests, nondeterministic execution of Par in tuple space should be replaced with deterministic version
+  // so that we can test tuple space access in all possible states (and not guess and wait).
+
+  // TODO: fuzzer test now creates complete on-disk tuple space for each execution, this can be replaced
+  //  with in-memory version used here.
+  // https://github.com/rchain/rchain/blob/1f9554f68a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala#L231
+
+  "multi joins (2/2)" should "execute successfully in replay" in {
+    val term =
+      """
+        |new x in {
+        |  x!() | for(<- x; <- x) { 0 } | x!()
+        |}
+        |""".stripMargin
+    testRholangTerm(term, 500, 30.seconds)
+  }
+
+  "multi joins (4/2)" should "execute successfully in replay" in {
+    val term =
+      """
+        |new x in {
+        |  x!() | x!() | for(<- x; <- x) { 0 } | x!() | x!()
+        |}
+        |""".stripMargin
+    testRholangTerm(term, 500, 30.seconds)
+  }
+
+  def testRholangTerm(term: String, iterations: Int, timeout: Duration) =
+    withRSpaceAndRuntime { runtime =>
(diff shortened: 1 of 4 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
```diff
@@ -0,0 +1,123 @@
...
+import coop.rchain.shared.Log
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.duration._
+
+class ReplaySpec extends FlatSpec with Matchers {
+
+  it should "not fail with unused COMM events ???" in {
```
Isaac-DeFrain: Is this test passing now that you have ordered the produces by channel hash?

tgrospic: @Isaac-DeFrain this is a new test that reveals a new bug when produces happens after consume e.g. `x!() | x!() | for(<- x; <- x) { 0 }`. If `x!()` happens after consume (or both sends), replay will save sends and it will not find COM event.

Ordering bug is more likely to happen than this new bug, so this test would detect ordering bug also.

9rb (approved):
Please complete additional changes as a separate PR. Thanks

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    val term =
+      """
+        |new x in {
+        |  x!() | x!() | for(<- x; <- x) { 0 } | x!() | x!()
+        |}
+        |""".stripMargin
+    testRholangTerm(term, 500, 30.seconds)
+  }
+
+  def testRholangTerm(term: String, iterations: Int, timeout: Duration) =
```
Isaac-DeFrain, line 50: Very nice!

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
```diff
@@ -0,0 +1,131 @@
...
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.duration._
+
+class ReplaySpec extends FlatSpec with Matchers {
+
+  "multi joins (2/2)" should "execute successfully in replay" in {
+    val term =
```
Isaac-DeFrain, line 31: Have you tested situations with different types of sends and receives? Persistent/nonpersistent/peek?

rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
```diff
@@ -57,7 +54,9 @@ object COMM {
       peeks: SortedSet[Int],
       produceCounters: (Seq[Produce]) => Map[Produce, Int]
   ): COMM = {
-    val produceRefs = dataCandidates.map(_.datum.source)
+    val produceRefs =
+      dataCandidates.map(_.datum.source).sortBy(p => (p.channelsHash, p.hash, p.persistent))
```
Isaac-DeFrain, line 58: Does this solve the replay issue witnessed in comm events with both persistent and nonpersistent sends/receives?

Isaac-DeFrain (approved):
Very nice work! I left a few comments/questions.

Have you tested this with more complicated examples? Maybe with nesting?

Isaac-DeFrain:
bore r+

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
tgrospic: @Isaac-DeFrain I added comments for these tests to better explain the purpose. I see it as temporary just until all bugs with replay are resolved. This 500 iteration lasts about 10sec which is waste of time. We need better way to control execution of _Par_ in tests.

rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
tgrospic: Produces are sorted by all fields from `Produce` so it should work with persistent and nonpersistent sends/receives.
https://github.com/rchain/rchain/blob/79060b3fa15b372e5ed233e845a01a7528228597/rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala#L69-L73

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReplaySpec.scala
tgrospic: One unfortunate thing, although it's completely executed in-memory, it cannot be run in parallel. First because of _Cost_ instance and also because of _replayData_ in `IReplaySpace`.

But if we could make snapshot of the whole `Runtime`, we could make a fork and execute terms in parallel from the same parent point.
https://github.com/rchain/rchain/blob/f7e46a9346990010bf32422f0c1abee17a1f85c5/rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala#L31-L39

PR #2878, opened 2020-02-10
Fixes contention bug and adds test for the bug.

ghost:
## Overview
Quantifier [discovered a bug](https://discordapp.com/channels/391435113415573504/420661762828926978/675797443660480543) in TreeHashMap.  I replicated the bug and found it was due to a get-then-update rather than an atomic update.  This fixes that bug.

### DIFF
--- a/casper/src/main/resources/Registry.rho
+++ b/casper/src/main/resources/Registry.rho
@@ -13,7 +13,7 @@ new _registryStore,
 in {
 
   // TreeHashMap is defined here because the registry uses it internally.
-  // We can't define it in another file like MakeMint or NonNegativeNumber because 
+  // We can't define it in another file like MakeMint or NonNegativeNumber because
   // that method assumes the registry already exists.
 
   // Rholang map desiderata: speedy insert & lookup,  no conflicts on lookup, no conflicts on inserts to different keys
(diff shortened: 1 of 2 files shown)


casper/src/main/resources/Registry.rho
```diff
@@ -188,28 +152,64 @@ in {
...
-                  // Node missing
-                  // Create node, set node to 0
-                  MakeNode!(0, (map, nybList.slice(0, n + 1))) |
-                  new ack in {
-                    // Atomically update current node to val | (1 << nybList.nth(n))
-                    nodeSetBit!((map, nybList.slice(0, n)), nybList.nth(n), *ack) |
-                    for (_ <- ack) {
+            match (map, nybList.slice(0, n)) {
+              node => {
+                for (@val <<- @[node, *storeToken]) {
```
Isaac-DeFrain, line 157: Nice. Now the lock is only acquired if there's a value present at the node. Cool.

Isaac-DeFrain (approved):
LGTM

tgrospic (approved):
Excellent! With this fix PoSTest in @Isaac-DeFrain PR #2875 now passes.

PR #2879, opened 2020-02-12
RCHAIN-4014 add SystemDeployUtil, change closeBlockDeploy random seed, remove

zsluedem:
## Overview
1. add SystemDeployUtil to generate random seed for systemDeploy. The logic of `PreCharge` and `Refund` doesn't change. I only change `CloseBlock` and `Slashing`
2. replace the `parentHash` with `seqNum`  in the `BlockData`. 

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4014

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -331,7 +331,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       Span[F].trace(computeGenesisLabel) {
         for {
           _ <- runtime.blockData.set(
-                BlockData(blockTime, 0, PublicKey(Array[Byte]()), Array[Byte]())
+                BlockData(blockTime, 0, PublicKey(Array[Byte]()))
               )
           _          <- Span[F].mark("before-process-deploys")
           evalResult <- processDeploys(runtime, startHash, terms, processDeploy(runtime))
@@ -359,7 +359,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                      terms,
                      systemDeploys,
                      replayDeploy(runtime, withCostAccounting = !isGenesis),
-                     replaySystemDeploy(runtime, blockData.parentHash)
+                     replaySystemDeploy(runtime, terms)
                    )
         } yield result
       }
@@ -386,7 +386,6 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       runtime: Runtime[F]
   )(deploy: Signed[DeployData])(implicit Log: Log[F]) = {
     import cats.instances.vector._
-    val rand = Tools.rng(deploy.sig.toByteArray)
     EitherT(
       WriterT(
         Log.info(
@@ -397,7 +396,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
             new PreChargeDeploy(
               deploy.data.totalPhloCharge,
               deploy.pk,
-              rand
+              SystemDeployUtil.generatePreChargeDeployRandomSeed(deploy)
             )
           )
       )
@@ -413,7 +412,10 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                 s"Refunding ${Base16.encode(deploy.pk.bytes)} with ${pd.refundAmount}"
               ) >>
                 playSystemDeployInternal(runtime)(
-                  new RefundDeploy(pd.refundAmount, rand.splitByte(64))
+                  new RefundDeploy(
+                    pd.refundAmount,
+                    SystemDeployUtil.generateRefundDeployRandomSeed(deploy)
+                  )
                 )
             )
           ).leftSemiflatMap(
@@ -546,17 +548,23 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
   }
   private def replaySystemDeploy(
       runtime: Runtime[F],
-      parentHash: Array[Byte]
+      deploys: Seq[ProcessedDeploy]
   )(processedSystemDeploy: ProcessedSystemDeploy): F[Option[ReplayFailure]] = {
     import processedSystemDeploy._
 
     systemDeploy match {
       case SlashSystemDeployData(invalidBlockHash, issuerPublicKey) =>
         val slashDeploy =
-          SlashDeploy(invalidBlockHash, issuerPublicKey, Tools.rng(invalidBlockHash.toByteArray))
+          SlashDeploy(
+            invalidBlockHash,
+            issuerPublicKey,
+            SystemDeployUtil.generateSlashDeployRandomSeed(deploys.map(_.deploy))
+          )
         replaySystemDeployInternal(runtime, slashDeploy, processedSystemDeploy)
       case CloseBlockSystemDeployData =>
-        val closeBlockDeploy = CloseBlockDeploy(Tools.rng(parentHash))
+        val closeBlockDeploy = CloseBlockDeploy(
+          SystemDeployUtil.generateCloseDeployRandomSeed(deploys.map(_.deploy))
+        )
         replaySystemDeployInternal(runtime, closeBlockDeploy, processedSystemDeploy)
       case Empty => ReplayFailure.internalError(new Exception("Expected system deploy")).some.pure
     }
@@ -589,10 +597,13 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
 
     def evaluatorT: EitherT[F, ReplayFailure, Boolean] =
       if (withCostAccounting) {
-        val rand            = Tools.rng(deploy.sig.toByteArray)
         val expectedFailure = processedDeploy.systemDeployError.map(SystemDeployError)
         replaySystemDeployInternal(runtime)(
-          new PreChargeDeploy(deploy.data.totalPhloCharge, deploy.pk, rand),
+          new PreChargeDeploy(
+            deploy.data.totalPhloCharge,
+            deploy.pk,
+            SystemDeployUtil.generatePreChargeDeployRandomSeed(processedDeploy.deploy)
+          ),
           expectedFailure
         ).flatMap(
           _ =>
@@ -608,7 +619,10 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                 .flatTap(
                   succeeded =>
                     replaySystemDeployInternal(runtime)(
-                      new RefundDeploy(refundAmount, rand.splitByte(64)),
+                      new RefundDeploy(
+                        refundAmount,
+                        SystemDeployUtil.generateRefundDeployRandomSeed(processedDeploy.deploy)
+                      ),
                       None
                     ).map(_ => succeeded)
                 )
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -331,7 +331,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       Span[F].trace(computeGenesisLabel) {
         for {
           _ <- runtime.blockData.set(
-                BlockData(blockTime, 0, PublicKey(Array[Byte]()))
+                BlockData(blockTime, 0, PublicKey(Array[Byte]()), 0)
               )
           _          <- Span[F].mark("before-process-deploys")
           evalResult <- processDeploys(runtime, startHash, terms, processDeploy(runtime))
@@ -359,7 +359,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
                      terms,
                      systemDeploys,
                      replayDeploy(runtime, withCostAccounting = !isGenesis),
-                     replaySystemDeploy(runtime, terms)
+                     replaySystemDeploy(runtime, blockData)
                    )
         } yield result
       }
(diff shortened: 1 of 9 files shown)


zsluedem:
After several thoughts on this, I think the case comes to `SlashingDeploy` and `CloseBlock` is a little bit uncertain for me.
Supposed that we have 5 validators -> v1, v2, v3, v4, v5.
If v1 proposes an invalid block b1, v2 receives it and create slashing deploy in b2, v3 also receives it and create slashing deploy in b3 but v3 haven't seen b2 yet. After b3 is proposed, v3 saw b2 and again propose a new block b4 with b2 as parent.Then actually in b4, v1 is slashed double times with merging.
I am not sure if this is the expected outcome.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,47 @@
...
+  // The refund and preCharge system deploy
+  // would use user deploy signature to generate the system deploy. The random seed of
+  // the refund and preCharge has to be exactly the same to make sure replay the user
+  // deploy would come out the exact same result.
+  //
+  // As for closeBlock and slashing, the rnode would execute closeBlock system deploy in every block.
+  // The closeBlock system deploy would use the sender and the seqNum of the validator.
+
+  def generateSystemDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    Tools.rng(sender.toByteArray() ++ sender.toByteArray())
```
tgrospic: `seqNum` is not used as input to create random generator.

Maybe we can use existing `Tools.unforgeableNameRng(deployer: PublicKey, timestamp: Long)`. This leads me to a question, seqNum can eventually be the same as the timestamp so theoretically it's possible to have the same seed as in user deploy. Maybe we should still add prefix to system deploy seed? @zsluedem, @nzpr

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -546,17 +548,24 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
   }
   private def replaySystemDeploy(
       runtime: Runtime[F],
-      parentHash: Array[Byte]
+      deploys: Seq[ProcessedDeploy],
```
tgrospic: `deploys` are not used anymore.

nzpr:
@zsluedem I believe we can check here https://github.com/rchain/rchain/blob/f9ed789ea35c1bdfe8ff6df55f7582c46c531069/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala#L73
if slashing deploy is present inside parents.
I created ticket for that https://rchain.atlassian.net/browse/RCHAIN-4028

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
nzpr: @tgrospic I would add prefix to be sure

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  def generateSlashDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    generateSystemDeployRandomSeed(sender, seqNum).splitByte(2)
+
+  def generatePreChargeDeployRandomSeed(deploy: Signed[DeployData]): Blake2b512Random =
+    Tools.rng(deploy.sig.toByteArray)
+
+  def generateRefundDeployRandomSeed(deploy: Signed[DeployData]): Blake2b512Random =
+    Tools.rng(deploy.sig.toByteArray).splitByte(64)
```
tgrospic: This `64` is like magic number, maybe we can use split index `0` for pre-charge and `1` for refund.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
zsluedem: > `seqNum` is not used as input to create random generator.
> 
> Maybe we can use existing `Tools.unforgeableNameRng(deployer: PublicKey, timestamp: Long)`. This leads me to a question, seqNum can eventually be the same as the timestamp so theoretically it's possible to have the same seed as in user deploy. Maybe we should still add prefix to system deploy seed? @zsluedem, @nzpr

different sender with seqNum for each block would be unique , why would it be same as timestamp? @tgrospic

tgrospic: Validator can new deploy with his own key and with some timestamp e.g. 12345. When seqNum will be at 12345, system deploy will have the same seed as this first deploy.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,47 @@
...
+  // The refund and preCharge system deploy
+  // would use user deploy signature to generate the system deploy. The random seed of
+  // the refund and preCharge has to be exactly the same to make sure replay the user
+  // deploy would come out the exact same result.
+  //
+  // As for closeBlock and slashing, the rnode would execute closeBlock system deploy in every block.
+  // The closeBlock system deploy would use the sender and the seqNum of the validator.
+
+  def generateSystemDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    Tools.rng(sender.toByteArray() ++ Array[Byte](seqNum.toByte))
```
tgrospic: `Array[Byte](seqNum.toByte)` will truncate prefix with this conversion. I'm thinking that we can use protobuf to create fixed length serialization to bytes.
Something like this should work.
```scala
import java.io.{ByteArrayInputStream, ByteArrayOutputStream}

import com.google.protobuf.{CodedInputStream, CodedOutputStream}

def serializeInt64Fixed(value: Long): Array[Byte] = {
  val stream = new ByteArrayOutputStream
  val proto = CodedOutputStream.newInstance(stream)
  proto.writeFixed64NoTag(value)
  proto.flush()
  stream.toByteArray
}

def deserializeInt64Fixed(bytes: Array[Byte]): Long = {
  val stream = new ByteArrayInputStream(bytes)
  val proto = CodedInputStream.newInstance(stream)
  proto.readSFixed64
}

val values = List(1234567890123456789L, 1L)

val valuesSer = values.map(serializeInt64Fixed)
// List(Array(21, -127, -23, 125, -12, 16, 34, 17), Array(1, 0, 0, 0, 0, 0, 0, 0))

val valuesCopy = valuesSer.map(deserializeInt64Fixed)
// List(1234567890123456789, 1)

assert(values == valuesCopy)
```

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
zsluedem: Yes. That can be the case. But the increasing speed of timestamp is different from seqNum. Take a splitByte is not enough?

tgrospic: Random used in user deploy is already splitted by each Par in the Reducer.
https://github.com/rchain/rchain/blob/93451ae8/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala#L186
If unforgeable name created with today's timestamp is stored privately in some contract, future system deploy (with seqNum as today's timestamp) will collide with that previously created unforgeable name and can change/read private info.

zsluedem: Make sense!

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
tgrospic: My concerns are similar as with [ETH signatures, hashes and byte messages](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#signatures-and-hashing-overview).
We need to be careful of which parts the "message" (seed) is constructed. With variable length of messages, length can be included to prevent collisions. We don't need that but we need to take care about the length of each part of the message and total size. This is why I'm suggesting protobuf fixed serializer.

So for a block seed it would be enough to have: `PREFIX ++ PublicKey ++ seqNum` serialized with protobuf. This way we can be completely sure that collision cannot happen. And we can you the same technique if we need seed for other purposes in the future.

zsluedem: ~~Are you sure the bytes would be the same? It seems it is different.  Replay would be wrong in this way.~~

zsluedem: > My concerns are similar as with [ETH signatures, hashes and byte messages](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#signatures-and-hashing-overview).
> We need to be careful of which parts the "message" (seed) is constructed. With variable length of messages, length can be included to prevent collisions. We don't need that but we need to take care about the length of each part of the message and total size. This is why I'm suggesting protobuf fixed serializer.
> 
> So for a block seed it would be enough to have: `PREFIX ++ PublicKey ++ seqNum` serialized with protobuf. This way we can be completely sure that collision cannot happen. And we can you the same technique if we need seed for other purposes in the future.

Your point totally makes sense and good. I was thinking to use `com.google.protobuf.wrappers.Int32Value`.But I am not sure if it is fixed length

tgrospic: With `Int32Value` wrapper you still need to write to OutputStream.
https://github.com/yeriomin/play-store-api/blob/4ec5513/src/main/java/com/google/protobuf/Int64Value.java#L56-L61
I find it easier to use `CodedOutputStream` and `CodedInputStream` directly without the wrapper where any supported type can have its own method.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  private def serializeInt32Fixed(value: Int): Array[Byte] = {
+    val stream = new ByteArrayOutputStream
+    val proto  = CodedOutputStream.newInstance(stream)
+    proto.writeFixed32NoTag(value)
+    proto.flush()
+    stream.toByteArray
+  }
+  def generateSystemDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    Tools.rng(serializeInt32Fixed(0) ++ sender.toByteArray() ++ serializeInt32Fixed(seqNum))
```
tgrospic: Please extract as a variable this prefix, something like `SYSTEM_DEPLOY_PREFIX`. We want to communicate that this is important info.
I was thinking about serialized string as a prefix, but I like your choice better.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  def generateCloseDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    generateSystemDeployRandomSeed(sender, seqNum).splitByte(0)
+
+  def generateCloseDeployRandomSeed(pk: PublicKey, seqNum: Int): Blake2b512Random = {
+    val sender = ByteString.copyFrom(pk.bytes)
+    generateSystemDeployRandomSeed(sender, seqNum).splitByte(0)
```
tgrospic: With `ByteString` you can call overload from previous definition that accepts `Validator`.

tgrospic (approved):
Very nice work, Will. Thanks!

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
```diff
@@ -81,23 +81,21 @@ object Runtime {
...
 
-  // parentHash in side the BlockData is only used for closeBlock system deploy random initial
-  // parentHash is not exposed in the rholang contract `rho:block:data`
-  // maybe future we will refactor this into a better way
+  // seqNum is not exposed to rholang api contract currently
   final case class BlockData private (
       timeStamp: Long,
       blockNumber: Long,
       sender: PublicKey,
-      parentHash: Array[Byte]
```
Isaac-DeFrain, line 91: Was this parentHash in BlockData only being used to generate random seeds for system deploys?

Isaac-DeFrain (approved):
LGTM. Nice work!

rholang/src/main/scala/coop/rchain/rholang/interpreter/Runtime.scala
tgrospic: Yes, you can see in the comments. ;)

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUtil.scala
```diff
@@ -0,0 +1,63 @@
...
+      serializeInt32Fixed(SYSTEM_DEPLOY_PREFIX) ++ sender
+        .toByteArray() ++ serializeInt32Fixed(seqNum)
+    )
+
+  def generateCloseDeployRandomSeed(sender: Validator, seqNum: Int): Blake2b512Random =
+    generateSystemDeployRandomSeed(sender, seqNum).splitByte(0)
+
+  def generateCloseDeployRandomSeed(pk: PublicKey, seqNum: Int): Blake2b512Random = {
+    val sender = ByteString.copyFrom(pk.bytes)
+    generateCloseDeployRandomSeed(sender, seqNum).splitByte(0)
```
tgrospic: Calling this overload should be without `.splitByte(0)`.

PR #2881, opened 2020-02-14
RCHAIN-4029: fix Web API prepare-deploy to return correct sequence number

tgrospic:
## Overview
Web API `prepare-deploy` should return `seqNum` for the last block created by the validator. For read-only node it returns `-1`.

NOTE: Also I removed `blockNumber` from `prepare-deploy` so it can be more lightweight (BlockAPI.getBlocks is overhead for that).
The latest block number can be retrieved with `/blocks/1`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4029

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -664,4 +664,26 @@ object BlockAPI {
     )
   }
 
+  sealed trait LatestBlockMessageError     extends Throwable
+  final case object ValidatorReadOnlyError extends LatestBlockMessageError
+  final case object NoBlockMessageError    extends LatestBlockMessageError
+
+  def getLatestMessage[F[_]: Sync: EngineCell: Log]: F[ApiErr[BlockMetadata]] = {
+    val errorMessage =
+      "Could not execute exploratory deploy, casper instance was not available yet."
+    EngineCell[F].read >>= (
+      _.withCasper(
+        implicit casper =>
+          for {
+            validatorOpt     <- casper.getValidator
+            validator        <- validatorOpt.liftTo[F](ValidatorReadOnlyError)
+            dag              <- casper.blockDag
+            latestMessageOpt <- dag.latestMessage(ByteString.copyFrom(validator.bytes))
+            latestMessage    <- latestMessageOpt.liftTo[F](NoBlockMessageError)
+          } yield latestMessage.asRight[Error],
+        Log[F].warn(errorMessage).as(s"Error: $errorMessage".asLeft)
+      )
+    )
+  }
+
 }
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -664,4 +664,26 @@ object BlockAPI {
...
+      "Could not execute exploratory deploy, casper instance was not available yet."
+    EngineCell[F].read >>= (
+      _.withCasper(
+        implicit casper =>
+          for {
+            validatorOpt     <- casper.getValidator
+            validator        <- validatorOpt.liftTo[F](ValidatorReadOnlyError)
+            dag              <- casper.blockDag
+            latestMessageOpt <- dag.latestMessage(ByteString.copyFrom(validator.bytes))
+            latestMessage    <- latestMessageOpt.liftTo[F](NoBlockMessageError)
```
tgrospic, line 682: @zsluedem we can have nicer and more typed errors. Return of `ApiErr` is just because of `withCasper` that accepts the default also.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -664,4 +664,26 @@ object BlockAPI {
...
+
+  def getLatestMessage[F[_]: Sync: EngineCell: Log]: F[ApiErr[BlockMetadata]] = {
+    val errorMessage =
+      "Could not execute exploratory deploy, casper instance was not available yet."
+    EngineCell[F].read >>= (
+      _.withCasper(
+        implicit casper =>
+          for {
+            validatorOpt     <- casper.getValidator
+            validator        <- validatorOpt.liftTo[F](ValidatorReadOnlyError)
```
Isaac-DeFrain, line 679: Are we making a distinction between readonly and non-readonly validators here?

9rb (approved):

Isaac-DeFrain (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: In read-only node there is no validator identity (`casper.getValidator` returns an Option) so this function is not total. If validator is not present, the last message cannot be also and this is marked with `ValidatorReadOnlyError` exception.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
zsluedem: Agree

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
zsluedem: First, my understanding of read-only node is that a node can not propose a block.
As for the definition above, there is also another situation when a node with validator-identity can be read-only -> a node in the current post-state of tuplespace is not an active validator can be a read-only node. https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala#L728-L738 some related codes I wrote.
Just want to mention that.

Aside from the conversations above, we need DOCS for terminology.

PR #2883, opened 2020-02-16
RCHAIN-4030 skip genesis validation when connecting exist network and have genesis

zsluedem:
## Overview
The original thoughts in discord was to introduce a flag to skip the genesis validation. After studying the [codes](https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala#L34-L56), I think we can just skip validation on case of connecting to the existing network with the genesis block stored. @tgrospic @nzpr @Isaac-DeFrain 

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4030

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -89,22 +89,37 @@ sealed abstract class MultiParentCasperInstances {
       validatorId: Option[ValidatorIdentity],
       genesis: BlockMessage,
       shardId: String,
-      finalizationRate: Int
+      finalizationRate: Int,
+      skipValidateGenesis: Boolean
   )(implicit runtimeManager: RuntimeManager[F]): F[MultiParentCasper[F]] =
     Span[F].trace(genesisLabel) {
       for {
         dag <- BlockDagStorage[F].getRepresentation
-        maybePostGenesisStateHash <- InterpreterUtil
-                                      .validateBlockCheckpoint(genesis, dag, runtimeManager)
-        postGenesisStateHash <- maybePostGenesisStateHash match {
-                                 case Left(BlockError.BlockException(ex)) =>
-                                   ex.raiseError[F, StateHash]
-                                 case Left(error) =>
-                                   new Exception(s"Block error: $error").raiseError[F, StateHash]
-                                 case Right(None) =>
-                                   new Exception("Genesis tuplespace validation failed!")
-                                     .raiseError[F, StateHash]
-                                 case Right(Some(hash)) => hash.pure
+        postGenesisStateHash <- if (skipValidateGenesis) {
+                                 Log[F].warn("Skip genesis block validation!") >> genesis.body.state.postStateHash.pure
+                               } else {
+                                 for {
+                                   maybePostGenesisStateHash <- InterpreterUtil
+                                                                 .validateBlockCheckpoint(
+                                                                   genesis,
+                                                                   dag,
+                                                                   runtimeManager
+                                                                 )
+                                   postGenesisStateHash <- maybePostGenesisStateHash match {
+                                                            case Left(
+                                                                BlockError.BlockException(ex)
+                                                                ) =>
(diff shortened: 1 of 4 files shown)


nzpr:
@zsluedem but what happens if someone start (e.g. accidentally) non-compatible RNode release with genesis block stored? IMO thats risky, and flag to bypass genesis validation should be used only when it's needed.

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -60,7 +60,13 @@ object CasperLaunch {
         validatorId <- ValidatorIdentity.fromConfig[F](conf)
         genesis     = approvedBlock.candidate.block
         casper <- MultiParentCasper
-                   .hashSetCasper[F](validatorId, genesis, conf.shardId, conf.finalizationRate)
+                   .hashSetCasper[F](
+                     validatorId,
+                     genesis,
+                     conf.shardId,
+                     conf.finalizationRate,
+                     true
```
tgrospic: Please use named parameter, it will be more understandable than just `true`. There are other two places also.

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -89,22 +89,37 @@ sealed abstract class MultiParentCasperInstances {
...
-                                 case Left(BlockError.BlockException(ex)) =>
-                                   ex.raiseError[F, StateHash]
-                                 case Left(error) =>
-                                   new Exception(s"Block error: $error").raiseError[F, StateHash]
-                                 case Right(None) =>
-                                   new Exception("Genesis tuplespace validation failed!")
-                                     .raiseError[F, StateHash]
-                                 case Right(Some(hash)) => hash.pure
+        postGenesisStateHash <- if (skipValidateGenesis) {
+                                 genesis.body.state.postStateHash.pure
```
tgrospic: Maybe it would be useful to log a warning that validation is skipped.

tgrospic:
This validation checks if node can create identical genesis block. If it cannot, this doesn't mean that node cannot run in the network. If we change something in PoS contract in the project, it can fail to create identical genesis but will still be a valid version that can run on the same network.
I don't see that this validation gives us very valuable information when node is connecting to existing network and can easily gives us false negatives although node is supported.

@nzpr if we have this as a parameter, in which situations it will not be used when connecting to existing network? I'm not against this parameter, but if it will be used almost always it will defeat the purpose.

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
zsluedem: Eagle eyes 👍

nzpr:
@tgrospic that's very good point. I'm not 100% sure what the genesis block validation when starting the node is for, so just being careful, as this validation is in the codebase for quite a long time I believe.
Though I found this comment by @KentShikama 
https://github.com/rchain/rchain/pull/2164#issue-249845890 
`Note the only time that genesis validation happens is
at the beginning of the chain.`

zsluedem:
@nzpr @tgrospic Got any conclusions? I am more favor with not adding parameter but fine with adding it.

nzpr:
@zsluedem I changed my mind and I'm for not adding parameter. If we will figure out why this validation is needed, we can add it. But for mow most probably we will always disable existing genesis block validation.

zsluedem:
> @zsluedem I changed my mind and I'm for not adding parameter. If we will figure out why this validation is needed, we can add it. But for mow most probably we will always disable existing genesis block validation.

Yes. It is actually very easy to add the parameter. We could do that later if that is more appropriate.

tgrospic (approved):
I'm also OK that we disable validation without additional parameter but please create a ticket for better chain compatibility validation.

nzpr (approved):

PR #2884, opened 2020-02-16
Update fork-choice tips when node is supposedly stuck

nzpr:
## Overview
As we introduced synchrony constraint - there might be situation when node is stuck.
As an edge case with `sync = 0.99`, if node misses the block that is the last one to meet sync constraint, it has no way to request it after it was broadcasted. So it will never meet synchrony constraint. To mitigate this issue we can update fork choice tips if current fork-choice tip has old timestamp, which means node does not propose new blocks and no new blocks were received recently.

### JIRA ticket:

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -6,6 +6,7 @@ import cats.{Applicative, Monad}
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.BlockStore
 import coop.rchain.casper._
+import coop.rchain.casper.engine.EngineCell.EngineCell
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.util.comm.CommUtil
 import coop.rchain.comm.protocol.routing.Packet
@@ -17,6 +18,7 @@ import coop.rchain.metrics.Metrics
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.Validator.Validator
 import coop.rchain.shared.{Cell, Log, Time}
+import monix.eval.Task
 
 import scala.concurrent.duration._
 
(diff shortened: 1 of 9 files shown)


nzpr:
It's WIP, I'm going to test after https://github.com/rchain/rchain/pull/2883 merged. Also I believe this params have to be adjusted https://github.com/rchain/rchain/blob/4bf93c2a212d86cd9c7dfb4605d73da47ba06eb8/node/src/main/scala/coop/rchain/node/NodeRuntime.scala#L71 IMO we should give more threads to loopScheduler.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -62,6 +62,17 @@ object Running {
       blockHash: BlockHash
   ) = requestForBlock(peer, blockHash) >> addNewEntry(blockHash, Some(peer))
 
+  /**
+    * As we introduced synchrony constraint - there might be situation when node is stuck.
+    * As an edge case with `sync = 0.99`, if node misses the block that is the last one to meet sync constraint,
+    * it has no way to request it after it was broadcasted. So it will never meet synchrony constraint.
+    * To mitigate this issue we can update fork choice tips on schedule.
```
zsluedem: ~~How is updating fork choice going to help? Can you elaborate?~~
Nvm, I got what you say

zsluedem:
I think your pr make sense and right. I had this idea before. But what we should care is that do we really send forkchoice request to all other peers periodicly. I think it would make more sense to send the forkchoice request after the node haven't received messages for a period.

nzpr:
@zsluedem I agree with `send the forkchoice request after the node haven't received messages for a period`

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -793,20 +793,24 @@ object NodeRuntime {
...
-              Metrics[F]
-            )
-        _ <- Time[F].sleep(conf.casper.casperLoopInterval.seconds)
-      } yield ()
+      casperLoop = {
+        implicit val cu = commUtil
+        for {
+          engine <- engineCell.read
+          _      <- engine.withCasper(_.fetchDependencies, Applicative[F].unit)
+          _      <- Running.updateForkChoiceTips[F]
```
tgrospic: Why not just use `commUtil.sendForkChoiceTipRequest` here?

Also my concern is if this will wait for peers to respond. It would be useful to test this scenario when peers are down.

tgrospic:
> IMO we should give more threads to loopScheduler.

@nzpr with async (IO) operations threads are released when e.g. reading from a file or when sending a packet over gRPC. So for _loopScheduler_ is ok to have fewer threads but they should not execute CPU heavy operations to block these threads.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -807,6 +816,17 @@ object NodeRuntime {
...
       } yield ()
+      // Broadcast fork choice tips request if current fork choice is more then 3 minutes old.
+      // For why - look at updateForkChoiceTipsIfStuck method description.
+      updateForkChoiceLoop = {
+        implicit val cu = commUtil
+        implicit val ec = engineCell
+        implicit val bs = blockStore
+        for {
+          _ <- Running.updateForkChoiceTipsIfStuck(3.minutes)
+          _ <- Time[F].sleep(1.minutes)
```
zsluedem: We should not hard-coded the parameter. Can you put it into configuration?

nzpr: @zsluedem 863a271

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -373,6 +373,20 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
...
+      opt[FiniteDuration](
+        default = Option(FiniteDuration(3, scala.concurrent.duration.MINUTES)),
+        descr =
+          "Node will request for fork choice tips if the latest FCT is more then " +
+            "forkChioceStaleThreshold old. Default 3 min."
+      )
+
+    val forkChioceCheckIfStaleInterval =
+      opt[FiniteDuration](
+        default = Option(FiniteDuration(1, scala.concurrent.duration.MINUTES)),
```
nzpr, line 387: IMO we should have default for each run arg set here. Now values are spread across codebase.

nzpr:
@tgrospic @zsluedem this PR is ready

zsluedem (approved):
LGTM

tgrospic (approved):
LGTM

PR #2890, opened 2020-02-21
fix None.get error in safetyOracle

zsluedem:
## Overview
fix None.get error in safetyOracle

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-3976

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
+++ b/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
@@ -73,27 +73,38 @@ sealed abstract class SafetyOracleInstances {
           candidateBlockHash: BlockHash
       ): F[Float] = Span[F].trace(SafetyOracleMetricsSource) {
         for {
-          candidateMetadata <- blockDag.lookup(candidateBlockHash).map(_.get)
-          totalWeight       <- computeTotalWeight(blockDag, candidateMetadata)
-          _                 <- Span[F].mark("total-weight")
-          agreeingValidatorToWeight <- computeAgreeingValidatorToWeight(
-                                        blockDag,
-                                        candidateMetadata
-                                      )
-          _ <- Span[F].mark("agreeing-validator-to-weight")
-          maxCliqueWeight <- if (2L * agreeingValidatorToWeight.values.sum < totalWeight) {
-                              0L.pure[F]
-                            } else {
-                              agreementGraphMaxCliqueWeight(
-                                blockDag,
-                                candidateMetadata,
-                                agreeingValidatorToWeight
-                              )
-                            }
-          _ <- Span[F].mark("max-clique-weight")
+          maybeCandidateMetadata <- blockDag.lookup(candidateBlockHash)
+          faultTolerance <- maybeCandidateMetadata match {
+                             case Some(candidateMetadata) =>
+                               for {
+                                 totalWeight <- computeTotalWeight(blockDag, candidateMetadata)
+                                 _           <- Span[F].mark("total-weight")
+                                 agreeingValidatorToWeight <- computeAgreeingValidatorToWeight(
+                                                               blockDag,
+                                                               candidateMetadata
+                                                             )
+                                 _ <- Span[F].mark("agreeing-validator-to-weight")
+                                 maxCliqueWeight <- if (2L * agreeingValidatorToWeight.values.sum < totalWeight) {
+                                                     0L.pure[F]
+                                                   } else {
+                                                     agreementGraphMaxCliqueWeight(
+                                                       blockDag,
+                                                       candidateMetadata,
+                                                       agreeingValidatorToWeight
+                                                     )
+                                                   }
+                                 _ <- Span[F].mark("max-clique-weight")
 
-          faultTolerance = 2 * maxCliqueWeight - totalWeight
-        } yield faultTolerance.toFloat / totalWeight
+                                 faultTolerance = 2 * maxCliqueWeight - totalWeight
+                               } yield faultTolerance.toFloat / totalWeight
+                             // if the node can to find the block, it would return -1 and regard that block as orphaned
+                             case None =>
+                               Log[F].info(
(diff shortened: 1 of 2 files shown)


nzpr:
@zsluedem do you have understanding why this error were showing itself?  O wonder why `normalizedFaultTolerance` was even called on a block that is not present in DAG.

tgrospic (approved):
LGTM

zsluedem:
> @zsluedem do you have understanding why this error were showing itself? O wonder why `normalizedFaultTolerance` was even called on a block that is not present in DAG.

When we add a block, firstly we add the block in the blockStore no matter whether is a valid block or invalid block and and after validation or something  then add it into blockDagStore.
Here is the code https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L88-L91 which gives better look. 
After propose , if you `show-block` right after the propose then you can get the error

zsluedem:
https://github.com/rchain/rchain/pull/2890#issuecomment-589954346
After some thinking, I think it is wrong to think like that.

nzpr:
@zsluedem actually I find this weird, as both here [1]
https://github.com/rchain/rchain/blob/ec9798d8442dd1521f1cf05bc04a85acc8e28f96/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L422

and 
here [2]
https://github.com/rchain/rchain/blob/ec9798d8442dd1521f1cf05bc04a85acc8e28f96/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L448

queries for a block should do the same.
But for some reason in [1] `implicit casper: MultiParentCasper[F]` is left unused and instead `MultiParentCasper[F]` queried.

nzpr:
Ah, in [1] query made on blockStore actually
https://github.com/rchain/rchain/blob/ec9798d8442dd1521f1cf05bc04a85acc8e28f96/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L508

nzpr:
@zsluedem I believe you are right with your reasoning, and error shows itself when block went into blockStore but was not added to DAG yet. And the purpose of geBlock is to obtain block from the DAG. So it should be fixed by using here 
https://github.com/rchain/rchain/blob/ec9798d8442dd1521f1cf05bc04a85acc8e28f96/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L423
instead of ` maybeBlock <- getBlock[F](hash, dag)` => `blockDag.lookup(candidateBlockHash).map(_.get)`. 
Or modifying getBlock, as now it queries on block store instead of DAG despite that it has dag passed into (which is unused).
How do you think?

nzpr:
Imo the solution in this PR can create ambiguity when an invalid block is received and sits in block store waiting for being added. If user query some node for a block - it expects that if node responds, it verified that block (added to the DAG). How about this option? https://github.com/rchain/rchain/pulls

zsluedem:
> Imo the solution in this PR can create ambiguity when an invalid block is received and sits in block store waiting for being added. If user query some node for a block - it expects that if node responds, it verified that block (added to the DAG). How about this option? https://github.com/rchain/rchain/pulls

This pr doesn't touch any logic of the store things, it just handles the case which isn't handled well. I don't know why it is ambiguity.

As for your comments, I think it is also useful to show the unverified block(It helps debug to get debug info). What we have to do is to provide more info in the return message.

zsluedem:
@nzpr Pick your pr in https://github.com/rchain/rchain/pull/2892. 
I think that satisfies everyone now.

nzpr (approved):
Yes, makes sense!

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -134,8 +146,9 @@ sealed abstract class SafetyOracleInstances {
           candidateMetadata: BlockMetadata
       ): F[Map[BlockHash, Long]] =
         candidateMetadata.parents.headOption match {
-          case Some(parent) => blockDag.lookup(parent).map(_.get.weightMap)
-          case None         => candidateMetadata.weightMap.pure[F]
+          case Some(parent) =>
+            blockDag.lookup(parent).map(_.fold(candidateMetadata.weightMap)(_.weightMap))
+          case None => candidateMetadata.weightMap.pure[F]
```
tgrospic, line 151: You can make this little nicer without duplicating default value. With _traverse_ (_sequence_) you can "invert" two monads (`F` can be applicative only) `Option[F[A]]` to `F[Option[A]]`.
Also we can save the creation of one Task. In runtime `F` will be a Monix Task, so wrapping a value with `pure[F]` will create a new Task just to immediately return this value. We should investigate how big is this overhead especially in large loops.
```scala
import cats.instances.option._
for {
  maybeMetadata <- candidateMetadata.parents.headOption.flatTraverse(blockDag.lookup)
} yield maybeMetadata.map(_.weightMap).getOrElse(candidateMetadata.weightMap)
```

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -73,27 +73,39 @@ sealed abstract class SafetyOracleInstances {
...
+                                                   }
+                                 _ <- Span[F].mark("max-clique-weight")
+
+                                 faultTolerance = 2 * maxCliqueWeight - totalWeight
+                               } yield faultTolerance.toFloat / totalWeight
+                             // if the node can to find the block, it would return -1 and regard that block as orphaned
+                             case None =>
+                               Log[F].info(
+                                 s"calculate faultTolerance blockHash ${candidateBlockHash} failed because it can not be found in store."
+                               ) >> (-1L).toFloat
```
nzpr, line 105: @zsluedem just to be clear what I meant by ambiguity - this will return valid `show-block` with faultToleranceThreshold=-1 for block that is not added yet to the DAG (previously we had error in this case).

PR #2891, opened 2020-02-22
Make block propagation aware of block store

nzpr:
Liveness of the protocol is supported by casper loop https://github.com/rchain/rchain/blob/ab3edcca40cf0c80eb6f6da31ee9271edc95ca37/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L105
which maintains queue of blocks that casper needs to get from another peers. Node adds to that queue when requesting block
https://github.com/rchain/rchain/blob/ab3edcca40cf0c80eb6f6da31ee9271edc95ca37/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L62-L65
and removes after block is added to the DAG.
https://github.com/rchain/rchain/blob/ab3edcca40cf0c80eb6f6da31ee9271edc95ca37/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L199-L211

So atm block request is removed from the queue only when it is in the DAG which causes faulty rerequests.
Node might actually receive block, put in the block store and send it to casper to add, where it will wait for `blockProcessingLock` to be released https://github.com/rchain/rchain/blob/a8ace3b28430f5af323bd0710eec0628cef297d2/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L98 
If there are lots of block being processed, or a single big block occupied lock for a long time - as we are unable to process blocks in parallel, on the next casper loop invocation this block still won't be in the DAG. This will trigger rerequest, which is redundant.

To disable this we can check for block store instead of dag store for hash presence on block message arrival.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -197,10 +197,10 @@ object Running {
     )
 
   def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks](peer: PeerNode, b: BlockMessage)(
-      casperContains: BlockHash => F[Boolean],
+      blockStoreContains: BlockHash => F[Boolean],
       casperAdd: BlockMessage => F[ValidBlockProcessing]
   ): F[Unit] =
-    casperContains(b.blockHash)
+    blockStoreContains(b.blockHash)
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
         Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> casperAdd(b) >>= (
@@ -299,7 +299,7 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
 
   override def handle(peer: PeerNode, msg: CasperMessage): F[Unit] = msg match {
     case blockHash: BlockHashMessage => handleBlockHashMessage(peer, blockHash)(casper.contains)
-    case b: BlockMessage             => handleBlockMessage(peer, b)(casper.contains, casperAdd(peer))
+    case b: BlockMessage             => handleBlockMessage(peer, b)(BlockStore[F].contains, casperAdd(peer))
     case br: BlockRequest            => handleBlockRequest(peer, br)
     case hbr: HasBlockRequest        => handleHasBlockRequest(peer, hbr)(casper.contains)
     case hb: HasBlock                => handleHasBlock(peer, hb)(casper.contains)
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -102,7 +102,7 @@ object Running {
     * and keep the requested blocks list clean.
     * See spec RunningMaintainRequestedBlocksSpec for more details
     */
-  def maintainRequestedBlocks[F[_]: Monad: RPConfAsk: RequestedBlocks: TransportLayer: Log: Time: Metrics](
+  def maintainRequestedBlocks[F[_]: Monad: RPConfAsk: RequestedBlocks: TransportLayer: Log: Time: Metrics: BlockStore](
       timeout: Int
   ): F[Unit] = {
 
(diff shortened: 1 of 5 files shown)


tgrospic (approved):
Nice catch!

zsluedem (approved):

nzpr:
Also we should move adding adding of block into block store out of blockProcessingLock. https://github.com/rchain/rchain/blob/a8ace3b28430f5af323bd0710eec0628cef297d2/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L105.
But I wonder if it is safe?

tgrospic:
> Also we should move adding adding of block into block store out of blockProcessingLock.
>
https://github.com/rchain/rchain/blob/a8ace3b28430f5af323bd0710eec0628cef297d2/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L105
> .
> But I wonder if it is safe?

I think you are right. It should be safe because `BlockStore.put` has already a lock. This change can be part of this PR.

nzpr:
I guess there might be some other changes needed. If node shut down while some `casperAdd` calls are waiting for the lock - whats the logic for these blocks knocking the casper door again after restart? Might be there are ways, but need to check.

tgrospic:
Now, when block is processing, new blocks are received and cannot be saved to disk. So after restart these blocks must be re-requested. If saving blocks is outside of the block processing lock, received blocks will be saved and after restart they will be in the BlockStore.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -196,14 +196,20 @@ object Running {
...
+  )(
       blockStoreContains: BlockHash => F[Boolean],
       casperAdd: BlockMessage => F[ValidBlockProcessing]
   ): F[Unit] =
     blockStoreContains(b.blockHash)
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
-        Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> casperAdd(b) >>= (
+        Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> {
+          // TODO check if block is valid here before placing in block store
```
zsluedem, line 222: create a ticket

nzpr: https://rchain.atlassian.net/browse/RCHAIN-4037

nzpr:
Might be this PR have to include one more change. With this PR merged we will be dropping blocks that are present in block store. So we enable situation when if due to some reason Casper failed to add block (e.g. due to node restart) but node have it already in block store - casper will be able to add that block only when receive its descendant and find this block as a parent. In general case this should not be the problem, but there might be edge cases when this can block the node.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -196,14 +208,20 @@ object Running {
...
   ): F[Unit] =
-    casperContains(b.blockHash)
+    blockStoreContains(b.blockHash)
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
-        Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> casperAdd(b) >>= (
+        Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> {
+          // TODO check if block is valid here before placing in block store
+          BlockStore[F].put(b)
+        } >> casperAdd(b) >>= (
```
tgrospic, line 224: This function already abstracts the notion of containing and adding a block. It can also accepts function to add to the store. Or maybe BlockStore.put can be done in `casperAdd` function which has BlockStore already. In both cases BlockStore instance in not necessary.
https://github.com/rchain/rchain/blob/591876bd3872c7e877a580942dfa41d6499eec12/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L302-L314

tgrospic (approved):

nzpr:
@tgrospic @zsluedem After reconsideration, I believe the issue can be handled with less code - ptal https://github.com/rchain/rchain/pull/2897

PR #2892, opened 2020-02-22
make getBlock aware if block is added

nzpr:
~~Alternative~~ Addition to https://github.com/rchain/rchain/pull/2890

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -422,8 +422,15 @@ object BlockAPI {
         dag        <- MultiParentCasper[F].blockDag
         maybeBlock <- getBlock[F](hash, dag)
         blockInfo <- maybeBlock match {
-                      case Some(block) =>
-                        getFullBlockInfo[F](block).map(_.asRight[Error])
+                      case Some(block) => {
+                        val blockAdded = dag.contains(block.blockHash)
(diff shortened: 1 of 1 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -422,8 +422,15 @@ object BlockAPI {
         dag        <- MultiParentCasper[F].blockDag
         maybeBlock <- getBlock[F](hash, dag)
         blockInfo <- maybeBlock match {
-                      case Some(block) =>
-                        getFullBlockInfo[F](block).map(_.asRight[Error])
+                      case Some(block) => {
+                        val blockAdded = dag.contains(block.blockHash)
```
tgrospic, line 426: I agree with your changes, but they should be addition to #2890.

Please fix `getBlock` (private). Remove Dag parameter because it's not used, the name will clash with public overload so `getBlockFromStore` would be more appropriate.

tgrospic (changes requested):

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
zsluedem: Fixed in https://github.com/rchain/rchain/pull/2890/commits/efd5d419141ea3b5569bd3a11490e622c1b93c82

PR #2897, opened 2020-03-01
[WIP] Prevent Casper from making redundant rerequest

nzpr:
Replacement for https://github.com/rchain/rchain/pull/2891

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -214,11 +214,8 @@ object Running {
   def handleBlockHashMessage[F[_]: Monad: Log: ConnectionsCell: TransportLayer: Time: RPConfAsk: RequestedBlocks](
       peer: PeerNode,
       blockHashMessage: BlockHashMessage
-  )(casperContains: BlockHash => F[Boolean]): F[Unit] =
-    (
-      casperContains(blockHashMessage.blockHash),
-      RequestedBlocks.contains(blockHashMessage.blockHash)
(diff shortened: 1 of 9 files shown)


nzpr:
I believe we need to get rid of this lock
https://github.com/rchain/rchain/blob/09359edf042c9c995afbf9cf1862e4fb023b5399/casper/src/main/scala/coop/rchain/casper/util/comm/FairRoundRobinDispatcher.scala#L30-L36
The makes the whole casper message processing atomic, acquiring lock right after GRPC endpoint receives it.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -30,6 +28,7 @@ object Running {
   final case class Requested(
       timestamp: Long,
       peers: Set[PeerNode] = Set.empty,
+      received: Boolean,
```
tgrospic, line 31: Maybe it makes sense to set by default to `false`.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -60,7 +60,7 @@ class GrpcTransportClient(
 
   // Start to consume the stream queue immediately
   private val _ = streamQueue.subscribe()(
-    Scheduler.fixedPool("tl-client-stream-queue", parallelism, reporter = UncaughtExceptionLogger)
+    scheduler
```
tgrospic, line 63: It can be in one line now.

PR #2899, opened 2020-03-04
Rchain 4045 get block info from height

zsluedem:
## Overview
get block info from height

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4045

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
@@ -129,38 +129,66 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
       }
     def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]] =
       blockHashesByDeploy.get(deployId).pure[F]
-    def topoSort(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
-      if (startBlockNumber >= sortOffset) {
-        val offset = startBlockNumber - sortOffset
-        assert(offset.isValidInt)
-        topoSortVector.drop(offset.toInt).pure[F]
-      } else if (sortOffset - startBlockNumber + topoSortVector.length < Int.MaxValue) { // Max Vector length
-        lock.withPermit(
-          for {
-            checkpoints          <- getCheckpoints
-            checkpointsWithIndex = checkpoints.zipWithIndex
-            checkpointsToLoad    = checkpointsWithIndex.filter(startBlockNumber < _._1.end)
-            checkpointsDagInfos <- checkpointsToLoad.traverse {
-                                    case (startingCheckpoint, index) =>
-                                      loadCheckpointDagInfo(startingCheckpoint, index)
-                                  }
-            topoSortPrefix = checkpointsDagInfos.toVector.flatMap { checkpointsDagInfo =>
-              val offset = startBlockNumber - checkpointsDagInfo.sortOffset
-              // offset is always a valid Int since the method result's length was validated before
-              checkpointsDagInfo.topoSort.drop(offset.toInt) // negative drops are ignored
-            }
-            result = topoSortPrefix ++ topoSortVector
-          } yield result
-        )
+
+    private def topoSortBeforeOffset(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
+      lock.withPermit(
+        for {
+          checkpoints          <- getCheckpoints
+          checkpointsWithIndex = checkpoints.zipWithIndex
+          checkpointsToLoad    = checkpointsWithIndex.filter(startBlockNumber < _._1.end)
+          checkpointsDagInfos <- checkpointsToLoad.traverse {
+                                  case (startingCheckpoint, index) =>
+                                    loadCheckpointDagInfo(startingCheckpoint, index)
+                                }
+          result = checkpointsDagInfos.toVector.flatMap { checkpointsDagInfo =>
+            val offset = startBlockNumber - checkpointsDagInfo.sortOffset
+            assert(offset.isValidInt)
+            // offset is always a valid Int since the method result's length was validated before
+            checkpointsDagInfo.topoSort.drop(offset.toInt) // negative drops are ignored
+          }
+        } yield result
+      )
+
+    /** Get the blocks from the dag based on the startBlockNumber and maybeEndBlockNUmber
+      *  @param startBlockNumber the start blockNumber from the fragment in the dag
+      *  @param maybeEndBlockNumber this param can be optional, if it is none, it would choose the last blockNumber
+      *                             in the dag store
+      */
+    def topoSort(
+        startBlockNumber: Long,
+        maybeEndBlockNumber: Option[Long]
+    ): F[Vector[Vector[BlockHash]]] = {
+      val endBlockNumber = maybeEndBlockNumber.getOrElse(sortOffset + topoSortVector.length)
+      if (startBlockNumber >= sortOffset && startBlockNumber <= endBlockNumber) {
+        val offsetHead = startBlockNumber - sortOffset
+        // minus 1 to include endBlockNumber
+        val offsetTail = endBlockNumber - sortOffset + 1
+        assert(offsetHead.isValidInt)
+        assert(offsetTail.isValidInt)
+        topoSortVector.slice(offsetHead.toInt, offsetTail.toInt).pure[F]
+      } else if (startBlockNumber < sortOffset && endBlockNumber > sortOffset) {
+        for {
+          topoSortPrefix <- topoSortBeforeOffset(startBlockNumber.toLong)
+          topoSortTail   = topoSortVector.take((endBlockNumber - sortOffset + 1).toInt)
+          result         = topoSortPrefix ++ topoSortTail
+        } yield result
+      } else if (startBlockNumber < sortOffset && endBlockNumber <= sortOffset) {
+        for {
+          topoSortPrefix <- topoSortBeforeOffset(startBlockNumber.toLong)
+          offsetTail     = sortOffset - endBlockNumber
+          result         = topoSortPrefix.dropRight(offsetTail.toInt)
+        } yield result
       } else {
         Sync[F].raiseError(
-          TopoSortLengthIsTooBig(sortOffset - startBlockNumber + topoSortVector.length)
+          TopoSortFragmentParameterError(startBlockNumber, endBlockNumber)
         )
       }
+    }
+
     // TODO should startBlockNumber have topoSortVector.length - 1 (off by one error)?
     def topoSortTail(tailLength: Int): F[Vector[Vector[BlockHash]]] = {
       val startBlockNumber = Math.max(0L, sortOffset - (tailLength - topoSortVector.length))
-      topoSort(startBlockNumber)
+      topoSort(startBlockNumber, none)
     }
     def latestMessageHash(validator: Validator): F[Option[BlockHash]] =
       latestMessagesMap.get(validator).pure[F]
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
@@ -149,7 +149,7 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
         } yield result
       )
 
-    def latestBlockNumber: F[Long] = (sortOffset + topoSortVector.length - 1).pure[F]
+    def latestBlockNumber: F[Long] = (sortOffset + topoSortVector.length).pure[F]
 
     /** Get the blocks from the dag based on the startBlockNumber and maybeEndBlockNUmber
       * There is one thing has to be careful when calculating the index.Because the genesis block number
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
@@ -130,21 +130,30 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
     def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]] =
       blockHashesByDeploy.get(deployId).pure[F]
 
-    private def topoSortBeforeOffset(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
+    private def topoSortBeforeOffset(
+        startBlockNumber: Long,
+        endBlockNumber: Long
+    ): F[Vector[Vector[BlockHash]]] =
       lock.withPermit(
         for {
           checkpoints          <- getCheckpoints
           checkpointsWithIndex = checkpoints.zipWithIndex
-          checkpointsToLoad    = checkpointsWithIndex.filter(startBlockNumber < _._1.end)
+          checkpointsToLoad = checkpointsWithIndex.filter(
+            cp => startBlockNumber < cp._1.end && endBlockNumber > cp._1.start
+          )
           checkpointsDagInfos <- checkpointsToLoad.traverse {
                                   case (startingCheckpoint, index) =>
                                     loadCheckpointDagInfo(startingCheckpoint, index)
                                 }
-          result = checkpointsDagInfos.toVector.flatMap { checkpointsDagInfo =>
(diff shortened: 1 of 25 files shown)


nzpr (approved):
LGTM, but minor comment - I suggest to add comments to each new method (I hope one day there will be rich comments across all codebase). Like this https://github.com/rchain/rchain/blob/a8ace3b28430f5af323bd0710eec0628cef297d2/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L30-L35

This would help creating automatic documentation.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
```diff
@@ -137,26 +155,54 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
...
-                                      loadCheckpointDagInfo(startingCheckpoint, index)
-                                  }
-            topoSortPrefix = checkpointsDagInfos.toVector.flatMap { checkpointsDagInfo =>
-              val offset = startBlockNumber - checkpointsDagInfo.sortOffset
-              // offset is always a valid Int since the method result's length was validated before
-              checkpointsDagInfo.topoSort.drop(offset.toInt) // negative drops are ignored
-            }
-            result = topoSortPrefix ++ topoSortVector
+            topoSortPrefix <- topoSortBeforeOffset(startBlockNumber)
+            result         = topoSortPrefix ++ topoSortVector
```
tgrospic: Why not move the lock also to `topoSortBeforeOffset`. This should not be responsibility of the caller to care about the lock.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
```diff
@@ -32,6 +32,7 @@ trait BlockDagRepresentation[F[_]] {
   def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]]
   def topoSort(startBlockNumber: Long): F[Vector[Vector[BlockHash]]]
   def topoSortTail(tailLength: Int): F[Vector[Vector[BlockHash]]]
+  def topoSortFragment(startBlockNumber: Long, endBlockNumber: Long): F[Vector[Vector[BlockHash]]]
```
tgrospic: Can you please see if we can only have one function for `topoSort` which accepts start block and optionally end block.

tgrospic (changes requested):

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
```diff
@@ -131,6 +140,11 @@ object WebApiRoutes {
       case GET -> Root / "blocks" =>
         webApi.getBlocks(none).handle
 
+      case GET -> Root / "blocks" :? StartBlockNumberParamMatcher(startBlockNumber) +& EndBlockNumberParamMatcher(
```
tgrospic: What do you think without query strings? E.g. `/block-range/2/42`.
```scala
case GET -> Root / "blocks-range" / IntVar(start) / IntVar(end) =>
```

casper/src/test/scala/coop/rchain/casper/api/BlocksResponseAPITest.scala
```diff
@@ -36,232 +39,182 @@ class BlocksResponseAPITest
...
+//        blocksResponse <- BlockAPI.getBlocks[Task](Some(2))(
+//                           Sync[Task],
+//                           engineCell,
+//                           logEff,
+//                           cliqueOracleEffect,
+//                           blockStore
+//                         )
+//      } yield blocksResponse.right.get.length should be(2) // TODO: Switch to 3 when we implement block height correctly
+//    }
+//  }
```
tgrospic: Leftover.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
```diff
@@ -46,8 +46,16 @@ final class InMemBlockDagStorage[F[_]: Concurrent: Sync: Log: BlockStore](
...
-    def topoSort(startBlockNumber: Long): F[Vector[Vector[BlockHash]]] =
-      topoSortVector.drop(startBlockNumber.toInt).pure[F]
+    def topoSort(
+        startBlockNumber: Long,
+        maybeEndBlockNumber: Option[Long]
+    ): F[Vector[Vector[BlockHash]]] = {
+      val endBlockNumber: Long = maybeEndBlockNumber.getOrElse(topoSortVector.length.toLong)
+      topoSortVector
+        .take(endBlockNumber.toInt)
+        .takeRight((endBlockNumber - startBlockNumber + 1).toInt)
```
tgrospic: This is important function so we want to be optimized. These two _take's_ will iterate two times. Can you replace this with `Vector#slice(start, end)`.
Maybe it would be cleaner with `if..then` on end number, so drop can be used as it's now.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
```diff
@@ -46,8 +46,16 @@ final class InMemBlockDagStorage[F[_]: Concurrent: Sync: Log: BlockStore](
...
+        .pure[F]
+    }
     def topoSortTail(tailLength: Int): F[Vector[Vector[BlockHash]]] =
       topoSortVector.takeRight(tailLength).pure[F]
```
tgrospic: Can we now get rid of `topoSortTail` and use your now `topoSort(start, end)`? :tada:

tgrospic: It can be extension function for `BlockDagRepresentation`. Smaller interface is easier to maintain and extension functions can be defined and used separately in projects.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
```diff
@@ -129,38 +129,66 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
...
+        maybeEndBlockNumber: Option[Long]
+    ): F[Vector[Vector[BlockHash]]] = {
+      val endBlockNumber = maybeEndBlockNumber.getOrElse(sortOffset + topoSortVector.length)
+      if (startBlockNumber >= sortOffset && startBlockNumber <= endBlockNumber) {
+        val offsetHead = startBlockNumber - sortOffset
+        // plus 1 to include endBlockNumber
+        val offsetTail = topoSortVector.length - endBlockNumber - sortOffset - 1
+        assert(offsetHead.isValidInt)
+        assert(offsetTail.isValidInt)
+        topoSortVector.drop(offsetHead.toInt).dropRight(offsetTail.toInt).pure[F]
```
tgrospic: `Vector#slice` seems like a good option here also.

Maybe first and third case can be combined and only start number is then calculated in two different ways.

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
zsluedem: Personally , I prefer query string. It is a design style. I don't mind that.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
zsluedem: Just want to share this with you https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/Vector.scala#L211-L213

but I agree that we should use `slice` for better code reading.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
zsluedem: > Maybe first and third case can be combined and only start number is then calculated in two different ways.

I don't get it.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
zsluedem: I don't understand how to do that on the `extension function` you referred to in codes aspect.
Can you describe more?

Because we got two implementations for `BlockDagRepresentation`, the `FileBlockDagRepresentation` is not like that.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
tgrospic: Good point!

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
tgrospic: Sorry, I see now that in third case it must load data before offset so it's not similar to the first case.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
tgrospic: `topoSortTail` can be written in terms of `topoSort(0, depth)` so it doesn't have to be directly on the interface.

zsluedem:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
tgrospic: Sorry, I'm wrong about this. For `topoSortTail` we need the length of _topoSortVector_ so it cannot be just rewritten with `topoSort`.

zsluedem:
@tgrospic Updated

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
```diff
@@ -129,39 +129,74 @@ final class BlockDagFileStorage[F[_]: Concurrent: Sync: Log: RaiseIOError] priva
...
+        topoSortVector.slice(offsetHead.toInt, offsetTail.toInt).pure[F]
+      } else if (startBlockNumber < sortOffset && endBlockNumber > sortOffset) {
+        for {
+          topoSortPrefix <- topoSortBeforeOffset(startBlockNumber.toLong)
+          topoSortTail   = topoSortVector.take((endBlockNumber - sortOffset + 1).toInt)
+          result         = topoSortPrefix ++ topoSortTail
+        } yield result
+      } else if (startBlockNumber < sortOffset && endBlockNumber <= sortOffset) {
+        for {
+          topoSortPrefix <- topoSortBeforeOffset(startBlockNumber.toLong)
```
tgrospic: This is situation where block hashes will be loaded from disk until the offset, even if only small part is needed. It becomes worst if selection is closer to zero.
This function is exposed as API, so we should do what is possible to optimize it.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -300,10 +300,11 @@ object BlockAPI {
...
     def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[A]] =
       for {
-        dag      <- MultiParentCasper[F].blockDag
-        depth    <- maybeDepth.fold(dag.topoSort(0L).map(_.length - 1))(_.pure[F])
-        topoSort <- dag.topoSortTail(depth)
-        result   <- doIt(casper, topoSort)
+        dag              <- MultiParentCasper[F].blockDag
+        totalBlockNumber <- dag.topoSort(0L, none).map(_.length)
+        depth            = maybeDepth.getOrElse(totalBlockNumber)
+        topoSort         <- dag.topoSort((totalBlockNumber - depth).toLong, none)
```
tgrospic: Even if `depth` value is specified `topoSort` will be called with start `0`, which will cause load of all block hashes. 

Previous code is also not correct, it calls `topoSort(0)` two times when depth is not specified. With all that, what is real use case for depth optional?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
zsluedem: https://rchain.atlassian.net/browse/RCHAIN-4049
You created the ticket for that. I want to fix it later.
But my idea is to limit the number in https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala. For internal usage, we don't limit that. What do you think?

zsluedem: nvm. I should also fix that in this pr.

tgrospic: We have something like this. I'm thinking on third case where start and end are less than sort offset.
```
0 ------- 10 ------- 20 ------- 30 ------- 40
                      |=====================|           < topoSortVector
                           S -------- E
      S ------------------ E
      S ------- E

  1. tails
                           S ---------------|
      S --------------|----|
      S --------------|

  2. cut ends
                           S -------- E
      S ------------------ E
      S ------- E
```

zsluedem: Did you see this comment https://github.com/rchain/rchain/pull/2899/files#diff-800b07b2c76799e3264de804716e6e00R158-R159?
Maybe I should extend more on it. I don't exactly know when the [sortoffset](https://github.com/rchain/rchain/blob/dev/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala#L83) would change. Do you have any informations for that? The dag codes looks to me that the `sortoffset` never changes.
Because the condition is unknown to me, I do not know the best solution for now(My assumption is like there would be serveral checkpoints).

Anyway I agree with you that is not a good-performance method to call the api.

zsluedem: My assumptions about  the dag is like below.
```
0----------10------------20------------30----------40
                                                    |===============|  -> toposortVector
                     |============| checkpoint 1
|========|   checkpoint2

```

zsluedem: Oh. Nvm. I think I am wrong. It doesn't have to know the `sortoffset` change time.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -300,28 +300,69 @@ object BlockAPI {
 
     def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[A]] =
       for {
-        dag      <- MultiParentCasper[F].blockDag
-        depth    <- maybeDepth.fold(dag.topoSort(0L).map(_.length - 1))(_.pure[F])
-        topoSort <- dag.topoSortTail(depth)
-        result   <- doIt(casper, topoSort)
+        dag               <- MultiParentCasper[F].blockDag
+        latestBlockNumber <- dag.latestBlockNumber
+        topoSort          <- dag.topoSort((latestBlockNumber - depth).toLong, none)
```
tgrospic, line 305: If you like, you can create `topoSortTail` extension method on BlockDagRepresentation as a higher order part of Dag algebra. It can be together with the interface or in the casper project.
Boolean extensions seems like a good example with companion object.
https://github.com/rchain/rchain/blob/5670b00bc/shared/src/main/scala/coop/rchain/catscontrib/BooleanF.scala

node/src/main/resources/reference.conf
```diff
@@ -99,6 +99,9 @@ rnode {
     }
 
     reporting = false
+
+    # the max block number limit aquire from api
+    api-max-blocks-limit = 50
```
tgrospic, line 104: @nzpr what do you think default should be?

For users to get data from validators very small number is enough.

tgrospic (approved):
Excellent! One comment if you want have fun with extensions. :)

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
zsluedem: To be honest, I don't really know how to do it. I need to understand that.

tgrospic: It's totally ok. We can play with extensions together.

PR #2900, opened 2020-03-04
RCHAIN-4047: Improved get deploy API method

zsluedem:
## Overview
get deploy

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4047

### Notes


node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
```diff
@@ -136,6 +137,9 @@ object WebApiRoutes {
 
       case GET -> Root / "deploy" / deployId =>
         webApi.findDeploy(deployId).handle
+
+      case GET -> Root / "v2" / "deploy" / deployId =>
+        webApi.getDeploy(deployId).handle
```
tgrospic, line 165: I think it's ok to replace existing method.

But `DeployInfo` object is not exactly what we need. It doesn't have any block info, at least we need block number and block hash.
Can you create `DeployResponse` object which will also contain basic info about block. Deploy `term` can also be big so you can omit it or return a blake hash to keep response small.

tgrospic (changes requested):

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
zsluedem: I think turning `term` into blake hash doesn't make sense to the user

tgrospic: I was thinking for comparison. If user has original term it can compare hashes. But not to important.

zsluedem: Currently the response of `getDeploy` is not compatible with the old version. That's why I put it in `v2`. I am not sure if that's ok to you.

PR #2902, opened 2020-03-07
Message propagation improvement

nzpr:
Result of WIP in of https://github.com/rchain/rchain/pull/2897 and https://github.com/rchain/rchain/pull/2891

This PR enable discarding of duplicate messages to lower down load on Casper and system in general.
Moreover, bloat was found in consumption of streamed GRPC messages (blocks) which lead to over a hundred of messages waiting for processing as per `tmp/comm`. It is fixed as well.

As a result, we can see nodes advancing nicely when creating blocks simultaneously (synchrony constraint 0.67).
Screenshot 2020-03-07 at 13 19 11

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -85,7 +85,6 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
 
     def add: F[ValidBlockProcessing] = spanF.trace(AddBlockMetricsSource) {
       for {
-        _      <- BlockStore[F].put(b)
         _      <- spanF.mark("block-store-put")
         dag    <- blockDag
         status <- internalAddBlock(b, dag)
@@ -95,8 +94,20 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
 
     import cats.instances.either._
     for {
+      // Save block in block store before waiting for blockProcessingLock
+      // This is called inside block message handler as well, but required here in case block is not received from peer
+      // but proposed by node itself.
+      _ <- BlockStore[F].contains(b.blockHash).ifM(().pure[F], BlockStore[F].put(b))
+      _ <- Log[F].info(
+            s"Block ${PrettyPrinter.buildString(b.blockHash)} asks for blockProcessingLock."
+          )
+      // ATM Casper is allowed to add only one block at a time
       status <- blockProcessingLock.withPermit {
                  val exists = for {
+                   _ <- Log[F].info(
+                         s"Block ${PrettyPrinter.buildString(b.blockHash)} " +
+                           s"got blockProcessingLock."
+                       )
                    dag         <- blockDag
                    cst         <- state.read
                    dagContains <- dag.contains(b.blockHash)
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -16,9 +16,8 @@ import coop.rchain.comm.transport.{Blob, TransportLayer}
 import coop.rchain.comm.PeerNode
 import coop.rchain.metrics.Metrics
 import coop.rchain.models.BlockHash.BlockHash
-import coop.rchain.models.Validator.Validator
 import coop.rchain.shared.{Cell, Log, Time}
-import monix.eval.Task
+import coop.rchain.catscontrib.Catscontrib.ToBooleanF
 
 import scala.concurrent.duration._
 
@@ -30,6 +29,7 @@ object Running {
   final case class Requested(
       timestamp: Long,
       peers: Set[PeerNode] = Set.empty,
+      received: Boolean = false,
       waitingList: List[PeerNode] = List.empty
   )
 
@@ -143,9 +143,14 @@ object Running {
       requests.keys.toList
         .traverse(hash => {
           val requested = requests(hash)
-          Time[F].currentMillis
-            .map(_ - requested.timestamp > timeout.seconds.toMillis)
-            .ifM(tryRerequest(hash, requested), (hash -> Option(requested)).pure[F])
+          for {
+            received <- requested.received.pure[F]
+            expired  <- Time[F].currentMillis.map(_ - requested.timestamp > timeout.seconds.toMillis)
+            rerequest <- if (expired && !received)
+                          tryRerequest(hash, requested)
+                        else
+                          (hash -> Option(requested)).pure[F]
+          } yield rerequest
         })
         .map(list => toMap(list))
     })
@@ -163,7 +168,7 @@ object Running {
       peer: PeerNode,
       hb: HasBlock
   )(
-      casperContains: BlockHash => F[Boolean]
+      repeatedCasperMessage: BlockHash => F[Boolean]
   ): F[Unit] = {
 
     val hashStr = PrettyPrinter.buildString(hb.hash)
@@ -175,7 +180,7 @@ object Running {
       )
 
     val logIntro = s"Received confirmation from $peer that it has block $hashStr."
-    Log[F].info(logIntro) >> casperContains(hb.hash).ifM(
+    Log[F].info(logIntro) >> repeatedCasperMessage(hb.hash).ifM(
       Applicative[F].unit,
       RequestedBlocks.get(hb.hash) >>= (_.fold(requestForNewBlock(peer, hb.hash))(
         req =>
@@ -197,10 +202,10 @@ object Running {
     )
 
   def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks](peer: PeerNode, b: BlockMessage)(
-      casperContains: BlockHash => F[Boolean],
+      repeatedCasperMessage: BlockHash => F[Boolean],
       casperAdd: BlockMessage => F[ValidBlockProcessing]
   ): F[Unit] =
-    casperContains(b.blockHash)
+    repeatedCasperMessage(b.blockHash)
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
         Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> casperAdd(b) >>= (
@@ -214,11 +219,8 @@ object Running {
   def handleBlockHashMessage[F[_]: Monad: Log: ConnectionsCell: TransportLayer: Time: RPConfAsk: RequestedBlocks](
       peer: PeerNode,
       blockHashMessage: BlockHashMessage
-  )(casperContains: BlockHash => F[Boolean]): F[Unit] =
-    (
-      casperContains(blockHashMessage.blockHash),
-      RequestedBlocks.contains(blockHashMessage.blockHash)
-    ).mapN(_ || _)
+  )(repeatedCasperMessage: BlockHash => F[Boolean]): F[Unit] =
+    repeatedCasperMessage(blockHashMessage.blockHash)
       .ifM(
         Log[F]
           .info(
@@ -281,6 +283,38 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
   private val F    = Applicative[F]
   private val noop = F.unit
 
+  /**
+    * Message that relates to a block (e.g. block message or block hash broadcast message) shall be considered
+    * as repeated and dropped if corresponding block is marked as received in RequestedBlocks or already added to DAG.
+    * Otherwise - that is a new message and shall be processed normal way.
+    *
+    * RequestedBlocks is a part of Casper loop maintaining liveness of the protocol.
+    * It keeping records of blocks being requested and re-request them on schedule if failed first time.
+    * It is expected to get lots of duplicate messages from different peers, so check for duplicate is required.
+    * Processing of duplicate blocks puts load on Casper and highly discouraged.
+    *
+    * When block is received from some peer first time - it is
+    * - marked as `received` in RequestedBlocks,
+    * - added to the block store and
+    * - sent to Casper.
+    * Block is being kept in RequestedBlocks until it is added to the DAG so we can detect duplicates.
+    *
+    * If either RequestedBlocks shows that block was received (but not yet added to Casper)
+    * or block was already added by Casper (so removed from RequestedBlocks) - related message
+    * considered as repeated.
+    *
+    * @param hash Block hash
+    * @return If message is repeated
+    */
+  private def repeatedCasperMessage(hash: BlockHash): F[Boolean] =
+    RequestedBlocks.get(hash).map(x => x.nonEmpty && x.get.received) ||^ casper.contains(hash)
+
+  /**
+    * Wrapper for Casper add block.
+    * @param peer originator of the block
+    * @param b block
+    * @return block processing result
+    */
   private def casperAdd(peer: PeerNode)(b: BlockMessage): F[ValidBlockProcessing] = {
     import cats.instances.option._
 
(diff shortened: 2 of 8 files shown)


casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -281,28 +282,112 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+    *
+    * @param hash Block hash
+    * @return If message is repeated
+    */
+  private def repeatedCasperMessage(hash: BlockHash): F[Boolean] =
+    (
+      // If either RequestedBlocks shows that block was received (but not yet added to Casper)
+      RequestedBlocks.get(hash).map(x => x.nonEmpty && x.get.received),
+      // or block was already added by Casper (so was removed from RequestedBlocks)
+      casper.contains(hash)
```
tgrospic: This will execute casper check even if requested block check returns true. You can use "higher" _or_ combinator.
`RequestedBlocks.get(hash).map(x => x.nonEmpty && x.get.received) ||^ casper.contains(hash)`

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -281,28 +282,112 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
         )
-      )
-    } >>
-      casper.addBlock(b)
+      } >> {
+      for {
+        maybeReq <- RequestedBlocks.get(b.blockHash)
+        _ <- maybeReq match {
+              // There might be blocks that are not maintained by RequestedBlocks, e.g. genesis ceremony messages
+              case None => Log[F].info("BlockMessage is not present in RequestedBlocks.")
```
tgrospic: Block hash would be useful also in the log message.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -58,15 +55,16 @@ object CommUtil {
           .flatMap(
             requested =>
               Applicative[F].unlessA(requested.contains(hash))(
-                Running.addNewEntry(hash) >> sendToPeers(HasBlockRequestProto(hash)) <* Log[F]
+                Log[F]
                   .info(s"Requested missing block ${PrettyPrinter.buildString(hash)} from peers")
               )
+                >> Running.addNewEntry(hash) >> sendToPeers(HasBlockRequestProto(hash))
```
tgrospic: The logic seems changed, _addNewEntry_ and _sendToPeers_ are now outside of the check if already requested.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -47,9 +44,9 @@ object CommUtil {
     new CommUtil[F] {
 
       def sendBlockHash(hash: BlockHash, blockCreator: ByteString): F[Unit] =
-        sendToPeers(BlockHashMessageProto(hash, blockCreator)) <* Log[F].info(
+        Log[F].info(
           s"Sent hash ${PrettyPrinter.buildString(hash)} to peers"
-        )
+        ) >> sendToPeers(BlockHashMessageProto(hash, blockCreator))
```
tgrospic: Is there a reason why logs are now printed before the executing functions? It makes sense to know when the call finished. It's similar with other two changes in this file.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -789,7 +796,7 @@ object NodeRuntime {
           conf.roundRobinDispatcher.giveUpAfterSkipped,
           conf.roundRobinDispatcher.dropPeerAfterRetries
         )
-      }
+      }*/
```
tgrospic, line 797: Can we keep fairness constraint in code but disable it with CLI arguments?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -67,10 +67,12 @@ class NodeRuntime private[node] (
     eventLog: EventLog[Task]
 ) {
 
+  private[this] val numCPU = Math.max(java.lang.Runtime.getRuntime.availableProcessors(), 2)
+
   private[this] val loopScheduler =
     Scheduler.fixedPool("loop", 4, reporter = UncaughtExceptionLogger)
   private[this] val grpcScheduler =
-    Scheduler.cached("grpc-io", 4, 64, reporter = UncaughtExceptionLogger)
+    Scheduler.cached("grpc-io", 4, numCPU * 10, reporter = UncaughtExceptionLogger)
```
tgrospic: This thread pool should be only for async IO operations with lower amount of threads. Is it used for something else like processing messages?
For reference, good description about different pool configurations.
https://gist.github.com/djspiewak/46b543800958cf61af6efa8e072bfd5c

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
```diff
@@ -87,15 +87,18 @@ object StreamHandler {
...
         case (stmd, Right(Left(_))) =>
-          gracefullyClose[Task](stmd.fos).as(()) >>
+          Log[Task].warn("Failed collecting stream.") >>
+            gracefullyClose[Task](stmd.fos).as(()) >>
             stmd.path.deleteSingleFile[Task]
-        // should not happend (errors handled witin bracket) but covered for safety
+        // should not happen (errors handled within bracket) but covered for safety
         case (stmd, Left(_)) =>
-          gracefullyClose[Task](stmd.fos).as(()) >>
+          Log[Task].error("Unexpected scenario. Please pay attention.") >>
```
tgrospic: This doesn't inform you in which way you can proceed. Can you add something more specific?

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala
```diff
@@ -103,10 +96,12 @@ class GrpcTransportServer(
       tellConsumer <- Task.delay(
                        tellBuffer
                          .mapParallelUnordered(parallelism)(dispatchSend)
-                         .subscribe()(queueScheduler)
+                         .subscribe()(scheduler)
                      )
       blobConsumer <- Task.delay(
-                       blobBuffer.mapEval(dispatchBlob).subscribe()(queueScheduler)
+                       blobBuffer
+                         .mapParallelUnordered(parallelism)(dispatchBlob)
```
tgrospic, line 110: :tada:

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -30,6 +29,7 @@ object Running {
   final case class Requested(
       timestamp: Long,
       peers: Set[PeerNode] = Set.empty,
+      received: Boolean,
```
tgrospic: It makes sense to set default _false_ for received field.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -95,8 +96,20 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
 
     import cats.instances.either._
     for {
+      // Save block in block store before waiting for blockProcessingLock
+      // This is called inside block message handler as well, but required here in case block is not received from peer
+      // but proposed by node itself.
+      _ <- BlockStore[F].contains(b.blockHash).ifM(().pure[F], BlockStore[F].put(b))
+      _ <- Log[F].debug(
+            s"Block ${PrettyPrinter.buildString(b.blockHash)} asks for blockProcessingLock."
+          )
```
tgrospic, line 103: Maybe this log would be useful not only in debug?

tgrospic (changes requested):
Nice work and bug fix! Some comments about the logs and minor changes.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: It is used for all GRPC related operation - kademlia p2p discovery, p2p messaging, GRPC transport both packets and streaming (which end up adding blocks). I guess there should be several schedulers for different things.

nzpr: @tgrospic does this make sense? https://github.com/rchain/rchain/pull/2902/commits/05ade87d830d969316784b9b2db4166f3435410a. As we had growing number of unconsumed block messages - I wonder if it is a good idea to dedicate scheduler for that particular task.
Or do you want to lower down the upper bound for `grpc-io` scheduler as well?

nzpr: Weird but I can see on one of 5 nodes growing number of unprocessed messages
```sh
pdsh -w ^sandhosts "ls -l /var/lib/rnode/tmp/comm/ | wc -l"
nzpr@node1: 1
nzpr@node0: 1
nzpr@node4: 1
nzpr@node2: 85
nzpr@node3: 1
```

tgrospic: Let's keep `grpc-io` as it is. Before change, we must be sure that is only executing async IO. It was used to process incoming blocks so your change to another schedule will address this.

tgrospic: We have enough schedulers, maybe even to many. They are all competing for fixed number of processors so there is no magic that can do. 
There is no reason to have more schedulers for async IO operations, or even more threads, because they are not taking CPU time. Instead, we can have opposite effect, many IO threads can interrupt main computation where context switching is much more expensive.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
tgrospic: Is it possible that this is causing many unprocessed messages?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -281,28 +282,112 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+    * RequestedBlocks is a part of Casper loop maintaining liveness of the protocol.
+    * It keeping records of blocks being requested and re-request them on schedule if failed first time.
+    * It is expected to get lots of duplicate messages from different peers, so check for duplicate is required.
+    * Processing of duplicate blocks puts load on Casper and highly discouraged.
+    *
+    * When block is received from some peer first time - it is
+    * - marked as `received` in RequestedBlocks,
+    * - added to the block store and
+    * - sent to Casper.
+    * Block is being kept in RequestedBlocks until it is added to the DAG so we can detect duplicates.
```
zsluedem, line 300: Nice comments

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -95,8 +96,20 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
 
     import cats.instances.either._
     for {
+      // Save block in block store before waiting for blockProcessingLock
+      // This is called inside block message handler as well, but required here in case block is not received from peer
+      // but proposed by node itself.
```
zsluedem, line 99: good comments.

zsluedem (commented):
Very good. But the tests are still failing

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: Actually this commit made it worse, need to investigate.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
nzpr: I don't think so, `sendBlockRequest ` is called in two places when casper tries to fetch missing deps. And if block in block store - this wont be triggered
https://github.com/rchain/rchain/blob/a8ace3b28430f5af323bd0710eec0628cef297d2/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L379-L391
and 
https://github.com/rchain/rchain/blob/a442ed2fb6af9f4bab1b6118e91be4c94976dea6/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L476-L485

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: It made it worse because of this piece of code
https://github.com/rchain/rchain/pull/2902/commits/49f8f576da7754587e81cf38ea843d6ced08ecc8#diff-22a8d68f65fb447029ffa3431a23b529R366-R379
I pushed the fix https://github.com/rchain/rchain/pull/2902/commits/d02b9266f04e40bcf06f4f8e51795a4522cf47cd

nzpr: And this bad piece of code was less destructive when block messaging was handled on grpc-io scheduler. So I'm not sure anymore if we should push block handling to dedicated (or main) scheduler as per https://github.com/rchain/rchain/commit/05ade87d830d969316784b9b2db4166f3435410a

@tgrospic how do you think?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: I'm not sure thats an easy task if we use fairDispatcher instead of straight CasperPacketHandler. Or we have to have condition here. As I'm not sure we need fair dispatcher at all - I would prefer to enable it later if there is a need.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/b94764d552d88a30e0bea77be8c951688d9024ca

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/b94764d552d88a30e0bea77be8c951688d9024ca

comm/src/main/scala/coop/rchain/comm/transport/StreamHandler.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/b94764d552d88a30e0bea77be8c951688d9024ca

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/b94764d552d88a30e0bea77be8c951688d9024ca

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/7256c18e958c0f1103a138d69985ef4da5c0036d

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
tgrospic: > Either way, you're almost always going to have some sort of singleton object somewhere in your application which just has these three pools, pre-configured for use. If you ascribe to the "implicit ExecutionContext pattern", then you should make the CPU pool the implicit one, while the others must be explicitly selected.

From previous link but mentioned in the second also. :)

> If you are looking for less manual work - monix.eval.Task is great middleground which by default shifts tasks automatically in batches preserving both great throughput and decent latency off the shelf and exposes very rich configuration options if you have more advanced use case.
https://typelevel.org/cats-effect/concurrency/basics.html#thread-scheduling

Three schedulers, CPU bounded, implicit as we have main, async IO like our grpc-io and blocking, which we should not have because we use "shifting" for most flatMaps with monix Task.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/80c745cecfaa8f19ec8e29c0e62452a4e656bcad

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: Ok, so I'm going to revert https://github.com/rchain/rchain/commit/05ade87d830d969316784b9b2db4166f3435410a

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -281,28 +281,111 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+    */
+  private def repeatedCasperMessage(hash: BlockHash): F[Boolean] =
+    // If either RequestedBlocks shows that block was received (but not yet added to Casper)
+    // or block was already added by Casper (so was removed from RequestedBlocks)
+    (RequestedBlocks.get(hash).map(x => x.nonEmpty && x.get.received) ||^ casper.contains(hash))
+      .ifM(
+        // message is repeated
+        true.pure[F],
+        false.pure[F]
+      )
```
tgrospic: No need for `ifM`, `_ ||^ _: F[Boolean]`.

casper/src/test/scala/coop/rchain/casper/engine/RunningMaintainRequestedBlocksSpec.scala
```diff
@@ -35,7 +35,7 @@ class RunningMaintainRequestedBlocksSpec extends FunSpec with BeforeAndAfterEach
     describe("maintainRequestedBlocks, for every block that was requested") {
       describe("if block request is still within a timeout") {
         it("should keep the request not touch") {
-          val requested                = Requested(timestamp = notTimedOut)
+          val requested                = Requested(timestamp = notTimedOut, received = false)
```
tgrospic: On many places now received doesn't have to be set, default is false.

nzpr: https://github.com/rchain/rchain/pull/2902/commits/372522ce494a5493354b7a3505316204e77eb715

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: https://github.com/rchain/rchain/pull/2902/commits/372522ce494a5493354b7a3505316204e77eb715

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -31,9 +31,10 @@ class GrpcTransportClient(
     maxMessageSize: Int,
     packetChunkSize: Int,
     tempFolder: Path,
-    clientQueueSize: Int
+    clientQueueSize: Int,
+    parallelism: Int
 )(
-    implicit scheduler: Scheduler,
+    implicit ioScheduler: Scheduler,
```
tgrospic: Maybe it makes sense to keep only main scheduler as implicit and others as explicit parameters. It's important thing to be configured correctly and also it's already passed explicitly.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -46,12 +47,12 @@ class GrpcTransportClient(
   private def certInputStream  = new ByteArrayInputStream(cert.getBytes())
   private def keyInputStream   = new ByteArrayInputStream(key.getBytes())
   private val streamObservable = new StreamObservable(clientQueueSize, tempFolder)
-  private val parallelism      = Math.max(Runtime.getRuntime.availableProcessors(), 2)
   private val streamQueue =
     streamObservable
       .flatMap { s =>
         Observable
           .fromIterable(s.peers)
+          .executeOn(ioScheduler)
```
tgrospic: This is not simply io operation. That why previously this was fixed pool with the number of proc size. This should be new on the main scheduler. Parallelism as the number of proc is fine because Observable does not create real OS threads but just Task instances that can be run concurrently if there is not enough parallel threads.

nzpr:
I'm leaving schedulers tuning for future PR. We should investigate whats the best way.

tgrospic (approved):

zsluedem (approved):
Congratulation! Good one!

PR #2907, opened 2020-03-16
make ReportingCasper works with cost accounting

zsluedem:
## Overview
1. makes ReportingCasper work with cost accounting.
2. abstract reportingTransformer
3. add the blockEvent api in the BlockApi

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4056
sub task of
https://rchain.atlassian.net/browse/RCHAIN-4055

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
@@ -125,15 +125,19 @@ object ReportingCasper {
             AtomicAny(replayStore),
             Branch.REPLAY
           )
-          runtime          <- ReportingRuntime.createWithEmptyCost[F](reporting, Seq.empty)
-          manager          <- fromRuntime(runtime)
-          dag              <- BlockDagStorage[F].getRepresentation
-          bmO              <- BlockStore[F].get(hash)
+          runtime <- ReportingRuntime.createWithEmptyCost[F](reporting, Seq.empty)
+          manager <- fromRuntime(runtime)
+          dag     <- BlockDagStorage[F].getRepresentation
+          bmO     <- BlockStore[F].get(hash)
+          genesis <- BlockStore[F].getApprovedBlock
+          isGenesis = genesis
+            .flatMap(a => bmO.map(b => b.blockHash == a.candidate.block.blockHash))
+            .getOrElse(false)
           _                <- Log[F].info(s"trace block ${bmO}")
           invalidBlocksSet <- dag.invalidBlocks
           invalidBlocks    = invalidBlocksSet.map(block => (block.blockHash, block.sender)).toMap
           result <- bmO.traverse(
-                     bm => replayBlock(bm, dag, manager, invalidBlocks)
+                     bm => replayBlock(bm, dag, manager, invalidBlocks, isGenesis)
                    )
         } yield result
 
@@ -143,16 +147,16 @@ object ReportingCasper {
       block: BlockMessage,
       dag: BlockDagRepresentation[F],
       runtimeManager: ReportingRuntimeManagerImpl[F],
-      invalidBlocks: Map[BlockHash, Validator]
+      invalidBlocks: Map[BlockHash, Validator],
+      isGenesis: Boolean
   ): F[Either[ReplayFailure, List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]] = {
     val hash          = ProtoUtil.preStateHash(block)
     val deploys       = block.body.deploys
-    val systemDeploys = block.body.systemDeploys
     runtimeManager.replayComputeState(hash)(
       deploys,
-      systemDeploys,
       BlockData.fromBlock(block),
-      invalidBlocks
+      invalidBlocks,
+      isGenesis
     )
   }
 
@@ -208,9 +212,9 @@ object ReportingCasper {
 
     def replayComputeState(startHash: StateHash)(
         terms: Seq[ProcessedDeploy],
-        sysDeploys: Seq[ProcessedSystemDeploy],
         blockData: BlockData,
         invalidBlocks: Map[BlockHash, Validator],
+        isGenesis: Boolean
     ): F[Either[ReplayFailure, List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]] =
       Sync[F].bracket {
         runtimeContainer.take
@@ -222,9 +226,7 @@ object ReportingCasper {
                      runtime,
                      startHash,
                      terms,
-                     sysDeploys,
-                     replayDeploy(runtime, true)
-//                     replaySystemDeploy(runtime)
+                     replayDeploy(runtime, !isGenesis)
                    )
         } yield result
       }(runtimeContainer.put)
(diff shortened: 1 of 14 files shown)


rspace/src/main/scala/coop/rchain/rspace/ReportingTransformer.scala
```diff
@@ -0,0 +1,67 @@
...
+import coop.rchain.rspace.ReportingRspace.{
+  ReportingComm,
+  ReportingConsume,
+  ReportingEvent,
+  ReportingProduce
+}
+
+/**
+  * The purpose of the reportingTransformer is to create transformer to transform reporting events like
+  * `ReportingProduce`, `ReportingProduce` and `ReportingComm`(see coop.rchain.rspace.ReportingRspace) into
```
tgrospic: Typo with `ReportingProduce` which is mentioned two times.

rspace/src/main/scala/coop/rchain/rspace/ReportingRspace.scala
```diff
@@ -64,9 +76,31 @@ class ReportingRspace[F[_]: Sync, C, P, A, K](
...
 
-  val report: SyncVar[Seq[ReportingEvent]] = create[Seq[ReportingEvent]](Seq.empty)
+  /**
+    * in order to distinguish the system deploy(precharge and refund) in the a normal user deploy
+    * It might be more easily to analyse the report with data structure
+    * Seq(Seq[ReportingEvent](Precharge), Seq[ReportingEvent](userDeploy), Seq[ReportingEvent](Refund))
+    * It would be seperated by the softcheckpoint creation.
+    */
+  val report: SyncVar[Seq[Seq[ReportingEvent]]] = create[Seq[Seq[ReportingEvent]]](Seq.empty)
+  val softReport: SyncVar[Seq[ReportingEvent]]  = create[Seq[ReportingEvent]](Seq.empty)
```
tgrospic, line 86: Why is the difference for pre-charge to be in a nested Seq? Can we just concatenate all soft reports?

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -240,88 +173,262 @@ object ReportingCasper {
...
+        } yield result
       }(runtimeContainer.put)
 
     private def replayDeploys(
         runtime: ReportingRuntime[F],
         startHash: StateHash,
         terms: Seq[ProcessedDeploy],
-        replayDeploy: ProcessedDeploy => F[Either[ReplayFailure, Seq[ReportingEvent]]]
-    ): F[Either[ReplayFailure, List[(ProcessedDeploy, Seq[ReportingEvent])]]] =
+        systemDeploys: Seq[ProcessedSystemDeploy],
```
tgrospic: System deploys are not used. Without system deploys REV transactions to charge deploys are not visible in the report. Do we need ticket for that or is it necessary now?

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -186,10 +108,10 @@ object ReportingCasper {
     new ReportingCasper[F] {
       val codecK                                                     = serializeTaggedContinuation.toCodec
       implicit val m: RSpaceMatch[F, BindPattern, ListParWithRandom] = matchListPar[F]
-      val prettyPrinter                                              = RhoPrinter()
-      val transformer                                                = setupTransformer(prettyPrinter)
 
-      override def trace(hash: BlockHash): F[Report] =
+      override def trace(
+          hash: BlockHash
+      ): F[Option[Either[ReplayFailure, List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]]] =
```
tgrospic: Nested Option and Either are always suspicious. Here both define possible error, so `None` case can also be another `Left` case with some wrapper like `ReportError` which can have cases for block errors and case to wrap `ReplayFailure`. `Report` trait in `ReportingRoutes` can then be simplified because it's doing similar thing.

I'm more for ApplicativeErrors because they are much easier to extend and handle. This wrapper is a small step into this direction.

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -4,21 +4,92 @@ import cats.effect.Sync
...
+        s match {
+          case None => BlockNotFound(hash.base16String)
+          case Some((Left(internalError @ InternalError(_)))) =>
+            BlockReplayFailed(
+              hash.base16String,
+              internalError.toString,
+              None
+            )
+          case Some(Left(failed)) =>
+            BlockReplayFailed(hash.base16String, failed.toString, None)
```
tgrospic: These two case seems to have the same result.

tgrospic (commented):
Nice work! Especially including difficulties with protobuf types and partial functions. I left a few comments about that.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
zsluedem: No. The preCharge and refund system deploys are all included in the ProcessedDeploy.
The sytemDeploys here are just for closeBlock and slashDeploy. We don't need that now so I didn't put my effort on it. The soft reports above are  the ways the distinguish the precharge , refund and userDeploy.

These https://github.com/rchain/rchain/blob/dev/models/src/main/protobuf/CasperMessage.proto#L148 deploy logs includes precharge , refund and userDeploy logs.
 
`closeBlock` and `slashDeploy` deploy logs are in https://github.com/rchain/rchain/blob/dev/models/src/main/protobuf/CasperMessage.proto#L170. I don't actually like the design which put precharge, refund and userDeploy event logs altogether.

rspace/src/main/scala/coop/rchain/rspace/ReportingRspace.scala
zsluedem: `concatenate all soft reports` means single `Seq[ReportingEvent]` for a report and no soft report?  
The soft report here is used to distinguish the precharge , refund and userdeploy.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
tgrospic: You are right, we don't need close block and slash logs. Parameter can be removed altogether.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
zsluedem: @tgrospic I updated the embedded Option and Either. Take a look and see if that is what you want.

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

nzpr:
@zsluedem should we refactor later ReportingCasper to something that speaks more about providing access to data rather than consensus as we discussed in standup?

zsluedem:
> @zsluedem should we refactor later ReportingCasper to something that speaks more about providing access to data rather than consensus as we discussed in standup?

I don't understand your question actually. I don't understand the relation to  consensus

nzpr:
> > @zsluedem should we refactor later ReportingCasper to something that speaks more about providing access to data rather than consensus as we discussed in standup?
> 
> I don't understand your question actually. I don't understand the relation to consensus

I mean why it is called Reporting**Casper** if it is not related to consensus?

PR #2909, opened 2020-03-22
RCHAIN-4063 replace hardcoded api maxMessageSize to a parameter

zsluedem:
## Overview
replace hardcoded api maxMessageSize to a parameter

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4063

### Notes

### DIFF
--- a/node/src/main/resources/reference.conf
+++ b/node/src/main/resources/reference.conf
@@ -116,7 +116,7 @@ rnode {
     port-internal = 40402
 
     # Maximum size of message that can be sent via gRPC API
-    max-message-size = 4M
+    max-message-size = 16M
   }
 
   # Casper configuration
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -559,7 +559,8 @@ class NodeRuntime private[node] (
                             .acquireExternalServer[Task](
                               conf.grpcServer.portExternal,
                               grpcScheduler,
-                              apiServers.deploy
+                              apiServers.deploy,
+                              conf.grpcServer.maxMessageSize
                             )
       internalApiServer <- api
                             .acquireInternalServer(
@@ -567,7 +568,8 @@ class NodeRuntime private[node] (
                               grpcScheduler,
                               apiServers.repl,
                               apiServers.deploy,
-                              apiServers.propose
+                              apiServers.propose,
+                              conf.grpcServer.maxMessageSize
                             )
 
       prometheusReporter = new NewPrometheusReporter()
--- a/node/src/main/scala/coop/rchain/node/api/package.scala
+++ b/node/src/main/scala/coop/rchain/node/api/package.scala
@@ -16,16 +16,13 @@ import monix.execution.Scheduler
 
 package object api {
 
-  // 16 MB is max message size allowed by HTTP2 RFC 7540
-  // grpc and netty can however work with bigger values
-  val maxMessageSize: Int = 16 * 1024 * 1024
-
   def acquireInternalServer(
       port: Int,
       grpcExecutor: Scheduler,
       replGrpcService: ReplGrpcMonix.Repl,
       deployGrpcService: DeployServiceV1GrpcMonix.DeployService,
-      proposeGrpcService: ProposeServiceV1GrpcMonix.ProposeService
+      proposeGrpcService: ProposeServiceV1GrpcMonix.ProposeService,
+      maxMessageSize: Int
   ): Task[Server[Task]] =
     GrpcServer[Task](
       NettyServerBuilder
@@ -50,7 +47,8 @@ package object api {
   def acquireExternalServer[F[_]: Concurrent: Log: Taskable](
       port: Int,
       grpcExecutor: Scheduler,
-      deployGrpcService: DeployServiceV1GrpcMonix.DeployService
+      deployGrpcService: DeployServiceV1GrpcMonix.DeployService,
+      maxMessageSize: Int
   ): F[Server[F]] =
     GrpcServer[F](
       NettyServerBuilder


nzpr (commented):
@zsluedem Do we set default `conf.server.maxMessageSize` value somewhere? What will be `maxMessageSize` after this fix? I worry that it looks like `maxMessageSize` when used in `TransportServer` is supposed to be something small, and for large messages exchange between nodes we enable streaming. 
And for API calls I'm not sure how do we handle large deploys.

tgrospic:
@nzpr I think default is 4MB in [references.conf](https://github.com/rchain/rchain/blob/c142258a45b59e28a78c61918b018f310518011e/node/src/main/resources/reference.conf#L118-L119). Is this means max size for a deploy? Should default be 16MB in conf as it's now hardcoded?

We also have another `max-message-size`. Is this only for "protocol" messages? It's set to smaller value and also in chunks?
```sh
server {
  # Maximum size of messages that can be received via transport layer
  max-message-size = 256K

  # Chunk size for streaming packets between nodes
  packet-chunk-size = 256K
}
```
https://github.com/rchain/rchain/blob/2e570b8da710d34a2c852150377b9c4914fee382/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala#L80-L94

zsluedem:
@tgrospic @nzpr  Just want to share more test cases I did.

The test cases are based on this pr.
Server sends about 17M messages to client(python client I used).

1. Server with `grpc-max-message-size`= 40M , client with default `max_receive_message_length`= 4M. `GRpc Resource exhausted error` is raised.
2. Server with `grpc-max-message-size`= 40M , client with `max_receive_message_length`= 20M. message sent successfully
3. Server with `grpc-max-message-size`= 10M , client with default `max_receive_message_length`= 4M. `GRpc Resource exhausted error` is raised.
4. Server with `grpc-max-message-size`= 10M , client with default `max_receive_message_length`= 20M. message sent successfully

The `grpc-max-message-size` in this pr effects is only for receive side and doesn't restrict the send side. That's what is documented in [grpc document](https://github.com/grpc/grpc-java/blob/master/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L348)

Maybe we should distinguish the `client` and `server` parameter instead of using the same ones and also distinguish the `transport` layer and the `api` layer parameter(actually we did).

tgrospic:
> @deprecated Call {@link #maxInboundMessageSize} instead.

It seems that in the newer version of Netty this method is renamed to `maxInboundMessageSize` to be the same as in channel builder that we set in gRPC clients.

I don't think we need sending message limit. So important to distinguish should be for _transport_ and _api_, as we have now.
Maybe internal endpoint can be without limit?

zsluedem:
@tgrospic @nzpr Back to this pr, should we introduce the new param or just use the exist one?

tgrospic:
I don't think we need new parameter. But do we want to reduce limit on API messages. Hardcoded limit was 16MB and now default in config is 4MB. Should we change config value?

zsluedem:
> I don't think we need new parameter. But do we want to reduce limit on API messages. Hardcoded limit was 16MB and now default in config is 4MB. Should we change config value?

Yes. I changed the default value to 16Mb

nzpr:
IMO namings here are misleading here. 
https://github.com/rchain/rchain/blob/c142258a45b59e28a78c61918b018f310518011e/node/src/main/resources/reference.conf#L108-L120
as well as here
https://github.com/rchain/rchain/blob/1402b8800b7df57b8b3743f6764e7e986817c6f3/node/src/main/scala/coop/rchain/node/NodeRuntime.scala#L560
as GRPC is used for all communications. 
We have two `maxMessageSize` - in `servers` and `grpc` config section. And those `maxMessageSize` in `server` section is for grpc as well, but for transport and client serving p2p communication.

I'll try to suggest some refactoring in separate PR.

nzpr:
I'm curious what is the benefit to have maxMessageSize 256k instead of max 16MB for p2p connections? Why not 16MB? Is that safer? https://github.com/rchain/rchain/blob/1402b8800b7df57b8b3743f6764e7e986817c6f3/node/src/main/resources/reference.conf#L62

tgrospic:
> I'm curious what is the benefit to have maxMessageSize 256k instead of max 16MB for p2p connections? Why not 16MB? Is that safer?

It should minimize how much resource one message can take. We should be concerned if we don't have any limit. I'm not sure how big impact can have difference between 256k and 16MB.
Interestingly, I couldn't find better documentation.
https://docs.microsoft.com/en-us/aspnet/core/grpc/security?view=aspnetcore-3.1#message-size-limits

@nzpr I agree that naming is confusing and better to change it now then later.

tgrospic (approved):
LGTM

nzpr (approved):
LGTM
Further work for cleaning up config here https://github.com/rchain/rchain/pull/2912

nzpr:
@zsluedem  Looks like merging of https://github.com/rchain/rchain/pull/2907 makes one casper test fail on this PR. Look at staging branch GH action for this merge https://github.com/rchain/rchain/runs/544548071.

https://github.com/zsluedem/rchain/blob/1402b8800b7df57b8b3743f6764e7e986817c6f3/casper/src/test/scala/coop/rchain/casper/MultiParentCasperReportingSpec.scala#L21-L50

```
[info] - should behave the same way as MultiParentCasper *** FAILED *** (10 seconds, 314 milliseconds) 
[info]   1202 was not equal to 1203 (MultiParentCasperReportingSpec.scala:42)```

zsluedem:
> @zsluedem Looks like merging of #2907 makes one casper test fail on this PR. Look at staging branch GH action for this merge https://github.com/rchain/rchain/runs/544548071.
> 
> https://github.com/zsluedem/rchain/blob/1402b8800b7df57b8b3743f6764e7e986817c6f3/casper/src/test/scala/coop/rchain/casper/MultiParentCasperReportingSpec.scala#L21-L50
> 
> ```
> [info] - should behave the same way as MultiParentCasper *** FAILED *** (10 seconds, 314 milliseconds) 
> [info]   1202 was not equal to 1203 (MultiParentCasperReportingSpec.scala:42)```
> ```

The latest dev branch  is not like https://github.com/zsluedem/rchain/blob/1402b8800b7df57b8b3743f6764e7e986817c6f3/casper/src/test/scala/coop/rchain/casper/MultiParentCasperReportingSpec.scala#L23. The bug is known https://rchain.atlassian.net/browse/RCHAIN-4070. It is not related to the config changed. It is more about the reportingCasper

PR #2910, opened 2020-03-22
Do not consume repeated genesis candidate block

nzpr:
ATM during genesis ceremony Ceremony master repeatedly broadcasts candidate block to all known peers. Each of this block triggers new validation routine on peers, even if that block is already being validated. This PR introduces check for duplicate.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/GenesisValidator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/GenesisValidator.scala
@@ -14,11 +14,11 @@ import coop.rchain.comm.transport.TransportLayer
 import coop.rchain.comm.PeerNode
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.shared._
-import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.dag.BlockDagStorage
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedStorage
 import coop.rchain.casper.util.comm.CommUtil
+import coop.rchain.models.BlockHash.BlockHash
 
 class GenesisValidator[F[_]: Sync: Metrics: Span: Concurrent: CommUtil: TransportLayer: ConnectionsCell: RPConfAsk: Log: EventLog: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: LastApprovedBlock: BlockDagStorage: LastFinalizedStorage: EngineCell: RuntimeManager: Running.RequestedBlocks: EventPublisher: SynchronyConstraintChecker: LastFinalizedHeightConstraintChecker: Estimator: DeployStorage](
     validatorId: ValidatorIdentity,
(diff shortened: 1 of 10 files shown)


casper/src/main/scala/coop/rchain/casper/engine/GenesisValidator.scala
```diff
@@ -30,13 +40,26 @@ class GenesisValidator[F[_]: Sync: Metrics: Span: Concurrent: CommUtil: Transpor
   private val F    = Applicative[F]
   private val noop = F.unit
 
+  implicit val state = Cell.unsafe[F, Map[BlockHash, Boolean]](Map.empty)
```
tgrospic: Implicits are powerful abstraction. If it's not necessary, it can make code harder to reason about. `seen` and `ack` can be local functions accessing `state`.
E.g. as mentioned in the conclusion for implicit functions _please keep the Principle of Least Power in mind_.
https://www.scala-lang.org/blog/2016/12/07/implicit-function-types.html#conclusion

By the way, implicit functions look very nice for something like program configuration. I'm not sure if they will be available only in Dotty?

casper/src/main/scala/coop/rchain/casper/engine/GenesisValidator.scala
```diff
@@ -30,13 +40,26 @@ class GenesisValidator[F[_]: Sync: Metrics: Span: Concurrent: CommUtil: Transpor
...
-        Engine.transitionToInitializing(shardId, finalizationRate, Some(validatorId), init = noop)
-      }
+      GenesisValidator.State
+        .seen(ub.candidate.block.blockHash)(state)
+        .ifM(
+          Log[F].warn(
+            s"UnapprovedBlock ${PrettyPrinter.buildString(ub.candidate.block.blockHash)} is already being verified." +
+              s"Dropping duplicate message."
+          )
+            >> ().pure[F],
```
tgrospic: `Log.warn` already returns unit.

casper/src/main/scala/coop/rchain/casper/engine/GenesisValidator.scala
nzpr: I'm not sure I follow the gist about configuration. Do you mean CLI options, config files etc?

nzpr: @tgrospic https://github.com/rchain/rchain/pull/2910/commits/68993b6c159adb4dce40eeeed137d8f2d519a144 how about this?

tgrospic: Excellent!

I was referring to the use of implicits for configuration.
> In a functional setting, the inputs to a computation are most naturally expressed as parameters. One could simply augment functions to take additional parameters that represent configurations, capabilities, dictionaries, or whatever contextual data the functions need.

I remember you question, where should be the default values for configuration. It depends from where and in which moment someone is asking. Implementation of _Config_ implicit can load it from DB or expose mutation to the API.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -137,7 +137,7 @@ class GrpcTransportClient(
             log.debug(
               s"Error while streaming packet to $peer (timeout: ${timeout(packet).toMillis}ms): ${error.message}"
             )
-          case Right(_) => log.info(s"Streamed packet $path to $peer")
+          case Right(_) => log.debug(s"Streamed packet $path to $peer")
```
tgrospic, line 140: :+1:

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

PR #2912, opened 2020-03-29
RCHAIN-4100: Refactor config

nzpr:
Names used in config files, cmd line options and across codebase are often confusing. This is an attempt to clean up and make everything configurable to be exposed in config files. 

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4100


nzpr:
@zsluedem @tgrospic when you have time, please take a look on proposed `reference.conf` config file. If you have any suggestions on naming/structure, I'll adjust and then go forward updating CLI options and codebase.

nzpr:
@tgrospic might be you have better name for this or at least explanation what it is for
https://github.com/rchain/rchain/blob/c26b1ecbb5053c5b53652e70f8d36ddc07d62b90/node/src/main/resources/reference.conf#L141-L142

casper/src/main/scala/coop/rchain/casper/CasperConf.scala
```diff
@@ -20,7 +20,7 @@ final case class CasperConf(
     privateKey: Option[Either[String, Path]],
     bondsFile: Option[String],
     knownValidatorsFile: Option[String],
-    numValidators: Int,
+    autogenShardSize: Int,
```
zsluedem: why is `Shard`?

nzpr: This defines size of dummy bondsfile generated that will go into genesis block's bonds map. I believe each shard should have its own bonds map in a block. Or is that network-wide table?

zsluedem: I think `numActiveValidators` is more appropriate.

nzpr: `numActiveValidators ` sounds like parameter that defines the network  behaviour. But this one is just about how much identities to generate.

And there is param already `numberOfActiveValidators` that talks about casper behaviour

zsluedem: Ok. I mixed them up~!

node/src/test/scala/coop/rchain/node/configuration/commandline/ConfigMapperSpec.scala
```diff
@@ -3,260 +3,235 @@ package coop.rchain.node.configuration.commandline
...
+      .resources("defaults.conf")
+      .withFallback(
+        ConfigSource.string(
+          s"default-data-dir = /var/lib/rnode"
+        )
+      )
+
+    // Custom reader for PeerNode type
+    implicit val PeerNodeReader = ConfigReader.fromString[PeerNode](
+      ConvertHelpers.catchReadError(s => PeerNode.fromAddress(s).getOrElse(null))
```
tgrospic: Is there a reason to have fallback to `null`?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -2,264 +2,124 @@ package coop.rchain.node.configuration
...
-        config.hasPath(CertificateConfigKey),
-        config.hasPath(KeyConfigKey),
-        command,
-        profile.name,
-        configFile
-      )
-
-    configuration.copy(printHelpToConsole = () => options.printHelp())
-  }
+  final case class Profile(name: String, dataDir: (() => Path, String))
```
tgrospic: Why `dataDir` is a function?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -2,264 +2,124 @@ package coop.rchain.node.configuration
...
-
-  private val DataDirConfigKey     = s"${hocon.Server.Key}.${hocon.Server.keys.DataDir}"
-  private val CertificateConfigKey = s"${hocon.Tls.Key}.${hocon.Tls.keys.Certificate}"
-  private val KeyConfigKey         = s"${hocon.Tls.Key}.${hocon.Tls.keys.Key}"
-  private val GenesisPathConfigKey = s"${hocon.Casper.Key}.${hocon.Casper.keys.GenesisPath}"
-
-  def build(arguments: Seq[String]): Configuration = {
-    val configurationE = Try(buildConfiguration(arguments)).toEither
+    // Throw an error is unknown keys found
+    //implicit val hint = ProductHint[NodeConf](allowUnknownKeys = false)
```
tgrospic: What is the behavior if key is unknown?

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
```diff
@@ -2,264 +2,124 @@ package coop.rchain.node.configuration
...
-    blockstorage: FileLMDBIndexBlockStore.Config,
-    blockDagStorage: BlockDagFileStorage.Config,
-    kamon: Kamon,
-    printHelpToConsole: () => Unit = () => ()
-) {
-  def printHelp(): Task[Unit] = Task.delay(printHelpToConsole())
+  // within range HTTP2 RFC 7540
+  // TODO do we need this constraint? Let user set any but warn?
+  // private val MaxMessageSizeMinimumValue: Int = 256 * 1024 // 0.25 MB
+  // private val MaxMessageSizeMaximumValue: Int = 10 * 1024 * 1024
```
tgrospic, line 136: Good question. Maybe we can print warnings if configured size is not in standard range?

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -109,208 +103,251 @@ object Options {
...
+      required = false,
+      descr = "Path to latest messages data file"
+    )
 
-    val casperBlockDagStorageSize =
-      opt[Long](required = false, descr = "Casper BlockDagStorage map size (in bytes)")
+    val casperLatestMessagesCrcPath = opt[Path](
+      required = false,
+      descr = "Path to latest messages crc file"
+    )*/
```
tgrospic: Is this still needed?

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -341,60 +400,119 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
...
-    val finalizationRate = opt[Int](
-      descr = "Block finalization is triggered after adding every 'n' blocks. " +
-        "Use this option to configure this 'n'."
+    /*
+    TODO remove
+    val knownValidators = opt[String](
+        descr = "Plain text file listing the public keys of validators known to the user (one per line). " +
+        "Signatures from these validators are required in order to accept a block which starts the local" +
+        "node's view of the blockDAG."
+      )*/
```
tgrospic: Should this be removed? It's easy to add it later if needed, right?

tgrospic (commented):
Excellent work! I have just few minor comments.

Can you please create short info for node operators how to migrate to new options (names).

node/src/test/scala/coop/rchain/node/configuration/commandline/ConfigMapperSpec.scala
nzpr: I believe no, changed to, change is force pushed
https://github.com/rchain/rchain/blob/383f71c62cd2cd331bbb4cac0989fc82217ae4d6/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala#L69-L71

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
nzpr: It will just pass through it without any effect. Thats  bad but we need to support unknown keys to store Kamon config in `rnode.conf` next to node config.
Though I was thinking that we can add one more conf file and name it `kamon.conf` and leave `rnode.conf` under strict check. From one standpoint it should be good because that kamon config syntax is not determined by us. But we need additional config file in that case for configuring metrics.
What do you think?

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
nzpr: Removed

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
nzpr: I'm still not totally sure what for was this, might be its useful to debug something. But I'd remove it  if you are for it also.

node/src/main/resources/defaults.conf
```diff
@@ -0,0 +1,366 @@
...
+  # blocks proposed by peers simultaneously. Taking into account blocks are not the only protocol message,
+  # and blocks can be received as well via `requesting missing blocks` routine, it is recommended to
+  # make this number high. Upper bounded by the numer of messages (blocks are the biggest of them)
+  # node is able to keep in memory. Block size upper bound size is `max-stream-message-size`.
+  max-message-consumers = 400
+}
+
+protocol-client {
+  # Should be the same as server network ID.
+  network-id = ${protocol-server.network-id}
```
zsluedem, line 57: Since the client `network-id` and the server `network-id` should be the same. I think it is better to only config one param.

node/src/main/resources/defaults.conf
```diff
@@ -0,0 +1,366 @@
...
+
+  # Finalization is called every `finalization-rate` blocks
+  finalization-rate = 1
+
+  # Maximum number of block parents
+  max-number-of-parents = 2147483647
+
+  # Maximum depth of block parents. https://github.com/rchain/rchain/pull/2816
+  # Limits the depth of secondary parents in regards to the depth of the main parent.
+  max-parent-depth = 2147483647
```
zsluedem, line 203: Just curious. How do you get this number? I saw 7-depth in the link you posted

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -109,208 +103,251 @@ object Options {
...
+      descr = "Float value representing that the node tolerates up " +
+        "to fault-tolerance-threshold fraction of the total weight to equivocate."
+    )
 
     val validatorPublicKey = opt[String](
       descr = "Base16 encoding of the public key to use for signing a proposed blocks. " +
         "Can be inferred from the private key for some signature algorithms."
     )
 
+    // We should not encourage exposing private keys in such way, so this param is hidden
```
zsluedem, line 342: Just want to say that it make test easy with this param.

scripts/p2p-test-tool.py
```diff
@@ -28,7 +28,7 @@
 parser.add_argument("--bootstrap-command",
                     dest='bootstrap_command',
                     type=str,
-                    default="run --port 40400 --standalone",
+                    default="run --protocol-port 40400 --standalone",
```
zsluedem, line 31: Just want to say that I don't think we are still using this file.

scripts/genesis-test/generate-setup
```diff
@@ -90,7 +90,7 @@ parser.add_argument(
 # Genesis block approval parameters
 # https://rchain.atlassian.net/wiki/spaces/CORE/pages/485556483/Initializing+the+Blockchain+--+Protocol+for+generating+the+Genesis+block
 parser.add_argument(
-        '--required-sigs',
+        '--required-signatures',
```
zsluedem, line 93: Just want to say that I don't think we are still using this file.

zsluedem (commented):
Very fantastic work! I left some comments.

One more thing is that, it looks like we are doing a hard update. Since currently only coop is running the node now, it would be better we migrate in a softer way in the future if we are doing big change like this. What I mean softer way is that the old and new configs should be both exist for some versions. I think the wallet app developers also starts their own node now. We'd better inform them the changes.

node/src/main/resources/defaults.conf
nzpr: I wanted to as much as possible show all info that is required for different parts of the system. Protocol client requires to be initiated with right network-id for its messages to be accepted by peers. Might be its too hard wording `Should be the same as server network ID`, its just an expected behaviour. I'd like to keep this duplicate, as a part of attempt to make node modular.

node/src/main/resources/defaults.conf
nzpr: I just used `max-number-of-parents` value as `some random big number` to make effectively `max-parent-depth` effectively unlimited. This option is experimental and I we don't have any justification for why it's needed and what number this should be. So I made it unlimited by default.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
nzpr: No idea, refactored it to be plain Path

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
nzpr: I cannot find anything related to this range 0.25 - 10 MB in RFC 7540.

scripts/p2p-test-tool.py
nzpr: Do you think we should remove it? Looks like there is some useful code there. I don't want to hoard stuff but it's in `script` folder so does not hurt imo.

scripts/genesis-test/generate-setup
nzpr: Might be we should clean all stuff from `scripts` as outdated, but I suggest to postpone it for now.

nzpr:
@tgrospic I'll update https://github.com/rchain-community/rchain-docker-shard and write some notice letter.
@zsluedem I agree we need make such changes smooth, indeed. I hope future changes will be much smaler.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
nzpr: I enabled strict check and moved kamon conf to separate file. Path for it is not configurable,it should be next to `rnode.conf` and called `kamon.conf` https://github.com/rchain/rchain/pull/2912/commits/67e60ab94ee993659a10910f140c0715799d5981. Now if there are unknown key in config file node won't start and throw an error.

nzpr:
I've added a wiki page on [how to add RNode server configuration parameter](https://github.com/rchain/rchain/wiki/HOW-TO:-Adding-configuration-parameter)

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -30,65 +30,108 @@ object Main {
...
     Thread.setDefaultUncaughtExceptionHandler((thread, ex) => {
       LoggerFactory.getLogger(getClass).error("Unhandled exception in thread " + thread.getName, ex)
     })
 
-    val configuration = Configuration.build(args)
-    System.setProperty("rnode.data.dir", configuration.server.dataDir.toString) // NonUnitStatements
-
+    // Main scheduler for all CPU bounded tasks
+    // Should always be passed as implicit dependency.
+    // All other schedulers should be explicit.
```
tgrospic, line 48: :+1:

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -128,26 +170,73 @@ object Main {
...
     ) >> program
   }
 
+  private def subcommand(options: commandline.Options): Command =
+    options.subcommand match {
+      case Some(options.eval) =>
+        Eval(options.eval.fileNames(), options.eval.printUnmatchedSendsOnly())
+      case Some(options.repl)   => Repl
+      case Some(options.deploy) =>
+        //TODO: change the defaults before main net
```
tgrospic: This comment is obsolete.

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
```diff
@@ -44,27 +45,55 @@ object ValidatorIdentity {
       conf: CasperConf,
       privateKey: PrivateKey
   ): F[Option[ValidatorIdentity]] = {
-    val maybePublicKey = conf.publicKeyBase16.map(pk => PublicKey(Base16.unsafeDecode(pk)))
+    val maybePublicKey = conf.validatorPublicKey.map(pk => PublicKey(Base16.unsafeDecode(pk)))
```
tgrospic: Public key can be extracted from private key. So this function only need `PrivateKey` parameter.
`val publicKey = Secp256k1.toPublic(privateKey)`

Then can also be renamed to `apply` as it's the main _constructor_ for `ValidatorIdentity`.

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
```diff
@@ -44,27 +45,55 @@ object ValidatorIdentity {
...
+      publicKey(maybePublicKey, Secp256k1.name, privateKey),
+      privateKey,
+      Secp256k1.name
+    ).some.pure[F]
   }
 
-  def fromConfig[F[_]: Sync: EnvVars: Log](conf: CasperConf): F[Option[ValidatorIdentity]] =
-    conf.privateKey match {
-      case Some(Left(privateKeyBase16)) =>
+  def fromPrivateKey[F[_]: Sync: EnvVars: Log](conf: CasperConf): F[Option[ValidatorIdentity]] =
```
tgrospic: With the change to ValidatorIdentity, only `conf.validatorPrivateKey` is required for this function and not the whole config.

In general, we don't want to write functions like `Option[A] -> Option[B]`. What we want are functions that can be composed with monads, so in this case `String -> Option[ValidatorIdentity]` is much more useful and "effect" of logging when input is not available should be done outside.

---
Also `unsafeDecode` is used to parse input hex string. It doesn't make sense to check for input string existence and report an error, and throw an error if string is not valid base16 encoding (what is possible because is coming from config).

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
```diff
@@ -44,27 +45,55 @@ object ValidatorIdentity {
...
-          validatorIdentity <- createValidatorIdentity(conf, privateKey)
-        } yield validatorIdentity
       case None =>
         Log[F]
           .warn("No private key detected, cannot create validator identification.")
           .map(_ => none[ValidatorIdentity])
     }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  def publicKey(
```
tgrospic: This function will not be used anymore.

Please don't throw exceptions, suppress warning should only be used for legacy code.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
tgrospic: Now that we will store block data to lmdb this can be removed for sure.

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -317,18 +395,29 @@ object Main {
...
+  private def logConfiguration(conf: NodeConf, options: commandline.Options): Task[Unit] = {
+    val profile = Configuration.profiles.getOrElse(
+      options.profile.getOrElse("default"),
+      Configuration.profiles.get("default").get
+    )
+    val dataDir = options.run.dataDir.getOrElse(profile.dataDir._1)
+    val configPath = options.run.configFile
+      .getOrElse(
+        dataDir.resolve("rnode.conf")
+      )
```
tgrospic: Resolving profile is very similar to code here. It would be much better if we can do this logic only in one place.
https://github.com/rchain/rchain/blob/bf0bbd3cb3ee8cb678569011b3f17d72b1220ff2/node/src/main/scala/coop/rchain/node/configuration/Configuration.scala#L26-L33

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -109,208 +103,251 @@ object Options {
 final case class Options(arguments: Seq[String]) extends ScallopConf(arguments) {
   import Converter._
   import Options.Flag
-  import Options.ports._
+  val width = 120
```
tgrospic, line 108: Please add some explanation why this should be 120.

node/src/main/scala/coop/rchain/node/configuration/model.scala
```diff
@@ -2,66 +2,85 @@ package coop.rchain.node.configuration
...
+    standalone: Boolean,
+    protocolServer: ProtocolServer,
+    protocolClient: ProtocolClient,
+    peersDiscovery: PeersDiscovery,
+    apiServer: ApiServer,
+    tls: TlsConf,
+    storage: Storage,
+    casper: CasperConf,
+    metrics: Metrics,
+    // This field is dynamic and computer according to profile and is not used in client code
```
tgrospic: A typo in `computer`.

node/src/main/scala/coop/rchain/node/configuration/model.scala
```diff
@@ -2,66 +2,85 @@ package coop.rchain.node.configuration
...
+    // But it is required in the model to enable strict configuration check for unknown keys
+    defaultDataDir: String
```
tgrospic, line 34: How this "strict check" works? Who is doing the checking and when?

node/src/test/scala/coop/rchain/node/configuration/hocon/HoconConfigurationSpec.scala
```diff
@@ -2,264 +2,154 @@ package coop.rchain.node.configuration.hocon
...
+          s"default-data-dir = /var/lib/rnode"
+        )
+      )
 
-    val expectedServer =
-      configuration.Server(
+    // Custom reader for PeerNode type is required
+    implicit val PeerNodeReader = ConfigReader.fromString[PeerNode](
+      ConvertHelpers.catchReadError(s => PeerNode.fromAddress(s).getOrElse(null))
+    )
```
tgrospic, line 42: Using `null` can be avoided here.
```scala
implicit val PeerNodeReader = ConfigReader.fromStringTry[PeerNode](
  ConvertHelpers.catchReadError(PeerNode.fromAddress(_).toTry)
)
```

node/src/test/scala/coop/rchain/node/configuration/commandline/ConfigMapperSpec.scala
tgrospic: This will not propagate the error from `Left(err)` returned from `fromAddress`. There is `ConfigReader.fromStringTry` that is convenient in this case because Either[Throwable, A] is similar to `Try[A]`.
I commented on the third place where similar code is used. :)

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
tgrospic: I agree to have `kamon.conf` in a separate file.

Is it possible to later add CLI flag to disable strict check? E.g. editor plugins supports multiple node versions and tracking what key is correct in which version can be challenging.

node/src/test/scala/coop/rchain/node/configuration/hocon/HoconConfigurationSpec.scala
nzpr: I changed this is other places to
```scala
implicit val PeerNodeReader = ConfigReader.fromString[PeerNode](
      ConvertHelpers.catchReadError(s => PeerNode.fromAddress(s).right.get)
    )
```
are you ok with this option?

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
nzpr: Are you ok if I fix this in https://github.com/rchain/rchain/pull/2915? It's more suitable for this change imo.

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
nzpr: Are you ok if I fix this in https://github.com/rchain/rchain/pull/2915? It's more suitable for this change imo.

casper/src/main/scala/coop/rchain/casper/ValidatorIdentity.scala
nzpr: Are you ok if I fix this in https://github.com/rchain/rchain/pull/2915? It's more suitable for this change imo.

node/src/main/scala/coop/rchain/node/Main.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/main/scala/coop/rchain/node/Main.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/test/scala/coop/rchain/node/configuration/hocon/HoconConfigurationSpec.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/7b1015b6b360d2d044f73a9546b773fb82a1d712

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
tgrospic: Why this must be specified? Should this works automatically with console width? What if console is less than 120 chars? 
I'm trying to understand reason why fixed number helps.

node/src/test/scala/coop/rchain/node/configuration/hocon/HoconConfigurationSpec.scala
tgrospic: I made a comment there https://github.com/rchain/rchain/pull/2912#discussion_r414403759.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
nzpr: Default width is 80 and it just does not look good, lines are too short. I believe if console is less then this `120` won't take effect. Here's how its now with default width, imo too narrow.
Screenshot 2020-04-24 at 13 41 58

node/src/test/scala/coop/rchain/node/configuration/commandline/ConfigMapperSpec.scala
nzpr: Ah, understood, will push fix now, there are 3 places where this is used.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
nzpr: Separate Kamon file is pushed.

But if user have some key in configuration - he's expecting for it to take an effect. And if we don't rise an error for unknown keys, it might be confusing.

node/src/test/scala/coop/rchain/node/configuration/hocon/HoconConfigurationSpec.scala
nzpr: Fixed

node/src/test/scala/coop/rchain/node/configuration/commandline/ConfigMapperSpec.scala
nzpr: Fixed

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
tgrospic: Ok, that's reasonable. Please put this explanation in comments. :)

node/src/main/scala/coop/rchain/node/configuration/model.scala
tgrospic: What will happen if you remove this field? I'm not sure how can someone understand where is the connection to behavior you mentioned in the comment.

node/src/main/scala/coop/rchain/node/configuration/Configuration.scala
tgrospic: This is true is you are assuming that human is running the node, but if you want to write program that control the node it can be much more difficult to manage configuration.
E.g. when `--allow-private-addresses` was added to RNode, all previous versions would refuse to start with this parameter although there is no real problem.

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: Looks like I should to improve wordings. 
https://github.com/rchain/rchain/blob/38596efe3b9fb79a8651bf2bf1a0999337094e9f/node/src/main/resources/defaults.conf#L130 Here is the default hocon file, it should be loaded into NodeConf case lass. This `default-data-dir` is initialized from profile provided through CLI and added to default config. So its in HOCON file. And therefore it should be in model as well, or when trying to create model an error will be thrown by Pureconfig - `unknown key`. And we need these unknown key to throw errors to make sure mistypes in config file are detected.

nzpr: Adjusted comment in https://github.com/rchain/rchain/pull/2912/commits/011f6cbe73a10cfd14344ca3cd2b1c1a1572d712

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
nzpr: https://github.com/rchain/rchain/pull/2912/commits/011f6cbe73a10cfd14344ca3cd2b1c1a1572d712

node/src/main/scala/coop/rchain/node/configuration/model.scala
tgrospic: Thanks for additional config and explanation.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -535,59 +536,66 @@ class NodeRuntime private[node] (
...
-                              conf.server.port,
-                              conf.tls.certificate,
-                              conf.tls.key,
-                              conf.server.maxMessageSize,
-                              conf.server.maxStreamMessageSize,
-                              conf.server.dataDir.resolve("tmp").resolve("comm"),
-                              conf.server.messageConsumers
+                              //conf.protocolServer
+                              //conf.tls
+                              //conf.storage
```
tgrospic, line 550: Leftover comments.

node/src/main/scala/coop/rchain/node/configuration/commandline/ConfigMapper.scala
```diff
@@ -19,107 +18,101 @@ object ConfigMapper {
...
+      add("tls.certificate-path", run.tlsCertificatePath)
+      add("tls.secure-random-non-blocking", run.tlsSecureRandomNonBlocking)
+
+      add("metrics.prometheus", run.prometheus)
+      add("metrics.influxdb", run.influxdb)
+      add("metrics.influxdb-udp", run.influxdbUdp)
+      add("metrics.zipkin", run.zipkin)
+      add("metrics.sigar", run.sigar)
+      //TODO remove
+      //add(keys.KnownValidatorsFile, run.knownValidators
```
tgrospic, line 113: Leftover comments.

tgrospic (approved):
LGTM, thanks for taking care of all comments.

zsluedem (approved):
Very good work!

PR #2913, opened 2020-03-31
RCHAIN-4054: Last finalized state

tgrospic:
## Overview
Node should be able to request last finalized state from other nodes in the network and continue execution from that point.

NOTE: Details about specification will be written later as part of RCHIP.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4054


nzpr:
@tgrospic I'm curious whats the Request Result and Ticket in this trait https://github.com/nzpr/rchain/blob/40324d4a8bae8906eeaeba1f211b87a8ef9c3cd7/shared/src/main/scala/coop/rchain/state/StateManager.scala ?
I'm assuming this API is supposed to be used during catch up process, am I thinking right direction?

tgrospic:
@nzpr you are right. _Synchronizer_ instance should have logic to create requests for partial state, combine results and verify. And also it should provide current status that can be queried over admin API.

I'd like for _StateManager_ to abstract over all storages like BlockStore, BlockDagStorage etc. To reduce number of implicit parameters (like on Engine implementations) but more importantly to provide control for read and write operations.
If we can ensure that some function can only read from the storage it can be more easily optimized and run concurrently.
For example, with dynamic configuration it will be very important to know if a function can change configuration or it's just a reader. To enable/disable casper is much more than just switching the flag.

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedState.scala
```diff
@@ -0,0 +1,162 @@
...
+  import LastFinalizedState._
+
+  private val reqPath = Ref.unsafe(Seq((stateHash, none[Byte])))
+
+  private val semaphore = Semaphore[F](1)
+
+  override def init: F[Unit] = ().pure[F]
+
+  override def handle(peer: PeerNode, msg: CasperMessage): F[Unit] = msg match {
+    case StoreItemsMessage(startPath, lastPath, historyItems, dataItems) =>
```
nzpr, line 63: `Store` here means RSpace coldStore?

nzpr (approved):
Great foundation! FS2 streams is cool thing. I'll continue in another PR

zsluedem (approved):
LGTM

zsluedem:
@tgrospic Do you think it is safe to merge this into dev since the last finalized codes are not reachable when the node is running?

tgrospic:
bors merge

PR #2917, opened 2020-04-28
RCHAIN-4076 add hotstore for reporting data

zsluedem:
## Overview
~~optimization.(too lazy, save time to develop. Emergent task)~~

Optimize the performance of reporting api. Save the result in memory to avoid replaying the block all the time

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4076

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
@@ -102,7 +102,7 @@ object ReportingCasper {
   type RhoReportingRspace[F[_]] =
     ReportingRspace[F, Par, BindPattern, ListParWithRandom, TaggedContinuation]
 
-  def rhoReporter[F[_]: Sync: ContextShift: Monad: Concurrent: Log: Metrics: Span: Parallel: BlockStore: BlockDagStorage](
+  def rhoReporter[F[_]: ContextShift: Concurrent: Log: Metrics: Span: Parallel: BlockStore: BlockDagStorage](
       historyRepository: RhoHistoryRepository[F]
   )(implicit scheduler: ExecutionContext) =
     new ReportingCasper[F] {
@@ -149,7 +149,7 @@ object ReportingCasper {
                      case Left(replayError) =>
                        Log[F].info(
                          s"Relay ${PrettyPrinter.buildStringNoLimit(block.blockHash)} error ${replayError} from reporting"
-                       ) >> Sync[F].delay(
+                       ) >> Concurrent[F].delay(
                          ReportReplayError(replayError)
                            .asLeft[List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]
                        )
@@ -171,7 +171,7 @@ object ReportingCasper {
           _          <- Log[F].info(s"trace block ${maybeBlock}")
           result <- maybeBlock match {
                      case None =>
-                       Sync[F].delay(
+                       Concurrent[F].delay(
                          ReportBlockNotFound(hash)
                            .asLeft[List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]
                        )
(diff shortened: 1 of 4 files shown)


nzpr:
@zsluedem I'm curious why did not you use https://github.com/rchain/rchain/blob/e3bb9c1180102aef0df1bad463fd483bc0ebd0ef/shared/src/main/scala/coop/rchain/shared/Compression.scala#L7-L30 for https://github.com/rchain/rchain/pull/2917/commits/1a76bafc6abed804039c9bb8875ebea51b5ee9c5?

zsluedem:
> @zsluedem I'm curious why did not you use
> 
> https://github.com/rchain/rchain/blob/e3bb9c1180102aef0df1bad463fd483bc0ebd0ef/shared/src/main/scala/coop/rchain/shared/Compression.scala#L7-L30
> 
> for [1a76baf](https://github.com/rchain/rchain/commit/1a76bafc6abed804039c9bb8875ebea51b5ee9c5)?

Didn't know

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -105,18 +102,25 @@ object ReportingCasper {
   type RhoReportingRspace[F[_]] =
     ReportingRspace[F, Par, BindPattern, ListParWithRandom, TaggedContinuation]
 
-  def rhoReporter[F[_]: ContextShift: Monad: Concurrent: Log: Metrics: Span: Parallel: BlockStore: BlockDagStorage](
+  def rhoReporter[F[_]: Sync: ContextShift: Monad: Concurrent: Log: Metrics: Span: Parallel: BlockStore: BlockDagStorage](
```
tgrospic: `Sync` and `Monad` are not needed here, they are part of `Concurrent` trait. `Sync[F].delay` can be instead `Concurrent[F].delay`.

casper/src/main/scala/coop/rchain/casper/ReportHotStore.scala
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+  This is a temporary solution for the performance problem which is caused by the
+  reporting api. Exchanges highly rely on this api to get transactions info.But
+  every time request to the event data, the node would replay the block once again
+  which makes the api so slow and can not processed concurrently.
+
+  The ReportHotStore is a in memory store for event data in every block.
+  */
+trait ReportHotStore[F[_]] {
```
tgrospic: The name HotStore can be confusing because HotStore in RSpace is used to collect changes for event log (hot store actions).

node/src/main/scala/coop/rchain/node/api/package.scala
```diff
@@ -43,6 +43,7 @@ package object api {
             .bindService(deployGrpcService, grpcExecutor)
         )
         .addService(ProtoReflectionService.newInstance())
+        .compressorRegistry(null)
```
tgrospic, line 46: We have compression defined here for the protocol transport. I'm not sure what this is doing?
https://github.com/rchain/rchain/blob/13b2c6e0ae90728375df44564251091c00b224eb/comm/src/main/scala/coop/rchain/comm/transport/Chunker.scala#L17-L25

zsluedem: I think that's only for the casperMessage.
The deployService doesn't support that now.
They are two different services.

nzpr: And not for all CasperMessage but only for streamed messages blocks.

tgrospic: I meant this just as an example. Can we enable this on deploy service or `compressorRegistry(null)` is doing this?

zsluedem: `compressorRegistry(null)` should do the trick.

casper/src/main/scala/coop/rchain/casper/ReportHotStore.scala
nzpr: @zsluedem might be ReportingCache?

tgrospic (approved):
Very nice! :+1:

nzpr (approved):

PR #2926, opened 2020-05-04
RCHAIN-4080: Optimizations for BlockStore and Engine

tgrospic:
## Overview
This PR addresses issues with growing memory due to blocks held in memory during retrival.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4080

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -1,6 +1,7 @@
 package coop.rchain.casper.engine
 
 import cats.effect.Sync
+import cats.effect.concurrent.Ref
 import cats.syntax.all._
 import cats.{Applicative, Monad}
 import com.google.protobuf.ByteString
@@ -201,11 +202,14 @@ object Running {
         )
     )
 
-  def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks](peer: PeerNode, b: BlockMessage)(
+  def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks: BlockStore](
+      peer: PeerNode,
+      b: BlockMessage
+  )(
       repeatedCasperMessage: BlockHash => F[Boolean],
       casperAdd: BlockMessage => F[ValidBlockProcessing]
   ): F[Unit] =
-    repeatedCasperMessage(b.blockHash)
+    (repeatedCasperMessage(b.blockHash) ||^ BlockStore[F].contains(b.blockHash))
       .ifM(
         Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
         Log[F].info(s"Received ${PrettyPrinter.buildString(b)}.") >> casperAdd(b) >>= (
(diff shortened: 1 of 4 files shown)


casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -201,11 +202,14 @@ object Running {
...
-  def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks](peer: PeerNode, b: BlockMessage)(
+  def handleBlockMessage[F[_]: Monad: Log: RequestedBlocks: BlockStore](
+      peer: PeerNode,
+      b: BlockMessage
+  )(
       repeatedCasperMessage: BlockHash => F[Boolean],
       casperAdd: BlockMessage => F[ValidBlockProcessing]
   ): F[Unit] =
-    repeatedCasperMessage(b.blockHash)
+    (repeatedCasperMessage(b.blockHash) ||^ BlockStore[F].contains(b.blockHash))
```
nzpr: Might be this should be part of `repeatedCasperMessage` itself. But we need to add logic that node process that blocks without requesting them again.

tgrospic: Makes sense.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -201,20 +202,27 @@ object Running {
...
       casperAdd: BlockMessage => F[ValidBlockProcessing]
-  ): F[Unit] =
-    repeatedCasperMessage(b.blockHash)
+  ): F[Unit] = {
+    val alreadyInStore = isInBlockStore(b.blockHash)
+    (repeatedCasperMessage(b.blockHash) ||^ alreadyInStore)
       .ifM(
-        Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again."),
+        Log[F].info(s"Received block ${PrettyPrinter.buildString(b.blockHash)} again.") >>
+          alreadyInStore.ifM(RequestedBlocks.remove(b.blockHash), ().pure[F]),
```
tgrospic, line 217: @nzpr I seen messages re-requested even when in the block store so that's why they are removed here also.

nzpr: Logic was that hash should be be removed from RequestedBlocks only after it is added to DAG. But we should revisit this logic

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: Actually no, I'm wrong - repeatedCasperMessage means casper message is being processed right now and node should not process it again. If node is restarted but block is in blockstore - node should process it again if it is still not in Casper.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: This should not be done, because if block in store it might be not in process of adding to casper - nor added to casper

nzpr:
Looks like the issue is not a small one and requires quite a lot of changes, so rebasing this on https://github.com/tgrospic/rchain/tree/RCHAIN-4054-last-finalized-state to persist CasperState.

tgrospic:
Superseded by #2943.

PR #2927, opened 2020-05-05
RCHAIN-4081: Specify a starting point to visualize dag 

zsluedem:
## Overview
Right now, the start point of the vdag is the latest block. It would be better that we can visualize the dag from anywhere we want for debugging.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4081

### Notes

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
+++ b/casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
@@ -220,7 +220,7 @@ class TestNode[F[_]](
 
   def shutoff() = transportLayerEff.clear(local)
 
-  def visualizeDag(): F[String] = {
+  def visualizeDag(startBlockNumber: Int): F[String] = {
 
     type G[A] = State[StringBuffer, A]
     import cats.mtl.implicits._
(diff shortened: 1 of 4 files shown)


nzpr (commented):
So with this PR we can print DAG not for the most recent part but somewhere in the middle, right? To me `visual specify start point` sounds a bit misleading. Why `visual`? And looks like we set the end point of the vDAG?

casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
```diff
@@ -220,7 +220,7 @@ class TestNode[F[_]](
 
   def shutoff() = transportLayerEff.clear(local)
 
-  def visualizeDag(): F[String] = {
+  def visualizeDag(startBlockNumber: Int): F[String] = {
```
nzpr, line 223: What is this number supposed to be? Is it offset relative to the latest block? Or it is `seqNum`?

zsluedem:
> So with this PR we can print DAG not for the most recent part but somewhere in the middle, right? To me `visual specify start point` sounds a bit misleading. Why `visual`? And looks like we set the end point of the vDAG?

Yes. You are right. With this PR, we can print the dag anywhere we want. 
I think I meant visualize.

casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
zsluedem: This is just the test. Actually this method is not in use anywhere. I think it is probably for debugging.

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

PR #2933, opened 2020-05-12
RCHAIN-4089: Add configs for api server

zsluedem:
## Overview
This PR adds additional config for gRPC API servers to control connection timeout.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4089

### DIFF
--- a/node/src/main/resources/defaults.conf
+++ b/node/src/main/resources/defaults.conf
@@ -123,6 +123,27 @@ api-server {
 
   # Enable reporting API
   enable-reporting = false
+
+  # Sets a custom keepalive time, the delay time for sending next keepalive ping
+  keep-alive-time = 2 hours
+
+  # Sets a custom keepalive timeout, the timeout for keepalive ping requests
+  keep-alive-timeout = 20 seconds
+
+  # The most aggressive keep-alive time clients are permitted to configure.
+  # The server would close the connection if clients exceeding this rate
+  permit-keep-alive-time = 5 minutes
+
+  # Sets a custom max connection idle time, connection being idle for longer than which will be gracefully terminated
+  max-connection-idle = 1 hours
+
+  # Sets a custom max connection age, connection lasting longer than which will be gracefully terminated
+  max-connection-age = 1 hours
+
+  # Sets a custom grace time for the graceful connection termination. Once the max connection age
+  # is reached, RPCs have the grace time to complete. RPCs that do not complete in time will be
+  # cancelled, allowing the connection to terminate
+  max-connection-age-grace = 1 hours
 }
 
 storage {
--- a/node/src/main/scala/coop/rchain/node/api/package.scala
+++ b/node/src/main/scala/coop/rchain/node/api/package.scala
@@ -1,6 +1,7 @@
 package coop.rchain.node
 
 import java.net.InetSocketAddress
+import java.util.concurrent.TimeUnit
 
 import cats.effect.Concurrent
 import coop.rchain.casper.protocol.deploy.v1.DeployServiceV1GrpcMonix
@@ -15,6 +16,8 @@ import io.netty.channel.local.LocalAddress
 import monix.eval.Task
 import monix.execution.Scheduler
 
+import scala.concurrent.duration.FiniteDuration
+
 package object api {
 
   def acquireInternalServer(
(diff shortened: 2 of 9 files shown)


node/src/main/resources/defaults.conf
```diff
@@ -123,6 +123,27 @@ api-server {
...
+  # Sets a custom max connection idle time, connection being idle for longer than which will be gracefully terminated
+  max-connection-idle = 9223372036854775807
+
+  # Sets a custom max connection age, connection lasting longer than which will be gracefully terminated
+  max-connection-age = 9223372036854775807
+
+  # Sets a custom grace time for the graceful connection termination. Once the max connection age
+  # is reached, RPCs have the grace time to complete. RPCs that do not complete in time will be
+  # cancelled, allowing the connection to terminate
+  max-connection-age-grace = 9223372036854775807
```
zsluedem: https://github.com/grpc/grpc-java/blob/1c50bcafd5c5faa671f3731df24ea12878029f0a/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L67-L107
Just want to share the default value. I don't change the default values

node/src/main/scala/coop/rchain/node/api/package.scala
```diff
@@ -42,6 +49,12 @@ package object api {
           DeployServiceV1GrpcMonix
             .bindService(deployGrpcService, grpcExecutor)
         )
+        .keepAliveTime(keepAliveTime, TimeUnit.MILLISECONDS)
+        .keepAliveTimeout(keepAliveTimeout, TimeUnit.MILLISECONDS)
+        .permitKeepAliveTime(permitKeepAliveTime, TimeUnit.MILLISECONDS)
+        .maxConnectionIdle(maxConnectionIdle, TimeUnit.MILLISECONDS)
+        .maxConnectionAge(maxConnectionAge, TimeUnit.MILLISECONDS)
+        .maxConnectionAgeGrace(maxConnectionAgeGrace, TimeUnit.MILLISECONDS)
```
tgrospic: Default values are specified in nanoseconds and here are milliseconds. A good reason to use duration. :)
https://github.com/grpc/grpc-java/blob/1c50bcafd/netty/src/main/java/io/grpc/netty/NettyServerBuilder.java#L101-L105

node/src/main/resources/defaults.conf
tgrospic: Can we use `FiniteDuration` type like with other time values? It's much less chance to make an error.
```scala
val duration = FiniteDuration(Long.MaxValue, TimeUnit.NANOSECONDS)
// duration: scala.concurrent.duration.FiniteDuration = 9223372036854775807 nanoseconds

val x1 = Long.MaxValue
val x2 = maxDuration.toNanos
val x3 = maxDuration.toMillis
// x1: Long = 9223372036854775807
// x2: Long = 9223372036854775807
// x3: Long = 9223372036854
```

node/src/main/scala/coop/rchain/node/api/package.scala
zsluedem: It makes sense.

tgrospic (approved):
LGTM

zsluedem:
@tgrospic I just want to tell you that  I add one more commit to put the param in the http server also https://github.com/rchain/rchain/pull/2933/commits/836614129c735e01a29e516c001683697c1c1e62 .Hope you don't mind

node/src/main/resources/defaults.conf
```diff
@@ -123,6 +123,27 @@ api-server {
 
   # Enable reporting API
   enable-reporting = false
+
+  # Sets a custom keepalive time, the delay time for sending next keepalive ping
+  keep-alive-time = 2 hours
```
tgrospic, line 128: Is this means that _keep alive_ pings will be sent every 2 hours? How this relates to _keep-alive-timeout_? Should timeout be set to larger value to detect missing pings?

tgrospic: I found the explanation. Timeout is how much server will wait if ping is not received.
> GRPC_ARG_KEEPALIVE_TIMEOUT_MS
This channel argument controls the amount of time (in milliseconds) the sender of the keepalive ping waits for an acknowledgement. If it does not receive an acknowledgment within this time, it will close the connection.

https://github.com/grpc/grpc/blob/master/doc/keepalive.md

tgrospic: Maybe we can set keep alive interval to a smaller value, in a matter of minutes?

nzpr: @zsluedem I agree with @tgrospic Lets have keepalive 5 min with timeout 20sec? Its better to notify client that channel failed, e.g. when waiting for propose accomplish or uploading some big deploy.

nzpr (approved):
I see that default values in conf file are defaults for grpc, lets keep it that way.

PR #2934, opened 2020-05-17
RCHAIN-4077: Key-value store with manager

tgrospic:
## Overview
This PR introduces key-value store abstraction with the manager for easier and general use than existing `Store`.
Manager gives more flexibility how implementation deals with database connections and environments. We can manage all store instances in one place (node project) and relief other places (like RSpace) from knowing about implementation details. After migration, LMDB can be removed as dependency from other projects (shared, block-storage and rspace) and can be moved to node project (and main test project).
This will also enable much easier integration and testing with other key-value databases like RocksDB or Badger.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4077

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
@@ -1,0 +1,131 @@
+package coop.rchain.casper.storage
+
+import java.nio.ByteBuffer
+import java.nio.file.{Files, Path}
+
+import cats.effect.concurrent.{Deferred, Ref}
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import coop.rchain.shared.Log
+import coop.rchain.store.{KeyValueStore, KeyValueStoreManager}
+import enumeratum.{Enum, EnumEntry}
+import org.lmdbjava.{DbiFlags, Env, EnvFlags}
+
+object LmdbStoreManager {
+  def apply[F[_]: Concurrent: Log](dirPath: Path): F[KeyValueStoreManager[F]] =
+    Deferred[F, Env[ByteBuffer]] map (LmdbStoreManagerImpl(dirPath, _))
+}
+
+private final case class LmdbStoreManagerImpl[F[_]: Concurrent: Log](
+    dirPath: Path,
+    envDefer: Deferred[F, Env[ByteBuffer]]
+) extends KeyValueStoreManager[F] {
+
+  sealed trait EnvRefStatus extends EnumEntry
+  object EnvRefStatus extends Enum[EnvRefStatus] {
+    val values = findValues
+    case object EnvClosed   extends EnvRefStatus
+    case object EnvStarting extends EnvRefStatus
+    case object EnvOpen     extends EnvRefStatus
+    case object EnvClosing  extends EnvRefStatus
+  }
+  import EnvRefStatus._
+
+  private case class DbState(
+      status: EnvRefStatus,
+      inProgress: Int,
+      envDefer: Deferred[F, Env[ByteBuffer]],
+      dbs: Map[String, Deferred[F, DbEnv[F]]] = Map.empty
+  ) {
+    override def toString() = s"DbState(status: $status, inProgress: $inProgress)"
+  }
+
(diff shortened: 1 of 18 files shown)


node/src/main/scala/coop/rchain/node/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  // Database name to store instance name mapping (subfolder for LMDB store)
+  // - keys with the same instance will be in one LMDB file (environment)
+  val instanceDBmapping: Map[String, String] = Map[String, String](
+    // RSpace
+    ("history", "v2/history"),
+    ("cold", "v2/cold"),
+    ("roots", "v2/roots"),
+    // Block store
+    ("block-metadata", "dagstorage")
```
tgrospic: `RNodeKeyValueStoreManager` should accept Config and so that this mapping is exactly created from config parameters.

nzpr:
@tgrospic will you add `remove` method to KeyValueStore trait in this PR?

tgrospic:
@nzpr, @zsluedem PR is ready for review. I've added `delete` to KeyValueStore and tests.
Base branch is `dev` so we can use it for trace persistent cache.

casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
```diff
@@ -0,0 +1,138 @@
...
+  // Create LMDB environment and update the state
+  private def createEnv: F[Unit] =
+    for {
+      _     <- Sync[F].delay(Files.createDirectories(dirPath))
+      flags = Seq(EnvFlags.MDB_NOTLS)
+      // Create environment
+      env <- Sync[F].delay(
+              Env
+                .create()
+                .setMapSize(10L * 1024L * 1024L * 1024L)
```
zsluedem, line 116: Makes the param configurable is prefered for me. But we can do that in another pr. Maybe create a ticket for that

casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
```diff
@@ -0,0 +1,138 @@
...
+      // Find DB ref / first time create deferred object.
+      newDbDefer <- Deferred[F, DbEnv[F]]
+      action <- varState.modify { st =>
+                 val maybeDbRef = st.dbs.get(dbName)
+                 maybeDbRef.fold {
+                   val newDbs = st.dbs.updated(dbName, newDbDefer)
+                   val newSt  = st.copy(dbs = newDbs)
+                   (newSt, (true, newDbDefer))
+                 }(defer => (st, (false, defer)))
+               }
```
zsluedem, line 76: So each defer in each db can be a guarantee that the db can be used only if it is created, right?
It looks like the `true` `false` are dups? Not sure if i am right.

zsluedem (commented):
Very good!! Leave some comments!

casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
tgrospic: @zsluedem the idea is for RNodeKeyValueManager to accepts configuration for all stores and instantiate LmdbStoreManager parameters like map size or database count. We can do this as part of PR where it will be used first.

casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala
tgrospic: @zsluedem yes, each caller will be blocked with Deferred until it's completed which is done when DB is created. This is the first step to implement reload of the environment where we need to wait operations on existing environment to finish.

Returning a flag if new environment should be created is a way to capture transition that must happen only once. Only the first request should trigger creation of a new database. But this "effect" cannot be executed inside Ref `modify` function because this function should only depend on the internal state and can be executed multiple times if the state is changed between reading and writing. So the flag `true` will be only signaled once per database for "effectful" code to execute.

nzpr (approved):
Great work! Beautiful!

zsluedem (approved):

PR #2935, opened 2020-05-17
RCHAIN-4092: DAG storage implementation with key-value store (LMDB)

tgrospic:
## Overview
This PR introduces new implementation of DAG storage that uses `KeyValueStore` to persist data.
In separate PR will be the migration logic to transfer data from existing DAG file storage.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4092

### CPU and memory usage with main net state (including #2943)

Start with DAG storage migration (catch up phase).
rnode-start-with-migration

Start with migrated storage (catch up phase).
rnode-start-normal

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -1,0 +1,252 @@
+package coop.rchain.blockstorage.dag
+
+import cats.effect.concurrent.Semaphore
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import com.google.protobuf.ByteString
+import coop.rchain.blockstorage._
+import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
+import coop.rchain.blockstorage.dag.BlockMetadataStore.BlockMetadataStore
+import coop.rchain.blockstorage.dag.EquivocationTrackerStore.EquivocationTrackerStore
+import coop.rchain.blockstorage.dag.codecs._
+import coop.rchain.blockstorage.util.BlockMessageUtil._
+import coop.rchain.casper.protocol.BlockMessage
+import coop.rchain.crypto.codec.Base16
+import coop.rchain.metrics.Metrics.Source
+import coop.rchain.metrics.{Metrics, MetricsSemaphore}
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.models.EquivocationRecord.SequenceNumber
+import coop.rchain.models.Validator.Validator
+import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
+import coop.rchain.shared.syntax._
+import coop.rchain.shared.{Log, LogSource}
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import scodec.Codec
+import scodec.bits.ByteVector
+
+import scala.collection.immutable.SortedMap
+
+final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
+    lock: Semaphore[F],
+    latestMessagesIndex: KeyValueTypedStore[F, Validator, BlockHash],
+    blockMetadataIndex: BlockMetadataStore[F],
+    deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
+    invalidBlocksIndex: KeyValueTypedStore[F, BlockHash, BlockMetadata],
+    equivocationTrackerIndex: EquivocationTrackerStore[F]
+) extends BlockDagStorage[F] {
+  implicit private val logSource: LogSource = LogSource(BlockDagKeyValueStorage.getClass)
+
+  private case class KeyValueDagRepresentation(
+      dagSet: Set[BlockHash],
+      latestMessagesMap: Map[Validator, BlockHash],
+      childMap: Map[BlockHash, Set[BlockHash]],
+      heightMap: SortedMap[Long, Set[BlockHash]],
+      invalidBlocksSet: Set[BlockMetadata]
+  ) extends BlockDagRepresentation[F] {
+
+    def lookup(blockHash: BlockHash): F[Option[BlockMetadata]] =
+      if (dagSet.contains(blockHash)) blockMetadataIndex.get(blockHash)
+      else none[BlockMetadata].pure[F]
+
+    def contains(blockHash: BlockHash): F[Boolean] =
+      (blockHash.size == BlockHash.Length && dagSet.contains(blockHash)).pure[F]
+
+    def children(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
+      childMap.get(blockHash).pure[F]
+
+    def latestMessageHash(validator: Validator): F[Option[BlockHash]] =
+      latestMessagesMap.get(validator).pure[F]
+
+    def latestMessageHashes: F[Map[Validator, BlockHash]] = latestMessagesMap.pure[F]
+
+    def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
+
+    // latestBlockNumber, topoSort and lookupByDeployId are only used in BlockAPI.
+    // Do they need to be part of the DAG current state or they can be moved to DAG storage directly?
+
+    private def getMaxHeight = if (heightMap.nonEmpty) heightMap.last._1 + 1L else 0L
+
+    def latestBlockNumber: F[Long] =
+      getMaxHeight.pure[F]
+
+    def topoSort(
+        startBlockNumber: Long,
+        maybeEndBlockNumber: Option[Long]
+    ): F[Vector[Vector[BlockHash]]] = {
+      val endBlockNumber = maybeEndBlockNumber.getOrElse(getMaxHeight)
+      if (startBlockNumber >= 0 && startBlockNumber <= endBlockNumber) {
+        Sync[F].delay(
+          heightMap
+            .filterKeys(h => h >= startBlockNumber && h <= endBlockNumber)
+            .map { case (_, v) => v.toVector }
+            .toVector
+        )
+      } else {
+        Sync[F].raiseError(
+          TopoSortFragmentParameterError(startBlockNumber, endBlockNumber)
+        )
+      }
+    }
+
+    def lookupByDeployId(deployId: DeployId): F[Option[BlockHash]] =
+      deployIndex.get(deployId)
+  }
+
+  private object KeyValueStoreEquivocationsTracker extends EquivocationsTracker[F] {
+    override def equivocationRecords: F[Set[EquivocationRecord]] =
+      equivocationTrackerIndex.data
+
+    override def insertEquivocationRecord(record: EquivocationRecord): F[Unit] =
+      equivocationTrackerIndex.add(record)
+
+    override def updateEquivocationRecord(
+        record: EquivocationRecord,
+        blockHash: BlockHash
+    ): F[Unit] = {
+      val updatedEquivocationDetectedBlockHashes =
+        record.equivocationDetectedBlockHashes + blockHash
+      val newRecord =
+        record.copy(equivocationDetectedBlockHashes = updatedEquivocationDetectedBlockHashes)
+      equivocationTrackerIndex.add(newRecord)
+    }
+  }
+
+  private def representation: F[BlockDagRepresentation[F]] =
+    for {
+      // Take current DAG state / view of the DAG
+      latestMessages <- latestMessagesIndex.toMap
+      dagSet         <- blockMetadataIndex.dagSet
+      childMap       <- blockMetadataIndex.childMapData
+      heightMap      <- blockMetadataIndex.heightMap
+      invalidBlocks  <- invalidBlocksIndex.toMap.map(_.toSeq.map(_._2).toSet)
+    } yield KeyValueDagRepresentation(
+      dagSet,
+      latestMessages,
+      childMap,
+      heightMap,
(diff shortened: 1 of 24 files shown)


9rb (approved):
LGTM

node/src/main/scala/coop/rchain/node/storage/LmdbManager.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import cats.effect.concurrent.{Deferred, Ref}
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import coop.rchain.shared.Log
+import coop.rchain.store.{KeyValueStore, KeyValueStoreManager}
+import enumeratum.{Enum, EnumEntry}
+import org.lmdbjava.{DbiFlags, Env, EnvFlags}
+
+object LmdbManager {
```
nzpr: Might be rename this to `LmdbStoreManager` to match `InMemoryStoreManager` and `KeyValueStoreManager `?

nzpr:
@tgrospic I can see implementations of `KeyValueStoreManager` like `InMemoryStoreManager` and `LmdbManager`. Also I see `RNodeKeyValueStoreManager` extending KeyValueStoreManager. What is the logical relation between these implementations and `RNodeKeyValueStoreManager`?

node/src/main/scala/coop/rchain/node/storage/LmdbManager.scala
nzpr: Also would be nice to have a comment explaining what's this like in InMemoryStoreManager

node/src/main/scala/coop/rchain/node/storage/LmdbKeyValueStore.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import java.nio.ByteBuffer
+
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.shared.Resources.withResource
+import coop.rchain.store.KeyValueStore
+import org.lmdbjava._
+
+final case class Status(
```
nzpr: Could you add a comment explaining what is Status?

node/src/main/scala/coop/rchain/node/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import java.nio.file.Path
+
+import cats.effect.Concurrent
+import cats.effect.concurrent.{Deferred, Ref}
+import cats.syntax.all._
+import coop.rchain.shared.Log
+import coop.rchain.store.{KeyValueStore, KeyValueStoreManager}
+
+object RNodeKeyValueStoreManager {
```
nzpr: Comment here explaining whats the purpose of `RNodeKeyValueStoreManager ` would be helpful.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -0,0 +1,105 @@
...
+      blockMetadataMap <- blockMetadataStore.toMap
+      dagState         = recreateInMemoryState(blockMetadataMap)
+    } yield new BlockMetadataStore[F](
+      blockMetadataStore,
+      new AtomicMonadState(AtomicAny(dagState))
+    )
+
+  private final case class DagState(
+      dagSet: Set[BlockHash],
+      childMap: Map[BlockHash, Set[BlockHash]],
```
nzpr, line 30: I see this notion of a `child` in many places. But aren't we actually more interested in parents not children?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -0,0 +1,105 @@
...
+//      (oldMetadaatMap, _, _) = result
+//      _ <- blockMetadataStore.put(oldMetadaatMap.toSeq)
+      blockMetadataMap <- blockMetadataStore.toMap
+      dagState         = recreateInMemoryState(blockMetadataMap)
+    } yield new BlockMetadataStore[F](
+      blockMetadataStore,
+      new AtomicMonadState(AtomicAny(dagState))
+    )
+
+  private final case class DagState(
```
nzpr, line 28: This is `BlockMetadataStore` but effectively it defines  the whole DAG, might be rename this object to `DagStore` ?

nzpr: Actually nvm, this is a raw suggestion

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  private def validateDagState(state: DagState): DagState = {
+    // Validate height map index (block numbers) are in sequence without holes
+    val m          = state.heightMap
+    val (min, max) = if (m.nonEmpty) (m.firstKey, m.lastKey + 1) else (0L, 0L)
+    assert(max - min == m.size.toLong, "DAG store height map has numbers not in sequence.")
+    state
+  }
+
+  private def recreateInMemoryState(
```
nzpr, line 91: @tgrospic hdy think, can we maintain DAG state directly in LMDB without in-memory proxy (is performance a concern here)?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -0,0 +1,254 @@
...
+import scodec.Codec
+import scodec.bits.ByteVector
+
+import scala.collection.immutable.SortedMap
+
+final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
+    lock: Semaphore[F],
+    latestMessagesIndex: KeyValueTypedStore[F, Validator, BlockHash],
+    blockMetadataIndex: BlockMetadataStore[F],
+    deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
```
zsluedem, line 32: I wonder what we should do with the deploy in multiple blocks.(When a block get orphaned, the deploy would propose in another block because the block is not finalized).In the pervious implementation, the record should be there but overwritten while reading the log file.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
tgrospic: @nzpr on this PR, I was trying to keep similarity with existing file storage. I agree to make more refactoring of block and dag storage after migration.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
tgrospic: @nzpr this in-memory part gives as ability like a snapshots of the DAG. With LMDB is not that easy, we are always seeing the latest state of the data.
Now we have combination of LMDB to store the whole dag data and in-memory part is just the keys to filter the "heavy" part in LMDB.
So performance is not the concern here and moving more data to LMDB should be way to go. Also with LMDB allocation of direct buffer memory will increase which must be taken into account.

tgrospic:
> I can see implementations of `KeyValueStoreManager` like `InMemoryStoreManager` and `LmdbManager`. Also I see `RNodeKeyValueStoreManager` extending KeyValueStoreManager. What is the logical relation between these implementations and `RNodeKeyValueStoreManager`?

@nzpr _manager_ should wrap/abstract one logical piece of storage. For LMDB this means to hold one environment. For RNode store manager this means the whole "environment" for RNode where multiple LMDB managers are managed.
So it's about how we want to group/manage different storage and split initialization logic from its data API.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
tgrospic: I see it's used in Estimator and LastFinalizedBlockCalculator. Do you think this can be rewritten in different way. Or maybe it's easy calculate childs from parents where is needed.

node/src/main/scala/coop/rchain/node/storage/LmdbKeyValueStore.scala
tgrospic: I left this by mistake. It's not used for now, but the idea was to expose some of storage info through Admin API.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
tgrospic: > In the pervious implementation, the record should be there but overwritten while reading the log file.

@zsluedem when the record is overwritten from the log? Do you mean when node is started?

~I don't think we need separate persistent storage for deploy signatures at all. We can refactor this and hold latest 50 deploy signatures just in memory. When node is starting signatures can be loaded from blocks.~
Correction, `deployIndex` is only used by `BlockAPI.findDeploy` and not in casper validation so it can be exposed directly on the DAG storage and not on DAG representation.

node/src/main/scala/coop/rchain/node/storage/LmdbManager.scala
tgrospic: @nzpr I renamed to `LmdbStoreManager`. I'm not sure what you mean in the second comment. Does my https://github.com/rchain/rchain/pull/2935#issuecomment-636499108 for `KeyValueStoreManager` answers to this?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
nzpr: > I see it's used in Estimator and LastFinalizedBlockCalculator. Do you think this can be rewritten in different way. Or maybe it's easy calculate childs from parents where is needed.

I'm not sure if there is a good way, lets better keep it as it is now.. We have parents in BlockMetadata, dagState is just  additional data, so my question is not relevant, this has been added fo particular usecase.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
zsluedem: When the node load the log files,  the node would load the files in sequences to write to a map in memory. So if a deploy exists in two blocks, the last index would be loaded in the map(The first one would be overwritten(updated) by the last one).

node/src/main/scala/coop/rchain/node/storage/LmdbManager.scala
nzpr: @tgrospic I was just for adding a short comment like `InMemoryStoreManager` has, explaining what is this for. But nvm, I believe the name should be self explaining.

tgrospic: I've added a comment as part of reporting cache PR.
https://github.com/rchain/rchain/blob/391b8f228/casper/src/main/scala/coop/rchain/casper/storage/LmdbStoreManager.scala#L19-L26

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
tgrospic: @zsluedem this means behavior will not change. Last deploy added to deploy index will overwrite previous record.

Maybe we need to save multiple blocks with some deploy to cover all possible cases?

tgrospic:
Results from testing with main net state. Migration is done in about 15-20 sec.
```
17:05:28.184 [main] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Starting DAG file storage migration, loading existing data.
17:05:38.605 [node-runner-22] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Create new DAG storage.
17:05:38.648 [node-runner-22] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Migrate metadata index.
17:05:44.621 [node-runner-21] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Migrate latest messages index.
17:05:44.630 [node-runner-21] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Migrate invalid blocks index.
17:05:44.638 [node-runner-21] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Migrate equivocations tracker index.
17:05:44.642 [node-runner-21] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - Migrate deploys index.
17:05:45.799 [node-runner-21] WARN  c.r.b.dag.BlockDagKeyValueStorage$ - DAG storage migration successful.
```

nzpr (approved):
I've seen these logs on sandbox as well. 

Great work! Waiting for merge!

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
zsluedem: I agree with covering all possible cases. It can be misleaded if the user found the deploy sometimes in one block and sometimes it is in another one.

tgrospic: I opened a new bug/enhancement issue to fix this #2954.

zsluedem (approved):
LGTM

tgrospic:
bors merge

PR #2940, opened 2020-05-21
RCHAIN-4095: Fix slashing condition, not include Slashed deploy in every block [v0.9.23LTS]

zsluedem:
## Overview
Sencond attempt to fix the problem.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4095

### Notes

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -42,6 +42,12 @@ object BlockCreator {
     } yield isActive
   }
 
+  private def isBonded[F[_]: Sync](
+      blockMeta: BlockMetadata,
+      validator: Validator
+  ): Boolean =
+    blockMeta.weightMap.getOrElse(validator, 0L) > 0L // consider stake greater than 0 as bonded
+
   /*
    * Overview of createBlock
    *
@@ -82,10 +88,13 @@ object BlockCreator {
         latestMessageOpt      <- dag.latestMessage(sender)
         seqNum                = latestMessageOpt.fold(0)(_.seqNum) + 1
         _                     <- Log[F].info(s"Creating block with seqNum ${seqNum}")
+        // if the node is already not bonded by the parent, the node won't slash once more
+        invalidLatestMessagesExcludeUnbonded = invalidLatestMessages.filter {
+          case (validator, _) => isBonded(parentMetadatas.head, validator)
+        }
         // TODO: Add `slashingDeploys` to DeployStorage
-        slashingDeploys = invalidLatestMessages.values.toList.map(
+        slashingDeploys = invalidLatestMessagesExcludeUnbonded.values.map(
           invalidBlockHash =>
-            // TODO: Do something useful with the result of "slash".
             SlashDeploy(
               invalidBlockHash,
               validatorIdentity.publicKey,
@@ -110,7 +119,7 @@ object BlockCreator {
         invalidBlocksSet <- dag.invalidBlocks
         invalidBlocks    = invalidBlocksSet.map(block => (block.blockHash, block.sender)).toMap
         // make sure closeBlock is the last system Deploy
-        systemDeploys = slashingDeploys :+ CloseBlockDeploy(
+        systemDeploys = slashingDeploys.toList :+ CloseBlockDeploy(
           SystemDeployUtil.generateCloseDeployRandomSeed(sender, seqNum)
         )
         unsignedBlock <- isActive.ifM(
--- a/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
@@ -65,15 +65,23 @@ object BlockCreator {
     spanF.trace(CreateBlockMetricsSource) {
       import cats.instances.list._
       for {
-        tipHashes             <- Estimator[F].tips(dag, genesis)
-        _                     <- spanF.mark("after-estimator")
-        parentMetadatas       <- EstimatorHelper.chooseNonConflicting(tipHashes, dag)
+        tipHashes <- Estimator[F].tips(dag, genesis)
+        _ <- Log[F].info(
+              s"Creating block with tip hashes ${tipHashes.map(PrettyPrinter.buildString)}"
+            )
+        _               <- spanF.mark("after-estimator")
+        parentMetadatas <- EstimatorHelper.chooseNonConflicting(tipHashes, dag)
+        _ <- Log[F].info(
+              s"Creating block with parents ${parentMetadatas.map(b => PrettyPrinter.buildString(b.blockHash))}"
+            )
         maxBlockNumber        = ProtoUtil.maxBlockNumberMetadata(parentMetadatas)
+        _                     <- Log[F].info(s"Creating block with maxBlockNumber ${maxBlockNumber}")
         invalidLatestMessages <- ProtoUtil.invalidLatestMessages(dag)
         deploys               <- extractDeploys(dag, parentMetadatas, maxBlockNumber, expirationThreshold)
         sender                = ByteString.copyFrom(validatorIdentity.publicKey.bytes)
         latestMessageOpt      <- dag.latestMessage(sender)
         seqNum                = latestMessageOpt.fold(0)(_.seqNum) + 1
+        _                     <- Log[F].info(s"Creating block with seqNum ${seqNum}")
         // TODO: Add `slashingDeploys` to DeployStorage
         slashingDeploys = invalidLatestMessages.values.toList.map(
           invalidBlockHash =>
@@ -96,6 +104,7 @@ object BlockCreator {
         isActive = parents
           .traverse(b => isActiveValidator(b, runtimeManager, validatorIdentity))
           .map(_.forall(identity))
+        _                <- Log[F].info(s"Check isActive:${isActive} for creating block")
         justifications   <- computeJustifications(dag, parents)
         now              <- Time[F].currentMillis
         invalidBlocksSet <- dag.invalidBlocks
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -94,8 +93,35 @@ object BlockCreator {
...
+                                                    .traverse(
+                                                      b =>
+                                                        isActiveValidator(
+                                                          b,
+                                                          runtimeManager,
+                                                          PublicKey(v._1)
+                                                        )
+                                                    )
+                                                    .map(_.forall(identity))
+                                              )
```
tgrospic: Slashed validator is not removed from active validator set until next epoch. We should check bonds list if validator exists on the list and has stake greater that zero. I'm not sure is it necessary to check if it's active but it makes sense to have this check also.

I don't like how `isActiveValidator` is called for each parent and with this new check we will use it in even more cases. Each call is doing Rholang execution and it would better is this function can accept list of keys and return results for all of them in just one Rholang execution. The same should be for calling bonds list.

zsluedem: Ok. You are right. I misunderstand in the first place. I was being stupid

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -42,6 +42,19 @@ object BlockCreator {
...
+      runtimeManager: RuntimeManager[F],
+      publicKey: PublicKey
+  ): F[Boolean] = {
+    val lastProposedTuplespace = ProtoUtil.postStateHash(block)
+    for {
+      bonds     <- runtimeManager.computeBonds(lastProposedTuplespace)
+      validator = ByteString.copyFrom(publicKey.bytes)
+      isBonded  = bonds.exists(b => b.stake > 0 && b.validator == validator)
+    } yield isBonded
+  }
```
tgrospic: If we return a `Map[Validator, Stake]` we need to call it only once even if the block has multiple parents. This is because each call to `computeBonds` is about 300ms.

tgrospic: I take this back, we need bonds for different roots so we need to call this for each. Sorry, my mistake.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -96,6 +107,32 @@ object BlockCreator {
...
+                                                     .traverse(
+                                                       b =>
+                                                         isBonded(
+                                                           b,
+                                                           runtimeManager,
+                                                           PublicKey(v._1)
+                                                         )
+                                                     )
+                                                     .map(_.forall(identity))
+                                               )
```
tgrospic: We have parent blocks here and the bonds list is already in the block (and validated with replay) so after all we don't need to execute Rholang to get it.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -65,26 +78,24 @@ object BlockCreator {
...
-            // TODO: Do something useful with the result of "slash".
-            SlashDeploy(
-              invalidBlockHash,
-              validatorIdentity.publicKey,
-              SystemDeployUtil.generateSlashDeployRandomSeed(sender, seqNum)
-            )
-        )
-        parents <- parentMetadatas.toList.traverse(p => ProtoUtil.getBlock(p.blockHash))
+        _                     <- Log[F].info(s"Creating block with seqNum ${seqNum}")
+        parents               <- parentMetadatas.toList.traverse(p => ProtoUtil.getBlock(p.blockHash))
```
tgrospic: I didn't look in detail but it might be that we don't need the full block here and just use block metadata instead. :thinking:

tgrospic: Ok, we will need at least bonds list which is not stored in bock metadata. Something for later refactoring with the DAG storage indexes.

zsluedem: `BlockMetaData` doesn't contain `postStateHash` which `isActive` wants

zsluedem: Maybe we should reconsider `blockMetaData` when we do the migrations.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -65,17 +70,31 @@ object BlockCreator {
...
+          v =>
+            parentMetadatas
+              .map(p => isBonded(p, v._1))
+              .forall(identity)
+        )
         // TODO: Add `slashingDeploys` to DeployStorage
-        slashingDeploys = invalidLatestMessages.values.toList.map(
+        slashingDeploys = invalidLatestMessagesExcludeUnbonded.values.toList.map(
           invalidBlockHash =>
             // TODO: Do something useful with the result of "slash".
```
tgrospic: This comment is obsolete. We are handling slashing funds in PoS contract.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -42,6 +42,11 @@ object BlockCreator {
     } yield isActive
   }
 
+  private def isBonded[F[_]: Sync](
+      blockMeta: BlockMetadata,
+      validator: Validator
+  ): Boolean = blockMeta.weightMap.exists(s => s._1 == validator)
```
tgrospic: Oh, we have bonds in metadata after all. :)

tgrospic (commented):
It looks OK. Please add tests for this new case. It's strange that with this change all tests are stil passing.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -65,19 +71,32 @@ object BlockCreator {
...
         deploys               <- extractDeploys(dag, parentMetadatas, maxBlockNumber, expirationThreshold)
         sender                = ByteString.copyFrom(validatorIdentity.publicKey.bytes)
         latestMessageOpt      <- dag.latestMessage(sender)
         seqNum                = latestMessageOpt.fold(0)(_.seqNum) + 1
+        _                     <- Log[F].info(s"Creating block with seqNum ${seqNum}")
+        invalidLatestMessagesExcludeUnbonded = invalidLatestMessages.filter(
+          v =>
+            parentMetadatas
+              .map(p => isBonded(p, v._1))
+              .forall(identity)
```
tgrospic: As Greg said, we need to filter parents and take into account only ones that makes 67% of the whole stake.

zsluedem: It can end up with several cases. How to deal with the cases propoerly?

tgrospic: Instead of just returning boolean, `isBonded` can also return the stake. So here we can have overall stake behind each result. To decide if validator is bonded we don't need all of the parents (stake) to say yes, we only need 67% of the whole stake to say yes.
For synchrony constraint there is similar checking. Although it has a bug if only one validator is bonded.
https://github.com/rchain/rchain/blob/037411fc62cbb0115d6cb93fa2b7df62a320b649/casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala#L46-L58

zsluedem: > we only need 67% of the whole stake to say yes.

Which `bondsMap` do we choose here?

zsluedem: What I mean is we are leading into a loop on choosing the right bondsMap here

zsluedem:
Fall back to  version like`7c4bdb0` but brand new commit

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -2,7 +2,7 @@ package coop.rchain.casper
 
 import cats.Monad
 import cats.effect.Sync
-import cats.syntax.all._
+import cats.implicits._
```
tgrospic: Please use _syntax_  to import because _implicits_ includes much more things and type inference can have more ambiguities which requires more explicitly written types.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -65,19 +71,33 @@ object BlockCreator {
...
         sender                = ByteString.copyFrom(validatorIdentity.publicKey.bytes)
         latestMessageOpt      <- dag.latestMessage(sender)
         seqNum                = latestMessageOpt.fold(0)(_.seqNum) + 1
+        _                     <- Log[F].info(s"Creating block with seqNum ${seqNum}")
+        // if the node is already not bonded by the parent, the node won't slash once more
+        invalidLatestMessagesExcludeUnbonded = invalidLatestMessages.filter(
+          v =>
+            parentMetadatas
+              .map(p => isBonded(p, v._1))
+              .forall(identity)
```
tgrospic: Only the first (base) parent needs to be checked, this is the same as getting bonds list from base parent block.

casper/src/main/scala/coop/rchain/casper/BlockCreator.scala
```diff
@@ -213,7 +234,7 @@ object BlockCreator {
   )(implicit spanF: Span[F]): F[CreateBlockStatus] =
     spanF.trace(ProcessDeploysAndCreateBlockMetricsSource) {
       (for {
-        blockData <- BlockData(now, maxBlockNumber + 1, sender, seqNum).pure
+        blockData <- BlockData(now, maxBlockNumber + 1, sender, seqNum).pure[F]
```
tgrospic: With _cats.syntax_ import, F here can be inferred.

tgrospic (approved):
Although we are moving to v0.9.25 version, it makes more sense to merge this PR then close it.

nzpr (approved):
LGTM

PR #2943, opened 2020-06-01
RCHAIN-4080: New block retrieval procedure + preparing for LFS

### LINKED ISSUE #2980: New block retrieval procedure + preparing for LFS
We need 
1. Redesign catch up logic so it allows catch up on long chains without consuming lots of memory
- Introduce BlockRetriever trait that should hold all logic related to serving Casper with blocks it needs.
- Remove related logic from CommUtils
- Make Casper buffer persistent to not lose catch up state on restart.

2. Make basic block validation that we can do without Casper, in message handlers before putting block in BlockStore. 
- Filter invalid blocks
- Filter blocks that are not needed when Last Finalised State is implemented

3. Improve filtering of duplicate messages.

4. Make Casper lock less aggressive (move thing that are not required to be inside lock, outside).

5. Make Casper|BlockRetriever work well with restarts.

nzpr:
## Overview
Continuation of https://github.com/rchain/rchain/pull/2926

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4080

### Notes

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
@@ -1,0 +1,117 @@
+package coop.rchain.blockstorage.casperbuffer
+
+import cats.effect.Concurrent
+import cats.effect.concurrent.{Ref, Semaphore}
+import cats.syntax.all._
+import coop.rchain.catscontrib.Catscontrib._
+import coop.rchain.blockstorage.dag.codecs._
+import coop.rchain.shared.syntax._
+import coop.rchain.blockstorage.util.{
+  BlockDependencyDag,
+  DoublyLinkedDag,
+  DoublyLinkedDagOperations
+}
+import coop.rchain.blockstorage.BlockStorageMetricsSource
+import coop.rchain.metrics.Metrics.Source
+import coop.rchain.metrics.{Metrics, MetricsSemaphore}
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.shared.Log
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import scodec.Codec
+
+/**
+  * @param parentsStore - persistent map {hash -> parents set}
+  * @param blockDependencyDag - in-memory dependency DAG, recreated from parentsStore on node startup
+  */
+final class CasperBufferKeyValueStorage[F[_]: Concurrent: Log] private (
+    parentsStore: KeyValueTypedStore[F, BlockHash, Set[BlockHash]],
+    blockDependencyDag: Ref[F, DoublyLinkedDag[BlockHash]],
+    lock: Semaphore[F]
+) extends CasperBufferStorage[F] {
+
+  override def addRelation(parent: BlockHash, child: BlockHash): F[Unit] =
+    lock.withPermit(
+      for {
+        parents <- parentsStore.get(child).map(_.getOrElse(Set.empty[BlockHash]))
+        _       <- parentsStore.put(child, parents + parent)
+        _ <- blockDependencyDag.update(
+              curState => DoublyLinkedDagOperations.add(curState, parent, child)
+            )
+      } yield ()
+    )
+
+  override def remove(hash: BlockHash): F[Unit] =
+    lock.withPermit(
+      for {
+        curState                                  <- blockDependencyDag.get
+        (newState, hashesAffected, hashesRemoved) = DoublyLinkedDagOperations.remove(curState, hash)
+        _                                         <- blockDependencyDag.set(newState)
+        changes                                   = hashesAffected.map(h => (h, newState.childToParentAdjacencyList(h))).toSeq
+        deletes                                   = hashesRemoved.toSeq
+        _                                         <- parentsStore.put(changes)
+        // If node crush here - KV store end up in inconsistent state. What's the good way to handle it?
+        _ <- parentsStore.delete(deletes)
+      } yield ()
+    )
+
+  override def getParents(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
+    blockDependencyDag.get.map(_.childToParentAdjacencyList.get(blockHash))
+
+  override def getChildren(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
+    blockDependencyDag.get.map(_.parentToChildAdjacencyList.get(blockHash))
+
+  override def getPendants: F[Set[BlockHash]] =
+    blockDependencyDag.get.map(_.dependencyFree)
+
+  // Block is considered to be in CasperBuffer when there is a records about its parents
+  override def contains(blockHash: BlockHash): F[Boolean] =
+    blockDependencyDag.get.map(
+      bdd => bdd.childToParentAdjacencyList.get(blockHash).exists(_.nonEmpty)
+    )
+
+  override def toDoublyLinkedDag: F[DoublyLinkedDag[BlockHash]] = blockDependencyDag.get
+
+  override def size: F[Long] =
+    blockDependencyDag.get.map(_.childToParentAdjacencyList.size.toLong)
+
+  override def isPendant(blockHash: BlockHash): F[Boolean] =
+    getPendants.map(_.contains(blockHash))
+}
+
+object CasperBufferKeyValueStorage {
+  implicit private val ms: Source =
+    Metrics.Source(BlockStorageMetricsSource, "dag-key-value-store")
+
+  def create[F[_]: Concurrent: KeyValueStoreManager: Log: Metrics]: F[CasperBufferStorage[F]] = {
+
+    def recreateInMemStore(
+        parentsStore: KeyValueTypedStore[F, BlockHash, Set[BlockHash]],
+        initState: DoublyLinkedDag[BlockHash] = BlockDependencyDag.empty
+    ): F[DoublyLinkedDag[BlockHash]] =
+      for {
+        parentsMap <- parentsStore.toMap
+        inMemStore = parentsMap.keySet.toList.foldLeft(initState) { (bdd, key) =>
+          val parents = parentsMap(key)
+          parents.toList.foldLeft(bdd)(
+            (bdd, p) => DoublyLinkedDagOperations.add(bdd, p, key)
+          )
+        }
+      } yield inMemStore
+
+    for {
+      parentsStore <- KeyValueStoreManager[F].database[BlockHash, Set[BlockHash]](
+                       "parents-map",
+                       codecBlockHash,
+                       codecBlockHashSet
+                     )
+      inMemStore    <- recreateInMemStore(parentsStore)
+      inMemStoreRef <- Ref.of[F, DoublyLinkedDag[BlockHash]](inMemStore)
+      lock          <- MetricsSemaphore.single[F]
+    } yield new CasperBufferKeyValueStorage[F](
+      parentsStore,
+      inMemStoreRef,
+      lock
+    )
+  }
+
+}
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferStorage.scala
@@ -1,0 +1,25 @@
+package coop.rchain.blockstorage.casperbuffer
+
+import coop.rchain.blockstorage.util.{BlockDependencyDag, DoublyLinkedDag}
+import coop.rchain.models.BlockHash.BlockHash
+
+/**
+  * CasperBuffer manages block dependency graph.
+  * Block is in scope of Casper Buffer until it is validated and added to DAG.
+  */
+trait CasperBufferStorage[F[_]] {
+  def getParents(blockHash: BlockHash): F[Option[Set[BlockHash]]]
+  def getChildren(blockHash: BlockHash): F[Option[Set[BlockHash]]]
+  def getPendants: F[Set[BlockHash]]
+  def addRelation(parent: BlockHash, child: BlockHash): F[Unit]
+  def remove(blockHash: BlockHash): F[Unit]
+  def contains(blockHash: BlockHash): F[Boolean]
+  def isPendant(blockHash: BlockHash): F[Boolean]
+  def size: F[Long]
+  // TODO this is used only in single place, refactor and remove this method
+  def toDoublyLinkedDag: F[DoublyLinkedDag[BlockHash]]
+}
+
+object CasperBufferStorage {
+  def apply[F[_]](implicit instance: CasperBufferStorage[F]): CasperBufferStorage[F] = instance
+}
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
@@ -1,4 +1,4 @@
-package coop.rchain.casper.util
+package coop.rchain.blockstorage.util
 
 import coop.rchain.models.BlockHash.BlockHash
 
@@ -64,7 +64,7 @@ object DoublyLinkedDagOperations {
     val parentToChildAdjacencyList: Map[A, Set[A]] = dag.parentToChildAdjacencyList
     val childToParentAdjacencyList: Map[A, Set[A]] = dag.childToParentAdjacencyList
     assert(!childToParentAdjacencyList.contains(element))
-    assert(!parentToChildAdjacencyList.values.toSet.contains(element))
+    assert(!parentToChildAdjacencyList.valuesIterator.flatten.contains(element))
     val maybeChildren = parentToChildAdjacencyList.get(element)
     val initAcc       = (childToParentAdjacencyList, Set.empty[A])
     val (updatedChildToParentAdjacencyList, newDependencyFree) = maybeChildren match {
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
@@ -115,6 +115,6 @@ object DoublyLinkedDagOperations {
           : Map[A, Set[A]]                                    = dag.parentToChildAdjacencyList - element
       override val childToParentAdjacencyList: Map[A, Set[A]] = updatedChildToParentAdjacencyList
       override val dependencyFree: Set[A]                     = dag.dependencyFree ++ newDependencyFree - element
-    }, childrenAffected.toSet, childrenRemoved.toSet)
+    }, childrenAffected, childrenRemoved)
   }
 }
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
@@ -3,12 +3,14 @@ package coop.rchain.blockstorage.util
 import coop.rchain.models.BlockHash.BlockHash
 
 import scala.collection.immutable.{HashMap, HashSet}
+import scala.collection.mutable
 
 trait DoublyLinkedDag[A] {
   val parentToChildAdjacencyList: Map[A, Set[A]]
   val childToParentAdjacencyList: Map[A, Set[A]]
   val dependencyFree: Set[A]
 }
+
 final case class BlockDependencyDag(
     parentToChildAdjacencyList: Map[BlockHash, Set[BlockHash]],
     childToParentAdjacencyList: Map[BlockHash, Set[BlockHash]],
@@ -60,37 +62,59 @@ object DoublyLinkedDagOperations {
 
   // If the element doesn't exist in the dag, the dag is returned as is
   @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
-  def remove[A](dag: DoublyLinkedDag[A], element: A): DoublyLinkedDag[A] = {
+  def remove[A](
+      dag: DoublyLinkedDag[A],
+      element: A
+  ): (DoublyLinkedDag[A], Set[A], Set[A]) = {
     val parentToChildAdjacencyList: Map[A, Set[A]] = dag.parentToChildAdjacencyList
     val childToParentAdjacencyList: Map[A, Set[A]] = dag.childToParentAdjacencyList
     assert(!childToParentAdjacencyList.contains(element))
     assert(!parentToChildAdjacencyList.valuesIterator.flatten.contains(element))
     val maybeChildren = parentToChildAdjacencyList.get(element)
-    val initAcc       = (childToParentAdjacencyList, Set.empty[A])
-    val (updatedChildToParentAdjacencyList, newDependencyFree) = maybeChildren match {
-      case Some(children) =>
-        children.foldLeft(initAcc) {
-          case ((childToParentAdjacencyListAcc, dependencyFreeAcc), child) =>
-            val maybeParents = childToParentAdjacencyListAcc.get(child)
-            maybeParents match {
-              case Some(parents) =>
-                val updatedParents = parents - element
-                if (updatedParents.isEmpty) {
-                  (childToParentAdjacencyListAcc - child, dependencyFreeAcc + child)
-                } else {
-                  (childToParentAdjacencyListAcc.updated(child, updatedParents), dependencyFreeAcc)
-                }
-              case None =>
-                throw new Error(s"We should have at least $element as parent")
-            }
-        }
-      case None => initAcc
-    }
-    new DoublyLinkedDag[A] {
+    val initAcc       = (childToParentAdjacencyList, Set.empty[A], Set.empty[A], Set.empty[A])
+    val (updatedChildToParentAdjacencyList, newDependencyFree, childrenAffected, childrenRemoved) =
+      maybeChildren match {
+        case Some(children) =>
+          children.foldLeft(initAcc) {
+            case (
+                (
+                  childToParentAdjacencyListAcc,
+                  dependencyFreeAcc,
+                  childrenAffected,
+                  childrenRemoved
+                ),
+                child
+                ) =>
+              val maybeParents = childToParentAdjacencyListAcc.get(child)
+              maybeParents match {
+                case Some(parents) =>
+                  val updatedParents = parents - element
+                  if (updatedParents.isEmpty)
+                    (
+                      childToParentAdjacencyListAcc - child,
+                      dependencyFreeAcc + child,
+                      childrenAffected,
+                      childrenRemoved + child
+                    )
+                  else
+                    (
+                      childToParentAdjacencyListAcc.updated(child, updatedParents),
+                      dependencyFreeAcc,
+                      childrenAffected + child,
+                      childrenRemoved
+                    )
+
+                case None =>
+                  throw new Error(s"We should have at least $element as parent")
+              }
+          }
+        case None => initAcc
+      }
+    (new DoublyLinkedDag[A] {
       override val parentToChildAdjacencyList
           : Map[A, Set[A]]                                    = dag.parentToChildAdjacencyList - element
       override val childToParentAdjacencyList: Map[A, Set[A]] = updatedChildToParentAdjacencyList
       override val dependencyFree: Set[A]                     = dag.dependencyFree ++ newDependencyFree - element
-    }
+    }, childrenAffected.toSet, childrenRemoved.toSet)
   }
 }
--- a/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
@@ -3,7 +3,7 @@ package coop.rchain.casper
 sealed trait BlockStatus
 object BlockStatus {
   def valid: ValidBlock                    = ValidBlock.Valid
-  def processing: BlockError               = BlockError.Processing
+  def processed: BlockError                = BlockError.Processed
   def casperIsBusy: BlockError             = BlockError.CasperIsBusy
   def exception(ex: Throwable): BlockError = BlockError.BlockException(ex)
   def admissibleEquivocation: BlockError   = InvalidBlock.AdmissibleEquivocation
@@ -45,7 +45,7 @@ object ValidBlock {
 
 sealed trait BlockError extends BlockStatus
 object BlockError {
-  case object Processing                         extends BlockError
+  case object Processed                          extends BlockError
   case object CasperIsBusy                       extends BlockError
   final case class BlockException(ex: Throwable) extends BlockError
 }
--- a/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
@@ -4,6 +4,7 @@ sealed trait BlockStatus
 object BlockStatus {
   def valid: ValidBlock                    = ValidBlock.Valid
   def processing: BlockError               = BlockError.Processing
+  def casperIsBusy: BlockError             = BlockError.CasperIsBusy
   def exception(ex: Throwable): BlockError = BlockError.BlockException(ex)
   def admissibleEquivocation: BlockError   = InvalidBlock.AdmissibleEquivocation
   def ignorableEquivocation: BlockError    = InvalidBlock.IgnorableEquivocation
@@ -45,6 +46,7 @@ object ValidBlock {
 sealed trait BlockError extends BlockStatus
 object BlockError {
   case object Processing                         extends BlockError
+  case object CasperIsBusy                       extends BlockError
   final case class BlockException(ex: Throwable) extends BlockError
 }
 
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -123,9 +123,7 @@ sealed abstract class MultiParentCasperInstances {
                                  } yield postGenesisStateHash
                                }
         blockProcessingLock <- MetricsSemaphore.single[F]
-        casperState         <- Cell.mvarCell[F, CasperState](CasperState())
       } yield {
-        implicit val state = casperState
         new MultiParentCasperImpl(
           validatorId,
           genesis,
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -3,13 +3,13 @@ package coop.rchain.casper
 import cats.effect.{Concurrent, Sync}
 import cats.syntax.all._
 import cats.{Applicative, Show}
-import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.BlockStore
+import coop.rchain.blockstorage.casperbuffer.CasperBufferStorage
 import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
 import coop.rchain.blockstorage.dag.{BlockDagRepresentation, BlockDagStorage}
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedStorage
-import coop.rchain.casper.engine.Running
+import coop.rchain.casper.engine.{BlockRetriever, Running}
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.syntax._
 import coop.rchain.casper.util.ProtoUtil
@@ -47,13 +47,26 @@ object DeployError {
 }
 
 trait Casper[F[_]] {
-  def addBlock(b: BlockMessage): F[ValidBlockProcessing]
+  def addBlockFromStore(b: BlockHash, allowAddFromBuffer: Boolean = false): F[ValidBlockProcessing]
+  def addBlock(b: BlockMessage, allowAddFromBuffer: Boolean = false): F[ValidBlockProcessing]
   def contains(hash: BlockHash): F[Boolean]
+  def dagContains(hash: BlockHash): F[Boolean]
+  def bufferContains(hash: BlockHash): F[Boolean]
   def deploy(d: Signed[DeployData]): F[Either[DeployError, DeployId]]
   def estimator(dag: BlockDagRepresentation[F]): F[IndexedSeq[BlockHash]]
   def createBlock: F[CreateBlockStatus]
   def getGenesis: F[BlockMessage]
   def getValidator: F[Option[PublicKey]]
+  def getVersion: F[Long]
+
+  /**
+    * Approved Block extends notion of Genesis Block - its the block that meets Last Finalized State.
+    * To be able to start building on top of the Approved Block, node have to have some additional information
+    * from the network.
+    *
+    * @return if node has all necessary information accompanying ApprovedBlock
+    */
+  def approvedBlockStateComplete: F[Boolean]
 }
 
 trait MultiParentCasper[F[_]] extends Casper[F] {
@@ -86,7 +99,7 @@ sealed abstract class MultiParentCasperInstances {
     Metrics.Source(CasperMetricsSource, "casper")
   private[this] val genesisLabel = Metrics.Source(MetricsSource, "genesis")
 
-  def hashSetCasper[F[_]: Sync: Metrics: Concurrent: CommUtil: Log: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: BlockDagStorage: LastFinalizedStorage: Span: Running.RequestedBlocks: EventPublisher: SynchronyConstraintChecker: LastFinalizedHeightConstraintChecker: Estimator: DeployStorage](
+  def hashSetCasper[F[_]: Sync: Metrics: Concurrent: CommUtil: Log: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: BlockDagStorage: LastFinalizedStorage: Span: EventPublisher: SynchronyConstraintChecker: LastFinalizedHeightConstraintChecker: Estimator: DeployStorage: CasperBufferStorage: BlockRetriever](
       validatorId: Option[ValidatorIdentity],
       genesis: BlockMessage,
       shardId: String,
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -85,13 +85,12 @@ object MultiParentCasper extends MultiParentCasperInstances {
   def ignoreDoppelgangerCheck[F[_]: Applicative]: (BlockMessage, Validator) => F[Unit] =
     kp2(().pure)
 
-  def forkChoiceTip[F[_]: Sync: BlockStore](casper: MultiParentCasper[F]): F[BlockMessage] =
+  def forkChoiceTip[F[_]: Sync](casper: MultiParentCasper[F]): F[BlockHash] =
     for {
       dag       <- casper.blockDag
       tipHashes <- casper.estimator(dag)
       tipHash   = tipHashes.head
-      tip       <- BlockStore[F].getUnsafe(tipHash)
-    } yield tip
+    } yield tipHash
 }
 
 sealed abstract class MultiParentCasperInstances {
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -58,6 +58,7 @@ trait Casper[F[_]] {
   def getGenesis: F[BlockMessage]
   def getValidator: F[Option[PublicKey]]
   def getVersion: F[Long]
+  def getDeployLifespan: F[Int]
 
   /**
     * Approved Block extends notion of Genesis Block - its the block that meets Last Finalized State.
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -507,7 +507,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
       )
       _ <- (missingDeps ++ depsInBuffer).traverse(casperBuffer.addRelation(_, b.blockHash))
       _ <- Log[F].info(
-            s"${PrettyPrinter.buildString(b.blockHash)} missing dependencies. " +
+            s"Block ${PrettyPrinter.buildString(b, short = true)} missing dependencies. " +
               s"Fetching: ${PrettyPrinter.buildString(missingDeps)}. " +
               s"Already in CasperBuffer: ${PrettyPrinter.buildString(depsInBuffer)}. " +
               s"Already in DAG: ${PrettyPrinter.buildString(depsInDag)}."
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -25,6 +25,11 @@ import coop.rchain.shared._
 import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedStorage
+import coop.rchain.blockstorage.util.{
+  BlockDependencyDag,
+  DoublyLinkedDag,
+  DoublyLinkedDagOperations
+}
 import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.Signed
 
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -7,7 +7,6 @@ import cats.syntax.all._
 import coop.rchain.blockstorage._
 import coop.rchain.blockstorage.dag.{BlockDagRepresentation, BlockDagStorage}
 import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
-import coop.rchain.casper.CasperState.CasperStateCell
 import coop.rchain.casper.protocol._
 import coop.rchain.casper.syntax._
 import coop.rchain.casper.util._
@@ -23,32 +22,13 @@ import coop.rchain.models.{EquivocationRecord, NormalizerEnv}
 import coop.rchain.models.Validator.Validator
 import coop.rchain.shared._
 import com.google.protobuf.ByteString
+import coop.rchain.blockstorage.casperbuffer.CasperBufferStorage
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedStorage
-import coop.rchain.blockstorage.util.{
-  BlockDependencyDag,
-  DoublyLinkedDag,
-  DoublyLinkedDagOperations
-}
 import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.Signed
 
-/**
-  Encapsulates mutable state of the MultiParentCasperImpl
-
-  @param blockBuffer - holds hashes of blocks that were received but were not added because DAG does not have their parents yet
-  @param dependencyDag - dependency dag for block buffer // TODO they should be one structure
-  */
-final case class CasperState(
-    blockBuffer: Set[BlockHash] = Set.empty[BlockHash],
-    dependencyDag: DoublyLinkedDag[BlockHash] = BlockDependencyDag.empty
-)
-
-object CasperState {
-  type CasperStateCell[F[_]] = Cell[F, CasperState]
-}
-
-class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: BlockDagStorage: LastFinalizedStorage: CommUtil: EventPublisher: Estimator: DeployStorage](
+class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: BlockDagStorage: LastFinalizedStorage: CommUtil: EventPublisher: Estimator: DeployStorage: BlockRetriever](
     validatorId: Option[ValidatorIdentity],
     genesis: BlockMessage,
     postGenesisStateHash: StateHash,
@@ -56,7 +36,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
     finalizationRate: Int,
     blockProcessingLock: Semaphore[F]
 )(
-    implicit state: CasperStateCell[F],
+    implicit casperBuffer: CasperBufferStorage[F],
     metricsF: Metrics[F],
     spanF: Span[F],
     runtimeManager: RuntimeManager[F]
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -3,6 +3,7 @@ package coop.rchain.casper
 import cats.data.EitherT
 import cats.effect.{Concurrent, Sync}
 import cats.effect.concurrent.{Ref, Semaphore}
+import cats.effect.syntax.bracket
 import cats.syntax.all._
 import coop.rchain.blockstorage._
 import coop.rchain.blockstorage.dag.{BlockDagRepresentation, BlockDagStorage}
@@ -15,6 +16,7 @@ import coop.rchain.casper.util.comm.CommUtil
 import coop.rchain.casper.util.rholang._
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
 import coop.rchain.catscontrib.BooleanF._
+import coop.rchain.catscontrib.Catscontrib.ToBooleanF
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.metrics.implicits._
 import coop.rchain.models.BlockHash._
@@ -25,12 +27,13 @@ import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.casperbuffer.CasperBufferStorage
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedStorage
+import coop.rchain.casper.engine.BlockRetriever
 import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.Signed
 
 class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: LastFinalizedBlockCalculator: BlockStore: BlockDagStorage: LastFinalizedStorage: CommUtil: EventPublisher: Estimator: DeployStorage: BlockRetriever](
     validatorId: Option[ValidatorIdentity],
-    genesis: BlockMessage,
+    approvedBlock: BlockMessage,
     postGenesisStateHash: StateHash,
     shardId: String,
     finalizationRate: Int,
@@ -45,7 +48,6 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
 
   implicit private val logSource: LogSource = LogSource(this.getClass)
   private[this] val syncF                   = Sync[F]
-  private[this] val noop                    = syncF.unit
 
   //TODO: Extract hardcoded version and expirationThreshold
   private val version             = 1L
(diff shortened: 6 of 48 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,148 @@
...
+    justificationsStore
+      .get(Seq(blockHash))
+      .map(_.head.exists(_.nonEmpty))
+
+  override def toDoublyLinkedDag: F[BlockDependencyDag] =
+    for {
+      childrenMap   <- justifiedByStore.toMap
+      parentsMap    <- justificationsStore.toMap
+      pendentBlocks <- pendantBlocksStore.toMap.flatMap(_.keySet.pure[F])
+    } yield BlockDependencyDag(childrenMap, parentsMap, pendentBlocks)
```
tgrospic: `toMap` on KeyValueStore should be called only if the whole database should be loaded and there is no other way to read only part of the data. Using _get_ and _put_ directly on KeyValueStore should be comparable in performance to in-memory structure.

`getPendants` and `toDoublyLinkedDag` are called for each attempt to add a block. This will copy both databases to heap and discard quickly for GC to collect. We could store only one copy and just share the reference.
So just storing the block metadata (parents) on disk, we can load the whole database only on startup and reconstruct in-memory DoublyLinkedDag. Which can be shared and updated together with metadata as the source.

Another aspect of this from consistency perspective. All three databases have the same meaning for the key as block hash. If data is updated in one database and not the other it can contain incomplete data from the whole view of all DBs. It makes sense to have one object to hold all values needed for particular block which can be stored in one DB and updated atomically.
```scala
case class BlockBufferState(children: Seq[Hash], parents: Seq[Hash], pendant: Boolean)
```

casper/src/main/scala/coop/rchain/casper/engine/BlocksRetriever.scala
```diff
@@ -0,0 +1,235 @@
...
+                  val _ = Log[F].debug(
+                    s"Adding ${PrettyPrinter.buildString(hash)} hash to RequestedBlocks. Already exists."
+                  )
+                  requestedBlocks
+                } else
+                  requestedBlocks +
+                    (hash -> RequestState(
+                      timestamp = now,
+                      peers = if (peerRequested.isEmpty) None.toSet else Set(peerRequested.get),
+                      received = markAsReceived
```
tgrospic: Please convert this to `for .. yield` expression. Logging in line 106 and 108 is "in the air" because expression returned by Log[F].info is not used in execution sequence like with `_ <- Log[F].info(...)`.

casper/src/main/scala/coop/rchain/casper/engine/BlocksRetriever.scala
```diff
@@ -0,0 +1,235 @@
...
+              reason
+            )
+        } >> RequestedBlocks
+          .get(hash)
+          .flatMap(
+            state =>
+              if (state.get.peers.size == 1)
+                CommUtil[F].requestForBlock(peer, hash) >>
+                  CommUtil[F].broadcastHasBlockRequest(hash)
+              else ().pure[F]
```
tgrospic: `for .. yield` here would be a) much more readable, b) requested blocks _get_ can be called only once and use the same reference in both places and c) it would be much easier to add some additional code later if needed.
This is not always the case but when more variables are involved in larger scope `for .. yield` is usually much nicer.

casper/src/main/scala/coop/rchain/casper/engine/BlocksRetriever.scala
```diff
@@ -0,0 +1,235 @@
...
+                      s"Trying ${nextPeer.endpoint.host} to query for ${PrettyPrinter.buildString(hash)} block. " +
+                        s"Remain waiting: ${waitingListTail.map(_.endpoint.host).mkString(", ")}."
+                    )
+                _  <- CommUtil[F].requestForBlock(nextPeer, hash)
+                ts <- Time[F].currentMillis
+              } yield hash -> Option(modifiedRequested(ts))
+            case _ =>
+              val warnMessage = s"Could not retrieve requested block ${PrettyPrinter.buildString(hash)} " +
+                s"from ${requested.peers.mkString(", ")}. Removing the request from the requested blocks list. " +
+                s"Casper will have to re-request the block."
```
tgrospic: What is the reason to force casper to re-request the block? What is gained with pushing this again to casper?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -98,140 +73,110 @@ object Running {
...
+        ) >> BlocksRetriever[F].admitHash(h, peer, s"Unknown hash broadcast received.")
+      }
+    })
   }
 
-  def handleHasBlockRequest[F[_]: Monad: RPConfAsk: TransportLayer](
+  /**
+    * Peer says it has particular block.
+    */
+  def handleHasBlockMessage[F[_]: Monad: RPConfAsk: BlocksRetriever: TransportLayer: Time: Log: CommUtil](
```
tgrospic: Only `BlocksRetriever` is used in this function.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -98,140 +73,110 @@ object Running {
...
-          s"Received block hash ${PrettyPrinter.buildString(blockHashMessage.blockHash)}. Requesting ..."
-        ) >> requestForNewBlock(peer, blockHashMessage.blockHash)
-      )
+    BlocksRetriever[F].admitHash(hb.hash, peer, s"HasBlock confirmation received.")
 
-  def handleBlockRequest[F[_]: Monad: RPConfAsk: BlockStore: Log: TransportLayer](
+  /**
+    * Peer asks for particular block
+    */
+  def handleBlockRequest[F[_]: Monad: RPConfAsk: BlockStore: Log: TransportLayer: CommUtil](
```
tgrospic: `RPConfAsk` and `TransportLayer` are unused. Please check other handlers also.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -241,37 +186,58 @@ object Running {
...
+    * Peer asks if this node has particular block
+    */
+  def handleHasBlockRequest[F[_]: Monad: RPConfAsk: TransportLayer: CommUtil](
+      peer: PeerNode,
+      hbr: HasBlockRequest
+  )(blockLookup: BlockHash => F[Boolean]): F[Unit] =
+    blockLookup(hbr.hash).flatMap(
+      present =>
+        Applicative[F].whenA(present)(
+          CommUtil[F].sendToPeer(peer, HasBlockProto(hbr.hash))
```
tgrospic: FYI `whenA` has also extension variant e.g.
```scala
CommUtil[F].sendToPeer(peer, HasBlockProto(hbr.hash)).whenA(present)
```
It looks like _if_ in Python `expr if true`. :)

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -241,37 +186,58 @@ object Running {
...
+        )
+    )
+
+  /**
+    * Peer asks for fork-choice tip
+    */
+  def handleForkChoiceTipRequest[F[_]: Sync: RPConfAsk: Log: TransportLayer: BlockStore: CommUtil](
       peer: PeerNode,
       fctr: ForkChoiceTipRequest.type
```
tgrospic: This parameter is not used.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -241,37 +186,58 @@ object Running {
...
+    Log[F].info(s"Received ForkChoiceTipRequest from ${peer.endpoint.host}") >> MultiParentCasper
+      .forkChoiceTip(
+        casper
+      ) >>= (
         tip =>
           Log[F].info(
-            s"Streaming block ${PrettyPrinter.buildString(tip.blockHash)} to $peer"
-          ) >> streamToPeer(peer)(ToPacket(tip.toProto))
+            s"Streaming block ${PrettyPrinter.buildString(tip.blockHash)} to ${peer.endpoint.host}"
+          ) >> CommUtil[F].streamToPeer(peer, ToPacket(tip.toProto))
```
tgrospic: I love combinators but if they are sprinkled everywhere in the code it's difficult to follow the execution. Just by assigning explicit names can make things much more readable and gives more insight what are the "dependencies" to execute the final expression.
```scala
val logRequestedTip = Log[F].info(s"Received ForkChoiceTipRequest from ${peer.endpoint.host}")
def logStreamingBlock(block: BlockMessage) = Log[F].info(
  s"Streaming block ${PrettyPrinter.buildString(block.blockHash)} to ${peer.endpoint.host}"
)
val forkChoiceTip = MultiParentCasper.forkChoiceTip(casper)
def streamToPears(tip: BlockMessage) =
  logStreamingBlock(tip) >> CommUtil[F].streamToPeer(peer, ToPacket(tip.toProto))

logRequestedTip >> forkChoiceTip >>= streamToPears
```
This is also why we need functions to be "referentially transparent" so that expression
```scala
val forkChoiceTip = MultiParentCasper.forkChoiceTip(casper)
```
will not actually do anything until used/run in monadic sequence.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -284,107 +250,115 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+  ): F[Boolean] =
+    withCasper(
+      casper => {
+        validatorId match {
+          case None =>
+          case Some(id) => {
+            val self = ByteString.copyFrom(id.publicKey.bytes)
+            val _ = F.whenA(b.sender == self)(
+              Log[F].warn(
+                s"There is another node $peer proposing using the same private key as you. Or did you restart your node?"
```
tgrospic: This expression is also "in the air", Log.warn will never be executed. Instead of `Unit` it should return `F[Unit]` and executed as part of for comprehension bellow.

casper/src/main/scala/coop/rchain/casper/engine/BlocksRetriever.scala
nzpr: If node did not receive block, e.g. due to networking issue - it will keep asking peers (re-requesting) that reported they have block, one by one. Once block is in Casper, hash is be removed from BlockRetriever scope, so its not subject to this routine.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: I was hoping to get rid of this `.toMap` and make everything more memory efficient. I was concerned about consistency as well. But I recall you've ben  something about consistency across all our KVstores, so was hoping to get some solution and didn't want to load the whole database in memory and update both memory and persisted views.

node/src/main/resources/logback.xml
```diff
@@ -8,19 +8,19 @@
...
+      <pattern>%d{HH:mm:ss.SSS} [%highlight(%-5level{5})] [%cyan(%-20.20thread{20})] [%cyan(%-29.29logger{29})] - %msg%n </pattern>
     </encoder>
   </appender>
 
   <appender name="FILE" class="ch.qos.logback.core.FileAppender">
     <file>${rnode.data.dir}/rnode.log</file>
     <append>true</append>
     <encoder>
-      <pattern>%d{ISO8601, UTC} [%thread] %-5level %logger - %msg%n</pattern>
+      <pattern>%d{ISO8601, UTC} [%highlight(%-5level{5})] [%cyan(%-20.20thread{30})] [%cyan(%-29.29logger{29})]- %msg%n </pattern>
```
tgrospic: Maybe this should be in a separate PR, because it's unrelated change.

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -10,23 +10,31 @@ object PrettyPrinter {
...
+                s"-- M Parent Hash ${buildString(mainParent)} " +
+                s"-- Contents ${buildString(b.body.state)}" +
+                s"-- Shard ID ${limit(b.shardId, 10)}"
+          )
+      case true =>
+        b.header.parentsHashList.headOption
+          .fold(s"Block ${buildString(b.blockHash)} with missing elements")(
+            mainParent => s"#${b.body.state.blockNumber} (${buildString(b.blockHash)})"
+          )
+    }
```
tgrospic: Why not just simple if..else? And also inside of fold because empty parent case is the same.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -98,140 +67,122 @@ object Running {
...
-            ts <- Time[F].currentMillis
-          } yield hash -> Option(modifiedRequested(ts))
-        case _ =>
-          val warnMessage = s"Could not retrieve requested block ${PrettyPrinter.buildString(hash)} " +
-            s"from ${requested.peers.mkString(",")}. Removing the request from the requested blocks list. " +
-            s"Casper will have to re-request the block."
-          Metrics[F].incrementCounter("block-retrieve-failed") >>
-            Log[F].warn(warnMessage).as(hash -> none[Requested])
-      }
+      _ <- lockMap.remove(blockMessage.blockHash).pure[F]
```
tgrospic, line 137: In a situation where multiple threads enter this method. The first will start execution and others will wait. After the first is finished, the lock will be removed. The same lock for the one currently executing (entered second) and others that are still waiting. So now `lockMap` does not have lock for this block and next request will start execution immediately in parallel.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: I just turned persistent representation into a map Hash -> Set[Hash]

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: I've adopted this approach in Running object

node/src/main/resources/logback.xml
nzpr: Sure, I agree

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: This is the intention here - we just need to hold all handlers for particular block except for one that should go through the code inside lock and verify it. Then, when the hoard will go forward because there is no lock - they should be stopped on `ignoreMessageF` call. The goal here is to not execute block validation in parallel.

nzpr: Thanks for highlighting this, actually when blocks are all invalid, we have to validate them all. But idk whats the best way to mitigate this attack vector.

nzpr:
We decided first to merge <https://github.com/rchain/rchain/pull/2935> into dev and then I'll rebase this one onto dev.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,107 @@
...
+  * @param blockDependencyDag - in-memory dependency DAG, recreated from parentsStore on node startup
+  */
+final class CasperBufferKeyValueStorage[F[_]: Concurrent: Log] private (
+    parentsStore: KeyValueTypedStore[F, BlockHash, Set[BlockHash]],
+    blockDependencyDag: Cell[F, DoublyLinkedDag[BlockHash]]
+) extends CasperBufferStorage[F] {
+
+  override def addRelation(parent: BlockHash, child: BlockHash): F[Unit] =
+    for {
+      parents <- parentsStore.get(child).flatMap(_.getOrElse(Set.empty[BlockHash]).pure[F])
```
tgrospic: Mapping function doesn't have to return `F` value, so functor is enough here. Similar situation is on other places like `getParents`.

In essence this is always true. With pure, value is wrapped in F just to be unwrapped with flatMap.
`x.flatMap(_.pure) == x.map(_)`

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,107 @@
...
+  override def remove(hash: BlockHash): F[Unit] =
+    for {
+      curState                                  <- blockDependencyDag.read
+      (newState, hashesAffected, hashesRemoved) = DoublyLinkedDagOperations.remove(curState, hash)
+      _                                         <- blockDependencyDag.set(newState)
+      changes                                   = hashesAffected.map(h => (h, newState.childToParentAdjacencyList(h))).toSeq
+      deletes                                   = hashesRemoved.toSeq
+      _                                         <- parentsStore.put(changes)
+      // If node crush here - KV store end up in inconsistent state. What's the good way to handle it?
+      _ <- parentsStore.delete(deletes)
```
tgrospic: We need to add transactions to KV store to abstract over support for transactions in LMDB.

For now, maybe if delete is first then some blocks will be missing and repeated, but the state will be consistent. Is this makes sense?

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,107 @@
...
+
+  override def getChildren(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
+    blockDependencyDag.read.flatMap(_.parentToChildAdjacencyList.get(blockHash).pure[F])
+
+  override def getPendants: F[Set[BlockHash]] =
+    blockDependencyDag.read.flatMap(_.dependencyFree.pure[F])
+
+  // Block is considered to be in CasperBuffer when there is a records about its parents
+  override def contains(blockHash: BlockHash): F[Boolean] =
+    parentsStore.contains(blockHash) &&^ parentsStore.get(blockHash).flatMap(_.get.nonEmpty.pure[F])
```
tgrospic: Contains on KV store is optimized to skip loading a value. When _get_ is used, key and value will be loaded so calling _contains_ before that does not add anything, just one more call to LMDB.
https://github.com/rchain/rchain/blob/cbd0d5624a3da648c9fc71c6daad93fe7eb7d83f/shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreCodec.scala#L74-L79

1. Theoretically only genesis can be without parents, for which this will return false. Is it possible to insert that kind of block so this additional check is necessary?
2. Why not check `blockDependencyDag` is hash exists?

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,107 @@
...
+  override def isPendant(blockHash: BlockHash): F[Boolean] =
+    getPendants.flatMap(_.contains(blockHash).pure[F])
+}
+
+object CasperBufferKeyValueStorage {
+
+  def create[F[_]: Concurrent: KeyValueStoreManager: Log: Metrics]: F[CasperBufferStorage[F]] = {
+
+    def createDB[K, V](name: String, kCodec: Codec[K], vCodec: Codec[V]) =
+      KeyValueStoreManager[F].database(name).map(_.toTypedStore(kCodec, vCodec))
```
tgrospic: This function is now repeated on many places. Please define it as an extension method to `KeyValueStoreManager`. It can be defined like overload to `database` method with additional Codec parameters.

[KeyValueTypedStoreSyntax](https://github.com/rchain/rchain/blob/cbd0d5624a3da648c9fc71c6daad93fe7eb7d83f/shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala) can be used as an example.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,107 @@
...
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import scodec.Codec
+
+/**
+  * @param parentsStore - persistent map {hash -> parents set}
+  * @param blockDependencyDag - in-memory dependency DAG, recreated from parentsStore on node startup
+  */
+final class CasperBufferKeyValueStorage[F[_]: Concurrent: Log] private (
+    parentsStore: KeyValueTypedStore[F, BlockHash, Set[BlockHash]],
+    blockDependencyDag: Cell[F, DoublyLinkedDag[BlockHash]]
```
tgrospic: Is there any reason to use `Cell` instead of cats `Ref`? Cell is under investigation for further use. FYI @zsluedem

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferStorage.scala
```diff
@@ -0,0 +1,24 @@
...
+  def addRelation(parent: BlockHash, child: BlockHash): F[Unit]
+  def remove(blockHash: BlockHash): F[Unit]
+  def contains(blockHash: BlockHash): F[Boolean]
+  def isPendant(blockHash: BlockHash): F[Boolean]
+  def size: F[Long]
+  def toDoublyLinkedDag: F[DoublyLinkedDag[BlockHash]]
+}
+
+object CasperBufferStorage {
+  def apply[F[_]](implicit B: CasperBufferStorage[F]): CasperBufferStorage[F] = B
```
tgrospic: Usually, function parameters are camel case. I'm OK if we always name this `instance` or similar.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,134 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+                   .info(
+                     s"Block ${PrettyPrinter.buildString(b, short = true)} " +
+                       s"has already been processed."
+                   )
+                   .as(BlockStatus.processing.asLeft[ValidBlock])
+               } else {
+                 for {
+                   validSender <- Validate.blockSenderHasWeight(b, genesis, dag)
+                   resultWithUpdDag <- if (!validSender)
+                                        // TODO What's the scenario when this can happen?
```
tgrospic, line 239: What you mean by this TODO?

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -46,13 +46,17 @@ object DeployError {
...
+  def addBlock(b: BlockMessage, allowAddFromBuffer: Boolean = false): F[ValidBlockProcessing]
   def contains(hash: BlockHash): F[Boolean]
+  def dagContains(hash: BlockHash): F[Boolean]
+  def bufferContains(hash: BlockHash): F[Boolean]
   def deploy(d: Signed[DeployData]): F[Either[DeployError, DeployId]]
   def estimator(dag: BlockDagRepresentation[F]): F[IndexedSeq[BlockHash]]
   def createBlock: F[CreateBlockStatus]
   def getGenesis: F[BlockMessage]
   def getValidator: F[Option[PublicKey]]
+  def getVersion: F[Long]
```
tgrospic, line 60: TODO: We need to rethink Casper interface. Basically only estimator is casper operation and all other stuff are about the dag state, which should be parameter to casper.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,134 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+      _        <- Span[F].mark("reattempt-buffer")
+      children <- casperBuffer.getChildren(blockMessage.blockHash)
+      dag      <- blockDag
+      depFreeChildren <- children match {
+                          case None => List.empty[BlockHash].pure[F]
+                          case Some(s) =>
+                            s.toList.filterA { child =>
+                              for {
+                                childBlock     <- BlockStore[F].get(child)
+                                justifications = childBlock.get.justifications
```
tgrospic: TODO: We should avoid directly extract values. If block must be in the store some more descriptive error should be thrown.
In these cases we can define _unsafe_ helper like the lookup on dag state.
https://github.com/rchain/rchain/blob/8fb7bc466ea9a7b516133da40b85da95d11595bb/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala#L23-L26

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -401,23 +408,42 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
-      missingDependencies <- dependenciesHashesOf(b).filterA(dag.lookup(_).map(_.isEmpty))
-      missingUnseenDependencies <- missingDependencies.filterA(
-                                    blockHash => ~^(BlockStore[F].contains(blockHash))
-                                  )
-      _ <- missingDependencies.traverse(addDependencyToDag(_, b))
-      _ <- missingUnseenDependencies.traverse(CommUtil[F].sendBlockRequest)
+      dag <- blockDag
+      // Request blocks that are not in DAG and not in CasperBuffer
+      allDeps      = dependenciesHashesOf(b)
+      depsInDag    <- allDeps.filterA(dag.lookup(_).map(_.nonEmpty))
```
tgrospic: Why not use `dag.contains` here? It's more lightweight then loading the whole block metadata with lookup.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -401,23 +408,42 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+      depsInBuffer <- allDeps.filterA(casperBuffer.contains)
+      missingDeps  = allDeps filterNot (d => depsInDag.contains(d) || depsInBuffer.contains(d))
+      _            <- (missingDeps ++ depsInBuffer).traverse(addDependencyToBuffer(_, b))
+      depsInDaglogStr = depsInDag
+        .foldLeft("") { (str, dep) =>
+          str + s" ${PrettyPrinter.buildString(dep)}"
```
tgrospic: This can be simplified to `s"$str ${PrettyPrinter.buildString(dep)}"`.
The same is repeated to all three cases. It would be useful to create simple helper function and maybe it will fit in one line.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -401,23 +408,42 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
 
-  private def addDependencyToDag(hash: BlockHash, childBlock: BlockMessage): F[Unit] =
-    Cell[F, CasperState].modify(
-      s =>
-        s.copy(
-          dependencyDag = DoublyLinkedDagOperations.add(s.dependencyDag, hash, childBlock.blockHash)
-        )
-    )
+  private def addDependencyToBuffer(hash: BlockHash, childBlock: BlockMessage): F[Unit] =
+    casperBuffer.addRelation(hash, childBlock.blockHash)
```
tgrospic: Is it necessary to have this function, it's called only from one place?

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -57,13 +58,15 @@ object Validate {
...
     for {
-      _ <- Log[F]
-            .info(
-              s"Block already signed by: ${signatories
-                .map(x => "<" + Base16.encode(x.toByteArray).substring(0, 10) + "...>")
-                .mkString(", ")}"
-            )
+      _ <- Log[F].info(logMsg)
 
       result <- if (signatories.size >= requiredSignatures)
```
tgrospic: I know it's not your code but this can be simplified.
```scala
result = signatories.size >= requiredSignatures

_ <- Log[F].warn(...).whenA(!result)
```

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -67,12 +68,29 @@ object CasperLaunch {
...
+          _ <- pendantsReceived
+                .traverse_(
+                  pBlock =>
+                    BlockStore[F].get(pBlock).flatMap {
+                      case None => ().pure[F]
+                      case Some(b) =>
+                        Log[F].info(
+                          s"Pendant ${PrettyPrinter.buildString(b, short = true)} " +
+                            s"is available in BlockStore, sending to Casper."
+                        ) >> casper.addBlockFromStore(b.blockHash, allowAddFromBuffer = true).void
```
tgrospic: It would be useful to have a comment about why result is ignored here.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -1,68 +1,38 @@
...
-  private def requestForNewBlock[F[_]: Monad: RPConfAsk: TransportLayer: RequestedBlocks: Log: Time](
-      peer: PeerNode,
-      blockHash: BlockHash
-  ) = requestForBlock(peer, blockHash) >> addNewEntry(blockHash, Some(peer))
+  trait CasperMessageStatus
+  final case object BlockIsInDag            extends CasperMessageStatus
+  final case object BlockIsInCasperBuffer   extends CasperMessageStatus
+  final case object BlockIsWaitingForCasper extends CasperMessageStatus
+  final case object DoNotIgnore             extends CasperMessageStatus
+  final case class IgnoreCasperMessageStatus(doIgnore: Boolean, status: CasperMessageStatus)
```
tgrospic, line 39: It looks like `IgnoreCasperMessageStatus` is another message state but it not. It would be more clear with one empty line in between.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: You are right, better to check `blockDependencyDag` as we have it.
But should not `&&^` be lazy? so if `contains` on KV store return false, `get` is not executed?

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: Not really, I just used same way it was used in RequestedBlocks. Should I switch 
https://github.com/rchain/rchain/blob/6734c6415e3e078bb08997cfa6f63fc316c2489d/casper/src/main/scala/coop/rchain/casper/engine/BlockRetriever.scala#L76
to Ref as well?

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
tgrospic: Yes you are right, for _false_ case execution is the same.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
tgrospic: Yes please, it should be straightforward to switch to Ref.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -58,13 +58,15 @@ object Validate {
         if verifySig(candidateBytesDigest, signature.sig.toByteArray, publicKey.toByteArray)
       } yield publicKey).toSet
 
+    val logMsg = signatories.isEmpty match {
+      case true => s"ApprovedBlock is self-signed by ceremony master."
+      case false =>
+        s"ApprovedBlock is signed by: ${signatories
+          .map(x => "<" + Base16.encode(x.toByteArray).substring(0, 10) + "...>")
```
zsluedem, line 65: I think we have `Pretty.buildString`.

block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
```diff
@@ -60,7 +62,12 @@ object DoublyLinkedDagOperations {
 
   // If the element doesn't exist in the dag, the dag is returned as is
   @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
-  def remove[A](dag: DoublyLinkedDag[A], element: A): DoublyLinkedDag[A] = {
+  def remove[A](
+      dag: DoublyLinkedDag[A],
+      element: A
+  ): (DoublyLinkedDag[A], mutable.Set[A], mutable.Set[A]) = {
+    val childrenAffected                           = mutable.Set.empty[A]
+    val childrenRemoved                            = mutable.Set.empty[A]
```
tgrospic: Better to use immutable Set. Mutable structure if contained inside function can be justified but returning mutable object is calling for trouble.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: @tgrospic I just figured out what you meant by `Only genesis can be without parents, for which this will return false`. 

Let me clarify something. Block first end up in buffer when it is met as a dependency. At that point it is inserted as some another blocks dependancy and is treated as pendant block - the bottommost leaf of the buffer DAG. https://github.com/rchain/rchain/blob/6424ec9a028795840141822a725ea04996b91431/block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala#L63-L64
It does not have any parents in buffer - Casper did not received yet this block and node is not aware what are its parents. Once parents info is available - it means Casper received block, so it is treated as `in buffer`.
So there is plenty of situations when parentsStore does not have hash as a key. To make sure block is in Casper buffer I have to check if it is at all in parents map as a key, and fail here fast returning `false`, and if it is  check if parents are populated.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: IMO delete at the end is safer, as it looks more like a cleanup of leafs which parents are all deleted.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
nzpr: Not that straightforward, but - done. Just for the note - using Cell is risky, as we figured out.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Removed

casper/src/main/scala/coop/rchain/casper/Validate.scala
nzpr: I believe I did not touch this line, and don't see such helper in PrettyPrinter

block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
nzpr: Is there an elegant way to use immutable Set taking into account that we disabled var? Using e.g. Ref just to update immutable structure looks excessive.  For now I'm converting mutable set to immutable on return.

tgrospic: It's easy to convert this code to use immutable structure. This additional state should be also part of accumulator used when folding over children.
```scala
val initAcc = (childToParentAdjacencyList, Set.empty[A], Set.empty[A], Set.empty[A])
```
Maybe even helper case class would be useful, to be more clear about this temporary state.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,209 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
-    }
+  private def safetyRangeFilled: F[Boolean] = {
+    import cats.instances.list._
+    for {
+      approvedBlock  <- getGenesis
+      abHasLowHeight = ProtoUtil.blockNumber(approvedBlock) < expirationThreshold
 
-    import cats.instances.either._
+      // validators as per approved block state
+      abValidators = approvedBlock.body.state.bonds.map(_.validator)
```
tgrospic: Do we want to exclude slashed validators?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,209 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+    * Add block to Casper
+    *
+    * @param b                  Block
+    * @param allowAddFromBuffer If block is in CaspeBuffer, try to add it instead of reporting already processed.
+    *                           This is required for continuing syncing after node restart
+    */
+  def addBlock(b: BlockMessage, allowAddFromBuffer: Boolean): F[ValidBlockProcessing] =
+    for {
+      // We force Casper to be single threaded. More investigations needed to allow it be multi threaded.
+      casperLocked <- ~^(blockProcessingLock.tryAcquireN(1L))
```
tgrospic: This is not safe if error is thrown and _release_ is never called.

Bracket can ensure that release is called always like with `try...finally`. We have similar situation in Block API.
https://github.com/rchain/rchain/blob/8fb7bc466ea9a7b516133da40b85da95d11595bb/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L86-L87

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,209 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+      dag      <- blockDag
+      pendants <- CasperBufferStorage[F].getPendants
+      depFreePendants <- pendants.toList.filterA { pendant =>
+                          for {
+                            pendantBlock   <- BlockStore[F].get(pendant)
+                            justifications = pendantBlock.get.justifications
+                            // If even one of justifications is not in DAG - block is not dependency free
+                            depFree <- justifications.foldLeftM(true)(
+                                        (r, j) => r.pure[F] &&^ dag.contains(j.latestBlockHash)
+                                      )
```
tgrospic: You could use `justifications.existsM` to just fold until first justification is not found.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -715,6 +716,8 @@ object BlockAPI {
           "No action taken since other thread is already processing the block."
             .asLeft[String]
             .pure[F]
+        case BlockError.CasperIsBusy =>
+          s"Casper put block in the queue: $status".asLeft[String].pure[F]
```
tgrospic, line 722: Can we merge `Processing` and `CasperIsBusy`? Seems like they both represent the same state.

casper/src/main/scala/coop/rchain/casper/engine/BlockRetriever.scala
```diff
@@ -0,0 +1,312 @@
...
+                      s" of $admitHashReason."
+                  )
+                case Ignore => ().pure[F]
+              }
+          _ <- if (result.broadcastRequest)
+                CommUtil[F].broadcastHasBlockRequest(hash)
+              else ().pure[F]
+          _ <- if (result.requestBlock)
+                CommUtil[F].requestForBlock(peer.get, hash)
+              else ().pure[F]
```
tgrospic, line 207: When result is not needed this can be in one line.
`CommUtil[F].broadcastHasBlockRequest(hash).whenA(result.broadcastRequest)`

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
tgrospic: Oh, I see. Casper is not involved here, just block validation and saving to store.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferStorage.scala
```diff
@@ -0,0 +1,24 @@
...
+trait CasperBufferStorage[F[_]] {
+  def getParents(blockHash: BlockHash): F[Option[Set[BlockHash]]]
+  def getChildren(blockHash: BlockHash): F[Option[Set[BlockHash]]]
+  def getPendants: F[Set[BlockHash]]
+  def addRelation(parent: BlockHash, child: BlockHash): F[Unit]
+  def remove(blockHash: BlockHash): F[Unit]
+  def contains(blockHash: BlockHash): F[Boolean]
+  def isPendant(blockHash: BlockHash): F[Boolean]
+  def size: F[Long]
+  def toDoublyLinkedDag: F[DoublyLinkedDag[BlockHash]]
```
tgrospic, line 20: `getParents`, `getChildren`, `size` and `isPendant` are not used. Do you have some plan for that?

They are also redundant because `toDoublyLinkedDag` gives all necessary info so these can be extension functions.

Or maybe from another perspective. Why to expose all internals with `DoublyLinkedDag[BlockHash]` and not just `getChildren` which is used here?
https://github.com/rchain/rchain/blob/36bf1e0f183ab04e925e0e49ca8c42537183c294/casper/src/main/scala/coop/rchain/casper/EquivocationDetector.scala#L52-L56

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,210 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+                   for {
+                     dag                   <- blockDag
+                     missingDepCheckResult <- Validate.missingBlocks(b, dag)
+                     dagMissingBlockDependencies = missingDepCheckResult.equals(
+                       Left(InvalidBlock.MissingBlocks)
+                     )
+                     isReadyApprovedBlockChild = (isApprovedBlockChild(b) &&^ safetyRangeFilled)
+                     blockIsReady              <- ~^(dagMissingBlockDependencies.pure[F]) ||^ isReadyApprovedBlockChild
+
+                     _ <- fetchMissingDependencies(b).unlessA(!dagMissingBlockDependencies)
```
tgrospic: This is like negation on both sides, in the function (unless) and in the argument. _When dagMissingBlockDependencies_ sounds more natural.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,210 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+                 case None =>
+                   returnNoBlockInStore
+                 case Some(b) =>
+                   for {
+                     dag                   <- blockDag
+                     missingDepCheckResult <- Validate.missingBlocks(b, dag)
+                     dagMissingBlockDependencies = missingDepCheckResult.equals(
+                       Left(InvalidBlock.MissingBlocks)
+                     )
+                     isReadyApprovedBlockChild = (isApprovedBlockChild(b) &&^ safetyRangeFilled)
```
tgrospic: Brackets are not necessary.

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -55,7 +57,40 @@ object CasperLaunch {
...
+                s"Checking pendant hashes: ${pendantsReceived.size} items in CasperBuffer."
+              )
+          _ <- pendantsReceived
+              // we just need to send blocks to Casper. Nothing to do with results of block processing here,
+              // so ignoring them
+                .traverse_(
+                  pBlock =>
+                    BlockStore[F].get(pBlock).flatMap {
+                      case None => ().pure[F]
+                      case Some(b) =>
```
tgrospic: `pendantsReceived` are checked for existence in BlockStore so there is no need to match on None case here.

Unsafe variant of _get_ would be useful to throw some more meaningful error like _BlockStore inconsistency error_. This is also similar to https://github.com/rchain/rchain/pull/2943#discussion_r441538509.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -70,113 +56,210 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+        .exception(
+          new Exception("Casper tries to add block that is not present in BlockStore")
         )
-        .as(BlockStatus.processing.asLeft)
+        .asLeft[ValidBlock]
+        .pure[F]
+
+      result <- blockAvailable match {
+                 case None =>
+                   returnNoBlockInStore
```
tgrospic, line 95: This should never happen, right? It's adding block from the store. This is also similar to https://github.com/rchain/rchain/pull/2943#discussion_r441538509.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
nzpr: `Processing` actually means that block was already processed - it is returned when block acquired block and node found that this block was already aded. We should rename it I believe https://github.com/rchain/rchain/blob/10889ddd8976473853804cceeccaac4e292a8215/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L202-L208

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: @tgrospic yes it should not, but as its public method, it an be called in any way, so I put this 
https://github.com/rchain/rchain/pull/2943#discussion_r441538509 totally makes sense

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
nzpr: Actually looks like `Processing` usage is not consistent 
https://github.com/rchain/rchain/blob/9e1cf7d9fcaeb098f22cc890a12abf9f4609eabd/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L715-L718
here it means that block is in process of adding. But its odd to see it being used in createBlock

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: Oh yes, that makes sense.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferStorage.scala
nzpr: @tgrospic yes I believe its better to have this methods on CasperBuffer then `blockBufferDependencyDag.parentToChildAdjacencyList.contains(block.blockHash) `
I'm not sure yet about usage of these methods but it looks natural to have them, so I've added

nzpr: Hard to follow whats the logic behind `requestedAsDependency`

nzpr: I'd like to leave it as it is for now. We will refactor later when working on Casper.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: I mean I'm not sure this case can happen, and want it to be investigated later

casper/src/main/scala/coop/rchain/casper/Casper.scala
nzpr: @tgrospic Should I create github issue for that?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: I overlooked that one, thanks! `contains` should be here

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Open github issue?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -79,199 +53,219 @@ object Running {
...
+    * @param ignoreMessageF Defines when to ignore message.
+    * @param validateBlockF Defines basic block validation before persisting.
+    * @param casperAddF Function that sends block to Casper for inserting in the DAG
+    * @param lockMap       BlockMessage is a heavy message, so excessive processing
+    *                      of duplicate BlockMessages should be discouraged. This is map of Semaphore to block hash.
+    * @return
     */
-  def maintainRequestedBlocks[F[_]: Monad: RPConfAsk: RequestedBlocks: TransportLayer: Log: Time: Metrics](
-      timeout: FiniteDuration
+  def handleBlockMessage[F[_]: Monad: TransportLayer: RPConfAsk: BlockRetriever: BlockStore: Concurrent: Time: Metrics: Log](
```
tgrospic: TransportLayer, RPConfAsk and Time are not used. And Monad is covered by Concurrent.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -79,199 +53,219 @@ object Running {
...
-        case _ =>
-          val warnMessage = s"Could not retrieve requested block ${PrettyPrinter.buildString(hash)} " +
-            s"from ${requested.peers.mkString(",")}. Removing the request from the requested blocks list. " +
-            s"Casper will have to re-request the block."
-          Metrics[F].incrementCounter("block-retrieve-failed") >>
-            Log[F].warn(warnMessage).as(hash -> none[Requested])
+  /**
+    * Peer broadcasted block hash.
+    */
+  def handleBlockHashMessage[F[_]: Sync: BlockRetriever: Log](
```
tgrospic: Monad is enough here instead of Sync.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -79,199 +53,219 @@ object Running {
...
-      ToPacket(approvedBlock.toProto)
-    ) >> Log[F].info(s"ApprovedBlock sent to $peer")
+    Log[F].info(s"Received ApprovedBlockRequest from ${peer.endpoint.host}") >>
+      TransportLayer[F].streamToPeer(peer, approvedBlock.toProto) >>
+      Log[F].info(s"ApprovedBlock sent to ${peer.endpoint.host}")
 
 }
 
-class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell: RPConfAsk: Log: Time: Running.RequestedBlocks](
+class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell: RPConfAsk: Log: Time: BlockRetriever: CasperBufferStorage: Concurrent: Metrics](
```
tgrospic: Concurrent can replace Sync because it's already part of it.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -881,20 +896,17 @@ object NodeRuntime {
...
-              Log[F],
-              Time[F],
-              Metrics[F]
-            )
-        _ <- Time[F].sleep(conf.casper.casperLoopInterval)
-      } yield ()
+      casperLoop = {
+        implicit val br = blockRetriever
+        for {
+          _      <- Time[F].sleep(conf.casper.casperLoopInterval)
```
tgrospic: I'd expect for delay to be at the end so that the first execution happens immediately.

shared/src/main/scala/coop/rchain/store/KeyValueStoreManagerSyntax.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.store
+
+import cats.Functor
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.shared.syntax._
+import scodec.Codec
+
+trait KeyValueStoreManagerSyntax {
+  implicit final def sharedSyntaxKeyValueStoreManager[F[_]: Sync: Functor](
```
tgrospic, line 10: Functor is already part of Sync.

shared/src/test/scala/coop/rchain/store/InMemoryKeyValueStoreSpec.scala
```diff
@@ -12,10 +12,9 @@ import scodec.codecs.{int64, utf8}
 class KeyValueStoreSut[F[_]: Sync: KeyValueStoreManager] {
   def copyToDb(data: Map[Long, String]): F[KeyValueTypedStore[F, Long, String]] =
     for {
-      db    <- KeyValueStoreManager[F].database("test")
-      store = db.toTypedStore(int64, utf8)
-      _     <- store.put(data.toSeq)
-    } yield store
+      db <- KeyValueStoreManager[F].database("test", int64, utf8)
+      _  <- db.put(data.toSeq)
+    } yield db
```
tgrospic, line 17: :+1:

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
nzpr: We don't need it to be executed immediately - as on startup node requests fork-choice tip, and we have logic in CaperLaunch.

But makes sense, maybe we can merge this logic.

nzpr: Order changed

nzpr:
We need to reenable this tests in new incarnation https://github.com/rchain/rchain/pull/2943/commits/e07fce38a8199fdba6a4838d0492ac612b6c0b58, I suggest to do it in another PR.

I tested this on sandbox with multiparent to make sure network is progressing well and add blocks from the queue.

Also on mainnet observer node to test catch up. Node successfully download all blocks and started to replay them.
Catch up mode requires about 1GB RAM.
<br/>
Screenshot 2020-07-01 at 12 20 09

nzpr:
Issue for tests rewriting https://github.com/rchain/rchain/issues/2982

casper/src/main/scala/coop/rchain/casper/engine/BlockRetriever.scala
```diff
@@ -0,0 +1,327 @@
...
+          result <- RequestedBlocks[F]
+                     .modify[AdmitHashResult] { state =>
+                       val unknownHash = !state.contains(hash)
+                       if (unknownHash)
+                         (
+                           addNewRequest(state, hash, now, sourcePeer = peer),
+                           AdmitHashResult(
+                             NewRequestAdded,
+                             broadcastRequest = if (peer.nonEmpty) false else true,
+                             requestBlock = if (peer.nonEmpty) true else false
```
tgrospic: These kind of expressions can be simplified without using if..then. On line 180 is similar case.
```scala
broadcastRequest = peer.isEmpty,
requestBlock = peer.nonEmpty
```

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -274,110 +279,142 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+      )
+    case b: BlockMessage =>
+      handleBlockMessage(b, peer)(
+        ignoreCasperMessage,
+        blockIsOk,
+        sendBlockFromStoreToCasper,
+        hashHandlerLock
+      )
+    case br: BlockRequest => handleBlockRequest(peer, br)
+    // TODO should node say it has block only after it is in DAG, or CasperBuffer is enough? Or even just BlockStore?
```
tgrospic, line 412: If block is not in the DAG yet we cannot be sure if execution is valid but on the other hand we are validating signature and if this is block from bonded validator it will be slashed.
I think for now it's better to restrict only to blocks in the DAG until we have more input validation and better buffering when receiving blocks.

models/src/main/scala/coop/rchain/casper/PrettyPrinter.scala
```diff
@@ -63,4 +68,11 @@ object PrettyPrinter {
 
   def buildString(b: Bond): String =
     s"${buildStringNoLimit(b.validator)}: ${b.stake.toString}"
+
+  def buildString(hashes: List[BlockHash]): String =
+    s"[${hashes
+      .foldLeft("") { (str, dep) =>
+        s"$str ${PrettyPrinter.buildString(dep)}"
+      }
+      .trim}]"
```
tgrospic: With `mkString` method it's easier to join strings. Similar to this code.
https://github.com/rchain/rchain/blob/8489c6ca0f2ff8c3df1ebb8a63ca717c1a6cb07a/node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala#L28-L29

block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
```diff
@@ -60,37 +62,59 @@ object DoublyLinkedDagOperations {
...
+          }
+        case None => initAcc
+      }
+    (new DoublyLinkedDag[A] {
       override val parentToChildAdjacencyList
           : Map[A, Set[A]]                                    = dag.parentToChildAdjacencyList - element
       override val childToParentAdjacencyList: Map[A, Set[A]] = updatedChildToParentAdjacencyList
       override val dependencyFree: Set[A]                     = dag.dependencyFree ++ newDependencyFree - element
-    }
+    }, childrenAffected.toSet, childrenRemoved.toSet)
```
tgrospic: This `toSet` is now not necessary.

casper/src/main/scala/coop/rchain/casper/Casper.scala
tgrospic: For the reference here issue you created #2971.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -71,113 +59,230 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+                            missingDep <- justifications
+                                           .map(_.latestBlockHash)
+                                           .existsM(dagContains(_).not)
+                          } yield !missingDep
+                        }
+      enqueued <- BlockRetriever[F].getEnqueuedToCasper
+      all      = enqueued ++ depFreePendants
+      _ <- Log[F].info(s"Blocks ready to be added: enqueued to Casper ${PrettyPrinter
+            .buildString(enqueued)}, buffer pendants ${PrettyPrinter.buildString(depFreePendants)}.")
+    } yield if (all.nonEmpty) all.head.some else None
```
tgrospic: This can be simplified to `all.headOption`.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -402,90 +507,64 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+      _ <- BlockRetriever[F].ackInCasper(b.blockHash)
+      _ <- Log[F].info(
+            s"Block ${PrettyPrinter.buildString(b, short = true)} missing dependencies. " +
+              s"Fetching: ${PrettyPrinter.buildString(missingDeps)}. " +
+              s"Already in CasperBuffer: ${PrettyPrinter.buildString(depsInBuffer)}. " +
+              s"Already in DAG: ${PrettyPrinter.buildString(depsInDag)}."
+          )
+      _ <- missingDeps.traverse(
+            BlockRetriever[F]
+              .admitHash(_, admitHashReason = BlockRetriever.MissingDependencyRequested)
```
tgrospic, line 537: Little more precise would be to use `traverse_` so that the result is `Unit` and not `List[Unit]`.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -402,90 +507,64 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
 
   def fetchDependencies: F[Unit] = {
     import cats.instances.list._
     for {
-      s <- Cell[F, CasperState].read
-      _ <- s.dependencyDag.dependencyFree.toList.traverse(CommUtil[F].sendBlockRequest)
+      pendants       <- casperBuffer.getPendants
+      pendantsUnseen <- pendants.toList.filterA(BlockStore[F].contains(_).not)
+      _ <- Log[F].debug(s"Requesting CasperBuffer pendant hashes, ${pendantsUnseen.size} items.") >>
+            pendantsUnseen.toList.traverse(
```
tgrospic: Here also with `traverse_` the result would be `Unit`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: Should we rename this Processing status to Processed?

casper/src/test/scala/coop/rchain/casper/util/rholang/Resources.scala
```diff
@@ -75,6 +76,17 @@ object Resources {
       }
     } yield blockDagStorage)
 
+  def mkCasperBuferStorate[F[_]: Concurrent: Sync: Log: Metrics](
```
tgrospic: Concurrent already extends Sync.

casper/src/test/scala/coop/rchain/casper/helper/NoOpsCasperEffect.scala
```diff
@@ -47,6 +49,11 @@ class NoOpsCasperEffect[F[_]: Sync: BlockStore: BlockDagStorage] private (
   def fetchDependencies: F[Unit]                                      = ().pure[F]
   def getGenesis: F[BlockMessage]                                     = Dummies.createBlockMessage().pure[F]
   def getValidator: F[Option[PublicKey]]                              = none.pure[F]
+  def getVersion: F[Long]                                             = 1L.pure[F]
+  def approvedBlockStateComplete: F[Boolean]                          = true.pure[F]
+  def addBlockFromStore(b: BlockHash, allowAddFromBuffer: Boolean): F[ValidBlockProcessing] =
+    for { _ <- ().pure[F] } yield BlockStatus.valid.asRight
```
tgrospic: This can be simplified to `BlockStatus.valid.asRight.pure[F]`.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -274,110 +279,143 @@ class Running[F[_]: Sync: BlockStore: CommUtil: TransportLayer: ConnectionsCell:
...
+          lastFinalizedStateBlock = approvedBlock.candidate.block
+          // TODO might be more checks can/should be applied here
+          validShard   = lastFinalizedStateBlock.shardId.equalsIgnoreCase(b.shardId)
+          validFormat  <- Validate.formatOfFields(b)
+          validSig     <- Validate.blockSignature(b)
+          validVersion <- casper.getVersion.flatMap(Validate.version(b, _))
+          // TODO 50 is hardcoded value expirationThreshold from MultiParentCasperImpl. Move to conf
+          oldBlock = (ProtoUtil.blockNumber(b) <= ProtoUtil.blockNumber(
+            lastFinalizedStateBlock
+          ) - 50).pure[F]
```
tgrospic: Expiration threshold should only be hardcoded on one place and use the same reference here.

block-storage/src/main/scala/coop/rchain/blockstorage/casperbuffer/CasperBufferKeyValueStorage.scala
```diff
@@ -0,0 +1,120 @@
...
+}
+
+object CasperBufferKeyValueStorage {
+  implicit private val ms: Source =
+    Metrics.Source(BlockStorageMetricsSource, "dag-key-value-store")
+
+  def create[F[_]: Concurrent: KeyValueStoreManager: Log: Metrics]: F[CasperBufferStorage[F]] = {
+
+    def createDB[K, V](name: String, kCodec: Codec[K], vCodec: Codec[V]) =
+      KeyValueStoreManager[F].store(name).map(_.toTypedStore(kCodec, vCodec))
```
tgrospic: One more place left for `KeyValueStoreManager[F].database` extension.

block-storage/src/main/scala/coop/rchain/blockstorage/util/DoublyLinkedDagOperations.scala
```diff
@@ -60,37 +62,59 @@ object DoublyLinkedDagOperations {
...
   @SuppressWarnings(Array("org.wartremover.warts.Throw")) // TODO remove throw
-  def remove[A](dag: DoublyLinkedDag[A], element: A): DoublyLinkedDag[A] = {
+  def remove[A](
+      dag: DoublyLinkedDag[A],
+      element: A
+  ): (DoublyLinkedDag[A], Set[A], Set[A]) = {
     val parentToChildAdjacencyList: Map[A, Set[A]] = dag.parentToChildAdjacencyList
     val childToParentAdjacencyList: Map[A, Set[A]] = dag.childToParentAdjacencyList
     assert(!childToParentAdjacencyList.contains(element))
     assert(!parentToChildAdjacencyList.values.toSet.contains(element))
```
tgrospic: Please if you can also fix this.

This is not correct and I'm not sure what Scala magic is involved to compile this. It's comparing _element_ of a Set with the Set of elements.
image

I think using iterators would make the most sense here.
`parentToChildAdjacencyList.valuesIterator.flatten.contains(element)`

casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
```diff
@@ -3,7 +3,8 @@ package coop.rchain.casper
 sealed trait BlockStatus
 object BlockStatus {
   def valid: ValidBlock                    = ValidBlock.Valid
-  def processing: BlockError               = BlockError.Processing
+  def processing: BlockError               = BlockError.Processed
+  def CasperIsBusy: BlockError             = BlockError.CasperIsBusy
```
tgrospic: Function should start with lowercase.

tgrospic (approved):
This has become a really big PR. Nice work!

I'm approving with a few minor comments. But more important I see tests for BlockRetriver which are now disabled as you mentioned in issue #2982. This should be done as part of 0.9.25.1 release.

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -131,7 +136,14 @@ object Setup {
     implicit val synchronyConstraintChecker   = SynchronyConstraintChecker[Task](0d)
     implicit val lastFinalizedConstraintChecker =
       LastFinalizedHeightConstraintChecker[Task](Long.MaxValue)
-    implicit val estimator = Estimator[Task](Estimator.UnlimitedParents, None)
+    implicit val estimator      = Estimator[Task](Estimator.UnlimitedParents, None)
+    implicit val blockRetriever = BlockRetriever.of[Task]
+
+    implicit val kvsManager = RNodeKeyValueStoreManager[Task](Path.of("tmp"))
+      .unsafeRunSync(monix.execution.Scheduler.Implicits.global)
```
tgrospic: Path for should be to system temporary directory. With `Path.of("tmp")` this is relative to the current executing directory.
https://github.com/rchain/rchain/blob/0dd10620378443aaa90da7dba3270202e437c7f2/casper/src/test/scala/coop/rchain/casper/engine/Setup.scala#L44

Another option is to use `InMemoryStoreManager[Task]` which is not persisted on disk.

zsluedem (approved):
Very good job. It is a very big pr. Currently, I can not find any problems because I only got a rough look at the pr. Since you and @tgrospic worked on this for long, I believe you solve most of the problems. Maybe some edge cases are not covered, but we can take that into next pr. 
Again, good work.

nzpr:
Thanks @zsluedem! Yes, I did quite thorough testing on live network and @tgrospic as well when working on LFS on top of this PR. Unit tests will go in separate PR soon.

nzpr:
bors merge

PR #2952, opened 2020-06-15
RCHAIN-4109: Alignment + colors for logs

nzpr:
This PR adds colors and alignment to logs for better readability to node console output. On logfile appender coloring is disabled - the only change is formatting.

Screenshot 2020-06-15 at 17 50 53

https://rchain.atlassian.net/browse/RCHAIN-4109

### DIFF
--- a/integration-tests/test/wait.py
+++ b/integration-tests/test/wait.py
@@ -45,7 +45,7 @@ class LogsContainMessage:
 
 class NodeStarted(LogsContainMessage):
     def __init__(self, node: 'Node') -> None:
-        super().__init__(node, 'coop.rchain.node.NodeRuntime - Listening for traffic on rnode')
+        super().__init__(node, 'Listening for traffic on rnode')
 
 
 class RunningStateEntered(LogsContainMessage):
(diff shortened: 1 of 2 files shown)


nzpr:
I need to fix that checkers in integration tests relying on particular log format.

zsluedem:
I mean this is configurable. Why does this need to go to dev?
People can do whatever configs they want on the log config. I don't understand why we should put this into dev.
It is also possible that this affects other sre tools like kibana.
I don't suggest we make this into dev.

zsluedem:
Alignment can take into consider. But color part is not that important.

nzpr:
@zsluedem one thing that I noticed about color - it is much easier to navigate ad find the source of bug in GH Actions logs when integration test failed. And in general to me it is much easier to navigate through logs. I suggest to put this into dev because default configuration is user facing, would be nice if they can see good looking log output right from the scratch. Most users won't even bother. And looking into logs is expected part of dApp dev workflow when running local node.

nzpr:
It should not interfere with kibana, because there is one more appender `LOGSTASH` for this, PR changes appenders for file and stdout. And anyway, we have logback conf file <https://github.com/rchain/rchain-testnet-node/blob/dev/logback.xml#L28> for our nodes which overrides defaults.

nzpr:
@zsluedem @tgrospic I disabled colours in logfile, leaving only formatting. Colouring is enabled only on stdout.

tgrospic (approved):
LGTM

zsluedem (approved):
Sorry for the late approve.

PR #2964, opened 2020-06-21
Fix LMDB commit transaction when execution freezes

### LINKED ISSUE #2963: LMDB store sometimes block in execution
Bug is reproducible with this test when number of validators are more than 19 (?).
https://github.com/rchain/rchain/blob/8fb7bc466ea/casper/src/test/scala/coop/rchain/casper/batch2/ValidateTest.scala#L391

Problem is showing is LMDB _commit_ is called with suspended execution (Sync.delay).
https://github.com/rchain/rchain/blob/98cbac41a0202840011d2d4977260029250fea78/casper/src/main/scala/coop/rchain/casper/storage/LmdbKeyValueStore.scala#L32-L37

tgrospic:
## Overview
LMDB has limitation for write transactions to be used on a single thread. 
https://lmdb.readthedocs.io/en/release/#threads

This PR fixes the implementation of LMDB store to not use suspended execution (Sync.delay) which can cause changing threads. This applies to read and write transactions because freezing of execution was noticed in both cases.

For the reference this was the buggy code.
https://github.com/rchain/rchain/blob/899954e422/casper/src/main/scala/coop/rchain/casper/storage/LmdbKeyValueStore.scala#L21-L50

---

In this PR is also fix for CI to pass tests. Memory limit is set separately for integration tests with lower value and with more memory for unit tests.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/storage/LmdbKeyValueStore.scala
+++ b/casper/src/main/scala/coop/rchain/casper/storage/LmdbKeyValueStore.scala
@@ -10,22 +10,22 @@ import coop.rchain.shared.Resources.withResource
 import coop.rchain.store.KeyValueStore
 import org.lmdbjava._
 
+import scala.util.control.NonFatal
+
 final case class DbEnv[F[_]](env: Env[ByteBuffer], dbi: Dbi[ByteBuffer], done: F[Unit])
 
 final case class LmdbKeyValueStore[F[_]: Sync](
     getEnvDbi: F[DbEnv[F]]
 ) extends KeyValueStore[F] {
 
-  private def withTxn[T](
-      isWrite: Boolean
-  )(op: (Txn[ByteBuffer], Dbi[ByteBuffer]) => F[T]): F[T] =
+  private def withReadTxn[T](op: (Txn[ByteBuffer], Dbi[ByteBuffer]) => F[T]): F[T] =
     for {
       // Acquire current LMDB environment and DBI interface
       dbEnv <- getEnvDbi
       // "Done" must be called to mark finished operation with the environment/dbi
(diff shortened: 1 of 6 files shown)


zsluedem (approved):
Very good found out~! I am thinking is this the cause of the tuple space mismatch also.

tgrospic:
@zsluedem this code is not used in RSpace currently. But we should investigate if similar scenario is involved with existing implementation when processing operations in parallel.
https://github.com/rchain/rchain/blob/2560ab8de280c727560d7819f23fed9dd4286a8f/rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala#L318-L321

casper/src/main/scala/coop/rchain/casper/storage/LmdbKeyValueStore.scala
```diff
@@ -10,78 +10,108 @@ import coop.rchain.shared.Resources.withResource
...
+//                   Sync[F].delay(txn.close()) *> done
+//                 case (txn, Error(_)) =>
+//                   // Close transaction on error.
+//                   Sync[F].delay(txn.close()) *> done
+//                 case (_, Canceled) =>
+//                   // When execution is canceled, closing transaction throws NotReadyException
+//                   //  in bracket `use` function which is not visible in bracket `release` (?).
+//                   done
+//               }
+//    } yield result
```
tgrospic: TEMP

nzpr (approved):
I cannot reproduce freeze anymore, so this helped!

tgrospic:
I moved change with memory limit on CI to this PR to be able to pass tests.

tgrospic:
bors merge

PR #2967, opened 2020-06-23
RCHAIN-4111: provide block hash on exploratory deploy 

zsluedem:
make explore deploy api can be configured by blockHash and isPostStateHash

## Overview
https://rchain.atlassian.net/browse/RCHAIN-4111

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -43,6 +43,10 @@ object BlockAPI {
 
   val reportTransformer = new ReportingProtoTransformer()
 
+  // TODO: we should refactor BlockApi with applicative errors for better classification
+  //  of errors and to overcome nesting when validating data.
+  final case class BlockRetrievalError(message: String) extends Exception
+
   def deploy[F[_]: Sync: EngineCell: Log: Span](
       d: Signed[DeployData]
   ): F[ApiErr[String]] = Span[F].trace(DeploySource) {
@@ -550,10 +554,6 @@ object BlockAPI {
     val errorMessage =
       "Could not get block, casper instance was not available yet."
 
-    // TODO: we should refactor BlockApi with applicative errors for better classification
-    //  of errors and to overcome nesting when validating data.
-    final case class BlockRetrievalError(message: String) extends Exception
-
     def casperResponse(
         implicit casper: MultiParentCasper[F]
     ): F[ApiErr[BlockInfo]] =
@@ -575,7 +575,7 @@ object BlockAPI {
         // Check if hash is complete and not just the prefix in which case
         // we can use `get` directly and not iterate over the whole block hash index.
         getBlock  = BlockStore[F].get(hashByteString)
-        findBlock = getBlockFromStore[F](hash)
+        findBlock = findBlockFromStore[F](hash)
         blockF    = if (hash.length == 64) getBlock else findBlock
         // Get block form the block store
         block <- blockF >>= (_.liftTo[F](
@@ -676,7 +676,9 @@ object BlockAPI {
       faultTolerance = faultTolerance
     ).pure[F]
 
-  def getBlockFromStore[F[_]: Monad: BlockStore](
+  // Be careful to use this method , because it would iterate the whole indexes to find the matched one which would cause performance problem
+  // Trying to use BlockStore.get as much as possible would more be preferred
+  private def findBlockFromStore[F[_]: Monad: BlockStore](
       hash: String
   ): F[Option[BlockMessage]] =
     for {
@@ -809,8 +811,18 @@ object BlockAPI {
     )
   }
 
-  def exploratoryDeploy[F[_]: Monad: EngineCell: Log: SafetyOracle: BlockStore](
-      term: String
+  /**
+    * Explore the data or continuation in the tuple space for specific blockHash
+    *
+    * @param term: the term you want to explore in the request. Be sure the first new should be `return`
+    * @param blockHash: the block hash you want to explore
+    * @param usePreStateHash: Each block has preStateHash and postStateHash. If usePreStateHash is true, the explore
+    *                       would try to execute on preState.
+    * */
+  def exploratoryDeploy[F[_]: Sync: EngineCell: Log: SafetyOracle: BlockStore](
+      term: String,
+      blockHash: Option[String] = none,
+      usePreStateHash: Boolean = false
   ): F[ApiErr[(Seq[Par], LightBlockInfo)]] = {
     val errorMessage =
       "Could not execute exploratory deploy, casper instance was not available yet."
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -810,7 +810,9 @@ object BlockAPI {
   }
 
   def exploratoryDeploy[F[_]: Monad: EngineCell: Log: SafetyOracle: BlockStore](
-      term: String
+      term: String,
+      blockHash: String,
+      isPostStateHash: Boolean
```
nzpr: Whats the reason behind this flag? Is it to compute something on state of merged parents for debug purposes?

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
```diff
@@ -124,7 +125,14 @@ object WebApiRoutes {
...
 
       case req @ POST -> Root / "explore-deploy" =>
-        req.handle[String, ExploratoryDeployResponse](webApi.exploratoryDeploy)
+        req.handle[String, ExploratoryDeployResponse](
+          term => webApi.exploratoryDeploy(term, "", true)
+        )
+
+      case req @ POST -> Root / "v2" / "explore-deploy" =>
+        req.handle[ExploreDeployRequest, ExploratoryDeployResponse](
+          req => webApi.exploratoryDeploy(req.term, req.blockHash, req.isPostStateHash)
```
tgrospic: We want to have this method in v1 also. I see this as an extension. Maybe we can call it `explore-deploy-by-block-hash` for clear separation from last finalized variant.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
zsluedem: There are `postStateHash` and `preStateHash` in a block. If the isPostStateHash is `false`, it would choose `preStateHash`. That's  my idea.

zsluedem: Maybe give another name for the flag which makes more sense. Any suggestions?

nzpr: I suggest make it `usePreStateHash`, with default value `false`, as general purpose of this method is to evaluate state of the network at particular block, so of `postStateHash`. I can see `usePreStateHash=true` for debug purposes, as this state does not represent something of end user's interest.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -810,7 +810,9 @@ object BlockAPI {
   }
 
   def exploratoryDeploy[F[_]: Monad: EngineCell: Log: SafetyOracle: BlockStore](
-      term: String
+      term: String,
+      blockHash: String,
+      usePreStateHash: Boolean
```
nzpr: @zsluedem why not set here default value `usePreStateHash: Boolean = false`? So this can be called as 
`BlockAPI.exploratoryDeploy(term, hash)`, and `BlockAPI.exploratoryDeploy(term, hash, true)` only when it is needed (I assume execution over preStatHash is a rare need)

tgrospic: I'm not sure that I see the reason to have pre-state hash. It can also be confusing for users and how we will explain when to use it. I'd be more for adding this kind of operations on admin (internal) API.

zsluedem: makes sense

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -826,12 +828,24 @@ object BlockAPI {
                            .pure[F]
                        case None =>
                          for {
-                           lastFinalizedBlock <- casper.lastFinalizedBlock
-                           runtimeManager     <- casper.getRuntimeManager
-                           postStateHash      = ProtoUtil.postStateHash(lastFinalizedBlock)
-                           res                <- runtimeManager.playExploratoryDeploy(term, postStateHash)
-                           lightBlockInfo     <- getLightBlockInfo[F](lastFinalizedBlock)
-                         } yield (res, lightBlockInfo).asRight[Error]
+                           targetBlock <- if (blockHash.isEmpty)
```
nzpr, line 841: @zsluedem How about making blockHash Option[String] with default None? Otherwise in the code to query state of LFB we have to do `exploratoryDeploy(term, "")` -> the purpose of "" is not really obvious. Better to have 
`exploratoryDeploy(term)` when querying LFB state and `exploratoryDeploy(term, hash)` for  some particular block

casper/src/test/scala/coop/rchain/casper/api/ExploratoryDeployAPITest.scala
```diff
@@ -36,7 +36,7 @@ class ExploratoryDeployAPITest
       log: Log[Task]
   ) =
     BlockAPI
-      .exploratoryDeploy(term)(
+      .exploratoryDeploy(term, none[String], true)(
```
nzpr: why put `, none[String], true` here? you can just omit, as they are defaults

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -686,6 +686,13 @@ object BlockAPI {
       case None             => none[BlockMessage]
     }
 
+  def getBlockFromStore[F[_]: Monad: BlockStore](
```
nzpr: `getBlockFromStore` and `findBlockFromStore` have same input and output args - why not have only `findBlockFromStore`? Is it performance optimisation?

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -826,12 +830,27 @@ object BlockAPI {
...
-                           runtimeManager     <- casper.getRuntimeManager
-                           postStateHash      = ProtoUtil.postStateHash(lastFinalizedBlock)
-                           res                <- runtimeManager.playExploratoryDeploy(term, postStateHash)
-                           lightBlockInfo     <- getLightBlockInfo[F](lastFinalizedBlock)
-                         } yield (res, lightBlockInfo).asRight[Error]
+                           targetBlock <- if (blockHash.isEmpty)
+                                           casper.lastFinalizedBlock.map(_.some)
+                                         else
+                                           BlockStore[F].get(
+                                             ByteString.copyFrom(Base16.unsafeDecode(blockHash.get))
```
tgrospic: This will thrown error if hash string is not valid. I know there is no elegant solution but please check if just `raiseError` with some meaningful message is enough to be reported correctly to the client.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -810,7 +812,9 @@ object BlockAPI {
   }
 
   def exploratoryDeploy[F[_]: Monad: EngineCell: Log: SafetyOracle: BlockStore](
-      term: String
+      term: String,
+      blockHash: Option[String] = none,
+      usePreStateHash: Boolean = false
```
tgrospic, line 825: Please add documentation comment for this method with description for parameters. Also this can be part of PR description which is necessary for release notes.

TODO: We should investigate what is easiest way to generate some API specification like Open API.

zsluedem: create an issue for that https://github.com/rchain/rchain/issues/2974

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -826,12 +838,35 @@ object BlockAPI {
...
+                                           casper.lastFinalizedBlock.map(_.some)
+                                         else
+                                           for {
+                                             hashByteString <- Base16
+                                                                .decode(blockHash.getOrElse(""))
+                                                                .map(ByteString.copyFrom)
+                                                                .liftTo[F](
+                                                                  BlockRetrievalError(
+                                                                    s"Input hash value is not valid hex string: $blockHash"
+                                                                  )
```
tgrospic, line 851: It's ok to throw this error here because it's handled here and converted to ApiErr.
https://github.com/rchain/rchain/blob/bf4cc6bf94542bfc919e0a0db6c00b28621a7a4e/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L596-L604

tgrospic (approved):
LGTM

zsluedem:
bors merge

PR #2975, opened 2020-06-28
Reorganize TransportLayer, CommUtil, BlockStore helpers

### LINKED ISSUE #2970: Reorganize TransportLayer, CommUtil, BlockStore helpers
## Overview
Idea is to have only basic operations on traits. Operations that can be defined by the base should be put in extension methods (syntax).
CommUtil can be refactored in this way and extensions can be defined on TransportLayer instance (+ config).

tgrospic:
## Overview
This PR is not changing any functionality it's just refactoring of CommUtil interface to reduce basic functions which are moved to TransportLayer extensions (syntax).

Also two methods to get block data unsafe (with applicative error) are transformed to extension methods.

Implicit classes for extension methods for KeyValueStore now extends AnyVal for slightly better optimization. 

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
@@ -5,13 +5,14 @@ import cats.effect.Sync
 import cats.implicits._
 import coop.rchain.blockstorage.BlockStore
 import coop.rchain.blockstorage.dag.BlockDagRepresentation
+import coop.rchain.casper.syntax._
 import coop.rchain.casper.util.{DagOperations, EventConverter, ProtoUtil}
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.BlockMetadata
-import coop.rchain.shared.Log
 import coop.rchain.rspace.Blake2b256Hash
 import coop.rchain.rspace.trace._
+import coop.rchain.shared.Log
 
 import scala.collection.BitSet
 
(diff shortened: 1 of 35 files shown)


casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -115,14 +116,13 @@ object EstimatorHelper {
...
       blockAncestorsMeta: List[BlockMetadata]
   ): F[BlockEvents] =
     for {
-      maybeAncestors <- blockAncestorsMeta.traverse(
-                         blockAncestorMeta => BlockStore[F].get(blockAncestorMeta.blockHash)
-                       )
-      ancestors = maybeAncestors.flatten
+      ancestors <- blockAncestorsMeta.traverse(
+                    blockAncestorMeta => BlockStore[F].getUnsafe(blockAncestorMeta.blockHash)
+                  )
```
tgrospic, line 125: This code can not throw errors which should be what we want and not to skip ancestors if not in the store.

nzpr: Why here you changed Monad to Sync?

tgrospic: Because of use of `Sync.raiseError` inside `BlockStore[F].getUnsafe`.

nzpr (approved):
LGTM
I rebased https://github.com/rchain/rchain/pull/2943 on this PR

zsluedem (approved):
This looks great to me. I got better understanding on syntax.
So we should define trait with the most generic method and then use syntax to combine these methods. Does that make sense?

tgrospic:
> This looks great to me. I got better understanding on syntax.
> So we should define trait with the most generic method and then use syntax to combine these methods. Does that make sense?

@zsluedem Yes. Generic methods on a trait can be seen as a definition of the core language (rules) and syntax (extensions) are like sentences written in that language.

tgrospic:
bors merge

nzpr:
bors merge

tgrospic:
bors retry

tgrospic:
bors retry

PR #2985, opened 2020-07-04
fix explore deploy backward not compatible

zsluedem:
## Overview
https://github.com/rchain/rchain/pull/2967 introduce new form api of explore deploy.
But the old version grpc api doesn't compatible with it.
Fix the incompatible of it.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
+++ b/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
@@ -237,7 +237,11 @@ object DeployGrpcServiceV1 {
       def exploratoryDeploy(request: ExploratoryDeployQuery): Task[ExploratoryDeployResponse] =
         defer(
           BlockAPI
-            .exploratoryDeploy[F](request.term, Some(request.blockHash), request.usePreStateHash)
+            .exploratoryDeploy[F](
+              request.term,
+              if (request.blockHash.isEmpty) none[String] else Some(request.blockHash),
+              request.usePreStateHash
+            )
         ) { r =>
           import ExploratoryDeployResponse.Message
           import ExploratoryDeployResponse.Message._
--- a/node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
+++ b/node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
@@ -131,7 +131,11 @@ object WebApiRoutes {
 
       case req @ POST -> Root / "explore-deploy-by-block-hash" =>
         req.handle[ExploreDeployRequest, ExploratoryDeployResponse](
-          req => webApi.exploratoryDeploy(req.term, Some(req.blockHash), req.usePreStateHash)
+          req =>
+            if (req.blockHash.isEmpty)
+              webApi.exploratoryDeploy(req.term, none[String], req.usePreStateHash)
+            else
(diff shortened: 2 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -839,12 +839,12 @@ object BlockAPI {
                            .pure[F]
                        case None =>
                          for {
-                           targetBlock <- if (blockHash.isEmpty)
+                           targetBlock <- if (blockHash.getOrElse("").isEmpty)
```
nzpr: ~~Isn't this always true? `.isEmpty` on empty string or on None~~
I mean why there is a need for `.getOrElse("")`? 
If its None, `isEmpty` on `None` is `true` as well

zsluedem: For backward compatible of grpc api.

zsluedem: @nzpr I remember that why I use an empty string in the very first place which you required to use option. It was because of the backward compatibility.

nzpr: I see, so the problem is related to this code 
https://github.com/rchain/rchain/blob/07ba590af2e0edc834aade0685a96c8ce1029561/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala#L237-L241
you pass "" as input when you need last finalised block?
Maybe better fix it there - to pass None if `blockHash` field is ""?

nzpr: and in WebAPI as well 
https://github.com/rchain/rchain/blob/07ba590af2e0edc834aade0685a96c8ce1029561/node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala#L132-L135

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: The underlying problem is that we are accepting `String` as typed value deep inside domain logic. For now this is OK because we don't have good strategy to validate inputs on protocol level. This means that we specify what is a Hash and how to validate it only once and just use this type as a definition for input type on API.

Similar situation is on reporting API. http4s has ability to define "matchers" as a way to validate expected type.
https://github.com/rchain/rchain/blob/8489c6ca0f2ff8c3df1ebb8a63ca717c1a6cb07a/node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala#L99-L109

Basically what we want is clean domain API with strict types. And the transformation API as a layer with conversions from/to primitive types like strings and bytes. Both layers can generate errors and we should/can treat these errors differently when returning to the user or adding to the log.
For example `WebApiRoutes.scala` from `WebApi` itself if we imagine some annotation for a method to specify "hints"  for the generator.
```scala
// Read data (listen)
@httpRoute(name = "api/data-at-name")
@grpcMethod(service = "deployService", package = "casper.v1")
def listenForDataAtName(request: DataRequest): F[DataResponse]
```

tgrospic (approved):
I've added some wishes for the future. :)

zsluedem:
bors merge

PR #2986, opened 2020-07-05
Shard config

### LINKED ISSUE #2983: Support on-chain configuration storage
### Impact
This will eliminate chance to be slashed due to node misconfiguration

### Dependencies
Requires 
1. Implementing rholang contracts serving as configuration storage
2. Changes to [RuntimeManager](https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala) . Most probably big refactor is needed
3. When some new block meets LastFinalisedState conditions -> read configuration from postStateHash of that block and update node state. _We cannot do this on each block adding as we will do this in parallel, and I'm not sure if it safe to say `last finalised block` instead of `LastFinalisedState`_

### LINKED ISSUE #3058: SafetyOracle does not work properly with bonding
## Overview
We ave two questionable thing with SafetyOracle
1) It uses main parent bonds map -> suggestion is to use bonds map of the block itself (per postStateHash of block). Might require investigating if we have access to that data/
2) It takes into account all records from the bonds map -> it has to use only active validators

### Impact
This method is a subject <https://github.com/rchain/rchain/blob/8fb7bc466ea9a7b516133da40b85da95d11595bb/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala#L114-L118>

We should change this as a part of shard config

UPD.
We need

1) Not add latest message for any validator until message is created. Bonding should not lead to LM creation. (TODO discuss)
2) Use bonds map of message itself and not main parent bonds map (TODO discuss)
3) Check if only active validators are used

nzpr:
## Overview
Issue https://github.com/rchain/rchain/issues/2983


casper/src/main/resources/ShardConf.rho
```diff
@@ -0,0 +1,49 @@
+/*
+ The table below describes the required computations and their dependencies
+
+ No. | Dependency | Computation method | Result
+ ----+------------+--------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------
+ 1.  |            | given              | sk = a9585a0687761139ab3587a4938fb5ab9fcba675c79fefba889859674046d4a5
```
nzpr, line 6: @tgrospic @zsluedem could you guide me on how to get al this `???` so I can deploy contract to sandbox and test? I know we are going to revisit logic of blessed contracts, but I would like to test now.

tgrospic: This data is generated with this tool. First unforgeable name is generated from public key + timestamp. Then as a tuple with with nonce it's signed.
https://rchain.atlassian.net/wiki/spaces/CORE/pages/660865046/Inserting+contracts+in+the+registry

I created example in JavaScript.
https://github.com/tgrospic/rnode-client-js/blob/964074d/src/nodejs/client-insert-signed.js

nzpr:
Will reopen when its ready, to not take GHA resources

PR #2993, opened 2020-07-13
add new api getDataAtChannel and getContinuationAtChannels

### LINKED ISSUE #2992: Implement api to get data from rspace
## Overview
Implement api to get data from rspace
We already got [listeningDataAtName and listeningContinuationAtName](https://github.com/rchain/rchain/blob/c76bd6668cf5f6ddd3acdc44f26fcb08edba08ee/models/src/main/protobuf/DeployServiceV1.proto#L42-L45) although I think they are not very good designed api which can be performance cost.
 We can implement something like [explore_deploy](https://github.com/rchain/rchain/blob/c76bd6668cf5f6ddd3acdc44f26fcb08edba08ee/models/src/main/protobuf/DeployServiceV1.proto#L57), but only for par.

### Impact
No impact for core.

### Dependencies
No

### Design
No

zsluedem:
## Overview
add new api getDataAtChannel and getContinuationAtChannels

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -157,6 +157,117 @@ object BlockAPI {
       )
     )
   }
+  def getDataAtChannel[F[_]: Concurrent: EngineCell: Log: SafetyOracle: BlockStore](
+      channel: Par,
+      blockHash: String,
+      usePreStateHash: Boolean
+  ): F[ApiErr[DataWithBlockInfo]] = {
+    val errorMessage =
+      "Could not execute get data at channel, casper instance was not available yet."
+    EngineCell[F].read >>= (
+      _.withCasper(
+        implicit casper =>
+          for {
+            isReadOnly <- casper.getValidator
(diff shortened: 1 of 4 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -157,6 +157,117 @@ object BlockAPI {
...
+
+                         } yield res.fold(
+                           s"Can not find block ${blockHash}".asLeft[ContinuationsWithBlockInfo]
+                         )(_.asRight[Error])
+                     }
+          } yield result,
+        Log[F].warn(errorMessage).as(s"Error: $errorMessage".asLeft)
+      )
+    )
+  }
```
tgrospic, line 270: It looks like these two functions are almost the same. Maybe `getDataWithBlockInfo` and `getContinuationsWithBlockInfo` can be abstracted as an input parameters to one common function which can be called from both functions.
This common function can `raiseError`s which can be handled and converted to `ApiErr[_]`.

zsluedem:
close this pr because I got some other ideas after runtime is refactored

PR #2998, opened 2020-07-18
Catch up optimisation tests

### LINKED ISSUE #2982: Rewrite legacy tests related to block retrieval
Rewrite tests disabled with this commit <https://github.com/rchain/rchain/pull/2943/commits/55b5d914516da14a666cf95d53067f822ef13161>

nzpr:
https://github.com/rchain/rchain/issues/2982


casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -163,7 +163,7 @@ object Running {
       if (r.doIgnore) {
         logIgnore(r.status)
       } else {
-        logSuccess >> processHash
+        logSuccess >> processHash.void
```
zsluedem, line 166: I never use void before. It make sure return `Unit`, right ?

casper/src/test/scala/coop/rchain/casper/sync/BlockRetrieverRequesAllSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+            val requested = RequestState(
+              timestamp = timedOut,
+              peers = Set(peerNode("peer")),
+              waitingList = waitingList
+            )
+            currentRequests.set(Map(hash -> requested)).runSyncUnsafe()
+            // when
+            blockRetriever.requestAll(timeout).runSyncUnsafe()
+            // then
+            val Some(requestedAfter) = currentRequests.get.runSyncUnsafe().get(hash)
```
zsluedem, line 124: Just want to say that you can wrap  these into a `for` and then you don't have to call `runSyncUnsafe()` every time.

zsluedem (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: Yes, instead of `>> ().pure[F]`

casper/src/main/scala/coop/rchain/casper/engine/BlockRetriever.scala
```diff
@@ -171,14 +171,28 @@ object BlockRetriever {
...
+                             else {
+                               (
+                                 addSourcePeerToRequest(state, hash, peer.get),
+                                 AdmitHashResult(
+                                   NewSourcePeerAddedToRequest,
+                                   broadcastRequest = false,
+                                   // if peer is the first one in waiting list - request block from that peer,
+                                   // otherwise requests should be triggered by casper loop
+                                   requestBlock =
+                                     if (state(hash).waitingList.isEmpty) true else false
```
tgrospic: This is the same as just `state(hash).waitingList.isEmpty`.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -289,79 +289,45 @@ object ProtoUtil {
...
+      shardId: String,
+      seqNum: Int = 0
   ): BlockMessage = {
-    val hash = hashUnsignedBlock(header, justifications)
-
     // TODO FIX-ME fields that can be empty SHOULD be optional
-    BlockMessage(
-      hash,
+    val block = BlockMessage(
+      ByteString.EMPTY,
```
tgrospic: Please use named assignment in these situations, it's more readable `hash = ByteString.EMPTY`.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -115,7 +79,7 @@ class BlockQueryResponseAPITest
         effects                                  <- effectsForSimpleCasperSetup(blockStore, blockDagStorage)
         spanEff                                  = NoopSpan[Task]()
         (logEff, engineCell, cliqueOracleEffect) = effects
-        hash                                     = secondBlockQuery
+        hash                                     = BlockHashOps(secondBlock.blockHash).base16String
```
tgrospic, line 82: `base16String` is an extension method but must be imported with `import coop.rchain.models.BlockHash._`.

This is already on the list to refactor to _syntax_ style #3001.

casper/src/test/scala/coop/rchain/casper/sync/BlockRetrieverRequesAllSpec.scala
tgrospic: I agree with @zsluedem. It's not super important in tests right now but using `runSyncUnsafe` implies that we definitely working with concrete instance (Task) which prevent us to write more general code with `F[_]` instead.

models/src/main/scala/coop/rchain/models/block/StateHash.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import com.google.protobuf.ByteString
+import coop.rchain.crypto.codec.Base16
+
+object StateHash {
+  type StateHash = ByteString
+
+  val Length = 32
+  implicit class StateHashOps(bs: StateHash) {
+    def base16String: String = Base16.encode(bs.toByteArray)
```
tgrospic, line 11: This extension is almost identical to [BlockHashOps](https://github.com/rchain/rchain/blob/003e2a8e/models/src/main/scala/coop/rchain/models/BlockHash.scala#L10).

We can have only one defined for `ByteString` which will work in both cases. It should be defined in _syntax_ style so it's easier to import.
I'm ok if this is done as part of refactoring case #3001.

tgrospic (approved):
Excellent work! Thanks for more unified generators!

Just a few nitpicking comments.

casper/src/test/scala/coop/rchain/casper/sync/BlockRetrieverRequesAllSpec.scala
nzpr: Ah, I see, that makes sense

nzpr:
bors merge

PR #3006, opened 2020-07-21
Use one gRPC channel per peer

### LINKED ISSUE #3005: Error with gRPC streaming: cancelled before receiving half close
gRPC server when node is under load might refuse to process streamed messages (blocks)
https://github.com/rchain/rchain/blob/2e870e1fb196fff4775b249b201af29346fdca54/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala#L77

with error
```
Could not receive stream! CANCELLED: cancelled before receiving half close
io.grpc.StatusRuntimeException: CANCELLED: cancelled before receiving half close
```
As per https://grpc.github.io/grpc/core/md_doc_core_transport_explainer.html
this means that server cancelled gRPC channel before client said that it has no more data to push.
It's not clear yet whether the problem is on server or on client side

Piece of code related as well

https://github.com/rchain/rchain/blob/7de591c89dc120abdbee7b93bfa791d003b4818b/comm/src/main/scala/coop/rchain/comm/transport/SslSessionServerInterceptor.scala#L34-L39

nzpr:
Attempt to address this issue https://github.com/rchain/rchain/issues/3005
ATM each p2p call opens new gRPC connection, which does TLS handshake, allocates TCP port and so on. 
This PR makes transport client open a single gRPC connection per peer and use it for all subsequent calls.

Also as per https://github.com/grpc/grpc/issues/7058 `channels are intended to be shared among stubs.`

For now it's not clear if we can/should go even further and use a single stub to execute all requests, that needs some research.
https://github.com/rchain/rchain/blob/ff17c3988c43ce7d038c30e20fb7da511634dd39/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala#L128-L131

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
@@ -3,11 +3,12 @@ package coop.rchain.comm.transport
 import java.io.ByteArrayInputStream
 import java.nio.file.Path
 
+import cats.Applicative
+import cats.effect.concurrent.{Deferred, Ref}
+
 import scala.concurrent.duration.{FiniteDuration, _}
 import scala.util._
-
 import cats.implicits._
-
 import coop.rchain.catscontrib.ski.kp
 import coop.rchain.comm._
 import coop.rchain.comm.protocol.routing._
@@ -16,7 +17,6 @@ import coop.rchain.grpc.implicits._
 import coop.rchain.metrics.Metrics
 import coop.rchain.shared.{Log, UncaughtExceptionLogger}
 import coop.rchain.shared.PathOps.PathDelete
-
 import io.grpc.ManagedChannel
 import io.grpc.netty._
 import io.netty.handler.ssl.SslContext
@@ -31,7 +31,8 @@ class GrpcTransportClient(
     maxMessageSize: Int,
     packetChunkSize: Int,
     tempFolder: Path,
-    clientQueueSize: Int
+    clientQueueSize: Int,
+    channelsMap: Ref[Task, Map[PeerNode, Deferred[Task, ManagedChannel]]]
 )(
     implicit scheduler: Scheduler,
     val log: Log[Task],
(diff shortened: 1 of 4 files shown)


nzpr:
@zsluedem @tgrospic I was able to successfully sync from scratch 250k blocks, no errors from this issue https://github.com/rchain/rchain/issues/3005

comm/src/main/scala/coop/rchain/comm/transport/TransportLayer.scala
```diff
@@ -15,6 +15,7 @@ trait TransportLayer[F[_]] {
   def broadcast(peers: Seq[PeerNode], msg: Protocol): F[Seq[CommErr[Unit]]]
   def stream(peer: PeerNode, blob: Blob): F[Unit]
   def stream(peers: Seq[PeerNode], blob: Blob): F[Unit]
+  def shutdown(): F[Unit]
```
tgrospic: This operation is not necessary on Transport interface it's more implementation detail. In NodeRuntime we have instance of concrete implementation which can be used directly.

But I don't think Netty channels needs special treatment to close it when process is closing. Is there a reason to do this, like preventing a thread to exit?

nzpr: @tgrospic I'm more thinking about releasing channels for peers. Transport is a set of channels for client calls, they are up and available while node is running. I'm not sure whats the procedure for releasing resources on the counterparty when node is just Ctrl+C-ed, so I put this to explicitly close all client connections. Do you think this is unnecessary?

zsluedem (approved):
LGTM

comm/src/main/scala/coop/rchain/comm/transport/TransportLayer.scala
nzpr: I see, Netty should close channels itself on shutdown, so this part of the code is not necessary. I'll remove commit https://github.com/rchain/rchain/pull/3006/commits/c12e23411b8899c374af6a77a8b0ea321b44af99

tgrospic (approved):
Nice work!

nzpr:
bors merge

PR #3053, opened 2020-08-02
Enable per peer outbound streaming queue

### LINKED ISSUE #3011: Make per peer transport client queue
All scarce resources that can be overflown (like message queue size) need to be allocated on per peer basis.

nzpr:
ATM all outbound block messages go through single queue with fixed length. If queue is overflown, messages are dropped. So different p2p connections can influence each other.  This PR creates introduces separate queues er each P2P connection.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
@@ -22,7 +22,7 @@ import io.grpc.netty._
 import io.netty.handler.ssl.SslContext
 import monix.eval.Task
 import monix.execution.Ack.Continue
-import monix.execution.{CancelableFuture, Scheduler}
+import monix.execution.{Cancelable, CancelableFuture, Scheduler}
 import monix.reactive.Observable
 
 /**
@@ -32,7 +32,8 @@ import monix.reactive.Observable
   */
 final case class BufferedGrpcStreamChannel(
     grpcTransport: ManagedChannel,
-    buffer: StreamObservable
+    buffer: StreamObservable,
+    buferSubscriber: Cancelable
 )
 
 class GrpcTransportClient(
@@ -74,6 +75,7 @@ class GrpcTransportClient(
 
   private def createChannel(peer: PeerNode): Task[BufferedGrpcStreamChannel] =
     for {
+      _                <- log.info(s"Creating new channel to peer ${peer.toAddress}")
       clientSslContext <- clientSslContextTask
       grpcChannel = NettyChannelBuilder
         .forAddress(peer.endpoint.host, peer.endpoint.tcpPort)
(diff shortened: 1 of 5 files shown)


comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -78,26 +72,34 @@ class GrpcTransportClient(
...
+
+      channel = BufferedGrpcStreamChannel(grpcChannel, buffer)
+
+      _ = buffer.subscribe(
+        sMsg =>
+          streamBlobFile(sMsg.path, peer, sMsg.sender)
+            .guarantee(sMsg.path.deleteSingleFile[Task])
+            .runToFuture >> Continue.pure[CancelableFuture]
+      )
+    } yield channel
```
tgrospic, line 99: Please add log message when new channel is created.

nzpr: @tgrospic here is it 
https://github.com/rchain/rchain/blob/8190cf547f03de903c8d6699e7842f23c7279151/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala#L111-L113

nzpr: I moved to `createChannel` method, thanks for hint.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -106,27 +110,25 @@ class GrpcTransportClient(
...
-            ) >>
+      _                  <- Applicative[Task].whenA(newChannel)(createChannel(peer) >>= cDef.complete)
+      c                  <- cDef.get
+      // In case underlying gRPC transport is terminated - clean resources,
+      // remove current record and try one more time
+      r <- if (c.grpcTransport.isTerminated)
+            Task.delay(c.buferSubscriber.cancel()) >>
+              log.info(
+                s"Channel to peer ${peer.toAddress} is terminated, removing from connections map"
+              ) >>
```
tgrospic, line 121: Kudos for adding canceling. I just have some concerns that can be theoretically called multiple times if multiple threads enter `if` block in the same time. Is the cancel safe or it will throw an error in that scenario.

nzpr: Cancellable sources says that `Guaranteed idempotency - calling it multiple times should have the same side-effect as calling it only once. Implementations of this method should also be thread-safe`.

The implementation of message buffer is standard Monix `PublishToOneSubject`, so it should be thread safe.
https://github.com/rchain/rchain/blob/ce06e7378531a601360d7fba52e281a61fdefe34/comm/src/main/scala/coop/rchain/comm/transport/buffer/LimitedBufferObservable.scala#L10-L20

tgrospic: Nice. Thanks!

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3055, opened 2020-08-06
Last Finalised State. MVP.

### LINKED ISSUE #3054: Node traverses the whole DAG on each replay
For some reason node traverses the whole DAG here on each replay https://github.com/rchain/rchain/blob/3e4ef77fdde277db4bb1397872283dec6f6af1ad/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala#L421-L442, it takes about 20-30 sec on testnet (300k blocks in a chain).

We should get rid of it because its 
1. Slow
2. LFS assumes that we don't need and don't have that old blocks, so need to investigate why at all we do this.

nzpr:
Continuation of work on LFS.

Now we do not make a distinction between Approved Block from genesis ceremony and LFS. Approved Block is a block which state some set of validators is agree on. With the very first block set is defined by `required-sigs`. With all other blocks it is defined by safety oracle via finalization.


nzpr:
With these changes I am able to connect to the network from LFS.

nzpr:
This commit <https://github.com/rchain/rchain/pull/3055/commits/200272ad90e105d92b6485896d748bb72bc6581f> should be reverted to not mess with DagStore code and another way to prepare blocks for insertion should be implemented.

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -38,121 +47,178 @@ class Initializing[F[_]
...
+                invalid = false
+              )
+              .unlessA(added)
+        missingDeps <- ProtoUtil.dependenciesHashesOf(block).filterA(dag.contains(_).not)
+        blocksToAdd <- missingDeps.traverse(BlockStore[F].get)
+        sortedDeps = blocksToAdd.flatten.sortWith(
+          _.body.state.blockNumber > _.body.state.blockNumber
+        )
+        _ <- sortedDeps.traverse(addBlockToDag(_, approvedBlock))
+      } yield ()
```
tgrospic, line 152: Please refactor this function using `tailRecM` to be stack safe.
https://typelevel.org/cats/typeclasses/monad.html#tailrecm

nzpr:
Imo we don't need this validation at all 
https://github.com/rchain/rchain/blob/70cf529ca48f41126c235d535bfcf70894a34062/casper/src/main/scala/coop/rchain/casper/Casper.scala#L103-L128
- Ceremony master created block itself
- Genesis validators should check their signature on Approved block after they signed Unapproved block
- Usual peers should just rely on consensus on the state.
So we just need postStateHash of approved block

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -38,121 +47,178 @@ class Initializing[F[_]
...
+                block,
+                approvedBlock.candidate.block,
+                invalid = false
+              )
+              .unlessA(added)
+        missingDeps <- ProtoUtil.dependenciesHashesOf(block).filterA(dag.contains(_).not)
+        blocksToAdd <- missingDeps.traverse(BlockStore[F].get)
+        sortedDeps = blocksToAdd.flatten.sortWith(
+          _.body.state.blockNumber > _.body.state.blockNumber
+        )
```
tgrospic: Because `Ordering[Long]` is defined there is more convenient function `sortBy` which only need selection of the field.
`blocksToAdd.flatten.sortBy(_.body.state.blockNumber)`

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -22,12 +24,19 @@ import coop.rchain.metrics.{Metrics, Span}
...
 import coop.rchain.shared._
+import fs2.concurrent.Queue
 
 import scala.concurrent.duration._
 
-/** Node in this state will query peers in the network with [[ApprovedBlockRequest]] message
-  * and will wait for the [[ApprovedBlock]] message to arrive. Until then  it will respond with
-  * `F[None]` to all other message types.
+object Initializing {
+  val pageSize = 3000
```
tgrospic: This is leftover from my previous changes, it's not used.

tgrospic:
> Imo we don't need this validation at all
> 
> * Ceremony master created block itself
> * Usual peers should just rely on consensus on the state.
>   So we just need postStateHash of approved block

Yes! This is correct. Even for the very first genesis, state can be downloaded in the same way as for any (last) finalized block.

> * Genesis validators should check their signature on Approved block after they signed Unapproved block

Genesis validators still need to create the block to validate the hash and add their signature.

nzpr:
> Genesis validators still need to create the block to validate the hash and add their signature.

Yes, but now genesis creates block (play+replay) - send to genesis validators - wait for approvals - and here again replay approved block. So we are removing the last step.
This also will make integration tests faster.

nzpr:
Just want to push the naming once more :) Sorry for being pesky. 
Let's call it `approved block` - I believe `last finalised block` is a bad word, because each node might finalise blocks a bit differently, so we have to find block that is finalised on a majority of nodes. That is `approved Block`.

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateBlockRequester.scala
```diff
@@ -112,10 +116,19 @@ object LastFinalizedStateBlockRequester {
...
 
+        allValidatorsSeen <- Stream.eval(validatorsRequired.modify[Boolean] { currSet =>
+                              val newSet = currSet - block.sender
+                              (newSet, newSet.isEmpty)
+                            })
+
+        inDeployLifeSpanRage = ProtoUtil.blockNumber(block) > ProtoUtil.blockNumber(
+          approvedBlock.candidate.block
+          // TODO extract this number from shard config
+        ) - 50
```
zsluedem, line 127: why 50? I remember we got a config for this. But I can not find it.

nzpr: Its still hardcoded n Casper implementation, we are going to make it consistent <https://github.com/rchain/rchain/pull/2986>

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -193,15 +192,16 @@ object CasperLaunch {
...
+    private def connectAndQueryApprovedBlock(trimState: Boolean): F[Unit] =
       for {
         validatorId <- ValidatorIdentity.fromPrivateKeyWithLogging[F](conf.validatorPrivateKey)
         _ <- Engine.transitionToInitializing(
               conf.shardName,
               conf.finalizationRate,
               validatorId,
-              CommUtil[F].requestApprovedBlock
-//              CommUtil[F].requestLastFinalizedBlock // To enable LFS
+              // TODO peer should be able to request approved blocks on different heghts
```
tgrospic: Typo in `heghts`.

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -38,121 +46,212 @@ class Initializing[F[_]
...
+              _ <- LastApprovedBlock[F].set(approvedBlock)
+              // Request last finalized block from bootstrap node
+              _ <- CommUtil[F].requestLastFinalizedBlock
+            } yield ()
+          } else
+            Log[F].info("Invalid ApprovedBlock received; refusing to add.")
 
     } yield ()
-  }
+  }*/
```
tgrospic: Do you have some plans for this code or it can be removed?

nzpr:
<https://github.com/rchain/rchain/pull/3055/checks?check_run_id=994259740#step:8:94> 
The only test is failing because dummy implementation of RSpace state manager is not complete. As we are creating tests in the follow up PR, we leave it for now.

tgrospic (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
nzpr: Yes, I think we should remove it

zsluedem (approved):
VEYR GOOD WORK! We will polish in more prs

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -144,58 +145,37 @@ class Initializing[F[_]
...
-                  (remainder, acc).asLeft[List[(BlockHash, Long)]].pure[F]
-          } yield r
+      def loopDependencies(params: RecParams): F[Either[RecParams, SortedBlocks]] = {
+        val (hashes, sorted) = params
+        hashes match {
+          case Nil => sorted.asRight[RecParams].pure[F]
+          case head +: tail =>
+            for {
+              block       <- BlockStore[F].getUnsafe(head)
+              deps        = ProtoUtil.dependenciesHashesOf(block)
```
nzpr, line 132: This will add to the tail blocks that are not in block store, eventually. Have to add some equivalent of `blockStored` from the previous version

tgrospic: All blocks should be in the blocks store downloaded in previous operation with `LastFinalizedStateBlockRequester`.

tgrospic: Oh, I see what you mean. There should be the check for last downloaded block where insertion should stop because we don't have all blocks to the first genesis.

tgrospic: Fix is here #3092.

PR #3067, opened 2020-08-14
Do not traverse the whole DAG on unseenBlock

### LINKED ISSUE #3054: Node traverses the whole DAG on each replay

nzpr:
On each replay node have to find invalid blocks that are in the scope of block being replayed. By `scope` I mean that there is a path in DAG between block replayed and invalid block. Also, node maintains only a single set of invalid blocks without any metadata. Therefore, before replaying new block node have to find out which invalid blocks are in the scope of that block. To do this node perform traversing on the DAG. 

ATM this traversing includes the whole DAG due to the bug. So takes the more time, the bigger DAG become. On testnet it is about 20-30 seconds. So we waste this time replaying each block.
This PR fixes it.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -362,43 +362,47 @@ object ProtoUtil {
     import cats.instances.stream._
 
     for {
-      latestMessages        <- dag.latestMessages
-      latestMessagesOfBlock <- toLatestMessage(block.justifications, dag)
-      unseenBlockHashesAndLatestMessages <- latestMessages.toStream
-                                             .traverse {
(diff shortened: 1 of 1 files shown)


tgrospic (approved):
LGTM

I've renamed some arguments and added a comment to explain to myself.

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3070, opened 2020-08-15
Add BlockIsInProcessing ignore reason

nzpr:
There is a flaw in message ignoring logic. When block is passed to Casper with all dependencies already filled, it omits casperBuffer, so node is not aware that this block is actually being processed. As block processing can last long, this leads to misleading requests of message that is already being processing.

This PR makes Casper maintaining set of hashes that are currently in processing + adjusts message ignoring logic.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -38,7 +38,8 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
     postGenesisStateHash: StateHash,
     shardId: String,
     finalizationRate: Int,
-    blockProcessingLock: Semaphore[F]
+    blockProcessingLock: Semaphore[F],
+    blocksInProcessing: Ref[F, Set[BlockHash]]
 )(
     implicit casperBuffer: CasperBufferStorage[F],
     metricsF: Metrics[F],
@@ -67,6 +68,8 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
   def getGenesis: F[BlockMessage] = approvedBlock.pure[F]
 
   def getValidator: F[Option[PublicKey]] = validatorId.map(_.publicKey).pure[F]
+
+  override def getBlocksInProcessing: F[Set[BlockHash]] = blocksInProcessing.get
 
   // Later this should replace addBlock, but for now one more method created, or there are too many tests to fix
   def addBlockFromStore(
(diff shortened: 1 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -196,6 +200,7 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
                     case false =>
                       ().pure[F]
                   }
+      _ <- blocksInProcessing.update(_ - b.blockHash)
```
tgrospic: If block is already in processing and if it's called the second time, it cannot acquire lock so it will be deleted from the processing set. This doesn't seems right.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -281,32 +283,47 @@ class Running[F[_]: Concurrent: BlockStore: CasperBufferStorage: BlockRetriever:
...
+                    .ifM(
+                      IgnoreCasperMessageStatus(doIgnore = true, BlockIsInCasperBuffer).pure[F],
+                      casper.blockDag.flatMap(
+                        _.contains(hash)
+                          .ifM(
+                            IgnoreCasperMessageStatus(doIgnore = true, BlockIsInDag).pure[F],
+                            if (received)
+                              IgnoreCasperMessageStatus(doIgnore = true, BlockIsReceived).pure[F]
+                            else
+                              IgnoreCasperMessageStatus(doIgnore = false, DoNotIgnore).pure[F]
```
tgrospic, line 325: This is little scary.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Yes you are right, thanks

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
nzpr: @tgrospic do these comments help? 
https://github.com/rchain/rchain/blob/f01e1b9bb798b4061195c9dd1f45cdfc806ae0ae/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L299-L330

tgrospic: Yes, thanks. I just wanted to say, it's not easy to follow all these if's in the head.

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3071, opened 2020-08-15
Fix timeout on HTTP server and HTTP propose response

### LINKED ISSUE #2979: Connection idle timeout is not respected for HTTP connections
HTTP connection should timeout after `max-connection-idle` time set in config file or as CLI argument.

### What did you see instead?
Configuration parameter `max-connection-idle` is not respected for HTTP connection. Whatever is set, connection will break after 30 seconds.

Push many deploys to a node, enough that proposing a block will last more than 30 seconds. After that call HTTP Admin API `/propose` to create a block. After 30 seconds HTTP connection will break and propose will stop.

### The problem with HTTP server configuration
If response header is not set `withIdleTimeout` parameter will not have effect. In this case `withResponseHeaderTimeout` will be be used and now it's default 30 seconds. 
https://github.com/rchain/rchain/blob/9da5b4ff6256f15e61128ca80da73d539f2b2d6e/node/src/main/scala/coop/rchain/node/web/package.scala#L63-L68

tgrospic:
## Overview
HTTP server has connection timeout option which only applies if response header is sent. This PR adds HTTP timeout to response header also.

Also added encoder in Admin API to correctly response JSON on _propose_ method.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -945,11 +945,9 @@ object NodeRuntime {
       adminWebApi = {
         implicit val ec     = engineCell
         implicit val sp     = span
-        implicit val or     = oracle
-        implicit val bs     = blockStore
         implicit val sc     = synchronyConstraintChecker
         implicit val lfhscc = lastFinalizedHeightConstraintChecker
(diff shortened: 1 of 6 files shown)


nzpr (approved):
LGTM

zsluedem (approved):
LGTM

tgrospic:
bors merge

PR #3073, opened 2020-08-16
Reorder ignore reason

nzpr:
This is add-on for <https://github.com/rchain/rchain/pull/3070>

`BlockIsWaitingForCasper` means that it is not missing any dependency, so not pushed to casper buffer, but cannot be added right now, so is waiting for casper to be able to process it. 
It is being detected with condition for BlockRetriever `received && !inCasperBuffer`
The issue is that this is also true for `BlockIsInProcessing`, because block being accepted straight away to casper for replay, not touching casper buffer.

This PR reorders check for ignore reason and put `BlockIsInProcessing` before `BlockIsWaitingForCasper`, to not shield the former one.

This must have some more elegant solution, but looks like it requires more cross dependency, e.g. BlockRetriever to be aware of Casper, which is better to avoid.


tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3079, opened 2020-08-18
Config for Last Finalized State

### LINKED ISSUE #3077: config for last finalized state
## Overview
add config for last finalized state

1. config to start a node from genesis or from last finalized state
2. config if a node can response with last finalized state request
3. make the node request the last finalized state only from bootstrap right now

zsluedem:
## Overview
add configs for last finalized state

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
@@ -40,7 +40,7 @@ object CasperLaunch {
     /* Casper */      : Estimator: SafetyOracle: LastFinalizedBlockCalculator: LastFinalizedHeightConstraintChecker: SynchronyConstraintChecker
     /* Storage */     : BlockStore: BlockDagStorage: LastFinalizedStorage: DeployStorage: CasperBufferStorage: RSpaceStateManager
     /* Diagnostics */ : Log: EventLog: Metrics: Span] // format: on
-  (conf: CasperConf): CasperLaunch[F] = new CasperLaunch[F] {
+  (conf: CasperConf, trimState: Boolean): CasperLaunch[F] = new CasperLaunch[F] {
     def launch(): F[Unit] =
       BlockStore[F].getApprovedBlock map {
         case Some(approvedBlock) =>
@@ -59,7 +59,7 @@ object CasperLaunch {
           (msg, action)
         case None =>
           val msg    = "Approved block not found, connecting to existing network"
-          val action = connectAndQueryApprovedBlock(true)
+          val action = connectAndQueryApprovedBlock(trimState)
           (msg, action)
       } >>= {
         case (msg, action) => Log[F].info(msg) >> action
--- a/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
@@ -40,170 +40,188 @@ object CasperLaunch {
     /* Casper */      : Estimator: SafetyOracle: LastFinalizedBlockCalculator: LastFinalizedHeightConstraintChecker: SynchronyConstraintChecker
     /* Storage */     : BlockStore: BlockDagStorage: LastFinalizedStorage: DeployStorage: CasperBufferStorage: RSpaceStateManager
     /* Diagnostics */ : Log: EventLog: Metrics: Span] // format: on
-  (conf: CasperConf, trimState: Boolean): CasperLaunch[F] = new CasperLaunch[F] {
-    def launch(): F[Unit] =
-      BlockStore[F].getApprovedBlock map {
-        case Some(approvedBlock) =>
-          val msg    = "Approved block found, reconnecting to existing network"
-          val action = connectToExistingNetwork(approvedBlock)
-          (msg, action)
-        case None if (conf.genesisCeremony.genesisValidatorMode) =>
-          val msg =
-            "Approved block not found, taking part in ceremony as genesis validator"
-          val action = connectAsGenesisValidator()
-          (msg, action)
-        case None if (conf.genesisCeremony.ceremonyMasterMode) =>
-          val msg =
-            "Approved block not found, taking part in ceremony as ceremony master"
-          val action = initBootstrap()
-          (msg, action)
-        case None =>
-          val msg    = "Approved block not found, connecting to existing network"
-          val action = connectAndQueryApprovedBlock(trimState)
-          (msg, action)
-      } >>= {
-        case (msg, action) => Log[F].info(msg) >> action
+  (
+      conf: CasperConf,
+      trimState: Boolean,
+      enableStateExporter: Boolean
+  ): CasperLaunch[F] =
+    new CasperLaunch[F] {
+      def launch(): F[Unit] =
+        BlockStore[F].getApprovedBlock map {
+          case Some(approvedBlock) =>
+            val msg = "Approved block found, reconnecting to existing network"
+            val action =
+              connectToExistingNetwork(approvedBlock, enableStateExporter)
+            (msg, action)
+          case None if (conf.genesisCeremony.genesisValidatorMode) =>
+            val msg =
+              "Approved block not found, taking part in ceremony as genesis validator"
+            val action = connectAsGenesisValidator()
+            (msg, action)
+          case None if (conf.genesisCeremony.ceremonyMasterMode) =>
+            val msg =
+              "Approved block not found, taking part in ceremony as ceremony master"
+            val action = initBootstrap(enableStateExporter)
+            (msg, action)
+          case None =>
+            val msg = "Approved block not found, connecting to existing network"
+            val action = connectAndQueryApprovedBlock(
+              trimState,
+              enableStateExporter
+            )
+            (msg, action)
+        } >>= {
+          case (msg, action) => Log[F].info(msg) >> action
+        }
+
+      private def connectToExistingNetwork(
+          approvedBlock: ApprovedBlock,
+          enableStateExporter: Boolean
+      ): F[Unit] = {
+        def askPeersForForkChoiceTips = CommUtil[F].sendForkChoiceTipRequest
+        def sendBufferPendantsToCasper(casper: Casper[F]) =
+          for {
+            pendants <- CasperBufferStorage[F].getPendants
+            // pendantsReceived are either
+            // 1. blocks that were received while catching up but not end up in casper buffer, e.g. node were restarted
+            // or
+            // 2. blocks which dependencies are in DAG, so they can be added to DAG
+            // In both scenarios the way to proceed is to send them to Casper
+            pendantsStored <- pendants.toList.filterA(BlockStore[F].contains)
+            _ <- Log[F].info(
+                  s"Checking pendant hashes: ${pendantsStored.size} items in CasperBuffer."
+                )
+            _ <- pendantsStored
+                // we just need to send blocks to Casper. Nothing to do with results of block processing here,
+                // so ignoring them
+                  .traverse_(
+                    hash =>
+                      for {
+                        block <- BlockStore[F].get(hash).map(_.get)
+                        _ <- Log[F].info(
+                              s"Pendant ${PrettyPrinter.buildString(block, short = true)} " +
+                                s"is available in BlockStore, sending to Casper."
+                            )
+                        dc <- casper.dagContains(hash)
+                        _ <- Log[F]
+                              .error(
+                                s"Pendant ${PrettyPrinter.buildString(block, short = true)} " +
+                                  s"is available in DAG, database is supposedly in inconsistent state."
+                              )
+                              .whenA(dc)
+                        _ <- BlockRetriever[F].ackReceive(hash)
+                        _ <- casper.addBlockFromStore(hash, allowAddFromBuffer = true)
+                      } yield ()
+                  )
+          } yield ()
+
+        for {
+          validatorId <- ValidatorIdentity.fromPrivateKeyWithLogging[F](conf.validatorPrivateKey)
+          ab          = approvedBlock.candidate.block
+          casper <- MultiParentCasper
+                     .hashSetCasper[F](
+                       validatorId,
+                       ab,
+                       conf.shardName,
+                       conf.finalizationRate
+                     )
+          init = for {
+            _ <- askPeersForForkChoiceTips
+            _ <- sendBufferPendantsToCasper(casper)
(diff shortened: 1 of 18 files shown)


node/src/main/resources/defaults.conf
```diff
@@ -317,6 +317,12 @@ casper {
     # Allow node being ceremony master
     # If this is `true` and no genesis block available on startup node will start genesis ceremony.
     ceremony-master-mode = ${standalone}
+
+    # Allow node starts from genesis block or Approved block
+    # If this is `true`,when a node without any data connecting to exist network, the node would starts from
+    # genesis block and validate blocks from scratch. Otherwise, the node would start from approved block which
+    # can save a lot of time of validating all the history blocks
+    from-last-finalized-state = false
```
nzpr: @zsluedem I believe better to have `trimState` option as I named it first place, and have it true by default.
Most of the time we need just connect to the network not thinking about what os needed for that.
And only in rare cases we need to go back to the very first block (which actually can be not genesis on the particular node)

zsluedem: Yeah. I agree with you. But I am not sure if last-finalized-state is mature enough. I want to make this default to be `true`  when we think it is really ready.

tgrospic: We are working on feature branch for LFS so we can assume it will be ready when merging to dev.

zsluedem: ok.

casper/src/main/scala/coop/rchain/casper/engine/GenesisCeremonyMaster.scala
```diff
@@ -52,14 +52,24 @@ object GenesisCeremonyMaster {
     /* Casper */      : Estimator: SafetyOracle: LastFinalizedBlockCalculator: LastFinalizedHeightConstraintChecker: SynchronyConstraintChecker
     /* Storage */     : BlockStore: BlockDagStorage: LastFinalizedStorage: DeployStorage: CasperBufferStorage: RSpaceStateManager
     /* Diagnostics */ : Log: EventLog: Metrics: Span] // format: on
-  (shardId: String, finalizationRate: Int, validatorId: Option[ValidatorIdentity]): F[Unit] =
+  (
+      shardId: String,
+      finalizationRate: Int,
+      validatorId: Option[ValidatorIdentity],
+      allowRespondLastFinalizedStateRequest: Boolean
```
nzpr: I strongly suggest not use `last finalized state` anywhere in the code. It's `approved state`.

Also, why there is a need for this param in Ceremony Master? It should always answer for request for state transfer

node/src/main/resources/defaults.conf
```diff
@@ -317,6 +317,12 @@ casper {
     # Allow node being ceremony master
     # If this is `true` and no genesis block available on startup node will start genesis ceremony.
     ceremony-master-mode = ${standalone}
+
+    # Allow node starts from genesis block or Approved block
+    # If this is `true`,when a node without any data connecting to exist network, the node would starts from
+    # genesis block and validate blocks from scratch. Otherwise, the node would start from approved block which
+    # can save a lot of time of validating all the history blocks
+    from-last-finalized-state = true
```
nzpr: This should not be in `genesis-ceremony` section, as its not related to GC. I suggest to put it next to `bootstrap`
And again - lets not use `last-finalized-state`.

 I suggest 

```
bootstrap = "botstrap_addr"
trim-state = true
```

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -190,6 +190,9 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
         "in `<genesis-path>/<public_key>.sk`" +
         "This param specifies number of validator identites to generate."
     )
+    val fromLastFinalizedState = opt[Flag](
+      descr = "node would start last finalized state"
```
nzpr: Lets start all descriptions with uppercase as this is user facing text

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
```diff
@@ -40,170 +40,184 @@ object CasperLaunch {
...
+            (msg, action)
+          case None if (conf.genesisCeremony.genesisValidatorMode) =>
+            val msg =
+              "Approved block not found, taking part in ceremony as genesis validator"
+            val action = connectAsGenesisValidator()
+            (msg, action)
+          case None if (conf.genesisCeremony.ceremonyMasterMode) =>
+            val msg =
+              "Approved block not found, taking part in ceremony as ceremony master"
+            val action = initBootstrap(allowRespondLastFinalizedStateRequest)
```
nzpr: I've mentioned this in another comment, but IMO CM have to respond to state transfer requests

nzpr:
I don't get why there is a need for such a big change for https://github.com/rchain/rchain/pull/3079/commits/5499e371056e1b59d4c92716dd2c4f2420cff6e6?
You only need to block requests for this handler 

https://github.com/rchain/rchain/blob/f71068f0ea0a50d2e913c2240be472cd6b7cfb20/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L458-L460

So you only need to pass config param into Running engine instance.

Also I suggest to name param `enableStateExporter`, as what it does is allowing the node to give out the state.

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -29,6 +29,8 @@ trait CommUtil[F[_]] {
   // Broadcast packet in chunks (stream)
   def streamToPeers(packet: Packet): F[Unit]
 
+  def sendToPeer(message: Packet, peer: PeerNode): F[Unit]
```
nzpr: https://github.com/rchain/rchain/blob/8a69a559f1d4c96761251a81c974b74883e2ccd9/comm/src/main/scala/coop/rchain/comm/transport/TransportLayerSyntax.scala#L22-L27
This is already implemented as TransportLayer syntax

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
zsluedem: What is CM?

nzpr: @zsluedem Ceremony Master

zsluedem: Oh nvm

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -482,8 +483,11 @@ class Running[F[_]
 
     // Approved state store records
     case StoreItemsMessageRequest(startPath, skip, take) =>
-      handleStateItemsMessageRequest(peer, startPath, skip, take)
-
+      if (allowRespondApprovedStateRequest) {
+        handleStateItemsMessageRequest(peer, startPath, skip, take)
+      } else {
+        Sync[F].delay(())
```
tgrospic: When returning a _value_ it can just be wrapped with _pure_ `().pure[F]`. This doesn't change the behavior like with calling a function.

casper/src/test/scala/coop/rchain/casper/engine/InitializingSpec.scala
```diff
@@ -40,7 +40,9 @@ class InitializingSpec extends WordSpec {
           Some(validatorId),
           theInit,
           blockResponseQueue,
-          stateResponseQueue
+          stateResponseQueue,
+          false,
+          true
```
tgrospic: Please use named arguments for literal values.

node/src/main/resources/defaults.conf
```diff
@@ -50,6 +50,10 @@ protocol-server {
   # make this number high. Upper bounded by the numer of messages (blocks are the biggest of them)
   # node is able to keep in memory. Block size upper bound size is `max-stream-message-size`.
   max-message-consumers = 400
+
+  # whether a node should respond the request of last finalized state(StateItermRequest).
+  # The forbidden request can remove the network impact for the validator
+  allow-respond-last-finalized-state-request = true
```
tgrospic: This can be just `allow-respond-state-request` because we can later implement feature to receive any state not just last finalized.

casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala
zsluedem: This is just argument pass to running state

casper/src/main/scala/coop/rchain/casper/util/comm/CommUtil.scala
```diff
@@ -150,6 +150,15 @@ final class CommUtilOps[F[_]](
...
     sendToPeers(StoreItemsMessageRequest.toProto(req))
 
+  def sendStoreItemsRequestToBootstrap(
+      req: StoreItemsMessageRequest
+  )(implicit m: Sync[F], r: RPConfAsk[F], t: TransportLayer[F]): F[Unit] =
+    for {
+      maybeBootstrap <- RPConfAsk[F].reader(_.bootstrap)
+      bootstrap      <- maybeBootstrap.liftTo(StandaloneNodeSendToBootstrapError)
+      _              <- TransportLayer[F].sendToPeer(bootstrap, ToPacket(StoreItemsMessageRequest.toProto(req)))
+    } yield ()
```
tgrospic: `CommUtil` is not used here so it makes more sense to add this method to [TransportLayerSyntax](https://github.com/rchain/rchain/blob/8a69a559f1/comm/src/main/scala/coop/rchain/comm/transport/TransportLayerSyntax.scala#L26) as extension to `TransportLayer`.
The exception defined here can be moved also, until we create better way of classifying errors.

casper/src/test/scala/coop/rchain/casper/engine/GenesisCeremonyMasterSpec.scala
```diff
@@ -55,7 +55,8 @@ class GenesisCeremonyMasterSpec extends WordSpec {
           .waitingForApprovedBlockLoop[Task](
             shardId,
             finalizationRate,
-            Some(validatorId)
+            Some(validatorId),
+            true
```
tgrospic: Please use named arguments for literal values.

node/src/main/scala/coop/rchain/node/configuration/model.scala
```diff
@@ -44,12 +44,14 @@ final case class ProtocolServer(
...
     grpcMaxRecvStreamMessageSize: Long,
-    maxMessageConsumers: Int
+    maxMessageConsumers: Int,
+    enableStateExporter: Boolean
 )
 
 final case class ProtocolClient(
     networkId: String,
     bootstrap: PeerNode,
+    trimState: Boolean,
```
tgrospic, line 54: I'm not sure if these new options are directly _protocol_ related. They are more general, maybe on the RSpace level but even more, as the node parameters (*) to determine how state will be shaped and accessed. Transport layer or RSpace will not change its behavior depending on these options.

\* Because we have Engine, where most components are connected, inside casper project, maybe it's easiest solution is to be part of casper options.

tgrospic (commented):
Good work! I left a few comments.

For tests I created a new issue #3097.

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: @tgrospic I tend not to agree here - this config and CLI options are user facing features. So looking from node operator standpoint, `trimState` is a parameter that accompanying `bootstrap`, and saying that node operator does not want to download the whole state (or he does in case of `false`).  We should make configuration as simple as possible without assuming node operator has some implicit knowledge about the platform. We have protocol server and API server in node software, everything that talks with protocol server of another node - is protocol client related here, IMO.

node/src/main/resources/defaults.conf
nzpr: @tgrospic How about `enableStateExporter`? https://github.com/rchain/rchain/pull/3079#issuecomment-677426337

tgrospic: I'm fine with that.

node/src/main/scala/coop/rchain/node/configuration/model.scala
tgrospic: Good point. This section is more _client_ then _protocol_ so it makes sense. Maybe `requestTrimState` would be appropriate?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
zsluedem: I think a debug log is better than just nothing

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -490,8 +491,11 @@ class Running[F[_]
 
     // Approved state store records
     case StoreItemsMessageRequest(startPath, skip, take) =>
-      handleStateItemsMessageRequest(peer, startPath, skip, take)
-
+      if (enableStateExporter) {
+        handleStateItemsMessageRequest(peer, startPath, skip, take)
+      } else {
+        Log[F].debug("The node is configured to not respond to StoreItemsMessage.")
```
tgrospic: Maybe peer info would be useful so that we know who is sending requests.

node/src/main/scala/coop/rchain/node/configuration/commandline/ConfigMapper.scala
```diff
@@ -91,6 +95,7 @@ object ConfigMapper {
       add("casper.genesis-ceremony.approve-interval", run.approveInterval)
       add("casper.genesis-ceremony.approve-duration", run.approveDuration) //TODO remove
       add("casper.genesis-ceremony.autogen-shard-size", run.autogenShardSize)
+      add("casper.genesis-ceremony.trim-state", run.trimState)
```
tgrospic: This should be in `protocol-client` section.

tgrospic (approved):
LGTM

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: `requestTrimState ` sounds good to me, but imo no too different from `trimState`. I don't se something else that can be called `trimState` in this context

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #3090, opened 2020-08-19
Fix Initializing engine test

tgrospic:
## Overview
Fix for Initializing Engine to work with RSpace state manager which contains import and export features used for Last Finalized State. 

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
@@ -62,7 +62,7 @@ class Initializing[F[_]
   override def init: F[Unit] = theInit
 
   override def handle(peer: PeerNode, msg: CasperMessage): F[Unit] = msg match {
-    case ab: ApprovedBlock        => onApprovedBlock(peer, ab.candidate.block)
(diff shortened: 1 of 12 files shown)


nzpr:
This push is indentation fix so it compile

nzpr:
The latest commit actually reverts some previous changes from LFS, which was was wrong. ApprovedBlock handler were reconstructing approved block from extracted block and pushing empty list of signatures. I just made it back using the whole received approved block.

So he only difference that we have between genesis and future approved blocks - is how they are constructed. Genesis is created with genesis ceremony, so has signatures gathered during the ceremony. And usual approved blocks do not have this feature yet. So we will need some kind of signature exchange on ongoing basis to justify some block as an approved one (or this can be a task for porter - to find out which block is approved and get sigs). Now we just trust the bootstrap.

tgrospic:
Maybe we don't need direct signatures for the received LFB (approved block). Just by downloading all dependencies for consensus and replaying new blocks can be enough. When node calculates new finalized block on top of the received approved block, will make it validated.

nzpr:
Makes sense, having blocks signed by validators on top of approved block received will definitely justify it. So node should detect that someone gave out bad approved block pretty soon - when it cannot finalise it itself.

nzpr (approved):

PR #3092, opened 2020-08-21
Approved state. Bug fixes.

nzpr:
1. When approved state has not been received in full, on restart node crashes. Because it assumes that it already has approved state.
2. Sorting of blocks before inserting to DAG has a bug.

PR fixes this 2 issues.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
@@ -88,8 +88,12 @@ class Initializing[F[_]
       _ <- if (isValid) {
             for {
               _ <- Log[F].info(
-                    s"Valid approved block received ${PrettyPrinter.buildString(approvedBlock)}"
+                    s"Valid approved block ${PrettyPrinter
+                      .buildString(approvedBlock.candidate.block, short = true)} received. Restoring approved state."
                   )
+
+              // Download approved state and all related blocks
+              _ <- requestApprovedState(approvedBlock)
 
               _ <- EventLog[F].publish(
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -167,19 +131,56 @@ class Initializing[F[_]
                            tupleSpaceQueue
                          )
 
+      _ <- BlockStore[F].put(approvedBlock.candidate.block)
```
tgrospic, line 136: Why to move this here? It makes sense to be where approved block is saved in `onApprovedBlock` handler.

nzpr: @tgrospic if we save approved bock before the whole state is complete - on restart node assumes that it has approved block, and this causes crash.

nzpr: this block (wrapped into approved block) is the same from standpoint of the whole approved state - just another block. But we start to save blocks to blockstore inside fs2 stream, starting with dependencies. So I moved saving of the actual starting block as close to fs2 stream as possible, as we still need it to be in blockstore.

tgrospic: Makes sense. I've added a comment to clarify why is this at the end of received state and I updated requester to check if block is already in the store so now node will continue on the same block is interrupted in the process of receiving the blocks. For the state I'll try to make something similar.

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

Issue #3093, opened 2020-08-22
Maintain cache of on-chain data.


nzpr:
## Overview
To be able to execute various methods in scala code we need to have on-chain data. E.g. to get fork-choice tips, we need list of active validators, other place when on-chain data is extracted is block creation and block replay.
This data is related to particular state (e.g. we have number of blocks that we merge -> this result in some state). So, to have quick access to on-chain data, after it is accessed once, it should be cached for some time (LRU cache looks the most reasonable). 

### Impact
After this is complete, we will be able to get fork-choice tips for the latest finalized block efficiently, as it requires active validators. Also this is part of work on shard configuration.

PR #3095, opened 2020-08-22
Throw away latest messages that are more then 1000 blocks behind

### LINKED ISSUE #3094: Fork-choice tip calculation grows linearly
After epoch change we have latest message from non active validators participating in calculation of FCT. This leads to increasing of FCT calculation time, because DAG height on which this calculation is being performed grows as network proceeds.

This code slows down 
https://github.com/rchain/rchain/blob/27787e1f47392e1998135db2316c2c32651abf4d/casper/src/main/scala/coop/rchain/casper/Estimator.scala#L51-L72
Screenshot 2020-08-22 at 13 55 27

On testnet this code takes milliseconds.

nzpr:
After epoch change we have latest message from unbonded validators, and right fix for that should be https://github.com/rchain/rchain/issues/3093. Until this is implemented, this is a quick fix - to ignore latest messages that are more then 1000 blocks behind.

Closes https://github.com/rchain/rchain/issues/3094

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -55,7 +55,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
     def latestMessageHash(validator: Validator): F[Option[BlockHash]] =
       latestMessagesMap.get(validator).pure[F]
 
-    def latestMessageHashes: F[Map[Validator, BlockHash]] = latestMessagesMap.pure[F]
+    def latestMessageHashes: F[Map[Validator, BlockHash]] =
+      latestMessagesMap
+        .filter(
+          // This filter is a quick fix until https://github.com/rchain/rchain/issues/3093 is implemented,
+          // as we need to filter latest messages with only active validators. Adn this info is available only
+          // on chain.
+          message => {
+            val messageHeight = heightMap.find(_._2.contains(message._2)) match {
+              // Invalid blocks are not stored in height map
+              case None => invalidBlocksSet.find(_.blockHash == message._2).get.blockNum
+              case m    => m.get._1
+            }
+            messageHeight > (getMaxHeight - 1000)
+          }
+        )
+        .pure[F]
 
     def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
 


block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -55,7 +55,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
 
-    def latestMessageHashes: F[Map[Validator, BlockHash]] = latestMessagesMap.pure[F]
+    def latestMessageHashes: F[Map[Validator, BlockHash]] =
+      latestMessagesMap
+        .filter(
+          // This filter is a quick fix until https://github.com/rchain/rchain/issues/3093 is implemented,
+          // as we need to filter latest messages with only active validators. Adn this info is available only
+          // on chain.
+          message => {
+            val messageHeight = heightMap.find(_._2.contains(message._2)) match {
```
tgrospic, line 65: Using _heightMap_ will iterate a lot of records to find some old latest message. Maybe it's faster to use _blockMetadataIndex_ which will be mapped in memory with LMDB and there is no need for iteration. Also invalid blocks will be included.

Now that we don't have just a _value_ but some more work is executed, the function is not referentially transparent. With _pure_ value must be created eagerly, but if wrapped in `Sync[F].delay` its execution will be suspended until `F` is executed by the scheduler.
Similar can be said for other methods like _contains_ or _lookup_. I'm not sure what to consider as values so that everything is not wrapped in Sync suspend.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -55,7 +55,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+              // Invalid blocks are not stored in height map
+              case None => invalidBlocksSet.find(_.blockHash == message._2).get.blockNum
+              case m    => m.get._1
+            }
+            messageHeight > (getMaxHeight - 1000)
```
zsluedem, line 70: I am fine with `Throw away latest messages that are more than X blocks behind` but the value needs to be considered well.

I think the value should be grown if we get more validators. Just do not hard-coded value in the middle of the codes. 
And I am fine with `1000` currently now.


BTW, We also have this `50` hard-coded thing which should be considered well.

tgrospic: This is just a fix until we have new PoS with shard config so it should be only temporary. Maybe just creating a variable is enough. We don't expect any change of validators until next epoch.

I agree for deploy threshold of 50 blocks, it is part of shard config.
https://github.com/rchain/rchain/pull/2986/files#diff-0e24b8a374b991ceffedf7d5af4aa696R35

My concern is for slashed validators, if they should be detected and their latest blocks should not be filtered?

nzpr:
I think we should not proceed with this fix, as it will influence justifications, I'll create another PR

PR #3100, opened 2020-08-26
Temp fix. Add constraints on traversing DAG when calculating LCA

### LINKED ISSUE #3094: Fork-choice tip calculation grows linearly

nzpr:
This is a replacement for <https://github.com/rchain/rchain/pull/3095>. It does not touch justifications. It changes ~~how LCA is calculated and latest messages are ranked, but only for very old blocks (~1000 blocks behind top of the DAG).~~. I t turned out that its enough to just not use very old blocks for LCA calculation. It makes ranking fast as well, because of specifics of ranker implementation.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Estimator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Estimator.scala
@@ -99,11 +99,15 @@ final class Estimator[F[_]: Sync: Log: Metrics: Span](
       latestMessages <- latestMessagesHashes.values.toStream
                          .traverse(hash => blockDag.lookup(hash))
                          .map(_.flatten)
-      result <- if (latestMessages.isEmpty) {
+      topBlockNumber <- blockDag.latestBlockNumber
+      filteredLM = latestMessages.filter(
+        _.blockNum > topBlockNumber - Estimator.latestMessageMaxDepth
+      )
+      result <- if (filteredLM.isEmpty) {
                  genesis.blockHash.pure[F]
                } else {
-                 latestMessages
-                   .foldM(latestMessages.head) {
+                 filteredLM
+                   .foldM(filteredLM.head) {
                      case (acc, latestMessage) =>
                        // TODO: Change to mainParentLCA
                        DagOperations.lowestUniversalCommonAncestorF[F](
@@ -211,6 +215,12 @@ final class Estimator[F[_]: Sync: Log: Metrics: Span](
 
 object Estimator {
   val UnlimitedParents = Int.MaxValue
+
+  // Latest messages with block number more then this value before the tip of the DAG
+  // won't be taken into account when calculating LCA
+  // TODO remove when proper fix for https://github.com/rchain/rchain/issues/3094 is implemented
+  //  Have to be removed before block merge
+  val latestMessageMaxDepth = 1000L
 
   def apply[F[_]](implicit ev: Estimator[F]): Estimator[F] =
     ev


casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -151,13 +151,31 @@ final class Estimator[F[_]: Sync: Log: Metrics: Span](
...
+    }*/
+
+    // Initial code above that commented out is slow when there is old latest message.
+    // The below solution is a temp solution until we have efficient way to calculate LCA.
+    latestMessagesHashes.toList.foldLeftM(Map.empty[BlockHash, Long]) {
+      case (acc, (validator: Validator, latestBlockHash: BlockHash)) =>
+        for {
+          topBlockNumber           <- blockDag.latestBlockNumber
+          latestMessageBlockNumber <- blockDag.lookup(latestBlockHash).map(_.get.blockNum)
+          r <- if (topBlockNumber - latestMessageBlockNumber > 1000)
```
zsluedem: I like this fix better than the before one~! Good work. 
However, I am still picky on this hard-coded `1000`. :)
I saw this `1000` is used in multiple place. We'd better define a `val` currently and use that value in one place for now.

casper/src/main/scala/coop/rchain/casper/util/DagOperations.scala
```diff
@@ -109,7 +110,12 @@ object DagOperations {
 
     if (b1 == b2) {
       b1.pure[F]
-    } else {
+    } else
+    // This constraints are a temp solution, we need to make LCA calculation fast. For now it is assumed
+    // that when one block is `staleLatestMessageThreshold` behind another - it is LCA for this pair.
+    if (b1.blockNum - b2.blockNum > Estimator.depth) b2.pure[F]
+    else if (b2.blockNum - b1.blockNum > Estimator.depth) b1.pure[F]
```
tgrospic: So this is just an optimization. It will not change the result because this very old block will be anyway returned as LCA. Is that right?

nzpr: @tgrospic yes, but in general case you never know what should be `Estimator.depth`. If we have some slashing conditions like `you have to propose at least each N blocks`, this condition can be justified.

nzpr: But even then you cannot be sure how temporary forking influence this. Looks like you don't have much options except traversing the DAG till you find one LCA. Thats why I'm thinking - might be we can keep a records for joint LCAs for different blocks and somehow incrementally update. But thats just a raw guess.

9rb: https://uhra.herts.ac.uk/bitstream/handle/2299/12152/906535.pdf
https://java2blog.com/lca-k-ary-tree-osqrtheight/

nzpr:
I changed the hotfix logic. Taking into account how the Estimator code is written - this is enough. We only ned to filter very old messages when calculating LCA. Why it works for other parts of the code - is not very clear from the first glance and hidden by the actual implementation. So anyway, we need to rewrite estimator at some point (I think when doing block merge).

casper/src/main/scala/coop/rchain/casper/Estimator.scala
```diff
@@ -99,16 +99,20 @@ final class Estimator[F[_]: Sync: Log: Metrics: Span](
...
+      )
+      result <- if (filteredLM.isEmpty) {
                  genesis.blockHash.pure[F]
                } else {
-                 latestMessages
-                   .foldM(latestMessages.head) {
-                     case (acc, latestMessage) =>
+                 filteredLM
+                   .foldM(filteredLM.head) {
+                     case (acc, filteredLM) =>
```
tgrospic: Now one item has the same name as the collection. Existing name seems reasonable.

nzpr: Thanks, indeed

tgrospic (approved):
I'm OK with this optimization until we have block merging.

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3117, opened 2020-09-08
use PROXY_SAFE for lmdb 

### LINKED ISSUE #3116: debug infor for history store
## Overview
The node would encounter bug like below.
```
java.lang.Exception: 0: cannot acquire 256 bits from a vector that contains 252 bits at coop.rchain.shared.AttemptOpsF$.ex(AttemptOpsF.scala:10) at coop.rchain.shared.AttemptOpsF$RichAttempt.get(AttemptOpsF.scala:19) at coop.rchain.rspace.history.HistoryStoreInstances$$anon$1.$anonfun$get$2(HistoryStore.scala:38) at cats.instances.OptionInstances$$anon$1.traverse(option.scala:87) at cats.instances.OptionInstances$$anon$1.traverse(option.scala:15) at cats.Traverse$Ops.traverse(Traverse.scala:19) at cats.Traverse$Ops.traverse$(Traverse.scala:19) at cats.Traverse$ToTraverseOps$$anon$2.traverse(Traverse.scala:19) at coop.rchain.rspace.history.HistoryStoreInstances$$anon$1.$anonfun$get$1(HistoryStore.scala:38) at scala.Function1.$anonfun$andThen$1(Function1.scala:57) at monix.eval.internal.TaskRunLoop$.startFull(TaskRunLoop.scala:170) at monix.eval.internal.TaskRunLoop$.$anonfun$restartAsync$1(TaskRunLoop.scala:222) at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) at java.base/java.util.concurrent.FutureTask.run(Unknown Source) at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) at java.base/java.lang.Thread.run(Unknown Source)
```

We don't know the condition to reproduce the situation. Current solution is to put more debug info into the history store when this bug happened.

### LINKED ISSUE #3127: What is the performance lost in lmdb proxy_safe compared to proxy_optimal?
Related documents:
https://github.com/lmdbjava/lmdbjava/wiki/Buffers#bytebuffer

java lmdb provided two different buffer proxy for `DirectBuffer` in jvm.

Currently rnode using `PROXY_OPTIMAL` which is not safe in jvm(jvm doesn't encourage the opertion does by `PROXY_OPTIMAL`). THere is another option for bufferproxy which is `PROXY_SAFE`.

What is the performance differences between these two bufferproxy?

zsluedem:
## Overview
~~put debug info in history store when error happen~~

Use PROXY_SAFE for lmdb

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
@@ -9,6 +9,7 @@ import coop.rchain.lmdb.LMDBStore
 import coop.rchain.shared.ByteVectorOps.RichByteVector
 import coop.rchain.rspace.Blake2b256Hash
 import org.lmdbjava.DbiFlags.MDB_CREATE
+import org.lmdbjava.ByteBufferProxy.PROXY_SAFE
 import org.lmdbjava.{Env, EnvFlags, Txn}
 import scodec.bits.BitVector
 
@@ -39,7 +40,7 @@ object StoreInstances {
     for {
       env <- Sync[F].delay {
               Env
-                .create()
+                .create(PROXY_SAFE)
                 .setMapSize(config.mapSize)
                 .setMaxDbs(config.maxDbs)
                 .setMaxReaders(config.maxReaders)
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -28,6 +28,7 @@ import coop.rchain.shared.ByteStringOps._
 import coop.rchain.shared.{AtomicMonadState, Log}
 import monix.execution.atomic.AtomicAny
 import org.lmdbjava.DbiFlags.MDB_CREATE
+import org.lmdbjava.ByteBufferProxy.PROXY_SAFE
 import org.lmdbjava._
 
 import scala.util.matching.Regex
@@ -335,7 +336,7 @@ object FileLMDBIndexBlockStore {
       env <- Sync[F].delay {
               val flags = if (config.noTls) List(EnvFlags.MDB_NOTLS) else List.empty
               Env
-                .create()
+                .create(PROXY_SAFE)
                 .setMapSize(config.mapSize)
                 .setMaxDbs(config.maxDbs)
                 .setMaxReaders(config.maxReaders)
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagFileStorage.scala
@@ -33,6 +33,7 @@ import coop.rchain.shared.{AtomicMonadState, Log, LogSource}
 import monix.execution.atomic.AtomicAny
 import org.lmdbjava.DbiFlags.MDB_CREATE
 import org.lmdbjava.{Env, EnvFlags}
+import org.lmdbjava.ByteBufferProxy.PROXY_SAFE
 
 import scala.ref.WeakReference
 import scala.util.matching.Regex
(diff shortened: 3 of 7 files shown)


rspace/src/main/scala/coop/rchain/rspace/history/HistoryStore.scala
```diff
@@ -35,7 +37,24 @@ object HistoryStoreInstances {
     override def get(key: Blake2b256Hash): F[Trie] =
       for {
         maybeBytes <- store.get(key)
```
nzpr: I know this is aesthetic but I have an allergy to these `maybe` names. 
Can we use `bytesOpt`?
Actually we should have everywhere `unsafeGet` extensions, so we don't have tons of such options across code that expresses useful logic.

nzpr (approved):
LGTM

rspace/src/main/scala/coop/rchain/rspace/history/HistoryStore.scala
tgrospic: In Haskell `Maybe` type represents the same type as `Option` in Scala so from that perspective both of these terms makes sense. But I agree if we want to be more precise in naming.

I agree that `unsafe` extensions can simplify the code but only in situations where the value is expected and generic error is thrown otherwise. So here it is more applicable to `Trie.codecTrie.unsafeDecode(bytes)` instead to store get. (which is similar what `Trie.codecTrie.decode(bytes).get` is doing)

rspace/src/main/scala/coop/rchain/rspace/history/HistoryStore.scala
```diff
@@ -34,8 +36,25 @@ object HistoryStoreInstances {
...
+                                  Sync[F].pure(res)
+                                case Attempt.Failure(err) =>
+                                  Sync[F]
+                                    .raiseError[DecodeResult[Trie]](
+                                      new Exception(
+                                        s"Key: ${Base16.encode(key.bytes.toArray)}\nValue: ${Base16
+                                          .encode(bytes.toByteArray)}\n${err.toString()}"
+                                      )
+                                    )
+                              }
```
tgrospic: The work of converting `Attempt` to Applicative error is already done in `decode(..).get` so here we can just "hook" on the error to display additional info. `onError` does not handle the error.
```scala
result <- maybeBytes.traverse { bytes =>
            Trie.codecTrie.decode(bytes).get.onError {
              case ex: Throwable =>
                val errMsg =
                  s"Key: ${Base16.encode(key.bytes.toArray)}\nValue: ${bytes.toHex}"
                Log[F].error(errMsg, ex)
            }
          }
```
Can you also try to implement retry logic so we can see if very next get will have the same error?

rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
```diff
@@ -34,7 +35,7 @@ object StoreInstances {
     for {
       env <- Sync[F].delay {
               Env
-                .create()
+                .create(PROXY_SAFE)
```
tgrospic, line 43: Do you know what is performance impact of "safe buffers"?

nzpr:
With merging https://github.com/rchain/rchain/pull/3121 we don't need this PR anymore, right?

tgrospic:
> With merging #3121 we don't need this PR anymore, right?

Let's keep it a little longer. Also I'm interested if it makes sense to use `PROXY_SAFE` option for buffers.

rspace/src/main/scala/coop/rchain/rspace/history/Store.scala
zsluedem: https://github.com/rchain/rchain/issues/3127

zsluedem:
@nzpr @tgrospic Wanna use ProxySafe?

tgrospic (approved):
LGTM

zsluedem:
bors merge

zsluedem:
https://github.com/rchain/rchain/issues/3127 performance impact for using PROXY_SAFE

zsluedem:
bors merge

PR #3120, opened 2020-09-08
Make Justification regression work with LFS

### LINKED ISSUE #3119: Justification Regression validation might be broken with LFS
Current logic of justification regression check has flaw:

Hope graph and `dot` notation or graph below helps understanding the issue.
```dot
digraph G {
  bX ->  LFS
  LFS -> LMv1
  LFS -> bA
  LFS -> LMv3
  LFS -> LMv4
  LFS -> LMv5


  LMv1 -> B1
  B1 -> bA
}
```
Screenshot 2020-09-08 at 16 18 28
Let assume validator A (`vA`) has been bonded in block `bA`, but does not produce any blocks.

When new block `bX` having `bA` as justification is being added, it is being checked for justification regression.
This involves comparing two justifications originated by `vA`: for block `bX` (name it `jXA`) and block `bA` (name it `jAA`).

As `vA` just been bonded in `bA`, this means that `jAA`is non existent.

In this case genesis block used for justification. 
https://github.com/rchain/rchain/blob/dac6dab1e7e21bbdd1b910a41adcdd591ba90349/casper/src/main/scala/coop/rchain/casper/Validate.scala#L593-L597
Which is wrong, because with introducing of approved (finalised) state genesis block can have greater height then `bA`

Suggested change - assume no justification regression in this case without involving genesis.

nzpr:
Testing of LFS revealed (unexpectedly) possible flaw in justification regression detector.

JR detector can be broken when validator bonds and does not propose blocks.

More info in related issue

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -187,7 +187,7 @@ object Validate {
       _ <- EitherT.liftF(Span[F].mark("before-sequence-number-validation"))
       _ <- EitherT(Validate.sequenceNumber(block, dag))
       _ <- EitherT.liftF(Span[F].mark("before-justification-regression-validation"))
-      _ <- EitherT(Validate.justificationRegressions(block, genesis, dag))
+      _ <- EitherT(Validate.justificationRegressions(block, dag))
       _ <- EitherT.liftF(Span[F].mark("before-shard-identifier-validation"))
       s <- EitherT(Validate.shardIdentifier(block, shardId))
     } yield s).value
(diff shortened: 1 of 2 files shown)


nzpr:
I made some refactoring as well, but the fix is basically is this part of the code
https://github.com/rchain/rchain/blob/dac6dab1e7e21bbdd1b910a41adcdd591ba90349/casper/src/main/scala/coop/rchain/casper/Validate.scala#L593-L597 to assume no regression if `!previousLatestMessages.contains(previousLatestMessages)`

tgrospic:
Should we add this to dev and rebase LFS?

nzpr:
@tgrospic I believe its ok if we have it in LFS only, its not some urgent bugfix

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -545,105 +545,73 @@ object Validate {
...
+                  .buildString(sender)} has a lower sequence number than ${PrettyPrinter
+                  .buildString(currentHash)}."
+                Log[F].warn(ignore(b, msg))
+              }
+              // Compare and check for regression
+              val regressionDetected = for {
+                curJustificationHash <- curLMs(sender).pure[F]
+                newJustification     <- dag.lookupUnsafe(newJustificationHash)
+                curJustification     <- dag.lookupUnsafe(curJustificationHash)
+                regression           = (!newJustification.invalid && newJustification.seqNum < curJustification.seqNum)
```
tgrospic: 3 LOC instead of 26. Nice :smile:

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -545,105 +545,73 @@ object Validate {
...
+            } else {
+              def logWarn(currentHash: BlockHash, regressiveHash: BlockHash): F[Unit] = {
+                val msg = s"block ${PrettyPrinter.buildString(regressiveHash)} by ${PrettyPrinter
+                  .buildString(sender)} has a lower sequence number than ${PrettyPrinter
+                  .buildString(currentHash)}."
+                Log[F].warn(ignore(b, msg))
+              }
+              // Compare and check for regression
+              val regressionDetected = for {
+                curJustificationHash <- curLMs(sender).pure[F]
```
tgrospic: There is no need to wrap this in `F` just to be used in `for` comprehension. It can be defined as "normal" variable before `for`.

tgrospic (approved):
Very nice fix!

tgrospic:
In equivocation check is another place where genesis is also used. We should see how this is reflected with LFS.
https://github.com/rchain/rchain/blob/006e728637bc363a273fda70f7b663c0d10cb956/casper/src/main/scala/coop/rchain/casper/EquivocationDetector.scala#L274-L285

casper/src/main/scala/coop/rchain/casper/Validate.scala
nzpr: Yeah, I do this to avoid additional bracket section, makes it less readable

zsluedem (approved):
Good work!

nzpr:
bors merge

nzpr:
bors merge

PR #3123, opened 2020-09-10
Refactor dagstore insert, allow reverse order adding

nzpr:
This fixes some flaws in bockDAG insert method, that are required for LFS.
1. do not update latest message if seqNum goes backwards
2. do not update latest message for newly bonded validators if blocks are added in reverse order.
3. Refactor

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -142,29 +142,29 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
     // Empty sender is valid for genesis
     val senderIsEmpty          = block.sender == ByteString.EMPTY
-    val senderHasInvalidFormat = (block.sender.size() != Validator.Length) && !senderIsEmpty
+    val senderHasInvalidFormat = !senderIsEmpty && (block.sender.size() != Validator.Length)
     val sendersNewLM           = (block.sender, block.blockHash)
 
     val logAlreadyStored =
-      Log[F].warn(s"Block ${PrettyPrinter.buildString(block, short = true)} is already stored.")
+      Log[F].warn(s"${PrettyPrinter.buildString(block, short = true)} is already stored.")
 
     val logEmptySender =
-      Log[F].warn(s"Block ${Base16.encode(block.blockHash.toByteArray)} sender is empty")
+      Log[F].warn(s"${PrettyPrinter.buildString(block, short = true)} sender is empty")
 
+    // Add LM either if there is no existing message for the sender, or if sequence number advances
+    // - assumes block sender is not valid hash
     def shouldAddAsLatest: F[Boolean] =
-      // Do not add LM for empty senders
-      (!senderIsEmpty).pure[F] &&^
-        // Add LM either if there is no one for sender, or if sequence number advances
-        (latestMessagesIndex.contains(block.sender).not ||^
-          latestMessagesIndex
-          // Try get sender's latest message
-            .get(block.sender)
-            // Get metadata from index
-            .flatMap(_.traverse(blockMetadataIndex.getUnsafe))
-            // Check if seq number is greater that existing
-            .map(_.map(block.seqNum >= _.seqNum))
-            // Evaluate option and result
-            .map(_.contains(true)))
+      latestMessagesIndex
+      // Try get sender's latest message
+        .get(block.sender)
+        // Get metadata from index
+        .flatMap(_.traverse(blockMetadataIndex.getUnsafe))
+        // Check if seq number is greater that existing
+        .map(_.map(_.seqNum))
+        // Evaluate option and result. Add if:
+        // - latest message is not found, or
+        // - is found with seq num greater then existing
+        .map(lmSeqNumOpt => lmSeqNumOpt.isEmpty || lmSeqNumOpt.exists(block.seqNum >= _))
 
     def newLatestMessages: F[Map[Validator, BlockHash]] = {
       val newlyBondedSet = bonds(block)
@@ -173,12 +173,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
         .diff(block.justifications.map(_.validator).toSet)
       for {
         // This filter is required to enable adding blocks backward from higher height to lower
-        newlyBondedUnseen <- newlyBondedSet.toList.filterA(
-                              lm => latestMessagesIndex.contains(lm).not
-                            )
-        newlyBondedLMs = newlyBondedUnseen.map(v => (v, block.blockHash)).toMap
-
-      } yield newlyBondedLMs
+        newlyBondedUnseen <- newlyBondedSet.toList.filterA(latestMessagesIndex.contains(_).not)
+      } yield newlyBondedUnseen.map((_, block.blockHash)).toMap
     }
 
     def doInsert: F[Unit] = {
@@ -208,11 +204,10 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
         _ <- invalidBlocksIndex.put(blockMetadata.blockHash, blockMetadata).whenA(invalid)
 
         // Resolve if block should be added as the latest message for the block sender
-        newLatestFromSender <- (senderIsEmpty.pure[F].not &&^ shouldAddAsLatest)
-                                .ifM(
-                                  Map(sendersNewLM).pure[F],
-                                  Map.empty[Validator, BlockHash].pure[F]
-                                )
+        emptyLM = Map.empty[Validator, BlockHash].pure[F]
+        newLatestFromSender <- if (!senderIsEmpty)
+                                shouldAddAsLatest.ifM(Map(sendersNewLM).pure[F], emptyLM)
+                              else emptyLM
 
         // Add/update validators latest messages
         newLatestFromNewValidators <- newLatestMessages
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -10,6 +10,7 @@ import coop.rchain.blockstorage.dag.BlockMetadataStore.BlockMetadataStore
 import coop.rchain.blockstorage.dag.EquivocationTrackerStore.EquivocationTrackerStore
 import coop.rchain.blockstorage.dag.codecs._
 import coop.rchain.blockstorage.util.BlockMessageUtil._
+import coop.rchain.casper.PrettyPrinter
 import coop.rchain.casper.protocol.BlockMessage
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.metrics.Metrics.Source
@@ -19,6 +20,7 @@ import coop.rchain.models.EquivocationRecord.SequenceNumber
 import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
 import coop.rchain.shared.syntax._
+import coop.rchain.blockstorage.syntax._
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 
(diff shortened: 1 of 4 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +134,95 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+      val senderIsNewlyBonded  = zeroBlock || newlyBonded.contains(block.sender)
+      val senderHasValidFormat = block.sender.size() == Validator.Length
+      val logEmptySender =
+        Log[F].warn(s"Block ${Base16.encode(block.blockHash.toByteArray)} sender is empty")
+
+      // LM should be updated only when block sequence number advances.
+      val senderCurLMSeqNum = latestMessagesIndex
+        .getUnsafe(block.sender)
+        .flatMap(blockMetadataIndex.getUnsafe)
+        .map(_.seqNum)
```
tgrospic: Nice!

tgrospic (approved):
Nice work!

nzpr:
I still need to fix some tests. Bonds map in genesis block for some tests is empty list, which does not produce latest messages (and which is wrong).

zsluedem (approved):
LGTM

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +134,116 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
-  ): F[BlockDagRepresentation[F]] =
-    lock.withPermit(
+  ): F[BlockDagRepresentation[F]] = {
+    import cats.instances.list._
+    import coop.rchain.catscontrib.Catscontrib.ToBooleanF
+
+    val logAlreadyStored =
+      Log[F].warn(s"Block ${PrettyPrinter.buildString(block, short = true)} is already stored.")
+
+    val updateLatestMessages = {
```
tgrospic: This block is tricky to understand. It's not executed until `doInsert` is called but it's forces the reader to really type check a lot in the head.
What about refactoring this block in private function(s) that can be called from `doInsert`? It can be more clear about each step of checking and how data is connected.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +134,116 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                        // genesis has empty sender, so we have to put additional check
+                        .whenA(!senderHasValidFormat && !senderIsEmpty)
+                  _ <- logEmptySender.whenA(senderIsEmpty)
+                  // As we can add blocks in backwards (e.g. when reconstructing finalized state) -
+                  // to ot overwrite existing LMs, we should have this check
+                  seqNumIsRegressive <- checkRegressiveSeqNum
+                  r = if (seqNumIsRegressive)
+                    newlyBondedLMs
+                  else
+                    newlyBondedLMs + sendersNewLM
```
tgrospic: Here is a situation where two things can happen (not counting error a few lines above), `sendersNewLM` will be added to `newlyBondedLMs` or not.
So to be clear it would be nice to have some king of "logical table" where is more visible the dependency between expressions.

E.g. `senderIsEmpty` is basically on top, if we don't have a sender all bets are off to add it to the latest messages, so it make sense to use it only in the last definition like:
`val shouldAddToLatest = senderIsEmpty && allOtherChecks`.
For example now I'm not sure it `checkRegressiveSeqNum` should be executed if sender is empty.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
```diff
@@ -9,12 +10,25 @@ trait KeyValueTypedStoreSyntax {
...
+  def getUnsafe(key: K)(
+      implicit f: Sync[F],
+      line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing
+  ): F[V] = {
+    def source = s"${file.value}:${line.value} ${enclosing.value}"
+    def errMsg = s"KeyValueTypedStore is missing key ${key.toString}\n  $source"
+    get(key) >>= (_.liftTo(KeyValueTypedStoreInconsistencyError(errMsg)))
+  }
```
tgrospic: As you noticed, `K` here will not be printed correctly because it's not serialized first.

Please create another instance of this extension in block-storage project with `K`  replaced with concrete type. This is a good example how it's easy to add more specific version of a function in the project where generic type is known.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Yeah, all this is quite cumbersome. But thats because of that super special genesis block, which I was hoping to refactor at some point. But I agree, let me think a bit on this again.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
```diff
@@ -9,6 +10,8 @@ trait KeyValueTypedStoreSyntax {
   ): KeyValueTypedStoreOps[F, K, V] = new KeyValueTypedStoreOps[F, K, V](store)
 }
 
+final case class KeyValueTypedStoreInconsistencyError(message: String) extends Exception(message)
```
tgrospic: Changes in this file can be reverted, right?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,100 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                    .flatMap(blockMetadataIndex.getUnsafe)
+                    .map(_.seqNum)
+                  val checkRegressiveSeqNum = senderCurLMSeqNum.map(block.seqNum < _)
+
+                  for {
+                    _ <- Sync[F]
+                          .raiseError[(ByteString, ByteString)](BlockSenderIsMalformed(block))
+                          // genesis has empty sender, so we have to put additional check
+                          .whenA(!senderHasValidFormat && !senderIsEmpty)
+                    _           <- logEmptySender.whenA(senderIsEmpty)
```
tgrospic: `senderIsEmpty` is not used to decide if `sendersNewLM` will be added to `newlyBondedLMs` so it doesn't have to be part of the same logic.
Additional helper function can be useful because it can only calculate case when sender is not empty.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,100 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                    senderHasLM <- latestMessagesIndex.contains(block.sender)
```
tgrospic: It's possible to get here with empty sender. Is this necessary to check?

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
nzpr: I believe this file is not changed now

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,109 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                                 sendersCurLM <- latestMessagesIndex.get(block.sender)
+                                 r <- sendersCurLM match {
+                                       case None => false.pure[F]
+                                       case Some(lm) =>
+                                         blockMetadataIndex.getUnsafe(lm).map(_.seqNum).map(_ == 0)
+                                     }
+                               } yield r
+
+          // Do not create LM for empty sender, which is a now valid sender for genesis
+          lms <- if (senderIsEmpty || (newlyBondedSet.contains(block.sender) && !sendersLmIsGenesis))
```
tgrospic: Why is necessary to have `sendersLmIsGenesis`?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,109 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                  logEmptySender.whenA(senderIsEmpty) >> newlyBondedLMs.pure[F]
```
tgrospic: Log will be printed that sender is empty even when sender is not empty, and right expression is true.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,109 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                // Update LM for sender
+                else {
+                  // LM should be updated only when block sequence number advances.
+                  val senderCurLMSeqNum = latestMessagesIndex
+                    .getUnsafe(block.sender)
```
tgrospic: Here block data is read again from the index as in line 172.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,109 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                    .flatMap(blockMetadataIndex.getUnsafe)
+                    .map(_.seqNum)
+                  val checkRegressiveSeqNum  = senderCurLMSeqNum.map(block.seqNum < _)
+                  val senderHasInvalidFormat = block.sender.size() != Validator.Length
```
tgrospic: This check can be done immediately and check can fail fast before any other checks.

nzpr: The thing is that empty sender is allowed, so `senderHasInvalidFormat` has to be in clause when LM is updated

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: I guess you was looking at outdated code? `.whenA(senderIsEmpty)` is in place

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Does not the comment above answer?
https://github.com/rchain/rchain/blob/d7b308bdfe4ef36b89734a70c87e5e3da53dc53e/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala#L168-L179

nzpr: LM for validator is created when processing genesis. But the very first block does not have self justification, so won't try to update lates message assuming it is regressive. Imagine adding block from the top to the bottom - sender has LM all the time and in the end meet block with no self-justification. In general, we should not update LM in this case. But here is exactly the same situation, except we should update LM.
If we would create LM when block is created by sender, not newly bonded - we wont need to have this workaround. Or if genesis can have a set of senders.

nzpr:
@tgrospic I simplified the code, does it look better now? One drawback that I can see now is that it relies on fact that `newlyBondedLMs` can implicitly contains senders new LM. So `newlyBondedLMs + sendersNewLM` can be equal to `newlyBondedLMs`. But thats much cleaner, though involves some magic.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -133,57 +135,95 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+
+    val logAlreadyStored =
+      Log[F].warn(s"Block ${PrettyPrinter.buildString(block, short = true)} is already stored.")
+
+    val logEmptySender =
+      Log[F].warn(s"Block ${Base16.encode(block.blockHash.toByteArray)} sender is empty")
+
+    // Add LM either if there is no existing message for the sender, or if sequence number advances
+    // - assumes block sender is not valid hash
+    def shouldAddAsLatest: F[Boolean] =
```
tgrospic, line 156: We don't need to check if sender is empty here because this function should be called only for valid sender.

nzpr:
bors merge

PR #3124, opened 2020-09-11
Enable per peer incoming message queues

### LINKED ISSUE #3056: Use one gRPC message consume queue per peer
## Overview
Enable per peer server queue as it was has been done with client channel <https://github.com/rchain/rchain/pull/3006>

### Impact
This class should be adjusted in a fashion similar to PR above https://github.com/rchain/rchain/blob/7a34cd358d799448a39ecd52168914e999bdc587/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportServer.scala

nzpr:
Closes https://github.com/rchain/rchain/issues/3056


integration-tests/run_tests
```diff
@@ -25,8 +25,9 @@ fi
 
 cleanup=0
 
-if [[ ! -v TEMP_RESOURCES_DIR ]]; then
-	TEMP_RESOURCES_DIR=$(mktemp --tmpdir -d integration-tests.XXXXXXXXXX)
+TMPDIR="/tmp"
+if [[ -z ${TEMP_RESOURCES_DIR+x} ]]; then
+	TEMP_RESOURCES_DIR=$(mktemp -d "${TMPDIR:-/tmp}/integratopn-tests.XXXXXXXXX")
```
tgrospic: Is `TMPDIR` defines absolute path, if yes, why is this necessary? What is the meaning of `+x` suffix?

integration-tests/test/rnode.py
```diff
@@ -376,7 +377,7 @@ def run(self) -> None:
                 if self.terminate_thread_event.is_set():
                     break
                 line = next(containers_log_lines_generator)
-                self.logger.info('{}: {}'.format(self.container.name, line.decode('utf-8').rstrip()))
+                self.logger.info('{:>11}: {}'.format(self.container.name[-11:], line.decode('utf-8').rstrip()))
```
tgrospic, line 380: Please add a comment here why 11 is special number here.

integration-tests/resources/logback.xml
```diff
@@ -0,0 +1,31 @@
...
+      <level>DEBUG</level>
+    </filter>
+    <!-- encoders are  by default assigned the type ch.qos.logback.classic.encoder.PatternLayoutEncoder -->
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%highlight(%-5level{5})] [%cyan(%-5.5thread{5})] [%cyan(%-15.15logger{15})] - %msg%n </pattern>
+    </encoder>
+  </appender>
+
+  <logger name="coop.rchain.shared.EventLogger" level="OFF" />
+  <logger name="io.opencensus" level="OFF" />
```
zsluedem, line 20: This line can make the error logs in `rnode.log` disappear, right?

zsluedem (approved):
LGTM

integration-tests/resources/logback.xml
nzpr: There are DEBUG messages from package `io.opencensus` with big output, this disables them

integration-tests/test/rnode.py
nzpr: But its just trimming word and making alignment, usual stuff for `format`, do you think it worth comment? It really does not matter, it just give more room for actual log messages

integration-tests/run_tests
nzpr: TMPDIR env var is required on OS X because default folder `mktemp` cannot be used by docker by default.
`+x` makes sure condition fails if `TEMP_RESOURCES_DIR ` is set but null. I've seen it somewhere as some more reliable way to detect if var is unset.

tgrospic: Can we at least use relative folder inside project? This is very invasive to create new folder on the root. It's not what would you expect when running a test.
Maybe to fail with meaningful error and ask user to set some environment variable.

nzpr: I'm not sure I follow - `mktemp` by default creates in `/tmp` on Linux, so nothing changes there, its only makes it work for OS X. `TEMP_RESOURCES_DIR ` some random folder in /tmp/integration-tests. Actually I have a typo here

tgrospic: Location of the temp folder is user configuration and can be any folder not only `/tmp`. So IMO, creating `/tmp` folder in the root of the drive may not be what user expects.
I'll agree that this is not so important so I'll approve this PR.

tgrospic (approved):

nzpr:
bors merge

integration-tests/run_tests
nzpr: But `/tmp` is expected to exist in any POSIX OS and server as temp data, do I miss smth?

PR #3130, opened 2020-09-16
Introduce dev mode && enable exploratory deploys in dev mode

nzpr:

### DIFF
--- a/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/NodeRuntime.scala
@@ -892,7 +892,8 @@ object NodeRuntime {
           evalRuntime,
           blockApiLock,
           scheduler,
-          conf.apiServer.maxBlocksLimit
+          conf.apiServer.maxBlocksLimit,
+          conf.devMode
         )(
           blockStore,
           oracle,
@@ -940,7 +941,7 @@ object NodeRuntime {
         implicit val sp = span
         implicit val or = oracle
         implicit val bs = blockStore
-        new WebApiImpl[F](conf.apiServer.maxBlocksLimit)
+        new WebApiImpl[F](conf.apiServer.maxBlocksLimit, conf.devMode)
       }
       adminWebApi = {
         implicit val ec     = engineCell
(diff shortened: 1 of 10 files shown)


shared/src/main/scala/coop/rchain/config/SharedConf.scala
```diff
@@ -0,0 +1,10 @@
+package coop.rchain.config
+
+final case class SharedConf[F[_]](
```
tgrospic: Why is `F` defined here if it's not used?

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -940,6 +943,7 @@ object NodeRuntime {
         implicit val sp = span
         implicit val or = oracle
         implicit val bs = blockStore
+        implicit val sc = sharedConf
         new WebApiImpl[F](conf.apiServer.maxBlocksLimit)
```
tgrospic: What is justification of adding new **implicit** value and not just add "normal" parameter.

We should be careful when introducing new implicits and reduce its use when it's possible.

nzpr: @tgrospic this is what I was going to ask. Basically - this `--dev-mode` option potentially can influence all modules: casper, blockstorage, etc. I was thinking that using this style of passing implicit w/o explicit argument is more suitable, and pushing argument from node runtime down to all methods that are using this arg is too much, as mostly this arg won't be used by intermediary caller methods.

nzpr: I basically want to have this config state available from everywhere without too much hustle. Maybe this is over-optimisation though

tgrospic: I'm not against implicits for configuration. https://github.com/rchain/rchain/pull/2910#discussion_r396403582

From the perspective of WebAPI implementation. It's calling BlockAPI implementation. So now the question is why WebAPI implementation needs to know what BlockAPI has in configuration? What we want is not rewrite in different form, but to not write it at all. :)

But we don't have this kind of separation to be able to do that so for now I'm suggesting to have it as a parameter so that can we can just remove it later when we create some configuration strategy.

This is also known as _configuration problem_ and there are multiple solutions.
http://okmij.org/ftp/Haskell/tr-15-04.pdf

nzpr:
I cut it down to just passing devMode option to exploratoryDeploy method for now

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3134, opened 2020-09-17
Tests for LFS block requester

tgrospic:
## Overview
This PR contains initial set of tests for receiver of blocks for Last Finalized State. Tests are separated in two pieces, tests for internal state and effects (network requests and block store operations).

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3136, opened 2020-09-18
fix data inconsistent bug in tests

zsluedem:
## Overview
Fix the tests data inconsistent bug

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/engine/InitializingSpec.scala
+++ b/casper/src/test/scala/coop/rchain/casper/engine/InitializingSpec.scala
@@ -116,10 +116,8 @@ class InitializingSpec extends WordSpec with BeforeAndAfterEach {
         // Assert requested messages for the state and fork choice tip
         _        = assert(transportLayer.requests.size == expectedRequests.size)
         messages = transportLayer.requests.map(_.msg)
-        _ = messages.zip(expectedRequests).map {
-          case (msg1, msg2) => assert(msg1 == msg2)
-        }
-        _ = transportLayer.reset()
+        _        = assert(messages.toSet == expectedRequests.toSet)
+        _        = transportLayer.reset()
 
         // assert that we really serve last approved block
         lastApprovedBlockO <- LastApprovedBlock[Task].get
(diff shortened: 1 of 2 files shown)


comm/src/test/scala/coop/rchain/p2p/EffectsTestInstances.scala
```diff
@@ -63,26 +64,36 @@ object EffectsTestInstances {
     type Responses = PeerNode => Protocol => CommErr[Unit]
     var reqresp: Option[Responses] = None
     var requests: List[Request]    = List.empty[Request]
+    val lock                       = new SyncVar[TransportLayerStub[F]]()
+    lock.put(this)
```
tgrospic: There is `SyncVarOps.create` helper function to initialize SyncVar.

casper/src/test/scala/coop/rchain/casper/engine/InitializingSpec.scala
```diff
@@ -116,10 +116,8 @@ class InitializingSpec extends WordSpec with BeforeAndAfterEach {
         // Assert requested messages for the state and fork choice tip
         _        = assert(transportLayer.requests.size == expectedRequests.size)
         messages = transportLayer.requests.map(_.msg)
-        _ = messages.zip(expectedRequests).map {
-          case (msg1, msg2) => assert(msg1 == msg2)
-        }
-        _ = transportLayer.reset()
+        _        = assert(messages.forall(m => expectedRequests.exists(e => e == m)))
+        _        = transportLayer.reset()
```
tgrospic, line 120: This works but it's not precise enough. We should expect all requests correctly and not just part of it.

Order can be non-deterministic so comparing sets can be enough.
`assert(messages.toSet == expectedRequests.toSet)`

Maybe also delay in line 102 of 1 second is not enough to create the same requests. :thinking:

zsluedem: `forall` is not enough?

tgrospic: Not sure what you mean.

tgrospic: P.S. Overall, I'm not sure what this test is supposed to check. If just Initializing engine why all other stuff need to be created, like everything (block store, dag store, casper, ...) and also in the same way as in normal node operation, with LMDB and on file system.

tgrospic: I changed delay to 5 seconds. This should be enough for now.
https://github.com/rchain/rchain/pull/3134/files#diff-91178b6525f827b57ac86c3d7c1a3cd1R102

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

PR #3145, opened 2020-09-23
Validation for LFS import

### LINKED ISSUE #3098: Testing of LFS
## Overview

This issue is to record several rounds of testing of **Last Finalized State** pre release versions. Tests are more performance oriented and some of the bug fixes and optimitations are done between each test setup.

Results are starting from the last setup. Shows how speed is increased with stronger machine. This is due to additonal computation needed for verification of the Rholang state (#3145). This is visible in high usage of CPU during sync.  

Another important factor is disk speed. All tests are done on SSD disks and on stronger machines faster speed reduces overall sync duration.

### Testing setup 4

This round is almost identical to _setup 3_ with additional optimization in receiving blocks (#3243) which include block sorting as part of download and with testing on 32 vCPU machine.

RNode Docker image version: [`rchain/rnode:v0.9.26-rc`](https://hub.docker.com/layers/rchain/rnode/v0.9.26-rc/images/sha256-af9d5c6467474eb6e47d1557dfdef26c312a5834eeb3fed5c94b8d6ce25a5ee9?context=explore)

| Nr. | Sync time<br>(fully catch up) | Sync Rholang state | Sync blocks | VPS | Log
| --- | ---        | ---    | ---    | ---                         | ---
| 1   | **2.25 h** | 1.5 h  | 1.75 h | 32vCPU 64GB (DigitalOcean)  | [test-4-do-32vcpu-64g.zip](https://github.com/rchain/rchain/files/5630012/test-4-do-32vcpu-64g.zip)
| 2   | **3.25 h** | 1.5 h  | 2 h    | 16vCPU 32GB CPX51 (Hetzner) | [test-4-cpx51-32g.zip](https://github.com/rchain/rchain/files/5630016/test-4-cpx51-32g.zip)
| 3   | **3 h**    | 1.75 h | 2 h    | 8vCPU 32GB (IBM)                   | [test-4-ibm-8vcpu-32g.zip](https://github.com/rchain/rchain/files/5630025/test-4-ibm-8vcpu-32g.zip)
| 4   | **4 h**    | 2.25 h | 2.5 h  | 8vCPU 16GB CPX41 (Hetzner)  | [test-4-cpx41-16g.zip](https://github.com/rchain/rchain/files/5630022/test-4-cpx41-16g.zip)
| 5   | **8 h**    | 4.5 h  | 3.5 h  | 4vCPU 8GB CPX31 (Hetzner)   | [test-4-cpx31-8g.zip](https://github.com/rchain/rchain/files/5630030/test-4-cpx31-8g.zip)

NOTE: CPU and memory observed in testing is the same as in setup 3, only overall duration is shorter.

[1] CPU, memory - 32vCPU 64GB (DigitalOcean)
test-4-do-32vcpu-64g-lfs-progress-cpu

### Testing setup 3

First two rounds of testing showed that LFS source node disk reading speed can make significant difference in overall speed for nodes to download Rholang state. In this testing setup,  to make reading from disk much faster RSpace folder on source node is converted to RAM disk which reduced reading from ~20sec to ~1sec for chunk of state sent over the network.

RNode Docker image version: [`tgrospic/rnode:v0.9.26-rc`](https://hub.docker.com/layers/tgrospic/rnode/v0.9.26-rc/images/sha256-85bb8d444b354b0628d871b561c2e908131f79cef880ece5b86678f64fb49195?context=repo)

NOTE: Bug with memory leak on network errors is still not resolved and one test with 8GB machine failed not listed in results.

| Nr. | Sync time<br>(fully catch up) | VPS
| --- | ---  | --- | 
| 1 | 4 hours (2.5 h blocks)<br>[test-3-cpx51-32g-full-ok-3.zip](https://github.com/rchain/rchain/files/5629935/test-3-cpx51-32g-full-ok-3.zip) | 16vCPU 32GB CPX51
| 2 | 5.5 hours (3.5 h blocks)<br>[test-3-cpx41-16g-full-ok-1.zip](https://github.com/rchain/rchain/files/5629947/test-3-cpx41-16g-full-ok-1.zip) | 8vCPU 16GB CPX41
| 3 | 9 hours (4.5 h blocks)<br>[test-3-cpx31-8g-full-ok-1.zip](https://github.com/rchain/rchain/files/5629957/test-3-cpx31-8g-full-ok-1.zip) | 4vCPU 8GB CPX31

#### Observations from testing

[1] CPU, memory - 16vCPU 32GB CPX51
test-3-cpx51-32g-lfs-progress-cpu

[1] Direct buffer - 16vCPU 32GB CPX51
test-3-cpx51-32g-lfs-progress-buffer

[2] CPU, memory - 8vCPU 16GB CPX41
test-3-16g-lfs-progress-cpu

[3] CPU, memory - 4vCPU 8GB CPX31
test-3-8g-lfs-progress-cpu

Survived network errors without crashing - 4vCPU 8GB CPX31
test-3-8g-lfs-progress-3-errors-cpu

### Testing setup 2

RNode Docker image version: [`tgrospic/rnode:v0.9.26-beta`](https://hub.docker.com/layers/tgrospic/rnode/v0.9.26-beta/images/sha256-0734a74f064939d43208d82c6914f64a0917b4de95cca0ae863a3999f2d53e3c?context=repo)

| Nr. | Source | Duration | Sync time<br>(fully catch up) | Environment
| --- | ---  | ---   | --- | ---
| 1 | Full  | 12.5 hours (2.5 h blocks) | - | One machine + 3 active nodes
| 2 | LFS  | ERROR interrupted<br>[rnode-restart-error.zip](https://github.com/rchain/rchain/files/5241197/rnode-restart-error.zip) | - | Target alone on another machine<br>(Docker on Hetzner cloud CPX31)
| 3 | LFS [2] | 4.5 hours (2.5 blocks)<br> [rnode-test-3.zip](https://github.com/rchain/rchain/files/5244082/rnode-test-3.zip) | 8 hours | Target alone on another machine<br>(Docker on Hetzner cloud CPX31)
| 4 | LFS [3] | 6 hours (2 blocks)<br> [rnode-test-4.zip](https://github.com/rchain/rchain/files/5248452/rnode-test-4.zip) | 10 hours | Target alone on another machine<br>(Docker on Hetzner cloud CX31)

#### Observations from testing

[3] After node received LFS, it's not responding to requests for tuple space (_StateItems_). After restart it works as expected.

[2] CPU, memory
image

[2] Direct buffer memory
image

### Testing setup 1

~Trie traversal is slow when number of records is bigger then 3,000 and it's rising faster then linear.~ Resolved in PR #3099.

Performance of LFS should be tested with this configuration.

RNode Docker image for all nodes: [`rchain/rnode:v0.9.26-alpha`](https://hub.docker.com/layers/rchain/rnode/v0.9.26-alpha/images/sha256-8dee109eb6bb377a8912111e09489ea105f926eb36d3d6a53e5195738d170d60?context=explore)

1. Observer node with full state
2. Observer node with empty state (bootstrap from `1.`)
3. Observer node with empty state (bootstrap from `2.`)

~Observer nodes with empty state (2. and 3.) must have higher limit for direct buffer memory to 3GB.
`-XX:MaxDirectMemorySize=3g`~
Memory leak with direct buffer is fixed in transport layer #3239 so now limit can be much lower.
`-XX:MaxDirectMemorySize=200m`

First operation is to test syncing of LFS from the full node. When this is done second syncing should use this node with _trimmed_ state as a source to sync the third node.

| Operation | Expected    | Measured
| ---       | ---         | ---
| 1. -> 2.  | 4-6 hours   | 7.5 hours (one machine + 1 active nodes)<br>11 hours (one machine + 4 active nodes)
| 2. -> 3.  | 1-3 hours   | 4 hours (one machine + 4 active nodes)

tgrospic:
## Overview
This PR introduces validation of received state for Last Finalized Block.

We found that hashing of history and cold store is not done on the whole binary value that is stored but only on inner class. This forced to do additional work to verify the content (decoding/encoding). To make thing faster this work is parallelized so that without proper fix we can at least get more speed with stronger machine.


casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala
```diff
@@ -88,6 +89,9 @@ object LastFinalizedStateTupleSpaceRequester {
         requestQueue: Queue[F, Boolean]
     ): Stream[F, Boolean] = {
 
+      // Get importer
+      val importer = RSpaceStateManager[F].importer
+
       def broadcastStreams(ids: Seq[StatePartPath]) = {
```
nzpr, line 94: Not related to to this particular PR, but general observation. This `StatePartPath` made me a bit confused about its meaning. I get that this is a starting point from where chunk of `pageSize` is read and sent as a StoreItem.
Just a suggestion - maybe there is another name for that ? `TuplespacePointer`, `StoreItemStartHash`?

tgrospic: It's definitely a path in the tree so it makes sense to have Path suffix. It's not path in the tuple space because tuple space is "encoded" on top of this tree.
_Start_ in the name can be misleading because state chunk have start and end path. Do you have some other suggestions? :)

NOTE: I'm saying a _tree_ and not a _trie_ because path is always referring to one _tree_ in a whole history _trie_.

nzpr: Ok, now its clear :) Though not completely. You are saying about end path - where is it encoded? I was thinking `StatePartPath` is start and pageSize defines end.
Would be great to have a description for this type https://github.com/rchain/rchain/blob/1f783b7762e46a7ef4fd78df8988307536df3090/casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala#L31
Why Is Seq here? Is it to enable to request multiple state chunks?
UPD. Oh, I see, its path in the tree! But whats `Option[Byte]`?

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala
```diff
@@ -138,22 +142,40 @@ object LastFinalizedStateTupleSpaceRequester {
...
+                val historyItemsWithByteVector = historyItems.map {
+                  case (k, v) => (k, ByteVector(v.toByteArray))
+                }
+                val dataItemsWithByteVector = dataItems.map {
+                  case (k, v) => (k, ByteVector(v.toByteArray))
+                }
+
                 for {
+                  // Validate received tuple space items
+                  _ <- Lib.time("VALIDATE STATE CHUNK")(
```
zsluedem: Just curious that is there a plan to remove this one?

zsluedem (approved):
LGTM

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
```diff
@@ -39,7 +44,8 @@ object GrpcTransportReceiver {
       buffersMap: Ref[Task, Map[PeerNode, Deferred[Task, MessageBuffers]]],
       messageHandlers: MessageHandlers,
       tempFolder: Path,
-      parallelism: Int
+      parallelism: Int,
+      chunks: TrieMap[String, Array[Byte]]
```
nzpr: We might need to put a constraint on block size or this Map can potentially exceed Heap size

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala
tgrospic: Yes, but before we have easy diagnostics setup, I'll transform this to Log output.

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala
tgrospic: This optional Byte is the _index_ of the child node, if trie node is the `PointerBlock`. It gives the path enough structure to index the whole trie in depth first traversal.
It's used as offset to position the child node.
https://github.com/rchain/rchain/blob/416936482a9632bf3732cfe7fcee46c603e028d3/rspace/src/main/scala/coop/rchain/rspace/state/RSpaceExporter.scala#L119-L122

node/src/test/scala/coop/rchain/state/TrieExporterSpec.scala
```diff
@@ -1,51 +0,0 @@
-package coop.rchain.state
```
tgrospic, line 1: This was only for debugging, I'll create RSpace exporter tests in separate PR.

casper/src/main/scala/coop/rchain/casper/engine/LastFinalizedStateTupleSpaceRequester.scala
tgrospic: I transformed this to [Stopwatch](https://github.com/rchain/rchain/blob/52aa8bc5/shared/src/main/scala/coop/rchain/shared/Stopwatch.scala) which can measure duration of some block of code.

comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportReceiver.scala
tgrospic: Agree, we also need to investigate how to get more info about the message before it's completely received. Streams can have some header announcing the length.

node/src/main/resources/logback.xml
```diff
@@ -25,7 +25,6 @@
   </appender>
 
   <logger name="coop.rchain.shared.EventLogger" level="OFF" />
-  <logger name="coop.rchain.rspace" level="warn" />
```
tgrospic, line 28: There is no log messages in RSpace so it's safe to be removed.

rspace/src/main/scala/coop/rchain/rspace/state/RSpaceImporter.scala
```diff
@@ -1,8 +1,169 @@
...
+    // - validate trie hash to match decoded trie hash
+    def tries: F[List[(Blake2b256Hash, Trie)]] = historyItems.toList traverse {
+      case (hash, trieBytes) =>
+        val trie = decodeTrie(trieBytes)
+        // TODO: this is strange that we need to encode Trie again to get the hash.
+        //  - the reason is that hash is not calculated on result of `Trie.codecTrie.encode`
+        //    but on each case of Trie instance (see `Trie.hash`)
+        //  - this adds substantial time for validation e.g. 20k records 450ms (with encoding 2.4sec)
+        // https://github.com/rchain/rchain/blob/4dd216a7/rspace/src/main/scala/coop/rchain/rspace/history/HistoryStore.scala#L25-L26
+        val trieHash = Trie.hash(trie)
```
zsluedem, line 79: I think you can get the hash by `val trieHash = Blake2b256Hash.create(trieBytes.bits.drop(2L).toByteVector)`
But I tried and it doesn't  turns out to be the same always.

It might related to the byte-Bit-convert situation like below.
```
val b = BitVector(something)
BitVector(b.toByteVector) != b  (sometimes)
```

tgrospic: You are right, this is direct manipulation of serialized data but in this situation it can be justified.

I'll try to change that later as part of additional optimizations with more testing. I also noticed comparison of BitVector's is not consistent but conversion to raw bytes shouldn't be a problem.

nzpr (approved):
LGTM

PR #3159, opened 2020-10-03
Internalise enqueued blocks set into Casper trait

nzpr:
When new block A comes in, but Casper is busy processing another block, block A is put in queue and added later.
Currently node recognises enqueued blocks via BlockRetriever state. But sometimes this does not work, as it has been seen recently when integration test failed. Investigations had not led to any solution. So, this PR introduces more reliable way to have set of blocks enqueued as a structure internal to Casper. Which is better in logical sense as well.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -298,12 +298,12 @@ class Running[F[_]: Concurrent: BlockStore: CasperBufferStorage: BlockRetriever:
     for {
       // This long chain of ifM created to minimise computation for ignore check and provide
       // readable output for each ignore reason
-      received <- BlockRetriever[F].received(hash)
+      received <- BlockRetriever[F].isReceived(hash)
       r <- casper.getBlocksInProcessing
             .map(_.contains(hash))
             .ifM(
               IgnoreCasperMessageStatus(doIgnore = true, BlockIsInProcessing).pure[F],
-              BlockRetriever[F].getEnqueuedToCasper
+              casper.getBlocksEnqueued
                 .map(_.contains(hash))
                 .ifM(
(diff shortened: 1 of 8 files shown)


casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -298,13 +298,13 @@ class Running[F[_]: Concurrent: BlockStore: CasperBufferStorage: BlockRetriever:
...
-      r <- casper.getBlocksInProcessing
-            .map(_.contains(hash))
+      received <- BlockRetriever[F].isReceived(hash)
+      r <- casper.getBlockProcessingState
+            .map(_.processing.contains(hash))
             .ifM(
               IgnoreCasperMessageStatus(doIgnore = true, BlockIsInProcessing).pure[F],
-              BlockRetriever[F].getEnqueuedToCasper
-                .map(_.contains(hash))
+              casper.getBlockProcessingState
```
zsluedem: maybe it makes more sense that only call `casper.getBlockProcessingState` once and reused the state.
I am wondering if it is possible that the block is in the `enqueued` first then in `processing` but the double get getBlockProcessingState would skip these situation.

nzpr: @zsluedem nice catch! This makes sense. Though this is not really harmful - purpose of this method if to lower down the unnecessary load. Changes pushed.

tgrospic (approved):
LGTM

zsluedem (changes requested):
LGTM

zsluedem (approved):

zsluedem:
It seems that  we are trying to fix the problem(I am actually not sure if they are the same problem) for serveal times. Maybe it's worth to create the failing testcase for this situation.

nzpr:
bors merge

PR #3167, opened 2020-10-07
Update README.md link to live api

jimscarver:

### DIFF
--- a/docs/rnode-api/README.md
+++ b/docs/rnode-api/README.md
@@ -1,4 +1,5 @@
 # RNode API
+[Use it live!](https://developer.rchain.coop/rnode-api)
 
 `rnode` communicates primarily using gRPC but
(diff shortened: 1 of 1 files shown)


9rb (approved):
lgtm

nzpr:
I'm for removing documentation for end users from this repo. If any, documentation in this repo should be related only to protocol. Saving particular URL here is not a good idea IMO. And we should not point end users to this repo as well.

tgrospic (approved):
Thanks Jim!

tgrospic:
@nzpr I can agree with you but currently it's better to have this info on multiple places. Let's create more focused and systematic way documentation can be kept up to date. I also thinking on maintaining the services mentioned in documentation.

I don't see a reason why this repo should not contain documentation for clients if this is generated from central place and pushed here automatically as PR (for example).

PR #3175, opened 2020-10-15
Cleaning up block processing logic

nzpr:
There is two filter cases missing in dependencies set construction - it should not include blocks that are being processed by Casper at the moment or enqueued to Casper.

This PR turned into some bigger attempt to refactor, bugfix and improve how block processing is handled.

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3179, opened 2020-10-20
Do not fetch blocks in processing

nzpr:
There were tho flaws: 
1. node has been requesting blocks in processing and enqueued (and it should not). This lead to unnecessary requests, and misleading logs.
2. node has not been not adding relation between block and its missing dependencies that are in processing. As a consequence, after dependency in processing is added, adding of block has not been triggered.

First attempt <https://github.com/rchain/rchain/pull/3175> is too invasive (please see closing comment), so this is created instead.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -538,20 +538,28 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
                                equivocations.flatMap(_.equivocationDetectedBlockHashes)
                              }
                            }
-      depsInEqTracker = allDeps.filter(equivocationHashes.contains)
+      depsInEqTracker  = allDeps.filter(equivocationHashes.contains)
+      depsInProcessing <- blockProcessingState.get.map(_.processing)
+      depsEnqueued     <- blockProcessingState.get.map(_.enqueued)
 
-      missingDeps = allDeps filterNot (
-          d => depsInDag.contains(d) || depsInBuffer.contains(d) || depsInEqTracker.contains(d)
-      )
-      _ <- (missingDeps ++ depsInBuffer).traverse(casperBuffer.addRelation(_, b.blockHash))
+      depsToFetch = allDeps filterNot (
+          d =>
+            depsInProcessing.contains(d) || depsEnqueued.contains(d) ||
+              depsInDag.contains(d) || depsInBuffer.contains(d) || depsInEqTracker.contains(d)
+        )
+      // Commit to Casper buffer all relations with deps that have not been validated yet
+      _ <- (allDeps diff depsInDag diff depsInEqTracker)
+            .traverse(casperBuffer.addRelation(_, b.blockHash))
       _ <- BlockRetriever[F].ackInCasper(b.blockHash)
       _ <- Log[F].info(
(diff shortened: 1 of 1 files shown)


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -538,17 +538,23 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
                                equivocations.flatMap(_.equivocationDetectedBlockHashes)
                              }
                            }
-      depsInEqTracker = allDeps.filter(equivocationHashes.contains)
+      depsInEqTracker  = allDeps.filter(equivocationHashes.contains)
+      depsInProcessing <- blockProcessingState.get.map(_.processing)
+      depsEnqueued     <- blockProcessingState.get.map(_.enqueued)
```
nzpr, line 543: I must note that this `enqueued` is introduced prematurely. In current dev branch casper is locked, and we try to acquire lock only when all dependencies are in DAG, which means in single parent version `enqueued` should always empty. But this `enqueued` set has been introduced in this PR <https://github.com/rchain/rchain/pull/3159> to make sure no race happen.
Eventually, in block merge imo we need to decouple block creation, DAG builder, block requester, and Casper into separate things (hopefully using fs2 streams).

tgrospic (approved):
LGTM

zsluedem (commented):
It seems that we are fixing this problem for long. 
I think the way we use `Ref` here can still cause the block state is not consistent sometimes. I think we might better use `MVar` here to make sure state is consistent here or some other tricks to like combine `Ref` and `deffer` to make state consistent like @tgrospic mentioned before. I am not sure if I am right on it.

zsluedem (approved):

nzpr:
@zsluedem I don't think this particular issue is related to some race conditions. Here Ref is used atomicaly, but how we connect DAG, casper buffer, blocks enqueued is q bit cumbersome. I hope to clean it all up in block merge branch.

nzpr:
bors merge

zsluedem:
> @zsluedem I don't think this particular issue is related to some race conditions. Here Ref is used atomicaly, but how we connect DAG, casper buffer, blocks enqueued is q bit cumbersome. I hope to clean it all up in block merge branch.


-----------------------------------------------------> Timeline

|                     | time1                       |  time2                          |  time3|
| ------------ | ---------------------- |--------------------------|---------|
| Thread1  | Ref.get  ->  state1 | Ref.update(state2) |  do nothing |
| Thread2  | Ref.get   -> state1 | do nothing                | doSomething with state which is get in time1|

In `time3`,  Thread2 is doing something with **outdated** state which in our cases the block can be in `blockProcessingState` or `in `Dag`


@nzpr @tgrospic Is the case above possible?

tgrospic:
@zsluedem I agree that pattern is suspicious, even more in general when state is just read with `Ref.get` and some effect is executed, it's a question what will happen when multiple threads enter the same code path and execute the effects multiple times.
From that, anything between `Ref.get` and `Ref.update` can happen multiple times in multi threaded execution.

Another example, between these two _gets_ state can change and result can be inconsistent. I don't think it's a big problem here but needs to be refactored.
https://github.com/rchain/rchain/blob/ab5a79c886db6826c8ccb1c57e72984d0a9f54f9/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L542-L543

PR #3189, opened 2020-10-24
Make Signed equals method count on signatures

nzpr:
We issue InvalidRepeatDeploy based on duplicate signature. But we are filtering deploys seen in the past using object equality. 
https://github.com/rchain/rchain/blob/dac6dab1e7e21bbdd1b910a41adcdd591ba90349/casper/src/main/scala/coop/rchain/casper/BlockCreator.scala#L175
This PR overrides Signed class equal method to check if signatures are equal.


tgrospic:
There is one more place where equality for `Signed` type is defined.
https://github.com/rchain/rchain/blob/d7cda5476778f19ba62a92414794fee3bd629d73/models/src/main/scala/coop/rchain/models/EqualsM.scala#L116-L121

Issue #3001, opened 2020-07-19
Refactoring and cleanup


tgrospic:
## Overview
This issue is an epic to collect a list of proposals for code refactoring, file rearrangement or cleanups with the purpose to unify the design and strategy in organization the source code in the project.

### Restriction to type of changes
The important restriction is that changes should not impact functionality. Exceptions must be well justified.

### How to use this issue
Proposals will be collected in the description of the issue and updated when new items are added or status of existing is changed. 
From this issue multiple PR's will be spawn and linked to the issue.

### Proposals
- [ ] __Create _SourceEnclosing_ helper and remove LogSource and Metrics.Source__
  With simple macro we can eliminate creating LogSource on many places. We already have macro introduced here https://github.com/rchain/rchain/pull/1670#discussion_r227012153.
  This can be improved without the need to have implicit parameter for logger functions although with Metrics it's little more complex.
  Little complication is for shared project itself because macros must be compiled separately. One solution is to create subproject.
  https://www.scala-sbt.org/1.x/docs/Macro-Projects.html
  
- [ ] __Remove _metric_ cache, Kamon already keep internal map__
  https://github.com/rchain/rchain/blob/66b93e3/node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala#L117
  See Kamon source _MetricRegistry#lookupMetric_. Maybe _timer_ should be in bracket to ensure stop (?).

- [ ] __Replace custom `Time` with cats `Timer`__
  We can define custom [Clock](https://typelevel.org/cats-effect/datatypes/clock.html) and use cats `Timer` object. Time is introduced here https://github.com/rchain/rchain/pull/1558.
  Examples of `Timer` in tests and virtual (logical) time.
  https://blog.softwaremill.com/time-traveling-in-tests-with-cats-effect-b22084f6a89
  https://monix.io/api/3.0/monix/execution/schedulers/TestScheduler.html
    
- [ ] __Mock of Event#publish to be referentially transparent__
  https://github.com/rchain/rchain/blob/b5ba61/comm/src/test/scala/coop/rchain/p2p/EffectsTestInstances.scala#L129
  Move mock from _comm_ project (?).

- [ ] __Define _syntax_ trait for all extensions in _models_ project for easy import__
  Implicit class should extend `AnyVal` to prevent boxing.
  https://github.com/rchain/rchain/blob/003e2a8e/models/src/main/scala/coop/rchain/models/BlockHash.scala#L10-L12
[...]

PR #3193, opened 2020-10-26
Refactor runtime based on lfs

zsluedem:
## Overview
Refactor runtime based on lfs

@tgrospic @nzpr 

Some review advise:
1. the first commit https://github.com/rchain/rchain/pull/3193/commits/1dc72d8bb19cab78bbfa9fdcace7814877621e35 which is not actually necessary but I still kept it because of history. The `InterpreterSyntax` is removed in https://github.com/rchain/rchain/pull/3193/commits/d27f101b7763af6d2525b99462e50b435404dbea
2. The second commit https://github.com/rchain/rchain/pull/3193/commits/2e7b273d4266e9bccc3c3d32eba611aec351a17c is actually some clean up which only moves code around.

3. The most important part is `rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala` the new Runtime trait. 
4. After introducing the trait, I can move all the methods in the runtimeManager into a syntax of a  Runtime which the syntax in `rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala`

5. After renaming the runtime, another thing I need to fix is to re-implement all the codes where initial the runtime and the runtimeManager


nzpr (approved):
LGTM

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperReportingSpec.scala
```diff
@@ -31,19 +33,43 @@ class MultiParentCasperReportingSpec extends FlatSpec with Matchers with Inspect
...
+        reportingCommEventsNum = trace match {
+          case Right(value) =>
+            value.head._2.foldLeft(0)(
+              (sum, e) =>
+                sum + e
+                  .filter(
+                    p =>
+                      p match {
+                        case ReportingComm(_, _) => true
+                        case _                   => false
```
tgrospic: These functions can be defined in short form, just with cases.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -17,129 +17,86 @@ final case class EvaluateResult(cost: Cost, errors: Vector[InterpreterError]) {
...
       term: String,
       initialPhlo: Cost,
       normalizerEnv: Map[String, Par]
-  ): F[EvaluateResult]
+  )(implicit rand: Blake2b512Random): F[EvaluateResult]
+}
 
-  def injAttempt(
+class InterpreterImpl[F[_]: Sync](implicit C: _cost[F]) extends Interpreter[F] {
+  case class ParserError(parseError: InterpreterError) extends Throwable
```
tgrospic, line 31: Please return comment and add some empty lines for this error definition.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -0,0 +1,687 @@
...
+    val space: RhoISpace[F],
+    val cost: _cost[F],
+    val blockDataRef: Ref[F, BlockData],
+    val invalidBlocksParam: InvalidBlocks[F]
+) extends RhoRuntime[F] {
+  private val emptyContinuation = TaggedContinuation()
+  def close: F[Unit] =
+    for {
+      _ <- space.close()
+    } yield ()
```
tgrospic: `for` can be removed for only one expression. Similar later in _checkpoint_ functions.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -0,0 +1,687 @@
...
+    *                Normally you can just
+    *                use [[coop.rchain.rholang.interpreter.accounting.noOpCostLog]]
+    * @return
+    */
+  def createRhoRuntime[F[_]: Concurrent: Log: Metrics: Span: Parallel](
+      rspace: RhoISpace[F],
+      extraSystemProcesses: Seq[Definition[F]] = Seq.empty,
+      initRegistry: Boolean = true
+  )(implicit costLog: FunctorTell[F, Chain[Cost]]): F[RhoRuntime[F]] = {
+    implicit val P = Parallel[F]
```
tgrospic: This implicit definition should not be needed because it's present in function signature. The same for later functions.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala
```diff
@@ -0,0 +1,57 @@
+package coop.rchain.rholang.interpreter
+
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.crypto.hash.Blake2b512Random
+import coop.rchain.models.Par
+import coop.rchain.rholang.interpreter.accounting.Cost
+
+trait RhoRuntimeSyntax {
+  implicit final def syntaxRuntime[F[_]: Sync](
```
tgrospic: I was using naming as in _cats_, and this would be `rholangSyntaxRhoRuntime`.

https://github.com/typelevel/cats/blob/51ac3ef/core/src/main/scala/cats/syntax/monad.scala#L5

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala
```diff
@@ -0,0 +1,57 @@
...
+      runtime: RhoRuntime[F]
+  ): RhoRuntimeOps[F] =
+    new RhoRuntimeOps[F](runtime)
+
+//  implicit final def syntaxReplayRuntime[F[_]: Sync](
+  //      runtime: ReplayRhoRuntime[F]
+  //  ): ReplayRhoRuntimeOps[F] =
+  //    new ReplayRhoRuntimeOps[F](runtime)
```
tgrospic: Leftover.

rholang/src/main/scala/coop/rchain/rholang/interpreter/package.scala
```diff
@@ -1,11 +1,14 @@
...
 
 package object interpreter {
 
   type _error[F[_]] = Bracket[F, Throwable]
 
   def _error[F[_]](implicit ev: _error[F]): _error[F] = ev
+  object syntax extends AllSyntaxRholang
 
 }
+trait AllSyntaxRholang extends RhoRuntimeSyntax
```
tgrospic, line 14: I'm concerned how developers can navigate thru different _syntax_ definitions and it make sense that each project has its own _all_ definition.
So these definitions are more for `rholang/package.scala` file.
Import is then `coop.rchain.rholang.syntax._` similar to e.g. casper `coop.rchain.casper.syntax._`.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -32,25 +36,36 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
...
+        implicit val c: FunctorTell[Task, Chain[Cost]] = costLog
+        Resource.make(
+          RhoRuntime
+            .createRhoRuntime[Task](space, Seq.empty, false)(
+              implicitly,
+              logF,
+              metricsEff,
+              noopSpan,
+              parallel,
+              c
```
tgrospic: Can this be simplified without writing implicit arguments? And _parallel_ should then already be in the scope.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -64,40 +79,32 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
     implicit val logF: Log[Task]           = new Log.NOPLog[Task]
     implicit val metricsEff: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
     implicit val noopSpan: Span[Task]      = NoopSpan[Task]()
-    implicit val ms: Metrics.Source        = Metrics.BaseSource
+    implicit val parallel: Parallel[Task]  = new CatsParallelForTask()
```
tgrospic: Implicit Parallel for Task should automatically be in the scope.

tgrospic (commented):
Very big refactoring but very nice! Only few minor comments.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: Another costLog https://github.com/rchain/rchain/blob/dev/rholang/src/main/scala/coop/rchain/rholang/interpreter/accounting/package.scala#L54-L58 is in the package scope `coop.rchain.rholang.interpreter.accounting`. It would cause `ambiguous implicit values` if I don't explicit specify the value.

rholang/src/main/scala/coop/rchain/rholang/interpreter/package.scala
zsluedem: Ok. After some tries, I understand why I put the syntax in `coop.rchain.rholang.interpreter.syntax._` way at first because there are auto-generated java codes by bnfc in `coop.rchain.rholang.interpreter.syntax` package.

I think the syntax meaning for these two things are different and ambiguous so I put it in the `interpreter` package.

Do you still think I should put it in `coop.rchain.rholang.syntax._`this way?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -0,0 +1,677 @@
...
+    for {
+      softCheckpoint <- space.createSoftCheckpoint()
+    } yield softCheckpoint
+
+  override def revertToSoftCheckpoint(
+      softCheckpoint: SoftCheckpoint[Name, BindPattern, ListParWithRandom, TaggedContinuation]
+  ): F[Unit] =
+    for {
+      _ <- space.revertToSoftCheckpoint(softCheckpoint)
+    } yield ()
```
tgrospic: These three functions have redundant _for_ comprehension.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -0,0 +1,677 @@
...
+    )
+  )
+
+  def setupReducer[F[_]: Concurrent: Log: Metrics: Span](
+      chargingRSpace: RhoTuplespace[F],
+      blockDataRef: Ref[F, BlockData],
+      invalidBlocks: InvalidBlocks[F],
+      extraSystemProcesses: Seq[Definition[F]],
+      urnMap: Map[String, Par]
+  )(implicit cost: _cost[F], P: Parallel[F]): Reduce[F] = {
```
tgrospic: `F` is a parameter in this function so implicits for it can be defined together with other implicits as [Context bound](https://docs.scala-lang.org/tutorials/FAQ/context-bounds.html) parameters. Similar is in `createRhoEnv` function.
```scala
def setupReducer[F[_]: Concurrent: Parallel: _cost: Log: Metrics: Span]
```

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
tgrospic: One way to resolve this is to have new implicit value with the same name which would shadow previous value. This solution seems good enough in tests.
```scala
implicit val noOpCostLog: FunctorTell[Task, Chain[Cost]] = costLog
```
In that situation, it makes sense to rename `noOpCostLog`.

rholang/src/main/scala/coop/rchain/rholang/interpreter/package.scala
tgrospic: I see, this is unfortunate. Maybe we could rename generated namespace to _parser_ instead of _syntax_.
https://github.com/rchain/rchain/blob/116c21bf647bc634251b13bd8021487663b560fb/project/BNFC.scala#L95

Renaming does not involve to many files but better to do this on dev branch and with this PR. Please just add this task as part of cleanup issue #3001.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: Cool. I didn't know that.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -79,270 +58,75 @@ trait RuntimeManager[F[_]] {
   def getContinuation(hash: StateHash)(
       channels: Seq[Par]
   ): F[Seq[(Seq[BindPattern], Par)]]
-  def emptyStateHash: StateHash
-  def withRuntimeLock[A](f: Runtime[F] => F[A]): F[A]
+  def emptyStateHash: F[StateHash]
```
tgrospic: Empty state hash is constant value, is there a reason to have it on a trait and not as a value on companion object?

rspace/src/main/scala/coop/rchain/rspace/RSpace.scala
```diff
@@ -308,10 +334,16 @@ object RSpace {
...
       par: Parallel[F]
   ): F[ISpace[F, C, P, A, K]] =
-    setUp[F, C, P, A, K](dataDir, mapSize, branch).map {
-      case (historyReader, store) =>
-        new RSpace[F, C, P, A, K](historyReader, AtomicAny(store), branch)
-    }
+    for {
+      setup                  <- setUp[F, C, P, A, K](dataDir, mapSize, branch)
+      (historyReader, store) = setup
+      hotStore               <- HotStore.empty(historyReader)(sk.toCodec, concurrent)
```
tgrospic: Can we just use `store`, created and returned from setup? The same in _createReplay_ function.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
zsluedem: Ok. My idea is that this pr is only refactoring runtime but not `runtimeManager`. I didn't change traits on runtime manager on purpose. Actually, most of the methods in the runtimeManager trait are not really neccessary right now.

I want to refactor this runtimeManager on another pr. Ideally, my idea is that there would be only `withRuntime` and `withReplayRuntime` left in the manager and each `withRuntime` would spawn a new runtime and each runtime can be run in parallel, which is what we discussed long before.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -220,338 +163,97 @@ object ReportingCasper {
...
+          startHash: StateHash,
+          terms: Seq[ProcessedDeploy],
+          isGenesis: Boolean
+      ): F[Either[ReplayFailure, List[(ProcessedDeploy, Seq[Seq[ReportingEvent]])]]] =
+        (for {
+          _ <- EitherT.right(runtime.reset(Blake2b256Hash.fromByteString(startHash)))
+          res <- EitherT.right(terms.toList.traverse { term =>
+                  for {
+                    rd <- runtime
+                           .replayDeployE(isGenesis)(term)
```
tgrospic: Does behavior of replay changed to not replay cost accounting? Or genesis parameter is used differently?

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -220,338 +163,97 @@ object ReportingCasper {
...
-                              }
-                              .semiflatMap(
-                                _ => runtime.reportingSpace.getReport
-                              )
-                          }.value
+    def checkCreateDataDir: F[Unit] =
+      for {
+        notexists <- Sync[F].delay(Files.notExists(dataDir))
+        _ <- if (notexists) Sync[F].delay(Files.createDirectories(dataDir)) >> ().pure[F]
+            else ().pure[F]
```
tgrospic: When result is _unit_ it can be simplified.
```scala
Sync[F].delay(Files.createDirectories(dataDir)).whenA(notexists)
```

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
tgrospic: But now to get `emptyStateHash` runtime must be taken from MVar which doesn't make sense because it's just the constant value, which also doesn't have to be wrapped in `F`. So it's not really a big change to move it to companion object for easy use.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
zsluedem: Good catch~!

tgrospic (approved):
Great work!

zsluedem:
bors merge

zsluedem:
bors merge

zsluedem:
bors merge

PR #3201, opened 2020-11-05
Block processing refactor + casper refactor

### LINKED ISSUE #3248: Remove casper lock
Now Casper is single threaded, which not allows several blocks to be processed in parallel.

nzpr:
Refactor of block processing in preparation to block merge.


nzpr:
I finally wrapped up block processing refactor work to compilable state, tried to make as little changes as possible. Its still not complete, but @tgrospic @zsluedem would be great if you take a look.

nzpr:
Unit tests are still not in good shape, I'll try to repair ones that still make sense, and comment out ones that are not. But PR code should not change anymore, integration tests completes, so ready for review.

node/src/main/scala/coop/rchain/node/instances/ProposerInstance.scala
```diff
@@ -0,0 +1,80 @@
...
+    // max number of concurrent attempts to propose. Actual propose can happen only one at a time, but clients
+    // are free to make propose attempt. In that case proposeID returned will be None.
+    val maxConcurrentAttempts = 100
+
+    val out = Stream
+    // propose permit
+      .eval(for {
+        lock    <- Semaphore[F](1)
+        trigger <- MVar[F].of()
+      } yield (lock, trigger))
```
nzpr: Please check this recent change - I've added this trigger to make sure no propose events triggered are left without following propose attempt.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -88,85 +92,66 @@ object BlockAPI {
...
+          } yield r,
+        default = logWarn("Failure: casper instance is not available.")
       )
     )
   }
 
+  // Get result of the propose
+  def getProposeResult[F[_]: Concurrent: Log](
+      proposerState: Ref[F, ProposerState[F]]
+  ): F[ApiErr[String]] =
```
tgrospic, line 122: _Ref_ is not really used here so accepting only `ProposerState[F]` is enough. This also say that argument will only be used for read.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -88,85 +92,66 @@ object BlockAPI {
...
+    for {
+      pr <- proposerState.get.map(_.currProposeResult)
+      r <- pr match {
+            // return latest propose result
+            case None =>
+              for {
+                result <- proposerState.get.map(
```
tgrospic, line 129: In situations like this second _get_ call can return inconsistent result. E.g. after first _get_  `currProposeResult` is none and before second _get_ in another thread value can be updated, so the result may be valid in specific 
context but looking at the whole function it's inconsistent with the defined control flow structure.

Instead of two optional values (_product_ type) `ProposerState` can be separated in multiple classes (_sum_ type) so that pattern matching like here can be clearly expressed e.g. when both optional values missing or is is possible at all.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -699,47 +684,47 @@ object BlockAPI {
...
+//      }
+//      .merge
+
+//  private def prettyPrintUnmatchedSends[F[_]: Concurrent](
+//      casper: MultiParentCasper[F],
+//      deploys: Seq[Signed[DeployData]]
+//  ): F[String] =
+//    casper.getRuntimeManager >>= (
+//      _.withRuntimeLock(runtime => StoragePrinter.prettyPrintUnmatchedSends(deploys, runtime))
+//    )
```
tgrospic, line 727: Do we still need this code or some modification of it?

casper/src/main/scala/coop/rchain/casper/blocks/BlockProcessor.scala
```diff
@@ -0,0 +1,207 @@
...
+  // check if block should be processed
+  def checkIfOfInterest(c: Casper[F], b: BlockMessage)(
+      implicit log: Log[F]
+  ): F[Boolean] = {
+    // TODO c.dagContains does not take into account equivocation tracker
+    val alreadyProcessed  = c.dagContains(b.blockHash) ||^ c.bufferContains(b.blockHash)
+    val shardOfInterest   = c.getApprovedBlock.map(_.shardId.equalsIgnoreCase(b.shardId))
+    val versionOfInterest = c.getApprovedBlock.flatMap(g => Validate.version(b, g.header.version))
+    val oldBlock =
+      c.getApprovedBlock.flatMap(g => (ProtoUtil.blockNumber(b) < ProtoUtil.blockNumber(g)).pure[F])
```
tgrospic, line 53: No need for monad here, functor is enough `a.flatMap(x => x.pure) == a.map(x => x)`.

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -0,0 +1,181 @@
...
+        } yield validUnique
+
+      def prepareSlashingDeploys(seqNum: Int): F[Seq[SlashDeploy]] =
+        for {
+          ilm <- s.dag.invalidLatestMessages
+          // if the node is already not active as per main parent, the node won't slash once more
+          ilmFromBonded = ilm.toList.filter {
+            case (validator, _) => s.onChainState.bondsMap.getOrElse(validator, 0L) > 0L
+          }
+          // TODO: Add `slashingDeploys` to DeployStorage
```
tgrospic, line 70: I don't think this comment is relevant. Slashing deploys can always be generated when block is created so no need to store them.

tgrospic (commented):
I'm OK to merge this PR although it's not finished completely but please create a new issue with a list of tasks that are left to do.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
nzpr: Then I need to unpack this Ref inside webApi and grpcApi, imo better to have it on one place

nzpr:
Issue for tasks left unfinished in this PR, shall be completed in block merge branch
https://github.com/rchain/rchain/issues/3277

PR #3121, opened 2020-09-09
RCHAIN-3952 - fix for LMDB read operation

tgrospic:
## Overview
This PR fixes long standing bug which was occurring very rarely when HistoryStore was reading data.

This fix is to convert ByteBuffer returned from LMDB get operation before transaction is closed.

Also in this PR are additional methods on LMDBStore to get and put multiple records in one transaction which will be needed later for last finalized state feature.

### JIRA issue
https://rchain.atlassian.net/browse/RCHAIN-3952

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3203, opened 2020-11-09
Fix LMDB get operation to safely use returned buffer

tgrospic:
## Overview
This PR adds additional fixes for LMDB get operation follows PR #3121. Returned byte buffer is deallocated when LMDB transaction is closed so any use after can cause corrupted value.

LMDB `get` - ByteBuffer is returned
https://github.com/lmdbjava/lmdbjava/blob/25fe08a/src/main/java/org/lmdbjava/Dbi.java#L240
On transaction _close_, close is also called on `KeyVal` and value buffer is deallocated
https://github.com/lmdbjava/lmdbjava/blob/25fe08af/src/main/java/org/lmdbjava/KeyVal.java#L68

### JIRA issue
https://rchain.atlassian.net/browse/RCHAIN-3952

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -115,7 +115,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
   override def get(blockHash: BlockHash): F[Option[BlockMessage]] =
     lock.withPermit(
       for {
-        indexEntryBytesOpt     <- index.get(blockHash.toDirectByteBuffer)
+        indexEntryBytesOpt     <- index.get_WARNING(blockHash.toDirectByteBuffer)
         indexEntryOpt          = indexEntryBytesOpt.map(IndexEntry.load)
         maybeBlockMessageProto <- indexEntryOpt.traverse(readBlockMessage)
       } yield maybeBlockMessageProto >>= (bmp => BlockMessage.from(bmp).toOption)
--- a/shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
+++ b/shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
@@ -2,10 +2,10 @@ package coop.rchain.lmdb
 
 import java.nio.ByteBuffer
 
-import cats.implicits._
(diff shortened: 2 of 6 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -115,7 +115,7 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
   override def get(blockHash: BlockHash): F[Option[BlockMessage]] =
     lock.withPermit(
       for {
-        indexEntryBytesOpt     <- index.get(blockHash.toDirectByteBuffer)
+        indexEntryBytesOpt     <- index.get_WARNING(blockHash.toDirectByteBuffer)
```
zsluedem, line 118: why not camel case?

shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
```diff
@@ -37,10 +37,16 @@ final case class LMDBStore[F[_]: Sync](env: Env[ByteBuffer], dbi: Dbi[ByteBuffer
       withTxn(env.txnWrite)(f)
     }
 
-  def get(key: ByteBuffer): F[Option[ByteBuffer]] =
+  def get(key: ByteBuffer): F[Option[BitVector]] =
```
zsluedem, line 40: Just a question on this. 
Is it reasonable to abstract this `BitVector` to `T`? Or that's too much

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
tgrospic: It's on purpose, because this function should be removed and not used in the new code. When block store will also be stored to LMDB as we did for DAG storage this function will be removed.

shared/src/main/scala/coop/rchain/lmdb/LMDBStore.scala
tgrospic: I wanted to make minimal change on this file because it will be removed after RSpace will be moved to `KeyValueStore`.

nzpr (approved):
LGTM

zsluedem (approved):

tgrospic:
bors merge

PR #3204, opened 2020-11-12
StateMerger spec

### LINKED ISSUE #3139: Block merge conflicts detectors
## Overview
In order to merge two blocks, we need a detector to detect comm-event conflicts between the comm events in different blocks in different branches.

We got some related works before -> 
1. https://github.com/rchain/rchain/blob/c07671c4543bd16d2ae3258a1288425273e1a71b/casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala (original test for mergibilty rules)
2. https://github.com/rchain/rchain/blob/c07671c4543bd16d2ae3258a1288425273e1a71b/casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala (merge spec tests)
3. https://docs.google.com/spreadsheets/d/1pABqArF9e8HRTO9zSefp93mIVUm91avekeDgqSEw0R8/edit#gid=0 merge rules created by Mike
4. https://github.com/rchain/rchain/blob/c07671c4543bd16d2ae3258a1288425273e1a71b/casper/src/main/scala/coop/rchain/casper/TuplespaceEvent.scala (original event conflicts comparing)
5. **key conflicts function** for detect conflicts now https://github.com/rchain/rchain/blob/c07671c4543bd16d2ae3258a1288425273e1a71b/casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala#L50-L76

### Design
1. By @tgrospicthe design spec for block merging which mentions the way to resolve conflicts.  https://github.com/rchain/rchip-proposals/issues/23
2. @zsluedem mentioned in the dev standup `The conflicts detect action should happen when a node adds a block instead of when a node tries to merge two blocks`. @leithaus Said `Yes` it means the node should keep track on the conflict comm events between branches in the dag so that we need a persistent store for that

zsluedem:
## Overview
1. add `getJoins` in RSpace[F]
2. implement `ChannelStore` for storing the channel `dataHash` and `continuationHash
3. Reorganize the whole `MergeabilityRules`-> add more rules and a new way to test mergeability
4. introduce `StateMerger`and test merged states 
5. re-organized `conflictDetect` in `EstimatorHelper`

LeftOver:
1. join test cases
2. non-determined cases


rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
```diff
@@ -0,0 +1,15 @@
+package coop.rchain.rspace.merger
+
+import coop.rchain.rspace.Blake2b256Hash
+import coop.rchain.rspace.trace.{Event, Produce}
+
+import scala.language.higherKinds
+
+trait StateMerger[F[_], C, P, A, K] {
+  def merge(
+      baseState: Blake2b256Hash,
```
nzpr, line 43: what do these states mean?

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepository.scala
```diff
@@ -61,13 +71,17 @@ object HistoryRepositoryInstances {
       historyStore     = HistoryStoreInstances.historyStore[F](historyLMDBStore)
       history          = HistoryInstances.merging(currentRoot, historyStore)
       // RSpace importer/exporter / directly operates on Store (lmdb)
-      exporter = RSpaceExporterImpl[F](historyLMDBStore, coldLMDBStore, rootsLMDBStore)
-      importer = RSpaceImporterImpl[F](historyLMDBStore, coldLMDBStore, rootsLMDBStore)
+      exporter         = RSpaceExporterImpl[F](historyLMDBStore, coldLMDBStore, rootsLMDBStore)
+      importer         = RSpaceImporterImpl[F](historyLMDBStore, coldLMDBStore, rootsLMDBStore)
+      channelLMDBStore <- StoreInstances.lmdbStore(config.channelHashStore)
```
nzpr, line 76: Is this eventLog channel to real channel map? So you already made it persistent?

zsluedem: yeah

rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
zsluedem: I just put comments on the method. Does that comments make sense to you?

nzpr: @zsluedem great! would be good to add also some words about why baseState matters.

nzpr: @zsluedem maybe this trait should also return event logs that should be marked as rejected? Now client have to extract deploys from `mergingState` and make a diff with returned `eventLogs `, is that correct?

rspace/src/main/scala/coop/rchain/rspace/merger/instances/StateMergerImpl.scala
```diff
@@ -0,0 +1,322 @@
...
+      val consumeChanges = comm.map(_.consume.channelsHashes).toSet ++ eventsLogs.collect {
+        case c: Consume => c.channelsHashes
+      }.toSet
+      ChannelChanges(produceChanges, consumeChanges)
+    }
+
+    case class JoinInfo(channelHash: Blake2b256Hash, joinHash: Blake2b256Hash, joins: Seq[Seq[C]])
+    private def getJoins(
+        mainState: Blake2b256Hash,
+        mergedState: Blake2b256Hash,
```
nzpr: should this be called `mergingState` instead of `mergedState`?

rspace/src/main/scala/coop/rchain/rspace/merger/instances/StateMergerImpl.scala
```diff
@@ -0,0 +1,322 @@
...
+        consumeChangesOpt <- changedChannels.consumeChanges.toList
+                              .traverse(
+                                getContinuationChanges(_, mainState, baseState, mergingState)
+                              )
+        consumeChanges     = consumeChangesOpt.collect { case Some(c) => c }
+        joinChanges        <- getJoinChanges
+        changes            = dataChanges ++ consumeChanges ++ joinChanges
+        transformedActions = changes.map(transform)
+        historyActions     <- storeLeaves(transformedActions)
+        his                = history.reset(mainState)
```
nzpr: I'd call this `h`, its shorter, and `his` word has a meaning, so confusing a bit.

rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
```diff
@@ -0,0 +1,35 @@
+package coop.rchain.rspace.merger
+
+import coop.rchain.rspace.Blake2b256Hash
+import coop.rchain.rspace.trace.Event
+
+import scala.language.higherKinds
+
+trait StateMerger[F[_], C, P, A, K] {
```
nzpr: I believe you don't need  `C, P, A, K` types here

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -75,10 +103,125 @@ object EstimatorHelper {
...
+      }
+  }
+
+  final case class MergeChanges(
+      validDeploys: List[ProcessedDeploy],
+      rejectedDeploys: List[ProcessedDeploy],
+      validEventLogs: Seq[RSpaceEvent]
+  )
+
+  def isDeploysConflict[F[_]: Sync, C, P, A, K](
```
nzpr: This method deserves description, as its important user facing one. Also name is misleading, as it tells that Boolean is returned. From what I see in the code its more like `computeMergeChanges`. 
Also - does `left` here corresponds to `main` and `right` to `merging`? So basically, return deploys will be subset of `right`.

nzpr: Also looks like this should be a method or extension for HistoryRepository?

zsluedem: > Also - does left here corresponds to main and right to merging? So basically, return deploys will be subset of right.

 Yes.

I agree with you the this should an extension. EstimatorHelp is not a good place to hold  the codes.

rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
zsluedem: put more comments now

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
zsluedem: Already refactor a little bit.

rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
```diff
@@ -0,0 +1,48 @@
...
+    * @param baseState
+    * @param mainState
+    * @param mergingState
+    * @param eventLogs
+    * @return
+    */
+  def merge(
+      baseState: Blake2b256Hash,
+      mainState: Blake2b256Hash,
+      mergingState: Blake2b256Hash,
```
nzpr, line 45: @zsluedem also I believe this API should take a list of merging states, so List[(Blake2b256Hash, List[Event])] + main state and one base state. This Merger only merges two blocks and produces new state hash. We need to merge several blocks into mainState and then produce new hash

zsluedem: This can be done by foldLeft, I think.
I'd like keep it in a smaller case and extend it which can be more flexible.

nzpr: But then you will rebalance trie after each merge?

nzpr: Actually looks like we rebalance after each insert/delete, so nvm

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
```diff
@@ -1,699 +1,870 @@
...
-
-  def remaining(rho1: Rho, rho2: Rho): Option[Rho] = {
-    assert(matches(rho1, rho2))
-    (rho1.maybeCardinality, rho2.maybeCardinality) match {
-      case (Some(NonLinear), _) => Some(rho1)
-      case (_, Some(NonLinear)) => Some(rho2)
-      case (Some(Peek), _)      => Some(rho2)
-      case (_, Some(Peek))      => Some(rho1)
-      case _                    => None
+    def ++(obj: Any): State = obj match {
```
tgrospic: Why not accept State instead of Any type?

Maybe it makes sense to define `Semigroup[State]` or even Monoid because we can define empty State and we know that `State == empty ++ State`.
https://typelevel.org/cats/typeclasses/monoid.html

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
```diff
@@ -1,699 +1,870 @@
...
+                continuationOther.count(
+                  p =>
+                    continuationEqual(p.continuation, c.continuation) && p.patterns == c.patterns &&
+                      p.persist == c.persist && p.peeks == c.peeks
+                ) == continuations.count(
+                  p =>
+                    continuationEqual(p.continuation, c.continuation) && p.patterns == c.patterns && p.persist == c.persist && p.peeks == c.peeks
+                )
+              }
+          ) && joins.toSet == joinOther.toSet
```
tgrospic: Maybe this code would be more clear if you use few intermediate variables. But we can do this later.

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
```diff
@@ -1,699 +1,870 @@
...
+    new MergeableCase(left, right, base, mergedState, mergedState)
 
   case class Rho(
-      value: String,
-      maybePolarity: Option[Polarity] = None,
-      maybeCardinality: Option[Cardinality] = None
+      value: String
   ) {
+
+    lazy val rstate        = state.runSyncUnsafe()
```
tgrospic: Is there a reason for this to be defined as lazy? It seems that is executed only once anyway.

rspace/src/main/scala/coop/rchain/rspace/channelStore/instances/ChannelStoreImpl.scala
```diff
@@ -0,0 +1,66 @@
...
+import coop.rchain.shared.Serialize
+import scodec.Codec
+import scodec.codecs.{discriminated, uint2}
+
+object ChannelStoreImpl {
+  def apply[F[_]: Sync, C](store: Store[F], sc: Serialize[C], codecC: Codec[C]) =
+    ChannelStoreImpl(store, sc, codecC)
+
+  final case class ChannelStoreImpl[F[_]: Sync, C](
+      store: Store[F],
```
tgrospic, line 19: Using KeyValueStore should be much nicer because you don't have to think about closing and we want to remove Store eventually.
Even better with `KeyValueTypedStore[F, Blake2b256Hash, ChannelHash]` it's almost just like using a Map.
For now you need to supply codecs as it is but later we can add some "communication" to underlying codes in typed store.

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
```diff
@@ -0,0 +1,196 @@
...
+  IsConflict,
+  NonConflict
+}
+import coop.rchain.rspace.{internal, Blake2b256Hash, StableHashProvider}
+import coop.rchain.shared.Serialize
+
+import scala.language.{higherKinds, implicitConversions}
+
+trait HistoryRepositorySyntax {
+  implicit final def syntaxHistoryRepository[F[_]: Sync, C, P, K, A](
```
tgrospic, line 24: Nitpick: please prefix with the project name. :trophy:  for defining syntax!

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
```diff
@@ -0,0 +1,196 @@
...
+  ): F[Seq[internal.Datum[K]]] =
+    for {
+      maybeDataHash <- historyRepo.getChannelHash(channelHash)
+      dataHash <- maybeDataHash match {
+                   case Some(DataJoinHash(dataHash, _)) => dataHash.pure[F]
+                   case _ =>
+                     Sync[F].raiseError[Blake2b256Hash](
+                       new Exception(s"not found data hash for $channelHash in channel store")
+                     )
+                 }
```
tgrospic, line 82: With Option you can just _lift_ it to F if none case is handled as argument of `liftTo`.
```scala
for {
  maybeDataHash <- historyRepo.getChannelHash(channelHash)
  dataHash <- maybeDataHash.liftTo[F](new Exception(s"not found data hash for $channelHash in channel store"))
  ...
}
```

rspace/src/test/scala/coop/rchain/rspace/history/HistoryRepositoryGenerativeSpec.scala
```diff
@@ -57,13 +60,17 @@ class LMDBHistoryRepositoryGenerativeSpec
       emptyHistory     = HistoryInstances.merging(History.emptyRootHash, historyStore)
       exporter         = RSpaceExporterImpl[Task](historyLmdbStore, coldLmdbStore, rootsLmdbStore)
       importer         = RSpaceImporterImpl[Task](historyLmdbStore, coldLmdbStore, rootsLmdbStore)
+      channelLMDBStore <- StoreInstances.lmdbStore[Task](lmdbConfig)
+      channelStore     = new ChannelStoreImpl(channelLMDBStore, stringSerialize, codecString)
```
tgrospic, line 64: To get instance of KV typed store, in tests `InMemoryStoreManager` can be used.

rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala
```diff
@@ -38,6 +38,139 @@ object Event {
...
+      )
+
+  /**
+    * @return EventGroup contains all the events which is not volatile in the scope
+    */
+  def extractRSpaceEventGroup(events: List[Event]): EventGroup = {
+    val rspaceEvents     = events.toSet
+    val allProduceEvents = rspaceEvents.collect { case p: Produce => p }
+    val allConsumeEvents = rspaceEvents.collect { case c: Consume => c }
+    val allCommEvents    = rspaceEvents.collect { case c: COMM => c }
```
tgrospic: For better performance this can be simplified to fold only once over rspaceEvents and collect results in 3 groups.

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
zsluedem: You are talking about this whole `equal` function not just ` joins.toSet == joinOther.toSet` this line, right?

rspace/src/main/scala/coop/rchain/rspace/channelStore/instances/ChannelStoreImpl.scala
zsluedem: Another pr for refactoring this.

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
zsluedem: I still have to pattern match after `liftTo`. In that case, I think a direct pattern match would be better.

rspace/src/test/scala/coop/rchain/rspace/history/HistoryRepositoryGenerativeSpec.scala
zsluedem: see https://github.com/rchain/rchain/pull/3204#discussion_r532149581

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
tgrospic: The point is to remove pattern matching, `liftTo` is "matching" on None case with an F error as a result. So `dataHash` is the value extracted from Option.
`liftTo` is like a _break_ where code will end if option has no value.

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
```diff
@@ -129,44 +116,63 @@ final case class HistoryRepositoryImpl[F[_]: Sync: Parallel, C, P, A, K](
...
+        val key = hashJoinsChannel(d.channel, codecC).bytes
         s"${key.toHex};delete-join;0"
     }
 
-  private def transformAndStore(hotStoreActions: List[HotStoreAction]): F[List[HistoryAction]] =
+  private def transformAndStore(hotStoreActions: List[HotStoreAction]): F[List[HistoryAction]] = {
+    import cats.instances.list._
     for {
-      transformedActions <- Sync[F].delay(hotStoreActions.map(transform))
+      transformedActions <- hotStoreActions.map(transform).pure[F]
```
tgrospic: This change will brake referential transparency for this function. `hotStoreActions.map(transform)` will be invoked when `transformAndStore` is called and this is not what is expected.

With `Sync[F].delay(...)` execution will be _delayed_ until the whole F expression is executed.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/StateMergerImpl.scala
```diff
@@ -0,0 +1,322 @@
...
+    ) =
+      for {
+        joinInfo <- getJoins(mainState, mergingState, channelHash)
+        index    = joinInfo.joins.indexOf(join)
+        _ <- if (index != -1) {
+              joinMap.update { m =>
+                {
+                  val updatedJoins = removeIndex(joinInfo.joins, index)
+                  m + (joinInfo.joinHash -> updatedJoins)
+                }
```
tgrospic: Extra braces are not needed.

rspace/src/test/scala/coop/rchain/rspace/history/HistoryRepositorySpec.scala
```diff
@@ -221,6 +236,44 @@ trait InMemoryHistoryRepositoryTestBase extends InMemoryHistoryTestBase {
...
+
+    override def put(key: ByteBuffer, value: ByteBuffer): Task[Unit] = Task { data.put(key, value) }
+
+    override def put[T](keys: Seq[(Blake2b256Hash, T)], toBuffer: T => ByteBuffer): Task[Unit] =
+      Task {
+        keys.map { case (k, v) => put(k.bytes.toByteBuffer, toBuffer(v)).runSyncUnsafe() }
+        ()
+      }
+  }
+
```
tgrospic, line 280: We have in-memory implementation for Store via KeyValueStoreManager.
https://github.com/rchain/rchain/blob/4b44814895d424253366a9d7ed11dacbfa6c4205/shared/src/main/scala/coop/rchain/store/InMemoryStoreManager.scala#L9

And to get Store from KeyValueStore.
https://github.com/rchain/rchain/blob/b38985d9d147d841a4cd75fb81bdb759849b0147/rspace/src/main/scala/coop/rchain/rspace/history/StoreFactory.scala#L14

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
```diff
@@ -1,699 +1,873 @@
...
-  implicit def timeEff: LogicalTime[Effect]
-  def genesis: GenesisContext
-
-  trait TestCase {
-
-    def precondition(left: Rho*)(right: Rho*)(base: Rho*)(
-        implicit pos: source.Position
-    ): Effect[_]
-
+  implicit val intAdditionMonoid: Monoid[State] = new Monoid[State] {
```
tgrospic: The name is not correct.

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
zsluedem: I think a big lines includes all codes should be faster sometimes. But it is only in a test and intermediate variables should be more readable.

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
zsluedem: `dataHash` can be https://github.com/rchain/rchain/pull/3204/commits/6083e2fefa869fe0a057e849086c879d01bd75ff#diff-d48de69398507c3b8a43f3b7254847b4159df93be5131225e3696ea0a9d79103R16 or https://github.com/rchain/rchain/pull/3204/commits/6083e2fefa869fe0a057e849086c879d01bd75ff#diff-d48de69398507c3b8a43f3b7254847b4159df93be5131225e3696ea0a9d79103R18.
Pattern match is not only for `Option` here.

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
zsluedem: Cool. Great to know that! But how does this break `brake referential transparency for this function`?

rspace/src/test/scala/coop/rchain/rspace/history/HistoryRepositorySpec.scala
zsluedem: together with  https://github.com/rchain/rchain/pull/3204#discussion_r532149581

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
tgrospic: It's just one more _map_ which can select the value and at the end we can decide what to do with _option_. E.g. a vlue inside `DataJoinHash` could be also inside _option_ and in this case we would not need multiple pattern matching.
```scala
maybeDataHash <- historyRepo
  .getChannelHash(channelHash)
  .map(_.map { case DataJoinHash(dataHash, _) => dataHash })
```

tgrospic (approved):
I'm ok to merge this PR and fix it in feature block-merge branch.

PR #3205, opened 2020-11-14
Block store (approved block) implementation with KeyValueStore

tgrospic:
## Overview
This PR introduces replacement implementation for file based blocks database which is not reliable for corruption and restricted in concurrent access for write but also read.

This implementation is based on abstract [KeyValueStore](https://github.com/rchain/rchain/blob/a8cb29eff84/shared/src/main/scala/coop/rchain/store/KeyValueStore.scala) interface which is not tied to specific binary storage layer restricted for concurrency access.
In addition it adds compression to binary data which shows more then double reduction in file size.

We have KV store implemented with LMDB which means that blocks (and approved block) will be stored in LMDB. Performanse on LMDB is much better so even with added compression overall speed is increased. :rocket:

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
@@ -1,0 +1,161 @@
+package coop.rchain.blockstorage
+
+import java.nio.ByteBuffer
+
+import cats.effect.Sync
+import cats.syntax.all._
+import com.google.protobuf.{ByteString, CodedInputStream}
+import coop.rchain.casper.PrettyPrinter
+import coop.rchain.casper.protocol._
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.shared.ByteStringOps.RichByteString
+import coop.rchain.shared.ByteVectorOps.RichByteVector
+import coop.rchain.shared.Compression
+import coop.rchain.shared.syntax._
+import coop.rchain.store.{KeyValueStore, KeyValueStoreManager}
+import net.jpountz.lz4.{LZ4CompressorWithLength, LZ4DecompressorWithLength}
+import scodec.bits.ByteVector
+
+class KeyValueBlockStore[F[_]: Sync](
+    private val store: KeyValueStore[F],
+    private val storeApprovedBlock: KeyValueStore[F]
+) extends BlockStore[F] {
+  import KeyValueBlockStore._
+
+  private def errorBlock(hash: BlockHash)(cause: String) = BlockStoreFatalError(
+    s"Block decoding error, hash ${PrettyPrinter.buildString(hash)}. Cause: $cause"
+  )
+
+  override def get(blockHash: BlockHash): F[Option[BlockMessage]] = {
+    import cats.instances.option._
+    for {
+      // Optional serialized block from the store
+      bytes <- store.get1(blockHash.toDirectByteBuffer, ByteVector(_))
+      // Decode protobuf message / throws if fail
+      proto <- bytes.map(_.toArray).traverse(bytesToBlockProto[F])
+      block <- proto.traverse(BlockMessage.from(_).leftMap(errorBlock(blockHash)).liftTo[F])
+    } yield block
+  }
+
+  override def find(p: BlockHash => Boolean, n: Int): F[Seq[(BlockHash, BlockMessage)]] = {
+    import cats.instances.list._
+    for {
+      filteredIndex <- store.iterate { iterator =>
+                        iterator
+                          .map { case (k, v) => (ByteString.copyFrom(k), v) }
+                          .withFilter { case (key, _) => p(key) }
+                          // Decode value only for filtered elements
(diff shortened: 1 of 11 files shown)


casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -19,21 +19,30 @@ private final case class RNodeKeyValueStoreManager[F[_]: Concurrent: Log](
...
+  private val reportingEnvConfig    = LmdbEnvConfig(name = "reporting", maxEnvSize = 10 * tb)
 
-  // Database name to store instance name mapping (subfolder for LMDB store)
+  // Database name to store instance name mapping (sub-folder for LMDB store)
   // - keys with the same instance will be in one LMDB file (environment)
   private val dbInstanceMapping: Map[String, LmdbEnvConfig] = Map[String, LmdbEnvConfig](
     // Block storage
+    ("blocks", blockStorageEnvConfig),
+    // Block metadata storage
+    ("blocks-approved", dagStorageEnvConfig),
```
tgrospic, line 45: Approved block is now stored in DAG storage DB which seems more related than block store which is more like raw storage of blocks. And I don't see a reason to keep in separate DB

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -751,21 +747,41 @@ object NodeRuntime {
...
       else Span.noop[F]
+
       // Key-value store manager / manages LMDB databases
       casperStoreManager <- RNodeKeyValueStoreManager(conf.storage.dataDir)
+
+      // Block storage
+      blockStore <- {
+        implicit val kvm = casperStoreManager
+        // Check if old file based block store exists
+        val oldBlockStoreExists = blockstorePath.resolve("storage").toFile.exists
```
tgrospic, line 765: Existing nodes will continue work with old file based block store.

block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
```diff
@@ -0,0 +1,161 @@
...
+import coop.rchain.shared.ByteVectorOps.RichByteVector
+import coop.rchain.shared.Compression
+import coop.rchain.shared.syntax._
+import coop.rchain.store.{KeyValueStore, KeyValueStoreManager}
+import net.jpountz.lz4.{LZ4CompressorWithLength, LZ4DecompressorWithLength}
+import scodec.bits.ByteVector
+
+class KeyValueBlockStore[F[_]: Sync](
+    private val store: KeyValueStore[F],
+    private val storeApprovedBlock: KeyValueStore[F]
```
nzpr, line 21: It's a bit confusing to see here approved block, as you've mentioned that approved block store is now part of DAG database

shared/src/main/scala/coop/rchain/store/KeyValueStoreSyntax.scala
```diff
@@ -12,6 +16,23 @@ final class KeyValueStoreOps[F[_]](
     // KeyValueStore extensions / syntax
     private val store: KeyValueStore[F]
 ) extends AnyVal {
+  // Suffix `1` because overload is not resolved and also because it gets one record
```
zsluedem, line 19: I am curious.
Does it make sense to make traits work with single key instead of `Seq[key]` and extend it to work with `Seq[Key]` in syntax?

zsluedem (approved):
LGTM. Left a comment.

Another thing is that 
Is it necessary to roll out a tool to convert original store data to lmdb store?

block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
tgrospic: Yes, for now I didn't want to touch existing BlockStore interface because I don't think they are needed at all. They are just two instances of `KeyValueTypedStore`.

To be able to do this, interface for searching should be defined separately, something like this.
```scala
trait Searchable[F[_], A, B] {
  def find(term: A): F[Seq[B]]
}
```
We can even have defaults for `KeyValueStore` e.g. `Searchable[F, Hex[String], Array[Byte]]` that can search keys in any store as hex text.

shared/src/main/scala/coop/rchain/store/KeyValueStoreSyntax.scala
tgrospic: It's the easiest way to follow transaction semantics of underlying key-value store. So when multiple keys/pairs are supplied they will be executed atomically.
In the current `Store` this is visible, started with singe record functions it's extended to collections to handle multiple records using one transaction (only for _put_, get is used only for single record).

tgrospic:
> Is it necessary to roll out a tool to convert original store data to lmdb store?

We can create migration function but it's not necessary right now, although it would be useful for full nodes to reduce space and speed up. With LFS node will use LMDB version.

nzpr (approved):
Thats great!

tgrospic:
bors merge

PR #3208, opened 2020-11-15
NodeRuntime refactoring (abstract effect type)

tgrospic:
## Overview
This PR introduces parameterized effect type in `NodeRuntime` and propagates this to RNode main entry function.

There are exceptions where we still use monix Task directly e.g. gRPC generated services for which new `Monixable` trait is defined as a conversion from `Task ~> F` and `F ~> Task` where is needed. This allows us to have parameterized `F[_]` functions that uses Task with explicit dependency on monix which we should aim to remove in the future.

With these changes we can start building NodeRuntime dependencies in more modular and dynamic way. :rocket: 


node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -104,230 +91,196 @@ class NodeRuntime private[node] (
...
+      _             <- initPeer.fold(().pure[F])(p => kademliaStore.updateLastSeen(p))
+      nodeDiscovery = effects.nodeDiscovery(id)(Sync[F], kademliaStore, kademliaRPC)
+
+      /**
+        * We need to come up with a consistent way with folder creation. Some layers create folder on their own
+        * (if not available), others (like blockstore) relay on the structure being created for them (and will fail
+        * if it does not exist). For now this small fix should suffice, but we should unify this.
+        */
+      _ <- mkDirs(dataDir)
+      _ <- mkDirs(blockdagStoragePath)
```
tgrospic, line 192: Folder management here will be obsolete soon with everything on KV store.

nzpr (approved):
Quite big PR, I went through but its hard to understand the consequences of these changes. Though, no logic has been changed, so looks safe. Are you sure we want this in 0.9.26 and not let it have some more testing in block merge branch?

tgrospic:
It should be completely safe on dev and also it touches many files which would complicate sync with feature branches.

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
```diff
@@ -681,14 +610,89 @@ object NodeRuntime {
...
+  )(implicit scheduler: Scheduler): F[Unit] = {
+
+    /**
+      * Current implementation of Span uses ReaderT layer to hold the local state for tracing.
+      *
+      * To be able to instantiate NodeRuntime dependencies we need ReaderT implementation for each of them.
+      * If it's possible to construct FunctorK implementation like we have for Log then this can be used as a
+      * more general implementation.
+      */
+    type TaskEnv[A] = ReaderT[F, NodeCallCtx, A]
```
zsluedem, line 625: I am still a littble bit confused on this NodeCallCtx and TaskEnv.

Let me put things in my words to make sure I understand it correctly.

So currently `Span` needs this unique `TraceId` for each `Task` which is used for marking span data.
With the TaskEnv which is a ReaderT holds a function generating `TraceId`for Task Effect.

The `F ~> TaskEnv` and `TaskEnv ~> F` conversions would happened on effect `runCancelable` and `runAsync`which would generated new `TraceId`.

Does that sound correct? Or is it a clear description?

zsluedem (approved):
This is Great~!! Nice work. Left a question on `TaskEnv`

node/src/main/scala/coop/rchain/node/NodeRuntime.scala
tgrospic: ReaderT is used to store NodeCallCtx state used by Span. But this state doesn't have to be global for the whole program, it can be "scoped" so that when used inside scope it can contain different instance of the state.
In this example program will start with [initial state](https://github.com/rchain/rchain/blob/c2659e43c2/node/src/main/scala/coop/rchain/node/NodeRuntime.scala#L693) `t0`. When `Span.trace(block)` is used it will [create a reader with new instance](https://github.com/rchain/rchain/blob/bc4f21f44/node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala#L91-L96) of `NodeCallCtx` (`t1`) which will be visible only inside `block`.
```scala
t0 {
  t0.start // ApplicativeAsk.ask here returns t0
  t1 {
    t1.start
    ... // ApplicativeAsk.ask here returns t1
    t1.end
  }
  t0.end
}
```
`F ~> TaskEnv` is necessary for all NodeRuntime dependencies because we need to _lift_ them to ReaderT (as this will be `F` parameter) and `TaskEnv ~> F` is necessary to _run_ it and obtain `F` back when we want to execute it and get the result.

In a sense it's similar how with fs2 `Strem.eval(F[A]): Stream[F, A]` lifts `F` to a stream algebra and `Stream[F, A].compile.toList: F[A]` brings it back so it can be executed.

tgrospic:
bors merge

PR #3211, opened 2020-11-16
Compiler refactorings 1

JosephDenman:
## Overview
This is the first of a series of PRs to clean up the compiler.


rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/IndexMapChain.scala
```diff
@@ -0,0 +1,44 @@
...
+  * @param chain A list of bound variable maps.
+  * @tparam T The typing discipline we're enforcing.
+  */
+final case class IndexMapChain[T](chain: Vector[DeBruijnIndexMap[T]]) {
+
+  def get(name: String): Option[IndexContext[T]] = chain.head.get(name)
+
+  def find(name: String): Option[(IndexContext[T], Int)] = {
+    import cats.implicits.catsStdInstancesForVector
+    import cats.Foldable.ops.toAllFoldableOps
```
tgrospic, line 18: We now have practice to put `import cats.syntax.all._` on top of the file to have all extensions in scope and for the rest of instances just include in local scope like `import cats.instances.vector._`.
This prevents more instance conflicts then just `import cats.implicits._` but it's also less verbose then specifying individual instance.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -66,38 +65,40 @@ object GroundNormalizeMatcher {
 }
 
 object RemainderNormalizeMatcher {
-  def handleProcVar[M[_]](pv: ProcVar, knownFree: DebruijnLevelMap[VarSort])(
+  def handleProcVar[M[_]](pv: ProcVar, knownFree: DeBruijnLevelMap[VarSort])(
```
tgrospic, line 68: On multiple methods `M[_]: Sync` can be simplified without "normal" implicit, if you are for more cleanup.

tgrospic (approved):
Nice improvements!

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: That'll be in the next PR.

9rb (approved):
bors r+

PR #3223, opened 2020-11-20
Update block-merge branch to final LFS version

nzpr:
Merging latest work in dev and LFS, same branch as in https://github.com/rchain/rchain/pull/3215 used


rholang/src/test/scala/coop/rchain/rholang/InterpreterSpec.scala
```diff
@@ -85,26 +85,23 @@ class InterpreterSpec extends FlatSpec with Matchers {
             """.stripMargin
               )
 
-          contents <- storageContents(runtime)
-        } yield contents
+          tupleSpace <- runtime.space.toMap
```
zsluedem: use ` tupleSpace <-runtime.getHotChanges`

nzpr: Thanks!

zsluedem (approved):
LGTM

PR #3238, opened 2020-11-22
Block merge features

nzpr:
This PR introduces several features. Only last 4 commits are subject.
1. Ability to put dummy deploy in each block, for testing purposes.
2. Autopropose flag for node, which makes node make automatically attempt to issue block after each new block added 
3. Extend autopropose on deploy event.
4. Make runtime manager spawning new runtime for each execution (removing runtime manager lock).
5. Modifications for conflict detector (polarity based conflict detector)
6. Modifications for state merger to make it faster


tgrospic (approved):
I'm ok to merge this PR and fix it in feature block-merge branch.

nzpr:
Ok, let me make sure it is updated to all recent changes from other branches

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -126,47 +126,66 @@ object EstimatorHelper {
     * @param sc
```
zsluedem: This can be removed if you don't need

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
```diff
@@ -126,47 +126,66 @@ object EstimatorHelper {
...
+      val (prod, cons, comms) = d.deployLog
+        .map(EventConverter.toRspaceEvent)
+        .foldLeft((Seq.empty[Produce], Seq.empty[Consume], Seq.empty[COMM])) { (op, event) =>
+          event match {
+            case p: Produce => (op._1 :+ p, op._2, op._3)
+            case c: Consume => (op._1, op._2 :+ c, op._3)
+            case c: COMM    => (op._1, op._2, op._3 :+ c)
+          }
+        }
+      val freeProd = prod.filter(p => p.persistent || !comms.exists(_.produces.contains(p)))
```
zsluedem: In this case, peek would be ignored.

If a peek matched, the comm event would be like Seq(COMM(Consume(X), Produce(X)), Consume(X), Produce(X), Produce(X)), which there should a free Prod left.
But the filter ignore this situation.

zsluedem:
I don't see any strategies for detecting conflicts for `Join` in this pr. What is the idea behind this pr?

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
nzpr: I think I do not understand - why peek produces that event log? If peek = produce following consume, why event log is not `Seq(COMM(Consume(X), Produce(X)), Consume(X), Produce(X)`?

Peek is supposed to not introduce any conflicts, and it should not influence polarity of the channel. As peek = a pair of bound produce and consume, I assume its not free IOEvent.

nzpr:
Could you elaborate on what is `detecting conflicts for Join`? There are two operations that can affect any rholang channel - produce and consume. What is there about join? 

As I understand discussion about `join`, it is that just checking polarity potentially can show conflicts when in two blocks the same channel is is used in a join. And additional logic can be applied to check conditions when this can be treated as non conflicting.

zsluedem:
> Could you elaborate on what is `detecting conflicts for Join`? There are two operations that can affect any rholang channel - produce and consume. What is there about join?
> 
> As I understand discussion about `join`, it is that just checking polarity potentially can show conflicts when in two blocks the same channel is is used in a join. And additional logic can be applied to check conditions when this can be treated as non conflicting.

https://github.com/rchain/rchain/issues/3139#issuecomment-727519123 how do you solve situation like this?

casper/src/main/scala/coop/rchain/casper/EstimatorHelper.scala
zsluedem: Try Rholang like 
```
@0!(1)|for (_ <<- @0) { 0 }
```
And see what is the eventLogs.

 My question is -> Is there supposed to be a free produce in this situation?

nzpr: I think yes, there should be free produce, because it has not been consumed. It just been read (peeked), which should effectively be noop for tuplespace.

I guess I see what you mean, free produce here will be part of comm event. Right?

zsluedem: > free produce here will be part of comm event. Right?

Yes

nzpr: I think the problem here is how peek is implemented, it should not mess up with produces and consumes.

zsluedem: But sadly, this is currently how Rholang is doing now.

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -315,6 +315,15 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
     comms.tailRecM(go).map(_.toOption)
   }
 
+  def spawn: F[IReplaySpace[F, C, P, A, K]] = {
```
zsluedem, line 318: I think this method should be in `RuntimeManager`(or some other manager?).
I don't actually like spawn here. Maybe it is just personal favor. @tgrospic talked about this manager idea. I am more for that

shared/src/main/scala/coop/rchain/dag/DagOps.scala
```diff
@@ -0,0 +1,159 @@
...
+  def children(vertex: A): F[Option[Set[A]]]
+  def parents(vertex: A): F[Option[Set[A]]]
+}
+
+trait DagUpdater[F[_], A] {
+  def addEdge(child: A, parent: A): F[InMemDAG[F, A]]
+  def remove(vertex: A): F[Unit]
+}
+
+final case class InMemDAG[F[_]: Applicative, A](
```
zsluedem, line 22: unit test for this, please.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -59,41 +63,47 @@ trait RuntimeManager[F[_]] {
   def getContinuation(hash: StateHash)(
       channels: Seq[Par]
   ): F[Seq[(Seq[BindPattern], Par)]]
-  def withRuntimeLock[A](f: RhoRuntime[F] => F[A]): F[A]
-  def withReplayRuntimeLock[A](f: ReplayRhoRuntime[F] => F[A]): F[A]
+  def withRuntime[A](f: RhoRuntime[F] => F[A]): F[A]
+  def withReplayRuntime[A](f: ReplayRhoRuntime[F] => F[A]): F[A]
   // Executes deploy as user deploy with immediate rollback
   def playExploratoryDeploy(term: String, startHash: StateHash): F[Seq[Par]]
+  def getHistoryRepo: RhoHistoryRepository[F]
```
zsluedem, line 70: Like the `spawn` in rspace. This should not be concerned in runtimeManager.

zsluedem:
From my perspective, I think it makes sense to split this pr into seperate prs especially because the policy allowing failing tests exist. Then when we wind back, we got better knowledges on what kind of changes failed the tests.
Just comment.

zsluedem:
Just comment.
I still insist that leaving tests failing is a disaster.
I don't know what is the expecting behaviour I should see and what is the goal you are trying to reach because mostly I am not in you two conversations.
Anyway, as long as you two know what you are doing and make sure you cover the things you wrote, I am OK. Current review is not really important.

shared/src/main/scala/coop/rchain/store/LazyCache.scala
```diff
@@ -0,0 +1,41 @@
+package coop.rchain.store
+
+import cats.effect.Concurrent
+import cats.effect.concurrent.{Deferred, Ref}
+import cats.syntax.all._
+
+class LazyCache[F[_]: Concurrent, K, V] private[LazyCache] (
```
zsluedem: This is nice.

shared/src/main/scala/coop/rchain/dag/DagOps.scala
```diff
@@ -0,0 +1,158 @@
...
+  /**
+    * branch = tip block + all descendants through which tip is connected to base block.
+    * @return sequence of blocks inside branch
+    */
+  def computeBranch[F[_]: Monad, A](tip: A, base: A, dag: DagReader[F, A]): F[(A, Seq[A])] =
+    for {
+      // tip + all ancestors down to LFB block height
+      // traverse from tip to LFB through parents, save blocks visited
+      tipAncestors <- (Seq(tip), Seq(tip)).tailRecM[F, Seq[A]] {
+                       case (acc, blocksToProceed) =>
```
zsluedem: I just want to say this can be a performance issue if the traverse goes very long. 
The performance issue would be showed like all memory is occupied and the node is doing gc from my perspective.
This is from the experience which I did on the pr which traverse the whole dag and accumulate finalized blocks.
https://github.com/rchain/rchain/pull/3258/files#diff-74cd1658ab1ea29ab3304579f96ff1f6a6f4488b0b85b5cbeedcada721016ce7R121-R132 this is the algorithm which would cause performance issue.
I fixed that by using `Stream`(not fs Stream) like this https://github.com/rchain/rchain/pull/3263/files#diff-b2346ef03013f08f23c57bb943e5e8e9417d11517e8e448c9aaa9703c7b83ce6R26-R41 .

Below is my guess for the reason.
Because `acc` is accumulating and each iteration would create a new Seq in memory, memory would accumulate in O(n**2) way which can easily consume all the memories. Maybe data structure with `Vector` would help in this way, but I am not sure because I didn't try.

shared/src/main/scala/coop/rchain/dag/DagOps.scala
```diff
@@ -0,0 +1,158 @@
...
+                                              .map(_.flatten)
+                               // stop when base block is met
+                               next = if (nextBlocks.contains(base)) Seq.empty else nextBlocks
+                             } yield (acc ++ next, next).asLeft[Seq[A]]
+                         }
+                     }
+      // blocks from tipSet connected to base
+      branchSet <- (Seq.empty[A], Seq(base))
+                    .tailRecM[F, Seq[A]] {
+                      case (acc, blocksToProceed) =>
```
zsluedem: Same above.

shared/src/main/scala/coop/rchain/dag/DagOps.scala
nzpr: I'm not concerned too much about it, because if traverse is vey long, we have a bigger problem because finalised block is far far away.
But I'm also not sure if this is valid `Because acc is accumulating and each iteration would create a new Seq in memory, memory would accumulate in O(n**2) way which can easily consume all the memories`. Imo with tail recursion it should be fine.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -38,6 +38,37 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+    val streams = hashes.map(
+      h =>
+        fs2.Stream.eval(
+          dag.children(h).map(_.getOrElse(Set.empty).toList) >>= lookupUnsafe
+        )
+    )
+    fs2.Stream
+      .emits(streams)
+      .parJoinUnbounded
+      .mapAccumulate(List.empty[BlockMetadata])((acc, v) => (acc ++ v, Nil))
```
tgrospic, line 66: List concatenation is slower then with Vector or Array.
https://www.lihaoyi.com/post/BenchmarkingScalaCollections.html

But we can avoid sequential folding and first _select_ all children as a stream and just map loading operation in parallel.
```scala
val childStream = fs2.Stream.emits(hashes).flatMap { h =>
  fs2.Stream
    .eval(dag.children(h).map(_.getOrElse(Set.empty)))
    .flatMap(xs => fs2.Stream.fromIterator(xs.iterator))
    .parEvalMapUnorderedProcBounded(lookupUnsafe)
}
childStream.compile.toList
```

shared/src/main/scala/coop/rchain/dag/DagOps.scala
zsluedem: I am not 100% sure about the cause either.Maybe it is because my implementation.
Anyway, I just want to let you and @tgrospic know about that.

The `Stream` way is copied from the old devs which is lazy -> good for memory.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -38,6 +38,37 @@ final class BlockDagRepresentationOps[F[_]: Sync](
     dag.lookup(hash) >>= (_.liftTo(BlockDagInconsistencyError(errMsg)))
   }
 
+  def lookupUnsafe(hashes: Seq[BlockHash])(
+      implicit line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing,
```
tgrospic, line 44: Source info is not used here and it can be removed.

But even more, as mentioned in the top comment, this whole _source_ parameters are not needed anymore because monix@v3 has async stack trace so we have source reference info on errors.

tgrospic: @vb-cs maybe you can take this task to remove _source_ parameter?

I want to point you to this pattern of extension methods written as class methods. In this situation class must accept only one parameter, extend AnyVal for better performance etc. 

In Scala 3 this pattern will have syntax representation. I didn't test this but it should be something like this.
``` scala
extension [F[_]: Sync] (dag: BlockDagRepresentation[F]):
  def lookupUnsafe:
    def errMsg = s"DAG storage is missing hash ${PrettyPrinter.buildString(hash)}"
    dag.lookup(hash) >>= (_.liftTo(BlockDagInconsistencyError(errMsg)))
```

shared/src/main/scala/coop/rchain/dag/DagOps.scala
```diff
@@ -0,0 +1,158 @@
...
+    for {
+      // compute branches in parallel, sort according to size primarily and tip secondary, in reverse order
+      branches <- fs2.Stream
+                   .emits(tips.map { top =>
+                     fs2.Stream.eval(computeBranch(top, base, dag))
+                   })
+                   .parJoinUnbounded
+                   .compile
+                   .toList
+                   .map(_.sortBy(b => (b._2.size, b._1)))
```
zsluedem: The sorting rules should be a core part of the consensus protocol. I am OK to use the rule here right now. But we should probably discuss more on this(Maybe on https://github.com/rchain/rchip-proposals/issues/23 or should we start a fresh new one totally on consensus part?). @tgrospic @nzpr

shared/src/main/scala/coop/rchain/dag/DagOps.scala
```diff
@@ -0,0 +1,158 @@
...
+      // make branches non intersecting
+      (_, nonIntersectingBranches) = branches.reverse
+        .foldLeft(
+          // initial list of branches and result list of non intersecting branches
```
zsluedem, line 161: What is intersacting branches?
B4 and B5 like below?
```
    B4       B5
   /    \    /    \
  B1    B2     B3
     \      |     /
        base
```

shared/src/main/scala/coop/rchain/dag/DagOps.scala
zsluedem: We can calculate total stake on `b._2` to sort.

casper/src/main/scala/coop/rchain/casper/blocks/merger/ConflictDetectors.scala
```diff
@@ -0,0 +1,308 @@
...
+    val biCommedConsume = merge.consumesDestroyed
+      .intersect(main.consumesDestroyed)
+    val biCommedProduce = merge.producesDestroyed
+      .intersect(main.producesDestroyed)
+
+    val competeForBaseEventChecks = biCommedConsume.toList.map(
+      c =>
+        fs2.Stream
+          .eval(for {
+            baseConts <- baseContReader.get(ChannelStoreImpl.continuationKey(c.channelsHashes))
```
zsluedem: I just want to say that my old method doesn't have to read from base to know if main and merging competes the same produce(or consume), it can tell from the eventLog. But it can not work with PEAK situation. With our current implementation of peak(consume then produce),  reading base seems unavoidable.

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -0,0 +1,180 @@
...
+                           case (main, merging) =>
+                             fs2.Stream.eval(for {
+                               conflictingChans <- ConflictDetectors.findConflics(
+                                                    main,
+                                                    merging,
+                                                    baseDataReader,
+                                                    baseContReader
+                                                  )
+                               rejectedDeploys = merging.calculateRejected(conflictingChans)
+                             } yield (main + merging, rejectedDeploys))
```
zsluedem: `main + merging`. What if something in `merging` is conflict? Can you still simply add?

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -0,0 +1,180 @@
...
+    // ordering is required for computing branches procedure to be replayable
+    implicit val ord: Ordering[MergingVertex] = (x: MergingVertex, y: MergingVertex) =>
+      Base16.encode(x.stateHash.toByteArray).compare(Base16.encode(y.stateHash.toByteArray))
+
+    for {
+      baseStateReader <- RuntimeManager[F].getHistoryRepo
+                          .reset(Blake2b256Hash.fromByteString(base.stateHash))
+      branchesSorted <- DagOps.computeSortedBranches(tips, base, dag)
+      // TODO extract syntax for streams to compute in parallel but preserve order
+      branchesIndexed <- Stopwatch.time(Log[F].info(_))(
```
zsluedem, line 44: @tgrospic Is `Stopwatch` still a temporary solution or for long term now?
Why not Span?

zsluedem: I am ok for long term usage.

casper/src/test/scala/coop/rchain/casper/util/rholang/MergingBranchMergerSpec.scala
```diff
@@ -0,0 +1,469 @@
...
+  def mkTailStates(stateHash: StateHash, seqNum: Int, blockNum: Long)(
+      implicit runtimeManager: RuntimeManager[Task]
+  ) =
+    Stream
+      .emits(mkStatesSteams(stateHash, seqNum, blockNum).tail)
+      .parJoinUnbounded
+      .compile
+      .toList
+
+  "serial execution and merging" should "be equal" in effectTest {
```
zsluedem, line 138: Just want to say this is a very nice test with wisdom.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/StateMergerImpl.scala
```diff
@@ -65,220 +69,282 @@ object StateMergerImpl {
...
-                       mainJoins    <- historyRepo.getJoins(mainState, joinHash)
-                       mergingJoins <- historyRepo.getJoins(merginState, joinHash)
-                       unionJoins   = (mainJoins.toSet ++ mergingJoins.toSet).toSeq
-                       _            <- joinMap.update(_ + (joinHash -> unionJoins))
-                     } yield unionJoins
-                   case Some(j) => j.pure[F]
-                 }
-      } yield JoinInfo(channelHash, joinHash, result)
+      val changedChannels = {
+        val comm = events.collect { case c: COMM => c }
```
zsluedem, line 79: The collect for three types can simplifies as only one `collect`

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
nzpr: I think yes. This sum is a sum of indexes, so the only effect here is that sum index is more "conflicting", Just makes more possible that channel will be conflicting again on the next round of educing. But If channel is conflicting, then it will be removed any way, so thats ok.

zsluedem: From your perspective, it makes sense. I am OK with the idea you got here now. 
But I think it is also a core part of consensus (incentive part) like calculating [the order of MergingVertex](https://github.com/rchain/rchain/pull/3238#discussion_r556280427) which we can discuss later.
From the validator perspective, I am more for greedy way.

shared/src/main/scala/coop/rchain/dag/DagOps.scala
nzpr: B4  B1  B2  base
B5  B3  B2  base
these are intersecting
to merge we deduplicate them so have
B4 B1 B2 base
B5 B3

casper/src/test/scala/coop/rchain/casper/util/rholang/MergingBranchMergerSpec.scala
nzpr: kudos to @tgrospic

shared/src/main/scala/coop/rchain/dag/DagOps.scala
nzpr: @zsluedem I agree, yes, how branches are constructed should be part of a protocol, as different order lead to different hashes. Getting stake involved is an interesting idea, but this will create incentive to put deploys into validator with the most stake to get smaller reject rate, which is not what we want, imo.

nzpr:
This PR has been merged via several smaller, so closing.

PR #2951, opened 2020-06-13
Set fault tolerance as 1 for valid old blocks (temp solution) for 0.9.23LTS

nzpr:
Apply https://github.com/rchain/rchain/pull/2947 to 0.9.23LTS

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3251, opened 2020-11-30
Set finalized as true for valid old blocks (temp solution)

nzpr:
Temp solution similar to <https://github.com/rchain/rchain/pull/2951>.
Both problems are the same, so the same fix can be applied here.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -786,16 +786,25 @@ object BlockAPI {
                          s"Could not find block with hash $hash"
                            .asLeft[Boolean]
                            .pure[F]
-                       case Some(givenBlockMetadata) =>
-                         DagOperations
-                           .bfTraverseF(List(lastFinalizedBlockMetadata)) { b =>
-                             b.parents.traverse(dag.lookup).map { parentOpts =>
-                               parentOpts.flatten.distinct
-                                 .filter(_.blockNum >= givenBlockMetadata.blockNum)
+                       case Some(givenBlockMetadata) => {
+                         // TODO this is temporary solution to not calculate fault tolerance for old blocks which is costly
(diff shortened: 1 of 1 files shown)


nzpr:
As we decided, proper solution would be to have this info stored in BlockMetadata store

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -786,16 +786,25 @@ object BlockAPI {
...
-                           .bfTraverseF(List(lastFinalizedBlockMetadata)) { b =>
-                             b.parents.traverse(dag.lookup).map { parentOpts =>
-                               parentOpts.flatten.distinct
-                                 .filter(_.blockNum >= givenBlockMetadata.blockNum)
+                       case Some(givenBlockMetadata) => {
+                         // TODO this is temporary solution to not calculate fault tolerance for old blocks which is costly
+                         val oldBlock =
+                           dag.latestBlockNumber.map(_ - givenBlockMetadata.blockNum).map(_ > 100)
+                         val isInvalid           = dag.lookup(givenBlockMetadata.blockHash).map(_.get.invalid)
+                         val oldBlockIsFinalized = isInvalid.map(if (_) false else true)
```
tgrospic, line 794: This can be simplified as negation `val oldBlockIsFinalized = isInvalid.not` (or`~^(isInvalid.not`)`).

nzpr: I hope to remove this code shortly anyway, and wanted to keep it similar to piece of code in faoultTolerance :)

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

zsluedem (changes requested):
Sorry. I don't agree with that.

PR #3255, opened 2020-12-01
Store finalized blockHashes 

### LINKED ISSUE #3252: Cache fault tolerance in block metdata
## Overview
Looking for fault tolerance for old blocks is resource consuming. There should be a way to avoid unnecessary computation.

### Impact
<https://github.com/rchain/rchain/pull/3251> and <https://github.com/rchain/rchain/pull/2947> should be reverted

### Design
it is reasonable to accumulate and cache fault tolerance in block metadata

zsluedem:
## Overview
Store finalized blockHashes 

### DIFF
--- a/casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
+++ b/casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
@@ -6,8 +6,10 @@ import cats._
 import cats.implicits._
 import cats.effect.concurrent.Ref
 import cats.effect.{Concurrent, ContextShift}
+import coop.rchain.shared.syntax._
 import coop.rchain.blockstorage._
 import coop.rchain.blockstorage.casperbuffer.CasperBufferKeyValueStorage
+import coop.rchain.blockstorage.dag.codecs.codecBlockHash
 import coop.rchain.blockstorage.dag.{BlockDagRepresentation, InMemBlockDagStorage}
 import coop.rchain.blockstorage.deploy.InMemDeployStorage
 import coop.rchain.blockstorage.finality.LastFinalizedMemoryStorage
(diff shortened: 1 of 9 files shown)


nzpr:
<https://github.com/rchain/rchain/pull/3251> this should be reverted as well

zsluedem:
> #3251 this should be reverted as well

It should be overwritten in this pr.

nzpr:
I believe it's better to store fault tolerances, as isFinalized just checks if fault tolerance is greater than threshold. And threshold potentially can be dynamic. Not sure though whats the best way to implement this

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -133,8 +136,12 @@ object Setup {
           estimateBlockHash: BlockHash
       ): Task[Float] = Task.pure(1.0f)
     }
-    implicit val lastFinalizedBlockCalculator = LastFinalizedBlockCalculator[Task](0f)
-    implicit val synchronyConstraintChecker   = SynchronyConstraintChecker[Task](0d)
+    val finalizedBlocksStore  = InMemoryKeyValueStore[Task]()
+    val finalizedBlocksStoreT = finalizedBlocksStore.toTypedStore(codecBlockHash, codecBlockHash)
```
nzpr: As you need to store only list of hashes, you can store only keys, so for value use codec `ignore(0)`, it is defined in scodec package. And use `KeyValueTypedStore[F, BlockHash, Unit]`

nzpr:
I have been thinking, and dynamic threshold, if present, should not matter - if shard decide to rise the threhold for finalization, it should not influence previous finalized block. In that case to finalize child bigger fault tolerance will be required.

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
zsluedem: Nice idea

zsluedem:
close because of https://github.com/rchain/rchain/pull/3258

PR #3259, opened 2020-12-09
Do not add blocks with tuplespace hash mismatch

nzpr:
Do not add block in case of tuplespace hash mismatch occurs, as it can lead to slashing. Should be reverted in block merge branch.


tgrospic (approved):
LGTM

zsluedem (approved):

PR #3263, opened 2020-12-09
Finalized block bugs

zsluedem:
## Overview

1. the calculation on finalized block in memory would lead to infinite loop
2. blockAPi blockInfo faultolerance get based on isfinalized

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -623,19 +623,20 @@ object BlockAPI {
   )(implicit casper: MultiParentCasper[F]): F[A] =
     for {
       dag <- casper.blockDag
-      // TODO this is temporary solution to not calculate fault tolerance for old blocks which is costly
-      oldBlock  = dag.latestBlockNumber.map(_ - block.body.state.blockNumber).map(_ > 100)
-      isInvalid = dag.lookup(block.blockHash).map(_.get.invalid)
+      // TODO this is temporary solution to not calculate fault tolerance all the blocks
+      oldBlock = dag.latestBlockNumber.map(_ - block.body.state.blockNumber).map(_ > 100)
       // Old block fault tolerance / invalid block has -1.0 fault tolerance
-      oldBlockFaultTolerance = isInvalid.map(if (_) -1f else 1f)
       normalizedFaultTolerance <- oldBlock.ifM(
-                                   oldBlockFaultTolerance,
+                                   dag
+                                     .isFinalized(block.blockHash)
+                                     .map(isFinalized => if (isFinalized) 1f else -1f),
                                    SafetyOracle[F]
                                      .normalizedFaultTolerance(dag, block.blockHash)
                                  )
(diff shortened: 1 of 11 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -623,18 +623,9 @@ object BlockAPI {
...
-      normalizedFaultTolerance <- oldBlock.ifM(
-                                   oldBlockFaultTolerance,
-                                   SafetyOracle[F]
-                                     .normalizedFaultTolerance(dag, block.blockHash)
-                                 )
-      initialFault   <- casper.normalizedInitialFault(ProtoUtil.weightMap(block))
-      faultTolerance = normalizedFaultTolerance - initialFault
+      // TODO this is temporary solution to not calculate fault tolerance all the blocks
+      isFinalized    <- dag.isFinalized(block.blockHash)
+      faultTolerance = if (isFinalized) 1f else -1f
```
tgrospic: I think we can calculate fault tolerance for non-finalized blocks. This calculation should be the most few hundred milliseconds, right?

zsluedem: not for very old blocks

zsluedem: Calculating fault tolerance in single-parent mode does not actually have too much meaning

zsluedem: After another thinking, we can calculate in the latest 100 block like we did before.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -623,19 +623,20 @@ object BlockAPI {
...
-      isInvalid = dag.lookup(block.blockHash).map(_.get.invalid)
+      // TODO this is temporary solution to not calculate fault tolerance all the blocks
+      oldBlock = dag.latestBlockNumber.map(_ - block.body.state.blockNumber).map(_ > 100)
       // Old block fault tolerance / invalid block has -1.0 fault tolerance
-      oldBlockFaultTolerance = isInvalid.map(if (_) -1f else 1f)
       normalizedFaultTolerance <- oldBlock.ifM(
-                                   oldBlockFaultTolerance,
+                                   dag
+                                     .isFinalized(block.blockHash)
+                                     .map(isFinalized => if (isFinalized) 1f else -1f),
```
tgrospic, line 632: I was thinking to eliminate any calculation for finalized blocks and calculate only if not finalized and not old.

For example if `/blocks/100` is called, most blocks is finalized and we can save ~90 calculations.

tgrospic: On dev standup we discussed that fault tolerance for finalized blocks can have value other then `1` so it make sense to calculate always latest blocks.

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #3264, opened 2020-12-09
Let syntax

JosephDenman:
## Overview
Introduces sequential and concurrent let syntax.

### DIFF
--- a/rholang/src/main/bnfc/rholang_mercury.cf
+++ b/rholang/src/main/bnfc/rholang_mercury.cf
@@ -65,6 +65,7 @@ PInput.          Proc2  ::= "for" "(" Receipt ")" "{" Proc "}" ;
 PChoice.         Proc2  ::= "select" "{" [Branch] "}" ;
 PMatch.          Proc2  ::= "match" Proc4 "{" [Case] "}" ;
 PBundle.         Proc2  ::= Bundle "{" Proc "}" ;
+PLet.            Proc2  ::= "let" Decl Decls "in" "{" Proc "}" ;
 PIf.             Proc1  ::= "if" "(" Proc ")" Proc2 ;
 -- Use precedence to force braces around an interior if.
 PIfElse.      Proc1 ::= "if" "(" Proc ")" Proc2 "else" Proc1 ;
@@ -72,6 +73,19 @@ PNew.         Proc1 ::= "new" [NameDecl] "in" Proc1 ;
 PPar.         Proc  ::= Proc "|" Proc1 ;
 
 separator Proc "," ;
+
+-- Let declarations
+DeclImpl.    Decl ::= [Name] NameRemainder "<-" [Proc] ;
+
+LinearDeclImpl.    LinearDecl ::= Decl ;
+separator nonempty LinearDecl ";" ;
+
+ConcDeclImpl.    ConcDecl ::= Decl ;
+separator nonempty ConcDecl "&" ;
+
+EmptyDeclImpl.      Decls ::= ;
+LinearDeclsImpl.    Decls ::= ";" [LinearDecl] ;
+ConcDeclsImpl.      Decls ::= "&" [ConcDecl] ;
 
 -- Process variables
 ProcVarWildcard. ProcVar ::= "_" ;
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
@@ -1,5 +1,7 @@
 package coop.rchain.rholang.interpreter.compiler
 
+import java.util.UUID
+
 import cats.effect.Sync
 import cats.implicits._
 import cats.{Applicative, MonadError}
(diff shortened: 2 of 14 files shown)


rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -73,6 +74,19 @@ PPar.         Proc  ::= Proc "|" Proc1 ;
...
+
+LinearDeclImpl.    LinearDecl ::= Decl ;
+separator nonempty LinearDecl ";" ;
+
+ConcDeclImpl.    ConcDecl ::= Decl ;
+separator nonempty ConcDecl "&" ;
+
+EmptyDeclImpl.      Decls ::= ;
+LinearDeclsImpl.    Decls ::= ";" [LinearDecl] ;
+ConcDeclsImpl.      Decls ::= "&" [ConcDecl] ;
```
nzpr, line 88: What is the difference between linear and concurrent let? 
My intuition is that with linear you can do `let a <- 0, b <- a-5` and with parallel it will throw error at compile time? Am I right here?

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+      ParBuilderUtil.mkTerm(s).right.value shouldBe ParBuilderUtil.mkTerm(t).right.value
+
+    "translate a single declaration of multiple variables into a list match process" in {
+      // Variables declared in let declarations are treated as name variables,
+      // but variables bound by match expressions are treated as process variables.
+      // As long as name variables are used in name contexts and process variables are
+      // used in process contexts, the following two programs are equivalent because the normalizer
+      // eliminates the distinction between names and processes with `Par`, relying on context to
+      // decide how to process a given `p: Par`.
+      val s = "let x, @y, @{a | b} <- 1, 2, 3 | 4 in { Nil }"
```
nzpr, line 63: is this equivalent to `let x <- 1 & let @y <-2 & let @{a | b} <- 3 | 4 in { Nil }`?

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+  */
+class LetSpec extends WordSpec with Matchers {
+
+  "The lexer" should {
+
+    "disallow a non-positive number of let declarations" in {
+      ParBuilderUtil.mkTerm("let in { Nil }").left.value shouldBe a[SyntaxError]
+    }
+
+    "disallow '&' and ';' from appearing in the same let declaration sequence" in {
```
nzpr, line 22: just curious - why this is not possible?

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
JosephDenman: No. What you write translates to a COMM rule with three consumes in a join and three produces.

rholang/src/main/bnfc/rholang_mercury.cf
JosephDenman: The difference is simply that the expressions on the RHS of `<-` are evaluated concurrently. So `let x <- [1, 2].slice(0, 1) & y <- 2 + 3 in { Nil }` evaluates `[1, 2].slice(0, 1)` and `2 + 3` concurrently, then binds them to their respective variables.

nzpr: My question is more about if I can with linear let use variables from previous bindings, so is `let a <- 4, b <- a - 5` valid?

JosephDenman: Yes. Your expression translates to:

```
match 4 {
    a => {
        match a - 5 {
            b => 
        }
    } 
}
```

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    "translate multiple concurrent let declarations into a COMM" in {
+      val s = "let x <- 1 & y <- 2 & z <- 3 in { Nil }"
+      val t =
+        """
+          # new x0, x1, x2 in {
+          #   x0!(1) | x1!(2) | x2!(3) |
+          #   for(x <- x0 ; y <- x1 ; z <- x2) { Nil }
+          # }""".stripMargin('#')
```
tgrospic, line 122: Is there a reason to use `#` for the margin?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -1072,6 +1074,201 @@ object ProcNormalizeMatcher {
...
+
+            /*
+             It is not necessary to use UUIDs to achieve concurrent let declarations.
+             While there is the possibility for collisions with either variables declared by the user
+             or variables declared within this translation, the chances for collision are astronomically
+             small (see analysis here: https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87).
+             A strictly correct approach would be one that performs a ADT rather than an AST translation, which
+             was not done here due to time constraints.
+             */
+            val variableNames = List.fill(listNames.size)(UUID.randomUUID().toString)
```
tgrospic, line 1101: What would be necessary to change if new `NameVar` and `NameDeclSimpl` would have been created so that can be in different space then strings variables?

And one more concrete question. Is it possible that users will see generated UUID variables with some error in their code?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -1072,6 +1074,201 @@ object ProcNormalizeMatcher {
...
+              val listLinearBind = new ListLinearBind()
+              for (linearBind <- variableNames.zip(listNames).zip(listNameRemainders).map {
+                                  case ((variableName, listName), nameRemainder) =>
+                                    new LinearBindImpl(
+                                      listName,
+                                      nameRemainder,
+                                      new NameVar(variableName)
+                                    )
+                                })
+                listLinearBind.add(linearBind)
```
tgrospic, line 1118: `.foreach(listLinearBind.add)` can do the trick here. On two other places can also be one liner.

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
JosephDenman: I do it out of habit. A while back Artur Gajowy told me that he'd seen bugs emerge because of the clash between the line separator `|` and the Rholang pipe `|`. I've never seen it myself, and you can consider it defensive programming.

rholang/src/test/scala/coop/rchain/rholang/interpreter/LetSpec.scala
JosephDenman: The Rholang 1.1 spec doesn't include a translation. It's possible once there's a translation. While there's not, we want to make sure the parser catches invalid let declarations.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: > What would be necessary to change if new `NameVar` and `NameDeclSimpl` would have been created so that can be in different space then strings variables?

I'm not sure what you mean.

> And one more concrete question. Is it possible that users will see generated UUID variables with some error in their code?

No. Generated names should not appear directly in error messages. They're always bound, and they can't be referenced in the source code.

JosephDenman:
These commits will be in the forthcoming PR that includes all Rholang 1.1 changes.

PR #3269, opened 2020-12-16
Fix synchrony constraint calculation when latest block is older than approved

tgrospic:
## Overview
When validator propose a block it must satisfy synchrony constraint which depends on reading active validators from tuple space. When LFS state is downloaded and validator latest block is older then approved block, we don't have tuple state for that block. This PR adds this detection and allow validator to propose.

Another small fix to Engine initialization. The bug manifested when LFS node received the state, LFS exporter was not enabled although the flag was set.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
+++ b/casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
@@ -8,14 +8,16 @@ import coop.rchain.casper.protocol.{BlockMessage, Justification}
 import coop.rchain.casper.syntax._
 import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.casper.util.rholang.RuntimeManager
+import coop.rchain.metrics.Span
+import coop.rchain.models.BlockMetadata
 import coop.rchain.models.Validator.Validator
 import coop.rchain.shared.Log
 
-final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Log](
+final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Estimator: Log: Span](
     synchronyConstraintThreshold: Double
 ) {
   private def calculateSeenSendersSince(
-      lastProposed: BlockMessage,
+      lastProposed: BlockMetadata,
       dag: BlockDagRepresentation[F]
   ): F[Set[Validator]] =
     for {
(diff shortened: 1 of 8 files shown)


casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
```diff
@@ -35,40 +36,53 @@ final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Log](
...
+                )
+          } yield synchronyConstraintValue >= synchronyConstraintThreshold
+
+          // Calculate if validator has latest proposed block older than approved block.
+          // NOTE: When LFS is restored, we don't have tuple space older then approved block,
+          //  so we cannot read active validators from PoS.
+          approvedBlockNumber       = ProtoUtil.blockNumber(approvedBlock)
+          latestBlockBehindApproved = approvedBlockNumber > lastProposedBlockMeta.blockNum
+
+          allowedToPropose <- if (latestBlockBehindApproved) true.pure[F] else checkConstraint
```
tgrospic: Maybe it would be better to just skip active validators filtering instead of allow without checking. @nzpr can we do this or this will bring the old problem again resolved in PR #3069?

integration-tests/resources/logback.xml
```diff
@@ -17,8 +17,8 @@
   </appender>
 
   <logger name="coop.rchain.shared.EventLogger" level="OFF" />
+  <logger name="coop.rchain.rspace" level="info" />
```
tgrospic, line 20: Enable info messages from RSpace.

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
nzpr: @tgrospic I feel that using lastProposedTuplespace for fetching active validators is not correct - you have to use state that you are trying to propose on top of, so, basically, state of the main parent. Wdy think?

tgrospic: @nzpr makes sense to use main parent. Do you know why is state from latest block used now?

nzpr: @tgrospic not really, I don't see why its more preferable then main parent.

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
```diff
@@ -35,40 +37,59 @@ final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Log](
...
+            // Loading the whole block is only needed to get post-state hash
+            mainParentBlock     <- BlockStore[F].getUnsafe(mainParentMeta.blockHash)
+            mainParentStateHash = ProtoUtil.postStateHash(mainParentBlock)
+
+            // Get bonds map from PoS
+            // NOTE: It would be useful to have active validators cached in the block in the same way as bonds.
+            activeValidators <- runtimeManager.getActiveValidators(mainParentStateHash)
+
+            // Validators weight map filtered by active validators only.
+            validatorWeightMap = lastProposedBlockMeta.weightMap.filter {
```
nzpr, line 66: I overlooked this, but I think this also should be main parent weight map. Not sure if this is crucial though.

tgrospic: Yes, we are using active validators from main parent and use in filter from the latest block. I'll create a fix.

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
```diff
@@ -35,40 +37,59 @@ final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Log](
...
+              case (validator, _) => activeValidators.contains(validator)
+            }
+            // Guaranteed to be present since last proposed block was present
+            seenSenders   <- calculateSeenSendersSince(lastProposedBlockMeta, dag)
+            sendersWeight = seenSenders.toList.flatMap(validatorWeightMap.get).sum
```
tgrospic, line 71: @nzpr I'm not sure is it possible here that `validatorWeightMap` doesn't have validator calculated as seen sender?

tgrospic: @nzpr nvm, this is already handled.

tgrospic: Fix in PR #3274.

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
tgrospic: Fix in PR #3274.

PR #3271, opened 2020-12-16
Refactor hotstore #2

nzpr:

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -1,310 +1,411 @@
 package coop.rchain.rspace
 
 import cats._
-import cats.implicits._
 import cats.effect._
-import cats.effect.implicits._
-import cats.effect.concurrent.Ref
-import coop.rchain.catscontrib.mtl.implicits._
-import coop.rchain.catscontrib.seq._
-import coop.rchain.shared.Language._
-import coop.rchain.shared.MapOps._
+import cats.effect.concurrent.{Deferred, Ref}
+import cats.implicits._
 import coop.rchain.rspace.internal._
+import coop.rchain.shared.MapOps._
 import coop.rchain.shared.Serialize._
 import scodec.Codec
 
-import scala.collection.concurrent.TrieMap
+import scala.collection.immutable.Map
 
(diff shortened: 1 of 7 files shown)


rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -34,277 +28,372 @@ trait HotStore[F[_], C, P, A, K] {
...
+      conts          = contsAtJoin ++ instContAtJoin
+      err <- joinsRef.modify[Boolean](state => {
+              val curJoins = state(channel)
+              val curConts = conts
+
+              val index       = state(channel).indexOf(join)
+              val outOfBounds = !curJoins.isDefinedAt(index)
+
+              // TODO should attimpting to remove join with non empty contnuation lead to error as well?
+              val doRemove = curConts.isEmpty
```
tgrospic: Hm, this looks like a race. Reading from two Refs and depending on a value in the third. Two threads can enter this function, read the same continuations and result in different state.

Check to remove join depends on one version of the state (when continuations are read) so here when checking if it's empty it can give inconsistent results. Maybe we cannot escape to have on Ref for the whole state. Or maybe something similar to RSpace per channel locking.

nzpr: I have been thinking on this. Installed continuations should be filled only once when runtime is spawned, so `instContAtJoin` should never change here. And `conts` which is `contsAtJoin ++ instContAtJoin` - sum of this two non atomic reads, is used to detect if code should proceed and delete. 

So only race here can be at `contsAtJoin` and can lead to attempting to remove the same join several times. Which should not be a problem
https://github.com/rchain/rchain/pull/3271/commits/ce22b4e74f292374d7b5c2736896cc2b927b10e1#diff-84d533e963703b6bf252ff208084a62af1116772648322604b3869ad15b8c39cR220-R224

nzpr: https://github.com/rchain/rchain/pull/3271/commits/ce22b4e74f292374d7b5c2736896cc2b927b10e1#diff-84d533e963703b6bf252ff208084a62af1116772648322604b3869ad15b8c39cR224

zsluedem:
https://github.com/rchain/rchain/pull/3526 since it is merged, close this one

PR #3279, opened 2020-12-28
Rholang 1.1 implementation

JosephDenman:
## Overview

Exactly as the title says, Rholang 1.1 implementation. Sorry it's so big.

https://drive.google.com/file/d/1FvPuiPm7ytGyklex_PEmVyJ8XgbTdIEg


rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -61,18 +61,38 @@ PAnd.            Proc5  ::= Proc5 "and" Proc6 ;
...
+separator nonempty ConcDecl "&" ;
+
+EmptyDeclImpl.      Decls ::= ;
+LinearDeclsImpl.    Decls ::= ";" [LinearDecl] ;
+ConcDeclsImpl.      Decls ::= "&" [ConcDecl] ;
+
+-- Continuation for synchronous send
+
+EmptyCont.      SynchSendCont ::= "." ;
+NonEmptyCont.   SynchSendCont ::= ";" Proc1 ;
```
zsluedem, line 94: I am trying to understand the bnfc grammar. 
Joe, do you have any recommend docs or examples for understand the whole design on bnfc?
I am just curious.

rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -101,20 +122,24 @@ LinearSimple. ReceiptLinearImpl ::= [LinearBind] ;
...
+SimpleSource.         NameSource ::= Name ;
+ReceiveSendSource.    NameSource ::= Name "?!" ;
+SendReceiveSource.    NameSource ::= Name "!?" "(" [Proc] ")" ;
 
 -- Repeated Receipts
 RepeatedSimple. ReceiptRepeatedImpl ::= [RepeatedBind] ;
 -- Single Repeated Bind
 RepeatedBindImpl. RepeatedBind ::= [Name] NameRemainder "<=" Name ;
-separator nonempty RepeatedBind ";" ;
+separator nonempty RepeatedBind "&" ;
```
zsluedem, line 136: Question:
The rule changed here requires you to change the join ";" to "&" in `MakeMint.rho` and etc., right?

This looks like is not backward compatible from my perspective.Is that correct?

rholang/src/main/bnfc/rholang_mercury.cf
zsluedem: I got https://bnfc.readthedocs.io/en/latest/lbnf.html#introduction. But if you got something else, I'd love to take a look also.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
```diff
@@ -12,36 +12,50 @@ import coop.rchain.rholang.interpreter.errors._
 import coop.rchain.rholang.syntax.rholang_mercury.Absyn.Proc
 import coop.rchain.rholang.syntax.rholang_mercury.{parser, Yylex}
 
-trait ParBuilder[F[_]] {
-  def buildNormalizedTerm(source: String, normalizerEnv: Map[String, Par]): F[Par]
+trait Compiler[F[_]] {
 
-  def buildNormalizedTerm(reader: Reader, normalizerEnv: Map[String, Par]): F[Par]
+  def sourceToADT(source: String): F[Par] =
+    sourceToADT(source, Map.empty[String, Par])
```
zsluedem, line 18: I just want to say that this overloaded method can be a syntax extension.

Syntax extension is brought by @tgrospic which can make trait simpler. Here is the example.
https://github.com/rchain/rchain/blob/dev/rspace/src/main/scala/coop/rchain/rspace/state/RSpaceExporterSyntax.scala

This is just a good to have request.

rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -61,18 +61,38 @@ PAnd.            Proc5  ::= Proc5 "and" Proc6 ;
...
+PInput.          Proc2  ::= "for" "(" [Receipt] ")" "{" Proc "}" ;
 PChoice.         Proc2  ::= "select" "{" [Branch] "}" ;
 PMatch.          Proc2  ::= "match" Proc4 "{" [Case] "}" ;
 PBundle.         Proc2  ::= Bundle "{" Proc "}" ;
+PLet.            Proc2  ::= "let" Decl Decls "in" "{" Proc "}" ;
 PIf.             Proc1  ::= "if" "(" Proc ")" Proc2 ;
 -- Use precedence to force braces around an interior if.
 PIfElse.      Proc1 ::= "if" "(" Proc ")" Proc2 "else" Proc1 ;
 PNew.         Proc1 ::= "new" [NameDecl] "in" Proc1 ;
+PSendSynch.   Proc1 ::= Name "!?" "(" [Proc] ")" SynchSendCont ;
```
zsluedem, line 73: Just a stupid question. Is it a typo or is there some meaning in the last letter `h`?
What I mean is the `h` at the end of `PSendSynch`.

zsluedem: I think it is just short for `synchronous` which I usuaslly use `sync`. So nvm.

rholang/src/test/scala/coop/rchain/rholang/interpreter/SynchronousOutputSpec.scala
```diff
@@ -0,0 +1,52 @@
+package coop.rchain.rholang.interpreter
+
+import coop.rchain.rholang.interpreter.ParBuilderUtil.assertCompiledEqual
+import org.scalatest.{FlatSpec, Matchers}
+
+class SynchronousOutputSpec extends FlatSpec with Matchers {
+
+  "';'" should "have a higher precedence than '|'" in {
+    val s = "@1!?(1) ; @2!?(2). | @3!?(3)."
```
zsluedem, line 9: I think it would be better to have another example like `@3!?(3). | @1!?(1) ; @2!?(2). ` which is a more convincing test case that ";" have higher precedence than "|" 
Because based on the perspective of seeing this program in sequence which ";" should consider first.
Does that make sense?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -1061,17 +1191,222 @@ object ProcNormalizeMatcher {
...
+             or variables declared within this translation, the chances for collision are astronomically
+             small (see analysis here: https://towardsdatascience.com/are-uuids-really-unique-57eb80fc2a87).
+             A strictly correct approach would be one that performs a ADT rather than an AST translation, which
+             was not done here due to time constraints.
+             */
+            val variableNames = List.fill(listNames.size)(UUID.randomUUID().toString)
+
+            val psends = variableNames.zip(listProcs).map {
+              case (variableName, listProc) =>
+                new PSend(new NameVar(variableName), new SendSingle(), listProc)
```
zsluedem, line 1233: The `NameVar` is not something new which this pr introduced. 
I am curious how codes before this pr work with `variableName` in NameVar.
I mean instead of `UUID`, what does these `variableName` come from?

rholang/src/test/scala/coop/rchain/rholang/interpreter/SynchronousOutputSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    val t =
+      """
+          # new x0 in {
+          #   @1!(*x0, 1) |
```
zsluedem, line 13: I am trying `PSendSynch`. I was thinking that `PSendSynch` is supposed to solve condition like below with `ack` thing.
  The example here  https://github.com/rchain/rchain/blob/dev/rholang/examples/stdoutAck.rho#L1-L6
  ```
  new ack, stdout(`rho:io:stdout`), stdoutAck(`rho:io:stdoutAck`) in {
    stdoutAck!("hello, world!", *ack) |
    for (_ <- ack) {
      stdout!("received")
    }
  }
  ```

  can be re-written into
 ```
  new stdout(`rho:io:stdout`), stdoutAck(`rho:io:stdoutAck`) in {
    stdoutAck!?("hello, world!"); stdout!("received")
  }
```
But it seems that `PSendSynch` is quite limited which here the `new x0` must be the first argument in `@1!(*x0, 1)`.
Is that correct?

File: rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala [hunk 8]
```diff
@@ -722,17 +734,52 @@ object ProcNormalizeMatcher {
...
-          dataResults._2.knownFree
+          dataResults._2.freeMap
         )
 
+      case p: PSendSynch =>
+        val identifier = UUID.randomUUID().toString
+        val nameVar    = new NameVar(identifier)
+
+        val send: PSend = {
+          p.listproc_.prepend(new PEval(nameVar))
```
zsluedem, line 753: I think I found that the new var can only be the first one because of this

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -722,17 +734,52 @@ object ProcNormalizeMatcher {
...
+          new PSend(p.name_, new SendSingle(), p.listproc_)
+        }
+
+        val receive: PInput = {
+
+          val listName = new ListName()
+          listName.add(new NameWildcard)
```
zsluedem, line 760: This can only bind to `NameWildcard` which means the continuation can use the return value. 
This results in that only `ack` situation can be used in this syntax while return value type is not applicable, is that correct?
I mean a function like https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/ListOps.rho#L45-L48 can not used in this situation.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
zsluedem: I think most of the existing libraries put this `Ack` name in the tail not the head.
But I am not sure which one is better. Just curious whether it is a problem to put it in the tail.

rholang/src/test/scala/coop/rchain/rholang/interpreter/SequentialConsumeSpec.scala
```diff
@@ -0,0 +1,24 @@
...
+import coop.rchain.rholang.interpreter.ParBuilderUtil.assertCompiledEqual
+import org.scalatest.{Matchers, WordSpec}
+
+class SequentialConsumeSpec extends WordSpec with Matchers {
+
+  "The normalizer" should {
+
+    "translate one sequential join into multiple nested consumes" in {
+
+      val s = "for(x0 <- @0 & x1 <- @1 & x2 <- @2 ; x3 <- @3 & x4 <- @4 & x5 <- @5){ Nil }"
```
zsluedem, line 12: Just want to say that this is a great feature I wanted. :+1:

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -796,188 +843,271 @@ object ProcNormalizeMatcher {
...
+          val receiptContainsComplexSource: Boolean = {
+            p.listreceipt_.head match {
+              case rl: ReceiptLinear =>
+                rl.receiptlinearimpl_ match {
+                  case ls: LinearSimple =>
+                    ls.listlinearbind_.exists {
+                      case lbi: LinearBindImpl =>
+                        lbi.namesource_ match {
+                          case _: SimpleSource => false
+                          case _               => true
```
zsluedem, line 877: How does complex source look like?
```
for (_ <- @1?! ) {}
```
something like above?

zsluedem: Oh. Nvm. I think I found your test cases.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
zsluedem: Oh. Actually the https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/ListOps.rho#L45-L48 can be solved by complexConsume, right?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -796,188 +843,271 @@ object ProcNormalizeMatcher {
...
+                                    case srs: SendReceiveSource =>
+                                      listNameDecl.add(new NameDeclSimpl(identifier))
+                                      listLinearBind.add(
+                                        new LinearBindImpl(
+                                          lbi.listname_,
+                                          lbi.nameremainder_,
+                                          new SimpleSource(r)
+                                        )
+                                      )
+                                      srs.listproc_.prepend(new PEval(r))
```
zsluedem, line 933: Same question as https://github.com/rchain/rchain/pull/3279#discussion_r550859155

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
```diff
@@ -1061,17 +1191,222 @@ object ProcNormalizeMatcher {
...
+                }
+                .map {
+                  case (vectorPar, knownFree, locallyFree, connectiveUsed) =>
+                    ProcVisitOutputs(
+                      EList(vectorPar.reverse, locallyFree, connectiveUsed, none[Var]),
+                      knownFree
+                    )
+                }
+
+            // Largely similar to how consume patterns are processed.
```
zsluedem, line 1332: It would be better to write a re-usable function for similar process

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
```diff
@@ -12,36 +12,50 @@ import coop.rchain.rholang.interpreter.errors._
...
+
+  def astToADT(proc: Proc): F[Par] =
+    astToADT(proc, Map.empty[String, Par])
+
+  def astToADT(proc: Proc, normalizerEnv: Map[String, Par]): F[Par]
+
+  def sourceToAST(source: String): F[Proc] =
+    sourceToAST(new StringReader(source))
+
+  def sourceToAST(reader: Reader): F[Proc]
```
zsluedem, line 36: Another good-to-have suggestion for a better modular project and future syntax sugar development though I am not sure if that's reasonable.
I think it would be better to have a method that translates syntax sugar to another AST. What I mean is that we can separate procedures into 
`sourceString -> RichAST(AST with syntax sugar) -> PureAST(AST without syntax sugar) -> ADT`.
Currently the `RichAST(AST with syntax sugar) -> PureAST(AST without syntax sugar)` methods are embedded into `astToADT`.

zsluedem:
@JosDenmark Very good one~!Left some comments. I like your test cases. They almost cover everything you write.

rholang/src/main/bnfc/rholang_mercury.cf
JosephDenman: That's all I used. If there's a specific part that confuses you, I can try to help.

rholang/src/main/bnfc/rholang_mercury.cf
JosephDenman: Yes. That's correct. No, it's not backward compatible.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: The variable name is supplied by the programmer in the source code.

rholang/src/test/scala/coop/rchain/rholang/interpreter/SynchronousOutputSpec.scala
JosephDenman: Yes. Unfortunately, in order for remainder syntax to still make sense, we have to switch to the convention that the return channel is the first argument received. So, to be consistent, we'll have to rewrite existing Rholang source to follow that convention. That includes having rho:io:stdoutAck (and the other system processes) expect the return channel to be the first argument. Then your example will work.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: If it goes in the tail, we can no longer use remainder syntax (as it currently functions). Do you see why?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: See answer above.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: Yes, it can be simplified with a complex source, assuming the return channel convention is followed.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
JosephDenman: @tgrospic Did you find this pattern somewhere, or is it your own?

JosephDenman: It looks like the "Pimp My Class" pattern: http://blog.rcard.in/jvm/programming/design-pattern/2019/12/15/pimp-my-library-pattern.html

JosephDenman: Still, that's not appropriate here. We have control of the `Compiler` interface. Moreover, each of these overloaded alternatives are common enough to merit adding them to the main API. This is the preferred alternative to adding a default parameter to each of the methods.

tgrospic: @JosDenmark this is the pattern used in cats. Here is the example for monad.
https://github.com/typelevel/cats/blob/e1a7cfcddc/core/src/main/scala/cats/syntax/monad.scala
And all extension can then be imported with one import _all_.
https://github.com/typelevel/cats/blob/e1a7cfcddc/core/src/main/scala/cats/syntax/all.scala

I like it because it gives more flexibility to define extensions only in projects that are needed or to combine with additional dependency to enrich the _syntax_.
My opinion is that default implementations are fine if they don't require additional implicit dependencies like here.
https://github.com/rchain/rchain/blob/b69eb8a457f67e7ca32581c515db6958515cd62c/block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala#L21-L22

So here I'm fine with either decision. I'm thinking more about how to make it more clear which methods are basic definitions and which are defaults/extensions.

P.S.
This is basically dependency injection pattern which will be used even more in v3 of cats. There are also changes to the inheritance hierarchy where `Sync` will be moved from top to bottom.
https://youtu.be/xEkSDiz3XO0?t=2449

rholang/src/main/bnfc/rholang_mercury.cf
```diff
@@ -87,6 +88,11 @@ EmptyDeclImpl.      Decls ::= ;
 LinearDeclsImpl.    Decls ::= ";" [LinearDecl] ;
 ConcDeclsImpl.      Decls ::= "&" [ConcDecl] ;
 
+-- Continuation for synchronous send
+
+EmptyCont.      SynchSendCont ::= "." ;
```
tgrospic, line 93: @JosDenmark do you see the need for the special syntax for empty continuation? Is there a reason why not just insert automatically on the last send?

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
JosephDenman: Yeah. I don't think these qualify as "extension" methods. They belong in the main interface IMO. Let's consider this issue resolved.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
JosephDenman: That's definitely a separate PR!

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
JosephDenman: Agreed, but that's not feasible without significant compiler refactoring.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/Compiler.scala
zsluedem: Yes.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalize.scala
tgrospic: @zsluedem I created an issue for refactoring of normalizer #3290.

tgrospic (approved):
Excellent work! Thanks Joe!

Just one question and please add additional test for precedence as Will mentioned.

zsluedem (approved):

rholang/src/main/bnfc/rholang_mercury.cf
JosephDenman: No, for the record I think this syntax is pointless. If the programmer can remember to type `.` at the end of a sequence of outputs, they know where the end is, so the syntax is unnecessary. Plus, any time a sequence of outputs is put in parallel with another sequence of outputs, they're separated by `|` anyway, so it's doubly unnecessary :)

zsluedem:
bors merge

zsluedem:
bors merge

zsluedem:
bors try

zsluedem:
bors merge

PR #3288, opened 2021-01-04
Refactor event log to be Vector instead of Seq (List)

### LINKED ISSUE #3287: Refactor event log to be Vector instead of Seq (List)
## Overview
We should investigate if using `Vector` for event log will speed up overall execution of Rholang when creating and replaying a block.

https://github.com/rchain/rchain/blob/9ef35159ee0ca5291b8ffdf769c9522a05ab6d9f/rspace/src/main/scala/coop/rchain/rspace/trace/package.scala#L7

vinay-bhamidipati:
## Overview

Addresses  #3287. Makes event log a Vector rather than a Seq. Fixes related issues with this change. Targeted toward improving performance.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
@@ -447,7 +447,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
             .copy(systemDeployError = Some(errorMsg))
       }
       .run // run the computation and produce the logs
-      .map { case (accLog, pd) => pd.copy(deployLog = accLog.toList) }
+      .map { case (accLog, pd) => pd.copy(deployLog = accLog) }
   }
 
   private def processDeploy(runtime: Runtime[F])(
@@ -462,7 +462,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
       deployResult = ProcessedDeploy(
         deploy,
         Cost.toProto(cost),
-        checkpoint.log.map(EventConverter.toCasperEvent).toList,
+        checkpoint.log.map(EventConverter.toCasperEvent),
         errors.nonEmpty
       )
       _ <- if (errors.nonEmpty) runtime.space.revertToSoftCheckpoint(fallback)
--- a/models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
+++ b/models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
@@ -313,7 +313,7 @@ object RChainState {
 final case class ProcessedDeploy(
     deploy: Signed[DeployData],
     cost: PCost,
-    deployLog: List[Event],
+    deployLog: Vector[Event],
     isFailed: Boolean,
     systemDeployError: Option[String] = None
 ) {
@@ -337,13 +337,13 @@ final case class ProcessedDeploy(
 
 object ProcessedDeploy {
   def empty(deploy: Signed[DeployData]) =
-    ProcessedDeploy(deploy, PCost(), List.empty, isFailed = false)
+    ProcessedDeploy(deploy, PCost(), Vector.empty, isFailed = false)
   def from(pd: ProcessedDeployProto): Either[String, ProcessedDeploy] =
     for {
       ddProto   <- pd.deploy.toRight("DeployData not available")
       dd        <- DeployData.from(ddProto)
       cost      <- pd.cost.toRight("Cost not available")
-      deployLog <- pd.deployLog.toList.traverse(Event.from)
+      deployLog <- pd.deployLog.toVector.traverse(Event.from)
     } yield ProcessedDeploy(
       dd,
       cost,
@@ -404,36 +404,42 @@ object SystemDeployData {
 
 sealed trait ProcessedSystemDeploy {
   val systemDeploy: SystemDeployData
-  def eventList: List[Event]
+  def eventList: Vector[Event]
   def failed: Boolean
-  def fold[A](ifSucceeded: List[Event] => A, ifFailed: (List[Event], String) => A): A
+  def fold[A](ifSucceeded: Vector[Event] => A, ifFailed: (Vector[Event], String) => A): A
 }
 
 object ProcessedSystemDeploy {
 
   final case class Succeeded(
-      eventList: List[Event],
+      eventList: Vector[Event],
       systemDeploy: SystemDeployData
   ) extends ProcessedSystemDeploy {
     val failed = false
 
-    override def fold[A](ifSucceeded: List[Event] => A, ifFailed: (List[Event], String) => A): A =
+    override def fold[A](
+        ifSucceeded: Vector[Event] => A,
+        ifFailed: (Vector[Event], String) => A
+    ): A =
       ifSucceeded(eventList)
   }
 
   final case class Failed(
-      eventList: List[Event],
+      eventList: Vector[Event],
       errorMsg: String
   ) extends ProcessedSystemDeploy {
     val failed                                  = true
     override val systemDeploy: SystemDeployData = SystemDeployData.empty
 
-    override def fold[A](ifSucceeded: List[Event] => A, ifFailed: (List[Event], String) => A): A =
+    override def fold[A](
+        ifSucceeded: Vector[Event] => A,
+        ifFailed: (Vector[Event], String) => A
+    ): A =
       ifFailed(eventList, errorMsg)
   }
 
   def from(psd: ProcessedSystemDeployProto): Either[String, ProcessedSystemDeploy] =
-    psd.deployLog.toList
+    psd.deployLog.toVector
       .traverse(Event.from)
       .map(
         deployLog =>
--- a/models/src/main/scala/coop/rchain/models/HashM.scala
+++ b/models/src/main/scala/coop/rchain/models/HashM.scala
@@ -83,15 +83,12 @@ object HashM extends HashMDerivation {
 
     override def hash[F[_]: Sync](value: Seq[A]): F[Int] =
       value.toList.traverse(HashM[A].hash[F]).map(_.hashCode)
-      
 
   }
   implicit def vectorHash[A: HashM]: HashM[Vector[A]] = new HashM[Vector[A]] {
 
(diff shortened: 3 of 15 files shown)


vinay-bhamidipati:
Further improvements may be made by removing toVector in some places where earlier toList was being used to convert a Seq to a List. From what I can see, this would have no conflicts, but I held off from doing it since I was unsure whether it would be necessary for other parts of the project. I can remove them if permissible, let me know.

vinay-bhamidipati:
Also, any input on how to actually determine whether these changes increase perf?

tgrospic:
bors try

.github/workflows/continuous-integration.yml
```diff
@@ -426,45 +426,51 @@ jobs:
...
   # Merge into "trying" branch (bors try) does not trigger release_* jobs.
 
-  bors_try_success:
+  bors_success:
     name: bors build finished
     needs:
-      - run_unit_tests
-      - run_integration_tests
-    if: "github.event_name == 'push' && github.ref == 'refs/heads/trying' && success()"
+      - release_docker_image
```
tgrospic: Previously we had different dependency for `try` and `merge` commands. For the _try_ only tests were run and for _merge_ also Docker image is published as mentioned in comments.

Can we have the same functionality as before or it's how bors works now?

Also I'd suggest that this change is done in a new PR and after merge this PR can be rebased.

project/Dependencies.scala
```diff
@@ -54,14 +54,14 @@ object Dependencies {
...
-  val logstashLogback     = "net.logstash.logback"        % "logstash-logback-encoder"  % "5.3"
+  val logstashLogback     = "net.logstash.logback"        % "logstash-logback-encoder"  % "6.6"
   val lz4                 = "org.lz4"                     % "lz4-java"                  % "1.7.1"
   val magnolia            = "com.propensive"             %% "magnolia"                  % "0.12.0"
   val monix               = "io.monix"                   %% "monix"                     % "3.3.0"
   val pureconfig          = "com.github.pureconfig"      %% "pureconfig"                % "0.14.0"
   val scalaLogging        = "com.typesafe.scala-logging" %% "scala-logging"             % "3.9.2"
   val scalaUri            = "io.lemonlabs"               %% "scala-uri"                 % "3.0.0"
-  val scalacheck          = "org.scalacheck"             %% "scalacheck"                % "1.15.1"
+  val scalacheck          = "org.scalacheck"             %% "scalacheck"                % "1.15.2"
```
tgrospic: Changes to dependency should not be in this PR. Also please squash/rebase commits so that intermediate changes will not be merged.

models/src/main/scala/coop/rchain/models/HashM.scala
```diff
@@ -86,6 +86,14 @@ object HashM extends HashMDerivation {
         hashes <- value.toList.traverse(HashM[A].hash[F])
       } yield hashes.hashCode()
 
+  }
+  implicit def vectorHash[A: HashM]: HashM[Vector[A]] = new HashM[Vector[A]] {
+
+    override def hash[F[_]: Sync](value: Vector[A]): F[Int] =
+      for {
+        hashes <- value.toVector.traverse(HashM[A].hash[F])
+      } yield hashes.hashCode()
```
tgrospic: `Vector` and `List` extends `Seq` so the new definition specific to Vector should not be needed. What is the reason to add these implementations for equals and hash code?

.github/workflows/continuous-integration.yml
vinay-bhamidipati: With bors now, there can be only one test named 'bors build finished'. If there are multiple and one or more skips, bors considers it a failure. One way to fix it would be to have 'release docker image' execute as a standalone test for ref == staging and have bors_success just depend on unit and integration tests. That way, if release docker image fails on a merge, bors will fail, but on a try, 'release docker image' wont execute and bors will succeed. Hopefully.

models/src/main/scala/coop/rchain/models/HashM.scala
vinay-bhamidipati: Without them, the code will not compile; several errors will appear in those files. I think it is because in the definition in HashM and EqualsM for Seq the type is explicitly seq and does not use < to indicate that child classes can be used. I don't think Lists were being used by EqualsM and HashM, I think it was Seq that was being used directly so there were no conflicts with the old code.

tgrospic: Makes sense. This will also ensure that no conversion is happening here. `value` is already and Vector so `.toVector` is redundant.

For comprehension is not needed, at the end it's just mapping `.map(_.hashCode)`.

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -309,7 +309,7 @@ class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log](
           (startHash, processedDeploys) = deployProcessResult
           systemDeployProcessResult <- {
             import cats.instances.list._
-            systemDeploys.toList.foldM((startHash, Vector.empty[ProcessedSystemDeploy])) {
+            systemDeploys.toVector.foldM((startHash, Vector.empty[ProcessedSystemDeploy])) {
```
tgrospic: `systemDeploys` will be a List here so better to not change it. Although this list should not be big enough to make a difference.

P.S.
With latest cats update it's no longer needed to import instance when `traverse` (`foldM`) is used. In cats v3 (or 2.3) Seq will have instance for Traverse so explicit conversion should also be redundant.
https://github.com/typelevel/cats/pull/3620

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
```diff
@@ -337,13 +337,13 @@ final case class ProcessedDeploy(
...
   def empty(deploy: Signed[DeployData]) =
-    ProcessedDeploy(deploy, PCost(), List.empty, isFailed = false)
+    ProcessedDeploy(deploy, PCost(), Vector.empty, isFailed = false)
   def from(pd: ProcessedDeployProto): Either[String, ProcessedDeploy] =
     for {
       ddProto   <- pd.deploy.toRight("DeployData not available")
       dd        <- DeployData.from(ddProto)
       cost      <- pd.cost.toRight("Cost not available")
-      deployLog <- pd.deployLog.toList.traverse(Event.from)
+      deployLog <- pd.deployLog.toVector.traverse(Event.from)
```
tgrospic, line 346: Is `deployLog` already a Vector?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
vinay-bhamidipati: Changed!

models/src/main/scala/coop/rchain/models/HashM.scala
vinay-bhamidipati: Changed!

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
vinay-bhamidipati: deployLog on the right comes from the class ProcessedDeployProto which defines it to be a Seq. That is why toVector is necessary so that it is converted from Seq to Vector. Maybe we make deployLog in ProcessedDeployProto a vector instead of Seq as well? @tgrospic

nzpr:
@vb-cs to see influence on perf, if you don't want manually measure parts of the code you are tuning, you should use jaeger. The simplest way to do that is to add `--zipkin` to your run command, and run jaeger instance locally https://www.jaegertracing.io/docs/1.21/getting-started/#all-in-one - its just one command and you have instance running locally. 

Rnode with `--zipkin` flag streams events to localhost jaeger server, so you will see events there. 
It will look like this, list of events:
Screenshot 2021-01-08 at 11 14 36
and detailed view of one particular event of block replay
Screenshot 2021-01-08 at 11 15 02

So you can see what is the difference there. 
You will need to create blocks on your local node, so I'm assuming you know how to do that.

9rb:
bors try

.github/workflows/continuous-integration.yml
tgrospic: I think we can release Docker image on bors Try also.

tgrospic:
@vb-cs everything now looks good. Please add some data as confirmation that these changes resulted in performance speedup. Measurement can be done with Jaeger as mentioned by @nzpr. Logs data can also be used to show the difference in duration of some operations e.g. genesis creation.

vinay-bhamidipati:
> 
> 
> @vb-cs everything now looks good. Please add some data as confirmation that these changes resulted in performance speedup. Measurement can be done with Jaeger as mentioned by @nzpr. Logs data can also be used to show the difference in duration of some operations e.g. genesis creation.

Having trouble with this. When I try to run rnode I get 

node 14:19:35.626 [ERROR] [main                ] [coop.rchain.node.Main$       ] - Unhandled exception in thread main
node coop.rchain.blockstorage.finality.LastFinalizedBlockNotFound$: null

@tgrospic

9rb (approved):
If @zsluedem or @nzpr can confirm the performance benefits, this can be merged after a second approval

zsluedem:
I don't see big difference in performance in my tests

zsluedem (approved):
LGTM

tgrospic:
If there is no difference in performance better to leave it as is because `Seq` is more general then `Vector`.

zsluedem:
> If there is no difference in performance better to leave it as is because `Seq` is more general then `Vector`.

@tgrospic  I am curious in what way you tested see performance better.

9rb:
> 
> 
> I don't see big difference in performance in my tests

@zsluedem could you please post some performance test data here, pre- and post- this PR ? 
In Scala 3.0 there's ArraySeq and other data structures. I think @tgrospic may be implying that we can try those later if vector doesn't increase performance and then make a choice of which structure to use.

zsluedem:
Get block and replay a block avg time:
Vector:  10.040824460983277 s
Seq: 10.043872594833374  s

9rb:
Closing this for now, since the performance improvement seems miniscule. We can revisit this issue with ArraySeq and other structures in Scala 2.13 and Scala 3.0 and re-open this at that time, if needed

PR #3294, opened 2021-01-07
Updating Kamon [1.x.x -> 2.0.0]

vinay-bhamidipati:
## Overview

Updates Kamon-related dependencies to 2.0.0 (in a recent update to Kamon, all its dependencies now share the same version).

### Notes

This one took a lot of work. Will need substantial review. I'm particularly concerned about the InfluxDBReporters and the Prometheus reporters. The start function that used to be overridden in these files is no longer in use, so I was unsure where to put the setup that was being done. Also important to note, it seems like these files are based off the template files provided in the Kamon library which have undergone a lot of change since 1.x.x., so if someone could verify whether Kamon is reporting what it should be with these changes, that would be very much appreciated. Thanks,


9rb (changes requested):
@vb-cs 
1. Latest kamon version is 2.1.9
2. There's also Kamon-http4s we don't currently seem to use https://github.com/kamon-io/kamon-http4s  May be you can look into that and work with @nzpr or @tgrospic to see if it's worth adding ? This article https://blog.softwaremill.com/tmwl-april20-kafka-consumer-lag-monitoring-microservices-hasura-2c3c5f4364d9 half way down has some advice on how to correctly configure kamon-http4s if you choose to use it. This or similar guidance may also need to be investigated and applied to correct our current Kamon configuration, to reduce resources consumed by Kamon.

9rb:
bors try

Issue #3277, opened 2020-12-26
Collection of issues left over from block-merge sync to dev branch


nzpr:
- ~~redesign proposer state~~ done in https://github.com/rchain/rchain/pull/3312
- ~~address minor comments~~ done in https://github.com/rchain/rchain/pull/3312
- reimplement tests disabled in <https://github.com/rchain/rchain/pull/3201/commits/53578aa8c579b434eee256f7971143fcc62d68c7>, if applicable
- ~~fix all tests~~ done in https://github.com/rchain/rchain/pull/3312

nzpr:
Reenable tests disabled in in https://github.com/rchain/rchain/pull/3303/checks?check_run_id=1811486455 + ~~fix failing leftover. It looks like deadlock is somewhere, because integration tests are hanging.
Anther test of a concern is https://github.com/rchain/rchain/pull/3303/checks?check_run_id=1811474536#step:8:1664 which were not seen before~~ UPD. This is fixed in https://github.com/rchain/rchain/pull/3312

nzpr:
Tests requiring attention:
[HashSetCasperSpec](https://github.com/rchain/rchain/blob/53578aa8c579b434eee256f7971143fcc62d68c7/casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala#L1-L247) <- https://github.com/rchain/rchain/pull/3351#discussion_r593023486 -> Requires sync version of propose (@zsluedem is working on this)
[MultiParentCasperAddBlockSpec](https://github.com/rchain/rchain/blob/5a9b6d418ebb4cc86781ae69070d629491a6414a/casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala#L39-L73) <- https://github.com/rchain/rchain/pull/3351#discussion_r593024291 -> I'm not sure this test makes sense. It's "MultiParentCasper" should "not allow multiple threads to process the same block". Its a test for legacy synchronous implementation imo.
[CreateBlockAPITest](https://github.com/rchain/rchain/blob/3afeda6ff0ac8a2711ce7448b2d8dcc2e186d60a/casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala#L60) <- https://github.com/rchain/rchain/pull/3351#discussion_r593026591 and https://github.com/rchain/rchain/pull/3351#discussion_r593027343 -> This should be a test for Proposer API (might be ProposerInstance)
[...]

PR #3298, opened 2021-01-14
Add autopropose option

### LINKED ISSUE #3247: Enable autopropose
## Overview
Now the only way to issue a block is a node operator to call internal API port.
This is not user friendly, sustainable, also validator will be subject to ejecting if it does not propose in correct way.
How node proposes block should defined by shard behaviour.

NOTE: triggering propose does not mean issuing a block. If synchrony constraint fails, node won't create block.

### Design
Proposer should maintain the following features:
1. only one propose should be happening at a time
2. return to caller propose ID, or None if another propose is in progress
3. should have an Mvar variable denoting that propose has been triggered, but Proposer as not able to process it because another propose is in progress. In that case propose should be triggered after one that is in progress finishes.
4. should maintain Deferred propose result inside node state, so checking proposer status is cheap.

There are several places where propose can be called:
1. internal API port
2. after block added
3. after deploy received

1. implement proposer (done)
2. enable propose triggering wen `--autopropose` run option is provided

nzpr:
## Overview

This PR intention is to make development and testing easier.

1.  Add `--deployer-private-key $key` CLI/config option. 
When set, with node in development mode (`--dev-mode` flag is set), makes node to put `Nil` deploy signed by this key into each block created. This makes it possible to just call `rnode propose` and make node issue a block without requirement to make a deploy first.
NOTE: corresponding private key has to be funded (e.g. through wallets file in genesis block).

2.  Add `--autopropose` CLI/config flag. When set, node makes attempt to propose after each new block added. 
NOTE: this does not guarantee block will be issued. If some propose constraint is not met (e.g. synchrony constraint), propose fails.

3.  Make node attempting to propose after each new deploy when `--autopropose` is set

Overall, after this PR it should be easy to seed network with a single propose command and then block creation should be automatic, according to propose constraints. Later this propose constraints will be read from on-chain shard config.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -8,11 +8,12 @@ import coop.rchain.blockstorage.BlockStore
 import coop.rchain.blockstorage.deploy.DeployStorage
 import coop.rchain.blockstorage.syntax._
 import coop.rchain.casper.protocol.{Header, _}
-import coop.rchain.casper.util.ProtoUtil
+import coop.rchain.casper.util.{ConstructDeploy, ProtoUtil}
 import coop.rchain.casper.util.rholang.RuntimeManager.StateHash
 import coop.rchain.casper.util.rholang._
 import coop.rchain.casper.util.rholang.costacc.{CloseBlockDeploy, SlashDeploy}
 import coop.rchain.casper.{CasperSnapshot, PrettyPrinter, ValidatorIdentity}
+import coop.rchain.crypto.PrivateKey
 import coop.rchain.crypto.codec.Base16
 import coop.rchain.crypto.signatures.Signed
 import coop.rchain.metrics.{Metrics, Span}
@@ -37,8 +38,8 @@ object BlockCreator {
    */
   def create[F[_]: Sync: Log: Time: BlockStore: DeployStorage: Metrics: RuntimeManager: Span](
       s: CasperSnapshot[F],
-      validatorIdentity: ValidatorIdentity
-      //dummyDeployOpt: Option[(PrivateKey, String)] = None
+      validatorIdentity: ValidatorIdentity,
+      dummyDeployOpt: Option[(PrivateKey, String)] = None
   )(implicit runtimeManager: RuntimeManager[F]): F[BlockCreatorResult] =
     Span[F].trace(ProcessDeploysAndCreateBlockMetricsSource) {
       val selfId         = ByteString.copyFrom(validatorIdentity.publicKey.bytes)
@@ -86,30 +87,31 @@ object BlockCreator {
               )
         } yield slashingDeploys
 
-      // TODO enable this when adding dummy deploys
-//      val prepareDummyDeploy: Seq[Signed[DeployData]] = dummyDeployOpt match {
-//        case Some(d) =>
-//          cs.getMaxBlockNumber >>= {vabn => Seq(
-//            ConstructDeploy.sourceDeployNow(
-//              source = d._2,
-//              sec = d._1,
-//              validAfterBlockNumber = vabn
-//            ))}
-//        case None => Seq.empty[Signed[DeployData]]
-//      }
+      def prepareDummyDeploy(blockNumber: Long): Seq[Signed[DeployData]] = dummyDeployOpt match {
+        case Some(d) =>
+          Seq(
+            ConstructDeploy.sourceDeployNow(
+              source = d._2,
+              sec = d._1,
+              vabn = blockNumber - 1
+            )
+          )
+        case None => Seq.empty[Signed[DeployData]]
+      }
 
       for {
         _ <- Log[F].info(
               s"Creating block #${nextBlockNum} (seqNum ${nextSeqNum})"
             )
         userDeploys     <- prepareUserDeploys(nextBlockNum)
+        dummyDeploys    = prepareDummyDeploy(nextBlockNum)
         slashingDeploys <- prepareSlashingDeploys(nextSeqNum)
         // make sure closeBlock is the last system Deploy
         systemDeploys = slashingDeploys :+ CloseBlockDeploy(
           SystemDeployUtil
             .generateCloseDeployRandomSeed(selfId, nextSeqNum)
         )
-        deploys = userDeploys -- s.deploysInScope //++ prepareDummyDeploy
+        deploys = userDeploys -- s.deploysInScope ++ dummyDeploys
         r <- if (deploys.nonEmpty || slashingDeploys.nonEmpty)
               for {
                 now           <- Time[F].currentMillis
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -88,11 +88,11 @@ object BlockCreator {
         } yield slashingDeploys
 
       def prepareDummyDeploy(blockNumber: Long): Seq[Signed[DeployData]] = dummyDeployOpt match {
-        case Some(d) =>
+        case Some((privateKey, term)) =>
           Seq(
             ConstructDeploy.sourceDeployNow(
-              source = d._2,
-              sec = d._1,
+              source = term,
+              sec = privateKey,
               vabn = blockNumber - 1
             )
           )
--- a/node/src/main/resources/defaults.conf
+++ b/node/src/main/resources/defaults.conf
@@ -5,6 +5,7 @@
 # 3. `required-signatures` = 0
 # for node to be able to create and approve genesis block if its not available
 standalone = false
+autopropose = false
 dev-mode = false
 
 protocol-server {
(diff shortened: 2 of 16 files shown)


nzpr:
Two tests failing are inherited from block-merge branch, see https://github.com/rchain/rchain/issues/3277

node/src/main/resources/defaults.conf
```diff
@@ -343,3 +344,10 @@ metrics {
   zipkin = false
   sigar = false
 }
+
+dev-mode = false
```
tgrospic, line 347: This seems duplicated `dev-mode` option?

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -180,6 +176,10 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
       descr = "ID of the RChain network to connect to."
     )
 
+    val autopropose = opt[Flag](
+      descr = "make node automatically trying to propose block after each new block added"
```
tgrospic: It seems that all other descriptions starts with capital letter.

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
```diff
@@ -87,6 +89,13 @@ object BlockProcessorInstance {
                 _ <- logFinished
               } yield ()
             }
+            .evalTap { v =>
+              // we don't care about result of propose here, whether its started or not.
+              // so let deferred be not used
+              (Deferred[F, Option[Int]] >>= { d =>
+                proposerQueue.enqueue1((v._1, d))
+              }).whenA(autopropose)
```
tgrospic, line 97: The same as with _deploy_ function, proposer can expose trigger auto propose function (_enqueue_) without leaking queue structure. It seems that it only needs instance of Casper.

node/src/main/scala/coop/rchain/node/configuration/model.scala
```diff
@@ -100,6 +102,10 @@ final case class Metrics(
     sigar: Boolean
 )
 
+final case class DevConf(
+    deployerPrivateKey: Option[String]
+)
```
tgrospic, line 107: Maybe _enabled_ field can be part of `DevConf` instead of _devMode_ option on root config?

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -86,30 +87,31 @@ object BlockCreator {
...
-//              validAfterBlockNumber = vabn
-//            ))}
-//        case None => Seq.empty[Signed[DeployData]]
-//      }
+      def prepareDummyDeploy(blockNumber: Long): Seq[Signed[DeployData]] = dummyDeployOpt match {
+        case Some(d) =>
+          Seq(
+            ConstructDeploy.sourceDeployNow(
+              source = d._2,
+              sec = d._1,
```
tgrospic: Please deconstruct tuples instead of shortcut `_x`.

node/src/main/resources/defaults.conf
nzpr: thanks, indeed!

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
nzpr: I'm not sure wdy mean, do you suggest to pass proposer instance inside block processor? I pass in only queue to minimize what I need to create BlockProcessorInstance, for better testability.

node/src/main/scala/coop/rchain/node/configuration/model.scala
nzpr: I'm not sure actually, @zsluedem wdy think?

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
tgrospic: This is all that is used here and on other places. 
```scala
def triggerPropose(casper: Casper[F]): F[Unit]
```

There is no need to push two arguments to a function and not use it. E.g why WebAPI needs to know about propose queue or auto propose flag if it's used there.

So what I mean is to create triggerPropose function in node runtime and supply it to the places where needed.

It would be even better that block API is instantiated with these data so that even less parameters are dragged through all these functions, but this is for future refactoring.

nzpr: I was thinking that the goal was to create streams/queues and connect different pieces of node together, so passing around queue for some worker entity (like proposer/block processor/etc..) as a generic principle is what we want.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -167,10 +167,6 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
       descr = "Start a stand-alone node."
     )
 
-    val devMode = opt[Flag](
-      descr = "Enable all developer tools."
```
zsluedem, line 171: why remove here?

nzpr: @zsluedem it is moved to the bottom if the file next to deployerPrivateKey option

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
tgrospic: Streams should be used for lazy data processing and concurrent control flow. The purpose of Queues is to be a bridge between push and pull data/events e.g. transforming callbacks to Streams. And different pieces are connected with functions.

But I'm talking about more basic principle to not supply parameters to a function which are not used. Here is more simplified example. `a && b` represents duplicated code.
```scala
def doWork(start: Boolean) = ()

// Both modules use parameters exclusively together
def module1(a: Boolean, b: Boolean) = doWork(a && b)
def module2(a: Boolean, b: Boolean) = doWork(a && b)

// Common part can be extracted and supply to both modules
def module1a(c: Boolean) = doWork(c)
def module2a(c: Boolean) = doWork(c)

// Values
val A = true
val B = false

module1(A, B)
module2(A, B)

// Modules doesn't have to know about specific A and B if they are not used separately
val C = A & B
module1a(C)
module2a(C)
```

nzpr: I'm not sure I agree here - block processor should be informed, that under certain conditions it should send proposer a command to propose. For now it is called autopropose, but in future we can make it more granular, or maybe do tha after N block added or in some probabilistic manner. The thing is that `triggerPropose` should be aware where propose request come from, and then potentially have some logic inside. That logic imo better to be inside caller code.

tgrospic: Usually we want to make components not tightly coupled, which in this case I see Queue and creation of Deferred as internal state of proposer that is shared. I don't even see how other component can use Queue for other purposes then to enqueue elements, even worse if someone try to dequeue it, it will break proposer logic.
If there is additional logic in block processor it can skip calling triggerPropose and also triggerPropose can be extended with additional parameters if needs data from other components.

But I think you will agree that this code is duplicated on three places which can be extracted to a helper function.

https://github.com/rchain/rchain/blob/36cb8acb009ad4f3afd578a41ce1a5fe7ba8ce44/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L70-L72

https://github.com/rchain/rchain/blob/36cb8acb009ad4f3afd578a41ce1a5fe7ba8ce44/node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala#L95-L97

https://github.com/rchain/rchain/blob/36cb8acb009ad4f3afd578a41ce1a5fe7ba8ce44/node/src/main/scala/coop/rchain/node/instances/ProposerInstance.scala#L69-L71

node/src/main/scala/coop/rchain/node/configuration/model.scala
zsluedem: Oh. I actually did some comments a few days before. But it is gone. 
Normally I agree with @tgrospic .

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
zsluedem: Just want to say I agree with @tgrospic on this.

zsluedem:
I am thinking maybe I should follow your path and merge it first then clear things up later. Others can not easily code something into this feature branch now.

zsluedem (approved):
I am going to approve this one and let's polish it later with some other prs. I am sorry to be picky on the details with previous pr. I found that I should work with your pace instead of my pace which would make it easier for me to make pr on the branch. In that way , I can help you fix tests, too.

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
nzpr: `even worse if someone try to dequeue it, it will break proposer logic` indeed, thanks, I will hide enqueuing behind some API

nzpr:
Merging this so @zsluedem can start merging dev into block-merge feature branch. Issue for leftover is here https://github.com/rchain/rchain/issues/3308

PR #3299, opened 2021-01-14
Enable parallel runtimes

### LINKED ISSUE #3096: Enable parallel execution of non-conflicting deploys
## Overview
This is the first step for block merge.
It includes 
- refactoring of RuntimeManager so it acts as a Runtime producer that node can use to execute some rholang code.
- making cost-accounting work with such Runtimes.

nzpr:
This PR enables parallel Runtimes to be constructed, which is a requirement for block merge.

This is a very simple minimal working fix, heavy refactoring of RSpace/ReplayRspace/RSpaceOps still needed.

The commit with adding syntax is not related to this, but as this PR is an attempt to split the huge https://github.com/rchain/rchain/pull/3238, it is small enough, so I decided to make it accompany the parallel runtimes commit.

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
@@ -315,6 +315,15 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
     comms.tailRecM(go).map(_.toOption)
   }
 
+  def spawn: F[IReplaySpace[F, C, P, A, K]] = {
+    val historyRep  = historyRepositoryAtom.get()
+    implicit val ck = serializeK.toSizeHeadCodec
+    for {
+      nextHistory <- historyRep.reset(historyRep.history.root)
+      hotStore    <- HotStore.empty(nextHistory)
+      _           <- restoreInstalls()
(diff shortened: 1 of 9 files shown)


nzpr:
Two tests failing are inherited from block-merge branch, see #3277

rspace/src/main/scala/coop/rchain/rspace/ReplayRSpace.scala
```diff
@@ -315,6 +315,16 @@ class ReplayRSpace[F[_]: Sync, C, P, A, K](
     comms.tailRecM(go).map(_.toOption)
   }
 
+  def spawn: F[IReplaySpace[F, C, P, A, K]] = {
```
zsluedem, line 318: I think we should keep the idea of manager to spawn new instance like KeyValueStoreManger and RUntimeManager.

zsluedem:
I am thinking maybe I should follow your path and merge it first then clear things up later. Others can not easily code something into this feature branch now.

zsluedem (approved):
I am going to approve this one and let's polish it later with some other prs. I am sorry to be picky on the details with previous pr. I found that I should work with your pace instead of my pace which would make it easier for me to make pr on the branch. In that way , I can help you fix tests, too.

PR #3301, opened 2021-01-20
Make tips return lca as well

nzpr:
## Overview
Fork choice tips are ranked tips of the DAG. They do not make sense without knowing LCA, which were used to rank tips.
This PR package LCA next to tips returned by estimator.

This is useful for easier management of data and doing less work when preparing casper snapshot.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -100,6 +100,8 @@ object MultiParentCasper extends MultiParentCasperInstances {
   */
 final case class CasperSnapshot[F[_]](
     dag: BlockDagRepresentation[F],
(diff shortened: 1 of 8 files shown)


nzpr:
Two tests failing are inherited from block-merge branch, see #3277

tgrospic (approved):
Very nice!

9rb:
Is lca in this context the same as nearest common parent or nearest common state ? It's better to use a more complete and accurate name since l and c in lca can be misinterpreted or interpreted in multiple ways. It slips my mind now, but the last time we discussed this, we added a further clarification / qualification / specificity that was helpful and self-explanatory

nzpr:
@9rb I think it is `latest common ancestor`. But yes, having L in LCA match also lowest, or least, brings some uncertainty.

PR #3304, opened 2021-01-22
Update block message field with rejectedDeploys

### LINKED ISSUE #3150: New BlockMessage data structure
## Overview

Below is how the current blockMessage structure looks like https://github.com/rchain/rchain/blob/cf6f2da2d1b8b30e953f60bff21cc5cde7cc49be/models/src/main/protobuf/CasperMessage.proto#L87-L99 .

Based on the [discussion](https://github.com/rchain/rchip-proposals/issues/23#issuecomment-697369252), currently the blockMessage data structure doesn't get the ability to describe whether the deploy is rejected in merging blocks. RChain should introduce some addtional messages in the block for that.

Besides that, there were discussions between the devs (@tgrospic , @nzpr and @zsluedem)  about the bonding info in the blockMessage doesn't contain `isActive` message which is very important. 

**The changes of the blockMessage should go into [RChips](https://github.com/rchain/rchip-proposals), too.**

### Impact

The blockMessage should be changed and the explorer or any infrastructures relate to blockMessage should be impacted.

This kind of update should be a **hard-fork**.

### Design

1. introduce  `reject deploy` field in the blockMessage
2. introduce `isActive` field in bondInfo in the blockMessage 
3. split eventLogs into precharge, refund and user
4. MergerStateHash?(need discussion)

zsluedem:
## Overview

1. introduce isActive field in bondInfo in the blockMessage
2. introduce reject deploy field in the blockMessage

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -670,7 +670,10 @@ object BlockAPI {
       blockSize = block.toProto.serializedSize.toString,
       deployCount = block.body.deploys.length,
       faultTolerance = faultTolerance,
-      justifications = block.justifications.map(ProtoUtil.justificationsToJustificationInfos)
+      justifications = block.justifications.map(ProtoUtil.justificationsToJustificationInfos),
+      rejectedDeploys = block.body.rejectedDeploys.map(
+        r => RejectedDeployInfo(PrettyPrinter.buildStringNoLimit(r.sig))
+      )
     ).pure[F]
 
   // Be careful to use this method , because it would iterate the whole indexes to find the matched one which would cause performance problem
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -605,7 +605,7 @@ object BlockAPI {
       constructor: (
           BlockMessage,
           Float
-      ) => F[A]
+      ) => A
   )(implicit casper: MultiParentCasper[F]): F[A] =
     for {
       dag <- casper.blockDag
@@ -622,34 +622,34 @@ object BlockAPI {
       initialFault   <- casper.normalizedInitialFault(ProtoUtil.weightMap(block))
       faultTolerance = normalizedFaultTolerance - initialFault
 
-      blockInfo <- constructor(block, faultTolerance)
+      blockInfo = constructor(block, faultTolerance)
     } yield blockInfo
 
   private def getFullBlockInfo[F[_]: Monad: SafetyOracle: BlockStore](
       block: BlockMessage
   )(implicit casper: MultiParentCasper[F]): F[BlockInfo] =
-    getBlockInfo[BlockInfo, F](block, constructBlockInfo[F])
+    getBlockInfo[BlockInfo, F](block, constructBlockInfo)
   private def getLightBlockInfo[F[_]: Monad: SafetyOracle: BlockStore](
       block: BlockMessage
   )(implicit casper: MultiParentCasper[F]): F[LightBlockInfo] =
-    getBlockInfo[LightBlockInfo, F](block, constructLightBlockInfo[F])
+    getBlockInfo[LightBlockInfo, F](block, constructLightBlockInfo)
 
-  private def constructBlockInfo[F[_]: Monad: SafetyOracle: BlockStore](
+  private def constructBlockInfo(
       block: BlockMessage,
       faultTolerance: Float
-  ): F[BlockInfo] =
-    for {
-      lightBlockInfo <- constructLightBlockInfo[F](block, faultTolerance)
-      deploys        = block.body.deploys.toSeq.map(_.toDeployInfo)
-    } yield BlockInfo(
+  ): BlockInfo = {
+    val lightBlockInfo = constructLightBlockInfo(block, faultTolerance)
+    val deploys        = block.body.deploys.map(_.toDeployInfo)
+    BlockInfo(
       blockInfo = Some(lightBlockInfo),
       deploys = deploys
     )
+  }
 
-  private def constructLightBlockInfo[F[_]: Monad: SafetyOracle: BlockStore](
+  private def constructLightBlockInfo(
       block: BlockMessage,
       faultTolerance: Float
-  ): F[LightBlockInfo] =
+  ): LightBlockInfo =
     LightBlockInfo(
       blockHash = PrettyPrinter.buildStringNoLimit(block.blockHash),
       sender = PrettyPrinter.buildStringNoLimit(block.sender),
@@ -674,7 +674,7 @@ object BlockAPI {
       rejectedDeploys = block.body.rejectedDeploys.map(
         r => RejectedDeployInfo(PrettyPrinter.buildStringNoLimit(r.sig))
       )
-    ).pure[F]
+    )
 
   // Be careful to use this method , because it would iterate the whole indexes to find the matched one which would cause performance problem
   // Trying to use BlockStore.get as much as possible would more be preferred
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -131,6 +131,7 @@ object BlockCreator {
                   preStateHash,
                   postStateHash,
                   processedDeploys,
+                  rejectedDeploys,
                   processedSystemDeploys
                 )             = checkpointData
                 newBonds      <- runtimeManager.computeBonds(postStateHash)
@@ -144,6 +145,7 @@ object BlockCreator {
                   preStateHash,
                   postStateHash,
                   processedDeploys,
+                  rejectedDeploys,
                   processedSystemDeploys,
                   newBonds,
                   shardId,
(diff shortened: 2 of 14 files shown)


models/src/main/protobuf/CasperMessage.proto
```diff
@@ -229,6 +235,7 @@ message PeekProto {
 message BondProto {
   bytes validator = 1;
   int64 stake     = 2;
+  bool  isActive  = 3;
```
tgrospic: We cannot set this on previous blocks so after decoding this field will have default value, which is for _bool_ `false` .
https://developers.google.com/protocol-buffers/docs/proto3#default

Maybe it is sufficient to call it something like `isSlashed` or `deactivated` and can be backward compatible.

zsluedem: Do you think we need `isSlashed` field?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -670,7 +670,10 @@ object BlockAPI {
       blockSize = block.toProto.serializedSize.toString,
       deployCount = block.body.deploys.length,
       faultTolerance = faultTolerance,
-      justifications = block.justifications.map(ProtoUtil.justificationsToJustificationInfos)
+      justifications = block.justifications.map(ProtoUtil.justificationsToJustificationInfos),
+      rejectedDeploys = block.body.rejectedDeploys.map(
+        r => RejectedDeployInfo(PrettyPrinter.buildStringNoLimit(r.sig))
+      )
```
tgrospic, line 676: Now that you editing this function, please convert it to be referentially transparent.

Wrapping with _pure_ means that LightBlockInfo will be created when function is called and not when is executed so it needs to be wrapped in _delay_.
The rule should be: _pure_ for values and simple expressions that cannot throw exceptions (like 2 + 2) and _delay_ for terms that needs additional reductions.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -192,38 +205,47 @@ object InterpreterUtil {
...
+    (StateHash, StateHash, Seq[ProcessedDeploy], Seq[ProcessedDeploy], Seq[ProcessedSystemDeploy])
+  ] =
     spanF.trace(ComputeDeploysCheckpointMetricsSource) {
       for {
         nonEmptyParents <- parents.pure
                             .ensure(new IllegalArgumentException("Parents must not be empty"))(
                               _.nonEmpty
                             )
-        preStateHash <- computeParentsPostState(nonEmptyParents, s, runtimeManager)
+        preStateHashReject              <- computeParentsPostState(nonEmptyParents, s, runtimeManager)
```
tgrospic: _Reject_ suffix is now little confusing, maybe `computedParentsInfo` is more descriptive.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -178,6 +178,11 @@ message BodyProto {
   repeated ProcessedDeployProto deploys      = 2;
   repeated ProcessedSystemDeployProto systemDeploys = 3;
   bytes                    extraBytes   = 4;
+  repeated RejectedDeployProto rejectedDeploys = 5;
```
tgrospic, line 181: Do we need additional type or it can be just `repeated bytes`. Do you know if there is a difference in serialized data?

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -229,6 +234,7 @@ message PeekProto {
 message BondProto {
   bytes validator = 1;
   int64 stake     = 2;
+  bool  deactivated  = 3;
```
tgrospic: If we are not using this right now, maybe it makes sense to postpone adding it later. It will not cause incompatibility in any way.

models/src/main/protobuf/CasperMessage.proto
zsluedem: Original plan is to also put the blockHash in the RejectedDeployProto. But it is difficult to get blockHash with current structure.

zsluedem: What do you think adding blockHash?

I got this idea because one deploy can possible be in two blocks(or more) when the deploy is not finalized. This design is too sticky.

tgrospic: For now let's keep it without block hash, but I think we will need additional index for quick lookup of deploy status.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -61,16 +64,34 @@ object InterpreterUtil {
...
-      _                  <- Span[F].mark("before-unsafe-get-parents")
-      parents            <- ProtoUtil.getParents(block)
-      _                  <- Span[F].mark("before-compute-parents-post-state")
-      preStateHashEither <- computeParentsPostState(parents, s, runtimeManager).attempt
-      _                  <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(block)}.")
-      result <- preStateHashEither match {
+      _                          <- Span[F].mark("before-unsafe-get-parents")
+      parents                    <- ProtoUtil.getParents(block)
+      _                          <- Span[F].mark("before-compute-parents-post-state")
+      preStateHashRejectedEither <- computeParentsPostState(parents, s, runtimeManager).attempt
```
tgrospic: Here is similar situation as with `computedParentsInfo`.

tgrospic (approved):
LGTM

models/src/main/protobuf/CasperMessage.proto
nzpr: I guess we did not talk about how rejected deploy will actually be used, so just want to put some thoughts.

This rejected deploy will be involved when: 
1) when preparing unfinalized deploys to be put in new block. We might cancel deploy that is already found in scope, because of rejection 
and 
2) when removing deploys from deploy store on finalize. We might not remove deploy if rejection is found in the scope of DAG we are finalizing on.

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -165,13 +167,15 @@ object BlockCreator {
...
       systemDeploys: Seq[ProcessedSystemDeploy],
       bondsMap: Seq[Bond],
       shardId: String,
       version: Long
   ): BlockMessage = {
-    val state  = RChainState(preStateHash, postStateHash, bondsMap.toList, blockData.blockNumber)
-    val body   = Body(state, deploys.toList, systemDeploys.toList)
+    val state = RChainState(preStateHash, postStateHash, bondsMap.toList, blockData.blockNumber)
+    val body =
+      Body(state, deploys.toList, List.empty, systemDeploys.toList)
```
nzpr: should not this `List.empty` be actually `rejectedDeploys`?

zsluedem: very good catch

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -61,15 +63,15 @@ object InterpreterUtil {
...
-      _                  <- Span[F].mark("before-unsafe-get-parents")
-      parents            <- ProtoUtil.getParents(block)
-      _                  <- Span[F].mark("before-compute-parents-post-state")
-      preStateHashEither <- computeParentsPostState(parents, s, runtimeManager).attempt
-      _                  <- Log[F].info(s"Computed parents post state for ${PrettyPrinter.buildString(block)}.")
-      result <- preStateHashEither match {
+      _                   <- Span[F].mark("before-unsafe-get-parents")
+      parents             <- ProtoUtil.getParents(block)
+      _                   <- Span[F].mark("before-compute-parents-post-state")
+      computedParentsInfo <- computeParentsPostState(parents, s, runtimeManager).attempt
```
nzpr, line 70: I'd better call this `preStateComputed`, which should include preStateHash and deploys rejected when merging. And maybe make a `PreState` case class to describe result of computing of preState for next state transition. But I'm ok with this, thought would be nice to make code more verbose.

zsluedem: @tgrospic @nzpr Get consensus. I am not really good at naming. Actually now it is the right time for me to vote to get a consensus in a three-node(person) network. Because i am lazy and don't want to change, i vote for `computedParentsInfo`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -622,34 +622,34 @@ object BlockAPI {
...
+    val lightBlockInfo = constructLightBlockInfo(block, faultTolerance)
+    val deploys        = block.body.deploys.map(_.toDeployInfo)
+    BlockInfo(
       blockInfo = Some(lightBlockInfo),
       deploys = deploys
     )
+  }
 
-  private def constructLightBlockInfo[F[_]: Monad: SafetyOracle: BlockStore](
+  private def constructLightBlockInfo(
```
nzpr, line 649: nice to get rid of [F] where its not needed

nzpr (approved):
@tgrospic please vote here, if you think it does not worth efforts, lets merge. https://github.com/rchain/rchain/pull/3304#discussion_r598290964

zsluedem:
bors merge

PR #3270, opened 2020-12-16
Refactor hotstore

nzpr:
This PR is an attempt to fix invalid tuplespace mismatch issue.

PR #3312, opened 2021-01-29
BlockMerge optimisation

### LINKED ISSUE #3308: Address comments on PR #3298
https://github.com/rchain/rchain/pull/3298#discussion_r559170112

nzpr:
This PR continues work on block merge.

These are results for merging 19 blocks from peers. This is very first merge, looks like JVM were not warmed up, so numbers are slightly bigger than with averaging.

```2021-02-25 17:33:42,307 [INFO ] [node-runner-85      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 278 consumes [1.051922584 sec]
2021-02-25 17:33:42,318 [INFO ] [node-runner-85      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 310 consumes [1.148264846 sec]
2021-02-25 17:33:42,369 [INFO ] [node-runner-64      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 350 consumes [1.137562705 sec]
2021-02-25 17:33:42,391 [INFO ] [node-runner-64      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 302 consumes [1.212630647 sec]
2021-02-25 17:33:42,426 [INFO ] [node-runner-82      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 306 consumes [1.193673457 sec]
2021-02-25 17:33:42,474 [INFO ] [node-runner-50      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 314 consumes [1.225670189 sec]
2021-02-25 17:33:42,477 [INFO ] [node-runner-81      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 338 consumes [1.266100753 sec]
2021-02-25 17:33:42,512 [INFO ] [node-runner-81      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 286 consumes [1.256947788 sec]
2021-02-25 17:33:42,515 [INFO ] [node-runner-82      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 354 consumes [1.317729183 sec]
2021-02-25 17:33:42,537 [INFO ] [node-runner-50      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 298 consumes [1.26495599 sec]
2021-02-25 17:33:42,545 [INFO ] [node-runner-85      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 342 consumes [1.326230389 sec]
2021-02-25 17:33:42,547 [INFO ] [node-runner-91      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 334 consumes [1.318260285 sec]
2021-02-25 17:33:42,556 [INFO ] [node-runner-59      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 322 consumes [1.296293919 sec]
2021-02-25 17:33:42,568 [INFO ] [node-runner-85      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 282 consumes [1.422848821 sec]
2021-02-25 17:33:42,568 [INFO ] [node-runner-90      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 346 consumes [1.291053695 sec]
2021-02-25 17:33:42,575 [INFO ] [node-runner-91      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 330 consumes [1.30369556 sec]
2021-02-25 17:33:42,582 [INFO ] [node-runner-87      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 318 consumes [1.205777585 sec]
2021-02-25 17:33:42,584 [INFO ] [node-runner-58      ] [c.r.r.m.i.DiffStateMerger    ]- compute changes: 32 produces + 326 consumes [1.198131412 sec]
2021-02-25 17:33:47,752 [INFO ] [node-runner-53      ] [c.r.r.m.i.DiffStateMerger    ]- process trieActions [5.150237549 sec]
2021-02-25 17:33:47,752 [INFO ] [node-runner-53      ] [c.r.c.b.m.CasperDagMerger$   ]- Merging done. [8.342220589 sec]
```

NOTE: process trieActions takes quite a lot of time - this is because 
https://github.com/rchain/rchain/blob/5a9b6d418ebb4cc86781ae69070d629491a6414a/rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala#L318-L327
here history actions processing is sequential. It makes sense at least for testing environment to use parallel processing.

UPD. with parallel processing of trieActions `process trieActions [1.443238755 sec]`

For historical purposes, let me record here results of testing on network of 10, 20, 30 and 40 nodes at this commit https://github.com/rchain/rchain/pull/3312/commits/1748cc6d7ea98d1ad75f80bf891e3d685b8be9b5, **but with parallel processing of trie actions**. This is quite raw estimate, but this is what we have now. 
<p align="center">Screenshot 2021-02-26 at 18 27 45</p>

Looks like conflict detection part grows exponentially, and thats reasonable - now we check all branches combinations . One solution is to use map-reduce, also caching might be implemented.

closes https://github.com/rchain/rchain/issues/3308
closes https://github.com/rchain/rchain/issues/3277 (except reimplement tests disabled in 53578aa, if applicable)


casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -26,59 +32,138 @@ object CasperDagMerger {
...
+
+      _ <- Log[F].info(s"Rejecting ${toReject.size} deploys").whenA(toReject.nonEmpty)
+
+      // main branch
+      main = branchesIndexed.head
+
+      // merging branches, rejecting conflicting deploys
+      toMerge = branchesIndexed.tail.map(
+        m =>
+          (
```
tgrospic: Redundant parenthesis.

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -26,59 +32,138 @@ object CasperDagMerger {
...
+          )
+      v <- Span[F].trace(calculatingConflictsSpanLabel)(
+            Stopwatch.time(Log[F].info(_))(s"Calculating conflicts done")(
+              branchesIndexed.foldLeftM(
+                (branchesIndexed, Set.empty[(Set[Blake2b256Hash], Set[ProcessedDeploy])])
+              ) {
+                case ((biggestBranch :: remainder, result), _) =>
+                  for {
+                    conflicting <- fs2.Stream
+                                    .emits(remainder.map { b =>
```
tgrospic: I'd suggest that we deconstruct tuples in these cases for easier code readability. `b._3` is mentioned twice so to understand what it is much larger chunk of the code must be kept in mind.

rspace/src/main/scala/coop/rchain/rspace/history/instances/CachingHashHistoryReaderImpl.scala
```diff
@@ -0,0 +1,127 @@
...
+        trie match {
+          case LeafPointer(dataHash) => leafStore.get(dataHash)
+          case EmptyPointer          => Applicative[F].pure(None)
+          case _ =>
+            Sync[F].raiseError(new RuntimeException(s"unexpected data at key $key, data: $trie"))
+
+        }
+    }
+
+  override def root: Blake2b256Hash = h.root
```
tgrospic: Let's have a rule for naming with single letter only in a _small_ scope to make the code more readable.

```scala
// OK
xs.map(x => x.doSomething)

// Not OK for class parameters (usually large scope)
class Foo(x: Bar)  {
  def doFooThing = x.doSomething
}

// Not OK for named functions
def foo(x: Bar) = {
  ... a lot of code
  val y = x.doSomething
}
```

rspace/src/main/scala/coop/rchain/rspace/history/HistoryReader.scala
```diff
@@ -0,0 +1,81 @@
...
+trait HashHistoryReader[F[_], C, P, A, K]
+    extends HistoryReader[F, C, P, A, K, Blake2b256Hash, Blake2b256Hash, Blake2b256Hash] {
+
+  /**
+    * transform reader into [[RhoHistoryReader]]
+    *
+    * @param codecC codec for serializing channel
+    * @return
+    */
+  def toRho(
```
nzpr, line 51: Sorry for OOP fever, but thinking how to make it more pretty turned into a rabbit hole for me. Lets maybe arrange something later.

casper/src/test/scala/coop/rchain/casper/batch2/EstimatorHelperTest.scala
```diff
@@ -1,150 +1,149 @@
...
-    )
-
-  private def consume(channelsHash: ByteString) =
-    ConsumeEvent(
-      List(channelsHash),
-      hash = ByteString.copyFromUtf8("Asdfg753213fdsadfueias9fje35mv43"),
-      persistent = false
-    )
-}
+//package coop.rchain.casper.batch2
```
nzpr: Maybe it's better to remove this completely, I'm not sure if this code is any useful now.

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
```diff
@@ -100,6 +100,9 @@ package object effects {
               }
         } yield r
 
+      override def trace[A](tag: String)(block: F[A])(implicit parentSource: Source): F[A] =
```
nzpr: PTAL at this new way to create traces, without need to always create explicit source for each trace.

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
```diff
@@ -1,32 +1,24 @@
...
+import fs2.concurrent.Queue
 
 object BlockProcessorInstance {
   def create[F[_]: Concurrent: Log](
       blocksQueue: Queue[F, (Casper[F], BlockMessage)],
       blockProcessor: BlockProcessor[F],
       state: Ref[F, Set[BlockHash]],
-      proposerQueue: Queue[F, (Casper[F], Deferred[F, Option[Int]])],
-      autopropose: Boolean
+      triggerProposeF: Option[ProposeFunction[F]]
```
nzpr, line 20: This, and hiding stream behind one `triggerProposeF` method in other places should close https://github.com/rchain/rchain/issues/3308

node/src/main/scala/coop/rchain/node/runtime/APIServers.scala
```diff
@@ -0,0 +1,66 @@
+package coop.rchain.node.runtime
```
nzpr, line 1: Now NodeRuntime is less monstrous

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -286,25 +287,25 @@ private class InMemHotStore[F[_]: Sync, C, P, A, K](
...
       ck: Codec[K]
   ): HotStore[F, C, P, A, K] =
     new InMemHotStore[F, C, P, A, K]
 
   def from[F[_], C, P, A, K](
       cache: Cache[C, P, A, K],
-      historyReader: HistoryReader[F, C, P, A, K]
-  )(implicit ck: Codec[K], sync: Sync[F]) =
+      historyReader: RhoHistoryReader[F, C, P, A, K]
+  )(implicit ck: Codec[K], concurrent: Concurrent[F]) =
```
nzpr, line 295: this Concurrent is not really necessary, but it will be required for HotStore refactoring (attempts are in https://github.com/rchain/rchain/pull/3271 and https://github.com/rchain/rchain/pull/3270), so I just left this in PR.

rspace/src/main/scala/coop/rchain/rspace/history/package.scala
```diff
@@ -0,0 +1,152 @@
+package coop.rchain.rspace
```
nzpr, line 1: Please review this file carefully. We have in codebase tuple space hash for empty trie hardcoded, so this should serve as a protection for encoding corruption. 
https://github.com/rchain/rchain/blob/1724688c15cda74dad2c8dd26302ea6a7a880334/casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala#L218-L222
Might be there should be a test for that, because encoding tests did not catch bug that I had in this code - only by empty trie hash mismatch I figured out whats wrong.

rspace/src/main/scala/coop/rchain/rspace/history/syntax/HistoryRepositorySyntax.scala
```diff
@@ -29,169 +18,54 @@ trait HistoryRepositorySyntax {
...
-    }
+    historyRepo.getHistoryReader(state).getContinuationFromChannelHash(channelHash)(historyRepo)
 
-  /**
-    * detect conflict events between rightEvents and rightEvents.
-    * In conflict case, conflict set contains the conflict channel hash.
-    * @param baseState baseState needed here for detect conflict in joins
-    * @return
-    */
-  def isConflict(
```
nzpr, line 150: This code is legacy conflict detector, so removed

rspace/src/main/scala/coop/rchain/rspace/internal.scala
```diff
@@ -13,6 +13,9 @@ import scala.collection.SortedSet
 @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
 object internal {
 
+  /** helper class to package data and serialization (encoded by scodec) */
+  final case class Encoded[D](item: D, byteVector: ByteVector)
```
nzpr, line 17: Having ByteVector next to Scala object helps greatly when doing diffs on pair of vectors.

rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala
```diff
@@ -1,55 +1,42 @@
...
-    startState: Blake2b256Hash,
-    endState: Blake2b256Hash,
+/**
+  * Sequence of events along with data about state.
+  * @param startState start state (History root)
+  * @param endState end state (History root)
+  * @param events events that transition state from start to end
+  */
+final case class EventChain[F[_]](
+    startState: History[F],
```
nzpr, line 31: I put History here, as eventually I believe we should expose reader interface on History and get rid of pointer to current history in historyRepo.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
+package coop.rchain.rspace.merger.instances
```
nzpr, line 1: Sorry for gigantic file, but there are 4 parts here - 
1. preparation of Refs to throw workers on them
2. preparation of worker streams
3. populating Refs
4. Calculating storage actions from Refs 
Maybe split this later when we have more abstract merging API.

nzpr:
Sorry for lots of changes, I'm trying to put a handle on that :)

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
tgrospic: I've mentioned this in the first bullet on #3001. Macro to support this is already created. There are maybe some cases where you want to differentiate multiple traces inside one function (?).

nzpr: ` There are maybe some cases where you want to differentiate multiple traces inside one function (?).` I'm not sure, I assume that if nested calls should give nested traces. So you create implicit parent source and then able to use this method to just provide tag one level down. But how to structure tracing in best way is a bit blurry for me.

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -2,83 +2,214 @@ package coop.rchain.casper.blocks.merger
...
       base: MergingVertex,
-      dag: DagReader[F, MergingVertex]
+      dag: DagReader[F, MergingVertex],
+      blockIndexCache: LazyKeyValueCache[F, MergingVertex, BlockIndex]
   ): F[(StateHash, Seq[ProcessedDeploy])] = {
 
-    // ordering is required for computing branches procedure to be replayable
+    // ordering is required to be able to replay computing branches procedure
     implicit val ord: Ordering[MergingVertex] = (x: MergingVertex, y: MergingVertex) =>
       Base16.encode(x.postStateHash.toByteArray).compare(Base16.encode(y.postStateHash.toByteArray))
```
tgrospic: Encoding to string can be avoided and at least compare bytes. Right now I don't see better solution, ByteString only have lexical comparison which should be slower.
`java.util.Arrays.compare(x.postStateHash.toByteArray, y.postStateHash.toByteArray)`

casper/src/main/scala/coop/rchain/casper/blocks/merger/package.scala
```diff
@@ -0,0 +1,8 @@
+package coop.rchain.casper.blocks
+
+import coop.rchain.metrics.Metrics
+
+package object merger {
+  val MergingMetricsSource: Metrics.Source =
+    Metrics.Source(Metrics.BaseSource, "merging")
```
tgrospic: Why this cannot be in CasperDagMerger with child sources?

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -616,55 +622,57 @@ final class ReplayRhoRuntimeOps[F[_]: Sync: Span: Log](
...
-                        refundAmount,
-                        SystemDeployUtil.generateRefundDeployRandomSeed(processedDeploy.deploy)
-                      ),
-                      None
-                    ).map(_ => succeeded)
+                    for {
+                      _ <- Span[F].mark("deploy-eval-done")
+                      r <- runtime.createSoftCheckpoint
+                            .whenA(evalResult.succeeded)
+                            .map(_ => evalResult.succeeded)
```
tgrospic: Mapping with ignore has function `.as(evalResult.succeeded)`.

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -616,55 +622,57 @@ final class ReplayRhoRuntimeOps[F[_]: Sync: Span: Log](
...
                 )
+                .flatTap(_ => EitherT.liftF(Span[F].mark("refund-started")))
+                .flatTap { succeeded =>
+                  replaySystemDeployInternal(
+                    new RefundDeploy(
+                      refundAmount,
+                      SystemDeployUtil.generateRefundDeployRandomSeed(processedDeploy.deploy)
+                    ),
+                    None
+                  ).map(_ => succeeded)
```
tgrospic: Same here.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -91,7 +98,9 @@ object InterpreterUtil {
       block: BlockMessage,
       dag: BlockDagRepresentation[F],
       runtimeManager: RuntimeManager[F]
-  )(implicit spanF: Span[F]): F[Either[ReplayFailure, StateHash]] =
+  )(
+      implicit spanF: Span[F]
+  ): F[Either[ReplayFailure, StateHash]] =
```
tgrospic: Is there a change here?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -207,7 +216,9 @@ object InterpreterUtil {
       parents: Seq[BlockMessage],
       s: CasperSnapshot[F],
       runtimeManager: RuntimeManager[F]
-  )(implicit spanF: Span[F]): F[StateHash] =
+  )(
+      implicit spanF: Span[F]
+  ): F[StateHash] =
```
tgrospic: Same here.

casper/src/test/scala/coop/rchain/casper/batch2/EstimatorHelperTest.scala
tgrospic: I don't think we need this.

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
tgrospic: I'm thinking on multiple traces on the same level with the same parent source. Basically, source can be derived from the code itself in which case any manual definition can be avoided.

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
tgrospic: Is the result of the `ProposeFunction` used anywhere?

node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
```diff
@@ -0,0 +1,580 @@
...
+      adminWebApi: AdminWebApi[F],
+      proposer: Option[Proposer[F]],
+      proposeRequestsQueue: Queue[F, (Casper[F], Deferred[F, Option[Int]])],
+      triggerProposeFOpt: Option[ProposeFunction[F]],
+      proposerStateRef: Ref[F, ProposerState[F]],
+      blockProcessor: BlockProcessor[F],
+      blockProcessingState: Ref[F, Set[BlockHash]],
+      incomingBlocksQueue: Queue[F, (Casper[F], BlockMessage)]
+  )(
+      )(
```
tgrospic: Unnecessary parenthesis. Idea is complaining but sbt compiles. :)

node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
```diff
@@ -0,0 +1,580 @@
...
+      // 5. run the node program.
+      program = {
+        implicit val tr = transport
+        implicit val cn = rpConfAsk
+        implicit val cc = rpConnections
+        implicit val cs = rpConfState
+        implicit val pn = peerNodeAsk
+        implicit val ks = kademliaStore
+        implicit val nd = nodeDiscovery
+        //implicit val bs = blockStore
```
tgrospic: Can this be removed?

node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
```diff
@@ -0,0 +1,580 @@
...
+          nodeConf,
+          dagConfig,
+          blockstorePath,
+          lastFinalizedStoragePath,
+          eventBus,
+          deployStorageConfig
+        )
+      }
+      (
+        blockStore,
```
tgrospic: Block store is not used. Can be removed from setupNodeProgram.

casper/src/main/scala/coop/rchain/casper/blocks/merger/package.scala
nzpr: I've seen metric sources for particular package in other package definitions, e.g. in `blockstorage`. I'm not sure though this is the right way to do it.

node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala
nzpr: Thanks! I did not finish work actually, there were still propose queue passed down to clients in some cases.
Ptal https://github.com/rchain/rchain/commit/bead752fe15dc44abed844563f08bfa8eb8226b0

Yes, `ProposeFunction` should be defined in casper module and return value is used propose API https://github.com/rchain/rchain/pull/3312/commits/862bc47689cb8753af67088821ac8ebee9a663a0#diff-71a7b671734067220382fb1243b725e73f70e08ad9e78d93c6bdcd23f3d9fb84R30

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
nzpr: so we are leaving it as it is?

node/src/main/scala/coop/rchain/node/diagnostics/effects/package.scala
nzpr: I see, so we should start using this macro

rspace/src/main/scala/coop/rchain/rspace/channelStore/syntax/ChannelStoreSyntax.scala
```diff
@@ -0,0 +1,80 @@
...
+trait ChannelStoreSyntax {
+  implicit final def syntaxChannelStore[F[_], C](
+      channelStore: ChannelStore[F, C]
+  ): ChannelStoreOps[F, C] =
+    new ChannelStoreOps[F, C](channelStore)
+}
+
+final case class DatumMapping(historyHash: Blake2b256Hash, eventLogHash: Blake2b256Hash)
+final case class ConsumeMapping(historyHash: Blake2b256Hash, eventLogHashes: Seq[Blake2b256Hash])
+final case class JoinMapping(joinHash: Blake2b256Hash, eventLogHashe: Blake2b256Hash)
```
zsluedem: typo. `eventLogHashe`

zsluedem: Another question on this JoinMapping. I don't see this is in use anywhere. I think that is because we do nothing on join now, right?

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
...
+  */
+final case class DiffStateMerger[F[_]: Concurrent: Log, C, P, A, K](
+    historyRepo: HistoryRepository[F, C, P, A, K],
+    serializeC: Serialize[C]
+) extends StateMerger[F] {
+
+  override def merge(mainState: History[F], toMerge: Seq[EventChain[F]]): F[History[F]] =
+    for {
+      // compute trie actions
+      trieActions <- Stopwatch.time(Log[F].debug(_))(
```
zsluedem, line 32: @tgrospic I think this one can be another implementation for `Span`. what do you think?

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
tgrospic: What I suggested is to use array comparison directly without conversion to string and comparing strings, which should be slower.

rspace/src/main/scala/coop/rchain/rspace/channelStore/syntax/ChannelStoreSyntax.scala
nzpr: @zsluedem nice catch! Logic of merger assumes that joins hashes are read on-demand with LazyCache
https://github.com/rchain/rchain/blob/79d42ba36857599a2fcc75ac8af18ec5b3e71184/rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala#L73-L87
 because joins changes computation depends on consume changes.
https://github.com/rchain/rchain/blob/79d42ba36857599a2fcc75ac8af18ec5b3e71184/rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala#L192-L197

So we preload only datums and konts changes, therefore `JoinMapping` is not used. But might be later it will find a use. 
https://github.com/rchain/rchain/blob/79d42ba36857599a2fcc75ac8af18ec5b3e71184/rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala#L127-L128

rspace/src/main/scala/coop/rchain/rspace/internal.scala
zsluedem: Helps in which way? Performance?

shared/src/main/scala/coop/rchain/store/LazyKeyValueCache.scala
```diff
@@ -25,6 +25,17 @@ class LazyKeyValueCache[F[_]: Concurrent, K, V] private[LazyKeyValueCache] (
           )
       r <- d.get
     } yield r
+
```
zsluedem, line 28: I think `LazyKeyValueCache` and `LazyAdHocKeyValueCache` are totally duplicates. Do I miss something?

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
```diff
@@ -0,0 +1,60 @@
...
+  def toMap: F[Map[K, V]]
+}
+
+final case class NoOpKeyValueCache[F[_]: Applicative, K, V]() extends KeyValueCache[F, K, V] {
+  override def get(key: K, fallback: F[V]): F[V] = fallback
+  override def put(key: K, value: V): F[Unit]    = ().pure[F]
+  override def toMap: F[Map[K, V]]               = Map.empty[K, V].pure[F]
+}
+
+class LazyAdHocKeyValueCache[F[_]: Concurrent, K, V] private[LazyAdHocKeyValueCache] (
```
zsluedem, line 20: I think `LazyKeyValueCache` and `LazyAdHocKeyValueCache` are totally duplicates. Do I miss something?

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
...
+      }
+      joinsActions = joinsOptActions
+        .filter(_.isDefined)
+        .map(_.get)
+
+//      _ = println(produceChanges)
+//      _ = println(consumeChanges)
+//      _ = println(produceActions)
+//      _ = println(consumeActions)
+//      _ = println(joinsActions)
```
zsluedem: Some leftovers.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
...
+        val mergeChanges = {
+          // only produces and consumes that are not volatile in the scope of event log are subject to merge
+          // volatile events cannot have any effect outside if the scope
+          val newActiveProduces =
+            prods.filter(
+              p => p.persistent || !comms.exists(c => c.produces.contains(p) && c.peeks.isEmpty)
+            )
+          val newActiveConsumes = conss.filter(c => c.persistent || !comms.exists(_.consume == c))
+          val destroyedExternalProduces =
+            comms.filter(_.peeks.isEmpty).flatMap(_.produces).filterNot(p => prods.contains(p))
```
zsluedem: what if the produce is persistent which should not be destroyed?

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
...
+          val destroyedExternalConsumes = comms.map(_.consume).filterNot(c => conss.contains(c))
```
zsluedem: what if the consume is persistent which should not be destroyed? same above.

rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
```diff
@@ -0,0 +1,18 @@
+package coop.rchain.rspace
+
+import cats.effect.Sync
+import cats.syntax.all._
+
+package object merger {
+  def computeChannelChange[F[_]: Sync, T](
```
zsluedem, line 7: Just want to say this is a very nice abstraction. Nice work.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
```diff
@@ -0,0 +1,317 @@
...
+      produceChanges     <- pDiffRef.get
+      consumeChanges     <- cDiffRef.get
+      joinsCumulative    <- jAccRef.get
+      channelsPerConsume <- channelsPerConsumesRef.get
+
+      // transform changes into actions over trie
+      produceActions = produceChanges.toList.par.map {
+        case (dataHash, changes) =>
+          val newDatumsAtProduce =
+            (initData(dataHash) ++ changes.added diff changes.removed).distinct
```
zsluedem: Why do we need distinct here? Is it because every datum is unique?

zsluedem: Another question for these `++` and `diff` in `changes` which is generated by parallel processing. Would the order be changed every time based on different execution? Is that gonna affect the dataHash because of the sequence order?

zsluedem: > Another question for these `++` and `diff` in `changes` which is generated by parallel processing. Would the order be changed every time based on different execution? Is that gonna affect the dataHash because of the sequence order?

Oh. I just found the answer that Seq[datum] would be sorted first before encoded by `encodeSortedRich` which can be ensure order which doesnt lead to different hash.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
nzpr: One of them requires function to populate cache value during object construction, another - when calling get method, because it might be not available when cache is created. I agree it is better to have one API, but I'm not sure whats the best way do to that.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr: Nice catch, we just do more work, but looks like it should be 
```scala 
val destroyedExternalProduces =
            comms                                // for all comm events
              .filter(_.peeks.isEmpty)           // peeks won't affect channel state
              .flatMap(_.produces)               // get produces
              .filterNot(p => prods.contains(p)) // that are created outside of this event chain
              .filterNot(_.persistent)           // and not persistent, therefore might be destroyed
```

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr:
```scala
          val destroyedExternalConsumes =
            comms                                // for all comm events
              .map(_.consume)                    // get consumes
              .filterNot(c => conss.contains(c)) // that are created outside of this event chain
              .filterNot(c => c.persistent)      // and not persistent, therefore might be destroyed
```

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
tgrospic: I see Span, Log, Metrics as a garbage inside the code. Doesn't really do anything, just makes code uglier. :)

But it's needed, so it should be minimally intrusive and just specify our intention. E.g. to measure time we just need to specify a block of code and unique identifier.

```scala
// We don't need to specify implementation
// It can be measured with Span and/or put inside the Log
_ <- Diag[F].measureTime(<unique-id>) {
  ...
}

// Or we just want to specify some event that can be configured
//  to put in the Log or Metrics endpoint
_ <- Diag[F].event(<unique-id>)
```
@zsluedem I'll create more elaborate info about that.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
tgrospic: @zsluedem Let's move refactoring of `KeyValueCache` to another PR. I created an issue #3340 to refactor get method. You can add additional stuff to that issue.

casper/src/main/scala/coop/rchain/casper/blocks/merger/package.scala
tgrospic: We already have metrics source defined in casper package so this can be moved there. We will save a file.
https://github.com/rchain/rchain/blob/6dd06305/casper/src/main/scala/coop/rchain/casper/package.scala#L13

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -616,55 +622,57 @@ final class ReplayRhoRuntimeOps[F[_]: Sync: Span: Log](
             SystemDeployUtil.generatePreChargeDeployRandomSeed(processedDeploy.deploy)
           ),
           expectedFailure
-        ).flatMap(
-          _ =>
+        ).flatTap(_ => EitherT.liftF(Span[F].mark("precharge-done")))
```
tgrospic, line 626: There is a method on EitherT to support returning `F` directly.
```scala
.semiflatTap(_ => Span[F].mark("refund-started"))
```

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -616,55 +622,57 @@ final class ReplayRhoRuntimeOps[F[_]: Sync: Span: Log](
...
-                    ).map(_ => succeeded)
+                    for {
+                      _ <- Span[F].mark("deploy-eval-done")
+                      r <- runtime.createSoftCheckpoint
+                            .whenA(evalResult.succeeded)
+                            .as(evalResult.succeeded)
+                      _ <- Span[F].mark("deploy-done")
+                    } yield r
                 )
+                .flatTap(_ => EitherT.liftF(Span[F].mark("refund-started")))
```
tgrospic, line 640: Same here.

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -616,55 +622,57 @@ final class ReplayRhoRuntimeOps[F[_]: Sync: Span: Log](
...
+                .flatTap { succeeded =>
+                  replaySystemDeployInternal(
+                    new RefundDeploy(
+                      refundAmount,
+                      SystemDeployUtil.generateRefundDeployRandomSeed(processedDeploy.deploy)
+                    ),
+                    None
+                  ).as(succeeded)
+                }
+                .flatTap(_ => EitherT.liftF(Span[F].mark("refund-done")))
```
tgrospic, line 650: And here.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr: `Why do we need distinct here? Is it because every datum is unique?` I feel that you are right, I think there is no need for this distinct. Need to think a more.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: I'm ok with this but in general what is not used it should be removed.

rspace/src/main/scala/coop/rchain/rspace/channelStore/syntax/ChannelStoreSyntax.scala
tgrospic: Please comment out the code if it's not used and why is left.

rspace/src/main/scala/coop/rchain/rspace/history/HistoryReader.scala
tgrospic: Why is this called `rho`?

rspace/src/main/scala/coop/rchain/rspace/history/package.scala
tgrospic: Please create an issue to write these tests and review it in detail.

shared/src/main/scala/coop/rchain/diag/Tracing.scala
```diff
@@ -0,0 +1,8 @@
+package coop.rchain.diag
+
+import coop.rchain.metrics.Metrics
+
+object Tracing {
+  val CasperMetricsSource: Metrics.Source = Metrics.Source(Metrics.BaseSource, "casper")
```
tgrospic: This is duplicated here and in casper project.

shared/src/main/scala/coop/rchain/diag/Tracing.scala
```diff
@@ -0,0 +1,8 @@
...
+  val RuntimeManagerMetricsSource         = Metrics.Source(CasperMetricsSource, "runtime-manager")
```
tgrospic: Both of these does not belong to shared project. It should be moved to the project where are used.

shared/src/main/scala/coop/rchain/metrics/Metrics.scala
```diff
@@ -8,6 +8,7 @@ import coop.rchain.metrics.Metrics.Source
 trait Span[F[_]] {
   def mark(name: String): F[Unit]
   def trace[A](source: Metrics.Source)(block: F[A]): F[A]
+  def trace[A](tag: String)(block: F[A])(implicit parentSource: Source): F[A]
```
tgrospic: What is the difference between these definitions (source and parentSource)? Why this is not just extension method?

shared/src/main/scala/coop/rchain/shared/Language.scala
```diff
@@ -11,4 +11,10 @@ object Language {
     val _: A = a
     ()
   }
+
+  def removeIndex[E](col: Seq[E], index: Int): Seq[E] = {
+    val (l1, l2) = col splitAt index
+    (l1 ++ l2.drop(1))
+  }
```
tgrospic, line 18: We have this function duplicated on three places. Also working with index on a Seq is not advisable because it can be slow (linear with elements).
For now at least we can use this one here until we investigate better solution.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
zsluedem: I am ok to refactor this in another pr. I thought I miss something.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr: Yes, you are right! I checked and tests failed if I remove `distinct`. Thats because I do double work - channel can be checked both agains produce and against consume, so I need to have here
https://github.com/rchain/rchain/blob/79d42ba36857599a2fcc75ac8af18ec5b3e71184/rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala#L148-L151

`changedProduceHashes.distinct`.

rspace/src/main/scala/coop/rchain/rspace/internal.scala
nzpr: @zsluedem yes, its faster

shared/src/main/scala/coop/rchain/diag/Tracing.scala
nzpr: I moved it here because to my understanding - we might need base tracing source from `parent` project, and we don't have access to source defined in that project.

zsluedem (approved):
Nice work.
It would be greate if we can cover more tests on new implementation which give me better feel how each part integrates. 
Actually some previous prs are not viewed nicely by myself. I am still in some kind of mist for some parts.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr: You are right, it should not be `distinct` here. But it should be in another place. We might have duplicate produces/consumes in event log if it is newly introduced persistent and immediately comm-ed. So I will move it to events computation

shared/src/main/scala/coop/rchain/metrics/Metrics.scala
nzpr: I agree, this should be extension. But looks like common understanding that we should use macro, so this will be removed. Am i correct?

shared/src/main/scala/coop/rchain/diag/Tracing.scala
nzpr: I removed that package completely

shared/src/main/scala/coop/rchain/metrics/Metrics.scala
tgrospic: Please create syntax, it should be also valid when we switch to macro.

rspace/src/main/scala/coop/rchain/rspace/merger/instances/DiffStateMerger.scala
nzpr: Also if several events touch the same channel, it can lead to duplicate work, so Set should be used when computing channels to check. This duplicate work is actually not harmless, it can lead to duplicate changes applied to main state. So it's great that you've spotted this out. Please see https://github.com/rchain/rchain/pull/3312/commits/5dac4d19ffbc84e85608dee780419b55009134d8

nzpr: `Is it because every datum is unique?` Answering this - I believe it should be unique, actually, but this is not related to merging logic, so better not assume this.

shared/src/main/scala/coop/rchain/metrics/Metrics.scala
nzpr: @tgrospic https://github.com/rchain/rchain/pull/3312/commits/bc95a9cc30333f465939f660c4f16de8415a1492

tgrospic (approved):
Great work involving big effort. Thanks!

PR #3337, opened 2021-02-26
Allow arbitrary java for deb and rpm package

### LINKED ISSUE #3336: Fix java dependency for packages
## Overview
Testing revealed that when using arbitrary java version `.deb` (e.g https://adoptopenjdk.net/) package cannot be installed, as dependency is too restrictive. We need to allow any java version greater than version 11.
RPM were not tested, but it looks like its the same.

### Impact
scala.sbt is the scope of issue

### Dependencies
--

### Design
--

1. Do research on best practices
2. Adjust build.sbt

I believe not needed, but we need to test different distributives with different java versions.

### Resource estimation
1 dev

nzpr:
This PR requires testing, and it's not totally clear whats the best way to allow any java version. Need more investigation. Please see issue attached

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,7 +136,7 @@ lazy val shared = (project in file("shared"))
       catsEffectLawsTest,
       catsLawsTest,
       catsLawsTestkitTest,
-      enumeratum,
+      enumeratum
     )
   )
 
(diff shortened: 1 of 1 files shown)


tgrospic:
@nzpr should this be a _draft_ PR if it's not ready for review?

nzpr:
Yes, indeed, thanks!

PR #3342, opened 2021-03-02
Migrate file-based blockstore to key-value implementation

zsluedem:
## Overview

1. implement a `iterateStream` in blockstore
2. implement migration binary 

Steps to migrate the old database:
```
$ mv $RNODE_DATA_DIR/blockstore $RNODE_DATA_DIR/blockstore-old
$ docker run -v $RNODE_DATA_DIR:/var/lib/rnode --entrypoint /opt/docker/bin/block-store-migrate coop.rchain/rnode:$VERSION --original-block-store /var/lib/rnode/blockstore-old --target-data-dir /var/lib/rnode
```

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
@@ -2,7 +2,6 @@ package coop.rchain.blockstorage
 
 import java.nio.ByteBuffer
 import java.nio.file._
-
 import cats.effect.concurrent.Semaphore
 import cats.effect.{Concurrent, Sync}
 import cats.implicits._
@@ -24,11 +23,14 @@ import coop.rchain.lmdb.LMDBStore
 import coop.rchain.metrics.Metrics.Source
 import coop.rchain.metrics.{Metrics, MetricsSemaphore}
 import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.rspace.Context
 import coop.rchain.shared.ByteStringOps._
 import coop.rchain.shared.{AtomicMonadState, Log}
 import monix.execution.atomic.AtomicAny
 import org.lmdbjava.DbiFlags.MDB_CREATE
 import org.lmdbjava.ByteBufferProxy.PROXY_SAFE
+import fs2._
+import monix.eval.Task
 import org.lmdbjava._
 
 import scala.util.matching.Regex
@@ -137,6 +139,26 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
                    readBlockMessage(indexEntry)
                      .map(block => List(blockHash -> BlockMessage.from(block).right.get)) // TODO FIX-ME
                })
+    } yield result
+
+  def iterateStream: F[Stream[F, BlockMessage]] =
+    for {
+      indexes <- index.iterate { iterator =>
+                  Stream.chunk(
+                    Chunk.iterable(
+                      iterator
+                        .map(kv => (ByteString.copyFrom(kv.key()), kv.`val`()))
+                        .map {
+                          case (key, value) => (key, IndexEntry.load(value))
+                        }
+                        .toIterable
+                    )
+                  )
+                }
+      result = indexes.evalMap {
+        case (_, indexEntry) =>
+          readBlockMessage(indexEntry).map(b => BlockMessage.from(b).right.get)
+      }
     } yield result
 
   // Default implementation will use `get` to load the whole block so
@@ -286,6 +308,16 @@ object FileLMDBIndexBlockStore {
       blockStoreDataDir.resolve("approved-block"),
       blockStoreDataDir.resolve("checkpoints")
     )
+
+  def create[F[_]: Concurrent: Sync: Log: Metrics](
+      blockStoreDataDir: Path
+  ): F[BlockStore[F]] = {
+    val blockstoreEnv = Context.env(blockStoreDataDir, 1024L * 1024L * 1024L)
+    for {
+      blockStore <- create(blockstoreEnv, blockStoreDataDir)
+                     .map(_.right.get)
+    } yield blockStore
+  }
 
   def create[F[_]: Monad: Concurrent: Sync: Log: Metrics](
       env: Env[ByteBuffer],
--- a/build.sbt
+++ b/build.sbt
@@ -278,6 +278,7 @@ lazy val node = (project in file("node"))
     ),
     buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion, git.gitHeadCommit),
     buildInfoPackage := "coop.rchain.node",
+    mainClass in Compile := Some("coop.rchain.node.Main"),
     mainClass in assembly := Some("coop.rchain.node.Main"),
     assemblyMergeStrategy in assembly := {
       case x if x.endsWith("io.netty.versions.properties") => MergeStrategy.first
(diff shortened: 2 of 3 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
```diff
@@ -31,6 +32,8 @@ trait BlockStore[F[_]] {
 
   def close(): F[Unit]
 
+  def iterateStream(): F[Stream[F, BlockMessage]]
```
tgrospic: What we need for migration is a function like this which will be used in NodeRuntime
```scala
def migrateBlockStore(FileBS, KVBS) = ???
```
This means it will operate on implementations so there is no need to touch the BlockStore interface.

zsluedem: I don't know how to migrate without offering this? Your function is too wide.
Sorry. I didn't include the function here actually.

node/src/main/scala/coop/rchain/node/BlockStoreMigrate.scala
```diff
@@ -0,0 +1,66 @@
...
+        val blockstoreEnv = Context.env(originalBlockStore, 1024L * 1024L * 1024L)
+        for {
+          blockStore <- FileLMDBIndexBlockStore
+                         .create[Task](blockstoreEnv, originalBlockStore)
+                         .map(_.right.get)
+        } yield blockStore
+      }
+      _ <- log.info("Migrating file based block store to key value store.")
+
+      dataStream    <- blockFileStorage.iterateStream()
```
tgrospic: Here `iterateStream` is called on FileBS instance so there is no need to have it on a BS trait.

node/src/main/scala/coop/rchain/node/BlockStoreMigrate.scala
```diff
@@ -0,0 +1,66 @@
...
+    val originalBlockStore = options.originalBlockStore()
+    val targetDataDir      = options.targetDataDir()
+    implicit val log       = effects.log
+    implicit val metrics   = new Metrics.MetricsNOP[Task]
+
+    val t = for {
+      casperStoreManager <- RNodeKeyValueStoreManager[Task](targetDataDir)
+      blockKVStore       <- KeyValueBlockStore[Task](casperStoreManager)
+
+      blockFileStorage <- {
```
tgrospic: The logic you have here can be inside function on companion object on FileBS. And this function can be called here with instance of FileBS and KVBS.
When we remove FileBS this migration function will not be needed anymore.

zsluedem: Then this file will be deleted also.

node/src/main/scala/coop/rchain/node/BlockStoreMigrate.scala
```diff
@@ -0,0 +1,66 @@
...
+        for {
+          blockStore <- FileLMDBIndexBlockStore
+                         .create[Task](blockstoreEnv, originalBlockStore)
+                         .map(_.right.get)
+        } yield blockStore
+      }
+      _ <- log.info("Migrating file based block store to key value store.")
+
+      dataStream    <- blockFileStorage.asInstanceOf[FileLMDBIndexBlockStore[Task]].iterateStream
+      operation     = dataStream.broadcastTo(4)(s => s.evalMap(b => blockKVStore.put(b)))
```
tgrospic: To _process_ the stream there is also parallel version of evalMap `parEvalMap`. Because we don0t care about ordering it can be faster unordered version. To use as many threads as processors we have our extension `parEvalMapUnorderedProcBounded` in shared project.
```scala
dataStream.parEvalMapUnorderedProcBounded(blockKVStore.put)
```

node/src/main/scala/coop/rchain/node/BlockStoreMigrate.scala
tgrospic: Makes sense. I was already start thinking about adding additional commands to help managing of export/import data from the databases.
For this migration it would be something like this.
`rnode dev migrateFileBlockStore [--source <fileBS-path>] [--target <kvBS-path>]`.

It's outside of the scope for this PR.

node/src/main/scala/coop/rchain/node/BlockStoreMigrate.scala
```diff
@@ -0,0 +1,58 @@
...
+  val targetDataDir = opt[Path](
+    descr = "Output block store directory",
+    required = true
+  )
+
+  verify()
+
+}
+object BlockStoreMigrate {
+  def main(args: Array[String]): Unit = {
```
tgrospic, line 34: How we run multiple _main_ methods?

zsluedem: A separate binary would be generated. We can run through docker.

tgrospic: How this additional binary is generated? Maybe it's easier to just extend existing commands as I suggested in the previous comment https://github.com/rchain/rchain/pull/3342#pullrequestreview-602022629?

zsluedem: it is with `rnode`. Nothing special just another binary like `rnode` which is named as `block-store-migrate`

zsluedem:
bors try

tgrospic (approved):
Looks very nice!

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
```diff
@@ -139,6 +141,26 @@ class FileLMDBIndexBlockStore[F[_]: Monad: Sync: RaiseIOError: Log] private (
                })
     } yield result
 
+  def iterateStream: F[Stream[F, BlockMessage]] =
+    for {
+      indexes <- index.iterate { iterator =>
+                  Stream.chunk(
```
nzpr, line 147: Could you explain for education purpose - why this `Stream.chunk` and then `Chunk.iterable` wrappers used?

build.sbt
```diff
@@ -278,6 +278,7 @@ lazy val node = (project in file("node"))
     ),
     buildInfoKeys := Seq[BuildInfoKey](name, version, scalaVersion, sbtVersion, git.gitHeadCommit),
     buildInfoPackage := "coop.rchain.node",
+    mainClass in Compile := Some("coop.rchain.node.Main"),
```
nzpr, line 281: why this is required? I see that new binary is created, but how exactly is a bit blurry for me. Also I'd expect to see `coop.rchain.node.BlockStoreMigrate` somewhere, does sbt build binary for migration without explicit pointer?

nzpr (approved):
LGTM, but few comments to understand more

build.sbt
zsluedem: sbt can detect `Main` function automatically.  If there are multiple main in `coop.rchain.node` , sbt doesn't know which one belongs to the name `rnode`,  `mainClass in Compile := Some("coop.rchain.node.Main")` is for explicitly specifying the `Main` function for `rnode`.

If we don't put `mainClass in Compile := Some("coop.rchain.node.Main")`, there would be two binary named `block-store-migrate` and `main`

copied pasted from https://discord.com/channels/@me/702142029953761371/816649478705512448

block-storage/src/main/scala/coop/rchain/blockstorage/FileLMDBIndexBlockStore.scala
zsluedem: https://fs2.io/#/guide?id=chunks

It is just a wrapper to convert `Iterable` to `Stream`

zsluedem:
bors merge

zsluedem:
bors try

zsluedem:
I did the migration twice. 

The first time takes **42mins 16s**
The second time takes **34mins 30s**


```
root@dev-support-de:/rchain/mainnet/observer-full-dev/rnode# du -d 1 -h
52K     ./eval
55G     ./blockstorage-2
20K     ./deploystorage
141G    ./blockstore-old
55G     ./blockstorage
5.8M    ./casperbuffer
407G    ./rspace
5.5G    ./dagstorage
663G    .
```

The original file base is **141GB**. The key value lmdb store is **55GB**

PR #3345, opened 2021-03-08
[TEST] Disabled override protobuf hashCode, equals, serializeSize

tgrospic:


zsluedem:
Maybe this is not the perfect place to discuss here.
I think we got 3 parts on protobuf which is not used in the same manner.

1. Comm communication(Network)
2. Block storage
3. RSpace(Par, New, Send, Receive and etc.)

In the **comm** layer, protobuf object call `toByteString` to serialized to bytes.
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/models/src/main/scala/coop/rchain/casper/protocol/PacketTypeTag.scala#L125

In the **blockstorage** layer, protobuf object call `toByteArray` to serialized to bytes.
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala#L132

Actually, for the same protobuf object `toByteString` and `toByteArray` should get the same bytes result behind.

While the biggest difference is the protobuf in **rspace** serilization.

The `Par` are serilized use custom `Serialize` triat
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/models/src/main/scala/coop/rchain/models/serialization/implicits.scala#L15-L23
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/models/src/main/scala/coop/rchain/models/ProtoM.scala#L17-L24

The `mergeFromM` and `serializedSizeM` in compilation are introduced because of this `protobuf Serializer`, I think.
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/project/StacksafeScalapbGenerator.scala#L94-L115
https://github.com/rchain/rchain/blob/b159823c503d945917ead913b12e07c3265bff78/project/StacksafeScalapbGenerator.scala#L145-L279

tgrospic:
bors try

tgrospic:
`toByteArray` and `topByteString` should return the same result as defined in scalapb library.
https://github.com/scalapb/ScalaPB/blob/a9dba47a/scalapb-runtime/src/main/scala/scalapb/GeneratedMessageCompanion.scala#L111-L126

Custom scalapb plugin for Rholang types (StacksafeScalapbGenerator), used in `models` project, should just wrap existing protobuf methods and return the same result also but serialization of deeply nested objects should be safe of stack overflow error.

tgrospic:
bors try

tgrospic:
Test shown that memory leak is not in custom implementation of equality for protobuf types.

Issue #3350, opened 2021-03-10
Modify the DAG representation to support block merge


tgrospic:
## Overview
Currently view on the DAG can only be created using the most recent state of BlockDagStore. To be able to validate merged blocks created by other validators (with relaxed synchrony constraint) we need to be able to construct DAG view from perspective of that block, i.e. view on the DAG of peer validator at the beginning of block creation. When allowing proposing without requirement to wait for all other peer validators to issue new block, this might lead to difference between view on the DAG by block creator and of block validator.

### Design
Adjust `representation` method of BlockDagStorage to be able to ignore messages (and adjust corresponding indices) that are not in scope of a particular block.

nzpr:
New version of finalizer https://github.com/rchain/rchain/pull/3388 makes sure that finalized block is defined only by latest messages. So merging scope is defined by latest messages. This covers part of the issue. 
Remaining part is to not have last finalized blocks persisted and treated as a mutable value but to compute it each time casper shapshot is created (for validation or block creation purpose), or maintain some map Set[LatestMessage] -> Last Finalized Block. Having caching it should not be computationally intensive.

tgrospic:
> Remaining part is to not have last finalized blocks persisted and treated as a mutable value but to compute it each time casper shapshot is created (for validation or block creation purpose), or maintain some map Set[LatestMessage] -> Last Finalized Block. Having caching it should not be computationally intensive.

We can extend BlockMetaData with finalized flag and have this info for every block. Also I think we need info if block is directly or indirectly finalized for easier view on merging scopes and rejected deploys.

Issue #3347, opened 2021-03-09
Conflict detection time grows exponentially


nzpr:
## Overview
As shown in [this PR](https://github.com/rchain/rchain/pull/3312) description, conflict detection time grows exponentially. This makes sense because it is combination of branches involved

### Impact
https://github.com/rchain/rchain/blob/5dac4d19ffbc84e85608dee780419b55009134d8/casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala#L114-L123

### Dependencies
_

### Design
This should be rewritten using map-reduce. Also might be some caching strategies can be implemented.

### Steps to implement the solution
_

### Steps for integration testing
_

### Resource estimation
_

dckc:
I suggest refining titles of issues like this to be clear about *how much* speed should be improved. "conflict detection time grows exponentially" would make a more clear title, with **bug** rather than **enhancement** as the label. Perhaps add a **performance** label too.

nzpr:
@dckc thanks! I'll do that. Thought I'm not sure why this is a bug? This just means that one possible implementation which is not optimal is in place now.

dckc:
Well, it's now phrased as "the code has this bad thing" i.e. "there is a bug".

If exponential time for conflict detection is actually acceptable, then I suppose we should re-phrase it as an enhancement: "reduce conflict detection time from exponential to polynomial time" or some such. But the way I discovered this is via the milestone currently scheduled for March 31, which strongly suggests exponential time is _not_ acceptable. If it _is_ acceptable, I don't think it belongs on the milestone. It can just be dealt with whenever.

nzpr:
I assigned this to block merge milestone because we want to have a version that can clearly show linear scalability asap. And this particular issue does not look that hard. I think I assign features to milestone using "would-be-really-nice-to-have" criteria.

tgrospic:
We have working implementation so this is definitely an improvement that we will try to include in the first version of block merge.

tgrospic:
@nzpr Can this issue be closed? We now have completely different code then referenced in this issue. Am I missing something?

Issue #3339, opened 2021-03-01
Can peek be explicit in consume data structure instead in comm?


zsluedem:
## Overview

Currently, the consume data structure is like below
https://github.com/rchain/rchain/blob/79060b3fa15b372e5ed233e845a01a7528228597/rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala#L117-L121

and the proto structure is like below
https://github.com/rchain/rchain/blob/79060b3fa15b372e5ed233e845a01a7528228597/models/src/main/protobuf/CasperMessage.proto#L211-L215

But when one peek happens but no match it would remain as the same data structure like consume. We can know if that consume is a peek or a real consume when a match happens which leads to a comm event that contains the peek information.

https://github.com/rchain/rchain/blob/79060b3fa15b372e5ed233e845a01a7528228597/models/src/main/protobuf/CasperMessage.proto#L217-L221

https://github.com/rchain/rchain/blob/79060b3fa15b372e5ed233e845a01a7528228597/rspace/src/main/scala/coop/rchain/rspace/trace/Event.scala#L43-L48

## Question
Can peek be part of field in consume instead of in comm? Even more can peek be an independent type?

zsluedem:
Another extended topic on this is that maybe `peek` is worth a method in https://github.com/rchain/rchain/blob/98de8105e7f1a92fdb7101eb368f8d10b299db88/rspace/src/main/scala/coop/rchain/rspace/Tuplespace.scala#L5-L70 trait

PR #3351, opened 2021-03-11
Merge block merge feature branch into main dev

nzpr:
## Overview
This PR merges feature/block-merge branch (tagged as v0.11.0-alpha) into dev branch.


nzpr:
This merge does not have any conflicts, so no special branch for merge.

casper/src/main/scala/coop/rchain/casper/LastFinalizedHeightConstraintChecker.scala
```diff
@@ -1,52 +1,54 @@
...
-                   ) >> (heightDifference <= heightConstraintThreshold).pure[F]
+                   ) >> result.pure[F]
                  case None =>
-                   Sync[F].raiseError[Boolean](
-                     new IllegalStateException("Validator does not have a latest message")
-                   )
+                   CheckProposeConstraintsResult.success.pure[F]
+//                   Sync[F].raiseError[CheckProposeConstraintsResult](
+//                     new IllegalStateException("Validator does not have a latest message")
+//                   )
```
tgrospic: Without changing how validator is bonded are we already considering valid block without validator latest block?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -570,8 +396,8 @@ object MultiParentCasperImpl {
     )
   }
 
-  def createdEvent(cbs: Created): RChainEvent = {
-    val (blockHash, parents, justifications, deployIds, creator, seqNum) = blockEvent(cbs.block)
+  def createdEvent(b: BlockMessage): RChainEvent = {
```
tgrospic, line 404: `Created` block message was more specific block and with this change any block can be supplied. What was the reason for this change?

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
```diff
@@ -84,25 +83,25 @@ final class SynchronyConstraintChecker[F[_]: Sync: BlockStore: Estimator: Log: S
...
-          allowedToPropose     <- if (latestBlockIsGenesis) true.pure[F] else checkConstraint
+          allowedToPropose <- if (latestBlockIsGenesis)
+                               CheckProposeConstraintsResult.success.pure[F]
+                             else checkConstraint
         } yield allowedToPropose
       case None =>
-        Sync[F].raiseError[Boolean](
-          new IllegalStateException("Validator does not have a latest message")
-        )
+        CheckProposeConstraintsResult.success.pure[F]
```
tgrospic: Here is the same question about validator's latest message.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -699,47 +685,47 @@ object BlockAPI {
...
+//      }
+//      .merge
+
+//  private def prettyPrintUnmatchedSends[F[_]: Concurrent](
+//      casper: MultiParentCasper[F],
+//      deploys: Seq[Signed[DeployData]]
+//  ): F[String] =
+//    casper.getRuntimeManager >>= (
+//      _.withRuntimeLock(runtime => StoragePrinter.prettyPrintUnmatchedSends(deploys, runtime))
+//    )
```
tgrospic: Is this still needed?

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
```diff
@@ -0,0 +1,250 @@
...
+//                         for {
+//                           results <- fs2.Stream.emits(workers).parJoinUnbounded.compile.toList
+//                         } yield (
+//                           (
+//                             results.map(_._1) ++ leftover,
+//                             acc ++ results.flatMap(_._2)
+//                           )
+//                         ).asLeft[Set[ProcessedDeploy]]
+//                       }
+//                     }
```
tgrospic, line 250: Is this still needed?

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperMergingDagReader.scala
```diff
@@ -0,0 +1,54 @@
...
+import cats.effect.Concurrent
+import cats.syntax.all.{none, _}
+import coop.rchain.blockstorage.BlockStore
+import coop.rchain.blockstorage.syntax._
+import coop.rchain.dag.DagReader
+import coop.rchain.models.BlockHash.BlockHash
+
+final class CasperMergingDagReader[F[_]: Concurrent: BlockStore](
+    hashDAGReader: DagReader[F, BlockHash]
+) extends DagReader[F, MergingVertex] {
```
tgrospic, line 12: What it means to extend `DagReader` and also accept it as a parameter?

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -59,7 +62,8 @@ object RNodeKeyValueStoreManager {
       (Db("eval-roots"), evalHistoryEnvConfig),
       (Db("eval-cold"), evalColdEnvConfig),
       // RSpace cache (used for block merge)
-      (Db("rspace-channels"), rspaceCacheEnvConfig)
+      (Db("rspace-channels"), rspaceCacheEnvConfig),
+      (Db("channels"), channelEnvConfig)
```
tgrospic: These names are not very descriptive.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -32,9 +40,9 @@ object InterpreterUtil {
 
   private[this] val ComputeParentPostStateMetricsSource =
     Metrics.Source(CasperMetricsSource, "compute-parents-post-state")
-
-  private[this] val ReplayIntoMergeBlockMetricsSource =
-    Metrics.Source(CasperMetricsSource, "replay-into-merge-block")
+//
+//  private[this] val ReplayIntoMergeBlockMetricsSource =
+//    Metrics.Source(CasperMetricsSource, "replay-into-merge-block")
```
tgrospic: Is this needed?

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -197,98 +210,61 @@ object InterpreterUtil {
...
-        case Seq() =>
-          runtimeManager.emptyStateHash.pure
+        case 0 =>
+          RuntimeManager.emptyStateHashFixed.pure[F]
 
-        case Seq((_, parentStateHash)) =>
-          parentStateHash.pure
+        // For single parent, get itd post state hash
+        case 1 =>
+          ProtoUtil.postStateHash(parents.head).pure
```
tgrospic: This way of matching on a list size is more prone to errors when code is refactored and `head` is called on empty list.
Preferred way is to match on the list itself so that head of the list is ensured to exist by definition of the case expression.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -197,98 +210,61 @@ object InterpreterUtil {
...
-                blockHashesToApply
-              )
-          }
+        // we might want to take some data from the parent with the most stake,
+        // e.g. bonds map, slashing deploys, bonding deploys.
+        // such system deploys are not mergeable, so take them from one of the parents.
+        case _ => {
+          implicit val r = runtimeManager
+          val tipsHashes = parents.map(_.blockHash)
+          val lfbHash    = s.lastFinalizedBlock
```
zsluedem, line 253: I just want to say that I still think it is wrong to introduce last finalized block here for state merging.  I think you should check if the base state is right when you checkout 0.67 synchrony constraint.

Maybe it would be better to think as @tgrospic mentioned the concept -> **deploy chain** before and stop thinking of base state.

models/src/main/protobuf/ProposeServiceV1.proto
```diff
@@ -20,6 +20,7 @@ option (scalapb.options) = {
 
 service ProposeService {
   rpc propose(PrintUnmatchedSendsQuery) returns (ProposeResponse) {}
+  rpc proposeResult(ProposeResultQuery) returns (ProposeResultResponse) {}
```
zsluedem, line 23: Currently we might need to change the propose script for this.

To be very honest, I don't want this feature in this version right now. Propose script is very stable right now after several modifications.

we can discuss this after this pr.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
nzpr: @zsluedem There is no common base state in this implementation. LFB here is used to determine the scope of merging.
Deploy chains should be calculated somehow - you might think about current code as one of possible implementations of how to calculate these event chains.
1. You have DAG with tips
2. You have a scope defined, which is subject to change (deploys within scope can be rejected). This scope is - from tips, down to finalised block.
3. Then you need to split this scope in set of deploy chains - this is what is done in https://github.com/rchain/rchain/blob/5dac4d19ffbc84e85608dee780419b55009134d8/casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala#L72-L73
All tips should be descendants of finalized block as per block creator (lets call it LFBc) (** *actually this statement might need more thinking - here I agree, with 0.67 it might require modification. I'm not sure if it is guaranteed that all tips have LFB as an ancestor, thanks for heads up*), so DAG is traversed and subdag connecting all tips to LFBc defines the scope of merge.
This particular way of splitting aims having one of tips as the mains state. This should minimise number of events to merge, because many of events in the scope are already inside tip state.
4. Then you have set of deploy chains that you can merge. Each deploy chain is defined by `end state` (post state of tip of the DAG), `start state` (pre state of the very bottom block which is starting point of event chain **Note: this is NOT LFB**) and all events that happened to transition start to end.
https://github.com/rchain/rchain/blob/cf7943ef955474b0f5b53c3530ee8a5625eadffc/rspace/src/main/scala/coop/rchain/rspace/merger/StateMerger.scala#L24-L34

--*PTAL at remark under ** in item 3. I agree that splitting the scope into event chains might require adjusting. But not sure.*

nzpr: This actually returns us again to older question - how to define the scope of merging. As synchrony constraint is not connected to safety oracle, it's not that straightforward. Node A might have last finalised block that is not connected to all tips when it is about to create block.

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -1,247 +1,247 @@
...
+//    }
+//
+//    override def nextState(state: State): State = state
+//
+//    override def preCondition(state: State): Boolean = true
+//
+//    override def postCondition(state: State, result: Try[Result]): Prop =
+//      result.isSuccess && result.get.isEmpty
+//  }
+//}
```
tgrospic: What is the status of this thing?

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
```diff
@@ -39,39 +38,39 @@ class MultiParentCasperAddBlockSpec extends FlatSpec with Matchers with Inspecto
...
+//                     }
+//        } yield result
+//      }
+//    val threadStatuses: (ValidBlockProcessing, ValidBlockProcessing) =
+//      testProgram.unsafeRunSync(scheduler)
+//
+//    threadStatuses should matchPattern {
+//      case (Left(CasperIsBusy), Right(Valid)) | (Right(Valid), Left(CasperIsBusy)) =>
+//    }
+//  }
```
tgrospic, line 73: Is the refactoring of this part mentioned in some issue?

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -62,208 +57,209 @@ class CreateBlockAPITest extends FlatSpec with Matchers with EitherValues {
...
+  //            b3Status <- createBlock(blockApiLock)(engineCell)
+  //
+  //            _ = b3Status.left.value should include(
+  //              "Must wait for more blocks from other validators"
+  //            )
+  //          } yield ()
+  //      }
+  //  }
+  //
+  //  it should "allow proposals with enough new blocks from other validators" in effectTest {
```
tgrospic: Do we have similar test like this with the Proposer?

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -62,208 +57,209 @@ class CreateBlockAPITest extends FlatSpec with Matchers with EitherValues {
...
+  //
+  //            _        <- n1.casperEff.addDeploy(deploys(3))
+  //            b4Status <- createBlock(blockApiLock)(engineCell)
+  //
+  //            _ = b4Status shouldBe a[Right[_, String]]
+  //          } yield ()
+  //      }
+  //  }
+  //
+  //  it should "check for new deploys before checking synchrony constraint" ignore effectTest {
```
tgrospic: With the Proposer we still need to check synchrony constraint, is this done in some other test?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -88,88 +90,88 @@ class MultiParentCasperCommunicationSpec extends FlatSpec with Matchers with Ins
...
+//          _ <- nodes(0).receive()
+//
+//          _ = nodes(1).logEff.infos.count(_ startsWith "Requested missing block") should be(10)
+//          result = nodes(0).logEff.infos.count(
+//            s => (s startsWith "Received request for block") && (s endsWith "Response sent.")
+//          ) should be(10)
+//          // Oh yes - 10!
+//        } yield result
+//      }
+//  }
```
tgrospic: We need note/issue to track removal/refactor of these tests (maybe this is already covered in some issue).

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
```diff
@@ -59,45 +46,38 @@ class MultiParentCasperMergeSpec
...
+//        """.stripMargin)
+//    val lookup =
+//      Rho(s"""
+//             |new rl(`rho:registry:lookup`) in {
+//             |  rl!(`$uri`, Nil)
+//             |}
+//        """.stripMargin)
+//
+//    merges(lookup, lookup, setup)
+//  }
```
tgrospic: Add reference to an issue where is tracked.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
```diff
@@ -229,61 +209,61 @@ class MultiParentCasperMergeSpec
...
+//      (MergeabilityRules.baseMergeabilityCases ++ MergeabilityRules.peekMergeabilityCases)
+//        .map(_._1)
+//    withClue(s"""Missing cases: ${allMergeabilityCases
+//      .diff(testedMergeabilityCases.toSet)
+//      .toList
+//      .sorted
+//      .mkString(", ")}\n""") {
+//      testedMergeabilityCases should contain allElementsOf allMergeabilityCases
+//    }
+//  }
```
tgrospic: Add reference to an issue where is tracked.

casper/src/test/scala/coop/rchain/casper/batch2/SingleParentCasperSpec.scala
```diff
@@ -40,30 +40,30 @@ class SingleParentCasperSpec extends FlatSpec with Matchers with Inspectors {
...
+//          dualParentB3 = b3.copy(header = b3.header.copy(parentsHashList = twoParents))
+//
+//          validateResult <- {
+//            import n1._
+//            n1.blockDagStorage.getRepresentation >>=
+//              (dag => Validate.parents(dualParentB3, n1.genesis, dag))
+//          }
+//        } yield validateResult shouldBe BlockStatus.invalidParents.asLeft[ValidBlock]
+//    }
+//  }
```
tgrospic: What is the reason this test is commented out? It seems it should work the same for single parent casper.

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
```diff
@@ -37,98 +37,98 @@ import scala.collection.concurrent.TrieMap
...
+//            hashHandlerLock
+//          )
+//          .unsafeRunSync
+//        sentToCasper.contains(b.blockHash) should be(false)
+//      }
+//
+//      // TODO this is after LFS comlete
+//      ignore("should drop old ignorable blocks") {}
+//    }
+//  }
```
tgrospic: Add reference to an issue where is tracked.

node/src/main/resources/defaults.conf
```diff
@@ -331,3 +331,10 @@ metrics {
   zipkin = false
   sigar = false
 }
+
+dev-mode = false
+
+dev {
+  # If set, on each propose node will add dummy deploy signed by this key.
+  # deployer-private-key =
+}
```
tgrospic: Add new line at the end.

rspace/src/main/scala/coop/rchain/rspace/trace/TuplespaceEvent.scala
```diff
@@ -39,6 +38,7 @@ object TuplespaceEvent {
 
   def from(consume: Consume): Option[(Blake2b256Hash, TuplespaceEvent)] = consume match {
     case Consume(singleChannelHash :: Nil, _, _) =>
+      // ???????????????? it is possible that the consume is peek
```
tgrospic, line 41: Do we have issue to track this?

tgrospic (commented):
I added comments on things that needs more work or attention.
To be able to track them, notes on the board or issues (if more description is needed) should be created and put reference in the comments.

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperMergingDagReader.scala
nzpr: Might be this should be done another way, but this is transformer from reader operating on one type of vertex into another type of vertex. DAG is the same, but verices are different
If you have an instance of HashDagReader - you can create MergingDagReader, where each vertex support all capabilities required for merging.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Previously block creation was using this status
https://github.com/rchain/rchain/blob/fb789e156f4b10e6960e1b6b34c852ddc3a0c37b/casper/src/main/scala/coop/rchain/casper/CreateBlockStatus.scala#L29-L37

but with propose implementation it is not used, so its ok to have just BlockMessage without wrapper

tgrospic:
> This merge does not have any conflicts, so no special branch for merge.

Creating a branch on your repo is useful for others to easily clone the same code locally and inspect it, so it's not only for situation where there are conflicts.

casper/src/main/scala/coop/rchain/casper/LastFinalizedHeightConstraintChecker.scala
nzpr: I think this error message is misleading and does not tell much, as it is not related to this particular constraint checker.   If validator is bonded, but had not issued its first message - it should be able to create one, and peers should be able to validate the very first message. **[1]**

If it is not bonded - propose status should return exactly this **[2]**. And here is where all propose constraints check happen. Active validator check should fail.
https://github.com/rchain/rchain/blob/bc95a9cc30333f465939f660c4f16de8415a1492/casper/src/main/scala/coop/rchain/casper/blocks/proposer/Proposer.scala#L82-L97 

**[1]** *We had a related discussion - when to create latest message, should that be message where bonding happened or the very first block proposed. I believe it should be in the second one, but in both cases, having this error raised does not make sense imo. If no latest message seen for validator - network should not restrict it to create its first message.*

**[2]** Only propose is mentioned here because checking of propose constraints (except `creator should be bonded`) during validation is not implemented - its some future task,  which is part of shard config feature. Boilerplate code is mostly ready, but <https://github.com/rchain/rchain/issues/3350> feature is needed as well. But when validating, the reasoning remains the same imo.

casper/src/main/scala/coop/rchain/casper/SynchronyConstraintChecker.scala
nzpr: https://github.com/rchain/rchain/pull/3351#discussion_r593666487

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
nzpr: I'm not sure where seeing unmatched sends were useful, so left it commented out.

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
nzpr: @tgrospic this is attempt to implement conflict detection using map-reduce. So yes, its a WIP code

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/batch2/SingleParentCasperSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618, its because of API change

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618 but this particular test is a duplicate for mergeablilySpec I believe

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
nzpr: No, here's issue https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-798109618

rspace/src/main/scala/coop/rchain/rspace/trace/TuplespaceEvent.scala
zsluedem: https://github.com/rchain/rchain/issues/3339
related

node/src/main/resources/defaults.conf
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-800021338

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-800021338

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-800021338

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
nzpr: https://github.com/rchain/rchain/issues/3277#issuecomment-800021338

nzpr:
@tgrospic I've put all remaining tasks into comments here https://github.com/rchain/rchain/issues/3277#issuecomment-798109618 and https://github.com/rchain/rchain/issues/3277#issuecomment-800021338

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperMergingDagReader.scala
```diff
@@ -0,0 +1,54 @@
...
+  override def children(vertex: MergingVertex): F[Option[Set[MergingVertex]]] =
+    hashDAGReader.children(vertex.blockHash).flatMap {
```
tgrospic, line 14: Both of these member functions only use `vertex` to get the `blockHash`. So this can be simply viewed as function composition.
We are starting from these functions (children, parents) with the type roughly like this `cpHash : Hash -> Opt[Hashes]`. They are functions on DagReader which is omitted as the first parameter.

What we need is `cpVertex : Vertex -> Opt[Vertexes]` which is like lifted function from Hash to Vertex.

Let's see what we already have. `vxHash : Vertex -> Hash` and `cpHV : Hashes -> Opt[Vertexes]` (now inside match).

We can compose it `cpVertex = vxHash andThen cpHash andThen cpHv`. For careful eye, of course there is one problem, `cpHash` returns `Opt[Hashes]` so it must be `flatMap`-ed. :exclamation: **No matching needed. Matching is already done in implementation of Monad.**

In this strange notation the types goes like this.
`cpVertex : Vertex -> Hash -> Opt[Hashes].flatMap( Hashes -> Opt[Vertexes] )`

NOTE: because we are in Scala class may be useful to group implicit dependencies and create extension methods. But this will be much easier in Scala 3. :partying_face:

casper/src/main/scala/coop/rchain/casper/blocks/merger/CasperDagMerger.scala
tgrospic: Should this be mentioned as part of #3347?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: Please be sure you add comments about unused code. It's very difficult later to decide if is really needed or not.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
tgrospic: Please add this info on a new issue so we can extend discussion.

casper/src/main/scala/coop/rchain/casper/LastFinalizedHeightConstraintChecker.scala
tgrospic: @nzpr from our discussion on standup, for now it makes sense to keep throwing error when validator has no latest message. Please add this change as part of PR #3355.

nzpr: https://github.com/rchain/rchain/pull/3355#issuecomment-813358683

models/src/main/protobuf/ProposeServiceV1.proto
tgrospic: Also empty message is already defined in protobuf.
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/empty.proto

tgrospic (approved):
Great work is done here, it's ready to run as single parent, let's merge it. 🎉

zsluedem (approved):
Great!!

nzpr:
bors merge

PR #3354, opened 2021-03-16
Per block DAG view

### LINKED ISSUE #3350: Modify the DAG representation to support block merge

nzpr:
This PR enables truncating the most recent view on the DAG that is stored in BlockDagStore to the view of some message (block), defined by its justifications. 
In addition, finalised block is included in this view, so finalisation process can be replayed. Which is required for block merge 
in conditions without leader.


block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
```diff
@@ -1,46 +1,48 @@
...
-    lastFinalizedBlockDb: KeyValueTypedStore[F, Int, BlockHash]
+    lastFinalizedBlockDb: KeyValueTypedStore[F, Validator, BlockHash]
 ) extends LastFinalizedStorage[F] {
-  val fixedKey = 1
-  override def put(blockHash: BlockHash): F[Unit] =
-    lastFinalizedBlockDb.put(Seq((fixedKey, blockHash)))
+  override def put(validator: Option[Validator] = None, blockHash: BlockHash): F[Unit] =
+    lastFinalizedBlockDb.put(
+      Seq((validator.getOrElse(LastFinalizedStorage.noValidatorKey), blockHash))
+    )
```
tgrospic: I don't think it's needed to extend existing methods to support new key `LastFinalizedStorage.noValidatorKey`.

Example for `put(validator.some, ...)`.
```scala
def put(validator: Validator, blockHash: BlockHash): F[Unit] =
  lastFinalizedBlockDb.put(Seq((validator, blockHash)))
```
Example for `put(none, ...)` (syntax).
```scala
def put(blockHash: BlockHash): F[Unit] =
  lastFinalizedBlockDb.put(Seq((LastFinalizedStorage.noValidatorKey, blockHash)))
```

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -120,7 +120,92 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                             acc.asRight[(Set[BlockMetadata], Set[BlockMetadata])].pure[F]
+                           case v =>
+                             for {
+                               parents <- replaceWithParents(v)
+                               nextLvl <- parents.toList.filterA(
+                                           p => isInValidatorView(p.blockHash, excessValidators).not
+                                         )
+                             } yield (nextLvl.toSet, acc ++ v).asLeft[Set[BlockMetadata]]
+                         }
+                     }
```
tgrospic: It would be nice to Streamify this and expand in parallel. :wink:

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -100,4 +104,44 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+      mainParent: BlockHash
+  )(implicit bs: BlockStore[F]): F[List[BlockHash]] =
+    dag
+      .children(mainParent)
+      .flatMap(
+        _.traverse(
+          _.toList.filterA(
+            hash =>
+              BlockStore[F]
+                .getUnsafe(hash)
```
tgrospic: We are waiting here in sequence for each block to be read from disk and decoded.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -100,4 +104,44 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+        b => b.header.parentsHashList.contains(targetMainParent)
+      )
+
+  def getNonFinalizedAncestors(blockHash: BlockHash): F[Set[BlockHash]] =
+    DagOps
+      .bfTraverseF[F, BlockHash](List(blockHash))(
+        hash =>
+          dag
+            .parents(hash)
+            .flatMap(_.getOrElse(Set.empty[BlockHash]).toList.filterA(v => dag.isFinalized(v).not))
```
tgrospic, line 115: For check for finalization we also need to wait in sequence.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,123 @@
...
+                hash =>
+                  Stream.eval(
+                    removeBLockDeploysF(hash) >>
+                      recordFinalizedBlockF(hash)
+                  )
+              )
+            )
+            .parJoinProcBounded
+            .compile
+            .drain
```
tgrospic: The whole mapping can be done inside Stream. Try not to allocate new memory for the whole structure you are starting with and do the transformation inside Stream where things are _loaded_ on demand (pull concept of fs2).
```scala
_ <- Stream
      .emits(toFinalize.toList)
      .parEvalMapProcBounded(hash => removeBLockDeploysF(hash) >> recordFinalizedBlockF(hash))
      .compile
      .drain
```

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -120,7 +120,92 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+    */
+  private def representation(
+      latestMessagesToUse: Option[Set[BlockHash]] = None
+  ): F[BlockDagRepresentation[F]] = {
+
+    /** trims view on the DAG to get view as per provided `latest messages` */
+    def calculateView(
+        fullDag: KeyValueDagRepresentation,
+        tips: Set[BlockHash]
+    ): F[KeyValueDagRepresentation] = {
```
tgrospic: `calculateView` depends on `BlockDagRepresentation` and it makes sense to be part of the trait. With that restricted version `getRepresentation(hash)` can be added as extension on `BlockDagKeyValueStorage`.

If you think it will be easier to test if `KeyValueDagRepresentation` is public I'm ok with that. Also if you don't want to modify file based version you can create another PR to remove it.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
nzpr: I'm not sure I understand - I've adjusted this to be able to store finalized blocks for different Validators, to replay finalization process. Option here is to also support local view for readonly nodes.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: @tgrospic can we remove file versions support both for dagstore and blockstore?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: I think this is it?
https://github.com/rchain/rchain/blob/6f66e3c28d12af75eb5ebd22d7b5992f35f0753d/shared/src/main/scala/coop/rchain/shared/DagOps.scala#L9-L24
I guess I made it too complex

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
tgrospic: We still have full nodes with file-based BlockStore so for now only file-based DAG store is safe to remove.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
tgrospic: In other words, why to decide about `none` case inside `put` when this can be additional (syntax) method. In that way current interface does not need to be changed.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedFileStorage.scala
```diff
@@ -61,7 +63,9 @@ object LastFinalizedFileStorage {
     for {
       lock                      <- Semaphore[F](1)
       lastFinalizedBlockHashOpt <- readLastFinalizedBlockHash(lastFinalizedPath)
-      state                     <- Cell.mvarCell[F, Option[BlockHash]](lastFinalizedBlockHashOpt)
+      state <- Cell.mvarCell[F, Map[Option[Validator], Option[BlockHash]]](
+                Map(none[Validator] -> lastFinalizedBlockHashOpt)
+              )
```
tgrospic: Try to replace Cell with Ref because we want to get rid of Cell completely.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
nzpr: but 
```scala
def put(validator: Validator, blockHash: BlockHash): F[Unit] =
  lastFinalizedBlockDb.put(Seq((validator, blockHash)))
```
cannot be syntax - it has to be trait method, because underlying `lastFinalizedBlockDb` requires validator. So in trait methods we have to be to supply validator. So either we need to add second method with validator, of change the only one to support Option for validator. 
Are you saying that the second option is better?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3363

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedFileStorage.scala
nzpr: This is removed along with file DAG store

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
tgrospic: In my first comment are two methods, one on trait is with required validator and on the syntax is using default validator value so don't require validator as a parameter. (*maybe syntax will fail to compile with the exact same name)

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
tgrospic: One correction, conversion to list is not needed. This seems like a best way to convert Scala collection to Stream.

```scala
_ <- Stream
      .fromIterator(toFinalize.iterator)
      .parEvalMapProcBounded(hash => removeBLockDeploysF(hash) >> recordFinalizedBlockF(hash))
      .compile
      .drain
```

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -176,43 +189,72 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+
     for {
-      dag         <- BlockDagStorage[F].getRepresentation
-      r           <- Estimator[F].tips(dag, approvedBlock)
-      (lca, tips) = (r.lca, r.tips)
+      dag <- BlockDagStorage[F].getRepresentation
+
+      // read or calculate parents
+      parents <- if (targetBlockOpt.isDefined) ProtoUtil.getParents(targetBlockOpt.get)
+                else calculateParents(dag)
```
tgrospic, line 260: Try to map _until the last value_ and return otherwise.
```scala
parents <- targetBlockOpt.map(ProtoUtil.getParents).getOrElse(calculateParents(dag))
```

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -176,43 +189,72 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
-          r = lms.toList
-            .map {
-              case (validator, blockMetadata) => Justification(validator, blockMetadata.blockHash)
-            }
-            .filter(j => onChainState.bondsMap.keySet.contains(j.validator))
-        } yield r.toSet
-      }
+      // read or calculate justifications
+      justifications <- if (targetBlockOpt.isDefined) targetBlockOpt.get.justifications.pure[F]
+                       else calculateJustifications(dag, onChainState)
```
tgrospic, line 274: Same here. Mapping is just field accessor `(_.justifications.pure[F])`.

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -113,8 +114,12 @@ class Initializing[F[_]
               )
             )
 
-        // Update last finalized block with received block hash
-        _ <- LastFinalizedStorage[F].put(block.blockHash)
+        // Update last finalized block for all bonded validators with received block hash
+        _ <- block.body.state.bonds
+              .map(_.validator)
+              .traverse(v => LastFinalizedStorage[F].put(v.some, block.blockHash))
```
tgrospic, line 120: We have LFStorage backed by KeyValueStore which can _put_ multiple kv pairs. It makes sense to expose this in finalized storage and save all validators in one DB transaction.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
tgrospic: On second taught, now this whole hierarchy of `LastFinalizedStorage` is not necessary.

It can be defined directly as typed kv store (with multi pair operations).
```scala
type LFStore = KeyValueTypedStore[F, Validator, BlockHash]
// or
final case class LFStrore(kvStore: KeyValueTypedStore[F, Validator, BlockHash])
```
Also `LastFinalizedMemoryStorage` can be removed because we have in-memory `InMemoryKeyValueStore`.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Yes, thats old code, now I do map/traverse over Options

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedKeyValueStorage.scala
nzpr: @tgrospic hdy think, maybe put lastFinalizedBlocks database inside BlockDagStorage? so modify API to have
```scala
  def setFinalizedHash(validator: Validator, blockHash: BlockHash): F[Unit]
  def getFinalizedHash(validator: Validator): F[BlockHash]
```
Actually there is already `addFinalisedBlock`, which lives in parallel universe with finalized storage and nly changes inmem metadata... So I think its reasonable to make dagstore responsible for persisting LFB as well.

tgrospic: It makes sense, but it would be nice that changing of finalization state is done internal to DagStore and (snapshot) read is exposed through DagRepresentation.
Implementation of DagStore can accept finalization KeyValueTypedStore and use setter for multiple kv pairs.

Issue #3346, opened 2021-03-09
Improve concurrent block validation (optimize or remove channels map)


nzpr:
## Overview
Parallel block validation should be improved. Currently disconnection between channels in event log an real channels requires maintaining a map, and its quite costly 
<img width="1783" alt="Screenshot 2021-02-10 at 22 49 11" src="https://user-images.githubusercontent.com/1746367/110436120-e2167500-80c4-11eb-9fdf-341ee9a8c122.png">

### Impact
https://github.com/rchain/rchain/blob/cf7943ef955474b0f5b53c3530ee8a5625eadffc/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala#L238-L240 

Also there might be another obstacles preventing efficient parallel validation, the entry point for investigations is 
https://github.com/rchain/rchain/blob/3afeda6ff0ac8a2711ce7448b2d8dcc2e186d60a/node/src/main/scala/coop/rchain/node/instances/BlockProcessorInstance.scala#L15-L21

### Dependencies
_

### Design
_

### Steps to implement the solution
At the moment main thesis why concurrent validation is slower than expected is because of maintaining channels map. Optimising this code is a way to go, as it won't be removed completely for some time. Other bottlenecks should be investigated as well.

### Steps for integration testing
_

### Resource estimation
_

nzpr:
After this issue is resolved, tests for block processor should be implemented as per https://github.com/rchain/rchain/pull/3355#discussion_r610344614

dckc:
In [227 Apr 23 Community Debrief](https://blog.rchain.coop/2021/04/23/debrief-227/), @9rb reports that a hard fork is needed in addition to block merge to get RChain to perform / scale as designed. I asked for a pointer to the relevant issue and @nzpr suggested this one.

Since this is preventing RChain from scaling as designed, I suggest changing the label from an enhancement to **bug** and changing the title to "channels map interferes with parallel block validation". (A **performance** label would be a nice touch.)

Also, the [Block merge milestone](https://github.com/rchain/rchain/milestone/4) is **Past due by 29 days**; please update it to represent your best guess.

And if addressing this issue requires a hard fork, perhaps it belongs on a later milestone?

tgrospic:
Channels map removed in linked PR.

PR #3355, opened 2021-03-18
Fix tests for block merge branch + misc leftover

### LINKED ISSUE #3277: Collection of issues left over from block-merge sync to dev branch

nzpr:
## Overview
This PR aims restoring all tests disabled in block-merge branch, as well as address minor comments on merge `block-merge` feature branch.


casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -90,88 +91,96 @@ class MultiParentCasperCommunicationSpec extends FlatSpec with Matchers with Ins
...
-//          _ = nodes(1).logEff.infos.count(_ startsWith "Requested missing block") should be(10)
-//          result = nodes(0).logEff.infos.count(
-//            s => (s startsWith "Received request for block") && (s endsWith "Response sent.")
-//          ) should be(10)
-//          // Oh yes - 10!
-//        } yield result
-//      }
-//  }
+  // TODO this test fails because `TestNode.propagate(nodes)` returns prematurely.
+  it should "ask peers for blocks it is missing and add them" ignore effectTest {
```
nzpr: this test is ignored, because it looks like TestNode.propagate(nodes) returns prematurely in presence of async message processing. Im not sure yet whats the way to fix this.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -14,55 +15,55 @@ import org.scalatest.{FlatSpec, Inspectors, Matchers}
...
+        //1 receives request for has block1; sends i have block1
+        //2 receives I have block1; asks for block1
+        //1 receives request block1; sends block1
+        //2 receives block2; asks for block1
+        //2 receives block1; adds both block1 and block2
+
+        // due to asynchronous processing of blocks syncWith ends before blocks are added to DAG
+        // as this test is supposed to check that node is asking for block, just checking blockstore
+        // for presence of block is enough
+        _ <- nodes(2).blockStore.contains(signedBlock1.blockHash) shouldBeF true
```
nzpr: this is fix for this tests, ptal at comment

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -1,247 +1,229 @@
...
+
+    override def postCondition(state: State, result: Try[Result]): Prop = isValidAndNoErrors(result)
+  }
+
+  case class DeployAndPropose(node: RNode) extends Command {
+    override type Result = (ValidBlockProcessing, List[String])
+
+    override def run(sut: Sut): Result = {
+      val validator = sut.nodes(node.idx).node
+      val result = (deploy(validator, deployment(0)) >> create(validator) >>= (
```
nzpr, line 190: this test required sync version of propose

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
```diff
@@ -46,39 +46,6 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
     }
   }
 
-  // TODO enable tests below
```
nzpr, line 49: this does not make sense in presence of new MergeSpec. We can fix it but it'll require changes to some MergeSpec API (`merges(lookup, lookup, setup)`) is not present anymore. I don't think it worth it. We have BranchMergerSpec to merge actual blocks with cost accounting and StateMergeSpec to check all mergeability rules.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
```diff
@@ -208,62 +175,4 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
       } yield ()
     }
   }
-
-//  "This spec" should "cover all mergeability cases" in {
```
nzpr, line 212: this is superseded by MergeSpec

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -62,7 +61,6 @@ object RNodeKeyValueStoreManager {
       (Db("eval-roots"), evalHistoryEnvConfig),
       (Db("eval-cold"), evalColdEnvConfig),
       // RSpace cache (used for block merge)
-      (Db("rspace-channels"), rspaceCacheEnvConfig),
```
nzpr: this is duplicate for `channels` and not used

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
```diff
@@ -1,134 +0,0 @@
-package coop.rchain.casper.engine
```
nzpr, line 1: There should be BlockProcessor test instead of this... I guess I need to write one

casper/src/main/scala/coop/rchain/casper/blocks/proposer/ProposeResult.scala
```diff
@@ -24,7 +24,10 @@ case object NotEnoughNewBlocks         extends CheckProposeConstraintsFailure
 case object TooFarAheadOfLastFinalized extends CheckProposeConstraintsFailure
 
 object CheckProposeConstraintsResult {
-  def success: CheckProposeConstraintsResult = CheckProposeConstraintsSuccess
+  def success: CheckProposeConstraintsResult                    = CheckProposeConstraintsSuccess
+  def notBonded: CheckProposeConstraintsResult                  = NotBonded
+  def notEnoughNewBlock: CheckProposeConstraintsResult          = NotEnoughNewBlocks
+  def tooFarAheadOfLastFinalized: CheckProposeConstraintsResult = TooFarAheadOfLastFinalized
```
zsluedem, line 30: these methods are more like renaming from my perspective. I am curious why we need these methods.

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -1,247 +1,229 @@
...
+  implicit def noShrink[T]: Shrink[T] = Shrink.shrinkAny
+  property("network") = HashSetCasperSpecification.property()
+}
+
+object HashSetCasperPropertiesApp {
+  def main(args: Array[String]): Unit =
+    HashSetCasperSpecification.property().check(_.withMinSuccessfulTests(1))
+}
+
+class NodeBox(val node: TestNode[Effect], var lastBlock: Option[BlockMessage]) {
```
zsluedem, line 39: Just curious that can we use `val` instead of `var`?

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
tgrospic: Do we have this fixed on some other place? Channels map DB should not be shared with evaluator.

Channels map DB in Casper RSpace should be in the `rspace` folder with other DBs and for evaluator we can use dummy DB to not create any mark on disk.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -14,55 +15,55 @@ import org.scalatest.{FlatSpec, Inspectors, Matchers}
 // casper and communication has little to do with each other, so commented out
```
tgrospic: Is this comment still relevant?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
tgrospic: This test doesn't check what it says in the title, it only checks that node 2 will eventually receive block 1.

Can we check that BlockRetriever received a request from Casper  for block 1 (to not go further to TransportLayer)?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -14,55 +15,55 @@ import org.scalatest.{FlatSpec, Inspectors, Matchers}
...
+              for {
+                maybeBlock1 <- node.blockStore.get(signedBlock1.blockHash)
+                maybeBlock2 <- node.blockStore.get(signedBlock2.blockHash)
+              } yield {
+                withClue(s"Assertion failed for node ${node.local} --") {
+                  maybeBlock1 shouldBe Some(signedBlock1)
+                  maybeBlock2 shouldBe Some(signedBlock2)
+                }
+              }
+            }*/
```
tgrospic: Is there a reason to keep this comment out part?

tgrospic:
Please address in this PR fix mentioned in https://github.com/rchain/rchain/pull/3351/files#r592805237 about  the result from last finalized height constraint checker.

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
tgrospic: Changes in PR #3367 (https://github.com/rchain/rchain/pull/3367/commits/fe7821a3bab32abea7ff361c312acd3582209a84) should cover this fix so not needed here.

casper/src/main/scala/coop/rchain/casper/blocks/proposer/ProposeResult.scala
nzpr: I made this to be able to assert in tests `assert(r == ProposeResult.notEnoughBlocks && b.isEmpty)`
It does not work with case objects

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: slowcooker is not executed I guess, and this is very old code. I think not worth looking into

nzpr:
> Please address in this PR fix mentioned in https://github.com/rchain/rchain/pull/3351/files#r592805237 about the result from last finalized height constraint checker.

This is fix ^

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
zsluedem: We can consider removing it later and come up with other plans for that purpose.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: Its old leftover, I'll remove it

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: https://github.com/rchain/rchain/pull/3355/commits/00eb5badd57112a8e68e41498402476caf77db2a

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: No, removed https://github.com/rchain/rchain/pull/3355/commits/00eb5badd57112a8e68e41498402476caf77db2a

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: https://github.com/rchain/rchain/pull/3355/commits/00eb5badd57112a8e68e41498402476caf77db2a

nzpr:
Hm... one integration test is failing because of this https://github.com/rchain/rchain/pull/3355/commits/fe7bac71b2cc8199467f606af528db99d4a85113. Let me check

casper/src/main/scala/coop/rchain/casper/blocks/proposer/Proposer.scala
```diff
@@ -84,18 +84,23 @@ class Proposer[F[_]: Concurrent: Log: Span](
...
-    val work = Stream(s1, s2, s3)
-    work
-      .parJoin(3)
-      .compile
-      .toList
-      // pick some result that is not Success, or return Success
-      .map(_.find(_ != CheckProposeConstraintsSuccess).getOrElse(CheckProposeConstraintsSuccess))
-  }
+  ): F[CheckProposeConstraintsResult] =
+    checkActiveValidator(s, validator) match {
```
nzpr, line 102: As sync constraints depend on last message - for not bonded validators there is no such message. So first we check bonding and then check constraint, that might rise an error if no latest message for bonded validator found.

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
nzpr: I suggest first to address concurrent block processing performance https://github.com/rchain/rchain/issues/3346 and then get back to writing tests

casper/src/test/scala/coop/rchain/casper/addblock/ProposerSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.models.blockImplicits.getRandomBlock
+import coop.rchain.shared.Log
+import monix.eval.Task
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.collection.immutable.HashMap
+
+class ProposerSpec extends FlatSpec with Matchers with BlockDagStorageFixture {
```
zsluedem, line 25: ~~Hm.....I don't know how to put this. The tests here seem to test themselves instead of testing something very important in production. I used to write tests like these in other projects.~~

zsluedem: Oh. Nvm. I missed something. They are actually good.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
```diff
@@ -46,39 +46,6 @@ class MultiParentCasperMergeSpec extends FlatSpec with Matchers with Inspectors
...
   }
 
-  // TODO enable tests below
-//  it should "handle multi-parent blocks correctly when they operate on stdout" in effectTest {
-//    def echoContract(no: Int) =
-//      Rho(s"""new stdout(`rho:io:stdout`) in { stdout!("Contract $no") }""")
-//    merges(echoContract(1), echoContract(2), Rho("Nil"))
-//  }
-//
-//  it should "not conflict on registry lookups" in effectTest {
```
zsluedem, line 56: I think this one is still a worthy one. We can create another issue seperate for that.

zsluedem: This can be extended to our base rholang library and see what would cause conflicts.

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -1,247 +1,229 @@
...
+  def update(bm: Option[BlockMessage]): Unit = this.lastBlock = bm
+}
+
+object HashSetCasperActions {
+  import GenesisBuilder._
+
+  def context(
+      amount: Int,
+      bondsGen: Seq[PublicKey] => Map[PublicKey, Long]
```
zsluedem: `bondsGen` is not used.

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -1,247 +1,229 @@
...
+  ): GenesisContext = {
+    val genesis =
+      buildGenesis(amount)
+    genesis
```
zsluedem: the expression here can be simplified.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
```diff
@@ -90,88 +70,96 @@ class MultiParentCasperCommunicationSpec extends FlatSpec with Matchers with Ins
...
+        // this block will be propagated to all nodes and force nodes(2) to ask for missing blocks.
+        br <- makeDeploy(0) >>= (nodes(0).addBlock(_)) // block h1
+
+        _ <- TestNode.propagate(nodes) // force the network to communicate
+
+        _ <- nodes(2).contains(br.blockHash) shouldBeF true
+
+        nr <- makeDeploy(0) >>= (nodes(2).addBlock(_))
+      } yield {
+        PrettyPrinter.buildString(nr.header.parentsHashList) shouldBe PrettyPrinter
```
zsluedem: `buildString` would truncate the whole message which can be a problem in some extreme cases. Maybe `buildStringNoLimit` or just compare the hashes

casper/src/test/scala/coop/rchain/casper/addblock/ProposerSpec.scala
```diff
@@ -0,0 +1,227 @@
...
+            val p = new Proposer[Task](
+              validateBlock = alwaysSuccesfullValidation,
+              checkFinalizedHeight = okProposeConstraint,
+              checkActiveValidator = alwaysActiveF,
+              checkEnoughBaseStake = okProposeConstraint,
+              getCasperSnapshot = getCasperSnapshotF,
+              createBlock = createBlockF,
+              proposeEffect = proposeEffect(10),
+              validator = dummyValidatorIdentity
+            )
```
tgrospic, line 215: Code for creating Proposer is almost identical in all tests. This should be extracted in separate function so only different parts are specified in each test. I'm ok with this now because we will change Proposer anyway.

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
tgrospic: Please update #3346 with this info and/or create new issue for block processor tests.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
tgrospic: In async version we have Deferred to be able to wait on propose to complete. Why we cannot use it here?

casper/src/main/scala/coop/rchain/casper/blocks/proposer/ProposeResult.scala
tgrospic: I think you are pointing to another class ProposeResult.

But the reason is not case object because the same is with case classes. It's about unifying the type, in this example each constructor has a function returning `CheckProposeConstraintsResult`.

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
```diff
@@ -1,265 +0,0 @@
-package coop.rchain.casper.api
```
tgrospic, line 1: Are there replacement tests for create block API?

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
tgrospic: > slowcooker is not executed I guess, and this is very old code. I think not worth looking into

We should be careful in assessing the existing code, just because it's old is not a reason to be ignored or abandon it.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: propagate uses `.handleReceive()` which is just adding block to the queue. Previously `.handleReceive()` was blocking until block is processed.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: Nice catch, I agree. I don't like comparing just hashes because in case of test failure it does not print valuable information - what the hashes are actually, only non readable byte string.

casper/src/test/scala/coop/rchain/casper/engine/RunningHandleBlockMessageSpec.scala
nzpr: https://github.com/rchain/rchain/issues/3346#issuecomment-818549965

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: It was not compiling due to changes in block merge branch, so I commented that out temporarily. This PR adds minor fixes for it to compile. Might be I should use /* ... */ comments style for that, so its more clear what has been changed actually

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: It's old code and not used, do you want me to fix it or leave it as it is?

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: It's old code and not used, do you want me to fix it or leave it as it is?

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
nzpr: https://github.com/rchain/rchain/blob/34146e9a8f29cd49fe2a814ba7d0a78fe5f69bd6/casper/src/test/scala/coop/rchain/casper/addblock/ProposerSpec.scala#L25 is the replacement

casper/src/test/scala/coop/rchain/casper/addblock/ProposerSpec.scala
nzpr: Just a general remark related to this - as we were discussing general approach to building logic inside pure state. Though I agree this might be extracted to dedicated function, still we need to override quite a lot, so not sure it will gain benefit. To me its like testing an equation -  you pass in variables and look at the output, having ability to set some defaults for variables in this equation tells that you might not even need this variables. 
But yes, I did not think a lot which values for functions are used in all or most tests, might be there is a room for shrinking the code.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: https://github.com/rchain/rchain/pull/3355/commits/fe99286a4e6aade733a95e754960f688475dc21e

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: https://github.com/rchain/rchain/pull/3355/commits/fe99286a4e6aade733a95e754960f688475dc21e

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
nzpr: https://github.com/rchain/rchain/pull/3355/commits/fe99286a4e6aade733a95e754960f688475dc21e

nzpr:
Tests were not compiling, odd because its working locally. Rebased on recent feature branch, lets see.

nzpr:
@tgrospic @zsluedem I guess I've addressed all  comments

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
tgrospic: Yeah, block comments should be better in this case. I tried running this test on dev branch but it's failing there also. We need to check if it make sense to fix it or if it's not needed anymore.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
tgrospic: What is printed is whatever is returned on calling `.toString` on that object. We can create override of `toString` to customize  friendly string representation which is also useful in debugging.
I'm not sure if ScalaTest have some other way to do it e.g. using cats with Show instances which would be more elegant solution.

casper/src/test/scala/coop/rchain/casper/addblock/ProposerSpec.scala
tgrospic: Here is simplified example of what I mean. With the _truthTable_ we immediately see what is tested and what results we expect.

```scala
def doSomethind(a: Int, b: String, c: List[Int]) = ???

val r1 = doSomething(1, "First", List(42))
val r2 = doSomething(2, "Second", List(42))
val r3 = doSomething(3, "Third", List(42))

assert(r1 == 11)
assert(r2 == 12)
assert(r3 == 13)

// Instead we can recognize what input and output we want to compare.
val truthTable = List(
  (1, "First", 11),
  (2, "First", 12),
  (3, "First", 13),
)
// And how it's called is not important and can be written only once.
truthTable.map((arg1, arg2, result) => assert(doSomething(arg1, arg2, List(42)) == result))
```

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
tgrospic: I don't see that BlockAPI is used at all in Proposer tests. I'm I missing something?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
tgrospic: I'm not sure what are you suggesting. Should we fix this test later or it's not needed anymore because it's replaced by some other test?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
tgrospic: Do we have test with stdout and registry lookup when checking *all mergeability rules*?

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
nzpr: I believe how CreateBlockAPITest is written - it is a test for proposer logic not for API. We might write another test for blockAPI to test if it is returning valid responses. Maybe do this after this <https://github.com/rchain/rchain/blob/956680dbef1063e518912405f909618be28d0570/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L118-L120>?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperMergeSpec.scala
zsluedem: https://github.com/rchain/rchain/issues/3381

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
nzpr: I think we should replace this test. Basically it is an integration test which ensures that 1) node will add dependencies to block retriever 2) block addition will trigger processing of pendent descendants 3) casper can validate block. Test for 3) already exist, test for 1 is in the same file. We can modify this particular test to check if dependencies are actually received and are in casper buffer. I've added commit for that https://github.com/rchain/rchain/pull/3355/commits/e52588e3cc1d55504429c51e08479c9d00fdf5f4

The remainder is logic for 2), which is now implemented in `node` project. I'm not sure what's the best way to do that - I'm thinking to look into concurrent block processing and maybe after that check this test.

casper/src/test/scala/coop/rchain/casper/api/CreateBlockAPITest.scala
tgrospic: No doubt we need to redefine block API. I'm ok to leave this as part of those changes.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperCommunicationSpec.scala
tgrospic: OK, let's keep it in reduced form.

tgrospic (approved):
LGTM

PR #3358, opened 2021-03-22
Propose on start (for --autopropose)

nzpr:
## Overview
When validator is connected to existing network, it should try to create block after Casper is initialised and it found peers, so operator does not have to manually issue propose.

This is just a usability improvement.


tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

PR #3359, opened 2021-03-23
Remove caching from RSpace

### LINKED ISSUE #3340: Refactor KeyValueCache (implement size limit)
## Overview

NOTE: Issue is extended to also implement size limit so that memory doesn't grow indefinitely. 

Change `KeyValueCache#get` method to return `F[Option[V]]` because *fallback* may not always be available in which case the question is what to supply as `F[V]`. When Option is returned it's easy to do _orElse_ with fallback parameter.
https://github.com/rchain/rchain/blob/230b2515f369af344e83af30fb31af83dce44bc8/shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala#L8-L9

Here is an example how `HashHistoryReader` can be refactored using `OptionT`.
```scala
override def getRichDatums(hash: Blake2b256Hash): F[Seq[RichDatum[A]]] = {
  def cachedData: F[Option[Seq[RichDatum[A]]]] = ??? // Returned from cache

  val data = OptionT(fetchData(hash))
    .semiflatMap {
      case DataLeaf(bytes) =>
        decodeDataWithIntermediaries[A](bytes).map(Function.tupled(RichDatum.apply)).pure[F]
      case p =>
        new RuntimeException(
          s"Found unexpected leaf while looking for data at key $hash, data: $p"
        ).raiseError[F, Seq[RichDatum[A]]]
    }
    .getOrElse(Seq.empty[RichDatum[A]])

  OptionT(cachedData) getOrElseF data
}
```

P.S.
`OptionT` is similar in spirit to *fs2.Stream*, it doesn't contain many values but only one. It's also a small DSL for `F[_]` which has **introduction** functions (lift `F` values) and **elimination** functions (exit back to `F`).
Here are some similarities to fs2.Stream.
```scala
// Introduction - from F[_]
OptionT.liftF(F[A]) ~ Stream.eval(F[A])
OptionT(F[Option[A]]) ~ Stream.emits(Seq[A])

// Elimination - to F[_]
optT.value            ~ stream.compile.last
optT.getOrElseF(F[A]) ~ stream.compile.compile.lastOrError // sort of

// Map, tap with F[_]
optT.semiflatMap(A => F[B]) ~ stream.evalMap(A => F[B])
optT.semiflatTap(A => F[B]) ~ stream.evalTap(A => F[B])
```

nzpr:
## Overview
This started as an attempt to create on/off switch for RSpace caching and implement cache invalidation, to not have memory consumption growing over the time.

But as outlined in https://github.com/rchain/rchain/issues/3340, caching does not improves merging much. So not worth pursuing this path right now as @tgrospic suggests. This PR clears the codebase removing everything related to RSpace caching.


shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
```diff
@@ -5,42 +5,81 @@ import cats.effect.Concurrent
...
 final case class NoOpKeyValueCache[F[_]: Applicative, K, V]() extends KeyValueCache[F, K, V] {
   override def get(key: K, fallback: F[V]): F[V] = fallback
   override def put(key: K, value: V): F[Unit]    = ().pure[F]
   override def toMap: F[Map[K, V]]               = Map.empty[K, V].pure[F]
 }
 
 class LazyAdHocKeyValueCache[F[_]: Concurrent, K, V] private[LazyAdHocKeyValueCache] (
-    cache: Ref[F, Map[K, Deferred[F, V]]]
+    cache: Ref[F, KVCache[F, K, V]],
+    disableCache: Boolean
```
tgrospic: The purpose of this class is to implement cache. It doesn't make sense to have a flag to disable it.
It should be responsibility of the consumer to not use if that is desirable.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
```diff
@@ -5,42 +5,81 @@ import cats.effect.Concurrent
...
+}
+
+final case class FixedSizeLRUInvalidationPolicy[V](
+    capacity: Long,
+    useOrder: LinearSeq[V] = LinearSeq()
+) extends InvalidationPolicy[V] {
+  def processNewInsert(item: V): (Option[V], InvalidationPolicy[V]) =
+    useOrder match {
+      case Seq() => (none[V], FixedSizeLRUInvalidationPolicy(capacity, useOrder))
+      case v     => (v.head.some, FixedSizeLRUInvalidationPolicy(capacity, useOrder.tail :+ item))
```
tgrospic: Append operation on a Seq is linear so not the best choice here.
https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
```diff
@@ -5,42 +5,81 @@ import cats.effect.Concurrent
...
+
 trait KeyValueCache[F[_], K, V] {
   def get(key: K, fallback: F[V]): F[V]
   def put(key: K, value: V): F[Unit]
   def toMap: F[Map[K, V]]
 }
 
+trait InvalidationPolicy[V] {
+  def processNewInsert(item: V): (Option[V], InvalidationPolicy[V])
+}
```
tgrospic: Before we have clear understanding how limited cache will be implemented, defining some abstract definition will not make to much sense. Why this trait is necessary?

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
```diff
@@ -5,42 +5,81 @@ import cats.effect.Concurrent
 import cats.effect.concurrent.{Deferred, Ref}
 import cats.syntax.all._
 
+import scala.collection.immutable.LinearSeq
+
 trait KeyValueCache[F[_], K, V] {
   def get(key: K, fallback: F[V]): F[V]
```
tgrospic: Please take a look at issue #3340 to remove fallback from cache.

tgrospic:
@nzpr please also add some statistics about cache performance, enabled/disabled and also what is behavior when limit of the cache is reached.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
nzpr: I agree, It's just the easiest way to do it. And this class actually not the cache itself - its a reader that caches everything passing through, Cache is not quite good name here. But I'm not sure yet whats the best way to handle this.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
nzpr: I see, I used linear seq because https://docs.scala-lang.org/overviews/collections/seqs.html says it's optimised for `head` and `tail`.

tgrospic: That's true, but you are also using `:+` (append) which will iterate the whole list.

shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala
nzpr: To my understanding how cache is invalidated is something different to how cache is used. So I tried to write cache logic without implementation of invalidation. 

So invalidation check should happen after each new request for data that is not cached yet, resulting in expanding the cache. But how exactly this invalidation should influence cache - not known. 
Simplest is LRU cache with fixed map size, which is defined by order of item requested, and capacity of the seq (or set)
https://github.com/rchain/rchain/blob/bf08980d59c68f04fec14cee32518aa8aed9fc52/shared/src/main/scala/coop/rchain/store/LazyAdHocKeyValueCache.scala#L24-L33

What I found common for invalidation functions is that one should inputs some value (it might influence invalidation decision) and outputs option values to invalidate (can be set). Plus - the state that allows to make decision.

I tried first put everything in cache instance, but then trying to split task into different chunks, ended up with this solution.

nzpr:
> @nzpr please also add some statistics about cache performance, enabled/disabled and also what is behavior when limit of the cache is reached.

https://github.com/rchain/rchain/issues/3340#issuecomment-806059368
This PR is repurposed to clear cache related code.

tgrospic (approved):
LGTM

PR #3361, opened 2021-03-23
Validate content of genesis block against local config

### LINKED ISSUE #3360: Validation of genesis didn't fail when genesis validator had different quarantine-length

nzpr:
## Overview
Tests for block approver are limited to only ensuring that message processing is correct. But test of actual validation is not implemented. This PR implements these tests to be able to reproduce https://github.com/rchain/rchain/issues/3360

In addition, this PR introduces check for content of genesis block - it should be equal to blessed contracts validating node is expecting. This closes https://github.com/rchain/rchain/issues/3360 as a side effect.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
@@ -179,19 +179,21 @@ object BlockApproverProtocol {
 
         // check if expected blessed contracts and ones reported by candidate are equal
         // order should be the same and terms should be the same
-        wrongDeploys = (blockDeploys zip genesisBlessedContracts)
-          .mapFilter {
-            case (candidateDeploy, expectedContract) => {
-              val term    = candidateDeploy.deploy.data.term
-              val isWrong = term != expectedContract.data.term
-              if (isWrong) term.lines.findFirst.get.some else none[String]
-            }
+        wrongDeploys = (blockDeploys.iterator zip genesisBlessedContracts.iterator)
+          .filter {
+            case (candidateDeploy, expectedContract) =>
+              candidateDeploy.deploy.data.term != expectedContract.data.term
           }
+          .map {
+            case (candidateDeploy, _) =>
+              candidateDeploy.deploy.data.term.substring(0, 100)
+          }
+          .take(5)
 
         _ <- wrongDeploys.isEmpty
               .either(())
               .or(
-                s"Genesis candidate deploys do not match expected blessed contracts.\nBad contracts:\n${wrongDeploys
+                s"Genesis candidate deploys do not match expected blessed contracts.\nBad contracts (5 first):\n${wrongDeploys
                   .mkString("\n")}"
               )
       } yield (blockDeploys, block.body.state)
--- a/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
@@ -9,7 +9,6 @@ import coop.rchain.casper.ValidatorIdentity
 import coop.rchain.casper.genesis.Genesis
 import coop.rchain.casper.genesis.contracts._
 import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.ProtoUtil
 import coop.rchain.casper.util.rholang.RuntimeManager
 import coop.rchain.catscontrib.Catscontrib._
 import coop.rchain.comm.PeerNode
@@ -173,16 +172,34 @@ object BlockApproverProtocol {
             vaults,
             Long.MaxValue
           )
-          .toSet
         blockDeploys = block.body.deploys
         _ <- (blockDeploys.size == genesisBlessedContracts.size)
               .either(())
               .or("Mismatch between number of candidate deploys and expected number of deploys.")
+
+        // check if expected blessed contracts and ones reported by candidate are equal
+        // order should be the same and terms should be the same
+        wrongDeploys = (blockDeploys zip genesisBlessedContracts)
+          .mapFilter {
+            case (candidateDeploy, expectedContract) => {
+              val term    = candidateDeploy.deploy.data.term
+              val isWrong = term != expectedContract.data.term
+              if (isWrong) term.lines.findFirst.get.some else none[String]
+            }
+          }
+
+        _ <- wrongDeploys.isEmpty
+              .either(())
+              .or(
+                s"Genesis candidate deploys do not match expected blessed contracts.\nBad contracts:\n${wrongDeploys
+                  .mkString("\n")}"
+              )
       } yield (blockDeploys, block.body.state)
 
     (for {
       result                    <- EitherT(validate.pure[F])
       (blockDeploys, postState) = result
+      // check if candidate blessed contracts match candidate postStateHash
       stateHash <- EitherT(
                     runtimeManager
                       .replayComputeState(runtimeManager.emptyStateHash)(
--- a/casper/src/test/scala/coop/rchain/casper/engine/BlockApproverProtocolTest.scala
+++ b/casper/src/test/scala/coop/rchain/casper/engine/BlockApproverProtocolTest.scala
@@ -2,6 +2,7 @@ package coop.rchain.casper.engine
 
 import cats.Traverse
 import cats.implicits._
+import com.google.protobuf.ByteString
 import coop.rchain.casper.genesis.contracts.Vault
 import coop.rchain.casper.helper.TestNode
 import coop.rchain.casper.helper.TestNode._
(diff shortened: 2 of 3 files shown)


casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -178,32 +179,51 @@ object BlockApproverProtocol {
         _ <- (blockDeploys.size == genesisBlessedContracts.size)
               .either(())
               .or("Mismatch between number of candidate deploys and expected number of deploys.")
-      } yield (blockDeploys, block.body.state)
```
nzpr, line 181: The bug was here, what had been validated were processed deploys from unapproved block. So it does not matter what you put in local config, validation only was insuring that unapproved block received is replayable against its own processed deploys.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -178,11 +177,31 @@ object BlockApproverProtocol {
...
+              .or("Genesis candidate deploys do not match expected blessed contracts.")
       } yield (blockDeploys, block.body.state)
 
     (for {
       result                    <- EitherT(validate.pure[F])
       (blockDeploys, postState) = result
+      // check if candidate blessed contracts match candidate postStateHash
       stateHash <- EitherT(
                     runtimeManager
                       .replayComputeState(runtimeManager.emptyStateHash)(
```
tgrospic, line 207: Why we didn't see different hash when genesis validator replayed a block with different blessed contracts?

nzpr: Because `blockDeploys` here are deploys from candidate block itself. So there were no check at all that local config match candidate block received. The only check was that block received is replayable (deploys inside match state hash), even if blessed contracts inside candidate do not match our expectation.
https://github.com/rchain/rchain/blob/a89a137ef094f0873c1370f0aa52c6c06c9a1995/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala#L199-L212

nzpr: We cannot create genesis ourselves to check if hash matches candidate - due to nondeterminism hash will be different. Se have are limited to replaying processed deploys along with event logs from candidate.  But then we need to be sure that code being replayed is actually what we expect. This last part is added in PR.

tgrospic: We need to have cryptographic proof so any other comparison is a convenience to report more detailed errors.

Let's acknowledge what genesis validator needs to be able to replay the block in the same way as ceremony master.
It needs the same deploys and system deploys - we know we have it because we should run the same code. As you mentioned, this is not enough to replay due to nondeterminism in Rholang.
But we have event logs from ceremony master block as the second part of input which is then enough to successfully replay a block and get the final hash for crypto proof.

nzpr: Why not leave as its now? To compare terms that node is expect with terms that relayed?

We can use event logs from processed deploys in candidate to recreate processed deploy from local blessed contracts - but that is effectively the same operation? Except it is more relaxed in a way that terms should not be equal up to string, it should be just the same rholang code, possibly formatted differently.

nzpr: We might try to substitute term from local blessed contract, but how do we match processed deploy from candidate with local blessed contract? We can use only term for that, so we end up with necessity to have the same term. So its easier to just check that all local blessed terms are present in candidate deploys, and size of deploys in candidate equal size of local blessed contracts. Which means code executed in genesis candidate is the same that we are expecting.

tgrospic: If you are sure about that, can I be a ceremony master? :smiley: What if I execute some additional code not as part of deploy s and system deploys?

nzpr: You won't be able to execute it because I will check if terms in list of processed deploys from candidate you've sent me match terms that I expect to be in genesis. And number of expected terms equal number of processed deploys. So your processed deploys have to contain blessed contracts that I expect to see and only them. And then I replay your candidate to verify that processed deploys in your candidate create state with hash that you've claimed. This is effectively the same as I will invoke my blessed contracts with event log you've sent me, as terms are equal as I checked at step one.

tgrospic: And if I add some additional code in one of the system deploys?

nzpr: System deploys are empty for genesis and they explicitly set when replaying
https://github.com/rchain/rchain/blob/a89a137ef094f0873c1370f0aa52c6c06c9a1995/casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala#L205-L213

tgrospic: Per our discussion on standup. @nzpr convinced me, and Greg confirmed, that only deploys should be initial inputs to genesis (which is great) so equality check is all that is needed to validate before replay.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -178,11 +177,31 @@ object BlockApproverProtocol {
...
+            if (inconsistencyFound) (leftover, inconsistencyFound)
+            else {
+              blockDeploys.find(_.deploy.data.term == blessedDeploy.data.term) match {
+                case Some(_) =>
+                  (leftover - blessedDeploy, false)
+                case None =>
+                  (leftover, true)
+              }
+            }
+        }
```
tgrospic: Order should be also equal as from genesis creation.
https://github.com/rchain/rchain/blob/6507f3cf94859ef906eb0b0a4dd7f566babe14e5/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala#L62-L74

Just for fun you can use fs2, zip it in a stream, map in parallel for equality and print to log top three unequal with 50 characters from deploy term.

zsluedem (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -173,16 +172,34 @@ object BlockApproverProtocol {
...
+        // check if expected blessed contracts and ones reported by candidate are equal
+        // order should be the same and terms should be the same
+        wrongDeploys = (blockDeploys zip genesisBlessedContracts)
+          .mapFilter {
+            case (candidateDeploy, expectedContract) => {
+              val term    = candidateDeploy.deploy.data.term
+              val isWrong = term != expectedContract.data.term
+              if (isWrong) term.lines.findFirst.get.some else none[String]
+            }
+          }
```
tgrospic, line 186: To better understand what is executed, try this experiment. I'm interested what would you change in your code.

```scala
val items = 1 to 10

object test1 {
  val results = items
    .filter(_ < 5)
    .take(2)
    .map { x => println(s"Test 1 - number $x"); x }

  println(s"Test 1: $results\n")
}

object test2 {
  val results = items
    .filter(_ < 5)
    .map { x => println(s"Test 2 - number $x"); x
    }
    .take(2)

  println(s"Test 2: $results\n")
}

object test3 {
  import cats.syntax.all._

  val results = items.toList
    .mapFilter { x =>
      println(s"Test 3 - number $x")
      if (x < 5) x.some else none
    }
    .take(2)

  println(s"Test 3: $results\n")
}

test1; test2; test3
```

nzpr: I'm not sure what do you mean. I want to calculate `isWrong` for all pairs and pass down inconsistencies (first line of inconsistent deploy term, to be precise).

tgrospic: It can be many deploys so we don't need to check all of them. We can fail on first but also we can display three for example, as I described in my initial comment.
Also, one line can hold large string so that's why it make sense to extract only prefix just to recognize which deploy does not match.

What I mean by my sample code is to show how using List (Seq) and Stream is different and it is advisable to check what is exactly running to eliminate unnecessary executions.

nzpr: `we can display three for example` - you mean three inconsistent? how do we get them if we fail fast on first?
So I assume this will stop traversing `(blockDeploys zip genesisBlessedContracts)` List on first failure?
```scala 
    wrongDeploys = (blockDeploys zip genesisBlessedContracts)
      .filter {
        case (candidateDeploy, expectedContract) => {
          candidateDeploy.deploy.data.term != expectedContract.data.term
        }
      }
      .take(1)
```

tgrospic: Yes, we can choose how many items will be returned, one or three, after which iteration will stop.

Now, if we add one small change and use `Iterator` instead of the collection directly, behavior will be more like lazy without allocating intermediate collections. You will notice this is very similar to fs2 Pure streams.
```scala
object test2_iterator {
  val results = items.iterator
    .filter(_ < 5)
    .map { x => println(s"Test 2 iter - number $x"); x
    }
    .take(2)

  println(s"Test 2 iter: ${results.toList}\n")
}
```
Maybe you will remember @marcin-rzeznicki's PR #2812 with optimizations using iterators.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -173,16 +172,36 @@ object BlockApproverProtocol {
...
           )
-          .toSet
         blockDeploys = block.body.deploys
         _ <- (blockDeploys.size == genesisBlessedContracts.size)
               .either(())
               .or("Mismatch between number of candidate deploys and expected number of deploys.")
+
+        // check if expected blessed contracts and ones reported by candidate are equal
+        // order should be the same and terms should be the same
+        wrongDeploys = (blockDeploys.iterator zip genesisBlessedContracts.iterator)
```
tgrospic, line 182: Kudos for zipping iterators. :rocket:

tgrospic (approved):
LGTM

nzpr:
bors merge

Issue #3349, opened 2021-03-10
Appending of pending rewards is conflicting for large networks


nzpr:
## Overview
When running tests on reasonably large number of validators (30+), conflict detection between some blocks can lead to false positive.
Conflict that is detected is race for base produce - it is destroyed in both event logs. Thought it should not. At least it is expected for conflict resolution outcome to be always the same. This happens in precharge or refund, it is not clear where exactly. But once blocks of two validators are considered conflicting, the same happens on each subsequent attempt to merge newly created blocks. Channel that is conflicting is always the same.

### Impact
-- not clear

### Dependencies
--

### Design
--

### Steps to implement the solution
Best solution is to have debugger for Rholang, so `term - AST - ADT - event log hashes` translation is visible. But intuition says that the bug might be related to peek (also see https://github.com/rchain/rchain/issues/3200, https://github.com/rchain/rchain/issues/3268). Destroyed produce means linear consume happened on the same produce both in main and merging branches. And prechrge + refund should use peeks on potentially conflicting produces. 

### Steps for integration testing
--

### Resource estimation
--

zsluedem:
https://github.com/rchain/rchain/runs/2490010239?check_suite_focus=true
This can be reproduced by test. But it is more difficult with less nodes.

nzpr:
I cannot reproduce this with new merging logic in `deploy-merge` feature branch 59f787d050219213ef2d8cd398dd87ebe1a1eff8 so closing this issue

zsluedem:
https://github.com/rchain/rchain/runs/2801580638?check_suite_focus=true
reopen this issue because of the action above

The test action is based on dev. Maybe the bug is fixed in deploymerge branch. @nzpr

nzpr:
@zsluedem yes, dev branch has initial implementation of block merge, which has this issue. It should not be present in `feature/deploymerge`

nzpr:
After this PR https://github.com/rchain/rchain/pull/3444#discussion_r654188163 the bug is observable again

nzpr:
Here is some more data for network of 50 validators:
49 blocks are merged into base state, so 49 branches are present and conflicts against combination of then is computed. 

Here is size of conflicts for each branch
```
[...]

PR #3364, opened 2021-03-29
Basic support to debug Rholang channels to source code

tgrospic:
## Overview
This PR should enable (maybe just temporarily) debugging of the bug with merging #3349 to locate the channel involved in false positive conflict detection.

### DIFF
--- a/models/src/main/scala/coop/rchain/models/rholang/implicits.scala
+++ b/models/src/main/scala/coop/rchain/models/rholang/implicits.scala
@@ -1,13 +1,10 @@
 package coop.rchain.models.rholang
 
+import com.google.protobuf.ByteString
 import coop.rchain.models.Connective.ConnectiveInstance
 import coop.rchain.models.Connective.ConnectiveInstance._
 import coop.rchain.models.Expr.ExprInstance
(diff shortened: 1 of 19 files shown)


models/src/main/scala/coop/rchain/models/rholang/implicits.scala
```diff
@@ -327,7 +324,7 @@ object implicits {
 
     def singleDeployerId(): Option[GDeployerId] =
       if (p.sends.isEmpty && p.receives.isEmpty && p.news.isEmpty && p.exprs.isEmpty && p.matches.isEmpty && p.bundles.isEmpty && p.connectives.isEmpty) {
-        p.unforgeables.toList match {
+        p.unforgeables match {
```
zsluedem, line 327: I think it would be better to submit a small pr for this kind of fixes

tgrospic: @zsluedem All of the commits so far are only preparation and basically clean up with changes like this. It makes sense to extract it to separate PR.
Even more because I didn't find some nice way to extend normalizer to return additional info for debugging so it will probably be just temporary solution not entering main branch.

PR #3363, opened 2021-03-29
Remove legacy file based block dag store

nzpr:
## Overview
File based dag store (along with finalised block store) is not needed anymore once KV store is introduced. Migration from file to KV store can be done by running previous releases, e.g. https://github.com/rchain/rchain/releases/tag/v0.10.1, process is automatic and happen on startup. In future releases with this PR merged, file storage wont be supported anymore and migration won't be possible.

PR #3365, opened 2021-04-01
Merge dev into block-merge

nzpr:
Merge recent dev to block merge feature branch


node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
```diff
@@ -158,22 +156,6 @@ class NodeRuntime[F[_]: Monixable: ConcurrentEffect: Parallel: Timer: ContextShi
         */
       _ <- mkDirs(dataDir)
 
-      dagConfig = BlockDagFileStorage.Config(
```
nzpr, line 161: Additional commit is to apply changes from https://github.com/rchain/rchain/pull/3363 due to NodeRuntime has been split and moved to different package in block-merge branch

tgrospic (approved):
LGTM

PR #3366, opened 2021-04-01
New finalizer

nzpr:
Old finaliser were not suitable for multiparent. This PR suggests improved version. Only last commit is of interest, others are from not merged yet https://github.com/rchain/rchain/pull/3379.


shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,31 @@
...
+  implicit final def syntaxDagReader[F[_], A](
+      dag: DagReader[F, A]
+  ): DagReaderOps[F, A] = new DagReaderOps[F, A](dag)
+}
+
+final class DagReaderOps[F[_], A](
+    private val dag: DagReader[F, A]
+) {
+
+  def gatherAncestors(start: A, stopCondition: A => F[Boolean])(
```
tgrospic: Type parameter `A` can be defined on the method itself. In that way it can be resolved when method is called and not when implicit (syntax) is resolved.
In general type parameters (the same as value level variables) should be defined as late as possible.

P.S.
Class with extension methods should extend `AnyVal` to escape boxing on each call.
https://typelevel.org/cats/guidelines.html
> Ops classes should be marked final and extend AnyVal, to take full advantage of inlining and prevent unnecessary allocations.

shared/src/main/scala/coop/rchain/fs2/Fs2StreamSyntax.scala
```diff
@@ -84,4 +84,11 @@ class Fs2StreamOps[F[_], A](
     */
   def parEvalMapUnorderedProcBounded[F2[x] >: F[x]: Concurrent, B](f: A => F2[B]): Stream[F2, B] =
     stream.parEvalMapUnordered[F2, B](availableProcessors)(f)
+
+  /**
+    * Variant of [[Stream.evalFilterAsync]] with parallelism bound to number of processors.
+    */
+  def evalFilterAsyncProcBounded[F2[x] >: F[x]: Concurrent, B](f: A => F2[Boolean]): Stream[F2, A] =
+    stream.evalFilterAsync[F2](availableProcessors)(f)
```
tgrospic, line 92: :rocket:

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
tgrospic: Correction, type parameter `A` is part of DagReader definition so it must be defined on class level in Ops class.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,93 @@
...
+import fs2.Stream
+
+object Finalizer {
+
+  /**
+    * Run finalization through children until no new finalized child found.
+    * @param dag - dependency DAG
+    * @param startHash - starting point for finalization, should be some hash directly finalized before
+    * @param finalisationEffect - effect to be executed when some hash is finalized (pass fault tolerance threshold)
+    * @param lastFnalizedAdvanceEffect - effect to be executed when `last finalised` advances.
```
tgrospic: A typo in `Fnalized`.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,93 @@
...
+    for {
+      _ <- checkArgs
+      r <- startHash.some.tailRecM[F, Option[A]] {
+            case Some(h) =>
+              for {
+                nextOpt <- computeNextLFBWithEffects(h)
+                _       <- nextOpt.traverse(lastFinalisedAdvanceEffect)
+              } yield nextOpt.asRight[Option[A]]
+            case None => none[A].asLeft[Option[A]].pure[F]
+          }
```
tgrospic: Recursion can be done with fs2 streams by unfolding from the start value.
```scala
val finalizationUnfold =
  Stream
    .unfoldLoopEval(startHash)(computeNextLFBWithEffects(_).map(x => (x, x)))
    .unNoneTerminate
    .evalTap(lastFinalisedAdvanceEffect)

checkArgs *> finalizationUnfold.compile.last
```

casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
```diff
@@ -149,7 +147,7 @@ class TestNode[F[_]: Timer](
   val postGenesisStateHash = ProtoUtil.postStateHash(genesis)
 
   val shardConf = CasperShardConf(
-    -1,
+    0,
```
tgrospic: Please use named arguments for literal values to prevent errors when order of parameters of the same type change and also it's easier to read the code.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
nzpr: Oh, thats awesome! I tried to make recursion with `noneTerminate` by passing `tap`ing emitted value back to to queue, but I did it wrong and it hangs. Thanks for example!

nzpr:
Converting this to draft for now, more work to be done

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -48,6 +51,16 @@ final class BlockDagRepresentationOps[F[_]: Sync](
     fs2.Stream.emits(streams).parJoinUnbounded.compile.toList
   }
 
+  def latestMessageHashUnsafe(v: Validator)(
+      implicit line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing
```
tgrospic, line 57: With monix 3.0 and sync stack trace support we don't need to manually write source info.

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,191 @@
...
+
+    val compute = Span[F].trace(SafetyOracleMetricsSource) {
+      for {
+        // weight map of main parent (fallbacks to message itself if no parents)
+        mainParentsWeightMap <- {
+          dag
+            .mainParent(target)
+            .flatMap {
+              case Some(mainParent) => getWeightMap(mainParent)
+              case None             => getWeightMap(target)
```
tgrospic: Why stop on Option? You know you want to call `getWeightMap` on something. So resolve that first, `if not` mainParent `else` target, do getWeightMap on result.

nzpr: You mean 
```scala
dag.mainParent(target)
            .map {
              case Some(mainParent) => mainParent
              case None             => target
            }
            .flatMap(getWeightMap)
```
?

casper/src/test/scala/coop/rchain/casper/helper/TestNode.scala
```diff
@@ -199,8 +197,13 @@ class TestNode[F[_]: Timer](
 
   def propagateBlock(deployDatums: Signed[DeployData]*)(nodes: TestNode[F]*): F[BlockMessage] =
     for {
-      block <- addBlock(deployDatums: _*)
-      _ <- nodes.toList.traverse_ { node =>
+      _       <- Log[F].debug((s"\n$name creating block"))
```
nzpr, line 200: this is just optimisation I did when debugging one test issue - now it node does not propagate to itself an prints some useful logs

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,172 @@
...
+import coop.rchain.casper.safety.CliqueOracle.WeightMap
+import coop.rchain.dag.{Casper, DagOps, DagReader}
+import coop.rchain.metrics.{Metrics, Span}
+import coop.rchain.shared.Log
+import coop.rchain.shared.syntax._
+import fs2.Stream
+
+import scala.reflect.ClassTag
+
+/**
```
nzpr, line 17: This is explanation how new finaliser is supposed to work

nzpr:
I think this construct is quite odd, explanation is in comment. https://github.com/rchain/rchain/blob/61726ac897b55c1326d637ae715333153d9cdbda/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala#L134-L150
But also now after working on finaliser I think that this premature creation of latest message is not something that we want - this way we put stake behind block that bonding validator did not validate. So it's easier to cheat in block where someone new is bonded.

PR #3367, opened 2021-04-02
Disable channels map for single parent

nzpr:
Maintaining channels map is required (temporary) for merging blocks. For single parent it is not needed, but still being populated which is computationally expensive. This PR disables channels map population if node configured for single parent.


tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

PR #3370, opened 2021-04-06
Fix score for binary data, different ByteString should not be equal

zsluedem:
## Overview

https://discord.com/channels/391435113415573504/417429638642532363/828667175140982815

fix failing test case in https://github.com/rchain/rchain/runs/2268175723
```
2021-04-05T07:34:33.3160040Z [info] - should sort so that unequal terms have unequal scores and the other way around *** FAILED *** (2 seconds, 158 milliseconds)
2021-04-05T07:34:33.3167230Z [info]   TestFailedException was thrown during property evaluation.
2021-04-05T07:34:33.3175460Z [info]     Message: Node(List(Leaf(ScoreAtom(IntAtom(116))), Leaf(ScoreAtom(StringAtom(�))))) equaled Node(List(Leaf(ScoreAtom(IntAtom(116))), Leaf(ScoreAtom(StringAtom(�)))))
2021-04-05T07:34:33.3183342Z [info]     Location: (SortTest.scala:100)
2021-04-05T07:34:33.3189985Z [info]     Occurred when passed generated values (
2021-04-05T07:34:33.3196766Z [info]       arg0 = List(Expr(GByteArray(<ByteString@65e14bdc size=1 contents="\200">)), Expr(GByteArray(<ByteString@2f125f4b size=1 contents="\331">))) // 3 shrinks
2021-04-05T07:34:33.3203504Z [info]     )
```

### DIFF
--- a/casper/src/test/resources/PoSTest.rho
+++ b/casper/src/test/resources/PoSTest.rho
@@ -706,13 +706,14 @@ match (
           contract closeBlock(ackCh) = {
             new ackCh0, activeValidatorsCh, setBlockData(`rho:test:block:data:set`), blockDataSet in {
               @PoS!("closeBlock", sysAuthToken, *ackCh0) |
-              @PoS!("getActiveValidators", *activeValidatorsCh) |
-              for (@ack              <- ackCh0;
-                   @activeValidators <- activeValidatorsCh) {
-                // For testing we need one of the active validators to be the block sender
-                setBlockData!("sender", activeValidators.nth(0), *blockDataSet) |
-                for (_ <- blockDataSet) {
-                  ackCh!(ack)
+              for (@ack              <- ackCh0) {
+                @PoS!("getActiveValidators", *activeValidatorsCh) |
+                for (@activeValidators <- activeValidatorsCh){
+                  // For testing we need one of the active validators to be the block sender
+                  setBlockData!("sender", activeValidators.nth(0), *blockDataSet) |
+                  for (_ <- blockDataSet) {
+                    ackCh!(ack)
+                  }
                 }
               }
             }
@@ -793,7 +794,7 @@ match (
                       @PoS!("refundDeploy", refundAmt, sysAuthToken, *refundCh) |
                       for (@(true, _) <- chargeCh;
                            @(true, _) <- refundCh) {
-                        match activeValidators.nth(1) {
+                        match activeValidators.nth(3) {
                           slashedValidator => {
                             computeHash!(slashedValidator, *hashCh) |
                             for (@hash <- hashCh) {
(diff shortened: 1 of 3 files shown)


zsluedem:
Changing the sorting for byteString should be would change the result like `map.toList()` in rholang because the `toList` must be sorted to get the same state hash.


Example
```
new stdout(`rho:io:stdout`), list in {
   list!({"0000".hexToBytes():10000,
"1111".hexToBytes():10000,
"2222".hexToBytes():10000,
"3333".hexToBytes():10000,
"4444".hexToBytes():10000,
"5555".hexToBytes():10000,
"6666".hexToBytes():10000,
"7777".hexToBytes():10000,
"8888".hexToBytes():10000,
"9999".hexToBytes():10000,
"aaaa".hexToBytes():10000,
"bbbb".hexToBytes():10000,
"cccc".hexToBytes():10000,
"dddd".hexToBytes():10000,
"eeee".hexToBytes():10000})|
   for (@l <- list) {
     stdout!(l.toList().nth(0))
   }
 }
```

tgrospic:
@zsluedem We can now continue with this change for HF2.

casper/src/test/resources/PoSTest.rho
```diff
@@ -706,13 +706,14 @@ match (
...
-                for (_ <- blockDataSet) {
-                  ackCh!(ack)
+              for (@ack              <- ackCh0) {
+                @PoS!("getActiveValidators", *activeValidatorsCh) |
+                for (@activeValidators <- activeValidatorsCh){
+                  // For testing we need one of the active validators to be the block sender
+                  setBlockData!("sender", activeValidators.nth(0), *blockDataSet) |
+                  for (_ <- blockDataSet) {
+                    ackCh!(ack)
+                  }
```
zsluedem, line 716: change the logic here to be sequence here to make sure the setting block sender would be an active validators.
Because in some tests, some validators would be slashed or unbonded which would cause the sender is not activate validator.

nzpr: Is this fix related to main topic of PR?

zsluedem: The goal here is to fix tests.

tgrospic: I wonder how it worked so far. It seems by pure luck.

Keys for bonded validators are generated each time so it's strange that we didn't seen some of them slashed or unbonded.
https://github.com/rchain/rchain/blob/bf001ddd520818e6dd28bc59ff4f8d82e07370cc/casper/src/test/scala/coop/rchain/casper/genesis/contracts/PoSSpec.scala#L14-L20

tgrospic (approved):
LGTM

casper/src/test/resources/PoSTest.rho
zsluedem: The random generated key would always get the same prefix byte which would get a determined order of the list every time. In tests, the codes use `nth` which would get the same group of validators(One group is generated ones and another one group is `"1" * 130`, `"2" * 130` and etc. ) each time for bonding and unbonding things. Changing the order make it break.

PR #3373, opened 2021-04-08
Introduce MergingDeploysStatus with rejected flag

### LINKED ISSUE #3372: Change rejected deploys in block to be a flag
## Overview

With PR #3304 rejected deploys list are added to block message. When validator creates a merged block it can specify which deploys are rejected (ignored) in the current merging scope which contains all blocks from tips to Last Finalized Block (LFB). Later, with new blocks, in the new merging scope rejected deploys can be changed and previously rejected now can again be included. So this means as part of process of merging some deploys can be rejected in one block and included again in later blocks (or vice versa).

To more easily track these transformations we need to extend the format of rejected deploys inside of block to have a flag for each deploy if it's rejected or not.

zsluedem:
## Overview

This PR changes RejectedDeploy used in a block to be MergingDeploysStatus with rejected flag.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -131,7 +131,7 @@ object BlockCreator {
                   preStateHash,
                   postStateHash,
                   processedDeploys,
-                  rejectedDeploys,
+                  mergingDeploysStatus,
                   processedSystemDeploys
                 )             = checkpointData
                 newBonds      <- runtimeManager.computeBonds(postStateHash)
@@ -145,7 +145,7 @@ object BlockCreator {
                   preStateHash,
                   postStateHash,
                   processedDeploys,
-                  rejectedDeploys,
+                  mergingDeploysStatus,
                   processedSystemDeploys,
                   newBonds,
                   shardId,
@@ -167,7 +167,7 @@ object BlockCreator {
       preStateHash: StateHash,
       postStateHash: StateHash,
       deploys: Seq[ProcessedDeploy],
-      rejectedDeploys: Seq[ProcessedDeploy],
+      mergingDeploysStatus: Seq[ProcessedDeploy],
       systemDeploys: Seq[ProcessedSystemDeploy],
       bondsMap: Seq[Bond],
       shardId: String,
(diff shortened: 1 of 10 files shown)


models/src/main/protobuf/CasperMessage.proto
```diff
@@ -183,6 +183,7 @@ message BodyProto {
 
 message RejectedDeployProto{
   bytes sig = 1;
+  bool  unRejected = 2;
```
tgrospic: This object now looks a little bit strange.

Maybe we can rename the flag to be just `rejected` and the whole object something like `ConflictingDeploy` or with merging in name like `MergingDeployStatus`?

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -178,11 +178,12 @@ message BodyProto {
   repeated ProcessedDeployProto deploys      = 2;
   repeated ProcessedSystemDeployProto systemDeploys = 3;
   bytes                    extraBytes   = 4;
-  repeated RejectedDeployProto rejectedDeploys = 5;
+  repeated MergingDeployStatusProto mergingDeployStatuses = 5;
 }
 
-message RejectedDeployProto{
+message MergingDeployStatusProto{
```
tgrospic, line 184: Why do we need to define the same message on two places? Here and in DeployServiceCommon.proto.

nzpr:
I've added a comment on issue <https://github.com/rchain/rchain/issues/3372#issuecomment-817338656> as I'm not totally understand the purpose of this change

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -178,7 +178,9 @@ object BlockCreator {
       Body(
         state,
         deploys.toList,
-        rejectedDeploys.map(r => RejectedDeploy(r.deploy.sig)).toList,
+        // TODO check if the deploys are unrejected, currently default is false
```
tgrospic: Default is _rejected_.

zsluedem: oh. Right.For get about the comment

tgrospic (commented):
All looks good! Let's wait until we are sure how will merging use this flag.

zsluedem:
@nzpr @tgrospic What is the decision now? I don't think this pr doesn't hurt the other logic even if we won't introduce `UnRejected` in the future although it would create a little bit ambigious on reading the codes.

tgrospic:
> What is the decision now? I don't think this pr doesn't hurt the other logic even if we won't introduce `UnRejected` in the future although it would create a little bit ambigious on reading the codes.

@zsluedem We can put it on hold in draft mode. The first version of block-merge will always have all rejected deploys is scope so every directly finalized block will have overall decision what is rejected.

PR #3377, opened 2021-04-13
Disable block index caching on single parent

nzpr:
Block index cache is optimisation for block merge. For now there is no purging policy implemented, so it source of memory leak. This PR disables cache population on block validation if node is configured for single parent, which is sufficient to disable this leak when node is running without merging. 

Another place where cache records can be created is
https://github.com/rchain/rchain/blob/dec10c2adb6f694e026b3a7fcb37922da4af5987/casper/src/main/scala/coop/rchain/casper/blocks/merger/Indexer.scala#L163-L172
but this code won't be executed with single parent.


tgrospic (approved):
LGTM

PR #3379, opened 2021-04-15
Clique oracle rewrite

nzpr:
This PR is first in a series aiming to enable leaderless merging.

New clique oracle is a drop-in replacement for old one written in abstract way. Main purpose is to have clear code and path towards new finalizer. No test are changed (except one) which should give confidence about correct implementation. 

The next step is https://github.com/rchain/rchain/pull/3366 which should use only `computeMaxCliqueWeight` public function from new oracle.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
+++ b/casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
@@ -6,10 +6,12 @@ import cats.instances.list._
 import coop.rchain.catscontrib._
 import Catscontrib._
 import cats.data.OptionT
-import cats.effect.Sync
+import cats.effect.{Concurrent, Sync}
 import coop.rchain.blockstorage.dag.BlockDagRepresentation
 import coop.rchain.blockstorage.syntax._
 import coop.rchain.casper.protocol.Justification
+import coop.rchain.casper.safety.CliqueOracle
+import coop.rchain.casper.safety.CliqueOracle.normalizedFaultTolerance
 import coop.rchain.casper.util.ProtoUtil._
 import coop.rchain.casper.util.{Clique, ProtoUtil}
 import coop.rchain.dag.DagOps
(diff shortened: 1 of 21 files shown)


casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -55,177 +57,29 @@ trait SafetyOracle[F[_]] {
...
-          candidateMetadata: BlockMetadata,
-          targetBlockHash: BlockHash
-      ): F[Boolean] =
-        isInMainChain(blockDag, candidateMetadata, targetBlockHash)
+      CliqueOracle.normalizedFaultTolerance(
+        blockDag,
+        candidateBlockHash,
+        (h: BlockHash) => blockDag.lookupUnsafe(h).map(_.weightMap),
+        (h: BlockHash) => blockDag.lookupUnsafe(h).map(_.blockNum),
+        (v: Validator) => blockDag.latestMessageHashUnsafe(v),
```
nzpr, line 79: Thought here we are reading twice unnecessary - is it that bad? Should `blockDag.lookupUnsafe` result be cached by LMDB in memory?

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
```diff
@@ -55,177 +57,29 @@ trait SafetyOracle[F[_]] {
...
-          blockDag: BlockDagRepresentation[F],
-          candidateMetadata: BlockMetadata,
-          targetBlockHash: BlockHash
-      ): F[Boolean] =
-        isInMainChain(blockDag, candidateMetadata, targetBlockHash)
+      CliqueOracle.normalizedFaultTolerance(
+        blockDag,
+        candidateBlockHash,
+        (h: BlockHash) => blockDag.lookupUnsafe(h).map(_.weightMap),
+        (h: BlockHash) => blockDag.lookupUnsafe(h).map(_.blockNum),
```
nzpr, line 78: I wonder - is it better to created BlockDagRepresentation extension that will be DagReader[F, BlockMetadata] and get rid of `F[_]` in CliqueOracle and some other places? Or this more fine grained access to F[_] is better? I'm curious how badly performance hit here - duplicate lookupUnsafe should be mitigated by LMDB cache at some portion?

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -59,10 +59,11 @@ class BlockQueryResponseAPITest
       setValidator = sender.some,
       setDeploys = randomDeploys.some,
       setJustifications = List(Justification(bondsValidator.validator, genesisBlock.blockHash)).some,
+      setParentsHashList = List(genesisBlock.blockHash).some,
```
nzpr, line 62: This test has been failing on `dag.unsafeLookup` method because default generated parent that is not in dagstore has been used. 
In previous code this was mitigated by how code is has been written
```scala
      private def computeMainParentWeightMap(
          candidateMetadata: BlockMetadata
      ): F[Map[BlockHash, Long]] =
        candidateMetadata.parents.headOption match {
          case Some(parent) =>
            blockDag.lookup(parent).map(_.fold(candidateMetadata.weightMap)(_.weightMap))
          case None => candidateMetadata.weightMap.pure[F]
        }
```

But in general - if we want to know fault tolerance of some block, I believe it's safe to assume (and we ned to force that) that if some message has main parent, that main parent has to be in dagstore. Otherwise throwing error is reasonable.

Fault tolerance is adjusted because target message is tip, so should have lowest possible value, and previously it had fault tolerance `+1` because default parent has been used.

nzpr:
@tgrospic @zsluedem feel free to review this, but I won't assign you on dev board as this PR should be targeted to another feature that we'll create after <https://github.com/rchain/rchain/pull/3355> is merged and then `block-merge` feature branch merged to dev

nzpr:
@tgrospic @zsluedem feel free to review this, but I won't assign you on dev board as this PR should be targeted to another feature that we'll create after <https://github.com/rchain/rchain/pull/3355> is merged and then `block-merge` feature branch merged to dev

casper/src/main/scala/coop/rchain/casper/SafetyOracle.scala
nzpr: I was thinking for a while and my guess is that better to leave for input functions of CliqueOracle possibility to not be pure, but inside caller decide if we need to get e.g. `DagReader[F, BlockMetadata]` and then pass in something like `meta.weightMap.pure[F]`, or acquire values from `F[_]` for each function separately. Otherwise we have to assume that `M` type have to have certain properties without ability to read from database.

PR #3387, opened 2021-04-25
Clique oracle rewrite

nzpr:
This is straight replacement for <https://github.com/rchain/rchain/pull/3379> as it has been closed  by bors and base `block-merge` branch has been deleted, so PR cannot be reopened.

This PR is an effort to refactor clique oracle to enable new finalizer. 

New clique oracle is a drop-in replacement for old one written in abstract way. Main purpose is to have clear code and path towards new finalizer. No test are changed (except one) which should give confidence about correct implementation.

The next step is #3366 which should use only `computeOutput` public function from new oracle.

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
@@ -3,8 +3,6 @@ package coop.rchain.blockstorage.dag
 import cats.effect.{Concurrent, Sync}
 import cats.syntax.all._
 import coop.rchain.casper.PrettyPrinter
-import coop.rchain.casper.protocol.Justification
-import coop.rchain.dag.Casper
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.BlockMetadata
 import coop.rchain.models.Validator.Validator
@@ -113,19 +111,4 @@ final class BlockDagRepresentationOps[F[_]: Sync](
       ib <- dag.invalidBlocks
       r  = ib.map(block => (block.blockHash, block.sender)).toMap
     } yield r
-
-  def getCasperJustificationsUnsafe(
-      blockHash: BlockHash
-  ): F[Set[Casper.Justification[BlockHash, Validator]]] =
-    lookupUnsafe(blockHash).map(
-      _.justifications.map {
-        case Justification(validator, latestBlockHash) =>
-          Casper.Justification(latestBlockHash, validator)
-      }.toSet
-    )
-
-  def toCasperJustificationUnsafe(
-      blockHash: BlockHash
-  ): F[Casper.Justification[BlockHash, Validator]] =
-    lookupUnsafe(blockHash).map(m => Casper.Justification(m.blockHash, m.sender))
 }
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
@@ -3,6 +3,8 @@ package coop.rchain.blockstorage.dag
 import cats.effect.{Concurrent, Sync}
 import cats.syntax.all._
 import coop.rchain.casper.PrettyPrinter
+import coop.rchain.casper.protocol.Justification
+import coop.rchain.dag.Casper
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.BlockMetadata
 import coop.rchain.models.Validator.Validator
@@ -14,6 +16,7 @@ trait BlockDagRepresentationSyntax {
 }
 
 final case class BlockDagInconsistencyError(message: String) extends Exception(message)
+final case class NoLatestMessage(message: String)            extends Exception(message)
 
 final class BlockDagRepresentationOps[F[_]: Sync](
     // BlockDagRepresentation extensions / syntax
@@ -46,6 +49,16 @@ final class BlockDagRepresentationOps[F[_]: Sync](
   ): F[List[BlockMetadata]] = {
     val streams = hashes.map(h => fs2.Stream.eval(lookupUnsafe(h)))
     fs2.Stream.emits(streams).parJoinUnbounded.compile.toList
+  }
+
+  def latestMessageHashUnsafe(v: Validator)(
+      implicit line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing
+  ): F[BlockHash] = {
+    def source = s"${file.value}:${line.value} ${enclosing.value}"
+    def errMsg = s"No latest message for validator ${PrettyPrinter.buildString(v)}\n  $source"
+    dag.latestMessageHash(v) >>= (_.liftTo(NoLatestMessage(errMsg)))
   }
 
   def childrensMetas(hashes: Seq[BlockHash])(
@@ -100,4 +113,19 @@ final class BlockDagRepresentationOps[F[_]: Sync](
       ib <- dag.invalidBlocks
       r  = ib.map(block => (block.blockHash, block.sender)).toMap
     } yield r
+
+  def getCasperJustificationsUnsafe(
+      blockHash: BlockHash
+  ): F[Set[Casper.Justification[BlockHash, Validator]]] =
+    lookupUnsafe(blockHash).map(
+      _.justifications.map {
+        case Justification(validator, latestBlockHash) =>
+          Casper.Justification(latestBlockHash, validator)
+      }.toSet
+    )
+
+  def toCasperJustificationUnsafe(
+      blockHash: BlockHash
+  ): F[Casper.Justification[BlockHash, Validator]] =
+    lookupUnsafe(blockHash).map(m => Casper.Justification(m.blockHash, m.sender))
 }
--- a/casper/src/main/scala/coop/rchain/casper/BlockDagRepresentationSyntax.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockDagRepresentationSyntax.scala
@@ -1,0 +1,37 @@
+package coop.rchain.casper
+
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.blockstorage.dag.BlockDagRepresentation
+import coop.rchain.blockstorage.syntax._
+import coop.rchain.casper.protocol.Justification
+import coop.rchain.casper.safety.CliqueOracle
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.models.Validator.Validator
+
+trait BlockDagRepresentationSyntax {
+  implicit final def casperSyntaxBlockDagRepresentation[F[_]](
+      dag: BlockDagRepresentation[F]
+  ): BlockDagRepresentationOps[F] = new BlockDagRepresentationOps[F](dag)
+}
+
+final class BlockDagRepresentationOps[F[_]](
+    private val dag: BlockDagRepresentation[F]
+) extends AnyVal {
+  def getCasperJustificationsUnsafe(
+      blockHash: BlockHash
+  )(implicit sync: Sync[F]): F[Set[CliqueOracle.Justification[BlockHash, Validator]]] =
+    dag
+      .lookupUnsafe(blockHash)
+      .map(
+        _.justifications.map {
+          case Justification(validator, latestBlockHash) =>
+            CliqueOracle.Justification(latestBlockHash, validator)
+        }.toSet
+      )
+
+  def toCasperJustificationUnsafe(
+      blockHash: BlockHash
+  )(implicit sync: Sync[F]): F[CliqueOracle.Justification[BlockHash, Validator]] =
+    dag.lookupUnsafe(blockHash).map(m => CliqueOracle.Justification(m.blockHash, m.sender))
+}
--- a/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
+++ b/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
@@ -1,0 +1,180 @@
+package coop.rchain.casper.safety
+
+import cats.Show
+import cats.data.OptionT
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import coop.rchain.casper.util.{Clique, ProtoUtil}
+import coop.rchain.dag.{Casper, DagOps, DagReader}
+import coop.rchain.metrics.Metrics.Source
+import coop.rchain.metrics.{Metrics, Span}
+import coop.rchain.shared.Log
+import coop.rchain.shared.syntax._
+import coop.rchain.metrics.implicits._
+import fs2.Stream
+
+import scala.reflect.ClassTag
+
+object CliqueOracle {
+  // map of validator id to stake bonded
+  type WeightMap[V] = Map[V, Long]
+
+  implicit val baseMetricsSource: Source = coop.rchain.casper.CasperMetricsSource
+
+  private def computeMaxCliqueWeight[F[_]: Concurrent: Span, M, V: ClassTag](
+      target: M,
+      dag: DagReader[F, M],
+      agreeingWeightMap: WeightMap[V],
+      agreeingLatestMessages: Map[V, M],
+      toJustificationF: M => F[Casper.Justification[M, V]],
+      justificationsF: M => F[Set[Casper.Justification[M, V]]],
+      heightF: M => F[Long]
+  )(implicit show: Show[M]): F[Long] = Span[F].traceI("compute-max-clique-weight") {
+
+    /**
+      * Prerequisite for this is that latest messages from a and b both are in main chain with target message
+      *
+      *     a    b
+      *     *    *  <- lmB
+      *      \   *
+      *       \  *
+      *        \ *
+      *         \*
+      *          *  <- lmAjB
+      *
+      * 1. get justification of validator b as per latest message of a (lmAjB)
+      * 2. check if any self justifications between latest message of b (lmB) and lmAjB are NOT in main chain
+      *    with target message. If one found - this is a source of disagreement.
+      * */
+    def neverEventuallySeeDisagreement(a: V, b: V): F[Boolean] = {
+      val noDisagreement = for {
+        // latest messages of validators matched
+        lmA <- OptionT.fromOption(agreeingLatestMessages.get(a))
+        lmB <- OptionT.fromOption(agreeingLatestMessages.get(b))
+        // message of `b` used as a justification by latest message of `a`
+        lmAjB <- OptionT(justificationsF(lmA).map(_.find(j => j.creator == b)))
+
+        disagreements = for {
+          // self justification of lmAjB or lmAjB itself. Used as a stopper for traversal
+          // TODO not completely clear why try to use self justification and not just message itself
+          stopper <- justificationsF(lmAjB.message).map(
+                      _.find(_.creator == lmAjB.creator)
+                        .getOrElse(lmAjB)
+                        .message
+                    )
+          selfJustificationsChain = DagOps
+            .bfTraverseF[F, M](List(lmB)) { jFromB =>
+              ProtoUtil
+                .getCreatorJustification(
+                  jFromB,
+                  (b: M) => b == stopper,
+                  toJustificationF,
+                  justificationsF
+                )
+                .map(_.map(List(_)).getOrElse(List.empty))
+            }
+          // find if any of traversed selfJustifications is NOT in main chain with target
+          r <- selfJustificationsChain
+                .filterF(dag.isInMainChain(_, target, heightF).not)
+                .flatMap(_.toList)
+        } yield r
+
+        r <- OptionT.liftF(disagreements.map(_.isEmpty))
+
+      } yield r
+
+      noDisagreement.getOrElse(false)
+    }
+
+    /** across combination of validators compute pairs that do not have disagreement */
+    def computeAgreeingValidatorPairs: F[List[(V, V)]] = {
+      val pairs = agreeingWeightMap.keys.toArray.combinations(2).map { case Array(a, b) => (a, b) }
+
+      Stream
+        .fromIterator(pairs)
+        .evalFilterAsyncProcBounded {
+          case (a, b) =>
+            neverEventuallySeeDisagreement(a, b) &&^ neverEventuallySeeDisagreement(b, a)
+        }
+        .compile
+        .toList
+    }
+
+    computeAgreeingValidatorPairs.map { edges =>
+      Clique.findMaximumCliqueByWeight[V](edges, agreeingWeightMap)
+    }
+  }
+
+  def normalizedFaultTolerance[F[_]: Concurrent: Log: Metrics: Span, M, V: ClassTag](
+      target: M,
+      dag: DagReader[F, M],
+      getWeightMapF: M => F[WeightMap[V]],
+      heightF: M => F[Long],
+      latestMessageF: V => F[M],
+      toJustificationF: M => F[Casper.Justification[M, V]],
+      getJustificationsF: M => F[Set[Casper.Justification[M, V]]]
+  )(implicit show: Show[M]): F[Float] = {
+    val nonExistingMessage =
+      Log[F].error(s"Fault tolerance for non existing message ${target.show} requested.").as(-1f)
+
+    val compute = for {
+      // weight map of main parent (fallbacks to message itself if no parents)
+      fullWeightMap <- {
+        dag
+          .mainParent(target)
+          .map {
+            case Some(mainParent) => mainParent
+            case None             => target
+          }
+      }.flatMap(getWeightMapF)
+      // stake that agrees on a target message
+      agreeingWeightMap <- fs2.Stream
+                            .emits(fullWeightMap.toList)
+                            .evalFilterAsyncProcBounded {
+                              case (validator, _) =>
+                                for {
+                                  lm <- latestMessageF(validator)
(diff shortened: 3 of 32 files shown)


models/src/test/scala/coop/rchain/models/blockImplicits.scala
```diff
@@ -35,7 +35,7 @@ object blockImplicits {
   val bondGen: Gen[Bond] = for {
     byteArray <- listOfN(Validator.Length, arbitrary[Byte])
     validator = ByteString.copyFrom(byteArray.toArray)
-    stake     <- arbitrary[Long]
+    stake     <- Gen.chooseNum(1L, 1024L)
```
zsluedem, line 38: Question: Are you changing this because of negative numbers?

models/src/main/scala/coop/rchain/models/block/StateHash.scala
```diff
@@ -10,4 +11,12 @@ object StateHash {
   implicit class StateHashOps(bs: StateHash) {
     def base16String: String = Base16.encode(bs.toByteArray)
   }
+
+  implicit def ordering: Ordering[StateHash] =
+    Ordering.by((b: ByteString) => b.toByteArray.toIterable)
```
zsluedem: Why use `toIterable` here? It is kind of reduntant here.

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,68 @@
...
+  ): F[A] = {
+    def source = s"${file.value}:${line.value} ${enclosing.value}"
+    def errMsg = s"Unsafe call for parents for block with no parents: ${item.show}\n at $source"
+    mainParent(item) >>= (_.liftTo(DagReader.VertexDoesNotExist(errMsg)))
+  }
+
+  def isInMainChain(
+      descendant: A,
+      ancestor: A,
+      height: A => F[Long]
```
zsluedem, line 43: Help me to understand these arguments.

`descendant` is the target which the function want to check if the target is in the `MainChain`.
`ancestor` is the marking vertex which means the start poin of the `MainChain`.

Is that correct?

zsluedem: For a dag like this
```
     b3
/         \
b1     b2
  \    /
   b0
```
blockHeightFunc for each blocks ->  `b0=1` , `b1=2`, `b2=2`, `b3=3`.

`isInMainChain(descendant=b1, ancestor=b3, height=blockHeightFunc)` what's the result of the function?

models/src/test/scala/coop/rchain/models/blockImplicits.scala
nzpr: Yes, bonds should not be negative. Also if max is `Long.max` - this might cause overflow when summing two stakes, thats why I set `1024L` here

models/src/main/scala/coop/rchain/models/block/StateHash.scala
nzpr: I copied this ordering from Estimator, not really sure why `toIterator` is used - maybe it has some performance benefits

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
nzpr: The result will be false, as b1 is ancestor of b3, so function return false on first iteration of tail recursion
https://github.com/rchain/rchain/blob/fe40fe34335003babb42688618a6436f1d735d8e/shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala#L74-L77


> Help me to understand these arguments.
> 
> `descendant` is the target which the function want to check if the target is in the `MainChain`.
> `ancestor` is the marking vertex which means the start poin of the `MainChain`.
> 
> Is that correct?

its other way. If you want to check if something is in the main chain - you start with some descendant (e.g. tip), as you want to traverse through main parents. There is no main child, you can traverse only upside down.
So `target` which we are going to evaluate agains `isInMainChain` predicate in caller function is **ancestor** here.

models/src/main/scala/coop/rchain/models/block/StateHash.scala
zsluedem: `toIterator` is differenct `toIterable`. Here `toIterable` just return instance itself so we don't need that.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -48,6 +51,16 @@ final class BlockDagRepresentationOps[F[_]: Sync](
     fs2.Stream.emits(streams).parJoinUnbounded.compile.toList
   }
 
+  def latestMessageHashUnsafe(v: Validator)(
+      implicit line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing
```
tgrospic, line 55: We don't need source information anymore, we have error stack trace with monix 3.0.

shared/src/main/scala/coop/rchain/dag/Casper.scala
```diff
@@ -0,0 +1,5 @@
+package coop.rchain.dag
+
+object Casper {
+  final case class Justification[M, V](message: M, creator: V)
```
tgrospic: This class is used by clique oracle and should be defined there.

Also `toCasperJustificationUnsafe` and `getCasperJustificationsUnsafe` extensions can be defined as private functions on SafetyOracle because they are used only there.

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,68 @@
...
+
+final class DagReaderOps[F[_], A](
+    private val dag: DagReader[F, A]
+) extends AnyVal {
+
+  def parentsUnsafe(item: A)(
+      implicit sync: Sync[F],
+      line: sourcecode.Line,
+      file: sourcecode.File,
+      enclosing: sourcecode.Enclosing,
```
tgrospic: Source info is not needed anymore.

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,68 @@
...
+        else
+          for {
+            dHeight <- height(d)
+            aHeight <- height(ancestor)
+            r <- if (dHeight <= aHeight)
+                  false.asRight[A].pure[F]
+                else
+                  mainParentUnsafe(d).map(_.asLeft[Boolean])
+          } yield r
+    )
```
tgrospic: With multi line function body `{}` block is usually shorter because parameter stays in the same line and the body is indented only one column.

models/src/main/scala/coop/rchain/models/block/StateHash.scala
tgrospic: This should be fixed in all places as Will described.

models/src/main/scala/coop/rchain/models/block/StateHash.scala
```diff
@@ -10,4 +11,12 @@ object StateHash {
...
+
+  implicit val show = new Show[StateHash] {
+    def show(validator: StateHash): String =
+      Base16.encode(validator.toByteArray).take(10)
+  }
```
tgrospic: But also all of these definitions (StateHashOps, ordering, show) can be removed. There is no sense to define them separately for aliases. Definitions for BlockHash alias works for both, so better to not use aliases in these situations.

models/src/main/scala/coop/rchain/models/Validator.scala
```diff
@@ -1,9 +1,19 @@
...
 
   val Length = 65
+
+  implicit def ordering: Ordering[Validator] =
+    Ordering.by((b: Validator) => b.toByteArray.toIterable)
+
+  implicit val show = new Show[Validator] {
+    def show(validator: Validator): String =
+      Base16.encode(validator.toByteArray).take(10)
+  }
```
tgrospic: See StateHash comment. These are duplicated definitions for ByteString ordering and show.

models/src/main/scala/coop/rchain/models/BlockHash.scala
```diff
@@ -10,4 +11,12 @@ object BlockHash {
...
     def base16String: String = Base16.encode(bs.toByteArray)
   }
+
+  implicit def ordering: Ordering[BlockHash] =
+    Ordering.by((b: ByteString) => b.toByteArray.toIterable)
+
+  implicit val show = new Show[BlockHash] {
+    def show(blockHash: BlockHash): String =
+      Base16.encode(blockHash.toByteArray).take(10)
+  }
```
tgrospic: Define all three implicits for ByteString directly without alias, it will work for all defined aliases.

Also definition of _show_ is usually in representation that can be parsed and initial value can be constructed. So better to define it as an extension method for short print.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -59,10 +59,11 @@ class BlockQueryResponseAPITest
       setValidator = sender.some,
       setDeploys = randomDeploys.some,
       setJustifications = List(Justification(bondsValidator.validator, genesisBlock.blockHash)).some,
+      setParentsHashList = List(genesisBlock.blockHash).some,
       setBonds = List(bondsValidator).some
     )
 
-  val faultTolerance = 1f
+  val faultTolerance = -1f
```
tgrospic: Should this be named `faultToleranceMin`? Or definition from `MIN_FAULT_TOLERANCE` from SafetyOracle can be used.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -94,25 +81,43 @@ object ProtoUtil {
     * requires a list to be returned. When we reach the goalFunc,
     * we return an empty list.
     */
-  def getCreatorJustificationAsListUntilGoalInMemory[F[_]: Monad](
+  def getCreatorJustificationAsListUntilGoalInMemory[F[_]: Sync](
```
tgrospic, line 84: BlockDagRepresentation can have syntax defined in casper project also with this function added.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -94,25 +81,43 @@ object ProtoUtil {
...
+        .map(_.toSet)
+    val toJ = (h: BlockHash) =>
+      blockDag
+        .lookupUnsafe(h)
+        .map(m => Casper.Justification(m.blockHash, m.sender))
+
+    for {
+      r <- getCreatorJustification[F, BlockHash, Validator](blockHash, goalFunc, toJ, getJ)
+            .map(_.map(List(_)).getOrElse(List.empty))
+    } yield r
```
tgrospic: `for` comprehension is not needed.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -94,25 +81,43 @@ object ProtoUtil {
...
+
+  }
+
+  def getCreatorJustification[F[_]: Sync, A, V](
```
tgrospic, line 104: Because `Justification` class is part of clique oracle, this function belongs there.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -100,4 +113,19 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+      _.justifications.map {
+        case Justification(validator, latestBlockHash) =>
+          Casper.Justification(latestBlockHash, validator)
+      }.toSet
+    )
+
+  def toCasperJustificationUnsafe(
+      blockHash: BlockHash
+  ): F[Casper.Justification[BlockHash, Validator]] =
+    lookupUnsafe(blockHash).map(m => Casper.Justification(m.blockHash, m.sender))
```
tgrospic: `getCasperJustificationsUnsafe` and `toCasperJustificationUnsafe` should be moved to BlockDagRepresentationSyntax in casper project.

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,185 @@
...
+        dag
+          .mainParent(target)
+          .map {
+            case Some(mainParent) => mainParent
+            case None             => target
+          }
+      }.flatMap(getWeightMapF)
+      // stake that agrees on a target message
+      agreeingWeightMap <- fs2.Stream
+                            .emits(fullWeightMap.toList)
```
tgrospic: With iterator conversion to stream is more lazy  `Stream.fromIterator(fullWeightMap.iterator)`.

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,185 @@
...
+        r <- OptionT.liftF(disagreements.map(_.isEmpty))
+
+      } yield r
+
+      noDisagreement.getOrElse(false)
+    }
+
+    /** across combination of validators compute pairs that do not have disagreement */
+    def computeAgreeingValidatorPairs: F[List[(V, V)]] = {
+      val pairs = agreeingWeightMap.keys.toArray.combinations(2).map { case Array(a, b) => (a, b) }
```
tgrospic: This can be done lazy with Seq, without creating intermediate Array.
```scala
val pairs = agreeingWeightMap.keys.toSeq.combinations(2).map { case Seq(a, b) => (a, b) }
```

models/src/main/scala/coop/rchain/models/block/StateHash.scala
nzpr: just `b.toByteArray` won't work because there is not implicit ordering by `Array[Byte]`. Thats why it is converted to iterable

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
nzpr: I'd better do that in another PR as the whole ProtoUtil object require refactoring

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
nzpr: I don't think so, it a fault tolerance that is expected in tests. It just should be minimal

tgrospic: My thoughts was more on the usage side where we can be more clear what is expected.
```scala
b.faultTolerance should be(minFaultTolerance)
// Instead of
b.faultTolerance should be(faultTolerance)
```

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,214 @@
...
+      target: M,
+      dag: DagReader[F, M],
+      getWeightMapF: M => F[WeightMap[V]]
+  ): F[WeightMap[V]] =
+    dag
+      .mainParent(target)
+      .map {
+        case Some(mainParent) => mainParent
+        case None             => target
+      }
```
tgrospic: This can be one liner `.map(_.getOrElse(target))`.

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,59 @@
...
+  ): F[Seq[A]] = {
+    def errMsg = s"Parents lookup failed: DAG is missing ${item.show}"
+    dag.parents(item) >>= (_.liftTo(DagReader.VertexDoesNotExist(errMsg)))
+  }
+
+  def contains(item: A)(implicit sync: Sync[F]): F[Boolean] =
+    dag.parents(item).map(_.isDefined) ||^ dag.children(item).map(_.isDefined)
+
+  def mainParent(item: A)(implicit sync: Sync[F]): F[Option[A]] =
+    dag.parents(item).map(_.flatMap(_.headOption))
```
tgrospic, line 30: These two methods only need Monad instead of Sync.

casper/src/main/scala/coop/rchain/casper/BlockDagRepresentationSyntax.scala
```diff
@@ -0,0 +1,37 @@
...
+        _.justifications.map {
+          case Justification(validator, latestBlockHash) =>
+            CliqueOracle.Justification(latestBlockHash, validator)
+        }.toSet
+      )
+
+  def toCasperJustificationUnsafe(
+      blockHash: BlockHash
+  )(implicit sync: Sync[F]): F[CliqueOracle.Justification[BlockHash, Validator]] =
+    dag.lookupUnsafe(blockHash).map(m => CliqueOracle.Justification(m.blockHash, m.sender))
```
tgrospic, line 36: These methods are not really extensions of `BlockDagRepresentation` but more of `BlockMetaData` type.

It's more visible when these methods are used as they are additional fields.
```scala
h => dag.lookupUnsafe(h).map(_.weightMap),
h => dag.lookupUnsafe(h).map(_.blockNum),
h => dag.lookupUnsafe(h).map(_.toCliqueJustification),
h => dag.lookupUnsafe(h).map(_.getCliqueJustifications)
```
Instead directly on BlockDagRepresentation.
https://github.com/rchain/rchain/blob/afdeee0b892c608696d4849cd5827968b452f186/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L167-L170

It's tempting to put many methods on type like `BlockDagRepresentation` but this can easily become very large file.
But it can be reasonable to have these methods here if used on many places but definition would be composed of call to BlockMetaData extension.
```scala
def getCliqueJustificationsUnsafe(
      blockHash: BlockHash
  )(implicit sync: Sync[F]): F[Set[CliqueOracle.Justification[BlockHash, Validator]]] =
    dag.lookupUnsafe(blockHash).map(_.getCliqueJustifications)
```

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,214 @@
...
+      latestMessageF(validator).flatMap(dag.isInMainChain(_, message, heightF))
+
+    def agreeingWeightMapF(weightMap: WeightMap[V]): F[WeightMap[V]] =
+      fs2.Stream
+        .fromIterator(weightMap.toIterator)
+        .evalFilterAsyncProcBounded {
+          case (validator, _) => agree(validator, target)
+        }
+        .compile
+        .to(Map)
```
tgrospic, line 173: Nice conversion to Map!

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -31,26 +31,13 @@ object ProtoUtil {
...
-            false.pure[F]
-          } else {
-            val mainParentOpt = targetBlock.parents.headOption
-            mainParentOpt.traverse(isInMainChain(dag, candidateMetadata, _)).map(_.getOrElse(false))
-          }
-        case None =>
-          false.pure[F]
-      }
+    val heightF: BlockHash => F[Long] = h => dag.lookupUnsafe(h).map(_.blockNum)
+    dag.isInMainChain(targetBlockHash, candidateMetadata.blockHash, heightF)
```
tgrospic, line 40: Only block hash is used from `BlockMetadata` argument.

This method can be moved to casper `BlockDagRepresentationSyntax`.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
tgrospic: You created already syntax for BlockDagRepresentation in this PR. :wink:

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  private def computeMaxCliqueWeight[F[_]: Concurrent: Span, M, V](
+      target: M,
+      dag: DagReader[F, M],
+      agreeingWeightMap: WeightMap[V],
+      agreeingLatestMessages: Map[V, M],
+      toJustificationF: M => F[Justification[M, V]],
+      justificationsF: M => F[Set[Justification[M, V]]],
+      heightF: M => F[Long]
+  )(implicit show: Show[M]): F[Long] = Span[F].traceI("compute-max-clique-weight") {
```
tgrospic: `Show` can be defined as context bound parameter instead of "normal" implicit parameter similar to `F`.
```scala
def computeMaxCliqueWeight[F[_]: Concurrent: Span, M: Show, V]
```

shared/src/main/scala/coop/rchain/dag/DagReaderSyntax.scala
```diff
@@ -0,0 +1,59 @@
...
+      descendant: A,
+      ancestor: A,
+      height: A => F[Long]
+  )(implicit sync: Sync[F], show: Show[A]): F[Boolean] =
+    descendant.tailRecM { d =>
+      if (d == ancestor) true.asRight[A].pure[F]
+      else
+        for {
+          dHeight <- height(d)
+          aHeight <- height(ancestor)
```
tgrospic: Height of the can be checked only one on the beginning.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
nzpr: I mean not touch ProtoUtils, as lots of that stuff there belong to dag syntax. I'd like to keep history of changing these functions before moving to another file

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
nzpr: Indeed, I did that to make more explicit that logic, as it is sort of weird and very important. But ok, I'll simplify that.

casper/src/main/scala/coop/rchain/casper/BlockDagRepresentationSyntax.scala
nzpr: This Justification is actually the same justification as in Casper protocol messages - I just created abstract version of it to be able to use in CliqueOracle. There is no special CliqueJustification. Justification is message + message creator. Thats why I created it in shared project under Casper package first.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
nzpr: I think here it is just testing that API returns valid response for block query. Whether returned fault tolerance is minimal or some else - out of scope. It just happen that given how test is defined, fault tolerance returned is -1.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
nzpr: isInMainChain is in DagReaderSyntax, but it requires heightF, so this function is a proxy that extracts height from BlockDagRepresentation.

casper/src/main/scala/coop/rchain/casper/BlockDagRepresentationSyntax.scala
tgrospic: Shared project should not be storage for all types, this `Justification` class does not have meaning by itself and it has no operation defined with it. Tuple is defined already with two abstract arguments `Tuple2[A, B]` so this class is basically to provide names for each field.

My point here is that conversion from protobuf Justification to new Justification, which CliqueOracle requires, is not available to other code without having instance of BlockDagRepresentation.
I'm ok to leave this here and cleanup later.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
tgrospic: What makes this confusing is that DagReader is also a BlockDagRepresentation but it receives function to extract height which is available to this instance.

My point here is that it requires the whole BlockMetadata just to read the hash. But this can be cleanup later also.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -94,25 +81,40 @@ object ProtoUtil {
...
+    import coop.rchain.blockstorage.syntax._
+    val getJ = (h: BlockHash) =>
+      blockDag
+        .lookupUnsafe(h)
+        .map(_.justifications.map { case Justification(m, v) => CliqueOracle.Justification(m, v) })
+        .map(_.toSet)
+    val toJ = (h: BlockHash) =>
+      blockDag
+        .lookupUnsafe(h)
+        .map(m => CliqueOracle.Justification(m.blockHash, m.sender))
```
tgrospic, line 98: If `getCasperJustificationsUnsafe` and `toCasperJustificationUnsafe` will stay as BlockDagRepresentation extensions, they can be used here.

nzpr: I agree that this belongs to BlockMetadata syntax as you've pointed in https://github.com/rchain/rchain/pull/3387#discussion_r641605484

nzpr:
Let me mention that we might have some future work on safety oracle here <https://github.com/rchain/rchain/issues/3058>.

nzpr:
closed in favour of https://github.com/rchain/rchain/pull/3421

PR #3388, opened 2021-04-25
New finalizer with multi parent support

### LINKED ISSUE #3350: Modify the DAG representation to support block merge

nzpr:
This is replacement for https://github.com/rchain/rchain/pull/3366 as it has been closed by bors. 

Old finalizer is not suitable for multi parent as it is possible that last finalized block is stuck. This happen due to the fact that only children that have current LFB as a main parent are subject to be next LFB. 

Finalizer introduced in this PR seeks for LFB **only using tips and DAG** and input data, without involving requirement to know previous LFB. Due to this change, algorithm for finalization is rewritten significantly, keeping the main logic of following CliqueOracle output do decide if some message is finalized .


nzpr:
Only the last commit is for review, other are pulled from https://github.com/rchain/rchain/pull/3387

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,216 @@
...
+  *   2. Execute [[CliqueOracle.computeOutput]] on these targets.
+  *   3. Candidate with the highest block number passing FT becomes the next LFB.
+  */
+object Finalizer {
+
+  /** Message that is agreed on + weight of this agreement. */
+  final case class Agreement[M, V](message: M, stakeAgreed: WeightMap[V])
+
+  /** If more then half of total stake agree on message - it is considered to be safe from orphaning. */
+  def cannotBeOrphaned[F[_]: Applicative, M, V](
```
tgrospic: `F` is not used in this function.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,216 @@
...
+    * If validator is not present in message bonds map or its stake is zero, None is returned
+    */
+  def recordAgreement[V](
+      messageWeightMap: WeightMap[V],
+      agreeingValidator: V
+  ): Option[WeightMap[V]] = {
+    // if validator is not bonded according to message weight map - there is no agreement translated.
+    val stakeAgreed = messageWeightMap.getOrElse(agreeingValidator, 0L)
+    if (stakeAgreed > 0) Map(agreeingValidator -> stakeAgreed).some else None
+  }
```
tgrospic, line 85: Why to return a Map when only one record is returned?

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,216 @@
...
+    * @return Stream of agreements propagated by message onto its ancestors down till stopHeight
+    */
+  def messageAgreements[F[_]: Sync, M, V](
+      message: M,
+      validator: V,
+      dag: DagReader[F, M],
+      stopHeight: Long,
+      heightF: M => F[Long],
+      messageWeightMapF: M => F[WeightMap[V]]
+  ): Stream[F, Agreement[M, V]] =
```
nzpr: Output of this can be cached, its defined by message and stopHeight (which is LFB that should change in a discrete way)

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -122,20 +126,44 @@ class MultiParentCasperImpl[F[_]: Sync: Concurrent: Log: Time: SafetyOracle: Las
...
+        block          <- BlockStore[F].getUnsafe(finalizedChildHash)
+        deploys        = block.body.deploys.map(_.deploy)
+        deploysRemoved <- DeployStorage[F].remove(deploys)
+        _ <- Log[F].info(
+              s"Removed $deploysRemoved deploys from deploy history as we finalized block ${PrettyPrinter
+                .buildString(finalizedChildHash)}."
             )
-            .whenA(lastFinalizedBlockHash != updatedLastFinalizedBlockHash)
-      blockMessage <- BlockStore[F].getUnsafe(updatedLastFinalizedBlockHash)
+        _ <- BlockDagStorage[F].addFinalizedBlockHash(finalizedChildHash)
```
tgrospic, line 140: Where is indirectly finalized block persisted? This is just in memory update.

nzpr: Oh, I did not realise it is not persisted.
Indeed, looks like finalized set is recreated on each startup as ancestors of last finalized block
https://github.com/rchain/rchain/blob/e5cafa4e184b72d243e925f8bc6a523624ba6fe2/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala#L148-L155 But why this is not a problem now when you restart having big DAG?  
And this is out of this PR scope, am I correct?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -159,4 +159,23 @@ final class BlockDagRepresentationOps[F[_]](
...
+        parents.map(p => p.nonEmpty.guard[Option].as(p, p))
+      }
+      .flatMap(Stream.emits)
+      .compile
+      .to(Set)
+
+  def withAncestors(blockHash: BlockHash, filterF: BlockHash => F[Boolean])(
+      implicit sync: Sync[F]
+  ): F[Set[BlockHash]] =
+    ancestors(blockHash, filterF).map(_ + blockHash)
```
tgrospic, line 180: If `ancestors` return Stream then filter can be be used on that stream directly without predicate function as a parameter. This is the point of Stream as being lazy and pull based so consumer can decide when to stop pulling.

```scala
def ancestors2(blockHash: BlockHash)(
    implicit sync: Sync[F]
): Stream[F, BlockHash] =
  Stream
    .unfoldEval(List(blockHash)) { lvl =>
      val parents = lvl
        .traverse(lookupUnsafe)
        .map(_.flatMap(_.parents).distinct)
      parents.map(p => p.nonEmpty.guard[Option].as(p, p))
    }
    .flatMap(Stream.emits)

def withAncestors(blockHash: BlockHash, filterF: BlockHash => F[Boolean])(
    implicit sync: Sync[F]
): F[Set[BlockHash]] =
  ancestors2(blockHash).evalFilter(filterF).compile.to(Set)
```

nzpr: I made it return Set because Stream might contain duplicate elements, and this is not something caller is aware of. Event if you make `_.flatMap(_.parents).distinct` deduplicating one level - the same message might be visited again via some another branch (after several intermediate parents).

tgrospic: Deduplication can be done on a Stream also, but this is not the issue here. What I mean is that caller can call `.compile.to(Set)` to get unique elements. There is not need to invent the same functionality that already exists on Stream like another way of filtering which means `withAncestors` is just code duplication.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: OK, so this means indirectly finalized blocks will also have finalized state after restart?

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,192 @@
...
+        // first candidate that meets finalization criteria is new LFB
+        .filter { case (_, faultTolerance) => faultTolerance > faultToleranceThreshold }
+        .head
+        // execute all effects
+        .evalMap {
+          case (lfb, _) =>
+            newLfbEffect(lfb.blockHash).as(lfb.blockHash) <*
+              dag
+                .withAncestors(lfb.blockHash, dag.isFinalized(_).not)
+                .flatMap(_.toList.traverse(finalisationEffect))
```
tgrospic: Because we don't have atomic operation on LMDB, here is possible for node to be killed between saving directly finalized block `newLfbEffect` and related operations in `finalisationEffect`.
Is it better to save LFB at the end? Can this be more safe so that `finalisationEffect` can be executed again?

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
```diff
@@ -0,0 +1,192 @@
...
+    require(
+      !agreeingWeightMap.exists { case (_, stake) => stake <= 0 },
+      message = "Agreeing map contains not bonded validators"
+    )
+    val activeStakeTotal    = messageWeightMap.values.sum
+    val activeStakeAgreeing = agreeingWeightMap.values.sum
+    // in theory if each stake is high enough, e.g. Long.MaxValue, sum of them might result in negative value
+    require(activeStakeTotal > 0, message = "Long overflow when computing total stake")
+    require(activeStakeAgreeing > 0, message = "Long overflow when computing total stake")
+    activeStakeAgreeing > activeStakeTotal.toFloat / 2
```
tgrospic, line 71: How is this calculation here in finalizer? I don't see that was in previous calculation. This should be information from safety oracle.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
nzpr: `What I mean is that caller can call .compile.to(Set) to get unique elements.` I understand that, I'm saying that if I would see 
```
def ancestors2(blockHash: BlockHash)(
    implicit sync: Sync[F]
): Stream[F, BlockHash] 
```
I would assume that I can just call `compile.toList` and thats it, I would not expect this list contain duplicate elements. I made it return Set to be specific, that this will explicitly return unique ancestors. So function user not introduce bug assuming stream does not contain duplicates.

How do you deduplicate stream? In current bfTraverse function there is internal Set keeping the state. But why then not just return Set?

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
nzpr: This is because previous finalizer were just running clique oracle on all LFB children. This particular logic were not on that version. New version traverses down from tips until it finds messages worth to be computed faultTolerance agains. If you see clique oracle code - if less than half of a stake is agreed on message, max clique is not computed for this message, and minimal fault tolerance is returned straight away. So this message for sure cannot be finalized.

Do you mean this function better to be moved inside CliqueOracle?

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
nzpr: Makes sense, indeed, otherwise if node is killed `finalisationEffect ` might be never called for some messages that are ancestor of LFB. I'l push fix in commit addressing comments

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
tgrospic: Deduplication must involve some state to keep track of previous elements. Here is some discussion.
https://gitter.im/functional-streams-for-scala/fs2?at=58c300d41465c46a56152134
With the link to this gist.
https://gist.github.com/zeryx/5d918a433dca167975ca047a7e1f4cc1
We can create little modified version with Ref which stores Set of previous hashes and only produce unique ones, similar to what you described.

The main problem is that now `ancestors` cannot be used lazily just because includes materialization of Set at the end and because of that also needs filter as a parameter.

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
tgrospic: My concern is if we are sure here that agreeing validators cannot have equivocations.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
nzpr: Yes, as last finalized block is persisted

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
nzpr: Ok, I will make it return stream with comment about duplicates

casper/src/main/scala/coop/rchain/casper/finality/Finalizer.scala
nzpr: I'm not sure I understand this. There is always only one main parent, right? So agreeing chains are well defined. We need to get to the point where enough of these chains join, and find first message that pass fault toleranceThreshold.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
tgrospic: From discussion on standup, we concluded that with filter directly on a stream it's difficult to take only non finalized blocks because some parents may not be included. So better to leave as is.

tgrospic (approved):
LGTM

PR #3392, opened 2021-04-29
[WIP] Rejection testing

tgrospic:


casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+      )
+      _ = println(s"RESULT:\n${res.mkString("\n")}")
+    } yield ()
+  }
+
+  it should "reject deploys built on top of rejected deploy (separate blocks)" in effectTest {
+    TestNode.networkEff(genesis, networkSize = 3).use { nodes =>
+      testCase(nodes) { deploys =>
+        // Set and read in separate blocks
+        deploys.traverse(nodes(2).addBlock(_))
```
tgrospic, line 134: When deploys in second branch (node 2) are in two consequent blocks, _read_ deploy in not rejected, although _set_ in previous block is rejected.
```
Deploys          -> Set(N1_nila, N1_set, N1_nilb, N2_set, N2_read)
Rejected deploys -> List(false, false, false, true, false)
```

casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+      }
+    }
+  }
+
+  it should "reject deploys built on top of rejected deploy (same block)" in effectTest {
+    TestNode.networkEff(genesis, networkSize = 3).use { nodes =>
+      testCase(nodes) { deploys =>
+        // Set and read in the same block
+        nodes(2).addBlock(deploys: _*).map(List(_))
```
tgrospic, line 143: When deploys are in the same block, then _set_ and _read_ are rejected.
```
Deploys          -> Set(N1_nila, N1_set, N1_nilb, N2_set, N2_read)
Rejected deploys -> List(false, false, false, true, true)
```

casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
nzpr: So you have this DAG
  ```
/**
    *          *
    *         / \
    *        |   *
    *        |   |
    * set -> *   * <- set & read
    *         \ /
    *          *
    *        1 0 2
    */ 
``` 
Here right branch will be selected as the main one, so `set` from left branch should be rejected. This is what you see, right?

File: casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala (hunk 2 above)
Comment (nzpr) [reply casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala:143]: Here is the DAG
  ```
/**
    *          *
    *         / \
    * set -> *   * <- set & read
    *         \ /
    *          *
    *        1 0 2
    */ 
``` 
Here it depends on ordering of block hashes which one picked as a main state, so either `set` from left is rejected - or `set` and `read` from right.

nzpr:
@tgrospic Now dependency rejection is not implemented, I guess when you see two vs one rejection - it is because of which main branch is selected. But it is not because of dependency rejection.
But I has been thinking that it can done by modifying BranchIndex class to be this  
```scala
final case class BranchIndex(
    chanToDeployMap: Map[Blake2b256Hash, Set[ProcessedDeploy]],
    deployToChanMap: Map[ProcessedDeploy, Set[Blake2b256Hash]],
    eventLogIndex: EventLogIndex,
    deploys: Set[ProcessedDeploy]
) {
  @tailrec
  def go(
      channelsToRemove: Set[Blake2b256Hash],
      acc: Set[ProcessedDeploy]
  ): Set[ProcessedDeploy] =
    if (channelsToRemove.isEmpty) {
      acc
    } else {
      val toReject = channelsToRemove.flatMap(
        c => chanToDeployMap.getOrElse(c, Set.empty[ProcessedDeploy])
      ) -- acc
      val nextChannels = toReject.flatMap(
        d => deployToChanMap.getOrElse(d, Set.empty[Blake2b256Hash])
      )
      go(nextChannels, acc ++ toReject)
    }

  def calculateRejectedDeploys(conflicts: Set[Blake2b256Hash]): Set[ProcessedDeploy] =
    go(conflicts, Set.empty[ProcessedDeploy])
}
```

casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
tgrospic: @nzpr _Set_ from right branch is rejected, this means left one is the main but _read_ in subsequent block in not rejected although it should because after merging reads non existing data.
```
/**
    *          *
    *         / \
    *        |   * <- read
    *        |   |
    * set -> *   * <- set (rejected)
    *         \ /
    *          *
    *        1 0 2
    */ 
```

nzpr: I see, I missed that node1 creates 3 blocks
https://github.com/rchain/rchain/blob/59bfc7a58249df19355bce43fb1eaad5d871de29/casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala#L84-L89

casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
tgrospic: @nzpr If dependency rejection is not implemented why is in this case _read_ deploy (N2_read) rejected?

nzpr: @tgrospic when channel is conflicting - all deploys touching this channel on are rejected. To test dependency you need to have some non conflicting channel which two or more deploys share. So it’s not a dependency that is tested here - it’s just rejection. I think I found a bug that prevents read deploy from rejecting in multi block settings.

tgrospic:
This is test for the first version of block-merge, it's not important anymore.

PR #3393, opened 2021-04-29
Re-enable shutdown hook

nzpr:
In https://github.com/rchain/rchain/commit/deff62b18fa82bb85079fee28389c3179f0ee81e when splitting Runtime into smaller pieces shutdown hook has been disabled. This PR enables it back.
What should be cleaned up is 1) store manager, 2) deploy storage, as it is still outside of store manager 3) Kamon reporters. 4) API servers 5) block store

### DIFF
--- a/node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
+++ b/node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
@@ -37,7 +37,7 @@ import coop.rchain.shared._
 import coop.rchain.shared.syntax._
 import coop.rchain.store.KeyValueStoreManager
 import coop.rchain.store.LmdbDirStoreManager.gb
-import fs2.concurrent.Queue
+import fs2.concurrent.{Queue, SignallingRef}
 import kamon._
 import monix.execution.Scheduler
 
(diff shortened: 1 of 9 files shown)


node/src/main/scala/coop/rchain/node/runtime/NodeRuntime.scala
```diff
@@ -405,29 +404,6 @@ class NodeRuntime[F[_]: Monixable: ConcurrentEffect: Parallel: Timer: ContextShi
...
-  def clearResources(
-      servers: ServersInstances[F],
-      runtimeCleanup: Cleanup[F],
-      blockStore: BlockStore[F]
-  ): Unit = {
-    val shutdown = for {
-      _ <- Sync[F].delay(Kamon.stopAllReporters())
-      _ <- runtimeCleanup.close()
-      _ <- Log[F].info("Bringing BlockStore down ...")
-      _ <- blockStore.close()
```
nzpr, line 424: blockstore is now managed by node store manager, so no need to close it separately

tgrospic: We didn't migrated old full nodes to LMDB based block store.

tgrospic: Block store is now removed completely #3463.

tgrospic:
Closing, with other changes to node setup code this will be resolved.
- #3679

PR #3397, opened 2021-05-03
Make reporting replay system deploys

### LINKED ISSUE #2981: make reporting api replay system deploys
## Overview
Currently the reporting api is only replaying the user deploy but not the system deploy(closeBlock or slashing).

Sometime we want to replay the system deploy to get debug infos.

We also need documentation on reporting api also.

### Impact
No big impact on the core protocol
It would increase the disk storage on reporting cache.

### Dependencies
No

1. make reporting Casper replay with system deploy
2. make it configuable for request to replay or just fetch data from cache

zsluedem:
## Overview

Make reporting replay system deploys and adds parameter `forceReplay` to _trace_ API method to execute reporting replay even when exists in cache. 

Related https://github.com/rchain/rchain/issues/2981

Reporting API is now also available together with other methods on web and admin `/api/trace` route. Old `reporting/trace` is not changed.


zsluedem:
bors try

zsluedem:
bors try

zsluedem:
bors try

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -786,7 +778,14 @@ In both cases we want to check reply data and see if everything is in order */
...
       blockData: BlockData
-  )(processedSystemDeploy: ProcessedSystemDeploy): F[Option[ReplayFailure]] = {
+  )(processedSystemDeploy: ProcessedSystemDeploy): F[Option[ReplayFailure]] =
+    replaySystemDeployE(blockData)(processedSystemDeploy).flatMap(
+      _.swap.value.map(_.toOption)
+    )
+
+  def replaySystemDeployE(
+      blockData: BlockData
+  )(processedSystemDeploy: ProcessedSystemDeploy): F[EitherT[F, ReplayFailure, Unit]] = {
```
tgrospic: Function should return `EitherT[F, _, _]` or `F[Either[_, _]]`. Wrapped EitherT in F can always be extracted with `EitherT.value`.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperReportingSpec.scala
```diff
@@ -29,33 +29,24 @@ class MultiParentCasperReportingSpec extends FlatSpec with Matchers with Inspect
...
           case Right(value) =>
-            value.head._2.foldLeft(0)(
+            value.deployReportResult.head.processedDeploy.deployLog.foldLeft(0)(
               (sum, e) =>
-                sum + e.count {
-                  case ReportingComm(_, _) => true
-                  case _                   => false
+                e match {
+                  case CommEvent(_, _, _) => sum + 1
+                  case _                  => sum
```
tgrospic: Can we use `collect` here?
```scala
value.deployReportResult.head.processedDeploy.deployLog.collect {
    case CommEvent(_, _, _) => 1
}.sum
```

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -180,10 +143,27 @@ object ReportingCasper {
...
+                           case Right(s) => s
+                         }
+                       } yield SystemDeployReportResult(term.systemDeploy, r)
+                     }
+                   )
+          checkPoint <- EitherT.right[ReplayFailure](runtime.createCheckpoint)
+          result <- EitherT.right[ReplayFailure](
+                     ReplayResult(res, sysRes, checkPoint.root.toByteString).pure[F]
+                   )
+        } yield result).value
```
tgrospic: What is the use of Either here, function can just return `ReplayResult`? Non of the expressions are returning Left so body can be rewritten like this.
```scala
val replayResult = for {
  _ <- runtime.reset(Blake2b256Hash.fromByteString(startHash))
  res <- terms.toList.traverse { term =>
          for {
            rd <- runtime
                   .replayDeployE(withCostAccounting)(term)
                   .map(_.semiflatMap(_ => runtime.getReport))
            res <- rd.value
            r = res match {
              case Left(_)  => Seq.empty[Seq[ReportingEvent]]
              case Right(s) => s
            }
          } yield DeployReportResult(term, r)
        }
  sysRes <- systemDeploys.toList.traverse { term =>
             for {
               rd <- runtime
                      .replaySystemDeployE(blockData)(term)
                      .map(_.semiflatMap(_ => runtime.getReport))
               res <- rd.value
               r = res match {
                 case Left(_)  => Seq.empty[Seq[ReportingEvent]]
                 case Right(s) => s
               }
             } yield SystemDeployReportResult(term.systemDeploy, r)
           }
  checkPoint <- runtime.createCheckpoint
} yield ReplayResult(res, sysRes, checkPoint.root.toByteString)

replayResult.asRight.pure[F]
```

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -1,111 +1,64 @@
...
-            )
-          case Right(deploys) =>
-            val t = transformDeploy(transformer)(_, _)
-            BlockTracesReport(
-              hash.base16String,
-              deploys.map(Function.tupled(t))
-            )
-          case _ => BlockReplayError(hash.base16String)
+          case Left(error) => BlockReportError(hash, error)
+          case Right(b)    => BlockTracesReport(b)
```
tgrospic: This can be simplified to pattern match with anonymous function.
```scala
state.map {
  case Left(error) => BlockReportError(hash, error)
  case Right(b) => BlockTracesReport(b)
}
```

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -1,111 +1,64 @@
...
-      reportingCasper: ReportingCasper[M]
-  )(implicit nt: M ~> F): HttpRoutes[F] = {
+  def service[F[_]: Concurrent](
+      blockReportAPI: BlockReportAPI[F]
+  ): ReportingHTTPRoutes[F] = {
     val dsl = org.http4s.dsl.Http4sDsl[F]
     import dsl._
     import io.circe.generic.extras.Configuration
     import io.circe.generic.extras.auto._
+    import coop.rchain.node.encode.JsonEncoder._
```
tgrospic, line 34: We can try to use this library to simplify this manual definitions.
https://github.com/scalapb-json/scalapb-circe

zsluedem: I was thinking of using this library, too. But I want to be consistent with what we had previously. https://github.com/rchain/rchain/blob/dev/node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala#L90-L109.

1. use this library only for this pr
2. use this library for other old encoding and this pr
3. use old fashion encoding

Which one do you think is better?

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperReportingSpec.scala
zsluedem: Yes. But it would take one more iteration.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -175,10 +137,24 @@ object ReportingCasper {
...
+                       r = rd match {
+                         case Left(_)  => Seq.empty[Seq[ReportingEvent]]
+                         case Right(s) => s
+                       }
+                     } yield SystemDeployReportResult(term.systemDeploy, r)
+                   }
+
+          checkPoint <- runtime.createCheckpoint
+          result     <- ReplayResult(res, sysRes, checkPoint.root.toByteString).pure[F]
+        } yield result.asRight
```
tgrospic: Why is return type Either when Left case will never be returned?

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -29,143 +34,100 @@ import coop.rchain.rspace.{RSpace, ReportingRspace, Match => RSpaceMatch}
...
+)
+final case class SystemDeployReportResult(
+    processedSystemDeploy: SystemDeployData,
+    events: Seq[Seq[ReportingEvent]]
+)
+final case class ReplayResult(
+    deployReportResult: List[DeployReportResult],
+    systemDeployReportResult: List[SystemDeployReportResult],
+    postStateHash: ByteString
+)
```
tgrospic, line 68: Please add short info about the use of these classes and also separate errors from results.

Is there a reason to have `ReportError` trait if we have only one class extend it?

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
```diff
@@ -0,0 +1,184 @@
...
+                  sysDeploys = createSystemDeployReport(
+                    result.systemDeployReportResult
+                  )
+                  blockEvent = BlockEventInfo(
+                    Some(lightBlock),
+                    deploys,
+                    sysDeploys,
+                    result.postStateHash
+                  )
+                } yield blockEvent.asRight[Error]
```
tgrospic: There exists `bitraverse` for Either type so it's not necessary to manually deconstruct and again construct each Left and Right case.

Also it makes sense to extract each case as a separate function to minimize nesting and make code easier to read. Something like this:
```scala
def toBlockInfo(result: ReplayResult): F[BlockEventInfo] = ???
def errorToString(err: ReportError): F[String] = ???

res <- reportResult.bitraverse(errorToString, toBlockInfo)
```

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
```diff
@@ -0,0 +1,184 @@
...
+                                 .asLeft[BlockEventInfo]
+                                 .pure[F]
+                           }
+                } yield result
+              },
+              Log[F]
+                .warn("Could not get event data.")
+                .as("Error: Could not get event data.".asLeft[BlockEventInfo])
+            )
+    } yield res
```
tgrospic: We can separate these kind of nesting with multiple functions which can be composed.
```scala
def createReport(casper: MultiParentCasper[F]): F[Either[Error, BlockEventInfo]] = {
  implicit val c = casper
  for {
    maybeBlock <- BlockStore[F].get(ByteString.copyFrom(Base16.unsafeDecode(hash)))
    report <- maybeBlock
               .traverse(blockReportWithinLock(forceReplay, _))
               .map(_.getOrElse(s"Block $hash not found".asLeft[BlockEventInfo]))
  } yield report
}

def validateReadOnlyNode(casper: MultiParentCasper[F]): F[Either[Error, MultiParentCasper[F]]] =
  casper.getValidator.map { pkOpt =>
    // Error for validator node
    if (pkOpt.isEmpty) casper.asRight[Error]
    else "Block report can only be executed on read-only RNode.".asLeft[MultiParentCasper[F]]
  }

// Process report if read-only node and block is found
def processReport(casper: MultiParentCasper[F]): F[Either[Error, BlockEventInfo]] =
  EitherT(validateReadOnlyNode(casper)).flatMapF(createReport).value

def casperNotInitialized: F[Either[Error, BlockEventInfo]] =
  Log[F]
    .warn("Could not get event data.")
    .as("Error: Could not get event data.".asLeft[BlockEventInfo])

EngineCell[F].read.flatMap(_.withCasper(processReport, casperNotInitialized))
```
This is also preparation for removal of EngineCell inside each method in API.

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -1,111 +1,61 @@
...
-  def transformDeploy[C, P, A, K](transformer: ReportingTransformer[C, P, A, K, RhoEvent])(
-      ipd: ProcessedDeploy,
-      events: Seq[Seq[ReportingEvent]]
-  ): DeployTrace =
-    DeployTrace(
-      ipd.deploy.sig.base16String,
-      ipd.deploy.data.term,
-      events.map(a => a.map(transformer.transformEvent).toList).toList
-    )
+  type ReportingHTTPRoutes[F[_]] = HttpRoutes[F]
```
tgrospic: `Http` should be in camel case like any other word.

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
tgrospic: Good point. It makes sense to use this library in all places. So better to leave it for separate PR.

node/src/main/scala/coop/rchain/node/web/package.scala
```diff
@@ -45,10 +50,7 @@ package object web {
       )
       extraRoutes = if (reporting)
         Map(
-          "/reporting" -> CORS({
-            implicit val n = natId[F]
-            ReportingRoutes.service[F, F](reportingCasper)
-          })
+          "/reporting" -> CORS(reportingRoutes)
```
tgrospic: It would be nice that force report is only available on internal port. It's expensive and can be abused. Can we easily split this?

casper/src/main/scala/coop/rchain/casper/ReportStore.scala
```diff
@@ -0,0 +1,42 @@
...
+    override def decode(bits: BitVector): Attempt[DecodeResult[BlockEventInfo]] =
+      Attempt.successful(
+        DecodeResult(
+          BlockEventInfo.parseFrom(decompressor.decompress(bits.toByteArray)),
+          BitVector.empty
+        )
+      )
+
+    override def sizeBound: SizeBound = unknown
+  }
```
tgrospic: There are helper functions to define new Codecs more easily without using Codec constructor directly.
```scala
val blockEventInfoCodecCompressed =
  scodec.codecs.bytes.xmap[BlockEventInfo](
    bv => BlockEventInfo.parseFrom(decompressor.decompress(bv.toArray)),
    bei => ByteVector(compressor.compress(bei.toByteArray))
  )
```

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
zsluedem: Haha. Thanks to your comments. I think I made some useless codes before.

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
zsluedem: Nice.

zsluedem: I have to say that's much better. I was unsatisfied with my codes. Good work.

casper/src/main/scala/coop/rchain/casper/ReportStore.scala
zsluedem: I don't like some of the builtin Codec in SCodec because it would append some wired strings in the codec implementation.

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
```diff
@@ -43,56 +43,40 @@ class BlockReportAPI[F[_]: Concurrent: Metrics: EngineCell: Log: SafetyOracle: B
...
+      lock      = blockLockMap.getOrElseUpdate(b.blockHash, semaphore)
       result <- lock.withPermit(
                  for {
                    cached <- reportStore.get(b.blockHash)
                    res <- if (cached.isEmpty || forceReplay)
-                           for {
-                             blockEvent <- replayBlock(b)
-                             _ <- blockEvent.fold(
-                                   _ => ().pure[F],
-                                   data => reportStore.put(b.blockHash, data)
```
zsluedem: @tgrospic The result would never be stored in cache because  `reportStore.put(b.blockHash, data)` is removed now.

tgrospic:
As we discussed on standup. Let's keep reporting API on external and internal the same with force parameter included. Reporting is protected with config flag.

casper/src/main/scala/coop/rchain/casper/ReportingCasper.scala
```diff
@@ -175,10 +153,25 @@ object ReportingCasper {
...
+                              .semiflatMap(_ => runtime.getReport)
+                              .value
+                       r = rd match {
+                         case Left(_)  => Seq.empty[Seq[ReportingEvent]]
+                         case Right(s) => s
+                       }
+                     } yield SystemDeployReportResult(term.systemDeploy, r)
+                   }
+
+          checkPoint <- runtime.createCheckpoint
```
zsluedem, line 161: `createCheckpoint` here would store the items again which is not neccessary at all.  All we need here is the proof while the history current implementation is difficult to change the storing part we would leave that into another pr.

tgrospic (approved):
Nice refactoring!

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #2629, opened 2019-07-29
Rchain 3711 remove r

kayvank:
## Overview
<sup>Remove redundant parameter R from rspace match trait</sup>

### JIRA ticket:
<sup>https://rchain.atlassian.net/browse/RCHAIN-3711</sup>

### Notes
<sup>For all known use-cases the Rspace Match parameter R is the same as the Data Parameter A. This pull request is to simplify the Match trait by removing the redundant parameter. 
Rsapace "source produce/consume, RCHAIN-3446, and Rspace sequence number, RCHAIN-3511 are dependent on this pull request.</sup>

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3398, opened 2021-05-04
Fix "tut" sbt plugin outaged problem

zsluedem:
## Overview

Currently the sbt using https://github.com/tpolecat/tut to generate doc while it is not really used by any of us.
The binaries is not available after May1 2021. That's why new github action tests are failing on fetching the dependency->https://github.com/rchain/rchain/runs/2501268886?check_suite_focus=true

image

Currently my solution is to remove the plugin. **I tried to resolve it by replace by https://scalameta.org/mdoc/docs/tut.html  but failed on fixing deeply dependencies resolve problems.**


build.sbt
```diff
@@ -474,7 +473,6 @@ lazy val rspace = (project in file("rspace"))
       guava
     ),
     /* Tutorial */
-    tutTargetDirectory := (baseDirectory in Compile).value / ".." / "docs" / "rspace",
```
tgrospic, line 477: Do we have generated files checked-in in the repo?

_Tutorial_ comment should also be removed, right?

zsluedem:
```
sbt:rchain> rspace/tutSourceDirectory
[info] Total time: 4 ms
[info] /${HOME}/rchain/rchain/rspace/src/main/tut
[info] Total time: 5 ms
```
The source directory is not exist. It should be some old developer's work.

The output file is https://github.com/rchain/rchain/blob/dev/docs/rspace/Tutorial.md.

@tgrospic Do you think we should remove https://github.com/rchain/rchain/blob/dev/docs/rspace/Tutorial.md?

tgrospic: Oh, I see now. Tutorial.md is not generated but type checked by tut so it should stay.

I added a task to Core board to include replacement mdoc later.

tgrospic (approved):
LGTM

build.sbt
tgrospic: Also Tutorial.md source file (tut) is deleted here https://github.com/rchain/rchain/pull/2629 ???

nzpr (approved):
LGTM

tgrospic:
bors merge

PR #3408, opened 2021-05-15
[PoC] RSpace with hashed channels

tgrospic:
## Overview

The purpose of this experiment is to show that RSpace can be defined without using _complete_ channel type but hash only.


tgrospic:
bors try

PR #3414, opened 2021-05-21
Send all tips on fork choice request

nzpr:
Currently there is only one message that gives information on latest blocks that some node seen - fork choice tip request. It is just tip that is a fork choice, so there is no way to see all tips. 

In multi parent environment, when node misses some block, e.g. due to network issues - it has no way to request this block because all requests for other nodes to give the latest view return only one tip, which is most probably not hash that has been missed.
 
This PR suggests to send all tips as a response for fork choice tip request.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Running.scala
@@ -168,19 +168,22 @@ object Running {
   /**
     * Peer asks for fork-choice tip
     */
+  // TODO name for this message is misleading, as its a request for all tips, not just fork choice.
   def handleForkChoiceTipRequest[F[_]: Sync: TransportLayer: RPConfAsk: BlockStore: Log](
       peer: PeerNode
   )(casper: MultiParentCasper[F]): F[Unit] = {
     val logRequest = Log[F].info(s"Received ForkChoiceTipRequest from ${peer.endpoint.host}")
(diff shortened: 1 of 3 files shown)


casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -57,13 +57,15 @@ object Running {
       _ <- engine.withCasper(
             casper => {
               for {
-                tipHash      <- MultiParentCasper.forkChoiceTip[F](casper)
```
tgrospic, line 60: Why not update this function instead of duplicating code in both places where this function was used?

nzpr: Because it's not fork choice tip - we need all tips. But looks like this function is not used anywhere else... Well, this is to my question why at all we need  to know fork choice tip. So might be we can remove/ rename the message. But I wanted this PR to not touch much

nzpr: I removed this function at all in <https://github.com/rchain/rchain/pull/3414/commits/0d3c799ac0be4643549554491d0c986e99174b1f>

tgrospic (approved):
LGTM

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3418, opened 2021-05-29
Make Rev vault logging mandatory and read-only

Isaac-DeFrain:
## Overview

We want to include transactions on-chain in the form of a log in the corresponding Rev vaults. We want this to be enabled automatically for all Rev vaults and we want the logs to be read-only. For compliance with exchanges, we also need to give each transaction a unique id. Lastly, we want to store the raw (unhashed) Rev address in the Rev vault map since the keys are hashed.

This PR automatically enables Rev vault logging and makes it read-only to nodes. Unique ids are now generated for each Rev vault transfer as an unforgable name. Logging is done in both the sender and receiver vaults. Logging in purses was unnecessary and is removed by this PR.

### Notes

These changes should take care of most of the Rev vault contract items.


Isaac-DeFrain:
bors try

casper/src/main/resources/RevVault.rho
```diff
@@ -175,100 +219,139 @@ in {
...
+              }
+            } |
+            // Returns an entry of the vault's log
+            contract revVault(@"readLogEntry", ret) = {
+              new thmRet in {
+                // Lookup the log channel associated with `fromRevAddress`
+                TreeHashMap!("get", vaultMap, fromRevAddress, *thmRet) |
+                for (@(_, _, *logCh) <- thmRet) {
+                  // Read a transfer entry and return it on `ret`
+                  for (@("transfer", entry) <<- logCh) {
```
zsluedem, line 289: I just want to say that `peek` would just consume **one** record on the `logCh` if `logCh` got some many of logs of it

tgrospic: You are right. I'm not sure that we can _extract_ all log entries inside Rholang but only with direct _get data at name_ API method.

zsluedem:
@tgrospic 
Here are the current APIs which are used by the [Revdefine](https://revdefine.io). 

What this on-chain transaction history can help is the `/transaction/:address`(without paginations.) endpoint. It can not help with other options. Other small endpoints needs smart indexes to query.

And in order to support revdefine front end by rnode (without revdefine endpoint), we could remove functions in front end. 

**I am not sure what is the ultimate goal you want from open source perspective.**

If community(coop) needs a commercial explorer like etherscan, we should not put these smart api in the rnode because it is clumsy and users-response-sensitive.

If coop needs a open source explorer like or [Expedition](https://expedition.dev/?network=Ethereum%20Classic), we really rethink from scratch what we can do.

casper/src/main/resources/NonNegativeNumber.rho
```diff
@@ -14,52 +14,71 @@
...
-    new this, valueStore in {
+    new
+      this,
+      valueStore
+    in {
+      // Adds `x` to the value if no overflow occurs, returns success message on success
       contract this(@"add", @x, success) = {
         if (x >= 0) {
-          for(@v <- @(*MergeableTag, *valueStore)){
+          for (@v <- @(*MergeableTag, *valueStore)){
```
zsluedem, line 31: Conflicts in `NonNegativeNumber` is because  Issac's comments changes and the `MergeableTag`.

casper/src/main/resources/RevVault.rho
```diff
@@ -175,100 +219,139 @@ in {
...
           } |
-          // Supplies balance, transfer, and setLog methods on revVault.
-          contract _newVault(revVault, @ownRevAddress, purse) = {
-            new logStore in {
-              logStore!(Nil) |
-              contract revVault(@"balance", ret) = {
-                purse!("getBalance", *ret)
-              } |
-              // Transfers from revVault to toRevAddress.
-              contract revVault(@"transfer", @targetAddress, @amount, authKey, ret) = {
```
zsluedem, line 186: @tgrospic 
Conflicts here because the renaming `targetAddress` here. I used Issac's name `fromRevAddress` now. Let me know if you have questions.

casper/src/main/resources/RevVault.rho
```diff
@@ -175,100 +219,139 @@ in {
...
           } |
           // Checks all necessary validation and authentication prerequisites. If everything checks out,
-          // amount is transferred from purse (ownRevAddress) to vault at targetAddress.
-          contract _transferTemplate(@ownRevAddress, purse, @targetAddress, @amount, authKey, ret) = {
-            new revAddressValid, revAddressValidEither, amountNonNegative,
-                authKeyValidCh, authKeyValidEitherCh,
-                parametersOkCh, parametersAndAuthOkCh,
-                split, eitherPurseCh, doDeposit
+          // `amount` is transferred from vault at `fromRevAddress` to vault at `toRevAddress`.
+          contract _transferTemplate(@fromRevAddress, purse, @toRevAddress, @amount, authKey, ret) = {
```
zsluedem, line 302: another naming conflicts `fromRevAddress` and `toRevAddress` compared to `ownerRevAddress` and `targetRevAddress`

casper/src/main/resources/RevVault.rho
```diff
@@ -175,100 +219,139 @@ in {
...
+                @{toRevAddress | bundle0{*_revVault}}!("_deposit", fromPurse, *retCh)
               }
             }
           } |
-          // Deposits the entire balance in the fromPurse to the toPurse.
-          contract _depositTemplate(fromPurse, toPurse, retCh) = {
-            new amountCh, depositSuccessCh in {
-              toPurse!("getBalance", *amountCh) |
+          // Deposits the entire balance in `fromPurse` to `toPurse`
+          contract _depositTemplate(toPurse, fromPurse, retCh) = {
```
zsluedem, line 340: @tgrospic I noticed you changed the order here before. I don't know why. Should I use your change? I think current naming is fine..

zsluedem:
bors try

zsluedem:
bors merge

zsluedem:
bors try

zsluedem:
And there is one limitation for current on-chain transactions. Only successful transaction can be recorded.

zsluedem:
bors try

zsluedem:
https://github.com/rchain/rchain/runs/5525318221?check_suite_focus=true#step:8:549 
https://github.com/rchain/rchain/actions/runs/1973196591
https://github.com/rchain/rchain/runs/5508273694?check_suite_focus=true#step:8:220
non-determined bug

casper/src/main/resources/RevVault.rho
```diff
@@ -175,100 +219,139 @@ in {
...
+                              new
+                                fromLogRet,
+                                toLogRet
+                              in {
+                                // Lookup vault associated with `toRevAddress`
+                                TreeHashMap!("get", vaultMap, toRevAddress, *toLogRet) |
+                                // Lookup log channel associated with `fromRevAddress`
+                                TreeHashMap!("get", vaultMap, fromRevAddress, *fromLogRet) |
+                                for (@(_, _, *fromLogCh) <- fromLogRet;
+                                     @(_, _, *toLogCh)   <- toLogRet) {
```
zsluedem, line 263: It is possible the `toAddress` vault is not created and the `get` above would return `Nil`. 
In this case the transfer is done and the `ret` won't return anything and the log would be recorded.

PR #3419, opened 2021-06-01
Syntax for bytestring based models

nzpr:
This PR unifies syntax for bytestring based models (Validator, StateHash, BlockHash). This is some work from deploymerge branch which belongs to dev.

### DIFF
--- a/shared/src/main/scala/coop/rchain/shared/package.scala
+++ b/shared/src/main/scala/coop/rchain/shared/package.scala
@@ -18,3 +18,4 @@ trait AllSyntaxShared
     with KeyValueStoreManagerSyntax
     with MonixableSyntax
     with Fs2StreamSyntax
+    with catscontrib.ToBooleanF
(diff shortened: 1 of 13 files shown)


shared/src/main/scala/coop/rchain/shared/package.scala
```diff
@@ -18,3 +18,4 @@ trait AllSyntaxShared
     with KeyValueStoreManagerSyntax
     with MonixableSyntax
     with Fs2StreamSyntax
+    with catscontrib.ToBooleanF
```
zsluedem, line 21: Just curious.
This is not related to byteString, right?

tgrospic: I'm ok for ToBooleanF to also be included in this PR. It needs to be renamed.

zsluedem (approved):
LGTM

tgrospic (approved):
LGTM

nzpr:
bors merge

PR #3421, opened 2021-06-02
Clique oracle rewrite v2

nzpr:
Previous attempt to rewrite clique oracle https://github.com/rchain/rchain/pull/3387 got some critique, particularly due to making some unnecessary changes and use DagReader trait heavily, which is questionable and subject to removal. This PR removes DagReader trait and, moving some of routines used by previous Clique oracle to BlockDagRepresentation syntax.


block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
```diff
@@ -142,7 +141,7 @@ final class InMemBlockDagStorage[F[_]: Concurrent: Sync: Log](
                                                     s"Block ${Base16.encode(block.blockHash.toByteArray)} sender is empty"
                                                   ) >> newValidatorsLatestMessages.pure[F]
                                                 } else if (block.sender
-                                                             .size() == BlockHash.Length) {
+                                                             .size() == Validator.Length) {
```
nzpr: We had this bug, but I guess it's ok to fix it in feature branch and not allocate PR on dev for that?

tgrospic: It's used in tests, how that is not failing now?

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
```diff
@@ -72,13 +54,10 @@ object DagMerger {
       applyTrieActions = (baseState: Blake2b256Hash, actions: Seq[HotStoreTrieAction]) =>
         historyRepository.reset(baseState).flatMap(_.doCheckpoint(actions).map(_.root))
 
-      // set that should be merged into LFB
-      lfbPostState = postState(lfb)
-
       r <- ConflictSetMerger.merge[F, R](
```
tgrospic: Maybe this is just me but reading `merge[F, DeployChainIndex]` is making more sense to me then `merge[F, R]`.

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,178 @@
...
+    *
+    *    If one found - this is a source of disagreement.
+    * */
+  private def neverEventuallySeeDisagreement[F[_]: Sync](
+      validatorA: V,
+      validatorB: V,
+      dag: BlockDagRepresentation[F],
+      targetMessage: M
+  ): F[Boolean] = {
+    def mightEventuallyDisagree(lmB: BlockHash, lmAjB: BlockHash): F[Boolean] =
```
tgrospic: Just from reading the signature of these functions here it's not clear if `M` represents the same type as for `lmB` or `lmAjB`. I'm not sure why is alias `M` and `V` here useful especially as `Validator` is already an alias.

Usually type alias is used to be give more specific name or to shorten the name of more complex types, and not as more general name.

> Type aliases are a useful way to refer to existing types with a local, specific name. A Tuple2[Int,String] used regularly inside a class may be more useful if it were named UserInfo. However, as with other advanced type features, type aliases should not replace careful object-oriented design. A real class named UserInfo will be more stable and intuitive in the long term than using type aliases.
https://www.oreilly.com/library/view/learning-scala/9781449368814/ch10.html

https://www.baeldung.com/scala/type-declaration#type-alias

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -79,25 +65,100 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+            .compile
+            .to(Set)
+    } yield r
+
+  def descendants(blockHash: BlockHash)(implicit sync: Sync[F]): F[Set[BlockHash]] =
+    Stream
+      .unfoldLoopEval(List(blockHash)) { lvl =>
+        for {
+          out  <- lvl.traverse(dag.children).map(_.flatten.flatten)
+          next = out
```
tgrospic, line 156: This assignment seems like a leftover?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -79,25 +65,100 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+
+  def nonFinalizedBlocks(implicit sync: Sync[F]): F[Set[BlockHash]] =
+    for {
+      tips <- latestMessages.map(_.values.map(_.blockHash).toList)
+      r <- Stream
+            .unfoldLoopEval(tips) { lvl =>
+              for {
+                out  <- lvl.filterA(dag.isFinalized(_).not)
+                next <- out.traverse(dag.lookup(_).map(_.map(_.parents))).map(_.flatten.flatten)
+              } yield (out, next.nonEmpty.guard[Option].as(next))
```
tgrospic, line 144: Nice use of Alternative type class.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -79,25 +65,100 @@ final class BlockDagRepresentationOps[F[_]: Sync](
...
+    val result = for {
+      aHeight <- OptionT(dag.lookup(ancestor).map(_.map(_.blockNum)))
+      is = mainParentChain(descendant, aHeight)
+        .filter(_ == ancestor)
+        .head
+        .compile
+        .last
+        .map(_.isDefined)
+      r <- OptionT.liftF(is)
+    } yield r
```
tgrospic: Usually with monad transformers like OptionT or fs2.Stream _for_ comprehension can add more boilerplate and in some situation with Streams even more safe.
https://gitter.im/functional-streams-for-scala/fs2?at=60032d6714cec811eca1b5f9

Start with transformer (DSL) of choice and compose other operations.
```scala
val result = OptionT(dag.lookup(ancestor).map(_.map(_.blockNum))).semiflatMap { aHeight =>
  mainParentChain(descendant, aHeight)
    .filter(_ == ancestor)
    .head
    .compile
    .last
    .map(_.isDefined)
}
```

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,178 @@
...
+    val noDisagreement = for {
+      lmA <- OptionT(dag.latestMessageHash(validatorA))
+      lmB <- OptionT(dag.latestMessageHash(validatorB))
+      // justification from `validatorB` as per latest message of `validatorA`
+      lmAjBF = dag.lookupUnsafe(lmA).map {
+        _.justifications.find(j => j.validator == validatorB).map(_.latestBlockHash)
+      }
+      lmAjB <- OptionT(lmAjBF)
+      r     <- OptionT.liftF(mightEventuallyDisagree(lmB, lmAjB).not)
+    } yield r
```
tgrospic: This is just as an example how OptionT can be used without _for_ comprehension using combinator functions like `flatMapF`, `tupled` and `semiflatMap`.
```scala
// justification from `validatorB` as per latest message of `validatorA`
val lmAjB = OptionT(dag.latestMessageHash(validatorA)).flatMapF(
  dag
    .lookupUnsafe(_)
    .map(_.justifications.find(_.validator == validatorB).map(_.latestBlockHash))
)
val lmB = OptionT(dag.latestMessageHash(validatorB))
// Check for disagreement, if messages are not found return false
(lmB, lmAjB).tupled.semiflatMap((mightEventuallyDisagree _).tupled).getOrElse(false)
```

NOTE: `(mightEventuallyDisagree _).tupled` <- this is not the same as `(OptionT, OptionT).tupled`

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
```diff
@@ -0,0 +1,178 @@
...
+
+    /** weight map containing only validators that agree on the the message */
+    def agreeingWeightMapF(weightMap: WeightMap): F[WeightMap] = {
+      def agree(validator: V, message: M): F[Boolean] =
+        dag
+          .latestMessageHash(validator)
+          .flatMap(_.traverse(dag.isInMainChain(message, _)))
+          .map(_.getOrElse(false))
+
+      fs2.Stream
```
tgrospic: `fs2` namespace is already imported.

casper/src/main/scala/coop/rchain/casper/util/rholang/InterpreterUtil.scala
```diff
@@ -253,61 +252,39 @@ object InterpreterUtil {
...
+                for {
+                  b <- BlockStore[F].getUnsafe(v)
+                  index <- BlockIndex(
+                            b.blockHash,
+                            b.body.deploys,
+                            Blake2b256Hash.fromByteString(b.body.state.preStateHash),
+                            Blake2b256Hash.fromByteString(b.body.state.postStateHash),
+                            runtimeManager.getHistoryRepo
+                          )
+                } yield index
```
tgrospic: With `getOrElse` we can escape deep nesting.
```scala
val cached = BlockIndex.cache.get(v).map(_.pure)
cached.getOrElse {
  for {
    b <- BlockStore[F].getUnsafe(v)
    index <- BlockIndex(
              b.blockHash,
              b.body.deploys,
              Blake2b256Hash.fromByteString(b.body.state.preStateHash),
              Blake2b256Hash.fromByteString(b.body.state.postStateHash),
              runtimeManager.getHistoryRepo
            )
  } yield index
}
```
Or without _for_ comprehension.
```scala
val cached = BlockIndex.cache.get(v).map(_.pure)
cached.getOrElse {
  BlockStore[F].getUnsafe(v).flatMap { b =>
    BlockIndex(
      b.blockHash,
      b.body.deploys,
      Blake2b256Hash.fromByteString(b.body.state.preStateHash),
      Blake2b256Hash.fromByteString(b.body.state.postStateHash),
      runtimeManager.getHistoryRepo
    )
  }
}
```

tgrospic (commented):
This is good improvement of clique oracle.

I added some comments but more focused on the use of _for_ comprehension and monad transformers.

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
nzpr: I'm thinking about R as minimal rejection unit, it just happens that now it is DeployChainIndex, it might be something else in theory, or have some another name.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
nzpr: I wanted to be explicit here, because `unfoldLoopEval ` always wants out and next. This logic assumes that next = out. But in function above, e.g. this is not the case. So I just found it more clear.

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
nzpr: This is a typo - I'd like to use M here as well. As it stands for message, and Capser and clique oracle is about messages. Thats why I don't want to have BlockHash. But as BlockDagRepresentation needs BlockHash, I defined type alias.

tgrospic: Why not call it `Message` or `Msg` instead of `M`?

The idea of an alias is also to make it easier for the reader to follow the code.

nzpr: It is said that this is a message here https://github.com/rchain/rchain/blob/bf23fb1ee2492f422fe6243fa04f9f96ecd9a379/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala#L22
And pretty often you have to pass in variables like this 
https://github.com/rchain/rchain/blob/bf23fb1ee2492f422fe6243fa04f9f96ecd9a379/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala#L67
So you duplicating name of the variable and type, which I want to avoid. As variable name is something that is used and exposed in code. But anyway, if you think this is a bad practice - I will adjust.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
nzpr: Im not totally sure , but It looks like this dagstore is used now mostly
https://github.com/rchain/rchain/blob/5a9b6d418ebb4cc86781ae69070d629491a6414a/casper/src/test/scala/coop/rchain/casper/helper/BlockDagStorageFixture.scala#L81-L90 which is not inmem

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
tgrospic: My concert is that type alias should not be used lightly. Especially if not used for the purpose of naming complex types or giving more specific name for general type. On top of that aliasing the alias is very confusing.

How can you explain this to a new developer joining the project? Can you find any reference that shows this is a good practice?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
tgrospic: Actually we don't need this in memory DAG storage anymore. It leftover of file-based storage and with key-value based storage we can easily create in memory variant and use the same code in testing and production.

https://github.com/rchain/rchain/blob/2d5aac8a270ee8053240eb340db9bb3624b52372/casper/src/test/scala/coop/rchain/casper/engine/Setup.scala#L119-L121

Here is the PR to remove it #3426.

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
nzpr: I pushed change to se direct types in DagMerger https://github.com/rchain/rchain/pull/3421/commits/3cadf0e801a9978adc9f861eb601750fe55f9172
I agree that as it is not abstracted over the type, I still pull the actual type inside scope, so it does not have much sense.

But are you ok to have ConflictSetMerger abstracted with R as minimal rejection unit?

casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala
tgrospic: I don't see duplication if existing alias is used. I'd expect that `M` is generic type.
```scala
targetMessage: BlockHash
```
If you like we can experiment on a more places to see if gives more clarity.

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
tgrospic: This can be part of the experiment. :+1:

casper/src/test/scala/coop/rchain/casper/util/rholang/MergingBranchMergerSpec.scala
```diff
@@ -339,64 +335,77 @@ class MergingBranchMergerSpec extends FlatSpec with Matchers {
...
+            _                            = assert(nextPreStateHash != base.body.state.postStateHash)
+            _                            = println(s"merge result ${PrettyPrinter.buildString(nextPreStateHash)}")
           } yield nextPreStateHash
 
         for {
-          // create base state - child of genesis
-          _ <- List.range(0, 4).foldLeftM(genesisPostStateHash) { (baseHash, i) =>
-                mkLayer(baseHash, i)
-              }
+          dagStore <- InMemBlockDagStorage.create[Task]
```
tgrospic: You can fix this test with with in-memory BlockDagKeyValueStorage.
https://github.com/rchain/rchain/blob/2d5aac8a270ee8053240eb340db9bb3624b52372/casper/src/test/scala/coop/rchain/casper/engine/Setup.scala#L119-L121

For easier sync with dev just undo changes to InMemBlockDagStorage and we can merge the PR. :tada:

block-storage/src/main/scala/coop/rchain/blockstorage/dag/InMemBlockDagStorage.scala
tgrospic: We can leave this file as it is. It will be removed after sync with dev.

tgrospic (approved):
Excellent work! Thanks for addressing the comments on this long PR. :rocket:

PR #3423, opened 2021-06-03
Enable protocol client timeout

nzpr:
Current configuration contains timeout for protocol client, but it is used not in client instantiated but in kademlia client and RPConf class. 

This PR removes field from RPConf as it is not used anywhere and adds timeout to GrpcTransportClient constructor. So both kademlia client and protocol client share the same timeout variable.

### DIFF
--- a/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
+++ b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
@@ -44,12 +44,11 @@ class GrpcTransportClient[F[_]: Monixable: Concurrent: Parallel: Log: Metrics](
     maxMessageSize: Int,
     packetChunkSize: Int,
     clientQueueSize: Int,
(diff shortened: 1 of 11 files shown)


comm/src/main/scala/coop/rchain/comm/transport/GrpcTransportClient.scala
```diff
@@ -162,7 +161,7 @@ class GrpcTransportClient[F[_]: Monixable: Concurrent: Parallel: Log: Metrics](
   ): F[Unit] = {
 
     def timeout(packet: Packet): FiniteDuration =
-      Math.max(packet.content.size().toLong * 5, DefaultSendTimeout.toMicros).micros
+      Math.max(packet.content.size().toLong * 5, networkTimeout.toMicros).micros
```
tgrospic, line 164: My concern here is that two different algorithms are mixed together. We are choosing from parameter that is relative to packet size and absolute value. This means slower node must opt-in to absolute value.
Although I'm not sure right no what would be easy way to fix it.

PR #3430, opened 2021-06-11
Add treeHashMap mergebility test 

### LINKED ISSUE #3381: Check if all the rholang libraries are mergeable
## Overview

Based on https://github.com/rchain/rchain/pull/3355/files#r610558141.

We should still keep some tests or even more tests to cover most of the basic [rholang library](https://github.com/rchain/rchain/tree/dev/casper/src/main/resources) to be sure that all the libraries are mergeble.

zsluedem:
## Overview

This pr is handling one of the library for core rholang -> treeHashMap.

Add tests to check if the libraries method is mergeable on read, write, update with the same map.

Related https://github.com/rchain/rchain/issues/3381. This test is removed in #3355 


zsluedem:
bors try

zsluedem:
bors try

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
```diff
@@ -15,12 +15,16 @@ import coop.rchain.shared.Log
...
+    (r: DeployChainIndex) => r.deploysWithCost.map(_.cost).sum
+
   def merge[F[_]: Concurrent: Log](
       dag: BlockDagRepresentation[F],
       lfb: BlockHash,
       lfbPostState: Blake2b256Hash,
       index: BlockHash => F[Vector[DeployChainIndex]],
-      historyRepository: RhoHistoryRepository[F]
+      historyRepository: RhoHistoryRepository[F],
+      optimalRejectionAlg: DeployChainIndex => Long
```
zsluedem: @nzpr @tgrospic Sorry. I still expose this method out. 
It would be a problem for some random generated test cases that I don't have to way to decide which part I want to reject.

.github/workflows/continuous-integration.yml
```diff
@@ -124,6 +124,7 @@ jobs:
           - "'casper/test:testOnly coop.rchain.casper.engine.*'"
           - "'casper/test:testOnly coop.rchain.casper.genesis.*'"
           - "'casper/test:testOnly coop.rchain.casper.util.*'"
+          - "'node/test:testOnly coop.rchain.node.mergeability.*'"
```
zsluedem: move all the mergeability test in the `node`

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+            i,
+            RhoTrieTraverser
+              .byteArrayToNybbleList(RhoTrieTraverser.keccakKey(i), 0, length = 1, Vector.empty)
+          )
+      )
+//    val head1 = insertKeyValues.groupBy(i => i._2.head)
+    val head2 = insertKeyValues.groupBy(i => i._2.slice(0, 2))
+//    println(head1(47))
+    println(head2(Vector(3, 10)))
+  }
```
tgrospic: This looks like leftover.

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+    *               /      \                         \
+    *             10        11(Add new node)          2
+    *            |            |                       |
+    *    {"XspL2n7hsd":3}   {"VbqWeRbRoT":1}      {"g6smo8DVcr": 4}
+    *
+    *
+    * Right:
+    *                           rootNode
+    *                     /                \
+    *                  47              4(Changed exist node)
```
tgrospic: This should be `3` instead of `47` for left node as in base.

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+      readTrieRho(initialKVS.map(_.key)),
+      trieInitializedRho(depth, initialKVS),
+      initialKVS,
+      depth
+    )
+  }
+
+  it should "read and write on different keys on the same treeHashMap is mergeable" in forAll(
+    listKeyValuesGen,
+    listKeyValuesGen,
```
tgrospic, line 282: Did you found somewhere if it's guarantied for strings in these lists to have different values? I'm wondering if this is the case or it can happen randomly to return the same string in both.

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+    Gen.chooseNum(0, 16),
+    minSuccessful(3)
+  ) { (initialKVS, writeKeyValues, depth) =>
+    val expectedKeyValues = (initialKVS ++ writeKeyValues)
+      .foldLeft(Map.empty[String, String]) {
+        case (acc, kv) => acc.updated(kv.key, kv.value)
+      }
+      .toList
+      .map(kv => KeyValue(kv._1, kv._2))
```
tgrospic: Small trick to convert tuple to functions arguments.
```scala
.map(KeyValue.tupled)
```

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+      treeHashMapDepth: Int
+  ): Unit =
+    runTest(left)(right)(base)(rejectRight = rejectRight)(isConflict = true)(
+      expectedKeyValue = expectedKeyValue
+    )(treeHashMapDepth = treeHashMapDepth)
+
+  private def runTest(left: String)(right: String)(base: String)(
+      rejectRight: Boolean
+  )(isConflict: Boolean)(expectedKeyValue: List[KeyValue])(treeHashMapDepth: Int): Unit = {
+    implicit val concurrent                = Concurrent[Task]
```
tgrospic: Using Task has already in scope Concurrent instance.

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
```diff
@@ -0,0 +1,542 @@
...
+      containTrieRho(keyValues.map(_.key)),
+      trieInitializedRho(depth, keyValues),
+      keyValues,
+      depth
+    )
+  }
+  private def readKeyRho(keys: List[String]): String =
+    keys.foldLeft("Nil") {
+      case (acc, key) => acc + s"""|new a in {TreeHashMap!("get", treeMap, "${key}", *a)}\n"""
+    }
```
tgrospic: `mkString` is handy in these cases with separator.

```scala
keys.map(key => s"""new ... $key ...""").mkString(" |\n")
```

node/src/test/scala/coop/rchain/node/mergeablity/MergeabilityRules.scala
```diff
@@ -0,0 +1,297 @@
...
+                  | isRejectRight: ${rejectRight}
+                  | Merged state
+                  | ${dataContinuationAtMergedState}
+                  | should be
+                  | ${mergedStateResult}
+                  |
+                  | conflicts found: ${mergedState._2.size}
+                  |
+                  | go see it at ${file.value}:${line.value}
+                  | """.stripMargin
```
tgrospic, line 279: Scala Test has support for custom `Matcher`s which can provide correct source reference where assertion is created so there is no need for manual manipulation of source info.

Maybe you can create a custom matcher for `State` similarly to this one for empty stream. It can provide the same info but with default Scala Test failed behavior.
https://github.com/rchain/rchain/blob/cba9de55821380fdf4994c05a2320cd78e229f9c/casper/src/test/scala/coop/rchain/casper/util/scalatest/Fs2StreamMatchers.scala#L17-L30

.github/workflows/continuous-integration.yml
tgrospic: No need to specify test here, it will run with other tests in node project. This also disables running of all other defined tests in node project.
I curious why you moved test to node project? Is it some reference unavailable?

https://github.com/rchain/rchain/blob/a17835b2cb5a6bb25097dd0a74debde0afa6f843/.github/run-unit-test-selection#L4-L11

zsluedem: > Is it some reference unavailable?

Yes. `import coop.rchain.node.revvaultexport.RhoTrieTraverser`.

node/src/test/scala/coop/rchain/node/mergeablity/TreeHashMapMergeabilitySpec.scala
zsluedem: You are right. We should avoid that.

.github/workflows/continuous-integration.yml
zsluedem: Should I move `coop.rchain.node.revvaultexport.RhoTrieTraverser` back to casper?

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
```diff
@@ -86,7 +84,6 @@ class JoinMergeability extends FlatSpec with Matchers with Inspectors with Basic
     R0.rstate ++ S1.rstate
   )
   it should "R J 4" in MergeableCase(J_)(F1)(R0)(J_.rstate ++ R0.rstate ++ F1.rstate)
-  it should "R J 4 2" in ConflictingCase(J_)(F0)(R0)(J_.rstate ++ R0.rstate)(R0.rstate)
```
zsluedem, line 88: dup with above

tgrospic (approved):
LGTM

Nice work!

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
nzpr: I'm ok with that, just imo its better to call it `rejectionCostF`, as in the end we are minimising total rejection cost of a branch

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
```diff
@@ -85,9 +84,8 @@ class JoinMergeability extends FlatSpec with Matchers with Inspectors with Merge
...
-  it should "R J C" in MergeableCase(C0)(J_)(R0)(J_.rstate ++ R0.rstate)
-  it should "R J C 2" in MergeableCase(C1)(J_)(R0)(J_.rstate ++ R0.rstate ++ C1.rstate)
+  it should "R J C" ignore MergeableCase(C0)(J_)(R0)(J_.rstate ++ R0.rstate)                // infinite loop
+  it should "R J C 2" ignore MergeableCase(C1)(J_)(R0)(J_.rstate ++ R0.rstate ++ C1.rstate) // infinite loop
   it should "R J R" in ConflictingCase(R0)(J_)(R0)(R0.rstate ++ R0.rstate)(
```
nzpr, line 87: I'm not sure - were these tests working before? There is nothing changed related to test execution, right?

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
```diff
@@ -1,484 +0,0 @@
...
-                        | Merged state
-                        | ${dataContinuationAtMergedState}
-                        | should be
-                        | ${referenceState}
-                        |
-                        | go see it at ${file.value}:${line.value}
-                        | """.stripMargin
-            _ <- Sync[Effect]
-                  .raiseError(new Exception(errMsg))
-                  .whenA(dataContinuationAtMergedState != referenceState)
```
zsluedem, line 476: @nzpr Codes only check the state. Never check if they are conflict or not

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
zsluedem: The codes didn't check if they are conflict or not. The codes only check the state before this pr. see line here https://github.com/rchain/rchain/pull/3430#discussion_r667345172

zsluedem: Let me check again. The commits are too old in my memory.

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
nzpr: but this effectively also checks conflict? if conflict detection is wrong - state will be wrong, do I miss something?

zsluedem: The name `MergeableCase` and `ConflictCase` don't make any difference.The test codes only check the state without caring about if they are conflict or not.

look at this pr https://github.com/rchain/rchain/pull/3447/files and the ci tests.

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
zsluedem: Ok. You made a good point. I remember why I ignore the test case. It is because I put a more strict rule for the test case -> https://github.com/rchain/rchain/blob/33d8a16d552762be05c973634521063144ba86fc/node/src/test/scala/coop/rchain/node/mergeablity/ComputeMerge.scala#L79-L83

when the deploy failed with error and raise the error directly. In this case, the deploy run into infinite loop which the deploy would fail with run out of phlo. And I notice that all other old InfiniteLoop tests are commented out -> https://github.com/rchain/rchain/blob/9cfb8954d5f44ccf8940ddb473268e9844e470b6/casper/src/test/scala/coop/rchain/casper/batch1/BaseMergeability.scala#L20-L22
 so I just ignore these tests.

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
nzpr: I believe they are different because `checkConflictAndMergeStatete` has `isConflict` field which is different. This is not checked as a test case, but printed when there is state mismatch, so you can see whether the problem is in wrong conflict detection.

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
```diff
@@ -1,16 +1,13 @@
-package coop.rchain.casper.batch1
+package coop.rchain.node.mergeablity
 
+import coop.rchain.node.mergeablity.OperationOn0Ch._
 import org.scalatest.{FlatSpec, Inspectors, Matchers}
 
-class JoinMergeability extends FlatSpec with Matchers with Inspectors with MergeabilityRules {
-  it should "J S N 2" in ConflictingCase(S0)(Nil)(J_)(J_.rstate ++ S0.rstate)(
```
zsluedem, line 6: dup with "J S N"

casper/src/test/scala/coop/rchain/casper/batch1/MergeabilityRules.scala
zsluedem: > This is not checked as a test case

why not check it?

zsluedem: In short, is it wrong to check it in the test?

I believe the tests are more robust with this additional check, aren't they?

And the result (in https://github.com/rchain/rchain/pull/3447/files) turns out to be that something is not as expected.

nzpr: I agree that this check makes sense, but are you saying that when you added 
```
 <- Sync[Effect]
                  .raiseError(new Exception(errMsg))
                  .whenA(rejectedDeployOpt.isDefined != isConflict)
```
this were failing?

I'm asking all this because I still don't understand what was wrong with regards to <https://github.com/rchain/rchain/pull/3430#discussion_r667340464>, trying to figure out why tests were passing when they should be not apparently?

zsluedem: I am sorry that I mislead you somewhere else here.

The comment here is actually not related to https://github.com/rchain/rchain/pull/3430#discussion_r667340464 ( I thought they were related.  My bad.)

I put another explanation on that comment below. Does that make sense? -> https://github.com/rchain/rchain/pull/3430#discussion_r667421501

node/src/test/scala/coop/rchain/node/mergeablity/JoinMergeability.scala
nzpr: I see, ok thanks for explanation

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #3433, opened 2021-06-12
Add transaction api

### LINKED ISSUE #3490: Remove old python typed transaction store and rewrite rev vault export
## Overview

The rev vault export for HF1 is using [python transaction server](https://github.com/zsluedem/transaction-server) to do that. But after https://github.com/rchain/rchain/pull/3433, the node would gain the ability like transaction server but even better than transaction server. 

After 3433 is merged, we would remove the old rev vault transaction export (including transaction store and etc.) and implement new rev vault transaction export.

### Impact

No impact
### Dependencies

No.
### Design

Only the transaction vault export would be changed. (StateBalance would stay the same.) The general logic would be the same as old implementation except getting data from new transaction store.

zsluedem:
## Overview

This pr is intended to provide a transaction api which currently only https://github.com/zsluedem/transaction-server provides.

#### Shortcoming in python transaction server.
1. only provide transaction in user deploy
2. people have to start another server and maintain additional infrastructure to support this function which it should be part of the node

This new transaction api solves the shortcoming above.
1. provide the transaction both in user deploys and system deploys
2. make node provide transaction function


node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
```diff
@@ -78,6 +86,104 @@ object JsonEncoder {
...
+                )
+                .flatMap(
+                  r =>
+                    Attempt.fromEither(
+                      r.leftMap(d => Err.General(d.message, List.empty))
+                        .map(DecodeResult(_, BitVector.empty))
+                    )
+                )
+          )
+    }
```
nzpr: I'd suggest we take some time to make code more compact, e.g. this can be rewritten as, and imo much more easy to understand
```scala
override def decode(bits: BitVector): Attempt[DecodeResult[A]] =
        for {
          string <- utf8.decode(bits).map(_.value)
          json <- Attempt.fromEither(parse(string).leftMap { e =>
                   Err.General(e.message, e.underlying.getStackTrace.map(_.toString).toList)
                 })
          result <- Attempt.fromEither(
                      decoder
                        .decodeJson(json)
                        .map(DecodeResult(_, BitVector.empty))
                        .leftMap(d => Err.General(d.message, List.empty))
                    )
        } yield result
```

zsluedem: https://github.com/rchain/rchain/pull/3454#discussion_r660026071
@tgrospic I am not sure if this is applicable here. 

But I agree with you @nzpr .

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,235 @@
...
+      ts <- report match {
+             case Right(b) =>
+               for {
+                 udts <- b.deploys.toList.traverse(d => {
+                          d.report.zipWithIndex.toList.traverse {
+                            case (report, index) => findUserDeployTransaction(d, report, index)
+                          }
+                        })
+                 userTransactions = udts.flatten.flatten
+                 systemDeployTransactions = b.systemDeploys.flatMap(
```
nzpr: I found it very handy to use curly brackets here so ` s =>` is in the same line as `b.systemDeploys.flatMap`, this way you save one LOC + line space inside map body. This is in many places.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,235 @@
...
+    convertCcodecToScodec(Encode.encodeTransactionResponse, Encode.decodeTransactionResponse)
+
+  def apply[F[_]: Sync: Concurrent](
+      blockReportAPI: BlockReportAPI[F],
+      // The transferUnforgeable can be retrieved based on the deployer and the timestamp of RevVault.rho
+      // in the genesis ceremony.
+      transferUnforgeable: Par
+  ): TransactionAPIImpl[F] = TransactionAPIImpl(blockReportAPI, transferUnforgeable)
+
+  def cacheTransactionAPI[F[_]: Sync: Concurrent](
```
tgrospic, line 248: Sync is redundant with Concurrent.

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
tgrospic: @zsluedem I'm not sure that relates.

But to create a SCodec it's easier to use helper functions then directly SCodec class constructor like here.
https://github.com/rchain/rchain/blob/7fa6000516b924494141b6c0970f4fb83435c9a7/shared/src/main/scala/coop/rchain/shared/Serialize.scala#L26-L36

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,235 @@
...
+sealed trait SystemTransaction                     extends TransactionType
+final case class CloseBlock(blockHash: String)     extends SystemTransaction
+final case class SlashingDeploy(blockHash: String) extends SystemTransaction
+
+final case class TransactionInfo(transaction: Transaction, transactionType: TransactionType)
+
+final case class TransactionResponse(data: Seq[TransactionInfo])
+
+trait TransactionAPI[F[_]] {
+  def getTransaction(blockHash: String): F[Seq[TransactionInfo]]
```
tgrospic: Can we make this block hash typed?

I know this relates to BlockReportAPI which also works with strings. Parsing error should be part of protocol layer.

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
```diff
@@ -78,6 +86,103 @@ object JsonEncoder {
...
+                    }
+                    .map(j => decoder.decodeJson(j))
+                )
+                .flatMap { r =>
+                  Attempt.fromEither(
+                    r.leftMap(d => Err.General(d.message, List.empty))
+                      .map(DecodeResult(_, BitVector.empty))
+                  )
+                }
+          )
```
tgrospic: To create a SCodec it's easier to use helper functions then directly SCodec class constructor like here.
https://github.com/rchain/rchain/blob/7fa6000516b924494141b6c0970f4fb83435c9a7/shared/src/main/scala/coop/rchain/shared/Serialize.scala#L26-L36

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
```diff
@@ -9,12 +9,20 @@ import coop.rchain.models.Expr.ExprInstance
 import coop.rchain.models.GUnforgeable.UnfInstance
 import coop.rchain.models.Var.{VarInstance, WildcardMsg}
 import coop.rchain.models._
+import scodec.SizeBound.unknown
+import scodec.bits.BitVector
+import io.circe.parser._
```
tgrospic, line 12: Why is circe parser needed. Is it because we have some JSON that is not covered by http4s-circe?
https://http4s.org/v0.19/json/

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,234 @@
...
+                     case SlashSystemDeployDataProto(_, _) =>
+                       transactions.map(
+                         t => TransactionInfo(t, SlashingDeploy(blockHash))
+                       )
+                     case CloseBlockSystemDeployDataProto() =>
+                       transactions.map(t => TransactionInfo(t, CloseBlock(blockHash)))
+                   }
+                 }
+                 transactions = userTransactions ++ systemDeployTransactions
+               } yield transactions
```
tgrospic: Let's make this more functional. Here are some hints for user deploy but similarly can be done for system also.
```scala
// Helper which accepts TransactionType constructor
private def makeDeployTransaction(
    d: DeployInfoWithEventData,
    report: SingleReport,
    txCtor: String => TransactionType
) = findTransactions(report).map(t => TransactionInfo(t, txCtor(d.deployInfo.get.sig)))

// List of TransactionType constructors that we want to zip with reports
val txCtors = Seq(PreCharge, UserDeploy, Refund)

// For Either, instead of matching directly on constructors we can use `map` and `getOrElse`
report map { b =>
  // ...
}
.getOrElse(Seq.empty)
```

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,234 @@
...
+                  val amount         = produce.data.get.pars(3).exprs.head.getGInt
+                  val retUnforgeable = produce.data.get.pars(5)
+                  val transaction    = Transaction(fromAddr, toAddr, amount, retUnforgeable, None)
+                  Some(acc :+ transaction)
+                } else None
+              }
+              .getOrElse(acc)
+
+          case _ => acc
+        }
```
tgrospic, line 94: This kind of collecting can be simplify without fold and with cleaner separation of what is collected and how is processed.
```scala
// First the collecting part
report.events
  .map(_.report.value)
  .collect {
    case ReportCommProto(Some(consume), produces) =>
      consume.channels.headOption.map((_, produces))
  }
  .flatten
  .flatMap {
    // And then when we have all what we need, comes processing part
    case (channel, produces) =>
      if (channel == transferUnforgeable) {
        // ...
        Seq(transaction)
      } else Seq.empty
  }
```
Or even step further.
```scala
report.events
  .map(_.report.value)
  .collect {
    case ReportCommProto(Some(consume), produces) =>
      consume.channels.headOption.map((_, produces))
  }
  .flatten
  .collect { case (channel, produces) if channel == transferUnforgeable => produces }
  .map { produces =>
    // ...
    transaction
  }
```

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,234 @@
...
+/**
+  * This API is totally based on how RevVault.rho is written. If the `RevVault.rho` is re-written or changed,
+  * this API might end up with useless.
+  */
+final case class TransactionAPIImpl[F[_]: Sync: Concurrent](
+    blockReportAPI: BlockReportAPI[F],
+    // The transferUnforgeable can be retrieved based on the deployer and the timestamp of RevVault.rho
+    // in the genesis ceremony.
+    transferUnforgeable: Par
+) extends TransactionAPI[F] {
```
tgrospic, line 63: Do we have tests for TransactionAPI?

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
zsluedem: To make https://github.com/rchain/rchain/pull/3433#discussion_r675856124 work for decoding.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
zsluedem: Nope. Let me add one.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
zsluedem: Isn't that going to iterate several times?

tgrospic: Please take a look on this comment https://github.com/rchain/rchain/pull/3361#pullrequestreview-622919165 related to multiple iteration and creation of intermediate structures.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,240 @@
...
+        val toAddr         = produce.data.get.pars(2).exprs.head.getGString
+        val amount         = produce.data.get.pars(3).exprs.head.getGInt
+        val retUnforgeable = produce.data.get.pars(5)
+        Transaction(fromAddr, toAddr, amount, retUnforgeable, None)
+      }
+
+    // traverse all the transactions to check if the transaction is completed or failed with reason
+    transactions.map { transaction =>
+      report.events
+        .find { p =>
```
tgrospic: This is another place where events will be iterated for each transaction.

In these kind of situations, first build the structure for quick access by the key you want to find (Map) or build a collection which you can join (like SQL inner join).

```scala
val failedMap = report.events.iterator
  .map(_.report.value)
  .collect {
    case ReportProduceProto(Some(channel), data) =>
      val success =
        data.get.pars.head.exprs.head.getETupleBody.ps.head.exprs.head.getGBool
      val failedReason =
        if (success) None
        else Some(data.get.pars.head.exprs.head.getETupleBody.ps(1).exprs.head.getGString)
      (channel, failedReason)
  }
  .toMap
```
This map can then be used directly when Transaction is constructed.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,240 @@
...
+          p.report.value match {
+            case ReportProduceProto(channel, _) =>
+              channel
+                .flatMap { c =>
+                  Some(c == transaction.retUnforgeable)
+                }
+                .getOrElse(false)
```
tgrospic: This can be simplified to `channel.contains(transaction.retUnforgeable)`.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,240 @@
...
+        val systemDeployTransactions = b.systemDeploys.flatMap { s =>
+          // system doesn't get precharge and refund , so it would always get one
+          val transactions = findTransactions(s.report.head)
+          s.systemDeploy.get.systemDeploy.value match {
+            case SlashSystemDeployDataProto(_, _) =>
+              transactions.map(
+                t => TransactionInfo(t, SlashingDeploy(blockHashStr))
+              )
+            case CloseBlockSystemDeployDataProto() =>
+              transactions.map(t => TransactionInfo(t, CloseBlock(blockHashStr)))
```
tgrospic: Here the same code is duplicated on both branches, only `SlashingDeploy` or `CloseBlock` are different. So by DRY principle we can first select what is different and have transactions transform specified only once.
```scala
val txCtor = s.systemDeploy.get.systemDeploy.value match {
  case SlashSystemDeployDataProto(_, _)  => SlashingDeploy
  case CloseBlockSystemDeployDataProto() => CloseBlock
}
transactions.map(t => TransactionInfo(t, txCtor(blockHashStr)))
```

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
tgrospic: I see, this is because we are storing JSON in KV store.

It would be much better if we have scodecs defined directly without going to JSON especially with fields specified as plain strings.
https://github.com/rchain/rchain/blob/1c4572f70da4109bd777b406e297cc5504d4a2de/node/src/main/scala/coop/rchain/node/revvaultexport/reporting/Transaction.scala#L71-L77

---

Do we need two TransactionStore types? Also keys should be typed and not just strings.
https://github.com/rchain/rchain/blob/945bbc9f03d21c077d6dc5922990d8237add9eb6/node/src/main/scala/coop/rchain/node/web/Transaction.scala#L179

https://github.com/rchain/rchain/blob/1c4572f70da4109bd777b406e297cc5504d4a2de/node/src/main/scala/coop/rchain/node/revvaultexport/reporting/Transaction.scala#L26

zsluedem: >Do we need two TransactionStore types? 

The second one is the python transaction server store. We will remove this one later. Let me create an issue.https://github.com/rchain/rchain/issues/3490.

> It would be much better if we have scodecs defined directly without going to JSON especially with fields specified as plain strings.

Actually I want it to be more like general encoding like json and protobuf which other languages can support.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
zsluedem: I wrote a lot(but deleted) to reply to you. But find that you are right. Haha. Good work.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,227 @@
...
+
+  def cacheTransactionAPI[F[_]: Sync: Concurrent](
+      transactionAPI: TransactionAPI[F],
+      kvm: KeyValueStoreManager[F]
+  ): F[CacheTransactionAPI[F]] =
+    kvm.database("transaction", utf8, transactionResponseCodec) >>= { s =>
+      CacheTransactionAPI(
+        transactionAPI,
+        s
+      ).pure
```
tgrospic: Why not just use map instead of fleatMap?

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -46,7 +48,15 @@ object ReportingRoutes {
 
     HttpRoutes.of[F] {
       case GET -> Root / "trace" :? BlockHashParam(hash) +& ForceReplayParam(forceReplay) =>
-        Ok(transforResult(hash, blockReportAPI.blockReport(hash, forceReplay.getOrElse(false))))
+        Ok(
+          transforResult(
+            hash,
+            blockReportAPI.blockReport(
+              ByteString.copyFrom(Base16.unsafeDecode(hash)),
```
tgrospic: ByteString is basically like String. Please create `QueryParamDecoderMatcher[Blake2b256Hash]` with corresponding `QueryParamDecoder`.
In that way we can validate user input and have friendly response error unlike _unsafeDecode_ which should be only used when we know error cannot happen.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,227 @@
...
+sealed trait SystemTransaction                     extends TransactionType
+final case class CloseBlock(blockHash: String)     extends SystemTransaction
+final case class SlashingDeploy(blockHash: String) extends SystemTransaction
+
+final case class TransactionInfo(transaction: Transaction, transactionType: TransactionType)
+
+final case class TransactionResponse(data: Seq[TransactionInfo])
+
+trait TransactionAPI[F[_]] {
+  def getTransaction(blockHash: BlockHash): F[Seq[TransactionInfo]]
```
tgrospic: This should be also `Blake2b256Hash`.

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
tgrospic: > The second one is the python transaction server store. We will remove this one later. Let me create an issue.#3490.

I see, this is basically to read externally created LMDB database.

> Actually I want it to be more like general encoding like json and protobuf which other languages can support.

Do you think that we need more general format, it's internal database used by RNode?

I'd be more for protobuf because JSON is slower. But protobuf requires textual proto files with code generation, which makes it much harder to maintain.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
zsluedem: Actually why `Blake2b256Hash`?

zsluedem: nvm

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
zsluedem: I fall back to scodec now

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
```diff
@@ -70,11 +70,11 @@ class BlockReportAPI[F[_]: Concurrent: Metrics: EngineCell: Log: SafetyOracle: B
                )
     } yield result
 
-  def blockReport(hash: String, forceReplay: Boolean): F[ApiErr[BlockEventInfo]] = {
+  def blockReport(hash: BlockHash, forceReplay: Boolean): F[ApiErr[BlockEventInfo]] = {
```
tgrospic, line 73: This is still untyped, String or ByteString has very little difference but it's ok to do this later.

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
```diff
@@ -254,7 +256,10 @@ object DeployGrpcServiceV1 {
 
       override def getEventByHash(request: ReportQuery): Task[EventInfoResponse] =
         defer(
-          blockReportAPI.blockReport(request.hash, request.forceReplay)
+          blockReportAPI.blockReport(
+            ByteString.copyFrom(Base16.unsafeDecode(request.hash)),
+            request.forceReplay
+          )
```
tgrospic: Just as a reminder. We need to fix error handling in APIs. Using unsafe here will not give friendly error to the API users.

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
```diff
@@ -78,6 +84,86 @@ object JsonEncoder {
...
+    Decoder.decodeUnit.map[AlwaysEqual[BitSet]](_ => AlwaysEqual(BitSet()))
+
+  // FIXME blake2b512Random decode. Question Is that really neccessary?
+  implicit val decodeBlake2b512Random: Decoder[Blake2b512Random] =
+    Decoder.decodeUnit.map[Blake2b512Random](_ => Blake2b512Random(1))
+
+  // convert the circe codec into scodec
+  def convertCcodecToScodec[A](encoder: Encoder[A], decoder: Decoder[A]): SCodec[A] =
+    utf8.exmap(
+      s =>
```
tgrospic: Please use curly braces for multi line code blocks. Code looks nicer without this extra line just for argument and arrow.

node/src/main/scala/coop/rchain/node/web/ReportingRoutes.scala
```diff
@@ -40,13 +40,20 @@ object ReportingRoutes {
         .withKebabCaseConstructorNames
         .withKebabCaseMemberNames
 
-    object BlockHashParam   extends QueryParamDecoderMatcher[String]("blockHash")
+    implicit val blake2b256HashParamDecoder =
+      QueryParamDecoder.stringQueryParamDecoder.map(Blake2b256Hash.fromHex)
```
tgrospic: This is not very friendly for users because `Blake2b256Hash.fromHex` uses unsafe and it will just throw exception for invalid hex.
We can also do this later as part of API redesign.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,257 @@
...
+          val transactions = findTransactions(s.report.head)
+          val txCtor = s.systemDeploy.get.systemDeploy.value match {
+            case SlashSystemDeployDataProto(_, _)  => SlashingDeploy
+            case CloseBlockSystemDeployDataProto() => CloseBlock
+          }
+          transactions.map(t => TransactionInfo(t, txCtor(Base16.encode(blockHash.bytes.toArray))))
+        }
+        (userTransactions ++ systemDeployTransactions).toList
+
+      case Left(_) => List.empty
```
tgrospic: In the spirit of tagless-final we don't need to pattern match on _tags_ and just use functions provided for each case.
```scala
// For Either, instead of matching directly on constructors we can use `map` and `getOrElse`
report map { b =>
  // ...
}
.getOrElse(Seq.empty)
```

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,257 @@
...
+      }
+      .flatten
+      .collect { case (channel, produces) if channel == transferUnforgeable => produces.head }
+      .map { produce =>
+        val fromAddr       = produce.data.get.pars.head.exprs.head.getGString
+        val toAddr         = produce.data.get.pars(2).exprs.head.getGString
+        val amount         = produce.data.get.pars(3).exprs.head.getGInt
+        val retUnforgeable = produce.data.get.pars(5)
+        Transaction(fromAddr, toAddr, amount, retUnforgeable, None)
+      }
```
tgrospic, line 113: To prevent intermediate structures created after each operation `report.events` should start with iterator `report.events.iterator` and ends up with materialization to concrete structure `toList`, `toVector`, etc. In this case all operation are _fused_ as one function and iteration will be done only once to produce concrete structure, similar how failed map is created.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  object SCodec {
+    import scodec._
+    import scodec.bits._
+    import scodec.codecs._
+    import coop.rchain.rholang.interpreter.storage._
+
+    val transactionCodec: Codec[Transaction] =
+      (utf8_32L :: utf8_32L :: long(63) :: serializePar.toSizeHeadCodec :: optional[String](
```
tgrospic: Why not use `uint64` for Long type instead of `long(63)`?

Also we are using big-endian format everywhere so maybe better not to mix will little-endian if not needed.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,257 @@
...
+  // This is not needed when onChain transfer history is implemented and deployed to new network in the future.
+  val transferUnforgeable = {
+    val seedForRevVault = Tools.unforgeableNameRng(
+      StandardDeploys.revVault.pk,
+      StandardDeploys.revVault.data.timestamp
+    )
+    // the 11th unforgeable name
+    val unfogeableBytes = (1 to 11).foldLeft(Array.empty[Byte]) {
+      case (_, _) => seedForRevVault.next()
+    }
```
tgrospic: I was thinking there is already some kind of support for Iterator and unfold but it seems it's a new stuff in 2.13.
https://github.com/scala/scala/pull/6851

But there is unfold on LazyList. But even that is too much, I found nice `continually` function on Iterator to just drop first 10 elements.
```scala
LazyList.unfold(Array.empty[Byte])(s => (seedForRevVault.next(), s).some).drop(10).head

// Even better without keeping intermediate values
Iterator.continually(seedForRevVault.next()).drop(10).next()
```

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
```diff
@@ -78,6 +84,86 @@ object JsonEncoder {
   implicit val encodeConnective: Encoder[Connective] = deriveEncoder[Connective]
   implicit val encodeAlwaysEqual: Encoder[AlwaysEqual[BitSet]] =
     Encoder.encodeUnit.contramap[AlwaysEqual[BitSet]](_ => ())
+
+  // FIXME blake2b512Random encode. Question Is that really neccessary?
```
tgrospic, line 88: We need this codec for definition but it does nothing in the resulting binary. Maybe to rename it as _dummy_?

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -130,6 +134,9 @@ object WebApi {
 
     def isFinalized(hash: String): F[Boolean] =
       BlockAPI.isFinalized(hash).flatMap(_.liftToBlockApiErr)
+
+    def getTransaction(hash: String): F[TransactionResponse] =
+      cacheTransactionAPI.getTransaction(hash)
```
tgrospic, line 139: Web API also needs to be typed and parsing errors should be moved to protocol level. Also task for API redesign.

node/src/main/scala/coop/rchain/node/encode/JsonEncoder.scala
zsluedem: Let's try to put this kind of format requirements into scala fmt. If there are nothing like this in scalafmt, let's be not picky.

zsluedem: https://github.com/scalameta/scalafmt/issues/1222
https://github.com/scalameta/scalafmt/issues/2019
find something related. It seems scalafmt didn't address this.

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
zsluedem: What do you mean "typed" here?

node/src/main/scala/coop/rchain/node/web/Transaction.scala
zsluedem: https://github.com/scodec/scodec/blob/95468730ae160f1598b47bf31da3aa65ede26afd/shared/src/main/scala/scodec/codecs/package.scala#L71

` long(63)` is big-endian

zsluedem: Oh. Sorry good catch . `utf8_32L` is little-endian

zsluedem: I should use `vlong`

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
tgrospic: I just mean something more specific/structured like Blake2b256Hash. E.g. it can validate raw input to be right length.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
tgrospic: Without converting to concrete structure like `toList` here, it will execute the whole composition from where the iterator is created.

First iteration is here. To existing iterator `transactions`, another map will be added to create a Set.
```scala
transactions.map(_.retUnforgeable).toSet
```

Second iteration is here. To existing iterator `transactions` another map will be added to produce a List.
```scala
transactions.map { t =>
  val failReason = failedMap.getOrElse(t.retUnforgeable, None)
  t.copy(failReason = failReason)
}.toList
```

Because iterators are lazy, they will run only when converted to concrete collection type like `toList` (Scala 3 `to(List)`). This is the same concept used in fs2.Stream which is run with `to(List)` or `drain`.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  object SCodec {
+    import scodec._
+    import scodec.bits._
+    import scodec.codecs._
+    import coop.rchain.rholang.interpreter.storage._
+
+    val transactionCodec: Codec[Transaction] =
+      (utf8_32 :: utf8_32 :: vlong :: serializePar.toSizeHeadCodec :: optional[String](
```
tgrospic, line 187: `vlong` nice! :+1:

tgrospic (approved):
Nice work and thanks to addressing my comments!

It would be great that we had this before. Maybe we can support pre HF1 state, so that the whole transaction history can be rebuild without transaction server.

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #3434, opened 2021-06-12
RSpace and Rholang project cleanup

tgrospic:
## Overview

The goal of this PR is to add more clarity to Rholang reducer and normalizer code. It removes implicit conversions of RSpace result types to be easier to understand.

**NOTE: It does not change any functionality.**

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
@@ -58,7 +58,6 @@ class DebruijnInterpreter[M[_]: Sync: Parallel: _cost](
     * @param chan  The channel on which data is being sent.
     * @param data  The par objects holding the processes being sent.
     * @param persistent  True if the write should remain in the tuplespace indefinitely.
-    * @return  An optional continuation resulting from a match in the tuplespace.
     */
   private def produce(
       chan: Par,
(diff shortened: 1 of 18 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -41,14 +41,10 @@ trait Reduce[M[_]] {
 
 }
 
-class DebruijnInterpreter[M[_]](
+class DebruijnInterpreter[M[_]: Sync: Parallel: _cost](
```
nzpr, line 44: I'm just curious - why such unusual name for type class? `_cost`

nzpr (approved):
LGTM

zsluedem (approved):
LGTM

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
tgrospic: This is very good question.

We also have `_error`. Maybe because both are aliases.
https://github.com/rchain/rchain/blob/95ac99c4c991df111c3d7278827758a7e3607997/rholang/src/main/scala/coop/rchain/rholang/interpreter/package.scala#L7

I tried to replace `_error` with `Sync but it's not that easy, it seems that MonadState replaced with Ref in _cost can simplify it. I was playing with Ref already https://github.com/tgrospic/rchain/pull/12.

tgrospic:
bors merge

PR #3439, opened 2021-06-16
Per block dag view

### LINKED ISSUE #3350: Modify the DAG representation to support block merge

nzpr:
## Overview
When concurrent block processing is enabled in the network, local view of the validating node might be different then view of block creator (there might be more messages processed). Therefore to be able to validate blocks, Casper view should be reconstructed according to justifications of a block being validated.
Details described in Issue https://github.com/rchain/rchain/issues/3350.

### Notes
This branch is running well against 0.67 sync constraint, but very quickly fails with well known "Bug found: refund failed" when multiple blocks from the same validator are in merging scope. So further testing is blocked until merging if REV balances is implemented.

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -55,7 +55,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       (blockHash.size == BlockHash.Length && dagSet.contains(blockHash)).pure[F]
 
     def children(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
-      childMap.get(blockHash).map(_ intersect dagSet).pure[F]
+      childMap.get(blockHash).pure[F]
 
     def latestMessageHash(validator: Validator): F[Option[BlockHash]] =
       latestMessagesMap.get(validator).pure[F]
@@ -71,46 +71,68 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
         targetLatestMessages: Map[Validator, BlockHash],
         findLfb: Map[Validator, BlockHash] => F[BlockHash]
     ): F[BlockDagRepresentation[F]] = {
+
       val lmAll       = this.latestMessagesMap.values.toSet
-      val lmSeen      = targetLatestMessages.values.toSet
       val seenSenders = targetLatestMessages.keySet
 
       for {
-        lfb             <- findLfb(targetLatestMessages)
-        latestFinalized <- this.latestFinalized(lfb, seenSenders).map(_.valuesIterator.toSet)
+        lfb         <- findLfb(targetLatestMessages)
+        lmFinalized <- this.latestFinalized(lfb, seenSenders).map(_.valuesIterator.toSet)
+        lmSeen      <- targetLatestMessages.values.toList.traverse(this.lookupUnsafe).map(_.toSet)
+
         // for all known latest messages, collect all self justifications until first finalized message found
-        toAdjust <- Stream
-                     .fromIterator(lmAll.iterator)
-                     .flatMap(
-                       this
-                       // message + all self justifications
-                         .fromSenderBelowWith(_, m => (m.blockHash, m.blockNum))
-                         // take while first finalized message is found
-                         .takeWhile {
-                           case (hash, _) => !latestFinalized.contains(hash)
-                         }
-                         .mapAccumulate(init = false) {
-                           case (childIsSeen, (hash, height)) =>
-                             val seen = childIsSeen || lmSeen.contains(hash)
-                             (seen, (hash, height))
-                         }
-                     )
-                     .compile
-                     .toList
+        toAdjust = {
+          val seenSeqNums = lmSeen.map(m => (m.sender, m.seqNum)).toMap
+          val finSeqNums  = lmFinalized.map(m => (m.sender, m.seqNum)).toMap
+          val seen        = (m: BlockMetadata) => seenSeqNums.get(m.sender).exists(_ >= m.seqNum)
+          val finalized   = (m: BlockMetadata) => finSeqNums.get(m.sender).exists(_ >= m.seqNum)
 
-        (u, r) = toAdjust.partition { case (seen, _) => seen }
+          Stream
+            .unfoldLoopEval(lmAll.toVector) { messages =>
+              for {
+                metas                    <- messages.traverse(this.lookupUnsafe).map(_.filterNot(finalized))
+                (toUnfinalize, toRemove) = metas.partition(seen)
 
-        toUnfinalize = u.map { case (_, (hash, height)) => (hash, height) }
-        toRemove     = r.map { case (_, (hash, height)) => (hash, height) }
+                parents = metas.flatMap(_.parents).distinct
+                next    = parents.nonEmpty.guard[Option].as(parents)
+                out     = toRemove.map((_, true)) ++ toUnfinalize.map((_, false))
+              } yield (Stream.emits(out.toList), next)
+            }
+            .flatten
+            .fold((Map.empty[Long, Set[BlockHash]], Map.empty[Long, Set[BlockHash]], childMap)) {
+              case ((removeAcc, unfinalizeAcc, childMap), (m, shouldRemove)) =>
+                val height = m.blockNum
+                if (shouldRemove)
+                  (
+                    removeAcc + (height -> (removeAcc.getOrElse(height, Set()) + m.blockHash)),
+                    unfinalizeAcc,
+                    // Remove from children map key for message, adjust keys for parents.
+                    // As some parent might be already removed by previous iterations, use filter.
+                    m.parents.filter(childMap.contains).foldLeft(childMap - m.blockHash) { (a, p) =>
+                      a + (p -> (a(p) - m.blockHash))
+                    }
+                  )
+                else
+                  (
+                    removeAcc,
+                    unfinalizeAcc + (height -> (unfinalizeAcc.getOrElse(height, Set()) + m.blockHash)),
+                    childMap
+                  )
+            }
+            .compile
+            .lastOrError
+        }
 
-        excessSet       = toRemove.map { case (hash, _)     => hash }.toSet
-        nonFinalizedSet = toUnfinalize.map { case (hash, _) => hash }.toSet
+        r                                        <- toAdjust
+        (toRemove, toUnfinalize, newChildrenMap) = r
+        excessSet                                = toRemove.flatMap { case (_, hashes) => hashes }.toSet
+        nonFinalizedSet                          = toUnfinalize.flatMap { case (_, hashes) => hashes }.toSet
 
         // new truncated values
         truncatedDagSet     = this.dagSet diff excessSet
         truncatedInvalidSet = this.invalidBlocksSet.filter(m => dagSet.contains(m.blockHash))
         truncatedHeightMap = toRemove.foldLeft(this.heightMap) {
-          case (acc, (hash, height)) => acc.updated(height, acc(height) - hash)
+          case (acc, (height, hashes)) => acc.updated(height, acc(height) -- hashes)
         }
         truncatedFinalizesSet = finalizedBlocksSet diff excessSet diff nonFinalizedSet
 
@@ -118,6 +140,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
           lastFinalizedBlockHash = lfb,
           finalizedBlocksSet = truncatedFinalizesSet,
           dagSet = truncatedDagSet,
+          childMap = newChildrenMap,
           latestMessagesMap = targetLatestMessages,
           heightMap = truncatedHeightMap,
           invalidBlocksSet = truncatedInvalidSet
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -20,6 +20,7 @@ import coop.rchain.models.EquivocationRecord.SequenceNumber
 import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
 import coop.rchain.shared.syntax._
+import coop.rchain.models.syntax._
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 import fs2.Stream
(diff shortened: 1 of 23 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,49 +54,107 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
 
     // latestBlockNumber, topoSort and lookupByDeployId are only used in BlockAPI.
     // Do they need to be part of the DAG current state or they can be moved to DAG storage directly?
 
-    private def getMaxHeight = if (heightMap.nonEmpty) heightMap.last._1 + 1L else 0L
+    private def getMaxHeight: F[Long] =
+      latestMessagesMap.values.toList
+        .traverse(lookup)
+        .map(_.map(_.get.blockNum))
+        .map(heights => if (heights.nonEmpty) heights.max else 0L)
```
tgrospic: The whole point of indices in DAG representation is not do this kind of calculation on every access to index.

Why not just remove all hashes from height map that are _older_ of latest messages, together with hashes from validators that are not in latest messages?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,49 +54,112 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
-        )
+    ): F[Vector[Vector[BlockHash]]] =
+      getMaxHeight.flatMap { maxNumber =>
+        val startNumber = Math.max(0, startBlockNumber)
+        val endNumber   = maybeEndBlockNumber.map(Math.min(maxNumber, _)).getOrElse(maxNumber)
+        if (startNumber >= 0 && startNumber <= endNumber) {
+          Sync[F].delay(
+            heightMap
+              .filterKeys(h => h >= startNumber && h <= endNumber)
+              .map { case (_, v) => v.toVector.filter(dagSet.contains) }
```
tgrospic: Filtering of hashes in height map can be done as part of truncate.

nzpr: We need to filter the whole height map, e.g. size of a million, on each DAG representation calculation. And here we only need to filter small part of this map which is of interest. Do you think it worth it to filter the whole map?

Well, we can do that in truncate, but logic looks to me more complex. Let me think about it

tgrospic: We have block heights of latest messages so this can be used to restrict going to far in the past.

nzpr: I implemented truncate by only heights, and still not fully convinced - are you sure that if message A has height lower then lowest latest message - this A is always in view?

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
```diff
@@ -558,4 +558,51 @@ class MultiParentCasperAddBlockSpec extends FlatSpec with Matchers with Inspecto
...
+          _  = b1.header.parentsHashList.size shouldBe 1
+          _  = b2.header.parentsHashList.size shouldBe 1
+          //         b3 b4
+          //   b1 b2
+          //b0
+          b3 <- n3.createBlockUnsafe(deploys(3))
+          b4 <- n4.createBlockUnsafe(deploys(4))
+          _  = (b3.body.state.preStateHash == b4.body.state.preStateHash) shouldBe true
+          _  <- nodes.toList.traverse(_.processBlock(b3))
+          // after b3 processed by all blocks, b4 processing should not be influenced by b3 (b3 should not be used as parent)
```
nzpr, line 591: this particular line is about per block view. I'm thinking how to write better tests, but having KeyValueDagRepresentation private class of BlockDagKeyValueStorage makes things not easy, so the easiest place to test this is in Casper in AddBlockSpec, which has way more context. Does it makes sense to just make it public? @tgrospic

tgrospic: Makes sense, but I'm not sure that public access will be enough.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,14 +53,71 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+      childMap.get(blockHash).map(_ intersect dagSet).pure[F]
 
     def latestMessageHash(validator: Validator): F[Option[BlockHash]] =
       latestMessagesMap.get(validator).pure[F]
 
     def latestMessageHashes: F[Map[Validator, BlockHash]] = latestMessagesMap.pure[F]
 
-    def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
+    def invalidBlocks: F[Set[BlockMetadata]] =
+      invalidBlocksSet.filter(m => dagSet.contains(m.blockHash)).pure[F]
```
tgrospic: Why not filter child map and invalid blocks as part of truncate?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,14 +53,71 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+    ): F[BlockDagRepresentation[F]] = {
+      final case class HeightView(
+          height: Long,
+          excess: Iterable[ByteString],
+          inView: Iterable[ByteString]
+      )
+      for {
+        lmMetas <- latestMessages.values.toStream
+                    .traverse(lookup)
+                    .map(_.map(_.get).toSet)
```
tgrospic: You can use _unsafe_ lookup here. Please avoid using _.get_ on an Option type directly because becomes the same as using _null_.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,14 +53,71 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+        // messages inside range of latest message heights might contain new messages from validator that are not seen yet
+        lmRangeViews <- heightMap
+                         .range(lowestHeight, highestHeight)
+                         .toList
+                         .traverse {
+                           case (h, messages) =>
+                             for {
+                               lms <- messages.toList
+                                       .traverse(lookup)
+                                       .map(_.map(_.map(m => (m.blockHash, m.sender)).get))
```
tgrospic: Same here with Option type.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: for childMap I'm not sure how to do this effectively and reliably, e.g. by analysing DAG. It is `Map[BlockHash, Set[BlockHash]]` and I need to examine value for each key. For invalidSet it makes sense and simple, thanks will adjust.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -53,14 +53,71 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+
+                             } yield r
+                         }
+        // messages below lowest height across input latest messages are always in view,
+        // as it was decided so the whole excess is this sum
+        excessMessages  = highExcess.values.flatten ++ lmRangeViews.flatMap(_.excess)
+        truncatedDagSet = this.dagSet diff excessMessages.toSet
+
+        truncatedHeightMap = lmRangeViews.foldLeft(withoutHighExcess)(
+          (acc, heightView) => acc.updated(heightView.height, heightView.inView.toSet)
```
tgrospic: `heightView.inView` is an iterator and here it's iterated for each element in _lmRangeViews_. Think always how to minimize iteration over collections you are working with.

Storing _iterator_ is usually not what you want, except the only thing in which will be used is single iteration which cannot be avoided. If you need multiple iterations make sense to choose some concrete collection.

nzpr: @tgrospic It's `Iterable`, not `Iterator`. So this means that this is just a collection, right?

tgrospic: Yes, you are right I think of Iterator but Iterable has only one method to provide Iterator.

Iterator is basically these two functions. This means the only way to do anything with this "collection" is to get next element (maybe all).
https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/Iterator.scala#L359-L380
```scala
trait Iterator[+A] extends TraversableOnce[A] {
  /** Tests whether this iterator can provide another element.
   *
   *  @return  `true` if a subsequent call to `next` will yield an element,
   *           `false` otherwise.
   *  @note    Reuse: $preservesIterator
   */
  def hasNext: Boolean

  /** Produces the next element of this iterator.
   *
   *  @return  the next element of this iterator, if `hasNext` is `true`,
   *           undefined behavior otherwise.
   *  @note    Reuse: $preservesIterator
   */
  def next(): A
```

So the question is how to choose the right type of collection or better to say what kind of interface should provide this collection. The main thing we want to consider is performance (do only necessary work) but also that code is easy to understand which will reduce bugs.
**For collections better performance === less iterations.** 

For example
`inView = inView.map { case (hash, _) => hash }`
when executed here has one iteration to extract `Iterable[ByteString]` so when Set is needed here
`(acc, heightView) => acc.updated(heightView.height, heightView.inView.toSet)`
it needs to iterate once again to create a Set. This is done with builder implementation which uses TraversableOnce created from iterator.
https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/immutable/Set.scala#L372-L384

So here, more precise type for `inView` would be a Set because this is what is needed at the end when is added to height map. So when inView is created it will be converted to Set.
But even this will still be the same. :) `map` will iterate and toSet will iterate.
```scala
// inView: Set[ByteString]
inView = inView.map { case (hash, _) => hash }.toSet
```
So to make it *lazy*, first we need to convert it to iterator which will be used by map and toSet for each element only once.
```scala
// inView: Set[ByteString]
inView = inView.toIterator.map { case (hash, _) => hash }.toSet
```
---

Another view can be, why not use iterator as a type for `inView` and have the same effect. So here mapping will not happen but only when toSet is called later. This is also correct but I'd say little harder to understand and follow.
```scala
// inView: Iterator[ByteString]
inView = inView.toIterator.map { case (hash, _) => hash }
```
---

This is implementation of `Iterator#map` function. It doesn't do nothing except changing the get next function with the wrapper, supplied mapping function.
https://github.com/scala/scala/blob/2.12.x/src/library/scala/collection/Iterator.scala#L459-L462
```scala
def map[B](f: A => B): Iterator[B] = new AbstractIterator[B] {
  def hasNext = self.hasNext
  def next() = f(self.next())
}
```

nzpr:
I target this on dev because, although this is required for leaderless merge, on merging with leader absence of this might lead to replay failure under certain conditions (e.g. when after nodes are restarted and added more blocks then they are supposed to).

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +63,130 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+        excessMessages = lmRangeViews.flatMap(_.excess) ++ highExcess.values.flatten
+
+        // check clause 3
+        fullLMMetas <- this.latestMessages
+        // filtering out genesis is required because there might be some validators bonded in genesis
+        // which are not yet in view. Its not quite pretty, but no other options.
+        notSeenSenders = fullLMMetas.filter { case (_, meta) => meta.blockNum != 0 }.keySet diff targetLatestMessages.keySet
+        lowExcess <- Stream
+                    // gather all messages of unseen validators
+                      .fromIterator(notSeenSenders.toIterator)
```
nzpr, line 152: This particular part is something that I'm not completely sure yet. This is supposed to remove messages that are not seen and are lower than lowest target latest message.

block-storage/src/test/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorageTest.scala
```diff
@@ -338,4 +339,127 @@ class BlockDagKeyValueStorageTest extends BlockDagStorageTest {
...
+    * O2 and O3 are high excess
+    * O1 is range excess
+    * O4 is unseenSender excess - some validator bonded in genesis has been waiting for a while before creating new block.
+    *   And created on top of genesis without referencing all new messages.
+    *
+    *   03
+    *   02     T
+    *   L1 01 L2
+    *      L0
+    *   B0 B1 B2 O4
```
nzpr, line 357: @tgrospic we've been talking about possibility of messages below lowest target LM to be outside if the view. I still this this is possible, O4 block is an example. I think it's not impossible scenario. Therefore I've added this check ti truncate
https://github.com/rchain/rchain/blob/4725a25a9e89d73dc9aa0dfd03e488d332e77479/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala#L145-L171

nzpr, line 357: @tgrospic we've been talking about possibility of messages below lowest target LM to be outside the view. I still not sure this is impossible, O4 block is an example. Therefore I've added this check to truncate
https://github.com/rchain/rchain/blob/4725a25a9e89d73dc9aa0dfd03e488d332e77479/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala#L145-L171

nzpr, line 353: @tgrospic we've been talking about possibility of messages below lowest target LM to be outside the view. I still not sure this is impossible, O4 block is an example. Therefore I've added this check to truncate
https://github.com/rchain/rchain/blob/4725a25a9e89d73dc9aa0dfd03e488d332e77479/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala#L145-L171

nzpr:
When testing with leaderless merge (thanks to @tgrospic for a workaround to enable that) I figured out that finalised set has to be adjusted as well on truncate, will push commit soon.

nzpr:
The commit above is disabling check for sender. It is required to handle this case down the pipe as this is valid possible scenario and a slashable offence.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -133,20 +145,7 @@ final class BlockDagRepresentationOps[F[_]](
...
-              for {
-                out  <- lvl.filterA(dag.isFinalized(_).not)
-                next <- out.traverse(dag.lookup(_).map(_.map(_.parents))).map(_.flatten.flatten)
-              } yield (out, next.nonEmpty.guard[Option].as(next))
-            }
-            .flatMap(Stream.emits)
-            .compile
-            .to(Set)
-    } yield r
+  def nonFinalizedBlocks(implicit sync: Sync[F]): F[Set[BlockHash]] = dag.nonFinalizedSet.pure
```
tgrospic: Why is this extension needed?

block-storage/src/test/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorageTest.scala
tgrospic: So these messages should be with higher sequence number then LMs. They are like attached sub-dag created after LMs from which we want to truncate.

Maybe we can use sequence number to cut off any newer messages. Traversing from LMs until oldest LM, collecting excess children, building part of new child map, building part of height map and unfinalized set.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
nzpr: Its a leftover, I'll remove it

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -55,7 +55,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       (blockHash.size == BlockHash.Length && dagSet.contains(blockHash)).pure[F]
 
     def children(blockHash: BlockHash): F[Option[Set[BlockHash]]] =
-      childMap.get(blockHash).map(_ intersect dagSet).pure[F]
```
nzpr: with the above change the whole state is truncated, children map as well

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +64,90 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+        excessSet                                = toRemove.flatMap { case (_, hashes) => hashes }.toSet
+        nonFinalizedSet                          = toUnfinalize.flatMap { case (_, hashes) => hashes }.toSet
+
+        // new truncated values
+        truncatedDagSet     = this.dagSet diff excessSet
+        truncatedInvalidSet = this.invalidBlocksSet.filter(m => dagSet.contains(m.blockHash))
+        truncatedHeightMap = toRemove.foldLeft(this.heightMap) {
+          case (acc, (height, hashes)) => acc.updated(height, acc(height) -- hashes)
+        }
+        truncatedFinalizesSet = finalizedBlocksSet diff excessSet diff nonFinalizedSet
```
tgrospic, line 136: I know it's the same but you are using `--` and `diff` with sets. I'm ok with both but if it's not mixed in the same scope.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +64,90 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+          val seenSeqNums = lmSeen.map(m => (m.sender, m.seqNum)).toMap
+          val finSeqNums  = lmFinalized.map(m => (m.sender, m.seqNum)).toMap
+          val seen        = (m: BlockMetadata) => seenSeqNums.get(m.sender).exists(_ >= m.seqNum)
+          val finalized   = (m: BlockMetadata) => finSeqNums.get(m.sender).exists(_ >= m.seqNum)
+
+          Stream
+            .unfoldLoopEval(lmAll.toVector) { messages =>
+              for {
+                metas                    <- messages.traverse(this.lookupUnsafe).map(_.filterNot(finalized))
+                (toUnfinalize, toRemove) = metas.partition(seen)
```
tgrospic: Why not just one mapping to final type?
```scala
out = metas.map(m => (m, !seen(m)))
```

P.S. I'd move this line here because it doesn't depend on `parents` and `next`.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +64,90 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+
+                parents = metas.flatMap(_.parents).distinct
+                next    = parents.nonEmpty.guard[Option].as(parents)
+                out     = toRemove.map((_, true)) ++ toUnfinalize.map((_, false))
+              } yield (Stream.emits(out.toList), next)
+            }
+            .flatten
+            .fold((Map.empty[Long, Set[BlockHash]], Map.empty[Long, Set[BlockHash]], childMap)) {
+              case ((removeAcc, unfinalizeAcc, childMap), (m, shouldRemove)) =>
```
tgrospic: Please `childMapAcc` :pray:

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +64,90 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+                    m.parents.filter(childMap.contains).foldLeft(childMap - m.blockHash) { (a, p) =>
+                      a + (p -> (a(p) - m.blockHash))
+                    }
+                  )
+                else
+                  (
+                    removeAcc,
+                    unfinalizeAcc + (height -> (unfinalizeAcc.getOrElse(height, Set()) + m.blockHash)),
+                    childMap
+                  )
```
tgrospic, line 119: I think we need to traverse into `children` for each parent to find newer blocks not seen by parent relation.
```
      c2 (a2, b2, c1) <- full dag
a3 b4             <- latest (a3, b4, c1)
   b3
a2 b2   
a1 b1 c1
```
`c2` should not be visible in truncated dag. It has parents (a2, b2, c1) which means child map for a1 and b1 will contain c2.

nzpr: But this piece of code is for unfinalized messages. 
If msg should be removed it is 
```scala
m.parents.filter(childMap.contains).foldLeft(childMap - m.blockHash) { (a, p) =>
                      a + (p -> (a(p) - m.blockHash))
                    }
```
as per several lines above

tgrospic: Yes, my comment is for remove case.

My concern is not for keys in the child map but for values which could have children not visible from the target latest messages.
I'm not sure how we will remove `c2` from children of `a1`, `b1` or `c1` in child map?

nzpr: c2 has parents `(a2, b2, c1)`, so for each of them when removing c2 `a + (p -> (a(p) - m.blockHash))` will be invoked? I must miss some point, Im not sure which one...

tgrospic: When we traverse from target latest blocks (a3, b4, c1) we will not see c2.

nzpr: But we traverse from the most recent latest blocks, so from `a3 b4 c2`
https://github.com/rchain/rchain/blob/c093633ccc19bd6a1e486cf07becb301163a27f9/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala#L90-L92
`val lmAll       = this.latestMessagesMap.values.toSet`
We traverse down from the most recent view, collecting what we should remove and what unfinalize.

tgrospic: OK, that makes sense. Is this means we will rebuild childs for a1, b1 and c1 which has c2 as a child?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
```diff
@@ -98,6 +98,18 @@ final class BlockDagRepresentationOps[F[_]](
...
+      implicit sync: Sync[F]
+  ): Stream[F, A] =
+    Stream.unfoldLoopEval(start)(
+      message =>
+        lookupUnsafe(message)
+          .map { meta =>
+            val selfJustification = meta.justifications.find(_.validator == meta.sender)
+            (proj(meta), selfJustification.map(_.latestBlockHash))
+          }
+    )
```
tgrospic: Is this extension used somewhere?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Nice catch!

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagRepresentationSyntax.scala
nzpr: No, will remove

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: I'm not sure - why we need to rebuild? When `c2` is removed - we remove `c2` key from children map (in this particular case there is no such key as no children), and update values for keys `a2,b2,c1`. Same is for every other message - remove key and update keys for parents.

tgrospic: I see, we will update childs for a1, b1 and c1 via parents of c2. This is what I mean rebuild.

tgrospic (approved):
Excellent work! Thanks for accepting all comments.

I'm happy that we finally have DAG of DAGs.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: I'm still not sure where is the problem :) if we are removing some hash - it should be removed as a parent, so key completely removed, plus we need to update records for all parents

PR #3443, opened 2021-06-17
[TEST] Rejection test

tgrospic:


casper/src/test/scala/coop/rchain/casper/wip/MergeRejectionSpec.scala
```diff
@@ -0,0 +1,147 @@
...
+        "All blocks      " -> node0Blocks,
+        "Parents         " -> mergedParents,
+        "Deploys         " -> deploys.keys,
+        "Rejected deploys" -> rejectedDeploys
+      )
+      _ = println(s"RESULT:\n${res.mkString("\n")}")
+    } yield ()
+  }
+
+  it should "reject deploys built on top of rejected deploy (separate blocks)" in effectTest {
```
nzpr, line 181: Do you think this test is not enough for such case (missing piece here is to test if depending deploy indices go into the same branch but this can be done as well)?  https://github.com/rchain/rchain/blob/ae2eb19695b43952ad0b64be0eb78b5245e2f44b/casper/src/test/scala/coop/rchain/casper/merging/BlockIndexSpec.scala#L176-L182

PR #3444, opened 2021-06-18
Fix rejection options bug. Refactor.

nzpr:
## Overview
https://github.com/rchain/rchain/pull/3443 revealed bug. This PR fixes it. Also some changes from  which are not related to tests from https://github.com/rchain/rchain/pull/3425 added here. PTAL at comments to know more.

### Notes
Bug with rejection options calculation were masking issue https://github.com/rchain/rchain/issues/3349. Now it is observable again, the issue is reopened.


rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
```diff
@@ -48,7 +48,11 @@ package object merger {
       }
     }
     // each rejection option is defined by decision not to reject a key in rejection map
-    conflictMap.keySet.map(k => process(conflictMap(k), Set.empty, reject = true))
+    conflictMap
+    // only keys that have conflicts associated should be examined
+      .filter { case (_, conflicts) => conflicts.nonEmpty }
+      .keySet
+      .map(k => process(conflictMap(k), Set.empty, reject = true))
```
nzpr: Without this filter rejection option to not reject anything is present in output in case there is a single branch that does not conflict with anything.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -73,29 +73,36 @@ object MergingLogic {
     (e.producesLinear ++ e.producesPersistent) diff e.producesCopiedByPeek
 
   /** consume created inside event log */
-  def consumesCreated(e: EventLogIndex): Set[Consume] = e.consumesLinearAndPeeks
+  def consumesCreated(e: EventLogIndex): Set[Consume] =
+    e.consumesLinearAndPeeks ++ e.consumesPersistent
```
nzpr, line 78: This is another bug fixed - persistent consumes were not accounted

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -73,29 +73,36 @@ object MergingLogic {
...
   def consumesAffected(e: EventLogIndex): Set[Consume] = {
     def externalConsumesDestroyed(e: EventLogIndex): Set[Consume] =
       e.consumesProduced diff consumesCreated(e)
 
-    consumesActive(e) ++ externalConsumesDestroyed(e)
+    consumesCreatedAndNotDestroyed(e) ++ externalConsumesDestroyed(e)
   }
+
+  /** if produce is copied by peek in one index and originated in another - it is considered as created in aggregate*/
+  def combineProducesCopiedByPeek(x: EventLogIndex, y: EventLogIndex): Set[Produce] =
```
nzpr, line 105: This is important function so moved to merging logic file

rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
nzpr: Given this change, issue <https://github.com/rchain/rchain/issues/3349> is present again. Previously it was masked by this bug.

casper/src/test/scala/coop/rchain/casper/util/rholang/MergingBranchMergerSpec.scala
```diff
@@ -29,7 +29,7 @@ import scala.collection.Seq
 
 class MergingBranchMergerSpec extends FlatSpec with Matchers {
 
-  val genesisContext             = GenesisBuilder.buildGenesis(validatorsNum = 50)
+  val genesisContext             = GenesisBuilder.buildGenesis(validatorsNum = 5)
```
nzpr, line 32: This change is to mitigate issue <https://github.com/rchain/rchain/issues/3349> of false positive conflict detection. Once it is closed, num of validators should be increased

rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
tgrospic: Why is this here and not in merging logic file?

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -48,8 +48,8 @@ object MergingLogic {
 
       // Search for match
       def check(left: EventLogIndex, right: EventLogIndex): Iterator[Blake2b256Hash] = {
-        val p = producesActive(left)
-        val c = consumesActive(right)
+        val p = producesCreatedAndNotDestroyed(left)
+        val c = consumesCreatedAndNotDestroyed(right)
```
tgrospic, line 53: This seems to be calculated multiple times, why not put this as part of event log index?

nzpr: when you combine two EventLogIndex, you need to recalculate. Because produce created in one of them might be destroyed in another. So in join index it won't be in this set, but it will be in one where it is created

rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
nzpr: I'm not sure, Its sort of general operations on Maps. Do you think the whole package content should be in merging logic?

nzpr: Actually might be makes sense to put this particular function to merging logic

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
tgrospic: I'm not saying for joined index but for one operation of merging this is calculated multiple times on the same event log index. I see that `producesCreatedAndNotDestroyed` will be called two times on the same event log index.

https://github.com/rchain/rchain/blob/3cadf0e801a9978adc9f861eb601750fe55f9172/casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala#L38

https://github.com/rchain/rchain/blob/3cadf0e801a9978adc9f861eb601750fe55f9172/casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala#L58

I'm not sure how much this impacts performance but there is a lot of diffs.

rspace/src/main/scala/coop/rchain/rspace/merger/package.scala
nzpr: https://github.com/rchain/rchain/pull/3444/commits/3c058c47f0c17d3d843be0102ba2e6ff0170cdec

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
nzpr: This is diff and append on sets, so I believe should be fast. We might cache if it will be slow

tgrospic (approved):
LGTM

Good catch of the other bug!

Issue #3268, opened 2020-12-15
Peek should be atomic


nzpr:
Consume and produce inside a peek should be executed inside a single RSpace lock.
Now first consume executed
https://github.com/rchain/rchain/blob/6dd06305020f02f9fed99c7eda6cbd980fdb0b7d/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala#L94-L101
 and then produce <https://github.com/rchain/rchain/blob/6dd06305020f02f9fed99c7eda6cbd980fdb0b7d/rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala#L109-L112> 
producePeeks(dataList) should be under the same channel lock
https://github.com/rchain/rchain/blob/fef59e6334b251e4688d6037d2c91fdecbd5dd00/rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala#L204-L213

zsluedem:
https://github.com/tgrospic/rchain/tree/peek-non-negative-number-bug related branch which can reveal some of the problems with `peek` though I didn't run that once. More infos would be put here.

PR #3446, opened 2021-06-18
Fix peek conflict detection

### LINKED ISSUE #3349: Appending of pending rewards is conflicting for large networks

nzpr:
## Overview
Due to this bug https://github.com/rchain/rchain/pull/3445 (issue https://github.com/rchain/rchain/issues/3349#issuecomment-863916745) peek cannot be reliably detected by COMM event itself, so we have to examine **all** produces (including inside COMM log) for existence in pre state, and adjust set of consumed produces accordingly

### Notes
Closes https://github.com/rchain/rchain/issues/3349, but still the problem of not having peek exposed in COMM event log remains active. It's not clear whether it can be fixed without hard fork. But also we want peek to be atomic which should cover this issue as well https://github.com/rchain/rchain/issues/3268. So no new issue is created.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
@@ -40,7 +40,6 @@ object DeployChainIndex {
                        channelsStore = historyRepository,
                        historyRepository.getSerializeC
                      )
-
     } yield DeployChainIndex(
       deploysWithCost,
       preStateHash,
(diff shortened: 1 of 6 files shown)


nzpr:
Given recent findings <https://github.com/rchain/rchain/issues/3349#issuecomment-864167814>, closing this PR, as it looks like conflicts are valid.

PR #3448, opened 2021-06-21
[POC] Radix tree for new history implementation

### LINKED ISSUE #3382: Performance measurements of History implementation
## Overview

RSpace data is stored as a collection of key-value pairs (tuple space). For each distinct state of this collection we should be able to calculate a hash to uniquely define it and to provide a way to verify the integrity of the data.

Implementation of *Merkle-Patricia Trie* is used for this purpose in RNode (currently v0.10.2) written in [MergingHistory](https://github.com/rchain/rchain/blob/a9c0f7ded230696e5f0649a276b3cdef385eec1f/rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala#L23) class. It's implementation of `History` interface with two basic operations, to update (`process`) and read (`find`) key-value pairs.

https://github.com/rchain/rchain/blob/a9c0f7ded230696e5f0649a276b3cdef385eec1f/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L12-L17

In its core it's a *Radix Tree* structure that enables storing keys in a tree with branch nodes representing common `key` prefixes and with leaf nodes holding `values`. Each node has a key which is hash of its children nodes (references).

This issue provides information about performance of insert, read, delete operations and distribution of nodes per levels in a tree.

### Setup

Measurements are done with RAM disk and with LMDB implementation as a underlying key-value store.
CPU used is *AMD Ryzen 7 2700X (8 cores) 3781MHz* using only one thread.

Details data is provided in a file [RChain history storage performance.pdf](https://github.com/rchain/rchain/files/6344453/RChain.history.storage.performance.pdf).

### Hashing of keys

Testing with real RSpace database is not very convenient so for these tests I used generated keys by hashing a sequence of increasing integers. I compared distribution of nodes per level to the main net state and variations is minimal.

Hashing algorithm used is [Blake2b256](https://github.com/rchain/rchain/blob/2ec4c8a136c9e788c8ea198d32de0073b4edc1c6/rspace/src/main/scala/coop/rchain/rspace/Blake2b256Hash.scala#L16).

Generating hashes for a sequence `(0 until 1,000,000)` is about `1.5` seconds.

###  Insert chunks of 100k pairs

image

### Insert chunks of 250k pairs

image

### Insert chunks of 500k pairs

image

### Insert 10k - 2M pairs (in one chunk)

image

### Read 10k records

image

### Read 100k records

image

### Delete 10k records

image

### Delete 100k records

image

### Tree levels 10k - 500k pairs

image

### Tree levels 1M - 2M pairs

image

### Tree levels on main net (250,000 block)

image

tgrospic:
image


rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala
```diff
@@ -0,0 +1,319 @@
...
+import scala.collection.concurrent.TrieMap
+
+object RadixTree {
+
+  /**
+    * Radix Tree structure (Node and Leaf)
+    */
+  sealed trait TreeR
+  final case class NodeR(prefix: ByteVector, refs: Vector[ByteVector]) extends TreeR
+  final case class LeafR(prefix: ByteVector, value: ByteVector)        extends TreeR
```
zsluedem, line 18: Are you considering `extensionNode`(extension in eth term and `Skip` in rchain term ) same as `LeafR` here?

tgrospic: I'd like to be simple as possible so `Tree` with `Branch` and `Leaf` nodes seems like a good choice.

Do you know what's the meaning of _extension_ node in ETH?

zsluedem: https://ethereum.stackexchange.com/questions/39915/ethereum-merkle-patricia-trie-extension-node
I think this is kind of a good answer.

zsluedem: The extension node doesn't seem like a good way to resolve inserting or deleting large amount of nodes.

zsluedem: Maybe some tests are better to evaluate it.

PR #3452, opened 2021-06-25
[WIP] Enable attestation messages

nzpr:
## Overview
This is WIP that might go to RCHIP, but for sake of clarity and to expose ideas in the code, it's a draft PR for now.

The suggestion is to make block message(BM) be presented in two flavours: attestation message (AM) and state transition message (STM). 
Both of them are exposed as block message on the wire, but having a bit different field content. So no hard fork is required, and casper message logic is preserved.

**Why this is needed?** _To not execute Rholang just for Casper needs when no state changes are required. So to achieve much faster finality._ 

**Impact:** this is not backward compatible change. 

**Formal difference:** 
- STM has preStateHash and postStateHash and they are not equal, so there is some state transition: whether this is some user deploy or system deploy 
- AM has preStateHash and postStateHash equal empty byte string. 
The purpose of AM is to serve as an approval for Casper that does not introduce any state change, so decoupled from Rholang. Attestation message is issued according to synchrony constraint if no user deploys are there to create state transition, and validator is not forced to make state change with some system deploy (e.g. issue CloseBlock deploy to make epoch change or other state change that is tied to particular block height). NOTE: slashing is not such change (look into point 7 below).

Below influence of introducing such messsage is outlined.

1. **Latest messages**: remains the same, all block messages (both STM and AM) are accounted

2. **Synchrony constraint**: Only state transition messages are accounted for synchrony constraint. 
When checking the constraint, node have to find by traversing through self justification first one which is which is STM (`lsjSTM`) and check if `lsjSTM` is not in the view of latest self-proposed message.
So nothing changes except requirement to optionally traverse several blocks n the past through self justification.

3. **Justifications**: both STM and AM can be a justification. Block as a justification unit remains the same. 
The difference here is that AM is defined by its justifications, so it virtually has multiple blockHashes. If some new STM uses multiple AM as justifications, parents used for computing pre state hash are join of justifications of parent AMs. If AM has AMs in its justification, process continues recursively - traverse justifications until all paths led to some STM.
This process exposes difference between STM and AM and decoupling Casper from global state. Only STMs matter for the global state. 

4. **Last Finalised Block**: AM cannot be LFB, if new LFB found is AM, pick the most recent STM ancestor.

5. **Main parent**: nothing changing, when creating STM, node have to compute preStateHash and find bonds map, so pick main parent according to stake of message creator. AM can be main parent. Only problem that is visible for now is odd way of computing bonds map that involves reading bonds map of main parent. https://github.com/rchain/rchain/blob/61726ac897b55c1326d637ae715333153d9cdbda/casper/src/main/scala/coop/rchain/casper/safety/CliqueOracle.scala#L134-L150

6. **Merging**: as AM does not introduce state change it does not require knowledge about merged pre state. Only when STM is created, it should find as per `3.` parents to merge.

7. **Slashing**: ~~currently invalid messages are accounted when computing fault tolerance, so local view does not matter network wise. The only thing validator can do is try to put slashing deploy in its each new block (and possibly die brave on that hill if majority disagrees) once it sees invalid block and still sees offender bonded. So slash deploy is just put in the next STM according to normal network operation.~~ If node sees invalid block, it cannot create attestation over it, as it incurs an agreement. So when slashing is required, STM is created.  

8. **Epoch change**: closeBlock should not be triggered in AM, as it should not changes state and does not have preState computed. Therefore, if block number is epoch change block, but no user deploys are available to put in a block, STM with only close block should be created. NOTE: this is the only case when AM becomes STM, when there are slashing deploys put in a block - node still proceeds with AM, as it is free to issue slashing deploy at any time

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -107,16 +107,22 @@ object BlockCreator {
         dummyDeploys    = prepareDummyDeploy(nextBlockNum)
         slashingDeploys <- prepareSlashingDeploys(nextSeqNum)
         // make sure closeBlock is the last system Deploy
-        systemDeploys = slashingDeploys :+ CloseBlockDeploy(
-          SystemDeployUtil
-            .generateCloseDeployRandomSeed(selfId, nextSeqNum)
-        )
-        deploys = userDeploys -- s.deploysInScope ++ dummyDeploys
-        r <- if (deploys.nonEmpty || slashingDeploys.nonEmpty)
+        deploys            = userDeploys -- s.deploysInScope ++ dummyDeploys
+        isEpochChangeBlock = nextBlockNum % s.onChainState.shardConf.epochLength == 0
+
+        // Issue state transition message if there are user deploys to put or its an epoch change block,
+        // so epoch change has to be performed
+        issueSTM = deploys.nonEmpty || slashingDeploys.nonEmpty || isEpochChangeBlock
+        r <- if (issueSTM) {
               for {
                 now           <- Time[F].currentMillis
                 invalidBlocks = s.invalidBlocks
                 blockData     = BlockData(now, nextBlockNum, validatorIdentity.publicKey, nextSeqNum)
+                // if any slashing deploys to put - pack them in a block, pushing closeBlock into the end
+                systemDeploys = slashingDeploys :+ CloseBlockDeploy(
+                  SystemDeployUtil
+                    .generateCloseDeployRandomSeed(selfId, nextSeqNum)
+                )
                 checkpointData <- InterpreterUtil.computeDeploysCheckpoint(
                                    parents,
                                    deploys.toSeq,
@@ -137,7 +143,7 @@ object BlockCreator {
                 _             <- Span[F].mark("before-packing-block")
                 shardId       = s.onChainState.shardConf.shardName
                 casperVersion = s.onChainState.shardConf.casperVersion
-                unsignedBlock = packageBlock(
+                unsignedBlock = packageSTM(
                   blockData,
                   parents.map(_.blockHash),
                   justifications.toSeq,
@@ -154,8 +160,21 @@ object BlockCreator {
                 signedBlock = validatorIdentity.signBlock(unsignedBlock)
                 _           <- Span[F].mark("block-signed")
               } yield BlockCreatorResult.created(signedBlock)
-            else
-              BlockCreatorResult.noNewDeploys.pure[F]
+              // else issue attestation message
+            } else
+              for {
+                now       <- Time[F].currentMillis
+                blockData = BlockData(now, nextBlockNum, validatorIdentity.publicKey, nextSeqNum)
+                unsignedBlock = packageAM(
+                  blockData,
+                  parents.map(_.blockHash),
+                  justifications.toSeq,
+                  shardId = parents.head.shardId,
+                  version = parents.head.header.version
+                )
+                _           <- Span[F].mark("block-created")
+                signedBlock = validatorIdentity.signBlock(unsignedBlock)
+              } yield BlockCreatorResult.createdAttestation(signedBlock)
       } yield r
 
       for {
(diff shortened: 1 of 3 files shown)


zsluedem:
I am so glad that you made a very detailed description of this new feature although I read none of your codes right now.

Currently, I got only one disagreement with your description above. 

I think LFB can be an attestation block like stm. 

Another thing that is kind of different in my head is that the prestateHash and postStateHash should be the same but not with empty one. I am not picky on this one. But I think it would make api life easier with useful information.

Generally, I think attestation blocks should be treated like normal blocks without too many other considerations.
I think it makes things complicated putting a lot of colors on it.

nzpr:
```
Another thing that is kind of different in my head is that the prestateHash and postStateHash should be the same but not with empty one. I am not picky on this one. But I think it would make api life easier with useful information.
```
the problem is which preStateHash to use. If we want preStateHash that means something then we need merge on each attestation, which is costly and potentially produce double work. I'd better merge when I need to make state transition.

```
Generally, I think attestation blocks should be treated like normal blocks without too many other considerations.
I think it makes things complicated putting a lot of colors on it.
```
I agree thats the plan, but there are caveats and tradeoffs, this preStateHash is one of them

zsluedem:
@nzpr 
> If we want preStateHash that means something then we need merge on each attestation, which is costly and potentially produce double work

Supposed that in the future RChain is going to the top 10 Blockchain project and get 1/10 transactions of ETH, maybe at that time every block would get some deploys. The costly double work would still manifest when the network becomes really busy because of the merging design not because of the attestation block. 

This is nothing about attestation block can solve.

Back to the point `costly double work on merging`, I got a vague memory on some comments you made -> "Merging as much as blocks is the fastest"(Something like this not exactly the same). I hold a different point of view back in Jan because of the costly double work you mentioned here. 
I am not sure if that is correct or not. We should definitely test what is the best max parent number.

nzpr:
```... maybe at that time every block would get some deploys.``` If blocks are very fast, this might not happen. Id depends on how blocks are packaged, are all deploys in queue put on each block creation attempt. But for now I assume you want to have next deploy hitting validator once it proposed block to pack it in full.

```
Back to the point costly double work on merging, I got a vague memory on some comments you made -> 
"Merging as much as blocks is the fastest"(Something like this not exactly the same). 
I hold a different point of view back in Jan because of the costly double work you mentioned here.
```
You do merge to create pre state hash for state transition, if you do not make state transition - this is not necessary. It is useful only if this merge result can be reused in later STM. And this is what I'm talking about as double work - for now it looks like having bunch of AM as parents, they might have different states. So you did merge in each AM, and then you have to do that again in STM. 

```
We should definitely test what is the best max parent number.
```
I'm not sure why max parents number? Blocks are supposed to be created according to synchrony constraint, so parents number is defined by this value + stake distribution.

**Comment by tgrospic** [comment]
Re. merging in AM. If two STM blocks are created and on top of them are only AM blocks to make finalization. Who will do the merge? Finalized block must have state hash with already merged state.

So I think AM can do the merging and all other AM blocks on top of it don't need to do double work.

---

> 4. Last Finalised Block: AM cannot be LFB, if new LFB found is AM, pick the most recent STM ancestor.

This is related. If AM cannot be finalized most recent STM can mean multiple blocks (or taking only one parent from AM).

**Comment by nzpr** [comment]
```
Re. merging in AM. If two STM blocks are created and on top of them are only AM blocks to make finalization. Who will do the merge? Finalized block must have state hash with already merged state.
```
considering uniformly distributed stake and 0.49 sync constraint
```
AM1    AM2    AM3        AM4  AM5  AM6
       X                     X  <- cross here means cross connections
STM1   STM2   STM3       STM4 STM5 STM6
```
so left branch and right branch has been progressing on their own. So AM{1-3} merged STM{1-3}, and AM{4-6} merged STM{4-6}. Then some node seen all AM{1-6} and when user deploy came in, tried to create `STM7` - so merging STM{1-6}. So result of merge in AM{1-6} is not relevant, it cannot be reused.


```
This is related. If AM cannot be finalized most recent STM can mean multiple blocks (or taking only one parent from AM).
```
main parent STM state should be used here.

**Comment by tgrospic** [comment]
```
STM1 STM2 
                       AM3  AM4  AM5  AM6
```
What if I do explore deploy on AMx blocks? They have two state hashes, which one is used (finalized) if they are not merged?

nzpr:
You should not be able to explore on AM messages, or just explore on main parent. As AM cannot be finalized and no deploys are there changing the state, why would you need explore on AM?

PR #3454, opened 2021-06-25
Add leaderful propose script

zsluedem:
## Overview

This pr is intended to add a new propose script in scala version while the [old propose script](https://github.com/rchain/rchain-testnet-node/blob/dev/scripts/propose_in_turn.py) is  not applicable in the new leader network.

The general workflow in propose script is like below

```
1.   propose in leader node
2. wait for all the followers getting the block from leader
3. propose in parallel among followers 
4. wait for the leader getting the blocks from leader
5. go into 1 and loop
 
```

### DIFF
--- a/node/src/main/resources/propose.conf
+++ b/node/src/main/resources/propose.conf
@@ -1,0 +1,33 @@
+leader = {
+    host = "localhost",
+    grpc-port = 40402,
+    http-port=40403,
+    propose-info= {
+        deploy-contract: "Nil",
+        phlo-limit: 100,
+        phlo-price: 1,
+        deploy-private-key: "34d969f43affa8e5c47900e6db475cb8ddd8520170ee73b2207c54014006ff2b"
+    }
+}
+followers = [
+    {host = "localhost",
+     grpc-port = 40302 ,
+     http-port=40403,
+     propose-info= {
+         deploy-contract: "Nil",
+         phlo-limit: 100,
+         phlo-price: 1,
+         deploy-private-key: "5438ac0bde91bb813c7b4b17f215cd6ad52b19e7bcbf1f7907adb1d69c8aa7b1"
+     }},
+     {host = "localhost",
+     grpc-port = 40202 ,
+     http-port=40403,
+     propose-info= {
+         deploy-contract: "Nil",
+         phlo-limit: 100,
+         phlo-price: 1,
+         deploy-private-key: "304b2893981c36122a687c1fd534628d6f1d4e9dd8f44569039ea762dae2d3e7"
+     }},
+]
+check-interval = 10s
+propose-interval = 1s
--- a/node/src/main/scala/coop/rchain/node/proposescript/ProposeConfig.scala
+++ b/node/src/main/scala/coop/rchain/node/proposescript/ProposeConfig.scala
@@ -1,0 +1,132 @@
+package coop.rchain.node.proposescript
+
+import cats.syntax.all._
+import cats.effect.{Concurrent, Sync}
+import coop.rchain.monix.Monixable
+import coop.rchain.shared.Log
+import fs2.Stream
+import retry.{retryingOnFailures, RetryDetails, RetryPolicies, Sleep}
+
+import scala.concurrent.duration.FiniteDuration
+
+final case class ProposeConfig(
+    leader: ValidatorProposeInfo,
+    followers: List[ValidatorProposeInfo],
+    checkInterval: FiniteDuration,
+    proposeInterval: FiniteDuration
+) {
+
+  def logWaitProposeInterval[F[_]: Log]: F[Unit] =
+    Log[F].info(s"Wait ${proposeInterval} seconds for proposing interval.")
+
+  def proposeInLeader[F[_]: Sync: Monixable: Log]: F[String] =
+    leader.deployAndProposeAsync
+      .flatTap(blockHash => leader.logValidatorProposedBlock(blockHash))
+
+  def checkLeaderBlockInFollower[F[_]: Concurrent: Monixable: Log](blockHash: String): F[Boolean] =
+    Stream
+      .emits(followers)
+      .parEvalMapUnordered[F, Boolean](followers.length)(
+        v =>
+          v.isContainBlock(blockHash).flatTap { isContain =>
+            v.logValidatorGotBlock(blockHash)
+              .whenA(isContain) >> v.logValidatorMissedBlock(blockHash).whenA(!isContain)
+          }
+      )
+      .forall(identity)
+      .compile
+      .last
+      .map(_.getOrElse(false))
+
+  def waitFollowersGotLeaderBlock[F[_]: Concurrent: Monixable: Log: Sleep](
+      blockHash: String
+  ): F[Boolean] = {
+    val retryPolicy = RetryPolicies.constantDelay(checkInterval)
+    def onFailure(failedValue: Boolean, details: RetryDetails): F[Unit] =
+      Log[F].info(
+        s"Retry check in ${checkInterval} because followers didn't get the block ${blockHash} from leader." +
+          s"Retry ${details.retriesSoFar} times so far, cumulate cumulativeDelay:${details.cumulativeDelay}, ${failedValue}."
+      )
+    retryingOnFailures[Boolean](
+      policy = retryPolicy,
+      wasSuccessful = identity,
+      onFailure = onFailure
+    ) {
+      checkLeaderBlockInFollower(blockHash)
+    }
+  }
+
+  def proposeInFollowers[F[_]: Concurrent: Monixable: Log]: F[List[String]] =
+    Stream
+      .emits(followers)
+      .parEvalMapUnordered[F, String](followers.length)(
+        v => v.deployAndProposeAsync.flatTap(blockHash => v.logValidatorProposedBlock(blockHash))
+      )
+      .compile
+      .toList
+
+  def checkFollowerBlocksInLeader[F[_]: Concurrent: Monixable: Log](
+      blockHashes: List[String]
+  ): F[Boolean] =
+    Stream
+      .emits(blockHashes)
+      .parEvalMapUnordered[F, Boolean](blockHashes.length)(
+        b =>
+          leader
+            .isContainBlock(b)
+            .flatTap { isContain =>
+              leader
+                .logValidatorGotBlock(b)
+                .whenA(isContain) >> leader.logValidatorMissedBlock(b).whenA(!isContain)
+            }
+      )
+      .forall(identity)
+      .compile
+      .last
+      .map(_.getOrElse(false))
+
+  def waitLeaderGotFollowerBlocks[F[_]: Concurrent: Monixable: Log: Sleep](
+      blocks: List[String]
+  ): F[Boolean] = {
+    val retryPolicy = RetryPolicies.constantDelay(checkInterval)
+    def onFailure(failedValue: Boolean, details: RetryDetails): F[Unit] =
+      Log[F].info(
+        s"Retry check in ${checkInterval} because leader didn't get the block from followers." +
+          s"Retry ${details.retriesSoFar} times so far, cumulate cumulativeDelay:${details.cumulativeDelay}, ${failedValue}."
+      )
+    retryingOnFailures[Boolean](
+      policy = retryPolicy,
+      wasSuccessful = identity,
+      onFailure = onFailure
+    ) {
+      checkFollowerBlocksInLeader(blocks)
+    }
+  }
+
+  def proposeRound[F[_]: Concurrent: Monixable: Log: Sleep]: F[Unit] =
+    for {
+      _                    <- logWaitProposeInterval
+      _                    <- Sleep[F].sleep(proposeInterval)
+      _                    <- Log[F].info(s"Going to propose in the leader ${leader.host}.")
+      proposeBlockInLeader <- proposeInLeader
+      _ <- Log[F].info(
+            s"Proposed successfully in the leader with new block ${proposeBlockInLeader} " +
+              s"and then waiting the followers got the block."
+          )
+      _ <- waitFollowersGotLeaderBlock(proposeBlockInLeader)
+      _ <- logWaitProposeInterval
+      _ <- Sleep[F].sleep(proposeInterval)
+      _ <- Log[F].info(
+            s"All followers got the newly proposed block ${proposeBlockInLeader} from leader."
+          )
+      proposeBlocksInFollowers <- proposeInFollowers
+      _ <- Log[F].info(
+            s"Proposed successfully among all the followers. " +
+              s"Then waiting the leader to receive all the new blocks ${proposeBlocksInFollowers}."
+          )
+      _ <- waitLeaderGotFollowerBlocks(proposeBlocksInFollowers)
+      _ <- Log[F].info(
+            "The leader successfully received all proposed blocks from followers."
+          )
+    } yield ()
+}
--- a/node/src/main/scala/coop/rchain/node/proposescript/ProposeMain.scala
+++ b/node/src/main/scala/coop/rchain/node/proposescript/ProposeMain.scala
@@ -1,0 +1,46 @@
+package coop.rchain.node.proposescript
+
+import cats.effect.ExitCode
+import coop.rchain.shared.Log
+import fs2.Stream
+import monix.eval.{Task, TaskApp}
+import org.rogach.scallop.ScallopConf
+import pureconfig.ConfigSource
+import pureconfig.generic.auto._
+import java.nio.file.Path
+
+final case class ProposeScriptOptions(arguments: Seq[String]) extends ScallopConf(arguments) {
+  val width = 120
+  helpWidth(width)
+  printedName = "propose-script"
+
+  val config = opt[Path](
+    descr = s"Target config file path",
+    required = true
+  )
+  verify()
+
+}
+object ProposeMain extends TaskApp {
+  override def run(args: List[String]): Task[ExitCode] = {
+    val options      = ProposeScriptOptions(args)
+    val configPath   = options.config()
+    val configSource = ConfigSource.file(configPath)
+    val configE      = configSource.load[ProposeConfig]
+    val config       = configE.right.get
+    implicit val log = Log.log[Task]
+
+    for {
+      _ <- log.info(s"Running propose script with config: ${config}")
+      _ <- Stream
+            .eval(
+              config.proposeRound.tapError(
+                e => log.error(s"Running propose script with config: ${config} with error ${e}")
+              )
+            )
+            .repeat
+            .compile
+            .drain
+    } yield ExitCode.Success
+  }
+}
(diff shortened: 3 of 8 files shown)


node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+    for {
+      result <- deployAndPropose(isAsync = true)
+      _ <- result match {
+            case ProposeAsyncSuccessRegex(seqNum) => seqNum.pure[Task]
+            case failedResult =>
+              Task.raiseError[String](
+                new Exception(s"Propose async failed with ${failedResult} on $host: $grpcPort")
+              )
+          }
+      proposeResult <- proposeResult
```
zsluedem: The async way doesn't actually make too much difference from sync way from my perspective. On this line, the result would still hang there until the result come. 
Maybe I miss something?

tgrospic: In async variant, propose should not block and return immediately, this requires to call `proposeResult` to get the status of propose.
https://github.com/rchain/rchain/blob/66c79b65d1d86a7031e810667812e2c1eae845ea/models/src/main/protobuf/ProposeServiceV1.proto#L23

Shadowing `proposeResult` on multiple places and also the same name on propose service does not make it easy to follow. :wink:

---

@nzpr You were talking about some identifier returned in async call that can be referenced by the client to check result but I only see string as a result from _propose_.
https://github.com/rchain/rchain/blob/66c79b65d1d86a7031e810667812e2c1eae845ea/models/src/main/protobuf/ProposeServiceV1.proto#L26-L31

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+)
+final case class ValidatorInfo(
+    host: String,
+    grpcPort: Int,
+    httpPort: Int,
+    proposeInfo: ProposeInfo
+) {
+  val ProposeSuccessRegex      = "Success! Block ([0-9a-f]+) created and added.".r
+  val ProposeAsyncSuccessRegex = "Propose started \\(seqNum ([0-9]+)\\)".r
+  private def startGrpcChannel: Resource[Task, ManagedChannel] =
```
tgrospic: ValidatorInfo should not reference Task directly.

---

Because with each API call the whole process will exit, it doesn't make real difference here...

But `ManagedChannel` is meant to be used as a single object for one host. It has connection pool and can be used multi threaded.

It's so difficult to find documentation about that.
- https://stackoverflow.com/questions/63749113/grpc-call-channel-connection-and-http-2-lifecycle
- https://stackoverflow.com/questions/42320492/do-i-need-to-pool-managedchannel-instances-for-a-multithreaded-java-grpc-1-1-2
- https://groups.google.com/g/grpc-io/c/1-KMubq1tuc
- https://stackoverflow.com/questions/33197669/are-channel-stubs-in-grpc-thread-safe

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+    startGrpcChannel.evalMap(DeployServiceV1GrpcMonix.stub(_).pure[Task])
+
+  def proposeApi: Resource[Task, ProposeServiceV1GrpcMonix.ProposeServiceStub] =
+    startGrpcChannel.evalMap(ProposeServiceV1GrpcMonix.stub(_).pure[Task])
+
+  def getLatestBlocks(depth: Int = 1): Task[List[LightBlockInfo]] =
+    deployApi
+      .use(
+        deployService =>
+          for {
```
tgrospic: Use of curly braces can prevent unnecessary nesting. In this case by 4 spaces.
```scala
def getLatestBlocks(depth: Int = 1): Task[List[LightBlockInfo]] =
  deployApi.use { deployService =>
    for {
      blocksResp <- deployService.getBlocks(BlocksQuery(depth = depth)).toListL
```
Even this format works.
```scala
def getLatestBlocks(depth: Int = 1): Task[List[LightBlockInfo]] = deployApi.use { service =>
  for {
    blocksResp <- service.getBlocks(BlocksQuery(depth = depth)).toListL
```

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+            blocksResp <- deployService
+                           .getBlocks(BlocksQuery(depth = depth))
+                           .mapEval(Task.pure)
```
tgrospic: `.mapEval(Task.pure)` here is basically `identity` function. It's the same as `.evalMap(Task.pure)` in fs2 Stream.

But it's important _combinator_ to easily combine functions like `A => F[B]` in a Stream context without any manual lifting.
```scala
val s: Stream[F, A] = ???
def fun(x: A): F[B] = ???
def gun(x: B): F[C] = ???
// It is composable with "normal" F functions
val r = s.evalMap(fun).evalMap(gun)
```

That is the reason to create functions in `F` context whenever is possible and use specialized DSL like Stream, OptionT, EitherT or Observable only when this has meaning for the result. E.g. we have multiple lazy values to use Stream, optional value for OptionT etc.
So functions like this doesn't make sense because caller can always and easily _lift_ `F[B]` to `Stream[F, B]` and on the other side function can be easily use with streams `stream.evalMap(fun)`.
```scala
def fun(x: A): Stream[F, B] = Stream.eval(gun(x))
```
So when using specialized DSL (monad transformers) and these combinator functions, _for_ comprehensions is not really recommended.

Here are more examples from @nzpr's PR.
- https://github.com/rchain/rchain/pull/3421#discussion_r646117032
- https://github.com/rchain/rchain/pull/3421#discussion_r646121788

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+
+  def main(args: Array[String]): Unit = {
+    val options      = ProposeScriptOptions(args)
+    val configPath   = options.config()
+    val configSource = ConfigSource.file(configPath)
+    val configE      = configSource.load[ProposeConfig]
+    val config       = configE.right.get
+    val log          = Log.log[Task]
+
+    def logInfo(info: String) = Stream.eval(log.info(info))
```
tgrospic: All of these functions can return in `F` context.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+        .parEvalMapUnordered[Task, Boolean](blockHashes.length)(
+          b =>
+            config.leader
+              .isContainBlock(b)
+              .flatTap { isContain =>
+                logValidatorGotBlock(config.leader, b)
+                  .whenA(isContain) >> logValidatorMissedBlock(config.leader, b).whenA(!isContain)
+              }
+        )
+        .forall(identity)
```
tgrospic: When you need parallelism you can opt-in for a Stream (do forking) and at the end collect results (do joining).
```scala
.forall(identity).compile.last.map(_.getOrElse(false))
```

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+      for {
+        _ <- logInfo(
+              s"Wait ${config.checkInterval} seconds and then check if leader got the block from followers."
+            )
+        _      <- wait(config.checkInterval)
+        result <- checkFollowerBlocksInLeader(blocks)
+        _      <- waitLeaderGotFollowerBlocks(blocks).whenA(!result)
+      } yield ()
+
+    log.info(s"Running propose script with config: ${config}").runSyncUnsafe()
```
tgrospic: This should be composed with other _effectful_ values and avoid calling unsafe.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,351 @@
...
+    descr = s"Target config file path",
+    required = true
+  )
+  verify()
+
+}
+
+object ProposeLeaderful {
+
+  def main(args: Array[String]): Unit = {
```
tgrospic: Monix (as IO) has very useful helper for defining app entry point which can directly return Task.

This is what it might look like with `task` in F and repeated with Stream.
```scala
object ProposeLeaderful extends TaskApp {
  def run(args: List[String]): Task[ExitCode] = {
    ...
    Stream.eval(task).repeat.compile.drain.as(ExitCode.Success)
  }
```

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: > ValidatorInfo should not reference Task directly.

Do you mean I should use F? Or I should use syntax to get rid of task method define with case class?

> Because with each API call the whole process will exit, it doesn't make real difference here...

What do you mean about that?

> But ManagedChannel is meant to be used as a single object for one host. It has connection pool and can be used multi threaded.

Yes. You are right. I was thinking considering channel as normal tcp connection and close it when it is done which can save some efforts of thinking higher concepts.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: The `ProposeResultQuery` is empty right now.

I was thinking that I send the `ProposeResultQuery` and the `proposeResult` can be returned **immediately** with `ProposeInProgress` or `ProposeDone`. But the query would get the result until the propose is done. Just mention this. Not sure about what is the original design goal on this.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
tgrospic: Yes, please use parameter F instead directly Task. You would also need Monixable because generated gRPC bindings uses monix Task directly.

It basically the same as in [DeployService](https://github.com/rchain/rchain/blob/a6457a836778/node/src/main/scala/coop/rchain/node/api/client/DeployService.scala#L37) and [ProposeService](https://github.com/rchain/rchain/blob/a6457a8367780/node/src/main/scala/coop/rchain/node/api/client/ProposeService.scala#L23) implementations.
NOTE: Creating channel in constructor and converting errors to string is wrong, but we will fix that later.
https://github.com/rchain/rchain/blob/a6457a836778041c51307fedd2ebfe0961b7f09a/node/src/main/scala/coop/rchain/node/api/client/ProposeService.scala#L23-L32

---

> Because with each API call the whole process will exit, it doesn't make real difference here...

Sorry, I was thinking for CLI client calls. So it is the difference here to reuse the same channel.

tgrospic: What I usually do with these kind of _builders_ is to create a wrapper which accepts function that can modify builder options.
NOTE: I'm not sure if we want to copy termination logic from [DeployService](https://github.com/rchain/rchain/blob/a6457a836778/node/src/main/scala/coop/rchain/node/api/client/DeployService.scala#L183-L191).
```scala
def createChannel[F[_]: Sync](
    setup: NettyChannelBuilder => NettyChannelBuilder
): Resource[F, ManagedChannel] =
  Resource.make[F, ManagedChannel](
    Sync[F].delay(setup(NettyChannelBuilder.forAddress(host, grpcPort)).build())
  )(channel => Sync[F].delay(channel.shutdown()) >> ().pure[F])
```
Using it is easy and configurable.
```scala
val channelResource1 = createChannel[Task](identity) 
val channelResource2 = createChannel[Task](_.usePlaintext()) // with additional options
```

zsluedem: > Yes, please use parameter F instead directly Task. You would also need Monixable because generated gRPC bindings uses monix Task directly.

Ok. I didn't find any way to turn Observerble into F. I got you now. I was thinking I am already at the end of the world so it is fine to be concrete. You make more sense.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: Very good point. I didn't know `for` comprehension could cause memory leak.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: @tgrospic i still got some problems with [Observable](https://monix.io/docs/current/reactive/observable.html#observable-and-functional-programming). I don't think Observable[A] can easily turn into F[A] without introducing Task([like the old dev did]( https://github.com/rchain/rchain/blob/dev/shared/src/main/scala/coop/rchain/monix/MonixableSyntax.scala#L19)). This is one of the reason I used Task instead of F.

tgrospic: Unfortunately we cannot avoid Task because how gRPC services are generated. That's why I created Monixable, to have conversion from Task to F (and reverse). So Task you get from observable can be converted to F.

Similar case is e.g. when sending a message with `TransportLayer#send` which uses Task directly.
https://github.com/rchain/rchain/blob/db297f119eded438744e57e6c7a9579db54c165b/comm/src/main/scala/coop/rchain/comm/transport/GrpcTransport.scala#L87-L102

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: I actually want to compose Stream instead of in `for` which is more flexible on paralleling and fun. But it doesn't matter.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
zsluedem: @tgrospic I don't think Observable works like Stream. I think I need to use mapEval here to make call on `onSubscription` which is needed on this Observable.

node/src/main/scala/coop/rchain/node/proposescript/ProposeLeaderful.scala
```diff
@@ -0,0 +1,340 @@
...
+      proposeBlocksInFollowers <- proposeInFollowers
+      _ <- log.info(
+            s"Proposed successfully among all the followers. " +
+              s"Then waiting the leader to receive all the new blocks ${proposeBlocksInFollowers}."
+          )
+      _ <- waitLeaderGotFollowerBlocks(proposeBlocksInFollowers)
+      _ <- log.info(
+            "The leader successfully received all proposed blocks from followers."
+          )
+    } yield ()
```
tgrospic: This code looks so nice that it deserves their own file and with abstracted F. Or keep it in this file and have new Main.

node/src/main/scala/coop/rchain/node/proposescript/ValidatorProposeInfo.scala
```diff
@@ -0,0 +1,192 @@
...
+    proposeInfo: ProposeInfo
+) {
+  private val ProposeSuccessRegex      = "Success! Block ([0-9a-f]+) created and added.".r
+  private val ProposeAsyncSuccessRegex = "Propose started \\(seqNum ([0-9]+)\\)".r
+
+  private val channel: ManagedChannel =
+    ManagedChannelBuilder
+      .forAddress(host, grpcPort)
+      .usePlaintext()
+      .build
```
tgrospic, line 44: Closing the channel is not done now.

Let's make this whole class something like `ValidatorGrpcClient[F[_]: ...]` which accepts a channel. This whole class can be parameterized by `F`.

Channel should be wrapped in Resource so it can be closed gracefully. We can reuse this function later.
https://github.com/rchain/rchain/pull/3454#discussion_r660412472

node/src/main/scala/coop/rchain/node/proposescript/ProposeConfig.scala
```diff
@@ -0,0 +1,132 @@
...
+import retry.{retryingOnFailures, RetryDetails, RetryPolicies, Sleep}
+
+import scala.concurrent.duration.FiniteDuration
+
+final case class ProposeConfig(
+    leader: ValidatorProposeInfo,
+    followers: List[ValidatorProposeInfo],
+    checkInterval: FiniteDuration,
+    proposeInterval: FiniteDuration
+) {
```
tgrospic, line 17: Classes used for configuration (only constructor fields) should be separated from classes used to define operations.

Functions here are the essence of leaderful proposer so makes sense to call it something like `LeaderfulProposer[F[_]: ...]`. Constructor parameters defines what is necessary to run it.

node/src/main/scala/coop/rchain/node/proposescript/ProposeConfig.scala
```diff
@@ -0,0 +1,132 @@
...
+    retryingOnFailures[Boolean](
+      policy = retryPolicy,
+      wasSuccessful = identity,
+      onFailure = onFailure
+    ) {
+      checkFollowerBlocksInLeader(blocks)
+    }
+  }
+
+  def proposeRound[F[_]: Concurrent: Monixable: Log: Sleep]: F[Unit] =
```
tgrospic, line 106: I'd put this function on top because it's definition of the whole process. It's like specification which I want to see first and from there explore inner components.

node/src/main/scala/coop/rchain/node/proposescript/ProposeMain.scala
```diff
@@ -0,0 +1,46 @@
...
+object ProposeMain extends TaskApp {
+  override def run(args: List[String]): Task[ExitCode] = {
+    val options      = ProposeScriptOptions(args)
+    val configPath   = options.config()
+    val configSource = ConfigSource.file(configPath)
+    val configE      = configSource.load[ProposeConfig]
+    val config       = configE.right.get
+    implicit val log = Log.log[Task]
+
+    for {
```
tgrospic, line 33: Here, netty channel (as Resource) should be constructed and supplied to ValidatorGrpcClient, which is then supplied to LeaderfulProposer.

node/src/main/resources/propose.conf
```diff
@@ -0,0 +1,33 @@
...
+         deploy-contract: "Nil",
+         phlo-limit: 100,
+         phlo-price: 1,
+         deploy-private-key: "5438ac0bde91bb813c7b4b17f215cd6ad52b19e7bcbf1f7907adb1d69c8aa7b1"
+     }},
+     {host = "localhost",
+     grpc-port = 40202 ,
+     http-port=40403,
+     propose-info= {
+         deploy-contract: "Nil",
```
nzpr, line 26: I think it would be very useful to have here folder configurable and read random `.rho` file, to test merging.

zsluedem:
It seems we don't need this anymore. Close it now.

PR #3459, opened 2021-06-29
Migrate LFB from LastFinalisedStore to BlockDagStorage

nzpr:
## Overview
Currently there is a single key value storage for last finalised block. 

Given https://github.com/rchain/rchain/pull/3439, Last Finalised Block is necessary part of DAG representation, therefore it should be stored in BlockMetadata. Also we should record at least two last directly finalised blocks (preferably all of them) to detect when finalised deploy cannot be rejected by merge. 

This PR is part of https://github.com/rchain/rchain/pull/3439 that belongs to dev. 
It introduces migration and deprecates LastFinaizedStorage. In addition, flag `finalized` is introduced for BlockMetadata to mark indirect finalization.

**Migration logic is the following:** 
- read last finalized record from LastFinaizedStorage, update BlockMetadata for this hash. 
- In addition - collect all ancestors and mark them as finalized **and directlyiFinalized** (this is because current networks are single parent, so all blocks are supposed to be finalized directly). This processing is done in chunks - chunk of blocks is read from metadata storage, updated and then put back.

**Performance results**
Full main net DAG storage contains 880155 blocks, size of 7.8 Gb
batch size 1000 => 10 min
```
07:45:46.579 [INFO ] [main                ] [c.rchain.node.runtime.Setup$ ] - Migrating LastFinalizedStorage to BlockDagStorage.
07:46:02.439 [INFO ] [main                ] [.b.d.BlockDagKeyValueStorage$] - Migration of LFB done.
07:46:04.616 [INFO ] [node-runner-30      ] [.b.d.BlockDagKeyValueStorage$] - Finalized blocks recorded: 1000 of 880155.
....
07:56:00.780 [INFO ] [node-runner-31      ] [.b.d.BlockDagKeyValueStorage$] - Finalized blocks recorded: 880155 of 880155.
```

for batch size 10000 => 6 min
```
08:00:30.338 [INFO ] [main                ] [c.rchain.node.runtime.Setup$ ] - Migrating LastFinalizedStorage to BlockDagStorage.
08:00:45.994 [INFO ] [main                ] [.b.d.BlockDagKeyValueStorage$] - Migration of LFB done.
08:00:50.571 [INFO ] [node-runner-29      ] [.b.d.BlockDagKeyValueStorage$] - Finalized blocks recorded: 10000 of 880188.
...
08:07:08.876 [INFO ] [node-runner-30      ] [.b.d.BlockDagKeyValueStorage$] - Finalized blocks recorded: 880188 of 880188.
```

Looks like 10k blocks (about 100MB chunk) is a reasonable size.

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -1,5 +1,6 @@
 package coop.rchain.blockstorage.dag
 
+import cats.Show
 import cats.effect.concurrent.Semaphore
 import cats.effect.{Concurrent, Sync}
 import cats.syntax.all._
@@ -232,9 +233,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
         _ <- latestMessagesIndex.put(newLatestToAdd.toList)
 
         // if block added as approved, record it as directly finalized.
-        _ <- blockMetadataIndex
-              .recordDirectlyFinalised(blockMetadata.blockHash)
-              .whenA(approved)
+        _ <- blockMetadataIndex.recordFinalized(blockMetadata.blockHash, Set.empty).whenA(approved)
 
       } yield ()
     }
@@ -249,11 +248,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def accessEquivocationsTracker[A](f: EquivocationsTracker[F] => F[A]): F[A] =
     lock.withPermit(f(KeyValueStoreEquivocationsTracker))
 
-  def addFinalizedBlockHashes(hashes: List[BlockHash]): F[Unit] =
-    blockMetadataIndex.recordFinalized(hashes)
-
-  def recordDirectlyFinalised(blockHash: BlockHash): F[Unit] =
-    blockMetadataIndex.recordDirectlyFinalised(blockHash)
+  /** Record that some hash is directly finalized (detected by finalizer and becomes LFB). */
+  def recordDirectlyFinalized(
+      directlyFinalizedHash: BlockHash,
+      finalizationEffect: Set[BlockHash] => F[Unit]
+  ): F[Unit] =
+    for {
+      dag    <- representation
+      errMsg = s"Attempting to finalize nonexistent hash ${PrettyPrinter.buildString(directlyFinalizedHash)}."
+      _      <- dag.contains(directlyFinalizedHash).ifM(().pure, new Exception(errMsg).raiseError)
+      // all non finalized ancestors should be finalized as well (indirectly)
+      indirectlyFinalized <- dag.withAncestors(directlyFinalizedHash, dag.isFinalized(_).not)
+      // invoke effects
+      _ <- finalizationEffect(indirectlyFinalized + directlyFinalizedHash)
+      // persist finalization
+      _ <- blockMetadataIndex.recordFinalized(directlyFinalizedHash, indirectlyFinalized)
+    } yield ()
 }
 
 object BlockDagKeyValueStorage {
@@ -337,7 +347,6 @@ object BlockDagKeyValueStorage {
   ): F[Unit] = {
     val errNoLfbInStorage   = "No LFB in LastFinalizedStorage when attempting migration."
     val errNoMetadataForLfb = "No metadata found for LFB when attempting migration."
-    val errNoMetadataForFb  = "No metadata found for finalized block when attempting migration."
 
     for {
       blockMetadataDb <- kvm.database[BlockHash, BlockMetadata](
@@ -363,13 +372,33 @@ object BlockDagKeyValueStorage {
           blocksInfoMap.get(bh).map(_.parents.toList).getOrElse(List.empty)
         }
         .toList
-      curVs <- finalizedBlockSet.traverse { h =>
-                blockMetadataDb.get(h).flatMap(_.liftTo[F](new Exception(errNoMetadataForFb)))
+
+      processChunk = (chunk: List[BlockHash]) => {
+        implicit val s = new Show[BlockHash] {
+          override def show(t: BlockHash): String = PrettyPrinter.buildString(t)
+        }
+
+        for {
+          curVs <- blockMetadataDb.getUnsafeBatch(chunk)
+          // WARNING: migration should be done before block merge, as it assumes all blocks are directly finalized.
+          newVs = curVs.map(_.copy(directlyFinalized = true, finalized = true))
+          _     <- blockMetadataDb.put(newVs.map(v => (v.blockHash, v)))
+        } yield ()
+      }
+
+      chunkSize = 10000L
+      _ <- fs2.Stream
+            .fromIterator(finalizedBlockSet.grouped(chunkSize.toInt))
+            .evalMapAccumulate(0L)(
+              (processed, chunk) => {
+                val processSoFar = processed + chunk.size.toLong
+                processChunk(chunk) >> Log[F]
+                  .info(s"Finalized blocks recorded: ${processSoFar} of ${finalizedBlockSet.size}.")
+                  .as((processSoFar, ()))
               }
-      // WARNING: migration should be done before block merge, as it assumes all blocks are directly finalized.
-      newVs = curVs.map(_.copy(directlyFinalized = true, finalized = true))
-      _     <- blockMetadataDb.put(newVs.map(v => (v.blockHash, v)))
-      _     <- Log[F].info(s"Finalized blocks recorded: ${newVs.size}.")
+            )
+            .compile
+            .drain
 
     } yield ()
   }
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -258,7 +258,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       errMsg = s"Attempting to finalize nonexistent hash ${PrettyPrinter.buildString(directlyFinalizedHash)}."
       _      <- dag.contains(directlyFinalizedHash).ifM(().pure, new Exception(errMsg).raiseError)
       // all non finalized ancestors should be finalized as well (indirectly)
-      indirectlyFinalized <- dag.withAncestors(directlyFinalizedHash, dag.isFinalized(_).not)
+      indirectlyFinalized <- dag.ancestors(directlyFinalizedHash, dag.isFinalized(_).not)
       // invoke effects
       _ <- finalizationEffect(indirectlyFinalized + directlyFinalizedHash)
       // persist finalization
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -1,6 +1,5 @@
 package coop.rchain.blockstorage.dag
 
-import cats.Show
 import cats.effect.concurrent.Semaphore
 import cats.effect.{Concurrent, Sync}
 import cats.syntax.all._
@@ -10,7 +9,7 @@ import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
 import coop.rchain.blockstorage.dag.BlockMetadataStore.BlockMetadataStore
 import coop.rchain.blockstorage.dag.EquivocationTrackerStore.EquivocationTrackerStore
 import coop.rchain.blockstorage.dag.codecs._
-import coop.rchain.blockstorage.finality.LastFinalizedStorage
+import coop.rchain.blockstorage.syntax._
 import coop.rchain.blockstorage.util.BlockMessageUtil._
 import coop.rchain.casper.PrettyPrinter
 import coop.rchain.casper.protocol.BlockMessage
@@ -21,8 +20,6 @@ import coop.rchain.models.EquivocationRecord.SequenceNumber
 import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
 import coop.rchain.shared.syntax._
-import coop.rchain.blockstorage.syntax._
-import coop.rchain.dag.DagOps
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 
@@ -340,66 +337,4 @@ object BlockDagKeyValueStorage {
       stores.invalidBlocks,
       stores.equivocations
     )
-
-  def migrateLfb[F[_]: Sync: Log](
-      lastFinalizedStorage: LastFinalizedStorage[F],
-      kvm: KeyValueStoreManager[F]
-  ): F[Unit] = {
-    val errNoLfbInStorage   = "No LFB in LastFinalizedStorage when attempting migration."
-    val errNoMetadataForLfb = "No metadata found for LFB when attempting migration."
-
-    for {
-      blockMetadataDb <- kvm.database[BlockHash, BlockMetadata](
-                          "block-metadata",
-                          codecBlockHash,
-                          codecBlockMetadata
-                        )
-      // record LFB
-      lfb  <- lastFinalizedStorage.get() >>= (_.liftTo(new Exception(errNoLfbInStorage)))
-      curV <- blockMetadataDb.get(lfb).flatMap(_.liftTo[F](new Exception(errNoMetadataForLfb)))
-      _    <- blockMetadataDb.put(lfb, curV.copy(directlyFinalized = true, finalized = true))
-      blocksInfoMap <- blockMetadataDb
-                        .collect {
-                          case (hash, metaData) =>
-                            (hash, BlockMetadataStore.blockMetadataToInfo(metaData()))
-                        }
-                        .map(_.toMap)
-      _ <- Log[F].info("Migration of LFB done.")
-
-      // record finalized blocks
-      finalizedBlockSet = DagOps
-        .bfTraverse(List(lfb)) { bh =>
-          blocksInfoMap.get(bh).map(_.parents.toList).getOrElse(List.empty)
-        }
-        .toList
-
-      processChunk = (chunk: List[BlockHash]) => {
-        implicit val s = new Show[BlockHash] {
-          override def show(t: BlockHash): String = PrettyPrinter.buildString(t)
-        }
-
-        for {
-          curVs <- blockMetadataDb.getUnsafeBatch(chunk)
-          // WARNING: migration should be done before block merge, as it assumes all blocks are directly finalized.
-          newVs = curVs.map(_.copy(directlyFinalized = true, finalized = true))
-          _     <- blockMetadataDb.put(newVs.map(v => (v.blockHash, v)))
-        } yield ()
-      }
-
-      chunkSize = 10000L
-      _ <- fs2.Stream
-            .fromIterator(finalizedBlockSet.grouped(chunkSize.toInt))
-            .evalMapAccumulate(0L)(
-              (processed, chunk) => {
-                val processSoFar = processed + chunk.size.toLong
-                processChunk(chunk) >> Log[F]
-                  .info(s"Finalized blocks recorded: ${processSoFar} of ${finalizedBlockSet.size}.")
-                  .as((processSoFar, ()))
-              }
-            )
-            .compile
-            .drain
-
-    } yield ()
-  }
 }
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -9,7 +9,7 @@ import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
 import coop.rchain.blockstorage.dag.BlockMetadataStore.BlockMetadataStore
 import coop.rchain.blockstorage.dag.EquivocationTrackerStore.EquivocationTrackerStore
 import coop.rchain.blockstorage.dag.codecs._
-import coop.rchain.blockstorage.finality.{LastFinalizedKeyValueStorage, LastFinalizedStorage}
+import coop.rchain.blockstorage.finality.LastFinalizedStorage
 import coop.rchain.blockstorage.util.BlockMessageUtil._
 import coop.rchain.casper.PrettyPrinter
 import coop.rchain.casper.protocol.BlockMessage
@@ -42,6 +42,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       childMap: Map[BlockHash, Set[BlockHash]],
       heightMap: SortedMap[Long, Set[BlockHash]],
       invalidBlocksSet: Set[BlockMetadata],
+      lastFinalizedBlockHash: BlockHash,
       finalizedBlocksSet: Set[BlockHash]
   ) extends BlockDagRepresentation[F] {
 
@@ -61,6 +62,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
     def latestMessageHashes: F[Map[Validator, BlockHash]] = latestMessagesMap.pure[F]
 
     def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
+
+    override def lastFinalizedBlock: BlockHash = lastFinalizedBlockHash
 
     // latestBlockNumber, topoSort and lookupByDeployId are only used in BlockAPI.
     // Do they need to be part of the DAG current state or they can be moved to DAG storage directly?
@@ -125,6 +128,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       childMap           <- blockMetadataIndex.childMapData
       heightMap          <- blockMetadataIndex.heightMap
       invalidBlocks      <- invalidBlocksIndex.toMap.map(_.toSeq.map(_._2).toSet)
+      lastFinalizedBlock <- blockMetadataIndex.lastFinalizedBlock
       finalizedBlocksSet <- blockMetadataIndex.finalizedBlockSet
     } yield KeyValueDagRepresentation(
       dagSet,
@@ -132,6 +136,7 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       childMap,
       heightMap,
       invalidBlocks,
+      lastFinalizedBlock,
       finalizedBlocksSet
     )
 
@@ -140,7 +145,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
   def insert(
       block: BlockMessage,
-      invalid: Boolean
+      invalid: Boolean,
+      approved: Boolean
   ): F[BlockDagRepresentation[F]] = {
     import cats.instances.list._
     import cats.instances.option._
@@ -223,11 +229,19 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
         // Add latest messages to DB
         _ <- latestMessagesIndex.put(newLatestToAdd.toList)
+
+        // if block added as approved, record it as directly finalized.
+        _ <- blockMetadataIndex
+              .recordDirectlyFinalised(blockMetadata.blockHash)
+              .whenA(approved)
+
       } yield ()
     }
 
     lock.withPermit(
-      blockMetadataIndex.contains(block.blockHash).ifM(logAlreadyStored, doInsert) >> representation
+      blockMetadataIndex
+        .contains(block.blockHash)
+        .ifM(logAlreadyStored, doInsert) >> representation
     )
   }
 
@@ -236,6 +250,9 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
   def addFinalizedBlockHash(blockHash: BlockHash): F[Unit] =
     blockMetadataIndex.addFinalizedBlock(blockHash)
+
+  def recordDirectlyFinalised(blockHash: BlockHash): F[Unit] =
+    blockMetadataIndex.recordDirectlyFinalised(blockHash) >> addFinalizedBlockHash(blockHash)
 }
 
 object BlockDagKeyValueStorage {
@@ -249,8 +266,7 @@ object BlockDagKeyValueStorage {
       equivocationsDb: KeyValueTypedStore[F, (Validator, SequenceNumber), Set[BlockHash]],
       latestMessages: KeyValueTypedStore[F, Validator, BlockHash],
       invalidBlocks: KeyValueTypedStore[F, BlockHash, BlockMetadata],
-      deploys: KeyValueTypedStore[F, DeployId, BlockHash],
-      lastFinalizedBlockDb: LastFinalizedStorage[F]
+      deploys: KeyValueTypedStore[F, DeployId, BlockHash]
   )
 
   private def createStores[F[_]: Concurrent: Log: Metrics](kvm: KeyValueStoreManager[F]) = {
@@ -262,15 +278,7 @@ object BlockDagKeyValueStorage {
                           codecBlockHash,
                           codecBlockMetadata
                         )
-      // last finalized block
-      lastFinalizedBlockDb <- KeyValueStoreManager[F].database[Int, BlockHash](
-                               "last-finalized-block",
-                               codecSeqNum,
-                               codecBlockHash
-                             )
-      lastFinalizedStore = LastFinalizedKeyValueStorage(lastFinalizedBlockDb)
-      lastFinalizedBlock <- lastFinalizedStore.get
-      blockMetadataStore <- BlockMetadataStore[F](blockMetadataDb, lastFinalizedBlock)
+      blockMetadataStore <- BlockMetadataStore[F](blockMetadataDb)
       // Equivocation tracker map
       equivocationTrackerDb <- KeyValueStoreManager[F]
                                 .database[(Validator, SequenceNumber), Set[BlockHash]](
@@ -305,8 +313,7 @@ object BlockDagKeyValueStorage {
       equivocationTrackerDb,
       latestMessagesDb,
       invalidBlocksDb,
-      deployIndexDb,
-      lastFinalizedStore
+      deployIndexDb
     )
   }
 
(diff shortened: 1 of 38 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
```diff
@@ -37,6 +39,8 @@ trait BlockDagRepresentation[F[_]] {
       startBlockNumber: Long,
       maybeEndBlockNumber: Option[Long]
   ): F[Vector[Vector[BlockHash]]]
+  // DAG representation has to have finalized block, or it does not make sense
+  def lastFinalisedBlock: BlockHash
```
nzpr: In the end, I decided to have LFB be present without Option. It required more work but I believe it worth it.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -41,11 +41,22 @@ object BlockMetadataStore {
       dagSet: Set[BlockHash],
       childMap: Map[BlockHash, Set[BlockHash]],
       heightMap: SortedMap[Long, Set[BlockHash]],
+      // In general - at least genesis should be LFB.
+      // But dagstate can be empty, as it is initialized before genesis is inserted.
+      // Also lots of tests do not have genesis properly initialised, so fixing all this is pain.
+      // So this is Option.
+      lastFinalizedBlock: Option[BlockHash],
```
nzpr: LFB is option here because there should be empty DagState as init value to recreate inMem representation. But this is not exposed to client code.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -93,6 +111,12 @@ object BlockMetadataStore {
     def heightMap: F[SortedMap[Long, Set[BlockHash]]] =
       dagState.get.map(_.heightMap)
 
+    def lastFinalizedBlock(implicit sync: Sync[F]): F[BlockHash] = {
+      val errMsg =
+        "DagState does not contain lastFinalizedBlock. Are you calling this on empty BlockDagStorage? Otherwise there is a bug."
+      dagState.get.flatMap(_.lastFinalizedBlock.liftTo[F](new Exception(errMsg)))
```
nzpr: this is public API of BlockMetadataStore, it forces that when this function is called, BlockDagStorage should already contain "approved block | genesis | LFB". No one should call this before there is some meaningful DAG present in DagStorage.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +63,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
     def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
 
+    override def lastFinalisedBlock: BlockHash = lastFinalizedBlock
```
tgrospic: I think we have finalized with `z` on most places.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -322,4 +329,30 @@ object BlockDagKeyValueStorage {
...
+                        )
+      lfb <- lastFinalizedStorage.get() >>= (_.liftTo(new Exception(errNoLfbInStorage)))
+      curV <- blockMetadataDb
+               .get(lfb)
+               .flatMap(
+                 _.liftTo[F](
+                   new Exception(errNoMetadataForLfb)
+                 )
+               )
+      _ <- blockMetadataDb.put(lfb, curV.copy(isDirectlyFinalized = true))
```
tgrospic: To be consistent, we need to update all ancestors with directly finalized flag.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -119,6 +119,7 @@ message BlockMetadataInternal {
   int64 blockNum                        = 6;
   int32 seqNum                          = 7;
   bool invalid                          = 8; // whether the block was marked as invalid
+  bool isDirectlyFinalized              = 9;
```
tgrospic: `is` prefix is not used for other boolean fields so it makes sense to not introduce it just for this field.

---

How _indirectly_ finalized blocks are handled?

models/src/main/scala/coop/rchain/models/BlockMetadata.scala
```diff
@@ -70,6 +73,8 @@ object BlockMetadata {
       weightMap(b.body.state),
       b.body.state.blockNumber,
       b.seqNum,
-      invalid
+      invalid,
+      // this value is not used anywhere down the call pipeline, so its safe to set it to false
+      isDirectlyFinalized = false
```
tgrospic: We should try to remove this function in the future.

For now maybe better to include directly finalized flag as a parameter, the same as invalid, to be more clear.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -130,32 +162,56 @@ object BlockMetadataStore {
...
+
+    // Calculate set of finalized blocks
+    val finalizedBlockSet = lfbOpt.fold(Set.empty[BlockHash])(
+      lfb =>
+        DagOps
+          .bfTraverse(List(lfb._1))(
+            bh => blocksInfoMap.get(bh).map(_.parents.toList).getOrElse(List.empty)
+          )
+          .toSet
+    )
```
tgrospic: This seems to complicated. Try to add fourth _update_ to `addBlockToDagState` for LFB.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Yeah, but spelling checker always wants 's'. Ok, lets use `z`

models/src/main/protobuf/CasperMessage.proto
nzpr: Indirectly finalized are only maintained in memory. All ancestors of LFB loaded from disk on startup, and then when finalization effect is invoked - this set is appended.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: This is done before BlockDagStorage is initialized. So when InMem state is created it should do the job for indirectly finalized. https://github.com/rchain/rchain/pull/3459#discussion_r661287349. 
I don't think we can recover directly finalized blocks before LFB read from legacy storage (as there is a record only for the latest one).

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
nzpr: I did not change logic here, don't want touch too much. You cannot get finalized set until you get LFB. So you need to push everything in `addBlockToDagState`, which to me looks like a complication.

models/src/main/scala/coop/rchain/models/BlockMetadata.scala
nzpr: Im not sure - do you want to change something here or you are ok?

tgrospic: Change parameters to include directly finalized flag. So the caller should be aware that finalized info is not read from block itself.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
tgrospic: BlockMetaData, BlockMetaDataStore, in memory representation are different levels of indexing what ever is stored in _raw_ blocks. So we need to design this index in a way that simplify our code and increase performance.

With that in mind, we can put in BlockMetaData what we need. E.g. we can add both flags for directly and indirectly finalized blocks which will reduce amount of traversals and creation of in memory index.

nzpr: Implemented in https://github.com/rchain/rchain/pull/3459/commits/208c4fe33d5963252631f1902b4dd9c58455acda

models/src/main/protobuf/CasperMessage.proto
nzpr: Indirectly finalized persisting in metadata done in https://github.com/rchain/rchain/pull/3459/commits/208c4fe33d5963252631f1902b4dd9c58455acda

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3459/commits/208c4fe33d5963252631f1902b4dd9c58455acda

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -248,11 +249,11 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def accessEquivocationsTracker[A](f: EquivocationsTracker[F] => F[A]): F[A] =
     lock.withPermit(f(KeyValueStoreEquivocationsTracker))
 
-  def addFinalizedBlockHash(blockHash: BlockHash): F[Unit] =
-    blockMetadataIndex.addFinalizedBlock(blockHash)
+  def addFinalizedBlockHashes(hashes: List[BlockHash]): F[Unit] =
```
nzpr: I made this accept list because indirectly finalized blocks are processed always in batches. And it's better to have an atomic update on KVStore level.

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
```diff
@@ -118,6 +117,10 @@ object Engine {
         init,
         disableStateExporter
       )
+      // this is required because before https://github.com/rchain/rchain/pull/3459 genesis as not added to
+      // last finalized storage on network init.
+      _ <- BlockDagStorage[F].recordDirectlyFinalised(approvedBlock.candidate.block.blockHash)
+      _ <- BlockDagStorage[F].addFinalizedBlockHashes(List(approvedBlock.candidate.block.blockHash))
```
tgrospic: Can we do this as part of migration?

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -98,6 +97,9 @@ class Initializing[F[_]
               s"Valid approved block ${PrettyPrinter.buildString(block, short = true)} received. Restoring approved state."
             )
 
+        // Record approved block in DAG
+        _ <- BlockDagStorage[F].insert(block, invalid = false, approved = true)
```
tgrospic, line 101: This insert should be after line 108 where is insert to BlockStore. We don't want to save approved block before LFS state is completely downloaded.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
```diff
@@ -14,8 +14,16 @@ final class KeyValueTypedStoreOps[F[_], K, V](
     // KeyValueTypedStore extensions / syntax
     private val store: KeyValueTypedStore[F, K, V]
 ) extends AnyVal {
+  def errKVStoreExpectValue = "Error when reading from KeyValueStore: value not found."
+
   def get(key: K)(implicit f: Functor[F]): F[Option[V]] = store.get(Seq(key)).map(_.head)
 
+  def getUnsafe(keys: Seq[K])(implicit f: Sync[F]): F[List[V]] =
```
tgrospic: I'm little sceptical about this unsafe calls without any indication of used keys.

Maybe just simple `"$key"` would be better then nothing.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -64,8 +76,28 @@ object BlockMetadataStore {
...
+        curMeta <- store.getUnsafe(hash)
+        _       <- store.put(hash, curMeta.copy(directlyFinalized = true)) // update persistent meta
+        _ <- dagState.modify(
+              st =>
+                if (st.lastFinalizedBlock.exists { case (_, height) => height > curMeta.blockNum })
+                  st
+                else // update inmem state if new LFB higher then existing
+                  st.copy(lastFinalizedBlock = (hash, curMeta.blockNum).some)
+            )
+      } yield ()
```
tgrospic, line 110: It would be nice to combine these two functions and have atomic update of all finalized blocks. :thinking:

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
nzpr: We don't have approved block when migration happens - it is done even before blockdag storage is created, which is required for Casper when approved block is available.

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
nzpr: It's not Approved block its just normal block insertion, same as done inside `requestApprovedState`, but insert with approved flag. `LastApprovedBlock[F].set(approvedBlock)` is on line 109, without it restoring will be not complete and start again after restart.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
nzpr: Maybe force is to have implicit `Show[K]`?

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -119,6 +119,8 @@ message BlockMetadataInternal {
   int64 blockNum                        = 6;
   int32 seqNum                          = 7;
   bool invalid                          = 8; // whether the block was marked as invalid
+  bool directlyFinalized                = 9; // whether the block has been last finalized block (LFB)
+  bool finalized                        = 10;// whether the block is finalized
```
zsluedem, line 123: Why do we need two fields for that?
What is the goal of `directlyFinalized` trying to solve?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
nzpr: Now I'm thinking that its quite odd that we run finalization on DagRepresentation, and then get back to DagStorage to update it, and DagStorage actually might have been already updated in between.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
```diff
@@ -14,8 +14,16 @@ final class KeyValueTypedStoreOps[F[_], K, V](
     // KeyValueTypedStore extensions / syntax
     private val store: KeyValueTypedStore[F, K, V]
 ) extends AnyVal {
+  def errKVStoreExpectValue = "Error when reading from KeyValueStore: value not found."
```
zsluedem: Please provide more useful info on this error message. At least include the `key` value.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
```diff
@@ -37,6 +42,8 @@ trait BlockDagRepresentation[F[_]] {
       startBlockNumber: Long,
       maybeEndBlockNumber: Option[Long]
   ): F[Vector[Vector[BlockHash]]]
+  // DAG representation has to have finalized block, or it does not make sense
+  def lastFinalizedBlock: BlockHash
```
zsluedem, line 45: I think this can be a syntax method and calculated by just traversing the dag from latestBlock to the latest finalized block. 
It is fine to leave it here. Just mention my idea.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedStorage.scala
```diff
@@ -6,6 +6,10 @@ import coop.rchain.models.BlockHash.BlockHash
 trait LastFinalizedStorage[F[_]] {
   def put(blockHash: BlockHash): F[Unit]
   def get(): F[Option[BlockHash]]
+
+  /** If LFB is migrated to BlockDagStorage metadata */
```
zsluedem: Question:
Do you mean `/** If LFB is migrated to BlockDagStorage metadata , we can remove this one*/`?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -322,4 +330,47 @@ object BlockDagKeyValueStorage {
...
+      _ <- Log[F].info("Migration of LFB done.")
+
+      // record finalized blocks
+      finalizedBlockSet = DagOps
+        .bfTraverse(List(lfb)) { bh =>
+          blocksInfoMap.get(bh).map(_.parents.toList).getOrElse(List.empty)
+        }
+        .toList
+      curVs <- finalizedBlockSet.traverse { h =>
+                blockMetadataDb.get(h).flatMap(_.liftTo[F](new Exception(errNoMetadataForFb)))
```
zsluedem: Not sure about how the performance of migrating the database. If that takes too long. We'd better parallel reading here to be faster.

models/src/main/protobuf/CasperMessage.proto
nzpr: `directlyFinalized` means it has been output by finalizer, which means it has been last finalized block at some point. And LFB is used to constrain merging scope, we need at least two (well, exactly two I guess) merging scopes to reliably check if deploy cannot be rejected.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
nzpr: We are storing `lastFinalizedBlock` in metadata, but what you describe is how it's calculated. DagRepresentation should have lastFinalizedBlock or it does not make sense because at least genesis or some block should be initial point of the DAG evolution.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedStorage.scala
nzpr: It's more `Whether LFB  should be migrated to BlockDagStorage`

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
nzpr: I'd better not touch it now, its quite a rabbit whole

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -249,11 +248,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
-
-  def recordDirectlyFinalised(blockHash: BlockHash): F[Unit] =
-    blockMetadataIndex.recordDirectlyFinalised(blockHash)
+  /** Record that some hash is directly finalized (detected by finalizer and becomes LFB). */
+  def recordDirectlyFinalized(
+      directlyFinalizedHash: BlockHash,
+      finalizationEffect: Set[BlockHash] => F[Unit]
+  ): F[Unit] =
+    for {
+      dag    <- getRepresentation
```
nzpr: I think it makes sense to have only one method on BlockDagStorage - to record directly finalized hash. And it should compute what else should be finalized indirectly.

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
tgrospic: Yes, you are right! Although BlockStore is checked for approved block in launch of Casper, LastApprovedBlock[F] is in memory.
https://github.com/rchain/rchain/blob/1370ee414b4336fc273b3d6d73014666dc331330/casper/src/main/scala/coop/rchain/casper/engine/CasperLaunch.scala#L71-L72

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
```diff
@@ -118,6 +117,10 @@ object Engine {
         init,
         disableStateExporter
       )
+      // this is required because before https://github.com/rchain/rchain/pull/3459 genesis as not added to
+      // last finalized storage on network init.
+      _ <- BlockDagStorage[F]
+            .recordDirectlyFinalized(approvedBlock.candidate.block.blockHash, _ => ().pure[F])
```
tgrospic: Why this is not already covered by migration? Approved block will be migrated through parents traversal. Am I missing something?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
tgrospic: I'd add to read in chunks because this will load all DAG store with around ~5GB data. Also the whole chunk can be loaded with list overload of _get_.
Sorted keys can also speed up write but I'm not completely sure if ordering is the same.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -322,4 +339,47 @@ object BlockDagKeyValueStorage {
       stores.invalidBlocks,
       stores.equivocations
     )
+
+  def migrateLfb[F[_]: Sync: Log](
```
tgrospic: This function should be defined on `LastFinalizedKeyValueStorage` instance which is used in Setup and will be deleted with it.

block-storage/src/main/scala/coop/rchain/blockstorage/finality/LastFinalizedStorage.scala
```diff
@@ -6,6 +6,10 @@ import coop.rchain.models.BlockHash.BlockHash
 trait LastFinalizedStorage[F[_]] {
   def put(blockHash: BlockHash): F[Unit]
   def get(): F[Option[BlockHash]]
+
+  /** Methods below are for migration of finalized block from LastFinalizedStorage to BlockDagStorage */
+  def requireMigration: F[Boolean]
+  def recordMigrationDone: F[Unit]
```
tgrospic: This comment is more for general situations with migrations. Methods don't need to be defined on interface because usually migration will be done in some initialization stage where we have concrete instance anyway.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
zsluedem: You understand me correctly.
What I meant ultimately is that, maybe we don't need to store lfb. Just an idea. Not a request.

models/src/main/protobuf/CasperMessage.proto
zsluedem: :) I still do not quite understand the scenario. Maybe I don't understand from the point of choosing merging scope by LFB from the beginning which makes me understand the `directlyFinalized` difficulty.

`All the finalized blocks should be last finalized block at some poing.` This is wrong, isn't it from your view ?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
zsluedem: That's better

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Makes sense, thanks

nzpr: I found it more simple to have it in BlockDagStorage, as I'm not constrained with API. But ok, I will move it to finalized storage.

nzpr: Actually its hard to do that in LastFinalizedKeyValueStorage, because API that BlockDagStorage offers is to write the whole finalization at once https://github.com/rchain/rchain/pull/3459#discussion_r662431601, so it requires calling representation. And representation cannot be created because there is no lastFinalizedBlock (as it has not been migrated)

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
nzpr: When new network is started, or node is bootstrapped from LFB - approved block has to be marked as directly finalized.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: Do you have an idea what chunk size would be preferable?

models/src/main/protobuf/CasperMessage.proto
nzpr: `All the finalized blocks should be last finalized block at some poing` yes this is wrong. E.g. LFB1 can be at height 1 and next one, LFB2 - at height 5. So all blocks that are ancestors of LFB2 should be finalized as well. But they are not found by Finalizer according to clique oracle rule.

In this scenario between LFB1 and LFB2 there will be multiple merge blocks, all of them using LFB1 as a base for merge (state that cannot be changed).

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
nzpr: I think we need to store this only for merging purposes. Actually only two the most recent LFBs are needed. But might be it worth to store them all though, not that much data.

nzpr:
Description is updated with performance measurements.

casper/src/test/scala/coop/rchain/casper/batch2/FinalizerTest.scala
```diff
@@ -138,16 +130,12 @@ class FinalizerTest extends FlatSpec with Matchers with BlockGenerator with Bloc
                 dag,
                 faultToleranceThreshold = -1,
                 currLFBHeight = 0L,
-                m => (lfbStore = m).pure[Task],
-                m => finalisedStore.update(v => v ++ m)
+                m => (lfbStore = m).pure[Task] <* finalisedStore.update(v => v + m)
```
nzpr, line 133: Finalizer test is adjusted, because now finalizer only cares about single effect - when new LFB is found. How to find hashes that have to be finalized indirectly (and finalize them) - now is the task for BlockDagStorage.
```
def recordDirectlyFinalized(
      directlyFinalizedHash: BlockHash,
      finalizationEffect: Set[BlockHash] => F[Unit]
  ): F[Unit] =
```

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: 10k looks reasonable, ptal at description.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -322,4 +340,66 @@ object BlockDagKeyValueStorage {
...
+
+        for {
+          curVs <- blockMetadataDb.getUnsafeBatch(chunk)
+          // WARNING: migration should be done before block merge, as it assumes all blocks are directly finalized.
+          newVs = curVs.map(_.copy(directlyFinalized = true, finalized = true))
+          _     <- blockMetadataDb.put(newVs.map(v => (v.blockHash, v)))
+        } yield ()
+      }
+
+      chunkSize = 10000L
```
zsluedem: Cool with the optimizations.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -62,6 +65,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
     def invalidBlocks: F[Set[BlockMetadata]] = invalidBlocksSet.pure[F]
 
+    override def lastFinalizedBlock: BlockHash = lastFinalizedBlockHash
```
zsluedem, line 65: I got another question about this new concept on finalization. Is it possible the last finalized block could be several blocks in the new definitions instead of only one last finalized block in single parent mode ?

nzpr: LFB search area is constrained by blocks that are on intersection of main parent chains of tips holding more then 50% of the stake. So it always can be only one block at particular height, whether it's single or multi parent.

block-storage/src/test/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorageTest.scala
```diff
@@ -264,4 +266,76 @@ class BlockDagKeyValueStorageTest extends BlockDagStorageTest {
       }
     }
   }
+
+  "recording of new directly finalized block" should "record finalized all non finalized ancestors of LFB" in
```
nzpr, line 270: This test is created instead of adjusted FinalizerTest, because what exactly should be finalized having new directly finalised hash - is defined by content of BlockDagStorage. As it is some mutable DAG that represents the most recent view.

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
tgrospic: Received LFB is inserted as approved in Initializing engine.
https://github.com/rchain/rchain/blob/1370ee414b4336fc273b3d6d73014666dc331330/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala#L100-L101

And this is called by genesis master.
https://github.com/rchain/rchain/blob/f2c030b0a906e26343211daad214411bb0add306/casper/src/main/scala/coop/rchain/casper/engine/Engine.scala#L63-L71

Is there some other case?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -223,19 +228,39 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+      dag    <- representation
+      errMsg = s"Attempting to finalize nonexistent hash ${PrettyPrinter.buildString(directlyFinalizedHash)}."
+      _      <- dag.contains(directlyFinalizedHash).ifM(().pure, new Exception(errMsg).raiseError)
+      // all non finalized ancestors should be finalized as well (indirectly)
+      indirectlyFinalized <- dag.ancestors(directlyFinalizedHash, dag.isFinalized(_).not)
+      // invoke effects
+      _ <- finalizationEffect(indirectlyFinalized + directlyFinalizedHash)
+      // persist finalization
+      _ <- blockMetadataIndex.recordFinalized(directlyFinalizedHash, indirectlyFinalized)
+    } yield ()
```
tgrospic: This is to dangerous to be without locking. I cannot say the use case right now where can fail but in parallel replay finalization can be called on overlapping blocks.

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
nzpr: If network is started and no finalized block has been found, genesis is not recorded as finalized. On switching to this migration build and reconnecting there will be no finalized block.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: But why this is a problem? Data put is the same. I see that this might race with some another change in another part of the code, if metadata adjusted between get and put. But thats a general concern to KVStore not this particular code?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
```diff
@@ -64,8 +74,41 @@ object BlockMetadataStore {
...
+                  st.copy(finalizedBlockSet = newFinalizedSet)
+                else
+                  st.copy(
+                    finalizedBlockSet = newFinalizedSet,
+                    lastFinalizedBlock = (directly, curMetaForDF.blockNum).some
+                  )
+              }
+            )
+        // persist new values all at once
+        _ <- store.put(newMetaForDF +: newMetasForIF)
```
tgrospic, line 109: Connected to this comment.
https://github.com/rchain/rchain/pull/3459/files#r663200787

Between store read and write, another thread can call the same function and it's possible to overwrite others data.

I know that we don't have any other update of BlockMetaData and now is safe. But until we have "real" transactions this is just as precaution.

nzpr: ~~Not sure wdy mean by "real" transactions?~~ I see, looks like you mean transactional update

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
tgrospic: Sorry, but I still don't understand why not fix this as part of migration. We have approved block in BlockStore and we can add it to BlockMetaData store.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
nzpr: https://github.com/rchain/rchain/pull/3459/commits/6b5d545bb0a07ec63bbdf28aafa1080cd48fbc98

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3459/commits/6b5d545bb0a07ec63bbdf28aafa1080cd48fbc98

casper/src/main/scala/coop/rchain/casper/engine/Engine.scala
nzpr: Makes sense, https://github.com/rchain/rchain/pull/3459/commits/cf6106b2b00ee6687706e192bc62876ca429bc61

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockMetadataStore.scala
tgrospic: Yes, exactly.

In pseudo Rholang code. :)
```scala
for(curMetaForDF  <- store.getUnsafe(directly) &
    curMetasForIF <- store.getUnsafeBatch(indirectly.toList)) {
  // Loaded keys (channels) are locked
  // ...
  store.put(newMetaForDF +: newMetasForIF)
  // Lock is released
}
```

tgrospic (approved):
Awesome PR! Thanks for addressing all comments.

zsluedem (approved):
LGTM

nzpr:
bors merge

PR #3462, opened 2021-07-03
Fix genesis blockhash validation bug

zsluedem:
## Overview

After https://github.com/rchain/rchain/pull/3390, if we config genesis with a non-zero block number, the validator who received this genesis block can not validate the block hash because the hashing block detail is different and validator would show the warning below.

```
rnode.readonly | 01:56:12.867 [WARN ] [node-runner-26      ] [coop.rchain.casper.Validate$ ] - Ignoring block b63f133db5... because block hash b63f133db5... does not match to computed value e25b677348....
rnode.readonly | 01:56:12.870 [WARN ] [node-runner-26      ] [c.r.c.e.LfsBlockRequester$   ] - Received Block #10 (b63f133db5...) with empty parents (supposedly genesis) with invalid hash. Ignored block.
```

This pr is intended to make all block hashing consistent and I think we **should not make any difference between hashing signed block and unsigned block**.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
+++ b/casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
@@ -137,6 +137,7 @@ object BlockCreator {
                 _             <- Span[F].mark("before-packing-block")
                 shardId       = s.onChainState.shardConf.shardName
                 casperVersion = s.onChainState.shardConf.casperVersion
+                // unsignedBlock got blockHash(hashed without signature)
                 unsignedBlock = packageBlock(
                   blockData,
                   parents.map(_.blockHash),
@@ -150,7 +151,9 @@ object BlockCreator {
                   shardId,
                   casperVersion
                 )
-                _           <- Span[F].mark("block-created")
+                _ <- Span[F].mark("block-created")
+                // signedBlock add signature and replace hashed-without-signature
+                // blockHash to hashed-with-signature blockHash
                 signedBlock = validatorIdentity.signBlock(unsignedBlock)
                 _           <- Span[F].mark("block-signed")
               } yield BlockCreatorResult.created(signedBlock)
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -278,12 +278,12 @@ object ProtoUtil {
       extraBytes = ByteString.EMPTY
     )
 
-    val hash = hashSignedBlock(block)
+    val hash = hashBlock(block)
 
     block.copy(blockHash = hash)
   }
 
-  def hashSignedBlock(blockMessage: BlockMessage): BlockHash =
+  def hashBlock(blockMessage: BlockMessage): BlockHash =
     ProtoUtil.hashByteArrays(
       blockMessage.header.toProto.toByteArray,
       blockMessage.body.toProto.toByteArray,
--- a/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
@@ -278,16 +278,9 @@ object ProtoUtil {
       extraBytes = ByteString.EMPTY
     )
 
-    val hash = hashUnsignedBlock(block)
+    val hash = hashSignedBlock(block)
 
     block.copy(blockHash = hash)
-  }
-
-  def hashUnsignedBlock(blockMessage: BlockMessage): BlockHash = {
-    val toHash = blockMessage.header.toProto.toByteArray +: blockMessage.justifications.map(
-      _.toProto.toByteArray
-    )
-    hashByteArrays(toHash: _*)
   }
 
   def hashSignedBlock(blockMessage: BlockMessage): BlockHash =
--- a/casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
@@ -220,12 +220,13 @@ final class RhoRuntimeOps[F[_]: Sync: Span: Log](
 
   def computeGenesis(
       terms: Seq[Signed[DeployData]],
-      blockTime: Long
+      blockTime: Long,
+      blockNumber: Long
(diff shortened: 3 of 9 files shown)


casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
```diff
@@ -220,12 +220,13 @@ final class RhoRuntimeOps[F[_]: Sync: Span: Log](
...
       terms: Seq[Signed[DeployData]],
-      blockTime: Long
+      blockTime: Long,
+      blockNumber: Long
   ): F[(StateHash, StateHash, Seq[ProcessedDeploy])] =
     Span[F].traceI("compute-genesis") {
       for {
         _ <- runtime.setBlockData(
-              BlockData(blockTime, 0, PublicKey(Array[Byte]()), 0)
+              BlockData(blockTime, blockNumber, PublicKey(Array[Byte]()), 0)
```
zsluedem, line 229: The fix here is not actually the real problem. I just want to change it to be consistent.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -307,19 +307,12 @@ object ProtoUtil {
       extraBytes = ByteString.EMPTY
     )
 
-    val hash = hashUnsignedBlock(block)
+    val hash = hashBlock(block)
 
     block.copy(blockHash = hash)
   }
 
-  def hashUnsignedBlock(blockMessage: BlockMessage): BlockHash = {
```
zsluedem, line 286: @tgrospic @nzpr I made the decision by myself to remove this method forever but I am open to see if you guys get any other ideas. I don't think we should keep two hashing methods for the same `BlockMessage` object.

tgrospic: I agree with this change. Please also fix validation function for LFS which was I left open until we unify hashing.
https://github.com/rchain/rchain/blob/1370ee414b4336fc273b3d6d73014666dc331330/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala#L207-L214

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -150,7 +151,9 @@ object BlockCreator {
                   shardId,
                   casperVersion
                 )
-                _           <- Span[F].mark("block-created")
+                _ <- Span[F].mark("block-created")
+                // signedBlock add signature and replace hashed-without-signature
+                // blockHash to hashed-with-signature blockHash
```
tgrospic, line 156: This whole confusion seems to be because _sender_ is not added for genesis block, which doesn't make sense why? Genesis master created the genesis block and can be added as a sender and sign. Other genesis validators will validate this block in the same way as any other block.

Also I couldn't find the reason why signing algorithm is included for hashing. In verification is only used here, selected from hard coded algorithms.
https://github.com/rchain/rchain/blob/a37ff7ef5172a370621ab415a33c39a33126030a/casper/src/main/scala/coop/rchain/casper/Validate.scala#L81-L85
Another questions are. Why signing algorithm is included in the block and why signing algorithm is stored and not hashing algorithm how is used with enveloped/detached signatures.
https://en.wikipedia.org/wiki/XML_Signature
https://www.researchgate.net/figure/4-Enveloped-enveloping-and-detached-signatures_fig12_33760566

Every block is hashed without a signature. Signature means signing of the block hash.
`BlockCreator#packageBlock` can accept sender and create block hash (the one and only) and signing should just add _signature_.

But this seems to much for now. I'll create an issue for HF2.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
zsluedem: I didn't actually notice here. We didn't actually validate genesis blockHash before.

zsluedem:
```scala
 private def validateBlock(block: BlockMessage): F[Boolean] = { 
     Validate.blockHash(block).map(_ == Right(Valid)) 
 } 
```
like this?

I am not sure whether we should change this or not. I got an ultimate question for this. Should we maintain the old network for user to replay and validate?

casper/src/main/scala/coop/rchain/casper/util/rholang/CasperRhoRuntimeSyntax.scala
tgrospic: Good catch!

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
tgrospic: Yes, that is the fix, now we can validate all block in the same way.

> We didn't actually validate genesis blockHash before.

Yes, but LFS on main net was not downloading genesis block anyway.

> I am not sure whether we should change this or not. I got an ultimate question for this. Should we maintain the old network for user to replay and validate?

Yes, but to work with history networks older versions of node should be used. Similar how latest Ethereum node cannot process all previous blocks.

zsluedem: > Yes, that is the fix, now we can validate all block in the same way.

The existing genesis block https://revdefine.io/#/blocks/986addc3dfa12b179eaa40e38d77aec3da0530b9ca2243271436a135055229dd/ can not be validated this way.

tgrospic: Do you see situation where this can be a problem? We have v0.10.2 which can.

We can include this PR with HF1 release?

nzpr:
~~PR looks good, I' just don't get, why `Received Block #10 (b63f133db5...) with empty parents (supposedly genesis) with invalid hash`? Why block 10 does not have parents? It's just a normal block, right, and it's hash should include parents?~~ Oh, wait, its nit about LFS, it's starting the network from non zero block. Never mind.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
tgrospic: PR is now rebased on `feature/hard-fork-1`.

tgrospic (approved):
LGTM

PR #3464, opened 2021-07-04
Simplify bonds and wallets file parser

### LINKED ISSUE #3460: Change the format of wallets.txt file necessary for Hard Fork 1
## Overview

Current format of `wallets.txt` file used for initial REV balance on genesis is using ETH address format. That was necessary for transfer of RHOC from Ethereum network to start with REV on RChain network.

ETH addresses are not used anymore so `wallets.txt` file should accept REV address format.

Line format should be:
```
<REV address 1>,<balance>
<REV address 2>,<balance>
```

Third parameter for the type of address (private key, contract) is not relevant anymore. And also for REV addresses we cannot find this information from address itself.

### Impact

`wallets.txt` file is used only for creation of a genesis block for fresh network so there is no impact on existing network.

tgrospic:
## Overview

This PR simplify reading/parsing bonds.txt and wallets.txt files. File menagement is done using fs2.io which enables removal of custom code for file operations.

Handling of parsing errors is improved and more specific regex is used for line parsing.

This is also preparation for Hard Fork 1 changes for wallets.txt file #3460.


casper/src/main/scala/coop/rchain/casper/util/BondsParser.scala
```diff
@@ -1,113 +1,132 @@
...
+        .evalTap {
+          case (pk, stake) => Log[F].info(s"Bond loaded ${Base16.encode(pk.bytes)} => $stake")
+        }
+        .compile
+        .to(Map)
+        .adaptErr {
+          case ex: Throwable =>
+            new Exception(s"FAILED PARSING BONDS FILE: $bondsPath\n$ex")
+        }
+    Blocker[F].use(readLines)
```
zsluedem, line 62: Question.
In cat-effect 2, `Blocker[F].use` would submit this job into special thread pools which handle blocking IO, right?

zsluedem (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/util/BondsParser.scala
tgrospic: Yes, `Blocker[F]` creates a new scheduler with infinite thread pool and its threads are used inside `Blocker[F].use`. I didn't want to create instance on object level so that after the call it will be disposed with GC. 
https://github.com/typelevel/cats-effect/blob/v2.3.1/core/jvm/src/main/scala/cats/effect/internals/BlockerPlatform.scala#L31-L38

In my test of RSpace with blocker, monix io scheduler is created in main method. `Sync.blocking` will be more convenient.
https://github.com/rchain/rchain/blob/42e97c17578082ec7436069c1cf1d3a8417e99d0/node/src/main/scala/coop/rchain/node/Main.scala#L52-L54

https://github.com/rchain/rchain/pull/3441/files#diff-6862710cfeaf5d850532b0e40e2eb9de55451c1a9992c9c4ef1d17f379ee656fR52-R54

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/util/BondsParser.scala
nzpr: Why not run it on main scheduler I'm curious?

tgrospic: Running on main scheduler (which has fixed number of Java thread reserved) means that we _loose_ a thread blocked even though CPU is not used at all because we are waiting on disk IO to return and thread continues. Monix cannot _reuse_ the thread because there is no _flatMap_ to suspend the Task as light abstraction above Java thread.

Cats Effect 3 has good explanation about blocking.
https://typelevel.org/cats-effect/docs/schedulers#handling-blocking

tgrospic:
bors merge

casper/src/main/scala/coop/rchain/casper/util/BondsParser.scala
nzpr: I see, I mean - this code is something that is run very rarely, when genesis is created. So as I get this is used here to make it proper way whether it is used a lot or not.

PR #3394, opened 2021-04-30
[WIP] History multi map

tgrospic:
## Overview
This PR changes encoding and hashing of channels in History implementation to be compatible between channels encoded in event log and storage (removing the _channel map_).

This PR depends on PR for initial preparation #3401. Completion of this PR is done in #3474.

This schema shows graphically what is changing on _history_ implementation level. Instead of storing one map with prefixes on keys we are moving to storing multiple maps with prefixes as children of the root node.

image

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3474, opened 2021-07-12
Remove channels map

### LINKED ISSUE #3346: Improve concurrent block validation (optimize or remove channels map)

### LINKED ISSUE #3384: Make encoding algorithm for channel hash consistent
## Current shortage of encoding

Despite of the protobuf encoding method we used for the `Channel Type` which is [Par](https://github.com/rchain/rchain/blob/13be1a71b8f59016179af832a0a00d7a0a96b282/models/src/main/protobuf/RhoTypes.proto#L31-L42), there are differences between the encoding of the [EventLog Channel Hash](https://github.com/rchain/rchain/blob/13be1a71b8f59016179af832a0a00d7a0a96b282/rspace/src/main/scala/coop/rchain/rspace/StableHashProvider.scala) and the encoding of the [HistoryReposity Channel Hash Key](https://github.com/rchain/rchain/blob/13be1a71b8f59016179af832a0a00d7a0a96b282/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala#L63-L75) .

```

				  encodingAlg A	                               hashingAlgX
				  ---------------------> encodedType A---------------------------------->  channelHashA (used in BlockMessage)
				 /	                                                                                                                                   |
                                /                                                                                                                                         X
ChannelDataType                                                                                                                                      X can not convert channelHashA to channelHashB
				\                                                                                                                                         X
				 \  encodingAlg B                                       hashingAlgX with suffix                  v
				  ---------------------> encodedType B ---------------------------------->channelHashB (used in HistoryReposity)
```

Because of the differences above, we need to introduce [ChannelMap]() in block merging for accessing the data in HistoryRepo through `ChannelHash` in `BlockMessage`.

### The goal of the changes 

We need to change the encoding pattern to be like below.

```

                                     encodingAlgA                                            hashingAlgX                                                           
ChannelDataType -----------------------> encodedTypeA --------------------> channelHashA(used in BlockMessage) 
```
THEN
```

                                DatumHashAlg
                              ----------------------->   DatumChannelHashA(used in HistoryRepo)
                            /
                           /   ContinuationHashAlg
channelHashA ------------------------->  ContinuationChannelHashA(used in HistoryRepo)
                             \
                                \   JoinHashAlg
                                    -------------------->  JoinChannelHashA(used in HistoryRepo)
```

@tgrospic and @zsluedem  discussed this before and @tgrospic suggested putting some prefix bits in front of the `channelHashA` then we don't have to hash it again.  But it seems that this adding-prefix-bit way can to applied to `Joins` easily. 

After the changes are done, the channelMap should be removed.

### Impact

Need hard fork

### Dependencies

Nope

### Design

Need some discussions

nzpr:
## Overview
This PR is rebase of https://github.com/rchain/rchain/pull/3394 by @tgrospic  onto hard-fork-2 branch.
During hard fork 2 encoding of channels will be adjusted so channels in event logs are in alignment with channels in history.

### Notes
- https://github.com/rchain/rchain/pull/3394 had join mergeability tests failing, so this PR fixes that, refactoring while fixing.
- also this PR fixes one bug and makes state change compute more strict and behaves as it is expected, which makes PR author quite happy.


rspace/src/main/scala/coop/rchain/rspace/merger/StateChangeMerger.scala
```diff
@@ -1,176 +1,138 @@
...
+                           case (historyPointer, changes) =>
+                             mkTrieAction(
+                               historyPointer,
+                               baseReader.getJoins(_).map(_.map(_.raw)),
+                               changes,
+                               TrieDeleteJoins,
+                               TrieInsertBinaryJoins
+                             )
+                         }
+    } yield produceTrieActions ++ consumeTrieActions ++ joinsTrieActions
```
tgrospic, line 137: This is expensive to concat lists. Better to use Vector here.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala
```diff
@@ -128,8 +128,22 @@ object StateChange {
...
+                .map { case (_, ChannelChange(add, del)) => add.isEmpty && del.isEmpty }
+                .exists(_ == true)
+            )
       consumeChanges <- kontsDiffRef.get
-      joinsMap       <- joinsMapRef.get
+      _ <- new Exception("State change compute logic error: empty channel change for consume.")
+            .raiseError[F, StateChange]
+            .whenA(
+              consumeChanges
+                .map { case (_, ChannelChange(add, del)) => add.isEmpty && del.isEmpty }
```
tgrospic, line 143: Where this can be created with empty channels?

rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala
```diff
@@ -350,12 +358,16 @@ object HistoryInstances {
...
+//                       val key = ByteVector(remainingPath)
+//                       println(s"$index INSERT $key")
                        insert(currentRoot, previousModificationOpt, remainingPath, value)
 
                      case (
                          (currentRoot, previousModificationOpt),
                          DeleteAction(remainingPath)
                          ) =>
+//                       val key = ByteVector(remainingPath)
+//                       println(s"$index DELETE $key")
```
tgrospic: Please delete my debugging leftovers. There is also print on line 282.

This special case is resolved, it can happen when last node in the trie is deleted so `case (_, EmptyTrie)` just return None.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala
```diff
@@ -128,8 +128,22 @@ object StateChange {
             .compile
             .drain
       produceChanges <- datumsDiffRef.get
+      _ <- new Exception("State change compute logic error: empty channel change for produce.")
```
nzpr, line 131: The logic behind this exception is that we have to (and are) able to determine that channel is adjusted by just looking into event log. And now I see that this works as expected: only necessary events are used to compute state change, an  changes on channels involved are always non empty. Which makes sure that `consumesAffected ` and `producesAffected`  from MergingLogic are correct, which gives more confidence that overall logic does not have flaws.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala
nzpr: for consumes - here, if data in preState and postState is equal. `change` will be empty
https://github.com/rchain/rchain/blob/a2883c1776a1d4325102e2ba1edf082cc960f8be/rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala#L94-L111

tgrospic (approved):
LGTM

PR #43, opened 2017-11-04
Reverse output of twoZip implementation to be what is expected

KentShikama:
Resolves https://rchain.atlassian.net/browse/ROS-262

Issue #3432, opened 2021-06-12
Implement traverse method in rholang treeHashMap which returns all the keys and values


zsluedem:
## Overview

Currently the [holang treeHashMap](https://github.com/rchain/rchain/blob/9899c89b91f81ddc126545ee3c805ba4f90fdb08/casper/src/main/resources/Registry.rho) got `set`, `contains`, `get`, `fastUnsafeGet`, `update` for basic operation in treeHashMap.

We still need a method which return all the key values from the map.

Currently we got a [scala implementation](https://github.com/rchain/rchain/blob/9899c89b91f81ddc126545ee3c805ba4f90fdb08/node/src/main/scala/coop/rchain/node/revvaultexport/RhoTrieTraverser.scala) for doing that but we'd better also got something like this in rholang which dapp developer can benefit.

### Impact

Because the speed of rholang execution is slow, this method should take very long and consume a lot of resource on it.

### Dependencies

No

### Design

The design can borrow the idea from [scala version](https://github.com/rchain/rchain/blob/9899c89b91f81ddc126545ee3c805ba4f90fdb08/node/src/main/scala/coop/rchain/node/revvaultexport/RhoTrieTraverser.scala)

zsluedem:
Close it.

PR #3485, opened 2021-07-19
System contract updates

Isaac-DeFrain:
## Overview
This PR introduces the soft-fork mechanism outlined in [RCHIP #43](https://github.com/rchain/rchip-proposals/issues/43).

1. remove the old *logCh* logic in purse
2. introduce `rho:registry:systemContractManager` which coop could  update contract without hard-fork
3. introduce `rho:rchain:configPublicKeyCheck` which check if the deployID is coop key
4. wrap `AuthKey`, `MakeMint`, `MultiSigRevVault`, `Pos`, `RevVault` with dispatcher

### Uncompatible rholang change

**Only for `rho:lang:treeHashMap`** , other would stay the same.
rholang code like 
```
new rl(`rho:registry:lookup`), ret in {
  rl!(`rho:lang:treeHashMap`, *ret) |
  for (@TreeHashMap <-ret ) {
    do_something
  }
}
```

Needs to change into 
```
new rl(`rho:registry:lookup`), ret in {
  rl!(`rho:lang:treeHashMap`, *ret) |
  for (@(nonce, TreeHashMap) <-ret ) {
    do_something
  }
}
```


zsluedem:
~~https://github.com/rchain/rchain/issues/3432 is needed to extract the **RevVault.rho** state.~~ no need

zsluedem:
The state we need to extract from the contract:

1. AuthKey.rho
None(we would fix the authkey reusable problem in https://github.com/rchain/rchain/issues/3668)

2. Either.rho
None(Function contract which doesn't have a state)

3. ListOps.rho
None(Function contract which doesn't have a state)

4. MakeMint.rho
None(It is a dependency package for RevVault.rho. From `MakeMint.rho` inside, there is no way to reach all the generated unf names.)

5. RevVault.rho
[vaultMap](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/RevVault.rho#L40) and [_revvault](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/RevVault.rho#L24)

6. MultiSigVault.rho
[_multisigvault](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/MultiSigRevVault.rho#L21)

7. NonNegative.rho
[MergeableTag](https://github.com/zsluedem/rchain/blob/update-pos/casper/src/main/resources/NonNegativeNumber.rho#L20)

8. POS.rho
https://github.com/zsluedem/rchain/blob/update-pos/casper/src/main/resources/PoS.rhox#L323-L326

9. Registry.rho
There are 3 key function in `Registry.rho` -> TreeHashMap, Registry(insertArbitrary, insertSigned, BlessedContractMultiSig )

a. Registry doesn't have extractState function
b. treeHashMap hold [registryMap](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/Registry.rho#L363) and [storeToken](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/Registry.rho#L67) as its state.
c. multisig extracts `agreementMap`, `proposalMap`, `blessedMap`

node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
```diff
@@ -56,11 +56,6 @@ object StateBalanceMain {
 
   // hard-coded value in RevVault.rho
   val genesisVaultMapDepth = 2
-  // The way to get this Unforgeable name needs reporting casper to get all the concrete comm events.
-  // Anyway, as long as RevVault.rho and genesis doesn't change, this value would be fixed.
-  val genesisVaultMapPar: Par = GPrivate(
-    "5e8200d74ba8689013cf86f075775a18a42f1bd3a5ea122e82dfca9ba485a924".unsafeHexToByteString
-  )
```
zsluedem, line 63: The `vaultMap` is no longer a fixed unforgeable name because of the changes we made. The future testnet and mainnet could be totally different. It depends on the genesis.

node/src/main/scala/coop/rchain/node/revvaultexport/StateBalances.scala
```diff
@@ -22,10 +36,33 @@ import scala.concurrent.ExecutionContext
...
+    val bundleRevVault     = Par(bundles = Seq(Bundle(body = revVault, writeFlag = true)))
+    val extractStateString = Par(exprs = Seq(Expr(GString("extractState"))))
+    val e                  = Par(exprs = Seq(Expr(ETupleBody(ETuple(Seq(bundleRevVault, extractStateString))))))
+    for {
+      c <- runtime.getContinuation(Seq(e))
+      unf = c.head.continuation.taggedCont.parBody.get.body.sends.head.data.head.exprs.head.getEMapBody.ps
+        .get(Par(exprs = Seq(Expr(GString("vaultMap")))))
+        .get
+    } yield unf
+  }
```
zsluedem, line 60: Get the `vaultMap` based on the newlly introduced `extractState`

casper/src/test/scala/coop/rchain/casper/util/GenesisBuilder.scala
```diff
@@ -61,6 +61,37 @@ object GenesisBuilder {
...
+          posMultiSigPublicKeys = defaultPosMultiSigPublicKeys,
+          posMultiSigQuorum = defaultPosMultiSigPublicKeys.length - 1
+        ),
+        vaults = genesisVaults.toList.map {
+          case (p, s) => Vault(RevAddress.fromPublicKey(Secp256k1.toPublic(p)).get, s)
+        },
+        supply = Long.MaxValue,
+        blockNumber = 0
+      )
+    )
```
zsluedem, line 94: I was trying to refactor codes in this file but end up with a lot of unrelated changes and failing tests. I would just keep it here and see if casper change need to consider this.

casper/src/main/resources/Registry.rho
```diff
@@ -332,115 +343,518 @@ in {
...
+      // -------------------------------------------------------------------------------------
+      {
+        let @quorumSize, @pubKeys <- 2, Set(
+          // The private keys and public keys are here just for testing now
+          // fc743bd08a822d544bfbe05a5663fc325039a44c8f0c7fbea95a85517da5c36b private
+          "04e2eb6b06058d10b30856043c29076e2d2d7c374d2beedded6ecb8d1df585dfa583bd7949085ac6b0761497b0cfd056eb3d0db97efb3940b14c00fff4e53c85bf".hexToBytes(),
+          // 6e88cf274735f3f7f73ec3d7f0362c439ab508427682b5bd788007aca665d810 private
+          "043405f36c6cc2327fbbdf1af442a99e9a04f650ccb415bc23d4379e63a5cb5d53f7c7861d5937db32700b001ab2024f1c078b0ba90f47af46c8c4dea33f972d7c".hexToBytes(),
+          // 87369d132ed6a7626dc4c5dfbaf41e954dd0ec55830e613a3f868c74d64a7a22 private
+          "0412eea943c13ed3e99d2de048a57dbaf4c977df83a50bf7ec79ae58255cf8d891a1cbe3cd51aeafec34241a1a9c9a54e861aa5b373667793c617c0310eab26653".hexToBytes()
```
zsluedem: https://github.com/rchain/rchain/issues/3676 created an issue for configuring this

casper/src/main/resources/MakeMint.rho
```diff
@@ -178,12 +150,20 @@ in {
         } |
         return!(bundle+{*thisMint})
       }
-    }
+    } |
+    // No internal state to return.
+    // Method required for updates.
+    contract @(bundle+{*MakeMint}, "extractState")(ret) = {
+      ret!([])
```
tgrospic: Should we return Nil for empty export state?

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -63,12 +70,10 @@ in {
...
       }
     } |
-
-    // MultiSig
+    // MultiSig vault creation method
     // Returns (false, errorMsg) or (true, (multiSigVault, revAddress, revVault))
     contract MultiSigRevVault(@"create", @publicKeys, @unsealers, @quorumSize, ret) = {
-      if (quorumSize > publicKeys.length() + unsealers.length() or
-          quorumSize <= 0) {
+      if (quorumSize > publicKeys.length() + unsealers.length() or quorumSize <= 0) {
```
tgrospic: With lazy boolean operators we can rewrite this to first check if quorum is valid and then compare it with pub keys.

casper/src/main/resources/PoS.rhox
```diff
@@ -70,22 +70,32 @@ User transfer the bonding amount directly to posVault. `allBonds` would update w
...
+  pickActiveValidators,
+  runMVar,
+  fst,
+  deposit,
+  distributeRewards,
+  perActiveValidatorPosVaultsCh,
+  transferFromBlockSender,
+  rs(`rho:registry:insertSigned:secp256k1`),
+  posId(`rho:rchain:deployerId`),
+  PendingRewards, MergeIntMaps,
```
tgrospic: Can we remove these if not used?

node/src/main/scala/coop/rchain/node/revvaultexport/RhoTrieTraverser.scala
```diff
@@ -73,13 +74,11 @@ object RhoTrieTraverser {
...
-  private val storeTokenUnforgeable = {
+  private val storeTokenUnforgeable: Par = {
     val rand =
       Tools.unforgeableNameRng(StandardDeploys.registryPubKey, StandardDeploys.registryTimestamp)
-    (0 to 5).foreach(_ => rand.next())
-    val newRand = rand.splitShort(6)
-    (0 to 6).foreach(_ => newRand.next())
-    val target = newRand.next()
+    (0 to 12).foreach(_ => rand.next())
+    val target = rand.next()
```
tgrospic: This can be simplified with the iterator (LazyList).
```scala
val target = LazyList.continually(rand.next()).drop(13).head
```

node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
tgrospic: We added this into _mainnet1_ folder to be able to support different versions. Can we add this change in _mainnet2_ and have support for both (if it's not too much extra work)?

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -251,7 +251,7 @@ object Transaction {
       StandardDeploys.revVaultTimestamp
     )
     // the 11th unforgeable name
-    val unfogeableBytes = Iterator.continually(seedForRevVault.next()).drop(10).next()
+    val unfogeableBytes = Iterator.continually(seedForRevVault.next()).drop(11).next()
```
tgrospic: Comment is now out of sync. Maybe better to not mention specific number in comment at all.

casper/src/main/resources/Registry.rho
```diff
@@ -332,115 +341,388 @@ in {
...
+        let @pubKey <- 
+          // fc743bd08a822d544bfbe05a5663fc325039a44c8f0c7fbea95a85517da5c36b private
+          "04e2eb6b06058d10b30856043c29076e2d2d7c374d2beedded6ecb8d1df585dfa583bd7949085ac6b0761497b0cfd056eb3d0db97efb3940b14c00fff4e53c85bf".hexToBytes(),
+        in {
+          new
+            multisig,
+            multisigLoc,       
+            multisigDispatcher
+            // Although the names above have `multisig`, they are not `multisig` currently because 
+            // we want to simplify the process and make it multisig in the future
```
tgrospic: This _multisig_ here is very confusing. Firstly, it's the same name as multi-sig REV vault contract and secondly it's not working right now as multi signature process.

Please rename it to to something like `systemContractManager`.

casper/src/main/resources/NonNegativeNumber.rho
```diff
@@ -59,10 +67,17 @@ new  NonNegativeNumber, MergeableTag, rs(`rho:registry:insertSigned:secp256k1`),
...
+  contract @(bundle+{*NonNegativeNumber}, "extractState")(ret) = {
+    ret!({"mergeableTag": *MergeableTag})
+  } |
+  // Insert the contract location in the registry location map
+  rb!(`rho:lang:nonNegativeNumber`, bundle+{*NonNegativeNumber}, *NonNegativeNumberDispatcher) |
+  // Inserts signed write-only dispatcher contract into the registry
   rs!(
     "04e1559d809924e564dce57e34646e155b144d2a504ce7ee519d7a5108fd42f1038d08d745e5ea21cb53d6aa7c7174a768fa373207a83bc947a20c6a02ece7a60e".hexToBytes(),
-    (9223372036854775807, bundle+{*NonNegativeNumber}),
+    (9223372036854775807, bundle+{*NonNegativeNumberDispatcher}),
```
tgrospic: I don't think we need updateable NNN contract. It has only number as internal state which is exposed so these is no need to transfer state.
Users which depends on this contract should create a new one if needed.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -431,7 +431,8 @@ object RhoRuntime {
     "rho:registry:insertSigned:secp256k1" -> Bundle(
       FixedChannels.REG_INSERT_SIGNED,
       writeFlag = true
-    )
+    ),
+    "rho:registry:insertBlessed" -> Bundle(FixedChannels.REG_INSERT_BLESSED, writeFlag = true)
```
tgrospic: Do we have test to check that anyone cannot call this process and insert arbitrary code?

zsluedem: No tests for that. 
it is guarded by pattern match. https://github.com/rchain/rchain/blob/e166d539376041758a17add4d18519d683468fbb/casper/src/main/resources/Registry.rho#L584-L601

casper/src/test/resources/updateAuthKey/UpdateAuthKey.rho
```diff
@@ -42,7 +42,7 @@ in {
     }
   } |
   stdout!(("propose start!!!!!"))|
-  rl!(`rho:registry:multisig`, *rlret)|
+  rl!(`rho:registry:systemContractManager`, *rlret)|
   for(@multiSigReg <- rlret){
```
tgrospic: Please update binder name also.

casper/src/test/scala/coop/rchain/casper/genesis/POSUpdate.scala
```diff
@@ -0,0 +1,175 @@
...
+                                    #  }
+                                    #}""".stripMargin('#')
+    TestNode.standaloneEff(genesis).use { node =>
+      val rm = node.runtimeManager
+      for {
+        b2  <- node.addBlock(updateDeploy)
+        _   = assert(b2.body.deploys.head.cost.cost > 0L, s"$b2 deploy cost is 0L")
+        _   = assert(b2.body.deploys.head.systemDeployError.isEmpty, s"$b2 system deploy failed")
+        _   = assert(!b2.body.deploys.head.isFailed, s"$b2 deploy failed")
+        ret <- rm.playExploratoryDeploy(exploreUpdateResultTerm, b2.body.state.postStateHash)
```
tgrospic: Please fix in other tests how data is returned from update contracts like in my example https://github.com/rchain/rchain/pull/3485/files#diff-a814de5ae24c79deeffeb8dbf45d34bf526f71651791d0463e81415c264d1571R127-R131.

casper/src/test/resources/updatePOS/ProposePOS.rho
```diff
@@ -0,0 +1,553 @@
+
```
tgrospic: Please create minimal example to update PoS. E.g. add one method so it's easy to see how this should work.

casper/src/test/resources/updateRegistry/AgreeAuthKey.rho
```diff
@@ -0,0 +1,31 @@
+
```
tgrospic: Please remove files that are related to multi signature confirmation of update. We can keep these files in a new draft PR for future update.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
tgrospic: I'd like to eliminate the need for `rho:registry:insertBlessed`. Can you please check if this is possible.

casper/src/test/resources/updateRegistry/AgreeAuthKey.rho
zsluedem: This is only for updating systemContractManager. It is safe to keep.

tgrospic: We have three files Agree... Propose... and Update... Are all three used?

casper/src/test/scala/coop/rchain/casper/genesis/POSUpdate.scala
```diff
@@ -0,0 +1,175 @@
...
+import coop.rchain.shared.scalatestcontrib._
+import coop.rchain.models.syntax._
+import coop.rchain.rholang.build.CompiledRholangTemplate
+import coop.rchain.rholang.interpreter.util.RevAddress
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Inspectors, Matchers}
+
+import scala.io.Source
+
+class POSUpdate extends FlatSpec with Matchers with Inspectors {
```
tgrospic: Naming should always be camel-case. Please rename POS to Pos on newly added entities.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
zsluedem: Do you have any backup plans for replacing this? I tried and could not get a good solution for that.

casper/src/test/scala/coop/rchain/casper/genesis/PosUpdate.scala
```diff
@@ -0,0 +1,177 @@
...
+import coop.rchain.shared.scalatestcontrib._
+import coop.rchain.models.syntax._
+import coop.rchain.models.rholang.implicits._
+import coop.rchain.rholang.build.CompiledRholangTemplate
+import coop.rchain.rholang.interpreter.RhoType.{Boolean, String, Tuple2}
+import coop.rchain.rholang.interpreter.util.RevAddress
+import monix.execution.Scheduler.Implicits.global
+import org.scalatest.{FlatSpec, Inspectors, Matchers}
+
+class PosUpdate extends FlatSpec with Matchers with Inspectors {
```
tgrospic: Please rename this test to PosUpdateSpec like we have naming for other tests.

The same for other tests in this folder.

casper/src/test/resources/updateRegistry/updateRegistry.rho
```diff
@@ -0,0 +1,71 @@
+new
```
tgrospic: Please start with uppercase with the name of the file,

casper/src/test/resources/MultiSigVault/posMultiSigConfirm.rho
```diff
@@ -1,11 +1,11 @@
-new RevVaultCh, MultiSigRevVaultCh, POSCh, multiSigCh, 
+new RevVaultCh, MultiSigRevVaultCh, PosCh, multiSigCh,
```
tgrospic, line 1: Files in this folder should also start with uppercase letter.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala
```diff
@@ -36,9 +36,9 @@ final case class ProofOfStake(
 }
 
 object ProofOfStake {
-  // TODO: Determine how the "initial bonds" map can simulate transferring stake into the PoS contract
+  // TODO: Determine how the "initial bonds" map can simulate transferring stake into the Pos contract
```
tgrospic: In comment it makes sense to write PoS as shorthand for Proof of Stake.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -440,15 +440,15 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
...
+      descr = "Minimum bond accepted by the Pos contract in the genesis block."
     )
 
     val genesisBlockNumber = opt[Long](
       descr = "Configure genesis blockNumber for hard fork."
     )
 
     val bondMaximum = opt[Long](
-      descr = "Maximum bond accepted by the PoS contract in the genesis block."
+      descr = "Maximum bond accepted by the Pos contract in the genesis block."
```
tgrospic: Here also makes sense to use PoS.

casper/src/main/resources/Registry.rho
```diff
@@ -332,115 +340,376 @@ in {
...
+                              // write new method entry point to location channel
+                              @loc!(meth) |
+                              ret!((true, *newRet))
+                            }
+                          }
+                        }
+                      } 
+                    }
+                  } else {
+                    ret!((false, "You have no permission to do that."))
```
tgrospic: This message is not descriptive. It should say exactly what operations is not permitted.

casper/src/main/resources/AuthKey.rho
```diff
@@ -66,13 +68,18 @@ in {
...
     }
   } |
-
+  // No internal state to return.
+  // Method required for updates.
+  contract @(bundle+{*AuthKey}, "extractState")(ret) = {
+    ret!({"authKeyUnf": *_authKey})
+  } |
+  // Insert location channel into the registry location map
+  rb!(`rho:rchain:authKey`, bundle+{*AuthKey}, *AuthKeyDispatcher) |
```
tgrospic: Why is necessary to send a bundle to `rho:registry:insertBlessed`?

casper/src/main/resources/Registry.rho
```diff
@@ -332,115 +340,376 @@ in {
...
+          } |
+          for (@`rho:rchain:multiSigRevVault`, @lib, @dispatcher <- insertBlessed) {
+            newInsert!(lib, `rho:rchain:multiSigRevVault`, dispatcher)
+          } |
+          for (@`rho:rchain:pos`, @lib, @dispatcher <- insertBlessed) {
+            newInsert!(lib, `rho:rchain:pos`, dispatcher)
+          } |
+          for (@`rho:rchain:revVault`, @lib, @dispatcher <- insertBlessed) {
+            newInsert!(lib, `rho:rchain:revVault`, dispatcher)
+          }
```
tgrospic: This is just extra complication and duplication of the shorthand names. We need to keep shorthands for contracts on multiple locations in the code and also keep track which are updatable.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
tgrospic: OK, it seems this can be simplified.

We know which public keys are used to create blessed contracts so we know URI's that will be created. We can match on them in insert signed method and create location map without this additional special name `rho:registry:insertBlessed`. The main point is that shorthand map must be defined only on one place and it refers to dispatcher which means it doesn't need to know real contract reference.

This should also eliminate the need for bundle on _extractState_ which is now supplied as a bundle to `insertBlessed`.

---

Another question is also how this pattern of updatable contract can be used by anyone and not just for blessed contracts. What I mean is that each contract can specify its own way to secure the access for update and not one for all.

casper/src/main/resources/AuthKey.rho
zsluedem: Good point

zsluedem:
> Another question is also how this pattern of updatable contract can be used by anyone and not just for blessed contracts. What I mean is that each contract can specify its own way to secure the access for update and not one for all.

Original `InsertSigned` ?

casper/src/test/scala/coop/rchain/casper/genesis/PosUpdateSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+  private val pub3               = Secp256k1.toPublic(p3)
+  private val validatorKeys      = Seq((p1, pub1), (noPermissionKey, noPermissionKeyPub), (p3, pub3))
+  private val genesisVaults =
+    Seq((p1, 100000000000L), (noPermissionKey, 100000000000L), (p3, 100000000000L))
+  private val bonds       = Map(pub1 -> 1000000L, noPermissionKeyPub -> 1000000L, pub3 -> 1000000L)
+  private val genesis     = buildGenesis(buildGenesisParameters(validatorKeys, genesisVaults, bonds))
+  private val epochLength = 1
+
+  private val updatePosTerm = CompiledRholangTemplate.loadTemplate(
+    "UpdatePos/ProposePos.rho",
```
tgrospic: Please create simplest possible example of updating PoS contract like updating minimum bonds. Update code should not repeat the whole PoS contract from scratch. Is this possible?

---

Please also rename ProposePos.rho to UpdatePos.rho to be consistent with other examples.

node/src/main/scala/coop/rchain/node/revvaultexport/RhoTrieTraverser.scala
```diff
@@ -73,13 +75,11 @@ object RhoTrieTraverser {
...
-  private val storeTokenUnforgeable = {
+  private val storeTokenUnforgeable: Par = {
     val rand =
       Tools.unforgeableNameRng(StandardDeploys.registryPubKey, StandardDeploys.registryTimestamp)
-    (0 to 5).foreach(_ => rand.next())
-    val newRand = rand.splitShort(6)
-    (0 to 6).foreach(_ => newRand.next())
-    val target = newRand.next()
+    val target = LazyList.continually(rand.next()).drop(11).head
+    println(s"ASDasdas ${target.toHexString}")
```
tgrospic: Is this leftover from debugging?

casper/src/test/scala/coop/rchain/casper/genesis/PosUpdateSpec.scala
zsluedem: > Please create simplest possible example of updating PoS contract like updating minimum bonds. 

That's how the codes work now.

> Update code should not repeat the whole PoS contract from scratch. Is this possible?

~~No.~~ I see now.

casper/src/main/resources/AuthKey.rho
```diff
@@ -69,7 +69,7 @@ in {
   } |
   // No internal state to return.
   // Method required for updates.
-  contract @(*AuthKey, "extractState")(ret) = {
+  for (ret <- @(*AuthKey, "extractState")) {
```
nzpr, line 75: Could you explain the reason and effect of this change? A I understand,  persistent receive is replaced with linear one, so I was expecting to have in this commit next to this change some produce somewhere that puts the extracted state back.

casper/src/main/resources/AuthKey.rho
```diff
@@ -74,11 +72,9 @@ in {
   contract @(*AuthKey, "extractState")(ret) = {
     ret!({"authKeyUnf": *_authKey})
   } |
-  // Insert location channel into the registry location map
-  rb!(`rho:rchain:authKey`, *AuthKey, *AuthKeyDispatcher) |
   // Inserts signed write-only AuthKey contract into the registry
   rs!(
-    (9223372036854775807, bundle+{*AuthKeyDispatcher}),
+    (9223372036854775807, *AuthKey),
```
nzpr: Why is it not bundled anymore?

zsluedem: https://github.com/rchain/rchain/pull/3485#discussion_r866737455

zsluedem: requested from @tgrospic

casper/src/main/resources/AuthKey.rho
zsluedem: Because right now `AuthKey` could be reused in update and we could bound new state with only one consume.

casper/src/main/resources/AuthKey.rho
zsluedem: The `AuthKey` is now hidden behind `Dispatcher` which users no longer have access.

casper/src/main/resources/Registry.rho
```diff
@@ -1,15 +1,26 @@
...
+  secpVerify(`rho:crypto:secp256k1Verify`),
+  blake2b256(`rho:crypto:blake2b256Hash`),
+  TreeHashMap,
+  TreeHashMapLoc,
+  TreeHashMapDispatcher,
+  secpVerify(`rho:crypto:secp256k1Verify`),
+  blake2b256(`rho:crypto:blake2b256Hash`),
+  hashVerify,
+  storeToken,
+  normalInsertSigned
```
nzpr: what does `normalInsertSigned` mean? I mean why it is `normal`?

zsluedem: How would you like to call that?

nzpr: Why not just `insertSigned?`

casper/src/main/resources/AuthKey.rho
tgrospic: This now implies some implicit management for updatable contract. I had something like this in mind where insert to the registry is the same for all. Updatable contract should be created explicitly.
```scala
proxify!(*RevVault, *updateSecurityLogicOrSomething, *RevVaultProxyCh) |
for(RevVaultProxy <- RevVaultProxyCh) {
  rs!(
    (9223372036854775807, bundle+{*RevVaultProxy}),
    *deployerId,
    *uriOut)
}
```

casper/src/main/resources/Registry.rho
```diff
@@ -330,104 +340,382 @@ in {
...
+    `rho:rchain:multiSigRevVault` : `rho:id:b9s6j3xeobgset4ndn64hje64grfcj7a43eekb3fh43yso5ujiecfn`
+    }, {
+    `rho:id:qrh6mgfp5z6orgchgszyxnuonanz7hw3amgrprqtciia6astt66ypn` : `rho:lang:either`,
+    `rho:id:6fzorimqngeedepkrizgiqms6zjt76zjeciktt1eifequy4osz35ks` : `rho:lang:listOps`,
+    `rho:id:hxyadh1ffypra47ry9mk6b8r1i33ar1w9wjsez4khfe9huzrfcytx9` : `rho:lang:nonNegativeNumber` ,
+    `rho:id:1qw5ehmq1x49dey4eadr1h4ncm361w3536asho7dr38iyookwcsp6i` : `rho:rchain:authKey`,
+    `rho:id:asysrwfgzf8bf7sxkiowp4b3tcsy4f8ombi3w96ysox4u3qdmn1wbc` : `rho:rchain:makeMint`,
+    `rho:id:m3xk7h8r54dtqtwsrnxqzhe81baswey66nzw6m533nyd45ptyoybqr` : `rho:rchain:pos`,
+    `rho:id:6zcfqnwnaqcwpeyuysx1rm48ndr6sgsbbgjuwf45i5nor3io7dr76j` : `rho:rchain:revVault`,
+    `rho:id:b9s6j3xeobgset4ndn64hje64grfcj7a43eekb3fh43yso5ujiecfn` : `rho:rchain:multiSigRevVault`
```
tgrospic: URIs literals should be only defined once, if used on multiple places it should be defined with aliases.
```scala
let uriEither <- `rho:id:qrh6mg...t66ypn`; uriListOps <- `rho:id:6fzorimqn...z35ks`; ... in { ... }
```

casper/src/main/resources/Registry.rho
```diff
@@ -330,104 +358,241 @@ in {
...
+                  @(*oldLib, "securityCheck")!(securityData, *securityCh)|
+                  for (@res <-securityCh){
+                    match res {
+                      (true, Nil) => {
+                        for (oldLib <- loc) {
+                          new tmp, newRet in {
+                            @(*oldLib, "extractState")!(*tmp) |
+                            for (@oldState <- tmp) {
+                              // launch new contract instance with initial state extracted from the old instance
+                              @con!(oldState, *oldLib, *newRet) |
```
tgrospic: So here owner of the original (oldLib) contract will receive the instance. It means it can call `@(*oldLib, "extractState")` if needed so there is no need to do it as part of _systemContractManager_ code.
~~We can do the same for security check.~~

casper/src/main/resources/Registry.rho
```diff
@@ -330,104 +358,241 @@ in {
...
+      } |
+    {
+      // ------
+      // Update
+      // deployerID: (`rho:rchain:deployerId`)
+      // uri:   the uri you want to update.  `rho:rchain:pos` etc.
+      // con:   the update consume which take the old state
+      // meth:  the target new contract which would replace the old one
+      // ret:   channel which returns the update result
+      contract systemContractManager(@"update", @securityData, @uri, @con, ret) = {
```
tgrospic: Please use more descriptive names instead of _con_ and _meth_ below (it's not part of contract params).

casper/src/main/resources/AuthKey.rho
```diff
@@ -67,12 +70,39 @@ in {
...
+        for (@deployerPubKeyBytes <- deployerPubKeyBytesCh){
+          if (deployerPubKeyBytes == pubKey){
+            ret!((true, Nil))
+          } else {
+            ret!((false, "You have no permissions to update system contract!"))
+          }
+        }
+      }
+    }
+  } |
```
tgrospic: This security check code is duplicated in all updatable contracts. It can be extracted and loaded from the registry. Maybe we don't need alias for it because it's just initial security check based on supplied config public key.

casper/src/main/resources/Registry.rho
```diff
@@ -330,104 +358,241 @@ in {
...
+    `rho:lang:either`             : `rho:id:qrh6mgfp5z6orgchgszyxnuonanz7hw3amgrprqtciia6astt66ypn`,
+    `rho:lang:listOps`            : `rho:id:6fzorimqngeedepkrizgiqms6zjt76zjeciktt1eifequy4osz35ks`,
+    `rho:lang:nonNegativeNumber`  : `rho:id:hxyadh1ffypra47ry9mk6b8r1i33ar1w9wjsez4khfe9huzrfcytx9`,
+    `rho:rchain:authKey`          : `rho:id:1qw5ehmq1x49dey4eadr1h4ncm361w3536asho7dr38iyookwcsp6i`,
+    `rho:rchain:makeMint`         : `rho:id:asysrwfgzf8bf7sxkiowp4b3tcsy4f8ombi3w96ysox4u3qdmn1wbc`,
+    `rho:rchain:pos`              : `rho:id:m3xk7h8r54dtqtwsrnxqzhe81baswey66nzw6m533nyd45ptyoybqr`,
+    `rho:rchain:revVault`         : `rho:id:6zcfqnwnaqcwpeyuysx1rm48ndr6sgsbbgjuwf45i5nor3io7dr76j`,
+    `rho:rchain:multiSigRevVault` : `rho:id:b9s6j3xeobgset4ndn64hje64grfcj7a43eekb3fh43yso5ujiecfn`
+    } in {
+      contract hashVerify(@data, @sig, @pubKey, ret) = {
```
tgrospic: This is not used anymore.

casper/src/main/resources/AuthKey.rho
```diff
@@ -67,12 +70,39 @@ in {
...
+  rl!(`rho:registry:systemContractManager`, *systemContractManagerCh)|
+  for (@(_, systemContractManager) <- systemContractManagerCh){
+    @systemContractManager!("createDispatcher", *AuthKey, *dispatcherCh)|
+    for (dispatcher <- dispatcherCh){
+      // Inserts signed write-only AuthKey dispatcher contract into the registry
+      rs!(
+        (9223372036854775807, *dispatcher),
```
tgrospic: Dispatcher should be bundled when inserting to the registry.

zsluedem: https://github.com/rchain/rchain/blob/1b8df5514347d2a658e726be7986ef05a0b17857/casper/src/main/resources/Registry.rho#L580 bundle here.

casper/src/main/resources/Registry.rho
zsluedem: > We can do the same for security check.

No. We can not. People are allowed to get `oldLib` only when security check passed. How to guarantee security in the way of your description? Could you elaborate more?

tgrospic: User is providing `@con` consume which will receive old lib. What I pointing here is that user can call `@(*oldLib, "extractState")` if needed so it doesn't have to be supplied together with old lib. This is of course after security check is done as you already done.

zsluedem: > We can do the same for security check.

I meant this part. How could we do that the same for security check?

tgrospic: You are right, we cannot do this for security check. Before passing security check old lib cannot be accessed by the caller.

casper/src/main/resources/Registry.rho
```diff
@@ -51,10 +61,27 @@ in {
   //     }
   //   }
   // }
-
+  contract createSecurityCheck(lib) = {
+    let @pubKey <-
+      // fc743bd08a822d544bfbe05a5663fc325039a44c8f0c7fbea95a85517da5c36b private
+      "04e2eb6b06058d10b30856043c29076e2d2d7c374d2beedded6ecb8d1df585dfa583bd7949085ac6b0761497b0cfd056eb3d0db97efb3940b14c00fff4e53c85bf".hexToBytes()
+    in {
+      contract @(*lib, "securityCheck")(@deployerID, ret) = {
```
tgrospic: Hm, we have similar situation with this contract as with TreeHashMap, it's used in Registry so it cannot be defined in its own file. Please use only this definition to register it and remove separate file.

---

What is the purpose of `lib` here? This contract should just be to check (compare) public key defined in config with supplied _deployerId_. So this wrapper contract is not needed.

---

Name should also not include Coop because it can be used by anyone running their network.
Maybe we can call it `configPublicKeyCheck`?

casper/src/main/resources/CoopSecurityCheck.rho
```diff
@@ -0,0 +1,31 @@
+new
```
tgrospic: Sorry, I forgot that this contract is used inside Registry also so it cannot be defined in a separate file, just like TreeHashMap.

casper/src/main/resources/Registry.rho
zsluedem: > What is the purpose of lib here? This contract should just be to check (compare) public key defined in config with supplied deployerId. So this wrapper contract is not needed.

How should I check then with some many libs `MakeMint`, `RevVault`?

tgrospic: If `configPublicKeyCheck` is defined like this.
```scala
contract configPublicKeyCheck(deployerID, ret) = {
  let @pubKey <- 
    // fc743bd08a822d544bfbe05a5663fc325039a44c8f0c7fbea95a85517da5c36b private
    "04e2eb6b06058d10b30856043c29076e2d2d7c374d2beedded6ecb8d1df585dfa583bd7949085ac6b0761497b0cfd056eb3d0db97efb3940b14c00fff4e53c85bf".hexToBytes()
  in {
    new DeployerIdOps(`rho:rchain:deployerId:ops`), deployerPubKeyBytesCh in {
      DeployerIdOps!("pubKeyBytes", deployerID, *deployerPubKeyBytesCh) |
      for (@deployerPubKeyBytes <- deployerPubKeyBytesCh){
        if (deployerPubKeyBytes == pubKey){
          ret!((true, Nil))
        } else {
          ret!((false, "You have no permissions to update system contract!"))
        }
      }
    }
  } 
}
```
Then each system contract should define its own securityCheck method.
```scala
contract @(*<lib>, "securityCheck")(@deployerID, @return) = {
  configPublicKeyCheck(deployerID, return)
}
```

casper/src/main/resources/AuthKey.rho
```diff
@@ -67,12 +71,31 @@ in {
...
+  }  |
+
+  rl!(`rho:registry:systemContractManager`, *systemContractManagerCh)|
+  rl!(`rho:rchain:pubKeyCheck`, *pubKeyCheckCh)|
+  for (@(_, systemContractManager) <- systemContractManagerCh & @(_, pubKeyCheck)<- pubKeyCheckCh){
+    @systemContractManager!("createDispatcher", *AuthKey, *dispatcherCh)|
+    contract @(*AuthKey, "securityCheck")(@deployerId, ret) = {
+      @pubKeyCheck!(
+        deployerId,
+        "04e2eb6b06058d10b30856043c29076e2d2d7c374d2beedded6ecb8d1df585dfa583bd7949085ac6b0761497b0cfd056eb3d0db97efb3940b14c00fff4e53c85bf",
```
tgrospic: This is going to be problem when we want to put this hard-coded public key in the configuration.

So instead to duplicate config public key in all files it makes sense to have it only in the Registry where we can substitute only once, similar how we have PoS contract (it not great solution but we don't have better right now).

casper/src/main/resources/AuthKey.rho
```diff
@@ -67,12 +71,27 @@ in {
...
     }
   } |
+  // No internal state to return.
+  // Method required for updates.
+  for (ret <- @(*AuthKey, "extractState")) {
+    ret!({"authKeyUnf": *_authKey})
+  }  |
+
+  rl!(`rho:registry:systemContractManager`, *systemContractManagerCh)|
+  rl!(`rho:rchain:checkSystemSecurity`, *checkSystemSecurityCh)|
```
tgrospic: `checkSystemSecurity` doesn't reflect what this contract is doing better to call it something like `configPublicKeyCheck` as I mentioned in https://github.com/rchain/rchain/pull/3485#discussion_r879071859.

casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala
```diff
@@ -42,8 +43,10 @@ object StandardDeploys {
   val revGeneratorPk      = "a06959868e39bb3a8502846686a23119716ecd001700baf9e2ecfa0dbf1a3247"
 
   val (registryPubKey, registryTimestamp) = (toPublic(registryPk), 1559156071321L)
-  val (listOpsPubKey, listOpsTimestamp)   = (toPublic(listOpsPk), 1559156082324L)
-  val (eitherPubKey, eitherTimestamp)     = (toPublic(eitherPk), 1559156217509L)
+  val (coopSecurityCheckPubKey, coopSecurityCheckTimestamp) =
+    (toPublic(coopSecurityCheckPk), 1559156071322L)
```
tgrospic: This is now not needed because config public key check contract is part of the registry.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/CheckSystemSecuritySpec.scala
```diff
@@ -0,0 +1,12 @@
...
+import coop.rchain.casper.helper.RhoSpec
+import coop.rchain.models.NormalizerEnv
+import coop.rchain.rholang.build.CompiledRholangSource
+
+class CheckSystemSecuritySpec
+    extends RhoSpec(
+      CompiledRholangSource("CheckSystemSecurityTest.rho", NormalizerEnv.Empty),
+      Seq.empty,
+      GENESIS_TEST_TIMEOUT
+    )
```
tgrospic: This also.

rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala
```diff
@@ -96,6 +96,7 @@ object SystemProcesses {
     val DEPLOYER_ID_OPS: Long    = 14L
     val REG_OPS: Long            = 15L
     val SYS_AUTHTOKEN_OPS: Long  = 16L
+    val REG_INSERT_BLESSED: Long = 17L
```
tgrospic: This is also not used anymore.

casper/src/test/scala/coop/rchain/casper/genesis/contracts/CheckSystemSecuritySpec.scala
zsluedem: You want to remove the test completely?

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -260,10 +268,22 @@ in {
         } // new
       } // quorum size
     } |
-    rs!(
-      (9223372036854775807, bundle+{*MultiSigRevVault}),
-      *deployerId,
-      *uriOut
-    )
+    // No internal state to return.
```
tgrospic: What this comment means?

casper/src/main/resources/MakeMint.rho
```diff
@@ -179,11 +154,23 @@ in {
         } |
         return!(bundle+{*thisMint})
       }
+    } |
+    // No internal state to return.
```
tgrospic: This comment would make more sense inside `extractState` method.

casper/src/main/resources/AuthKey.rho
```diff
@@ -67,12 +71,27 @@ in {
       }
     }
   } |
+  // No internal state to return.
```
tgrospic: This comment doesn't makes sense here.

tgrospic (approved):
Excellent work @Isaac-DeFrain and @zsluedem! :1st_place_medal:

Thanks for addressing all of my comments!

zsluedem (approved):
Just want to approve to merge

zsluedem:
bors merge

PR #3492, opened 2021-07-27
Merging number channels (REV vaults) / fix conflicts in PoS

### LINKED ISSUE #3513: REV vaults (number channel) merging
## Overview

## Solution

## Design

    #3492
    #3492
    #3492
    #3492
- [x] #3514
  #3559
  #3577
  #3591

tgrospic:
## Overview

This PR should resolve this by recognizing special type of channel that can contain values which are mergeable, so even if multiple actors changing the channel it can still be merged as part of block-merge.
As a first implementation this will be done for channels holding REV balances which means that PoS don't need per validator vaults and can directly work with PoS vault.

### Issues not finished in this PR

1. Create unforgeable name to tag number channels, now fixed `__MERGEABLE__` name is used
1. Implement rejections for merged number channels which are negative or overflow
1. Add additional tests for number channel merging logic 


casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeployUserError.scala
```diff
@@ -1,21 +1,16 @@
...
 
-sealed trait SystemDeployFailure
-
-sealed abstract class SystemDeployUserError(val errorMessage: String) extends SystemDeployFailure
-
-object SystemDeployUserError {
-  final case class SystemDeployError(errorMsg: String) extends SystemDeployUserError(errorMsg)
-  def apply(errorMsg: String) = SystemDeployError(errorMsg)
-}
+final case class SystemDeployUserError(errorMessage: String)
```
tgrospic, line 7: `SystemDeployUserError` is now exclusively error inside Rholang code.

casper/src/main/scala/coop/rchain/casper/util/rholang/SystemDeploy.scala
```diff
@@ -48,12 +50,8 @@ abstract class SystemDeploy(initialRand: Blake2b512Random) {
...
   val source: String
 
-  final def extractResult(output: Par): Either[SystemDeployFailure, Result] =
-    extractor
-      .unapply(output)
-      .fold[Either[SystemDeployFailure, Result]](Left(UnexpectedResult(List(output))))(
-        processResult
-      )
+  final def extractResult[F[_]: Sync](output: Par): F[Either[SystemDeployUserError, Result]] =
+    extractor.unapply(output).map(processResult).liftTo(UnexpectedResult(Seq(output)))
```
tgrospic, line 54: Fatal errors are now thrown with F context (applicative errors) so SystemDeployUserError is only for Rholang errors.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -36,6 +36,21 @@ import org.scalatest.{FlatSpec, Matchers}
...
+
+final case class ReplayFailed[A](systemDeployError: SystemDeployUserError)
+    extends SystemDeployReplayResult[A]
+
+object SystemDeployReplayResult {
+  def replaySucceeded[A](stateHash: StateHash, result: A): SystemDeployReplayResult[A] =
+    ReplaySucceeded(stateHash, result)
+  def replayFailed[A](systemDeployError: SystemDeployUserError): SystemDeployReplayResult[A] =
+    ReplayFailed(systemDeployError)
+}
```
tgrospic, line 52: This whole `SystemDeployReplayResult` hierarchy objects was only used in tests so I moved it here.

casper/src/test/scala/coop/rchain/casper/util/rholang/RuntimeManagerTest.scala
```diff
@@ -161,6 +177,38 @@ class RuntimeManagerTest extends FlatSpec with Matchers {
...
+                           SystemDeployReplayResult.replaySucceeded(
+                             checkpoint.root.toByteString,
+                             value
+                           )
+                         }
+                     case (Left(failure), _) =>
+                       SystemDeployReplayResult.replayFailed[S#Result](failure).pure[F]
+                   }
+    } yield replayed
+  }
```
tgrospic, line 210: This function was previously used in tests only and had some duplicated logic. Now it's basically just tiny wrapper on runtime replay syntax to create SystemDeployReplayResult.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -14,9 +15,12 @@ import coop.rchain.rholang.interpreter.errors.{
   InterpreterError,
   OutOfPhlogistonsError
 }
-import coop.rchain.metrics.implicits._
 
-final case class EvaluateResult(cost: Cost, errors: Vector[InterpreterError]) {
+final case class EvaluateResult(
+    cost: Cost,
+    errors: Vector[InterpreterError],
+    mergeable: Set[Par]
```
tgrospic, line 22: Reducer will collect mergeable channels as part of evaluation results.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Interpreter.scala
```diff
@@ -56,18 +61,19 @@ class InterpreterImpl[F[_]: Sync: Span](implicit C: _cost[F]) extends Interprete
                      case err: InterpreterError => ParserError(err).raiseError[F, Par]
                    }
                }
-      _         <- Span[F].traceI("reduce-term") { reducer.inj(parsed) }
-      phlosLeft <- C.get
-    } yield EvaluateResult(initialPhlo - phlosLeft, Vector())
+      // Empty mergeable channels
+      _ <- mergeChs.update(_.empty)
```
tgrospic, line 65: Before each run of Rholang code, mergeable channels are set to empty. It works similar as Cost works now.

rholang/src/main/scala/coop/rchain/rholang/interpreter/Reduce.scala
```diff
@@ -143,6 +148,20 @@ class DebruijnInterpreter[M[_]: Sync: Parallel: _cost](
...
+
+  private def updateMergeableChannels(chan: Par) = Sync[M].defer {
+    val isMergeable = isMergeableChannel(chan)
+
+    mergeChs.update(_ + chan).whenA(isMergeable)
+  }
+
+  private def isMergeableChannel(chan: Par) = {
+    val tupleElms     = chan.exprs.flatMap(y => y.getETupleBody.ps)
+    val mergeStr: Par = GString("__MERGEABLE__")
```
tgrospic, line 161: In this PR number channels are "tagged" with fixed channel but it must be some unforgeable name which will be added in separate PR.

rholang/src/main/scala/coop/rchain/rholang/interpreter/compiler/normalizer/processes/PNewNormalizer.scala
```diff
@@ -38,32 +35,22 @@ object PNewNormalizer {
...
+    normalizeMatch[F](p.proc_, ProcVisitInputs(VectorPar(), newEnv, input.knownFree)).map {
+      bodyResult =>
+        val resultNew = New(
+          bindCount = newCount,
+          p = bodyResult.par,
+          uri = uris,
+          injections = env,
+          locallyFree = bodyResult.par.locallyFree.from(newCount).map(x => x - newCount)
+        )
+        ProcVisitOutputs(input.par.prepend(resultNew), bodyResult.knownFree)
```
tgrospic, line 53: `deployId` and `deployerId` were not used in normalizer but only in reducer. So check here and throwing errors makes no sense.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -53,15 +55,26 @@ object RholangAndScalaDispatcher {
...
       new RholangAndScalaDispatcher(dispatchTable)
 
     implicit lazy val reducer: Reduce[M] =
-      new DebruijnInterpreter[M](tuplespace, dispatcher, urnMap)
+      new DebruijnInterpreter[M](tuplespace, dispatcher, urnMap, mergeChs)
 
     (dispatcher, reducer)
   }
+
+  def create2[M[_]: Sync: Parallel: _cost](
```
tgrospic: Sadly, overload method with the same name doesn't work here.

rholang/src/test/scala/coop/rchain/rholang/interpreter/merging/RholangMergingLogicSpec.scala
```diff
@@ -0,0 +1,48 @@
...
+      Map((chA, 15L), (chC, 10L))
+    )
+
+    def getDataOnHash[F[_]: Applicative](hash: String): F[Option[Long]] =
+      initValues.get(hash).pure[F]
+
+    RholangMergingLogic.calculateNumChannelDiff(input, getDataOnHash[Task]).map { res =>
+      res shouldBe Seq(Map(("A", 10)), Map(("B", 3)), Map(("A", -5), ("C", -10)))
+    }
+  }
```
tgrospic, line 45: More tests will be added in separate PR as mentioned in the PR description.

rspace/src/main/scala/coop/rchain/rspace/RSpaceOps.scala
```diff
@@ -327,8 +327,7 @@ abstract class RSpaceOps[F[_]: Concurrent: ContextShift: Log: Metrics: Span, C,
       _           <- createNewHotStore(nextHistory.getHistoryReader(root))
       _           <- restoreInstalls()
 
-      // TODO: temp fix to release Semaphores inside TwoStepLock
-      //  Adjust when runtime changes got in, create instance on spawn runtime.
+      // Clean channel locks
       _ <- lockF.cleanUp
```
tgrospic, line 331: Better to leave locks cleanup so that reset puts RSpace in initial state.

rspace/src/main/scala/coop/rchain/rspace/hashing/StableHashProvider.scala
```diff
@@ -15,10 +15,11 @@ object StableHashProvider {
...
 
-  def hashSeq[C](channels: Seq[C])(implicit serializeC: Serialize[C]): Seq[Blake2b256Hash] = {
-    val encodedSeq = channels.map(serializeC.encode).sorted(util.ordByteVector)
-    encodedSeq.map(Blake2b256Hash.create).sortBy(_.bytes)(util.ordByteVector)
-  }
+  def hashSeq[C](channels: Seq[C])(implicit serializeC: Serialize[C]): Seq[Blake2b256Hash] =
+    channels
+      .map(serializeC.encode)
+      .map(Blake2b256Hash.create)
+      .sortBy(_.bytes)(util.ordByteVector)
```
tgrospic, line 22: Removed unnecessary ordering of intermediate list of ByteVectors.

rspace/src/main/scala/coop/rchain/rspace/merger/EventLogIndex.scala
```diff
@@ -145,6 +165,11 @@ object EventLogIndex {
     Seq(x, y).map(_.producesTouchingBaseJoins).reduce(_ ++ _),
     Seq(x, y).map(_.consumesLinearAndPeeks).reduce(_ ++ _),
     Seq(x, y).map(_.consumesPersistent).reduce(_ ++ _),
-    Seq(x, y).map(_.consumesProduced).reduce(_ ++ _)
+    Seq(x, y).map(_.consumesProduced).reduce(_ ++ _),
+    // Combine mergeable produces
+    x.producesMergeable |+| y.producesMergeable,
+    x.consumesMergeable |+| y.consumesMergeable,
+    // Merge number channels (add differences)
+    x.numberChannelsData |+| y.numberChannelsData
```
tgrospic, line 173: Both number channels difference data will apply Monoid operation for values (Long) which means they will be added together.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -9,10 +9,26 @@ import scala.annotation.tailrec
...
+
+  type NumberChannelsDiff = Map[Blake2b256Hash, Long]
+
   /** If target depends on source. */
-  def depends(target: EventLogIndex, source: EventLogIndex): Boolean =
-    (producesCreatedAndNotDestroyed(source) intersect target.producesConsumed).nonEmpty ||
-      (consumesCreatedAndNotDestroyed(source) intersect target.consumesProduced).nonEmpty
+  def depends(target: EventLogIndex, source: EventLogIndex): Boolean = {
+    val producesSource = producesCreatedAndNotDestroyed(source) diff source.producesMergeable
+    val producesTarget = target.producesConsumed diff source.producesMergeable
```
tgrospic, line 22: Dependency is now skipping number channels.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -30,10 +46,13 @@ object MergingLogic {
...
-      val consumeRaces =
-        (a.consumesProduced intersect b.consumesProduced).toIterator.filterNot(_.persistent)
+      val consumeRaces = {
+        (a.consumesProduced intersect b.consumesProduced -- a.consumesMergeable -- b.consumesMergeable).toIterator
+          .filterNot(_.persistent)
+      }
       val produceRaces =
-        (a.producesConsumed intersect b.producesConsumed).toIterator.filterNot(_.persistent)
+        (a.producesConsumed intersect b.producesConsumed -- a.producesMergeable -- b.producesMergeable).toIterator
+          .filterNot(_.persistent)
```
tgrospic: The same for conflicts, number channels are skipped for detection.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala
```diff
@@ -148,11 +147,8 @@ object StateChange {
 
   def empty: StateChange = StateChange(Map.empty, Map.empty, Map.empty)
   def combine(x: StateChange, y: StateChange): StateChange = {
-    def sumMaps[A, B: Monoid](map1: Map[A, B], map2: Map[A, B]): Map[A, B] = map1 ++ map2.map {
-      case (k, v) => k -> (v combine map1.getOrElse(k, Monoid[B].empty))
-    }
-    val newDC = sumMaps(x.datumsChanges, y.datumsChanges)
-    val newKC = sumMaps(x.kontChanges, y.kontChanges)
+    val newDC = x.datumsChanges |+| y.datumsChanges
+    val newKC = x.kontChanges |+| y.kontChanges
```
tgrospic, line 151: `sumMaps` is basically Monoid implementation for maps in StateChange.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChangeMerger.scala
```diff
@@ -25,7 +26,14 @@ object StateChangeMerger {
...
       changes: StateChange,
-      baseReader: HistoryReaderBinary[F, C, P, A, K]
+      baseReader: HistoryReaderBinary[F, C, P, A, K],
+      mergeableChs: NumberChannelsDiff,
+      // Override channel change
+      handleChannelChange: (
+          Blake2b256Hash,
+          ChannelChange[ByteVector],
+          NumberChannelsDiff
+      ) => F[Option[HotStoreTrieAction]]
```
tgrospic, line 36: This additional handler can return a value if channel implements some special behavior like merging for number channels.

rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
```diff
@@ -37,6 +37,12 @@ object ScodecSerialize {
     encodeSortedSeq[Datum[A]](datums, codec)
   }
 
+  def encodeDatum[A](datum: Datum[A])(implicit sa: Serialize[A]): ByteVector = {
+    val codec = serializeToCodecDatumMemo(sa)
+
+    codec.encode(datum).getUnsafe.toByteVector
+  }
```
tgrospic, line 44: Used in `RholangMergingLogic` to encode single datum which is carried in `ChannelChange[ByteVector]` added/removed lists.

shared/src/main/scala/coop/rchain/catscontrib/monadOps.scala
```diff
@@ -1,11 +1,19 @@
...
-  def forever: F[A]                                         = MonadOps.forever(self)
+// TODO: Rename this syntax to include functor, monad, ... (HKTSyntax?)
+
+final class MonadOps[F[_], A](val self: F[A]) extends AnyVal {
+  def liftM[G[_[_], _]](implicit M: Monad[F], G: MonadTrans[G]): G[F, A] = G.liftM(self)
+  def forever(implicit M: Monad[F]): F[A]                                = MonadOps.forever(self)
+
+  /* EitherT syntax */
+
+  def liftEitherT[L](implicit f: Functor[F]): EitherT[F, L, A] = EitherT.liftF[F, L, A](self)
```
tgrospic, line 31: This function is useful for ~list~ lift `F[A]` without specifying all type parameters together with F, left and right types.
https://github.com/rchain/rchain/blob/9317f35b5e78550dc5573307d417d28fde6f279c/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeReplaySyntax.scala#L141

casper/src/main/resources/NonNegativeNumber.rho
```diff
@@ -33,39 +33,37 @@ new  NonNegativeNumber, rs(`rho:registry:insertSigned:secp256k1`), uriOut in {
...
       return!(bundle+{*this}) |
       match init { //Initial balance is zero if given is negative on non-integer
         Int => {
-          if (init >= 0) { valueStore!(init)  }
-          else           { valueStore!(0) }
+          if (init >= 0) { @("__MERGEABLE__", *valueStore)!(init)  }
+          else           { @("__MERGEABLE__", *valueStore)!(0) }
         }
-         _ => { valueStore!(0) }
+         _ => { @("__MERGEABLE__", *valueStore)!(0) }
```
tgrospic, line 58: In separate PR fixed channel will be replaced with unforgeable channel.

rspace/src/main/scala/coop/rchain/rspace/hashing/StableHashProvider.scala
zsluedem: nice catch

shared/src/main/scala/coop/rchain/catscontrib/ApplicativeError_.scala
```diff
@@ -2,17 +2,11 @@ package coop.rchain.catscontrib
...
-/** We should REALLY use Scalaz */
 trait ApplicativeError_[F[_], E] {
 
   implicit val instance: ApplicativeError[F, E]
 
-  def raiseError[A](e: E): F[A]                        = instance.raiseError(e)
-  def handleErrorWith[A](fa: F[A])(f: E => F[A]): F[A] = instance.handleErrorWith(fa)(f)
-  def fromEither[A](x: E Either A): F[A]               = instance.fromEither(x)
-  def attempt[A](fa: F[A]): F[Either[E, A]]            = instance.attempt(fa)
-
```
zsluedem, line 15: I don't understand what is the difference between `cats.ApplicativeError` from this one. Maybe we should remove this file completely?

tgrospic: I tried but not hard enough. :) I looked more carefully and found it's used in two places, `PacketHandler` and `MonadError_` which both can be fixed. https://github.com/rchain/rchain/pull/3492/commits/1d087b1890a91cd86ebe5522acc478c96c3e54cd

This one does not extend `Applicative` instead holds the instance as implicit field. The purpose is to prevent ambiguous implicit resolution when it's used in combination with other interfaces extending Applicative. 

`MonadError_` is used here (seems used only in tests) so it can implement `raiseError` directly.
https://github.com/rchain/rchain/blob/53b2cc19efc68aafd7e207d7b082b7af8d5c9fac/rholang/src/main/scala/coop/rchain/rholang/interpreter/matcher/package.scala#L29

And `PacketHandler` can also use `ApplicativeError` directly.
https://github.com/rchain/rchain/blob/1a90135086723203b8a590d6679abf0f91c08012/comm/src/main/scala/coop/rchain/p2p/effects/PacketHandler.scala#L19-L22

zsluedem: I was just curious. We can fix this later.

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
```diff
@@ -59,6 +59,8 @@ object RNodeKeyValueStoreManager {
       (Db("reporting-cache"), reportingEnvConfig),
       // CasperBuffer
       (Db("parents-map"), casperBufferEnvConfig),
+      // Runtime mergeable store (cache of mergeable channels for block-merge)
+      (Db("mergeable-channel-cache"), casperBufferEnvConfig),
```
nzpr: I guess this should not be in `casperBuffer` env but in `dagStorageEnvConfig`?

casper/src/main/scala/coop/rchain/casper/util/rholang/RuntimeManager.scala
```diff
@@ -63,10 +68,13 @@ trait RuntimeManager[F[_]] {
...
 }
 
+final case class DeployMergeableData(channels: Seq[Blake2b256Hash])
+
 final case class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log: ContextShift: Parallel](
     space: RhoISpace[F],
     replaySpace: RhoReplayISpace[F],
-    historyRepo: RhoHistoryRepository[F]
+    historyRepo: RhoHistoryRepository[F],
+    mergeableStore: KeyValueTypedStore[F, ByteVector, DeployMergeableData]
```
nzpr: Would it make sense to create a type for this KV store? As it is used in several places.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChange.scala
nzpr: Oh... thats nice thanks, I was searching for this but looks like were asking wrong questions

shared/src/main/scala/coop/rchain/catscontrib/monadOps.scala
nzpr: Wdy mean by `list` here? Looking into this code ` val refT = Ref.of(Set[Par]()).liftEitherT[ReplayFailure] ` I dont understand what it is doing, could you elaborate?

casper/src/main/resources/NonNegativeNumber.rho
nzpr: Just curious, what would be the benefit from having unf name?

rholang/src/main/scala/coop/rchain/rholang/interpreter/merging/RholangMergingLogic.scala
```diff
@@ -0,0 +1,215 @@
...
+          .map(_.flatten.toMap)
+      }
+    }
+  }
+
+  /**
+    * Merge number channel value from multiple changes and base state.
+    *
+    * @param channelHash Channel hash
+    * @param changes Channel changes
```
nzpr: Maybe It would be good to add comment here that changes are only needed to merge random generators. The value change is in `diff`, which is not that visible. Or maybe swap diff and changes?

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
nzpr: I think there is no need to subtract data from both indices ` -- a.producesMergeable -- b.producesMergeable`  
Might be this is good place for sanity check 
```scala
val result = a.producesConsumed intersect b.producesConsumed -- a.producesMergeable 
assert((b.producesMergeable intersect result).isEmpty, "produce is mergeable in one index but not in another")
```

rspace/src/main/scala/coop/rchain/rspace/merger/StateChangeMerger.scala
nzpr: Can we move logic for mergeable channels (handleChannelChange) into `combine` method for state change? So here API is the same but StateChange contains data with merge.

casper/src/test/scala/coop/rchain/casper/merging/MergeNumberChannelSpec.scala
```diff
@@ -0,0 +1,259 @@
...
+        // Create store actions / uses handler for number channels
+        computeTrieActions = (changes: StateChange, mergeableChs: NumberChannelsDiff) => {
+          StateChangeMerger
+            .computeTrieActions(changes, baseReaderBinary, mergeableChs, overrideTrieAction)
+        }
+
+        applyTrieActions = (actions: Seq[HotStoreTrieAction]) =>
+          rm.getHistoryRepo.reset(baseCp.root).flatMap(_.doCheckpoint(actions).map(_.root))
+
+        r <- ConflictSetMerger.merge[F, DeployChainIndex](
```
nzpr, line 206: This ConflictSetMerger will be removed in state casper, just want to mention

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -68,13 +61,13 @@ object RholangAndScalaDispatcher {
     (dispatcher, reducer)
   }
 
-  def create2[M[_]: Sync: Parallel: _cost](
+  def apply[M[_]: Sync: Parallel: _cost](
```
nzpr, line 64: Cool, so this allows overloading, right?

node/src/test/scala/coop/rchain/node/mergeablity/MergingBranchMergerSpec.scala
```diff
@@ -423,19 +424,27 @@ class MergingBranchMergerSpec extends FlatSpec with Matchers {
...
-
+                          val seqNum        = b.seqNum
+                          val sender        = b.sender
                           for {
+                            numberChsData <- runtimeManager.loadMergeableChannels(
+                                              postStateHash,
+                                              sender.toByteArray,
+                                              seqNum
+                                            )
+                            // TODO: Something is wrong with the number of deploys/sysDeploys when block is created
```
nzpr: Could you elaborate whats the problem?

zsluedem:
Since `Purse` layer in the middle doesn't help at all in all kinds of perspectives, I think we should remove the `Purse` layer and make `transfer` would just be simple number increase and decrease which would make transfer much faster.

casper/src/main/scala/coop/rchain/casper/storage/RNodeKeyValueStoreManager.scala
tgrospic: Yes, I'll correct this.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
tgrospic: Yes, this works!

rholang/src/main/scala/coop/rchain/rholang/interpreter/merging/RholangMergingLogic.scala
tgrospic: I agree that _diff_ parameter is not visible like other parameters. I'll add more comments for changes and try to move _diff_ together with others.
I did it this way so that calling site can partially apply it but with Scala this can make it even worse. :man_facepalming:

casper/src/main/resources/NonNegativeNumber.rho
tgrospic: We don't want to allow creation of mergeable channels for everyone, but only in specific case like in NonNegativeNumber contract. At least for now.

shared/src/main/scala/coop/rchain/catscontrib/monadOps.scala
tgrospic: Uh, it's a typo, I meant **lift** F[A]. Fixed. I've added explanation in comments with the example https://github.com/rchain/rchain/pull/3492/commits/95e73d65d63873098c6138bdc4c42b5edead21fc.

rspace/src/main/scala/coop/rchain/rspace/merger/StateChangeMerger.scala
tgrospic: No, because StateChange is inside RSpace and we don't have access to types defined by Rholang project. So this to apply logic of merging on the Rholang level inside RSpace.

shared/src/main/scala/coop/rchain/catscontrib/ApplicativeError_.scala
tgrospic: ApplicativeError_.scala is removed in https://github.com/rchain/rchain/commit/1d087b1890a91cd86ebe5522acc478c96c3e54cd.

rholang/src/main/scala/coop/rchain/rholang/interpreter/merging/RholangMergingLogic.scala
tgrospic: I've changed order of parameters and more comments https://github.com/rchain/rchain/pull/3492/commits/8815804a8fdf52a56b4b8c9ded35a9ae4b6e03b0.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -28,12 +44,17 @@ object MergingLogic {
...
-        (a.consumesProduced intersect b.consumesProduced).toIterator.filterNot(_.persistent)
-      val produceRaces =
-        (a.producesConsumed intersect b.producesConsumed).toIterator.filterNot(_.persistent)
+      val consumes          = a.consumesProduced intersect b.consumesProduced
+      val mergeableConsumes = a.consumesMergeable intersect b.consumesMergeable
+      val consumeRaces      = (consumes diff mergeableConsumes).toIterator.filterNot(_.persistent)
+
+      val produces          = a.producesConsumed intersect b.producesConsumed
+      val mergeableProduces = a.producesMergeable intersect b.producesMergeable
+      val produceRaces      = (produces diff mergeableProduces).toIterator.filterNot(_.persistent)
```
tgrospic: Ref: https://github.com/rchain/rchain/pull/3492#discussion_r710308020

@nzpr I made a fix but without throwing exception. If channel is not mergeable in both indexes it can still proceed but it will be detected as normal conflict. Good catch!

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
tgrospic: Fixed in https://github.com/rchain/rchain/pull/3492/commits/8a71a4766f1917823be3751f80591ed0ab261ca5.

node/src/test/scala/coop/rchain/node/mergeablity/MergingBranchMergerSpec.scala
tgrospic: System deploys are empty in merging blocks so _numberChsData_ has records for 2 deploys and inside blocks is only 1.

I found the problem in generated block, here is fix https://github.com/rchain/rchain/pull/3492/commits/760cad43c7e8528338b1dbf1adb2498dedc840b6.

tgrospic:
> Since `Purse` layer in the middle doesn't help at all in all kinds of perspectives, I think we should remove the `Purse` layer and make `transfer` would just be simple number increase and decrease which would make transfer much faster.

@zsluedem We need to investigate what is needed for this simplification. Purse is protecting that only the same mints can interact.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
nzpr: But can it ever be the case?

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -44,15 +44,17 @@ object MergingLogic {
...
       */
     val racesForSameIOEvent = {
-      val consumeRaces = {
-        (a.consumesProduced intersect b.consumesProduced -- a.consumesMergeable -- b.consumesMergeable).toIterator
-          .filterNot(_.persistent)
-      }
-      val produceRaces =
-        (a.producesConsumed intersect b.producesConsumed -- a.producesMergeable -- b.producesMergeable).toIterator
-          .filterNot(_.persistent)
+      val consumes          = a.consumesProduced intersect b.consumesProduced
```
nzpr: just a wish - maybe call it sharedConsumes

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
tgrospic: We need to investigate if there is possibility that merging can have a problem when channel is created for the first time. For now number channel is recorded as mergeable even in this case.

nzpr (approved):
LGTM

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -28,12 +44,17 @@ object MergingLogic {
...
-        (a.consumesProduced intersect b.consumesProduced).toIterator.filterNot(_.persistent)
-      val produceRaces =
-        (a.producesConsumed intersect b.producesConsumed).toIterator.filterNot(_.persistent)
+      val sharedConsumes    = a.consumesProduced intersect b.consumesProduced
+      val mergeableConsumes = a.consumesMergeable intersect b.consumesMergeable
+      val consumeRaces      = (sharedConsumes diff mergeableConsumes).toIterator.filterNot(_.persistent)
+
+      val sharedProduces    = a.producesConsumed intersect b.producesConsumed
+      val mergeableProduces = a.producesMergeable intersect b.producesMergeable
+      val produceRaces      = (sharedProduces diff mergeableProduces).toIterator.filterNot(_.persistent)
```
zsluedem, line 57: Just curious. Does this fix https://discord.com/channels/391435113415573504/488440523326226433/898048166820184134 ?

tgrospic: I don't see connection of this change with the error for multiple values on number channel. I need to investigate this separately.

zsluedem (approved):
LGTM

PR #3497, opened 2021-08-03
Fix history trie encoding problem

### LINKED ISSUE #3491: Fix history (Trie) and cold store value hash
## Overview

Rholang state in RNode is stored in a Merkle tree with channels as keys and channel data as values (tuple space).

Merkle tree implementation has two components: History and Cold (leaf) store. History represents the tree structure with hashes on the leaf nodes that points to values in the Cold store.

History key is calculated from the hash value of its children nodes. Cold store key is calculated from the value itself that is stored on a key.
The problem is that input used to get the hash in History and Cold store does not represent the whole value (base type codec) but the sub-type which is missing discriminator encoded data.

This made LFS sync much harder, because validation needs to deserialize each value to get to the input for hashing.

For History.

https://github.com/rchain/rchain/blob/dd95fb0f014a4c254afe9476c2e810192fcaf1a4/rspace/src/main/scala/coop/rchain/rspace/state/RSpaceImporter.scala#L68-L78

And Cold store.

https://github.com/rchain/rchain/blob/dd95fb0f014a4c254afe9476c2e810192fcaf1a4/rspace/src/main/scala/coop/rchain/rspace/state/RSpaceImporter.scala#L99-L107

## Solution

Use base type codec to serialize the value and use as input for hashing.

Remove additional work done in LFS validation.

### History (Trie)

Hash of the `Trie` type is executed lazily and stored on the object itself. Function to calculate hash is just matching on sub-type and reads this value. The bug is that sub-type is using its own specific codec instead of the base type `Trie`.

https://github.com/rchain/rchain/blob/01c4a1e6d24fe0813377dabbe8031a806e5deeb6/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L105-L115

This should be fixed by using `Trie` codec to get the serialized value for hashing, because this is the complete binary that is stored.

`PointerBlock`

https://github.com/rchain/rchain/blob/01c4a1e6d24fe0813377dabbe8031a806e5deeb6/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L88

```scala
lazy val encoded: BitVector = codecTrie.encode(this).getUnsafe
```

`Skip`

https://github.com/rchain/rchain/blob/01c4a1e6d24fe0813377dabbe8031a806e5deeb6/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L72

```scala
lazy val encoded: BitVector = codecTrie.encode(this).getUnsafe
```

Surprisingly `EmptyTrie` is using correct codec.

https://github.com/rchain/rchain/blob/01c4a1e6d24fe0813377dabbe8031a806e5deeb6/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L42

### Cold store 

Here is the example for Produce (it's similar with Consume and Joins) where the value that is used as input for hashing is not the serialized `PersistedData` as the base type but the specific inherited class.

https://github.com/rchain/rchain/blob/ad9514a903dcff0e265c0a509ab5adfae5b26b5d/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala#L114-L120

This should be fixed by using `PersistedData` codec to get the serialized value for hashing, because this is the complete binary that is stored.

```scala
val dataLeaf    = DataLeaf(data)
val dataEncoded = codecPersistedData.encode(dataLeaf).getUnsafe.toByteVector
val dataHash    = Blake2b256Hash.create(dataEncoded)
```

zsluedem:
## Overview

This issue is trying to address https://github.com/rchain/rchain/issues/3491.

The issue itself explain very details on the problem. So I would skip explain here.

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/History.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/History.scala
@@ -69,7 +69,7 @@ sealed trait NonEmptyTrie extends Trie
 case object EmptyTrie extends Trie
 
 final case class Skip(affix: ByteVector, ptr: ValuePointer) extends NonEmptyTrie {
-  lazy val encoded: BitVector = codecSkip.encode(this).getUnsafe
+  lazy val encoded: BitVector = codecTrie.encode(this).getUnsafe
 
   lazy val hash: Blake2b256Hash = Blake2b256Hash.create(encoded.toByteVector)
 
@@ -85,7 +85,7 @@ final case class PointerBlock private (toVector: Vector[TriePointer]) extends No
 
   def countNonEmpty: Int = toVector.count(_ != EmptyPointer)
 
-  lazy val encoded: BitVector = codecPointerBlock.encode(this).getUnsafe
+  lazy val encoded: BitVector = codecTrie.encode(this).getUnsafe
 
   lazy val hash: Blake2b256Hash = Blake2b256Hash.create(encoded.toByteVector)
 
--- a/rspace/src/main/scala/coop/rchain/rspace/history/History.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/History.scala
@@ -104,8 +104,6 @@ object Trie {
 
   /**
     * Creates hash of Merkle Trie
-    *
-    * TODO: Fix encoding to use codec for the whole [[Trie]] and not for specific inherited variant.
     */
   def hash(trie: Trie): Blake2b256Hash =
     trie match {
--- a/rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
@@ -166,8 +166,6 @@ object ScodecSerialize {
     (codecConsume :: codecSeq(codecProduce) :: sortedSet(uint8) :: codecMap(codecProduce, int32))
       .as[COMM]
 
-  val codecSkip: Codec[Skip] = (codecByteVector :: codecTrieValuePointer).as[Skip]
-
   val memoizingSkipCodec: Codec[Skip] =
     Codec.apply((s: Skip) => Attempt.successful(s.encoded), codecSkip.decode)
 
@@ -180,27 +178,17 @@ object ScodecSerialize {
   /*
    * scodec for History types
    */
-
-  val codecPointerBlock: Codec[PointerBlock] =
-    vectorOfN(
-      provide(length),
-      codecTriePointer
-    ).as[PointerBlock]
-
-  val codecTrie: Codec[Trie] =
-    discriminated[Trie]
-      .by(uint2)
-      .subcaseP(0) {
-        case e: EmptyTrie.type => e
-      }(provide(EmptyTrie))
-      .subcaseP(1) {
-        case s: Skip => s
-      }(memoizingSkipCodec)
-      .subcaseP(2) {
-        case pb: PointerBlock => pb
-      }(memoizingPointerBlockCodec)
-
-  implicit def codecTriePointer: Codec[TriePointer] =
+  /**
+    * CodecTriePointer
+    *
+    * Encoded format:
+    * |head 2 bits|content|
+    * |0b00|(nothing)|  -> emptyPointer
+    * |0b01|32bytes|    -> LeafPointer
+    * |0b10|32bytes|    -> SkipPointer
+    * |0b11|32bytes|    -> NodePointer
+    */
+  val codecTriePointer: Codec[TriePointer] =
     discriminated[TriePointer]
       .by(uint2)
       .subcaseP(0) {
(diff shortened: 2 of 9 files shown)


rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
```diff
@@ -216,7 +195,31 @@ object ScodecSerialize {
...
 
-  implicit def codecTrieValuePointer: Codec[ValuePointer] =
+  /**
+    * FIXME There some problem with fixed length vector for `VectorOfN(provide(2), int32)` which when you encode
+    *        Vector(1) with only one element would be successful but decode the generated byteVector would fail.
+    *        example:
+    *         val codecVector = vectorOfN(provide(2), int32)
+    *         val encoded = codecVector.encode(Vector(1))  // success
+    *         val decoded = codecVector.decode(encoded.toOption.get)  // fail
+    *        There are limitation on creating method of PointerBlock which garantees length 256.
```
zsluedem: Maybe I should fix this now?

rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
```diff
@@ -166,41 +166,20 @@ object ScodecSerialize {
     (codecConsume :: codecSeq(codecProduce) :: sortedSet(uint8) :: codecMap(codecProduce, int32))
       .as[COMM]
 
-  val codecSkip: Codec[Skip] = (codecByteVector :: codecTrieValuePointer).as[Skip]
-
-  val memoizingSkipCodec: Codec[Skip] =
-    Codec.apply((s: Skip) => Attempt.successful(s.encoded), codecSkip.decode)
```
zsluedem, line 172: Remove this because it cause infinite loop on calling method which full the stack

zsluedem: Skip and PointerBlock are not abstract but concrete here.

rspace/src/main/scala/coop/rchain/rspace/history/History.scala
```diff
@@ -113,6 +113,8 @@ object Trie {
       case s: Skip           => s.hash
       case _: EmptyTrie.type => History.emptyRootHash
     }
+
+  def hash(bytes: ByteVector): Blake2b256Hash = Blake2b256Hash.create(bytes)
```
tgrospic: This function does not relates to Trie and importer can use Blake2b256Hash directly as it already does.

TODO comment can now be removed.

rspace/src/main/scala/coop/rchain/rspace/state/RSpaceImporter.scala
```diff
@@ -90,34 +85,30 @@ object RSpaceImporter {
...
+                  case DataLeaf(bytes)          => decodeDatums[A](bytes)
+                  case ContinuationsLeaf(bytes) => decodeContinuations[P, K](bytes)
+                }
+                .void
+                .handleErrorWith { ex: Throwable =>
+                  raiseErrorEx(
+                    s"Decode inner class of key: ${hash.bytes.toHex}.",
+                    ex
+                  )
+                }
```
tgrospic: Now, we don't need decoding part of the code anymore. When hash matches no other checks are needed.

rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
tgrospic: This is expected behavior with `provide` codec because information about length is not part of encoded data, so decoding the value which is not the same length as specified in _provide_ will fail.

Here is the example, using _provide_ will not encode length so decoder will expect the size 3.
```scala
val c1 = vectorOfN(provide(3), uint4)
c1.encode(Vector(1, 2, 3)).getUnsafe // BitVector(12 bits, 0x123)

val c2 = vectorOfN(uint4, uint4)
c2.encode(Vector(1, 2, 3)).getUnsafe // BitVector(16 bits, 0x3123)
```

Because PointerBlock is always created with fixed length (256), _provide_ codec can be used here.
https://github.com/rchain/rchain/blob/02c276ff4fe04415b0262e02fb3dd69ebda48533/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L120-L123

rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
tgrospic: :+1:

rspace/src/main/scala/coop/rchain/rspace/serializers/ScodecSerialize.scala
zsluedem: This is an unsafe function from my perspective. Anyway, just want to mention that.

zsluedem: I would remove the comment.

tgrospic (approved):
LGTM

PR #3498, opened 2021-08-04
Cleanup runtime codes

zsluedem:
## Overview

1. rename `emptyStateHash` to `preGenesisStateHash` because `emptyStateHash` doesn't represent empty state.
2. remove `initRegistry` arg in creating  RhoRuntime method which it should be only used in genesis and tests with registry.
3. Make Log[F] and Span[F] in smaller scope in RhoRuntimeSyntax.


casper/src/test/scala/coop/rchain/casper/util/rholang/DeployIdTest.scala
```diff
@@ -42,13 +43,7 @@ class DeployIdTest extends FlatSpec with Matchers {
...
     val result =
       runtimeManager
-        .use(
-          mgr =>
-            for {
-              hash <- RuntimeManager.preGenesisStateHashFixed.pure[Task]
-              res  <- mgr.captureResults(hash, d)
-            } yield res
-        )
+        .use(mgr => mgr.captureResults(emptyRootHash.toByteString, d))
```
zsluedem, line 46: tests like these don't need the registry so we could just use emptyRootHash for testing.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -493,6 +493,18 @@ object RhoRuntime {
...
+    * For a newly created rspace, you might need to bootstrap registry
+    * in the runtime to use rholang registry normally. Actually this initRegistry
+    * is not the only thing you need for rholang registry, after the bootstrap
+    * registry, you still need to insert registry contract on the rspace.
+    *
+    * For an exist rspace which bootstrap registry before, you can skip this.
+    *
+    * For some test cases, you don't need the registry then you can skip this
+    * init process which can be faster.
+    * @return
```
tgrospic, line 506: _Return_ can be removed if empty.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -201,7 +202,9 @@ object BlockApproverProtocol {
     (for {
       result                    <- EitherT(validate.pure[F])
       (blockDeploys, postState) = result
-      emptyStateHash            <- EitherT.right(RuntimeManager.emptyStateHashFixed.pure[F])
+      emptyStateHash <- EitherT.right(runtimeManager.spawnRuntime >>= { r =>
+                         r.preGenesisStateHash
+                       })
```
tgrospic, line 207: Why not use hard-coded value here?

On the other hand, we can compare hard-coded value to computed and throw error if not equal.

casper/src/main/scala/coop/rchain/casper/engine/BlockApproverProtocol.scala
```diff
@@ -201,7 +201,7 @@ object BlockApproverProtocol {
     (for {
       result                    <- EitherT(validate.pure[F])
       (blockDeploys, postState) = result
-      emptyStateHash            <- EitherT.right(RuntimeManager.emptyStateHashFixed.pure[F])
+      emptyStateHash            <- EitherT.right(RuntimeManager.preGenesisStateHashFixed.pure[F])
```
nzpr: I would better name it genesisPreStateHash, as its pre state for genesis.

node/src/test/scala/coop/rchain/node/mergeablity/ComputeMerge.scala
```diff
@@ -61,6 +61,7 @@ trait ComputeMerge {
       .use {
         case (runtime, _, historyRepo) =>
           for {
+            _           <- runtime.preGenesisStateHash
```
nzpr, line 64: Hm... I'm not sure what this is supposed to do?

zsluedem: Insert neccessary part for registry working

zsluedem: This  is not neccessary for basic mergebility test. It is only used for treeHashMao tests. I would come up with something else later.

nzpr: I mean, it looks weird.
Does this call invokes some effects on runtime and updates its internal state? As output is ignored. 
And also it is called `preGenesisStateHash ` so I expect it to just give me stateHash. Maybe rename to `initRegistry`?

zsluedem:
Not needed because of https://github.com/rchain/rchain/pull/3492

PR #3503, opened 2021-08-09
Rejection options compute bug fix + tests

nzpr:
## Overview
This PR introduces bug fix for rejection options calculations. Current rejection options compute logic were not complete and did not account all the cases. Tests also added to check correctness.

Benchmarks revealed that rejection options calculation in real node is slow, it turned out that operating on DeployChainIndex in comparison to, e.g. Integers slows down the code significantly. To overcome this issue, `equals` and `hashCode` overrides added to DeployChainIndex.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
@@ -8,6 +8,8 @@ import coop.rchain.rspace.history.HistoryRepository
 import coop.rchain.rspace.merger.{StateChange, _}
 import coop.rchain.rspace.syntax._
 
+import scala.util.Random
+
 final case class DeployIdWithCost(id: ByteString, cost: Long)
 
 /** index of deploys depending on each other inside a single block (state transition) */
@@ -59,4 +61,19 @@ object DeployChainIndex {
       stateChanges
     )
   }
+
+  def random: Iterator[DeployChainIndex] =
+    Iterator.continually[Int](Random.nextInt(10)).map { size =>
+      val deployIds = Range(0, size)
+        .map(
+          _ => ByteString.copyFrom(Array.fill(64)((scala.util.Random.nextInt(256) - 128).toByte))
+        )
+      DeployChainIndex(
+        deployIds.map(id => DeployIdWithCost(id, 0)).toSet,
+        Blake2b256Hash.fromByteArray(new Array[Byte](32)),
+        Blake2b256Hash.fromByteArray(new Array[Byte](32)),
+        EventLogIndex.empty,
+        StateChange.empty
+      )
+    }
 }
--- a/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
@@ -8,6 +8,7 @@ import coop.rchain.rspace.history.HistoryRepository
 import coop.rchain.rspace.merger.{StateChange, _}
 import coop.rchain.rspace.syntax._
 
+import java.util.Objects
 import scala.util.Random
 
 final case class DeployIdWithCost(id: ByteString, cost: Long)
@@ -18,16 +19,17 @@ final case class DeployChainIndex(
     preStateHash: Blake2b256Hash,
     postStateHash: Blake2b256Hash,
     eventLogIndex: EventLogIndex,
-    stateChanges: StateChange
+    stateChanges: StateChange,
+    private val hashCodeVal: Int
 ) {
   // equals and hash overrides are required to make conflict resolution faster, particularly rejection options calculation
   override def equals(obj: Any): Boolean = obj match {
-    case that: DeployChainIndex => that.deploysWithCost.map(_.id) == this.deploysWithCost.map(_.id)
+    case that: DeployChainIndex => that.deploysWithCost == this.deploysWithCost
     case _                      => false
   }
-
-  override def hashCode(): Int =
-    deploysWithCost.map(_.id).foldLeft(0)((acc, v) => acc + v.hashCode() * 31)
+  // caching hash code helps a lot to increase performance of computing rejection options
+  // TODO mysterious speedup of merging benchmark when setting this to some fixed value
+  override def hashCode(): Int = hashCodeVal
 }
 
 object DeployChainIndex {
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
```diff
@@ -17,7 +19,16 @@ final case class DeployChainIndex(
...
-)
+) {
+  // equals and hash overrides are required to make conflict resolution faster, particularly rejection options calculation
+  override def equals(obj: Any): Boolean = obj match {
+    case that: DeployChainIndex => that.deploysWithCost.map(_.id) == this.deploysWithCost.map(_.id)
+    case _                      => false
+  }
+
+  override def hashCode(): Int =
+    deploysWithCost.map(_.id).foldLeft(0)((acc, v) => acc + v.hashCode() * 31)
```
zsluedem: why `31` here?

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
```diff
@@ -147,23 +148,44 @@ object MergingLogic {
...
+              gatherRejOptions(remainingConflictsMap).map { v =>
+                v.copy(rejectedSoFar = rejectedAcc ++ v.rejectedSoFar)
+              }
+          }
+          // emit rejection option that do not have more conflicts, pass others further into the loop
+          .partition { case RejectionOption(_, cMap) => cMap.values.flatten.isEmpty }
+
+        (out.map(_.rejectedSoFar).toList, if (next.nonEmpty) Some(next) else None)
+      }
+      .flatMap(Stream.emits)
```
zsluedem, line 187: Is this line neccessary?

casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
```diff
@@ -50,4 +61,19 @@ object DeployChainIndex {
       stateChanges
     )
   }
+
+  def random: Iterator[DeployChainIndex] =
```
zsluedem, line 68: we can put this function in tests

rspace/src/test/scala/coop/rchain/rspace/merging/MergingLogic.scala
```diff
@@ -0,0 +1,50 @@
+package coop.rchain.rspace.merging
+
+import coop.rchain.rspace.merger.MergingLogic._
+import coop.rchain.shared.Stopwatch
+import org.scalatest.{FlatSpec, Matchers}
+
+class MergingLogic extends FlatSpec with Matchers {
+  // some random conflict maps and rejection options, computed manually
+  "rejection options" should "be computed correctly" in {
```
zsluedem: Thanks for the tests. Good work.

rspace/src/main/scala/coop/rchain/rspace/merger/MergingLogic.scala
nzpr: `.flatMap(Stream.emits)`? yes, because `unflold` returns Stream of Lists, and what I want is a flat stream. I think this is the only way to flatten such stream

casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
nzpr: Its just a prime number (used in other places to override hashCode). https://www.artima.com/articles/how-to-write-an-equality-method-in-java this is useful reference. I'm not sure this is the best implementation for hashCode for this case, just did as in other parts of the code. The goal here is to get good distribution of hash codes, so operations on Sets of these data is fast, as hash code defines bucket and then equals resolves collisions.

casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
nzpr: I encountered a problem with this when trying to create benchmark suite - you cannot easily reference test code from main project. So if defined in test, you cannot somewhere in benchmark code inside node project use it. Might be pushing benchmark suite in main project is not a good idea, but I'm not sure, as it would be nice to have the same binary for benchmarks and for node operation.

casper/src/main/scala/coop/rchain/casper/merging/DeployChainIndex.scala
tgrospic: There is helper function to build hash code from arrays. It's plain for loop so maybe little faster.
https://www.baeldung.com/java-hashcode#standard-hashcode-implementations

nzpr: I've tried several implementation. For some reason behaviour on live network and in synthetic test in this PR is different. E.g. if I remove benefit from hashCode completely, making it fixed number, test is much faster
```
override def hashCode(): Int = 0

5 => 2.980869 ms vs 585.726633 ms 
10 => 19.294805 ms vs 13.1713 ms 
15 => 37.885004 ms vs 14.579285 ms 
20 => 37.878952 ms vs 17.656214 ms 
25 => 126.036481 ms vs 58.800682 ms 
30 => 163.905676 ms vs 56.152606 ms 
35 => 329.156289 ms vs 68.314901 ms 
40 => 654.245934 ms vs 97.642741 ms 
45 => 1.58764128 sec vs 158.560366 ms 
50 => 3.521030067 sec vs 210.773154 ms 
```
but in real network slower
```
rejection options (215) in 1.11669206875 min 
rejection options (287) in 2.31602600755 min

```
This is the reference as per current code
```
// Current version
override def hashCode(): Int = deploysWithCost.map(_.id).foldLeft(0)((acc, v) => acc + v.hashCode() * 31)

2.033679 ms vs 605.306271 ms 
12.59387 ms vs 45.038413 ms 
58.595065 ms vs 42.617381 ms 
74.587689 ms vs 152.469986 ms 
46.475547 ms vs 695.93373 ms 
198.669401 ms vs 612.737869 ms 
573.537734 ms vs 2.038174953 sec 
897.53519 ms vs 8.299223488 sec 
3.634718809 sec vs 9.800560232 sec 
7.892337706 sec vs 19.930452911 sec 

rejection options (265) in 3.618607435 sec
rejection options (252) in 1.413084469 sec
```

We might want to optimise it in future, for now I just use helper as @tgrospic suggested.

rspace/src/test/scala/coop/rchain/rspace/merging/MergingLogic.scala
```diff
@@ -0,0 +1,50 @@
+package coop.rchain.rspace.merging
+
+import coop.rchain.rspace.merger.MergingLogic._
+import coop.rchain.shared.Stopwatch
+import org.scalatest.{FlatSpec, Matchers}
+
+class MergingLogic extends FlatSpec with Matchers {
```
tgrospic: Should we add suffix `Spec` or `Test` to testing classes? I think Spec is mostly used.

tgrospic (approved):
LGTM

Issue #3501, opened 2021-08-08
Refund time rises with number of bonded validators


nzpr:
## Overview
Profiling of state transitions having a single REV transfer revealed that increasing number of validators bonded in genesis increases execution time. It turned out that the root cause of the problem is refund, it generates more events.
PTAL at the following data for more details. As it can be seen, `rchain.casper.rho-runtime.refund` is rising. Precharge is rising as well but not drastically.

**1 bonded validator**
```sh
DONE: 1 State transitions (avg 1.0 TX per ST), user events: 620 P, 320 C 604 COMM, sys events : 54 P, 26 C 55 COMM, time: 3.098472169 sec.
           rchain.casper.rho-runtime.compute-state: avg 2.9989614 sec., total 2.9989614 sec., calls 1, 
            rchain.rholang.interpreter.reduce-term: avg 0.4428498 sec., total 1.7713993 sec., calls 4, 
  rchain.casper.rho-runtime.evaluate-system-source: avg 0.4797729 sec., total 1.4393188 sec., calls 3, 
                  rchain.casper.rho-runtime.refund: avg 0.6693681 sec., total 0.6693681 sec., calls 1, 
               rchain.casper.rho-runtime.precharge: avg 0.5722498 sec., total 0.5722498 sec., calls 1, 
                      rchain.rspace.locked-produce: avg 0.0006747 sec., total 0.4547590 sec., calls 674, 
                      rchain.rspace.locked-consume: avg 0.0011849 sec., total 0.4135375 sec., calls 349, 
             rchain.casper.rho-runtime.user-deploy: avg 0.3379702 sec., total 0.3379702 sec., calls 1, 
                rchain.rholang.runtime.create-play: avg 0.0945450 sec., total 0.0945450 sec., calls 1, 
                               rchain.rspace.reset: avg 0.0113873 sec., total 0.0227746 sec., calls 2, 
  rchain.rholang.interpreter.build-normalized-term: avg 0.0010468 sec., total 0.0041873 sec., calls 4, 
   rchain.casper.rho-runtime.consume-system-result: avg 0.0004258 sec., total 0.0012775 sec., calls 3, 
    rchain.rholang.interpreter.charge-parsing-cost: avg 0.0000153 sec., total 0.0000611 sec., calls 4, 
       rchain.rholang.interpreter.set-initial-cost: avg 0.0000008 sec., total 0.0000031 sec., calls 4, 
```
**2 bonded validators**
```sh
[...]

PR #3504, opened 2021-08-09
Benchmarks: concurrency + leaderful simulation

nzpr:
## Overview
This PR introduces two benchmarks:

1) **Node is following leaderful network**. Behaviour is identical, without networking layer. Given N validators, N-1 block added concurrently, then merged block added.  This pattern repeats. For hardening purposes, REV transfers are simulated in the network, and balances are verified after each new layer created.
```rnode bench leaderful --validators-num 30 --users-num 1000 --max-block-tx 1 --epoch-length 10000```
It is useful to test scaling. By adjusting `--validators-num` from 1 to X scaling ability can be estimated. In this case better to set epochLength to big value so it does not happen in test, because epoch change block takes longer to process. In addition `--max-block-tx 1` should be used, so each block is created with the same number of deploys.
The difference of this benchmark from real network is that each block creation in benchmark is play+replay+validation in real network, and networking overhead (which should not be big) is not accounted for.

2) **Scaling of concurrent block validation**. More simple test which just emulate concurrent block processing, without logic to harder merging (multiple transfers in block, balances validation).
```rnode bench concurrency --max-concurrent-tx 100```

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -292,19 +292,7 @@ in {
                                       match transferResult {
                                         (true, _) => {
                                           new newRewardsCh, pendingUpdatedCh in {
-                                            distributeRewards!(
-                                              initialPaymentAmount - refundAmount,
-                                              state.get("activeValidators"),
-                                              state.get("allBonds"),
-                                              *newRewardsCh
-                                            ) |
-                                            for (@newRewards <- newRewardsCh) {
-                                              // Add new pending rewards to validator pending rewards state
-                                              PendingRewards!("append", sender, newRewards, *pendingUpdatedCh) |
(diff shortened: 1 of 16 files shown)


nzpr:
https://github.com/rchain/rchain/pull/3504/commits/30d193cf74f5d3d8e08d16b7924fe8078626e1ff this commit removes rewards distribution for demonstration purposes which slows down computation significantly <https://github.com/rchain/rchain/issues/3501>, as it will be removed in https://github.com/rchain/rchain/pull/3492

nzpr:
https://github.com/rchain/rchain/pull/3504/commits/befd9c6207ee63ec8315aa3b55d3910eaf603f3b this commit makes simulation equal to live network, but with 0 overhead on network. Each state transition is done 3 times, representing play+replay on block sender + validation by peer.

PR #3509, opened 2021-08-14
Define reasons for propose. Enable attestation messages.

### LINKED ISSUE #3249: Enable confirmation/attestation blocks
## Overview
Finality time depends on how fast enough stake will be put on top of a block.
Currently node have to wait for new deploys to even confirm validation status of block it received. Also, this involves precharge and refund which makes it slower.
Node should be able to issue some flavour of lightweight block that servers for attestation purposes.

### Impact
Lower time to finalization

### Design
It looks reasonable to allow node to issue blocks with a single CloseBlock system deploy, without any user deploys. This block might have some additional service information.

nzpr:
## Overview
Define reasons for block creation.
Another approach to https://github.com/rchain/rchain/pull/3506 and https://github.com/rchain/rchain/pull/3452.

Motivation for this PR is upcoming block merge PR, which, given current bock producing logic, leads to high resource requirements. At the moment all blocks should contain some state transition, so include precharge and refund, which are costly until efficient REV vault inside History is implemented (as planned for HF3). 

Another motivation is making network lean, so not spending resources unless it serves the purpose of blockchain - maintaining state that network is agree on. Block merge release emphasises this issue.

Another point to add is that block merge not only motivated this solution, but also made it possible.

Attestation messages logic boils down to allowing proposing messages without user deploy and defining cases where this is appropriate. This replaces requirements for attestation messages and difficulties tied to using synchrony constraint. it looks like more elegant and straightforward solution.

**Please find description of proposed solution below.**

Casper messages serve only two purposes:
1. offering state transitions for validation by the network, and
2. participating in finalization of past state transitions.

Therefore message producing makes sense and should be allowed only if it is justified by these reasons.

Given said above here are rules for detecting if propose is justified:
1. State transition is offered (reason to create `STM` from from https://github.com/rchain/rchain/pull/3452).
   **a. there is a user deploy to be included in a block, or
   b. parents have different post state, so merge block should be issued.**
   Both of these rules can be validated, so offence an be detected.
2. Finalization of user deploys (reaching acquiescence) is not done. (reason to create AM from https://github.com/rchain/rchain/pull/3452.)
   **There are non finalized messages which do not share post state with some finalized message. So basically, if there are non finalized states known.**
   This rule also can be validated and offence detected.
   *TODO this might be narrowed down to "if sender already propagated its weight down to all states known", not finalization requirement*

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -18,6 +18,7 @@ import coop.rchain.metrics.{Metrics, MetricsSemaphore}
 import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.EquivocationRecord.SequenceNumber
 import coop.rchain.models.Validator.Validator
+import coop.rchain.models.block.StateHash.StateHash
 import coop.rchain.models.{BlockHash, BlockMetadata, EquivocationRecord, Validator}
 import coop.rchain.shared.syntax._
 import coop.rchain.models.syntax._
@@ -44,7 +45,9 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       heightMap: SortedMap[Long, Set[BlockHash]],
       invalidBlocksSet: Set[BlockMetadata],
       lastFinalizedBlockHash: BlockHash,
-      finalizedBlocksSet: Set[BlockHash]
+      finalizedBlocksSet: Set[BlockHash],
+      // how many times particular valid state is met in DAG as postState for block
+      validStatesCounter: Map[StateHash, Int]
   ) extends BlockDagRepresentation[F] {
 
     def lookup(blockHash: BlockHash): F[Option[BlockMetadata]] =
@@ -135,6 +138,20 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
         }
         truncatedFinalizesSet = finalizedBlocksSet diff excessSet diff nonFinalizedSet
 
+        excessStates <- excessSet.toList.traverse(
+                         this.lookupUnsafe(_).map(_.postStateHash)
+                       )
+        newValidStatesCounter = excessStates.foldLeft(validStatesCounter) {
+          case (acc, s) =>
+            assert(
+              acc.contains(s),
+              s"validStatesCounter does not have state from excess message when truncating DAG."
+            )
+            val curV = acc(s)
+            val newV = curV - 1
+            if (newV == 0) acc - s else acc.updated(s, curV - 1)
+        }
+
         view = this.copy(
           lastFinalizedBlockHash = lfb,
           finalizedBlocksSet = truncatedFinalizesSet,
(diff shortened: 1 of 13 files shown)


block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -186,6 +204,22 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+        // how many time each state is met across non finalized blocks
+        stateSeenInNonFinalizedCounter = statesOfNotFinalizedBlocks
+          .groupBy(identity)
+          .mapValues(_.size)
+
+        // if all occurrences of a state are in not finalized blocks - state is not finalized, no acquiescence
+        hasNotFinalizedState = stateSeenInNonFinalizedCounter.exists {
+          case (state, qty) => validStatesCounter(state) == qty
+        }
+      } yield !hasNotFinalizedState
```
tgrospic, line 222: What will be the state in non-finalized blocks when DAG is in acquiescence state?

Can we just look at non-finalized blocks to calculate this?

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -68,6 +69,35 @@ trait Casper[F[_]] {
...
+    *    TODO this can be narrowed down to "if sender already propagated its weight down to all states known"
+    */
+  def shouldPropose[F[_]: Monad](
+      s: CasperSnapshot[F],
+      deploysWaiting: F[Set[Signed[DeployData]]]
+  ): F[Boolean] = {
+    val hasParentsToMerge = (s.parents.map(_.body.state.postStateHash).distinct.size > 1).pure[F]
+    val hasDeploysToOffer = deploysWaiting.map(_ -- s.deploysInScope).map(_.nonEmpty)
+    val noAcquiescence    = s.dag.reachedAcquiescence.not
+    hasDeploysToOffer ||^ hasParentsToMerge ||^ noAcquiescence
```
tgrospic: Do we need to cover slashing deploys also?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
nzpr: On acquiescence non finalized blocks have the same state as some (one or several) finalized blocks. 

We need to be sure that state is met only in non finalized blocks. Thats why we need to have a visibility of which states are in finalized blocks. Or (as implemented) - have counter of how many times state is met, so by looking into non finalized blocks we are able to know that all occurrences are in non finalized blocks.

casper/src/main/scala/coop/rchain/casper/Casper.scala
nzpr: Yes, good catch!

nzpr: https://github.com/rchain/rchain/pull/3509/commits/6afae3cb02e9ac6bf28fcdd4bf8e9fc1528922dd

tgrospic (approved):
LGTM

PR #3512, opened 2021-08-31
[WIP] Partial state Casper

nzpr:
## Overview
New approach to Casper, consensus on state transitions.


casper-v2/src/main/scala/coop/rchain/casper/v2/core/DependencyGraph.scala
```diff
@@ -0,0 +1,57 @@
...
+import cats.effect.Sync
+import cats.syntax.all._
+
+trait DependencyGraph[F[_], M] {
+
+  /** Each DAG message is justified by set of other messages. */
+  def justifications(message: M): F[List[M]]
+
+  /**
+    * Parent is a justification which cannot be viewed through other justifications.
```
nzpr, line 11: I'm not sure If we should call this parent - its first observation of the message in the whole view, parent might be quite misleading imo.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -106,22 +106,30 @@ message BlockHashMessageProto {
...
+}
+
 message BlockMetadataInternal {
   // This message in mapped to a different Scala class because of protobuf's inability to create map<bytes, int64> for
   // bonds.
   option (scalapb.message).type = "coop.rchain.models.BlockMetadata";
 
   bytes blockHash                       = 1;
-  repeated bytes parents                = 2 [(scalapb.field).collection_type="collection.immutable.List"];
+  bytes postStateHash                   = 2;
```
zsluedem, line 127: I am very curious what is the intention to remove parents?

zsluedem: oh. It is just meta data.But still curious

nzpr: We do not need parents - only justifications. Parents were computed by Estimator with ranking, to compute main parent, which we do not have now. Some version of parents might be an optimisation for traversal of the DAG to not traverse multiple times through the same message if justification A sees justification B, but this is not done now, and will require some more thinking with regards to distance from the tips.

But long story short - we do not need a notion of parents. It should be removed from block as well.

zsluedem: I am not sure I am following.

nzpr: What is the purpose of having parents?

nzpr: If its not clear about distance - we need to find highest messages, and this `highest` can be detected by measuring Stream length when traversing. https://github.com/rchain/rchain/pull/3512/commits/7481cb404bbc3c1019ba3d6790451585785d5dcb#diff-8ca81baa47046b2e5e687927b9ad064e12dd12af17415e3003adc662d45e979dR48-R57

casper-v2/src/test/scala/CasperSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+    0 -> List()
+  )
+  val sendersMap: Map[Message, Validator] = Map(
+                            32 -> 2,                33 -> 3,
+    21 -> 1,                22 -> 2,
+                            12 -> 2,                13 -> 3,
+    0 -> 1
+  )
+  val seqNumsMap: Map[Message, Int] = Map(
+                            32 -> 3,                33 -> 3,
```
zsluedem, line 32: why `33 -> 3` is `->3` here? Not `2`?

casper-v2/src/main/scala/coop/rchain/casper/v2/core/syntax/CasperSyntax.scala
```diff
@@ -0,0 +1,75 @@
...
+
+trait CasperSyntax {
+  implicit final def casperSyntax[F[_], M, S](c: Casper[F, M, S]): CasperOps[F, M, S] =
+    new CasperOps[F, M, S](c)
+}
+
+final class CasperOps[F[_], M, S](val c: Casper[F, M, S]) extends AnyVal {
+  import c._
+
+  def messageScope(
```
zsluedem: @tgrospic Not sure if this is a good place to use `syntax` here. This method doesn't really need `Casper` but only `faultToleranceThreshold`. It would be better to put `faultToleranceThreshold` directly in argument.

shared/src/main/scala/coop/rchain/fs2/Fs2StreamSyntax.scala
```diff
@@ -12,6 +12,19 @@ import scala.concurrent.duration.{DurationInt, FiniteDuration}
 trait Fs2StreamSyntax {
   implicit final def sharedSyntaxFs2Stream[F[_], A](stream: Stream[F, A]): Fs2StreamOps[F, A] =
     new Fs2StreamOps[F, A](stream)
+
+  def zipStreamList[F[_], A](streams: List[Stream[F, A]]): Stream[F, A] =
```
zsluedem, line 16: I am not very sure what you are trying to do here.

```
  def zipStreamList[F[_], A](streams: List[Stream[F, A]]): Stream[F, A] =
    streams.foldLeft[Stream[F,A]](Stream.empty) { (acc, s) => acc ++ s}
```
I think this is enough.

casper-v2/src/test/scala/SafetyOracleSpec.scala
```diff
@@ -0,0 +1,83 @@
...
+    }
+  implicit val s = Sync[Task]
+
+  // Oracle that does not account compatibility and just uses justifications to propagate agreements
+  val o = oracle[Task](AlwaysCompatible)
+  val d = dag[Task]
+
+  val fts = List(
+    List((13, -1.0), (21, -1.0), (22, -1.0)),
+    List((0, 1.0), (12, 1.0), (13, 1.0)),
```
zsluedem, line 61: Does this line mean `0`, `12` and `13` is finalized?

casper-v2/src/test/scala/CasperSpec.scala
nzpr: Yes, you are right, good catch! Thought it should not influence the test

casper-v2/src/test/scala/SafetyOracleSpec.scala
nzpr: Yes

shared/src/main/scala/coop/rchain/fs2/Fs2StreamSyntax.scala
nzpr: Yes you are right, it should be just a function, not need to put in a syntax

casper-v2/src/test/scala/SafetyOracleSpec.scala
zsluedem: These 3 blocks are the `FinalizedFringe` , right?

zsluedem: And the deploy within these blocks are ok to consider finalized?

nzpr: Yes, they are. Each block(message) represents a state transition (or just state). So once message is finalized - state transition is finalized. What is rejected - rejected finally, what is accepted and proposed - finaized.

nzpr:
Closed as another implementation is merged https://github.com/rchain/rchain/blob/10d6383b6da0c6a697a5239fc28bdb3dc937e8bf/block-storage/src/main/scala/coop/rchain/blockstorage/dag/Finalizer.scala

PR #3517, opened 2021-10-01
New transaction balance script

### LINKED ISSUE #3490: Remove old python typed transaction store and rewrite rev vault export

zsluedem:
## Overview

This is a new replacement of the old transaction balance script.

Shortcoming of the old version:
1. Hard-coded transaction on old mainnet
2. Only applicable the old mainnet

New Version:
1. Can be used in all kinds of private network with transaction store enabled
2. No hard-coded transaction 


hilltracer (approved):
I can't say anything useful, sorry

node/src/main/scala/coop/rchain/node/revvaultexport/reporting/TransactionBalances.scala
```diff
@@ -318,32 +230,64 @@ object TransactionBalances {
...
+              ByteString
+                .copyFrom(
+                  Base16.unsafeDecode(blockHash)
+                )
+                .pure[F]
+            case SlashingDeploy(blockHash) =>
+              ByteString
+                .copyFrom(
+                  Base16.unsafeDecode(blockHash)
+                )
```
tgrospic: Two functions seems to be duplicated here, one to convert String to ByteString and lookup for block hash by deploy ID. It would make sense to extract them as helper functions in the same scope to make it more clear what are differences between transaction types.

node/src/main/scala/coop/rchain/node/revvaultexport/reporting/TransactionBalances.scala
```diff
@@ -318,32 +230,64 @@ object TransactionBalances {
...
+                .pure[F]
+          }
+        allTransactions.toList.traverse(
+          t =>
```
tgrospic: With curly braces `t` don't have to be in a new line. :wink:

tgrospic (commented):
Very nice simplification!

shared/src/main/scala/coop/rchain/shared/ByteStringOps.scala
```diff
@@ -20,4 +20,9 @@ object ByteStringOps {
   implicit class RichByteArray(arr: Array[Byte]) {
     def toByteString: ByteString = ByteString.copyFrom(arr)
   }
+
+  implicit class RichHexString(v: String) {
+    def unsafeToByteString: ByteString   = ByteString.copyFrom(Base16.unsafeDecode(v))
+    def toByteString: Option[ByteString] = Base16.decode(v).map(ByteString.copyFrom)
+  }
```
tgrospic: Please use _new_ (cats) way of defining extensions. We can have the same ByteStringSyntax in shared project or maybe it's enough to use this in _models_ project.
https://github.com/rchain/rchain/blob/22b937a2f8e49d01b6061887345bbe16a06be506/models/src/main/scala/coop/rchain/models/ByteStringSyntax.scala#L9-L30

models/src/main/scala/coop/rchain/models/StringSyntax.scala
```diff
@@ -0,0 +1,16 @@
...
+import coop.rchain.shared.Base16
+
+trait StringSyntax {
+  implicit final def modelsSyntaxString(s: String): StringOps =
+    new StringOps(s)
+
+}
+class StringOps(private val s: String) extends AnyVal {
+  def unsafeToByteString: ByteString   = ByteString.copyFrom(unsafeDecode16)
+  def toByteString: Option[ByteString] = decode16.map(ByteString.copyFrom)
```
tgrospic: These names are not precise enough. String representation can be in different encoding like utf8, ASCII or hex.

Better to call it something like `hexToByteString` so if necessary we can add `utf8ToByteString`.

models/src/main/scala/coop/rchain/models/StringSyntax.scala
```diff
@@ -0,0 +1,16 @@
...
+  def decode16: Option[Array[Byte]]    = Base16.decode(s)
+  def unsafeDecode16: Array[Byte]      = Base16.unsafeDecode(s)
```
tgrospic: I'd be more for `decodeHex` then just 16. As an example, `ByteVector` from scodes have `toHex` function. But this is just my preference.

node/src/main/scala/coop/rchain/node/configuration/commandline/Base16Converter.scala
```diff
@@ -9,8 +9,7 @@ object Base16Converter extends ValueConverter[Array[Byte]] {
       case (name, strings) => strings.map((name, _))
     } match {
       case List((name, v)) =>
-        Base16
-          .decode(v.toLowerCase)
+        v.toLowerCase.decode16
```
tgrospic: Not sure why is necessary to make it lower case. Hex encoding is case insensitive.

rholang/src/test/scala/coop/rchain/rholang/interpreter/ReduceSpec.scala
```diff
@@ -1076,7 +1076,7 @@ class ReduceSpec extends FlatSpec with Matchers with AppendedClues with Persiste
   "eval of bytesToHex" should "transform byte array to hex string (not the rholang term)" in {
     val splitRand    = rand.splitByte(0)
     val base16Repr   = "0123456789abcdef"
-    val testBytes    = ByteString.copyFrom(Base16.decode(base16Repr).get)
+    val testBytes    = ByteString.copyFrom(base16Repr.unsafeDecode16)
```
tgrospic: Can we use new `hexToByteSting` extension here?

casper/src/main/scala/coop/rchain/casper/api/MachineVerifiableDag.scala
```diff
@@ -1,21 +1,14 @@
...
 
 import coop.rchain.casper._
-import coop.rchain.casper.protocol._
-import coop.rchain.casper.util.ProtoUtil
-import coop.rchain.crypto.codec.Base16
 import coop.rchain.models.BlockHash.BlockHash
-import cats._, cats.data._, cats.implicits._
+import coop.rchain.models.syntax._
+import cats._
+import cats.implicits._
```
tgrospic: When you changing these lines can you please switch to `import cats.syntax.all._` instead of _implicits_.

For new developers it's better that we have correct version. Also I don't think *cats._* is needed.

node/src/main/scala/coop/rchain/node/revvaultexport/reporting/TransactionBalances.scala
```diff
@@ -318,32 +230,46 @@ object TransactionBalances {
...
+      allTransactions    = allTransactionsMap.values.flatMap(_.data)
+      blockDagStorage    <- BlockDagKeyValueStorage.create[F](rnodeStoreManager)
+      dagRepresantation  <- blockDagStorage.getRepresentation
+      allWrappedTransactions <- {
+        def findTransaction(transaction: TransactionInfo): F[ByteString] =
+          transaction.transactionType match {
+            case PreCharge(deployId) =>
+              dagRepresantation
+                .lookupByDeployId(deployId.unsafeToByteString)
+                .map(_.get)
```
tgrospic: Please don't use directly `get` method on option type because error in node case will not have any info, instead raise an error with some additional context like deployId. Also this lookup can be extracted in small helper function so that error message is not repeated.

node/src/main/scala/coop/rchain/node/revvaultexport/reporting/TransactionBalances.scala
```diff
@@ -318,32 +230,46 @@ object TransactionBalances {
...
+            case CloseBlock(blockHash) =>
+              blockHash.unsafeToByteString.pure[F]
+            case SlashingDeploy(blockHash) =>
+              blockHash.unsafeToByteString.pure[F]
+          }
+        allTransactions.toList.traverse { t =>
+          for {
+            blockHash    <- findTransaction(t)
+            blockMetaOpt <- dagRepresantation.lookup(blockHash)
+            blockMeta    = blockMetaOpt.get
```
tgrospic: This option can also be lifted to an error in F when None.

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -207,7 +207,7 @@ object Setup {
         val dummyDeployerKeyOpt       = conf.dev.deployerPrivateKey
         val dummyDeployerKey =
           if (dummyDeployerKeyOpt.isEmpty) None
-          else PrivateKey(Base16.decode(dummyDeployerKeyOpt.get).get).some
+          else PrivateKey(dummyDeployerKeyOpt.get.unsafeDecode16).some
```
tgrospic: This can be simplified.
```scala
val dummyDeployerKey = dummyDeployerKeyOpt.flatMap(Base16.decode(_)).map(PrivateKey(_))
```

node/src/main/scala/coop/rchain/node/configuration/commandline/Base16Converter.scala
zsluedem: not sure.

models/src/main/scala/coop/rchain/models/StringSyntax.scala
```diff
@@ -0,0 +1,16 @@
...
+import com.google.protobuf.ByteString
+import coop.rchain.shared.Base16
+
+trait StringSyntax {
+  implicit final def modelsSyntaxString(s: String): StringOps =
+    new StringOps(s)
+
+}
+class StringOps(private val s: String) extends AnyVal {
+  def unsafeToByteString: ByteString      = ByteString.copyFrom(unsafeDecodeHex)
```
tgrospic: This should be `unsafeHexToByteString`, just with prefix _unsafe_ like decode. Please add a line of separation between these two converters.

tgrospic (approved):
LGTM

Thanks for all addressing all comments!

zsluedem:
bors merge

zsluedem:
bors merge

PR #3521, opened 2021-10-18
Use DAG to find BlockMessage by truncated hash string to increase performance

### LINKED ISSUE #3495: Implement `getBlock` by hash via DAG representation (dagSet).
## Overview

Currently we iterate over all records when block hash is searched with prefix (`BlockStore#find`). This is slow because hashes are read from LMDB database via BlockStore implementation.

We have in `DagBlockRepresentation` all block hashes in memory to do the same operation much much faster.

## Solution

Implement search for a block with block hash prefix with `BlockDagRepresentation`.

EDIT: Extend `BlockDagRepresentation` with the method to find complete hash from hash prefix and use it directly in `BlockAPI`.

stanislavlyalin:
## Overview

Solution uses hash lookups from `BlockDagRepresentation`. A `find` method has been added in `BlockDagRepresentation` and implemented in `KeyValueDagRepresentation`, which takes a truncated hash and returns an optional `BlockHash`

### Performance results
| node | find block request time | factor |
|---|---|---|
| Main net, node0 | ~10.2s | 102x |
| Main net, observer-us | ~100ms | 1x |

bors r+
bors try
bors delegate+

### DIFF
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
@@ -37,31 +37,6 @@ class KeyValueBlockStore[F[_]: Sync](
     } yield block
   }
 
-  override def find(p: BlockHash => Boolean, n: Int): F[Seq[(BlockHash, BlockMessage)]] = {
-    import cats.instances.list._
-    for {
-      filteredIndex <- store.iterate { iterator =>
-                        iterator
-                          .map { case (k, v) => (ByteString.copyFrom(k), v) }
-                          .withFilter { case (key, _) => p(key) }
-                          // Decode value only for filtered elements
-                          .map { case (k, v) => (k, ByteVector(v)) }
-                          // Return only the first n results / stop iteration
-                          .take(n)
-                          .toList
-                      }
-
-      result <- filteredIndex.traverse {
-                 case (hash, bytes) =>
-                   // Convert bytes to proto object, it's fatal error if fails
-                   bytesToBlockProto(bytes.toArray) map
-                     BlockMessage.from flatMap
-                     (_.leftMap(errorBlock(hash)).liftTo[F]) map
-                     ((hash, _))
-               }
-    } yield result
-  }
-
   override def put(data: => (BlockHash, BlockMessage)): F[Unit] = Sync[F].defer {
     def toBuffer(b: BlockMessage) = blockProtoToBuffer(b.toProto)
     val (hash, block)             = data
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -13,6 +13,7 @@ import coop.rchain.blockstorage.syntax._
 import coop.rchain.blockstorage.util.BlockMessageUtil._
 import coop.rchain.casper.PrettyPrinter
 import coop.rchain.casper.protocol.BlockMessage
+import coop.rchain.crypto.codec.Base16
 import coop.rchain.metrics.Metrics.Source
 import coop.rchain.metrics.{Metrics, MetricsSemaphore}
 import coop.rchain.models.BlockHash.BlockHash
@@ -74,6 +75,10 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
     def isFinalized(blockHash: BlockHash): F[Boolean] =
       finalizedBlocksSet.contains(blockHash).pure[F]
+
+    override def find(truncatedHash: String): F[Option[BlockHash]] = Sync[F].delay {
+      dagSet.find(hash => Base16.encode(hash.toByteArray).startsWith(truncatedHash))
+    }
 
     def topoSort(
         startBlockNumber: Long,
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
@@ -77,7 +77,8 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
       finalizedBlocksSet.contains(blockHash).pure[F]
 
     override def find(truncatedHash: String): F[Option[BlockHash]] = Sync[F].delay {
-      dagSet.find(hash => Base16.encode(hash.toByteArray).startsWith(truncatedHash))
+      val truncatedByteString = ByteString.copyFrom(Base16.unsafeDecode(truncatedHash))
+      dagSet.find(hash => hash.startsWith(truncatedByteString))
     }
 
     def topoSort(
--- a/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
+++ b/block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
@@ -44,6 +44,7 @@ trait BlockDagRepresentation[F[_]] {
   // DAG representation has to have finalized block, or it does not make sense
   def lastFinalizedBlock: BlockHash
   def isFinalized(blockHash: BlockHash): F[Boolean]
+  def find(truncatedHash: String): F[Option[BlockHash]]
 }
 
 trait EquivocationsTracker[F[_]] {
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -631,15 +631,20 @@ object BlockAPI {
 
   // Be careful to use this method , because it would iterate the whole indexes to find the matched one which would cause performance problem
   // Trying to use BlockStore.get as much as possible would more be preferred
-  private def findBlockFromStore[F[_]: Monad: BlockStore](
+  private def findBlockFromStore[F[_]: Monad: EngineCell: BlockStore](
       hash: String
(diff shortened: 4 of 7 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -633,13 +633,7 @@ object BlockAPI {
...
   private def findBlockFromStore[F[_]: Monad: BlockStore](
       hash: String
-  ): F[Option[BlockMessage]] =
-    for {
-      findResult <- BlockStore[F].find(h => Base16.encode(h.toByteArray).startsWith(hash), 1)
-    } yield findResult.headOption match {
-      case Some((_, block)) => Some(block)
-      case None             => none[BlockMessage]
-    }
+  ): F[Option[BlockMessage]] = BlockStore[F].find(hash)
```
tgrospic: Exposing `BlockDagRepresentation` with all block hashes is what we want so here we can get it directly via Casper instance, similar how it's done in other methods like `findDeploy`.

That way we can completely remove `find` method from `BlockStore` without the need to change anything in `Setup`.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
```diff
@@ -19,6 +19,7 @@ trait BlockDagStorage[F[_]] {
       direct: BlockHash,
       finalizationEffect: Set[BlockHash] => F[Unit]
   ): F[Unit]
+  def find(truncatedHash: String): F[Option[BlockHash]]
```
tgrospic: This `find` method can be directly implemented on `BlockDagRepresentation` and called from `BlockAPI` in which case _dagSet_ can stay private.

tgrospic:
@stanislavlyalin A few tips for creating PRs.

- Please use the existing PR template with the checklist and bors info.
- PR title should be short summary of the changes.
- Linked issue can be selected in the right column and don't have to be part of the description.
- PR overview should contain textual description of the changes that should help non-developers to have basic understanding what PR is changing, useful to be used in weekly debrief and specific enough for developers to later understand why changes are done in this way.

PR can be created without linked issue in which case it should also contain more elaborate description which is usually part of the issue.

For questions in the PR you can use comments or comments related to some specific line of code.

block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
```diff
@@ -98,14 +104,19 @@ class KeyValueBlockStore[F[_]: Sync](
...
-  def apply[F[_]: Sync](kvm: KeyValueStoreManager[F]): F[BlockStore[F]] =
+  def apply[F[_]: Sync](
+      kvm: KeyValueStoreManager[F],
+      hashResolver: Option[String => F[Option[BlockHash]]] = None
+  ): F[BlockStore[F]] =
     for {
       store              <- kvm.store("blocks")
       storeApprovedBlock <- kvm.store("blocks-approved")
-      blockStore         <- KeyValueBlockStore(store, storeApprovedBlock)
+      blockStore         <- KeyValueBlockStore(store, storeApprovedBlock, hashResolver)
```
tgrospic: Changes in this file are not used now, right?

tgrospic:
Please cleanup your commits because we are merging PR without squashing.

There are some useful instructions how to use `git rebase -i HEAD~n` to rewrite Git history.
https://www.google.com/search?q=github+cleanup+PR+commits

tgrospic:
bors try

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -631,15 +631,23 @@ object BlockAPI {
...
+        implicit casper =>
+          for {
+            dag          <- casper.blockDag
+            blockHashOpt <- dag.find(hash)
+            message <- blockHashOpt match {
+                        case Some(blockHash) => BlockStore[F].get(blockHash)
+                        case None            => none[BlockMessage].pure[F]
+                      }
+          } yield message,
+        none[BlockMessage].pure[F]
```
tgrospic, line 645: Let's better throw exception when Casper is not ready because block can be in the DAG but node is just starting up.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -631,15 +631,23 @@ object BlockAPI {
...
-    }
+    EngineCell[F].read >>= (
+      _.withCasper[Option[BlockMessage]](
+        implicit casper =>
+          for {
+            dag          <- casper.blockDag
+            blockHashOpt <- dag.find(hash)
+            message <- blockHashOpt match {
+                        case Some(blockHash) => BlockStore[F].get(blockHash)
+                        case None            => none[BlockMessage].pure[F]
```
tgrospic: More functional approach is to not destructure _manually_ but use functions, in this case `flatTraverse` will behave like `map` in `F` context together with flattening Option.

block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala
```diff
@@ -39,6 +44,7 @@ class KeyValueBlockStore[F[_]: Sync](
 
   override def find(p: BlockHash => Boolean, n: Int): F[Seq[(BlockHash, BlockMessage)]] = {
```
tgrospic: Please remove `find` method from `BlockStore` completely.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -75,6 +75,9 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
     def isFinalized(blockHash: BlockHash): F[Boolean] =
       finalizedBlocksSet.contains(blockHash).pure[F]
 
+    override def find(truncatedHash: String): F[Option[BlockHash]] =
+      dagSet.find(_.toStringUtf8.startsWith(truncatedHash)).pure[F]
```
tgrospic: You need to convert hash `ByteString` to hex format because this is the format of input prefix string. We have `Base16.encode` function for that.

Also, because this function is in `F` context, using `pure` means that the whole find operation will be executed immediately when F expression is constructed and not suspended.
The rule is when you don't have just the value `Sync[F].delay` must be used instead so that operation (function) is suspended until it will be executed in F context.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: Please ignore this one, it will just complicates the code. We need to redesign the API by throwing errors for unexpected cases which should be handled on the protocol level (transformed to JSON, gRPC errors, ...) so that Scala code can be clean and contain only domain logic.

tgrospic:
bors try

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
```diff
@@ -75,6 +76,10 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
     def isFinalized(blockHash: BlockHash): F[Boolean] =
       finalizedBlocksSet.contains(blockHash).pure[F]
 
+    override def find(truncatedHash: String): F[Option[BlockHash]] = Sync[F].delay {
+      dagSet.find(hash => Base16.encode(hash.toByteArray).startsWith(truncatedHash))
```
nzpr: I'm curious, would something like this work / improve performance?
```scala
 override def find(truncatedHash: String): F[Option[BlockHash]] = Sync[F].delay {
      dagSet.find { hash =>
        // Use iterators to make it as lazy as possible, and transform only short string, not the whole hash
        val searchPrefix = Base16.unsafeDecode(truncatedHash).toIterator
        // This fails fast on the first array element mismatch
        !(hash.toByteArray.toIterator zip searchPrefix).exists { case (l, r) => l != r }
      }
    }
```

nzpr (approved):
Congratulations on the first PR! 
I left a minor performance related comment, as a nice addition if it makes sense. But ok to merge as it is.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
tgrospic: @nzpr If you suspecting that bottleneck is in byte comparison but I think this is already optimized in Java `String` class in `startWith` method.

I'm more suspicious in our custom implementation of encoding bytes to hex. This is used in `Base16.encode`.
https://github.com/rchain/rchain/blob/57eb112cf7275b31316b4692d2faf0f62fab0dad/crypto/src/main/scala/coop/rchain/crypto/codec/Base16.scala#L32-L33

nzpr: I see, anyway as it was shown by @stanislavlyalin on standup, there is `startsWith` on ByteString, so the fastest way is to convert `truncatedHash` into ByteString

tgrospic:
bors try

tgrospic (approved):
LGTM

Requests for unknown block hash on test net shows ~20 times improvement what is expected because iteration through all blocks hashes is done in memory.
https://de1a.dev.rchain.coop/api/block/1234abcd 80ms
https://observer.testnet.rchain.coop/api/block/1234abcd 1.7sec

@stanislavlyalin Excellent work for your first PR! :partying_face:

Please merge your PR now with 2 approvals.

stanislavlyalin:
bors merge

PR #3523, opened 2021-10-27
WIP

nzpr:


v2/src/main/scala/coop/rchain/v2/casper/data/CasperScope.scala
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @param latestMessages      Upper boundary, latest messages known. Define the whole [[CasperScope]].
+ * @param finalizationFringe  Bottom boundary, set of finalized messages for all known senders.
+ * @param conflictScope       The body of the scope - all messages above [[finalizationFringe]].
+ */
+final case class CasperScope[M, S](
+    latestMessages: LatestMessages[M, S],
+    finalizationFringe: FinalizationFringe[M, S],
+    conflictScope: ConflictScope[M]
+) { require(finalizationFringe.v.keySet == latestMessages.v.keySet) }
```
zsluedem, line 17: `Casper scope is valid only if finalization fringe is complete: finalized messages are found for all known senders.`

What does `valid` mean in the above sentence?

Can I say "For a particular block, there is always a fixed value `CasperScope` corresponding to this particular block" is a correct property for a block?

v2/src/main/scala/coop/rchain/v2/casper/data/FinalizationFringe.scala
```diff
@@ -0,0 +1,17 @@
...
+ */
+final case class FinalizationFringe[M, S](v: Map[S, M])
+
+object FinalizationFringe {
+
+  /**
+   * Find messages that are finalized in next compared to prev.
+   */
+  def diff[M, S](next: FinalizationFringe[M, S], prev: FinalizationFringe[M, S]): Set[M] = ???
+}
```
zsluedem, line 17: Sorry that I am still kind of unkown on the term `FinalizedFringe`. Is it possible to explain it in an example?

v2/src/main/scala/coop/rchain/v2/casper/Casper.scala
```diff
@@ -0,0 +1,198 @@
...
+ * @param dg            Dependency graph exposing justification structure.
+ * @param so            Safety oracle, defining fault tolerances.
+ * @param maxComplexity Maximum number of traversals through dependency graph.
+ * @tparam M            Type of a message.
+ * @tparam S            Type of a sender.
+ */
+final case class Casper[M, S](
+    dg: DependencyGraph[M, S],
+    so: SafetyOracle[M, S],
+    maxComplexity: Long = Long.MaxValue
```
zsluedem, line 19: It is kind of wired to have a limit here.

v2/src/main/scala/coop/rchain/v2/casper/data/CasperScope.scala
nzpr: Yes, you are right, for each block there is Casper scope associated. When you replay - you reconstruct the scope. To be able to create a block, proposer first have to find the scope, as well.

`Valid` means that if there is no finalized message for all senders in a fringe - Casper scope cannot be created. So if you will try to create such an instance - you will hit an error. I need to add here error msg actually.

v2/src/main/scala/coop/rchain/v2/casper/data/FinalizationFringe.scala
nzpr: `Finalization fringe` is a set of messages: highest finalized message for each known sender. Known senders are defined by latest messages, which you use to run Casper.

Screenshot 2021-11-02 at 12 33 08

Here you can see grey blocks: they are finalized. Collectively they are finalization fringe. Everything below them is finalized (conflict resolution has been done and acceptances and rejects are final).
 
(Also you can see there on extra grey block at the bottom - this is highest common message for messages of the fringe, it is required to merge fringe and get finalized state).

v2/src/main/scala/coop/rchain/v2/casper/Casper.scala
nzpr: I'm trying to push a narrative that Casper is an algorithm, and we can limit its complexity if we want. So we cannot end up with Casper running for hours without knowing about that. We put a constraint on amount of resources that are allocated for Casper to search for a scope. If DAG is very diverged, Casper might not find a scope. Then we might increase complexity and try again, having the same DAG. Or just fail and throw some error.

nzpr:
Closing as outdated

PR #3526, opened 2021-11-08
Hotstore rewrite (tuple space hash mismatch fix)

### LINKED ISSUE #3200: Problem with use of peek in NonNegativeNumber.rho
https://github.com/rchain/rchain/blob/0fb0c616dadb553b7b1bea09e448e7a0e1ff9765/casper/src/main/resources/NonNegativeNumber.rho#L48-L50

zsluedem:
## Overview

rebase the work in https://github.com/rchain/rchain/pull/3270 to the latest dev.

The original pr is intended to fix hotstore error which would cause tuple space mismatch

- [x] create a companion object of `HotStoreState`

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -31,13 +31,11 @@ trait HotStore[F[_], C, P, A, K] {
 }
 
 final case class HotStoreState[C, P, A, K](
-    continuations: Map[Seq[C], Seq[WaitingContinuation[P, K]]] =
-      Map.empty[Seq[C], Seq[WaitingContinuation[P, K]]],
-    installedContinuations: Map[Seq[C], WaitingContinuation[P, K]] =
-      Map.empty[Seq[C], WaitingContinuation[P, K]],
-    data: Map[C, Seq[Datum[A]]] = Map.empty[C, Seq[Datum[A]]],
-    joins: Map[C, Seq[Seq[C]]] = Map.empty[C, Seq[Seq[C]]],
-    installedJoins: Map[C, Seq[Seq[C]]] = Map.empty[C, Seq[Seq[C]]]
+    continuations: Map[Seq[C], Seq[WaitingContinuation[P, K]]],
+    installedContinuations: Map[Seq[C], WaitingContinuation[P, K]],
+    data: Map[C, Seq[Datum[A]]],
+    joins: Map[C, Seq[Seq[C]]],
+    installedJoins: Map[C, Seq[Seq[C]]]
 ) {
   def snapshot(): Snapshot[C, P, A, K] =
     Snapshot(
@@ -48,6 +46,17 @@ final case class HotStoreState[C, P, A, K](
         joins = this.joins,
         installedJoins = this.installedJoins
       )
+    )
+}
+
+object HotStoreState {
+  def apply[C, P, A, K](): HotStoreState[C, P, A, K] =
+    HotStoreState[C, P, A, K](
+      Map.empty[Seq[C], Seq[WaitingContinuation[P, K]]],
+      Map.empty[Seq[C], WaitingContinuation[P, K]],
+      Map.empty[C, Seq[Datum[A]]],
+      Map.empty[C, Seq[Seq[C]]],
+      Map.empty[C, Seq[Seq[C]]]
     )
 }
 
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -287,7 +287,7 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
                   (state, state.installedContinuations.get(join) ++: continuations)
               })._2
 
-              val index       = state.joins.getOrElse(channel, Seq.empty).indexOf(join)
+              val index       = curJoins.indexOf(join)
               val outOfBounds = !curJoins.isDefinedAt(index)
 
               // TODO should attimpting to remove join with non empty contnuation lead to error as well?
@@ -297,7 +297,7 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
                 if (outOfBounds)
                   (state.copy(joins = state.joins.updated(channel, curJoins)), true)
                 else {
-                  val newVal = removeIndex(state.joins.getOrElse(channel, Seq.empty), index)
+                  val newVal = removeIndex(curJoins, index)
                   (state.copy(joins = state.joins.updated(channel, newVal)), false)
                 }
               } else (state.copy(joins = state.joins.updated(channel, curJoins)), false)
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -7,7 +7,6 @@ import coop.rchain.rspace.history.HistoryReaderBase
 import coop.rchain.shared.MapOps._
 import coop.rchain.rspace.internal._
 
-final case class Snapshot[C, P, A, K](private[rspace] val cache: HotStoreState[C, P, A, K])
 import scala.collection.immutable.Map
 
 trait HotStore[F[_], C, P, A, K] {
@@ -27,7 +26,7 @@ trait HotStore[F[_], C, P, A, K] {
 
   def changes(): F[Seq[HotStoreAction]]
   def toMap: F[Map[Seq[C], Row[P, A, K]]]
-  def snapshot(): F[Snapshot[C, P, A, K]]
+  def snapshot(): F[HotStoreState[C, P, A, K]]
 }
 
 final case class HotStoreState[C, P, A, K](
@@ -36,18 +35,7 @@ final case class HotStoreState[C, P, A, K](
     data: Map[C, Seq[Datum[A]]],
     joins: Map[C, Seq[Seq[C]]],
     installedJoins: Map[C, Seq[Seq[C]]]
-) {
-  def snapshot(): Snapshot[C, P, A, K] =
-    Snapshot(
-      this.copy(
-        continuations = this.continuations,
-        installedContinuations = this.installedContinuations,
-        data = this.data,
-        joins = this.joins,
-        installedJoins = this.installedJoins
-      )
-    )
-}
+)
 
 object HotStoreState {
   def apply[C, P, A, K](): HotStoreState[C, P, A, K] =
@@ -73,7 +61,7 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
     historyReaderBase: HistoryReaderBase[F, C, P, A, K]
 ) extends HotStore[F, C, P, A, K] {
 
-  def snapshot(): F[Snapshot[C, P, A, K]] = hotStoreState.get.map(_.snapshot())
+  def snapshot(): F[HotStoreState[C, P, A, K]] = hotStoreState.get
 
   // Continuations
 
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -1,15 +1,11 @@
 package coop.rchain.rspace
 
-import cats._
-import cats.implicits._
+import cats.syntax.all._
 import cats.effect._
 import cats.effect.concurrent.{Deferred, Ref}
 import coop.rchain.rspace.history.HistoryReaderBase
-import coop.rchain.shared.Language._
 import coop.rchain.shared.MapOps._
 import coop.rchain.rspace.internal._
-import coop.rchain.shared.Serialize._
-import scodec.Codec
 
 final case class Snapshot[C, P, A, K](private[rspace] val cache: HotStoreState[C, P, A, K])
 import scala.collection.immutable.Map
--- a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
@@ -1,16 +1,18 @@
 package coop.rchain.rspace
 
-import cats.effect.Concurrent
-import cats.effect.concurrent.Ref
-import cats.syntax.all._
+import cats._
+import cats.implicits._
+import cats.effect._
+import cats.effect.concurrent.{Deferred, Ref}
 import coop.rchain.rspace.history.HistoryReaderBase
-import coop.rchain.rspace.internal._
 import coop.rchain.shared.Language._
 import coop.rchain.shared.MapOps._
+import coop.rchain.rspace.internal._
+import coop.rchain.shared.Serialize._
+import scodec.Codec
 
-import scala.collection.concurrent.TrieMap
-
-final case class Snapshot[C, P, A, K](private[rspace] val cache: Cache[C, P, A, K])
+final case class Snapshot[C, P, A, K](private[rspace] val cache: HotStoreState[C, P, A, K])
+import scala.collection.immutable.Map
 
 trait HotStore[F[_], C, P, A, K] {
   def getContinuations(channels: Seq[C]): F[Seq[WaitingContinuation[P, K]]]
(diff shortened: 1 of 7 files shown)


rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+                }
+              } else (state.copy(joins = state.joins.updated(channel, curJoins)), false)
+            })
+//      _ <- Sync[F]
+//            .raiseError(
+//              new IndexOutOfBoundsException(
+//                s"Index out of bounds when removing join"
+//              )
+//            )
+//            .whenA(err)
```
zsluedem: The old implementation ignore error like this. Not sure if we should raise error for this. It looks we'd better raise it while some tests are failing because the original tests are too sticky on implementations.

nzpr: I think this error was there before in `checkIndex`? It's just checking of index has been done in the middle two state modifications that are supposed to be one atomic thing. So index check is now performed inside pure computation and error is raised outside.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
-final case class Cache[C, P, A, K](
-    continuations: TrieMap[Seq[C], Seq[WaitingContinuation[P, K]]] =
-      TrieMap.empty[Seq[C], Seq[WaitingContinuation[P, K]]],
-    installedContinuations: TrieMap[Seq[C], WaitingContinuation[P, K]] =
-      TrieMap.empty[Seq[C], WaitingContinuation[P, K]],
-    data: TrieMap[C, Seq[Datum[A]]] = TrieMap.empty[C, Seq[Datum[A]]],
-    joins: TrieMap[C, Seq[Seq[C]]] = TrieMap.empty[C, Seq[Seq[C]]],
-    installedJoins: TrieMap[C, Seq[Seq[C]]] = TrieMap.empty[C, Seq[Seq[C]]]
+final case class HotStoreState[C, P, A, K](
+    continuations: Map[Seq[C], Seq[WaitingContinuation[P, K]]] =
```
hilltracer: Can you please say to me: what the difference is Map with TrieMap and which benefit of this change?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+private final case class HistoryStoreCache[F[_], C, P, A, K](
+    continuations: Map[Seq[C], Deferred[F, Seq[WaitingContinuation[P, K]]]],
+    datums: Map[C, Deferred[F, Seq[Datum[A]]]],
+    joins: Map[C, Deferred[F, Seq[Seq[C]]]]
+)
+
 private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
-    cacheRef: Ref[F, Cache[C, P, A, K]],
-    historyReader: HistoryReaderBase[F, C, P, A, K]
+    hotStoreState: Ref[F, HotStoreState[C, P, A, K]],
```
hilltracer, line 58: Can you please say to me: what is the point of Ref?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: Good question!

Difference:

1. TrieMap is a trie-based **concurrent** scalable map implementation. which is like the `history trie` you are trying to change.It provides an **efficient, non-blocking, O(1) time snapshot operation** while the performance is a little bit slower than `Map` which is just a java `ConcurrentHashMap`.(Getit  from some reading)
2. `TrieMap` is a **mutable** data structure while `Map` here is an **immutable** data structure. Functional programming languages favor immutable data structures which can provide safe concurrent operations.

> which benefit of this change?

There is a bug on hotstore which the data is not consistent after the operation. However, I can not really know the deep detail of the bug which you have to replay thoudsands of time to reproduce it. @nzpr got a pr the change the data into immutable ones and I just rebase the work from him.

@tgrospic @nzpr correct me if I am wrong.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: `Ref` is An asynchronous, concurrent mutable reference. There is an easy example here https://typelevel.org/cats-effect/docs/std/ref.

You can wrap any data structure into a `Ref`and use the operation of `Ref` like `get`, `set` , `getAndUpdate` and etc. to make sure the data structure mutation is totally safe and consistent. The [example](https://typelevel.org/cats-effect/docs/std/ref#concurrent-counter) above is a very good example says how to mutate a data wrapped in a `Ref` in multi threads.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: Not sure what you mean here. But I made some changes to include it.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
hilltracer: Thanks, for the answers. I have the same structure (TrieMap) for hashing in HistoryStore. And in my code, I don't use the snapshot function. I don't know, what better use in my case. My priority is speed. @tgrospic  @nzpr @zsluedem what do you think?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
hilltracer: Thanks, I try to understand)

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: I am still in a dilemma to do that.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: Oh. In my opinion, You don't have to compare your implementation with scala TrieMap. They are for different use cases. 
1. Your implementation should be able to used either in-memory storage or disk storage while scala TrieMap is only memory. 
2. Your implementation should be able to keep the history track of storage (by stateHash) while scala TrieMap don't do that(But I think you can consider snapshot as somthing similar.)

nzpr: We want to allow HotStore state to change only atomically. TrieMap allows concurrent access which has been used previously. New implementation puts the whole `HotStoreState ` inside a `Ref`, which manages concurrent access. In general, we don't want to have TrieMap-s, because it allows concurrency that is detached from our scheduler and managed by JVM. We want to use `Concurrent` typeclass for our F[_] (which is monix.Task) in places that require concurrency.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
nzpr: Its related to https://github.com/rchain/rchain/pull/3526#discussion_r752948893, you can check https://typelevel.org/cats-effect/docs/std/ref also.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -1,16 +1,18 @@
 package coop.rchain.rspace
 
-import cats.effect.Concurrent
-import cats.effect.concurrent.Ref
-import cats.syntax.all._
+import cats._
+import cats.implicits._
```
tgrospic: Please import `cats.syntax` instead. Do we need `cats._`?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: @hilltracer TrieMap cannot be just replaced with Map. Update/insert operation on a Map will return a new _immutable_ instance . TrieMap can be mutated by changing single reference.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: @hilltracer Ref gives the capability to update the value it holds in an atomic way. It's important that this is done without locking, update/modify operations **can** be executed multiple times if concurrently value has changed between the entry and exit of update/modify functions.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
-        continuations = this.continuations.snapshot(),
-        installedContinuations = this.installedContinuations.snapshot(),
-        data = this.data.snapshot(),
-        joins = this.joins.snapshot(),
-        installedJoins = this.installedJoins.snapshot()
+        continuations = this.continuations,
+        installedContinuations = this.installedContinuations,
+        data = this.data,
+        joins = this.joins,
+        installedJoins = this.installedJoins
```
tgrospic: This is now immutable object so there is no need to make a copy. Maybe we can even remove Snapshot class as well, if not to invasive change.

Also please create companion object for HotStoreState to construct object with default empty maps.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+)
+
 private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
-    cacheRef: Ref[F, Cache[C, P, A, K]],
-    historyReader: HistoryReaderBase[F, C, P, A, K]
+    hotStoreState: Ref[F, HotStoreState[C, P, A, K]],
+    // this is what is inside history store, lazily populated. Starting data for HotStoreState
+    historyStoreCache: Ref[F, HistoryStoreCache[F, C, P, A, K]]
+)(
+    implicit HR: HistoryReaderBase[F, C, P, A, K]
```
tgrospic: There is not need for history reader to be implicit argument (and be specially named uppercase).

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+                    joinsInHistoryStore
                   )
-                ignore(
-                  cache.joins.put(
-                    channel,
-                    updated
+                // just return what is in hot store already
+                case Some(joins) =>
+                  (state, joins)
+              })._2
```
tgrospic: State returned is not used, it's updated at the end of _modify_ function. This is just:
`state.joins.getOrElse(channel, joinsInHistoryStore)`

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+                    ),
+                    state.installedContinuations.get(join) ++: contsInHistoryStore
                   )
-                )
-                cache
-              } else ().pure[F]
+                // just return what is in hot store already
+                case Some(continuations) =>
+                  (state, state.installedContinuations.get(join) ++: continuations)
+              })._2
```
tgrospic: State returned is not used, it's updated at the end of _modify_ function.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -32,261 +34,388 @@ trait HotStore[F[_], C, P, A, K] {
...
+
+              val index       = curJoins.indexOf(join)
+              val outOfBounds = !curJoins.isDefinedAt(index)
+
+              // TODO should attimpting to remove join with non empty contnuation lead to error as well?
```
tgrospic: _Remove join_ is called when continuation is removed, so it can be called when continuations are present in which case we just want to skip removal.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -29,264 +26,356 @@ trait HotStore[F[_], C, P, A, K] {
...
-                continuations,
-                index
+      fromHistoryStore <- getContFromHistoryStore(channels)
+      r <- hotStoreState.modify[(Boolean, Boolean)](state => {
+            val curVal       = state.continuations.getOrElse(channels, fromHistoryStore)
+            val installedCon = state.installedContinuations.get(channels)
+            val isInstalled  = installedCon.nonEmpty
+
+            val removingInstalled = (isInstalled && index == 0)
+            val removedIndex      = if (isInstalled) index - 1 else index
```
tgrospic: What is the logic here to use index -1 if _installed_ continuation is trying to be removed?

Also it seems strange to check installed only if index is zero. I don't see when should be allowed to remove installed continuations at all.

@nzpr Is this part of your previous code?

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
```diff
@@ -39,35 +41,45 @@ trait HotStoreSpec[F[_], M[_]] extends FlatSpec with Matchers with GeneratorDriv
...
+                                   .arbitrary[
+                                     TrieMap[
+                                       Seq[String],
+                                       WaitingContinuation[Pattern, StringsCaptor]
+                                     ]
+                                   ]
+                                   .map(_.toMap)
+        data           <- Arbitrary.arbitrary[TrieMap[String, Seq[Datum[String]]]].map(_.toMap)
+        joins          <- Arbitrary.arbitrary[TrieMap[String, Seq[Seq[String]]]].map(_.toMap)
+        installedJoins <- Arbitrary.arbitrary[TrieMap[String, Seq[Seq[String]]]].map(_.toMap)
```
tgrospic: `immutable.Map` can be used here instead of TrieMap.

zsluedem: The whole method is not used actually . I just removed it .

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: > What is the logic here to use index -1 if installed continuation is trying to be removed?

I think this `-1` is added by me. Let me check why is that.

zsluedem: If the channels in Seq[C] contain `installedContinuation`,  the `hotStoreState.getContinuations` would look like `Vector(InstallContinuation ,  normalContinuation)`.

When you remove `continuation` with the `index` provided by `hotStoreState.getContinuations`, the first one is `installed` while `hotStoreState.continuations` doesn't contain the `installedContinuation`. That's why we need to remove 1 in index.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -29,264 +26,363 @@ trait HotStore[F[_], C, P, A, K] {
...
+            val isInstalled  = installedCon.nonEmpty
+
+            val removingInstalled = (isInstalled && index == 0)
+            // why -1 here
+            // from users who use `HotStore` trait perspective
+            // the users need to call getContinuations first to retrieve the continuation of the channels
+            // then find which index of the continuation users want to remove
+            // while getContinuations would return results like `Vector(installedContinuation, normalcontinuations, ...)`
+            // The first one would be always `installedContinuation` if it exist which need the
+            // index to -1 to exclude installedContinuation
```
nzpr, line 118: But index supplied here already accounts the presence of installed continuations. When you match produce over Seq of continuations, you will see the installed one, as you aways get it through `getContinuations`. I don't think you can modify index here.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
nzpr: Yes, this is a change compared to my code and old implementation as well.
In previous code index to remove is supplied as input arg is unchanged in the method.

```scala
def removeContinuation(channels: Seq[C], index: Int): F[Unit] =
    for {
      continuations <- getContinuations(channels)
      cache         <- cacheRef.get
      installed     = cache.installedContinuations.get(channels)
      _ <- if (installed.isDefined && index == 0) {
            new IllegalArgumentException("Attempted to remove an installed continuation").raiseError
          } else
            checkIndex(continuations, index) >> cacheRef.update { cache =>
              val updated = removeIndex(
                continuations,
                index
              )
              ignore(installed match {
                case Some(_) => cache.continuations.put(channels, updated tail)
                case None    => cache.continuations.put(channels, updated)
              })
              cache
            }
    } yield ()
```
@zsluedem I think your change masks the situation when previously error has been thrown. Now it just does nothing, as removing of index -1 does nothing. I've added one more comment https://github.com/rchain/rchain/pull/3526#discussion_r760451380

zsluedem: @nzpr In previous code，
```scala
              val updated = removeIndex(
                continuations,
                index
              )
```

The `continuations` above include `installedContinuation` and `normalContinuation` while in the new codes without my fix(index - 1) below.

```scala
  def removeContinuation(channels: Seq[C], index: Int): F[Unit] =
    for {
      fromHistoryStore <- getContFromHistoryStore(channels)
      r <- hotStoreState.modify[(Boolean, Boolean)](state => {
            val curVal      = state.continuations.getOrElse(channels, fromHistoryStore)
            val isInstalled = state.installedContinuations.contains(channels)

            val removingInstalled = (isInstalled && index == 0)
            val outOfBounds       = !curVal.isDefinedAt(index)

            if (removingInstalled || outOfBounds)
              (state, (removingInstalled, outOfBounds))
            else {
              val newVal = removeIndex(curVal, index)
              // TODO Logic behind this isInstalled check is uncertain
              if (isInstalled)
                (
                  state.copy(continuations = state.continuations.updated(channels, newVal.tail)),
                  (false, false)
                )
              else
                (
                  state.copy(continuations = state.continuations.updated(channels, newVal)),
                  (false, false)
                )
            }

          })
      (removingInstalled, invalidIndex) = r
      _ <- Sync[F]
            .raiseError {
              new IllegalArgumentException("Attempted to remove an installed continuation")
            }
            .whenA(removingInstalled)
      _ <- Sync[F]
            .raiseError(
              new IndexOutOfBoundsException(
                s"Index ${index} out of bounds when removing continuation"
              )
            )
            .whenA(invalidIndex)
    } yield ()
```

the `val newVal = removeIndex(curVal, index)` which `curVal` is just `continuations` without `installedContinuations`.

nzpr: I see, thanks, so old hotstore has installedContinuations in `curVal` and my code does not. Ok, thanks!

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -29,264 +26,363 @@ trait HotStore[F[_], C, P, A, K] {
...
-                    updated
-                  )
-                )
-                cache
-              } else ().pure[F]
+      joinsInHistoryStore <- getJoinsFromHistoryStore(channel)
+      contsInHistoryStore <- getContFromHistoryStore(join)
+      err <- hotStoreState.modify[Boolean](state => {
+
+              val curJoins = state.joins.getOrElse(channel, joinsInHistoryStore)
```
nzpr, line 261: @zsluedem I'm thinking - is here with joins situation the same as with continuations? `curJoin` here does not account installed.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: Previously continuations are retrieved with `getContinuations` which included installed, but now they are accessed directly from the state without installed.

I'm OK with this solution for now (it's works the same as previous) until we came up with more elegant one.

nzpr: Yes I agree https://github.com/rchain/rchain/pull/3526#discussion_r761162294

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
zsluedem: `removeJoin` got very specific target on removing

```
def removeJoin(channel: C, join: Seq[C])`
```

Not like continuations want `index`

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

tgrospic:
bors merge

PR #3528, opened 2021-11-10
[PoC] Dag manager and related interfaces (SDK)

tgrospic:
## Overview

This PR if proof of concept for `DagManager` as an abstraction to manage node state (blocks, blocks metadata, indexes).

### DIFF
--- a/node/src/main/scala/coop/rchain/node/dag/implementation/RNodeDagManager.scala
+++ b/node/src/main/scala/coop/rchain/node/dag/implementation/RNodeDagManager.scala
@@ -1,0 +1,50 @@
+package coop.rchain.node.dag.implementation
+
+import cats.effect.concurrent.Ref
+import cats.effect.{Concurrent, Sync}
+import coop.rchain.sdk.dag.data.{DagChecker, DagManager, DagView}
+import fs2.Stream
+
+object RNodeDagManager {
+  def apply[F[_]: Concurrent, M, MId, S](
+      st: Ref[F, Map[MId, M]],
+      requestMsg: MId => F[Unit],
+      msgInput: Stream[F, M],
+      dagChecker: DagChecker[F, M, MId]
+  ): F[RNodeDagManager[F, M, MId, S]] =
+    Sync[F].delay(new RNodeDagManager(st, requestMsg, msgInput, dagChecker))
+}
+
+final case class RNodeDagManager[F[_]: Concurrent, M, MId, S] private (
+    /**
+      * DagManager internal in-memory state.
+      */
+    st: Ref[F, Map[MId, M]],
+    /**
+      * Request external module for required messages.
+      */
+    requestMsg: MId => F[Unit],
+    /**
+      * Receiving stream of messages.
+      */
+    messagesInput: Stream[F, M],
+    /**
+      * All necessary validation to maintain the DAG.
+      *
+      * This interface should bring the Casper logic into DAG manager.
+      */
+    dagChecker: DagChecker[F, M, MId]
+) extends DagManager[F, M, S] {
+
+  /**
+    * Returns DAG read-only representation starting with specified latest messages.
+    *
+    * NOTE: Getting view for invalid latest messages should throw error.
+    */
+  override def getDagView(latestMessages: Set[M]): F[DagView[F, M, S]] = ???
+
+  /**
+    * Thread safe function to insert new node to the DAG.
+    */
+  override def updateDag(m: M): F[Unit] = ???
+}
--- a/shared/src/main/scala/coop/rchain/sdk/casper/syntax/CasperDagDataSyntax.scala
+++ b/shared/src/main/scala/coop/rchain/sdk/casper/syntax/CasperDagDataSyntax.scala
@@ -1,0 +1,76 @@
+package coop.rchain.sdk.casper.syntax
+
+import coop.rchain.sdk.dag.data.DagData
+import coop.rchain.sdk.dag.syntax._
+
+trait CasperDagDataSyntax {
+  implicit def sdkCasperSyntaxDagDataMessage[F[_], M, S](m: M): CasperDagDataMessageOps[M, S] =
+    new CasperDagDataMessageOps(m)
+}
+
+sealed trait MalformedMessageReason
+
+/**
+  * Extensions of basic DAG types related to Casper.
+  *
+  * @param m message instance
+  * @tparam M represents a message
+  * @tparam S represents a message sender
+  */
+final class CasperDagDataMessageOps[M, S](private val m: M) extends AnyVal {
+
+  /**
+    * Message should be well formed.
+    *
+    * If sender creates and signs message that contains junk data, there are two options: just ignore, or act.
+    * Just ignoring introduces DoS attack vector when lots of blocks have to be verified, because the whole
+    * message verification takes more resources then just verifying the signature.
+    * Therefore ill formed message is an offence.
+    *
+    * TODO: If this function needs other data then M it should be moved to [[DagViewSyntax]] syntax.
+    */
+  def invalidMalformedMessage(checkFormat: M => MalformedMessageReason): MalformedMessageReason =
+    checkFormat(m)
+
+  /**
+    * Message sender should be present and have non zero stake in map.
+    */
+  def inactiveSender(implicit dagData: DagData[M, S]): Boolean =
+    !m.bondsMap.exists { case (s, stake) => m.sender == s && stake > 0 }
+
+  /**
+    * Message should have justification for each senders in the bonds map (both active and inactive).
+    */
+  def invalidJustificationFollows(bondedSenders: Set[S])(implicit dagData: DagData[M, S]): Boolean =
+    m.justifications.map(_.sender) != bondedSenders
+
+  /**
+    * Message should have sequence number equal to sequence number of self justification + 1.
+    * [[incompleteJustificationSet]] guarantees that self justification s present.
+    */
+  def invalidSequenceNumber(selfJustification: M)(implicit dagData: DagData[M, S]): Boolean =
+    m.seqNum != selfJustification.seqNum + 1
+
+  /**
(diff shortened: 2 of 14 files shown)


shared/src/main/scala/coop/rchain/sdk/casper/syntax/CasperDagDataSyntax.scala
```diff
@@ -0,0 +1,76 @@
...
+
+import coop.rchain.sdk.dag.data.DagData
+import coop.rchain.sdk.dag.syntax._
+
+trait CasperDagDataSyntax {
+  implicit def sdkCasperSyntaxDagDataMessage[F[_], M, S](m: M): CasperDagDataMessageOps[M, S] =
+    new CasperDagDataMessageOps(m)
+}
+
+sealed trait MalformedMessageReason
```
zsluedem, line 11: It seems that `sealed` can only be expanded in the same source file. Not sure if you want to expand this in another file

shared/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
```diff
@@ -0,0 +1,21 @@
+package coop.rchain.sdk.dag.data
+
+trait DagData[M, S] {
```
zsluedem, line 3: I was trying to implement things like this several months before and I think `trait` like this is kind of over-abstract from my perspective. The abstraction gives very high flexibility indeed while the implementation of the trait seems more codes and long.

I prefer something like 
```scala
trait DagData {
  def seqNum: Long
  def blockNum: Long
  def justifications: Set[M]
  def sender[S]: S # or put S in DagData[S], not sure which is better
  def bondsMap: Map[S, Long]
  def stake[S](s: S): Long
}
```
and when we need type works like `DagData`, we could use like below
```scala
trait DagManager[F[_], M <: DagData, S] {
  def getDagView(latestMessages: Set[M]): F[DagView[F, M, S]]
  def updateDag(m: M): F[Unit]
}
```

In this way, we can build things with some relation to each other but these constraints might be troubles in some way.

shared/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
```diff
@@ -0,0 +1,21 @@
...
+  /* Message */
+
+  def seqNum(m: M): Long
+
+  def blockNum(m: M): Long
+
+  // TODO: this is not possible because message has IDs
+  def justifications(m: M): Set[M]
+
+  // TODO: this is not possible because message has IDs
```
zsluedem, line 13: What do you mean `not possible` here?

node/src/main/scala/coop/rchain/node/dag/implementation/RNodeDagManager.scala
```diff
@@ -0,0 +1,50 @@
...
+
+final case class RNodeDagManager[F[_]: Concurrent, M, MId, S] private (
+    /**
+      * DagManager internal in-memory state.
+      */
+    st: Ref[F, Map[MId, M]],
+    /**
+      * Request external module for required messages.
+      */
+    requestMsg: MId => F[Unit],
```
zsluedem, line 26: I am kind of headache and horror reading a class with higher function after reading all the `BlockProcessor` and `ConflictSetMerger`. Every time I come back to codes after some time, I have to start from scratch.
Maybe `DagManager`doesn't care about how the message is requested. We can have a `MessageRequester` and we can implement a interactor between `DagManager`and `MessageRequester`?

zsluedem: You already got `NetworkBlockRequester` which is good.

Maybe all we need is an **interactor** like `DagData` which is working on `M` and `S` and the interactor would be working on `DagManager` and `NetworkMessageRequester`.

zsluedem (approved):
Very excellent implementations and demonstrations of syntax. I can not wait to work on it.
I only have some minor problems with it. But the general direction and goal are clear and nice.

shared/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
zsluedem: Actually, after going over all the codes, I think this idea can hold back and try with what you have.

tgrospic: It will be much easier to do this with Scala 3 with less boilerplate code.

Inheritance has two major problems:
1) requires modification to types to extend the interface
2) return values can only have the type that is inherited which prevents composition.

But I agree, abstraction like this should only be used when necessary and not everywhere. Here is just an example to show how we can implement this pattern and what is the boilerplate code.

node/src/main/scala/coop/rchain/node/dag/implementation/RNodeDagManager.scala
tgrospic: Input function `MId => F[Unit]` has purpose to show that there can be interaction with other components. In many cases components has circular dependency and there is not way to create instance of one without the other.

E.g. `MessageRequester` can depend on the result of `DagManager` validation its state to request new messages and `DagManager` in the same time depends on `MessageRequester` to acquire missing messages.

In simplified view something like this.
```scala
// Mutually dependent components cannot be combined
val msgReq = new MessageRequester(dagMngr, ...)
val dagMngr = new DagManager(msgReq, ...)
```
So we need a way to create both and connect them.
```scala
// Creating components
val msgReq = new MessageRequester(...)
val dagMngr = new DagManager(...)

// Combining components
msgReq.validationHandler(dagMngr.validate)
dagMngr.requestHandler(msgReq.requestMsg)
```
So what I want to show here is that we need to think about composition/interaction of components as a way to build complexity and not how they are directly dependent.

shared/src/main/scala/coop/rchain/sdk/casper/syntax/CasperDagDataSyntax.scala
tgrospic: I was just using part of Anton's code in this example and it should be removed. Managing exceptions should be done separately.

shared/src/main/scala/coop/rchain/sdk/casper/syntax/CasperDagDataSyntax.scala
```diff
@@ -0,0 +1,76 @@
...
+    assert(selfJOpt.isDefined, "Message does not have self justification.")
+    val selfJ   = selfJOpt.get
+    val selfJjs = selfJ.justifications
+    js.toIterator.map { j =>
+      selfJjs.exists { selfJj =>
+        // Previous justification from the same sender has seq num higher then the current one.
+        selfJj.sender == j.sender && selfJj.seqNum > j.seqNum
+      }
+    }.nonEmpty
+  }
```
tgrospic, line 75: All of these functions are to show how the code can be structured and not relevant for this example like mentioned in previous comment.

shared/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
tgrospic: I'm assuming that messages must be rad from some storage and it's not possible to get it without F. So having `M` it would be not possible to get `Set[M]` except having `F[Set[M]]` because we need side effects.

This shows that with traits without `F` parameter we are limited in what they can do. This is OK if this is the purpose e.g. to access only the fields or calculations without side-effects.

shared/src/main/scala/coop/rchain/sdk/dag/data/DagView.scala
```diff
@@ -0,0 +1,25 @@
+package coop.rchain.sdk.dag.data
+
+import fs2.{Chunk, Stream}
+
+/**
+  * Represents read-only view of the DAG starting from latest messages.
+  */
+trait DagView[F[_], M, S] {
+  implicit val dd: DagData[M, S]
+  implicit val ordM: Ordering[M]
```
tgrospic, line 10: This is important part of the abstraction. It gives additional dependency to interface that can be used to implement domain logic directly on the interface and not on the implementation (like we have on MultiParentCasper).

Implementation should be straightforward without domain logic and should only be connection to underlying side-effects like DB, FS and network operations or communication with other components.
**So implementation should deal with components dependency and not the interfaces.**

shared/src/main/scala/coop/rchain/sdk/dag/syntax/DagDataSyntax.scala
```diff
@@ -0,0 +1,32 @@
...
+    new DagDataSenderOps(s)
+}
+
+final class DagDataMessageOps[M, S](private val m: M) extends AnyVal {
+  /* Boilerplate code to enable field like access to M type */
+  def seqNum(implicit dagData: DagData[M, S]): Long           = dagData.seqNum(m)
+  def blockNum(implicit dagData: DagData[M, S]): Long         = dagData.blockNum(m)
+  def justifications(implicit dagData: DagData[M, S]): Set[M] = dagData.justifications(m)
+  def sender(implicit dagData: DagData[M, S]): S              = dagData.sender(m)
+  def bondsMap(implicit dagData: DagData[M, S]): Map[S, Long] = dagData.bondsMap(m)
```
tgrospic, line 20: This will be easier in Scala 3 with implicit on the extension level. It can be done with Scala 2 but with additional boxing because class cannot extend `AnyVal` anymore.
```scala
trait DataTypesSyntax:
  extension [M, S](m: M)(using dt: DataTypes[M, S])
    def sender         = dt.sender(m)
    def seqNum         = dt.seqNum(m)
    def justifications = dt.justifications(m)
```

shared/src/main/scala/coop/rchain/sdk/dag/data/DagView.scala
zsluedem: Yes~ This part is the most educational for me .

PR #3535, opened 2021-11-25
History::reset method is wrapped into F

### LINKED ISSUE #3533: Change History reset method to return value in F
## Overview

New implementation of History requires for instance to be created in `F` type.

## Solution

Change History interface `reset` method.

https://github.com/rchain/rchain/blob/01c4a1e6d24fe0813377dabbe8031a806e5deeb6/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L36

Instead it should return `F[History[F]]`.

```scala
def reset(root: Blake2b256Hash): F[History[F]]
```

This change should be done on `dev` branch and propagated to `feature/hard-fork-2` branch.

- [x] Create PR for `dev` branch with the changes
- [x] Merge PR to `dev` branch
- [x] Create _merge/sync_ PR from `dev` to `feature/hard-fork-2`
  example: https://github.com/rchain/rchain/pull/3496
  completed: #3555

stanislavlyalin:
## Overview
Issue #3533. Solution wraps `History::reset` method into monadic type `F`. Wrapping of `HistoryRepository::getHistoryReader` is related.

### Notes
If `Sync[F].delay` is the correct way to wrap into `F`?

bors try

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala
@@ -557,8 +557,10 @@ object HistoryInstances {
       (start, key, TriePath.empty).tailRecM(traverse)
     }
 
-    override def reset(root: Blake2b256Hash): History[F] =
-      this.copy(root = root, historyStore = CachingHistoryStore(historyStore.historyStore))
+    override def reset(root: Blake2b256Hash): F[History[F]] =
+      Sync[F].delay(
+        this.copy(root = root, historyStore = CachingHistoryStore(historyStore.historyStore))
+      )
 
   }
 
--- a/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
@@ -238,7 +238,7 @@ final case class HistoryRepositoryImpl[F[_]: Concurrent: Parallel: Log: Span, C,
   override def reset(root: Blake2b256Hash): F[HistoryRepository[F, C, P, A, K]] =
     for {
       _    <- rootsRepository.validateAndSetCurrentRoot(root)
-      next = history.reset(root = root)
+      next <- history.reset(root = root)
     } yield this.copy[F, C, P, A, K](currentHistory = next)
 
   override def exporter: F[RSpaceExporter[F]] = Sync[F].delay(rspaceExporter)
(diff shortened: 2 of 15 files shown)


rspace/src/main/scala/coop/rchain/rspace/history/HistoryInstances.scala
```diff
@@ -557,8 +557,10 @@ object HistoryInstances {
       (start, key, TriePath.empty).tailRecM(traverse)
     }
 
-    override def reset(root: Blake2b256Hash): History[F] =
-      this.copy(root = root, historyStore = CachingHistoryStore(historyStore.historyStore))
+    override def reset(root: Blake2b256Hash): F[History[F]] =
+      Sync[F].delay(
+        this.copy(root = root, historyStore = CachingHistoryStore(historyStore.historyStore))
+      )
```
tgrospic, line 563: :+1: The correct way to choose between Sync.delay and Applicative.pure to lift a value to F is to use delay when wrapping some expression which needs to be evaluated (function call) and pure when wrapping a single instance of some primitive or complex object (42.pure[F]).

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
```diff
@@ -253,8 +253,10 @@ final case class HistoryRepositoryImpl[F[_]: Concurrent: Parallel: Log: Span, C,
 
   override def getHistoryReader(
       stateHash: Blake2b256Hash
-  ): HistoryReader[F, Blake2b256Hash, C, P, A, K] =
-    new RSpaceHistoryReaderImpl(history.reset(root = stateHash), leafStore)(
+  ): F[HistoryReader[F, Blake2b256Hash, C, P, A, K]] =
+    for {
+      targetHistory <- history.reset(root = stateHash)
+    } yield new RSpaceHistoryReaderImpl(targetHistory, leafStore)(
```
tgrospic: This _for_ comprehension uses only the functor _map_ operation, so in these cases we usually don't use it and use _map_ directly.

```scala
history
  .reset(root = stateHash)
  .map(
    new RSpaceHistoryReaderImpl(_, leafStore)(
      Concurrent[F],
      serializeC,
      serializeP,
      serializeA,
      serializeK
    )
  )
```

rspace/src/test/scala/coop/rchain/rspace/history/SimplisticHistory.scala
```diff
@@ -306,7 +306,7 @@ final case class SimplisticHistory[F[_]: Sync](
     case (trie, path) => (trie, path.nodes)
   }
 
-  def reset(root: Blake2b256Hash): History[F] = this.copy(root = root)
+  def reset(root: Blake2b256Hash): F[History[F]] = Sync[F].delay(this.copy(root = root))
```
tgrospic, line 309: In this case use of _pure_ here would be fine because only object instance is created with field assignment, but it makes sense to use _delay_ which seems more error prone for future changes.

tgrospic (approved):
Excellent work!

zsluedem (approved):
LGTM

stanislavlyalin:
bors merge

Issue #3539, opened 2021-11-26
[Draft] Implement API to retrieve peer table


zsluedem:
## Overview

Currently, we can use `/metrics` to know how many peers the node finds while we can not know what are those peers.

We can expose the peer tables to an API and users can easily check that.

### Dependencies

TBD.

### Design

TBD.

PR #3543, opened 2021-11-26
Added network-id and shard name to API response

### LINKED ISSUE #3532: Add network-id and shard name to API response
## Overview

Configuration parameters `network-id`  and `shard-name` (`shardId` in a block) are necessary to configure the node correctly and connect to a network.

These parameters are not directly exposed to API so it's difficult to find out what they are when configuring the node and connecting to an existing network. To get the shard name, `/api/blocks` method can be used to read `shardId` from any block because it's part of the block data returned by the API but this is not the case for `network-id` which is not available.

## Solution

Add network and shard info to API status method. This means to change existing WebAPI and gRPC service endpoints.

### WebAPI

1. Change existing `/api/status` method:
    - include fields from existing `/status` method
    - remove `message` field
1. Add new fields `networkId` and `shardId` to `/api/status`.

WebAPI implementation is defined here and some additional dependencies will be needed to get the info from existing `/status` method.
https://github.com/rchain/rchain/blob/0a06ca896a0c3651cd2f3b4dfc12d5424d53b3c1/node/src/main/scala/coop/rchain/node/api/WebApi.scala#L120-L124

Existing status and version methods are defined here [coop/rchain/node/web](https://github.com/rchain/rchain/tree/c0bebe29bd8ad61fbffc382a5bbd8003be6b10e7/node/src/main/scala/coop/rchain/node/web).

### gRPC

1. Add new rpc method `status` to **DeployService**.
    `rpc status(google.protobuf.Empty) returns (StatusResponse){}`
1. Add `StatusResponse` message type with parameters the same as in WebAPI response.
1. Extend CLI options with the new _status_ method to make request to running node.

Status method don't have input parameters so built in `Empty` message type should be used as request type. Import in proto files is needed for `Empty` type.
```proto
import "google/protobuf/empty.proto";
```
In Scala this is the full path to `com.google.protobuf.empty.Empty` type.

Here are definitions for existing CLI options which supports gRPC calls to other node.
https://github.com/rchain/rchain/blob/45f5d2a4c1c867501e88a5d7606baef06cf3a3be/node/src/main/scala/coop/rchain/node/Main.scala#L134-L176

NOTE: Unchecking _Configure automatically_ in settings for IntelliJ plugin for protobuf and without any path in the list helped to better recognize references.
https://github.com/jvolkman/intellij-protobuf-editor#path-settings

stanislavlyalin:
## Overview
Issue #3532. Added network-id and shard name to WebAPI and gRPC response. Additionally added CLI parameter `status` to get node status info from another node.

### Notes
* Is it correct to place call of gRPC `status` method to `DeployRuntime`?
* Is it correct to use `VersionInfo.get` to getting version info instead of `versionInfo <- Sync[F].delay(VersionInfo.get)` such as [`StatusInfo.scala`](https://github.com/rchain/rchain/blob/c0bebe29bd8ad61fbffc382a5bbd8003be6b10e7/node/src/main/scala/coop/rchain/node/web/StatusInfo.scala#L20)?

bors try

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
@@ -110,6 +110,8 @@ object DeployRuntime {
       )
     )
 
+  def status[F[_]: Sync: DeployService]: F[Unit] = gracefulExit(DeployService[F].status)
+
   private def gracefulExit[F[_]: Monad: Sync, A](
       program: F[Either[Seq[String], String]]
   ): F[Unit] =
--- a/models/src/main/protobuf/DeployServiceCommon.proto
+++ b/models/src/main/protobuf/DeployServiceCommon.proto
@@ -222,7 +222,7 @@ message BlockEventInfo{
 }
 
 message Status {
-  int32 versionNumber = 1;
+  int32 version = 1;
   string versionInfo  = 2;
   string address      = 3;
   string networkId    = 4;
--- a/models/src/main/protobuf/DeployServiceCommon.proto
+++ b/models/src/main/protobuf/DeployServiceCommon.proto
@@ -222,7 +222,7 @@ message BlockEventInfo{
 }
 
 message Status {
-  VersionInfo version = 1;
+  VersionInfo version = 1 [(scalapb.field).no_box = true];
   string address      = 2;
   string networkId    = 3;
   string shardId      = 4;
--- a/models/src/main/protobuf/DeployServiceCommon.proto
+++ b/models/src/main/protobuf/DeployServiceCommon.proto
@@ -220,3 +220,14 @@ message BlockEventInfo{
   repeated SystemDeployInfoWithEventData systemDeploys = 3;
   bytes postStateHash = 4;
 }
+
+message Status {
+  int32 versionNumber = 1;
+  string versionInfo  = 2;
+  string address      = 3;
+  string networkId    = 4;
+  string shardId      = 5;
+  int32 peers         = 6;
+  int32 nodes         = 7;
+}
+
--- a/models/src/main/protobuf/DeployServiceCommon.proto
+++ b/models/src/main/protobuf/DeployServiceCommon.proto
@@ -222,12 +222,15 @@ message BlockEventInfo{
 }
 
 message Status {
-  int32 version = 1;
-  string versionInfo  = 2;
-  string address      = 3;
-  string networkId    = 4;
-  string shardId      = 5;
-  int32 peers         = 6;
-  int32 nodes         = 7;
+  VersionInfo version = 1;
+  string address      = 2;
+  string networkId    = 3;
+  string shardId      = 4;
+  int32 peers         = 5;
+  int32 nodes         = 6;
 }
 
+message VersionInfo {
+  string api  = 1;
+  string node = 2;
+}
--- a/models/src/main/protobuf/DeployServiceV1.proto
+++ b/models/src/main/protobuf/DeployServiceV1.proto
@@ -15,6 +15,8 @@ import "DeployServiceCommon.proto";
 
 import "scalapb/scalapb.proto";
 
+import "google/protobuf/empty.proto";
+
 option (scalapb.options) = {
   package_name: "coop.rchain.casper.protocol.deploy.v1"
   flat_package: true
@@ -60,6 +62,8 @@ service DeployService {
   rpc getBlocksByHeights(BlocksQueryByHeight) returns (stream BlockInfoResponse){}
   // temporary api for testing
   rpc getEventByHash(ReportQuery) returns (EventInfoResponse){}
+  // Get node status
+  rpc status(google.protobuf.Empty) returns (StatusResponse){}
 }
 
 message EventInfoResponse{
@@ -182,5 +186,12 @@ message BondStatusResponse {
   oneof message {
     ServiceError error = 1;
     bool isBonded   = 2;
+  }
+}
+
+message StatusResponse {
+  oneof message {
+    ServiceError error  = 1;
+    Status status = 2;
   }
 }
--- a/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
+++ b/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
@@ -297,7 +297,7 @@ object DeployGrpcServiceV1 {
           .flatMap(Observable.fromIterable)
 
       def status(request: com.google.protobuf.empty.Empty): Task[StatusResponse] =
-        Monixable.apply.toTask(for {
+        (for {
           address <- RPConfAsk[F].ask
           peers   <- ConnectionsCell[F].read
           nodes   <- NodeDiscovery[F].peers
@@ -311,6 +311,6 @@ object DeployGrpcServiceV1 {
             nodes.length
           )
           response = StatusResponse().withStatus(status)
-        } yield response)
+        } yield response).toTask
     }
 }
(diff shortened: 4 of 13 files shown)


stanislavlyalin:
bors try

stanislavlyalin:
bors try

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -214,7 +228,12 @@ object WebApi {
 
   final case class ApiStatus(
       version: Int,
-      message: String
+      versionInfo: String,
```
tgrospic: This two versions seems confusing.

Let's create separate VersionInfo object which can have multiple fields for specific module version.

```scala
ApiStatus(version: VersionInfo, ...)
VersionInfo(api: String, node: String)
```
Maybe better also to have API version to be string if we want to add some minor versions in the future.

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -274,7 +276,8 @@ object Setup {
           conf.devMode,
           if (conf.autopropose && conf.dev.deployerPrivateKey.isDefined) triggerProposeFOpt
           else none[ProposeFunction[F]],
-          blockReportAPI
+          blockReportAPI,
+          conf.protocolServer.networkId
```
tgrospic: Shard name should be supplied from configuration also (`conf.casper.shardName`).

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
```diff
@@ -291,5 +295,25 @@ object DeployGrpcServiceV1 {
...
+            version = 1,
+            VersionInfo.get,
+            address.local.toAddress,
+            networkId,
+            shardId,
+            peers.length,
+            nodes.length
+          )
+          response = StatusResponse().withStatus(status)
+        } yield response)
```
tgrospic: There is defined [extension method for `F[A]` with Monixable implicit](https://github.com/rchain/rchain/blob/a3807977676d3dbebc8cb30787ed2c0125a35fe5/shared/src/main/scala/coop/rchain/monix/Monixable.scala#L22) in scope and it's imported with `import coop.rchain.shared.syntax._`.

So this should work also:
```scala
(for { ... } yield response).toTask
```
---

Note also that definition of _apply_ functions means additional constructor so these two expressions have the same meaning.
```scala
// Directly use apply
Monixable.apply[F].toTask(...)
// or as a constructor like
Monixable[F].toTask(...)
```
In both cases Monixable instance must be present as implicit in scope.

stanislavlyalin:
bors try

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
```diff
@@ -110,6 +110,8 @@ object DeployRuntime {
       )
     )
 
+  def status[F[_]: Sync: DeployService]: F[Unit] = gracefulExit(DeployService[F].status)
```
zsluedem, line 113: It is fine to have it in command line although I don't think people use that now

models/src/main/protobuf/DeployServiceCommon.proto
```diff
@@ -220,3 +220,17 @@ message BlockEventInfo{
   repeated SystemDeployInfoWithEventData systemDeploys = 3;
   bytes postStateHash = 4;
 }
+
+message Status {
+  VersionInfo version = 1;
```
tgrospic: In protobuf default generated fields to complex types will be wrapped in Scala `Option` type. To avoid that _no-box_ option can be specified like here.
https://github.com/rchain/rchain/blob/382576eb9030c7a03fdff0d75a8a2a41e80dc378/models/src/main/protobuf/RhoTypes.proto#L59

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -122,10 +125,20 @@ object WebApi {
...
+        address <- RPConfAsk[F].ask
+        blocks  <- getBlocks(depth = 1)
+        shardId = blocks.headOption.getOrElse(LightBlockInfo()).shardId
+        peers   <- ConnectionsCell[F].read
+        nodes   <- NodeDiscovery[F].peers
+      } yield ApiStatus(
+        version = VersionInfo(api = 1.toString, node = coop.rchain.node.web.VersionInfo.get),
+        address.local.toAddress,
+        networkId,
+        shardId,
```
tgrospic, line 137: ShardId should also be used from configuration like in gRPC service.

models/src/main/protobuf/DeployServiceV1.proto
```diff
@@ -184,3 +188,10 @@ message BondStatusResponse {
     bool isBonded   = 2;
   }
 }
+
+message StatusResponse {
+  oneof message {
+    ServiceError error  = 1;
```
zsluedem, line 194: @tgrospic Should we use some new style of error message for grpc API?
Actually there is no internal error for the statusAPI. Maybe we can just skip this?

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
```diff
@@ -291,5 +295,21 @@ object DeployGrpcServiceV1 {
             }
           )
           .flatMap(Observable.fromIterable)
+
+      def status(request: com.google.protobuf.empty.Empty): Task[StatusResponse] =
+        (for {
+          address <- RPConfAsk[F].ask
+          peers   <- ConnectionsCell[F].read
+          nodes   <- NodeDiscovery[F].peers
```
zsluedem, line 303: Just want to say that an easy extend on this method can implement https://github.com/rchain/rchain/issues/3539

models/src/main/protobuf/DeployServiceV1.proto
zsluedem: Oh. I thought this api would just static information except peer count.

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
```diff
@@ -291,5 +295,21 @@ object DeployGrpcServiceV1 {
...
+          status = Status(
+            version = VersionInfo(api = 1.toString, node = coop.rchain.node.web.VersionInfo.get),
+            address.local.toAddress,
+            networkId,
+            shardId,
+            peers.length,
+            nodes.length
```
zsluedem, line 310: What is the difference between nodes and peers actually?

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
zsluedem: Since we add this, why not make it complete?

tgrospic: @zsluedem Are you suggesting to add complete peer table in status response? This can be a big list so better to have separate method.

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
tgrospic: Peers are maintained by ConnectionsCell and nodes are managed by Kademlia. That should mean that peers are actively connected nodes.

tgrospic (approved):
LGTM

node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala
zsluedem: Yes. You are right. An new api would be better.

models/src/main/protobuf/DeployServiceV1.proto
tgrospic: @zsluedem The _error_ here is just to be consistent with other results. But yes, we should get rid of errors inside service definitions and use _protocol_ errors. E.g for gRPC https://www.grpc.io/docs/guides/error/.

casper/src/main/scala/coop/rchain/casper/util/comm/DeployRuntime.scala
tgrospic: We should be consistent with API interface in any supported protocol. Eventually we should have only one API implementation with many protocols preferably just generated from the API interface.

zsluedem (approved):
LGTM

stanislavlyalin:
bors merge

PR #3544, opened 2021-11-28
Implement new method on BlockAPI to get data at name

### LINKED ISSUE #3538: Implement replace API for listenForDataAtName
## Overview

**Current shortcoming of `listenForDataAtName`**

When users try to request `listenForDataAtName` API, the node would traverse blocks based on the request param `depth` which is very low efficiency and waste.  Because it requires traversal through blocks _depth_ is limited to only 3 blocks by default to prevent attacking the node in this way.

We want to remove this API and provide improved way for users to make request for data at some name (channel). This also includes data sent on `rho:rchain:deployId`.

### Impact

Any clients which use this API should consider move to the new API.

### Design

For those users who want to explore deploy effect on some unforgeable name, **the user need to find
which block the deploy is packed and then explore the unforgeable name like `ExploreDeploy`**.

Send a request with fields below: 
```scala
final case class(
  blockHash: String,
  par: Par,
  usePreStateHash: Boolean
)
```
Response should be the same as for exploratory deploy `(Seq[Par], LightBlockInfo)`.

- [x] Create new method on BlockAPI `getDataAtName` with corresponding tests
  #3544
- [x] Create gRPC binding for the new method
  #3592
- [x] Create HTTP binding for the new method
  #3584

zsluedem:
## Overview

First step for https://github.com/rchain/rchain/issues/3538

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -27,7 +27,7 @@ import coop.rchain.models.BlockHash.BlockHash
 import coop.rchain.models.syntax._
 import coop.rchain.models.rholang.sorter.Sortable._
 import coop.rchain.models.serialization.implicits.mkProtobufInstance
-import coop.rchain.models.{BlockMetadata, Par}
+import coop.rchain.models.{BindPattern, BlockMetadata, Par}
 import coop.rchain.rspace.ReportingRspace.ReportingEvent
 import coop.rchain.rspace.hashing.StableHashProvider
 import coop.rchain.rspace.trace._
@@ -791,5 +791,18 @@ object BlockAPI {
       )
     )
   }
+
+  def getDataAtPar[F[_]: Concurrent: RuntimeManager: BlockStore](
+      blockHash: String,
+      par: Par,
+      usePreStateHash: Boolean
+  ): F[Seq[Par]] =
+    for {
+      block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
+      stateHash = if (usePreStateHash) block.body.state.preStateHash
+      else block.body.state.postStateHash
+      sortedPar <- parSortable.sortMatch[F](par).map(_.term)
+      data      <- RuntimeManager[F].getData(stateHash)(sortedPar)
+    } yield data
 
 }


casper/src/main/scala/coop/rchain/casper/api/RuntimeAPI.scala
```diff
@@ -0,0 +1,71 @@
...
+import coop.rchain.casper.MultiParentCasper
+import coop.rchain.casper.engine.EngineCell
+import coop.rchain.casper.engine.EngineCell.EngineCell
+import coop.rchain.casper.util.rholang.RuntimeManager
+import coop.rchain.models.{BindPattern, Par}
+import coop.rchain.models.rholang.sorter.Sortable.parSortable
+
+object RuntimeAPI {
+
+  def getFromRuntime[F[_]: Concurrent: EngineCell: BlockStore, A, O](
```
nzpr: Why pull bock store and engine cell here? And not just RuntimeManager. Plus input stateHash instead of blockHash as you might want to call this on latest finalized state, which is not attributed to a block.

zsluedem: Then I don't actually need to implement any codes. Just use RuntimeManager API Then. @tgrospic agree?

We can ask @stanislavlyalin to proceed to the next step directly.

zsluedem: We can ask Stas to do the next.

tgrospic: @zsluedem We can use `getData` on RuntimeManager (right now we don't need to get continuation). This can be very similar method added to BlockAPI like exploratory deploy but instead of executing Rholang we just call getData on RM.

zsluedem: @tgrospic Not sure what you mean. Is my current pr needed?

tgrospic: Yes, I'm saying this new method can be on BlockAPI almost the same as exploratory deploy but instead of accepting Rholang term it should accept Par.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -792,4 +792,56 @@ object BlockAPI {
...
+          runtimeManager <- casper.getRuntimeManager
+          output         <- getFunc(runtimeManager, source, stateHash)
+        } yield output
+      result <- engine.withCasper(
+                 getData,
+                 Concurrent[F].raiseError[O](new Exception("Casper instance is not ready."))
+               )
+    } yield result
+
+  def getDataAtPar[F[_]: Concurrent: EngineCell: BlockStore](
```
tgrospic: Let's keep it simple and just expose access to data, continuations are not needed right now.

This function can accept `RuntimeManager` directly (available in Setup) which can spawn new Runtimes. It's not necessary to have EngineCell and Casper instance.

zsluedem:
@tgrospic 
RSpace direct API would something like
```scala
  def getDataAtPar[F[_]: Concurrent: BlockStore](
      blockHash: String,
      par: Par,
      usePreStateHash: Boolean
  )(
      implicit space: ISpace[F, Par, BindPattern, Datum[ListParWithRandom], WaitingContinuation[
        BindPattern,
        TaggedContinuation
      ]]
  ): F[Seq[Datum[ListParWithRandom]]] =
    for {
      block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
      stateHash = if (usePreStateHash) block.body.state.preStateHash
      else block.body.state.postStateHash
      sortedPar <- parSortable.sortMatch[F](par).map(_.term)
      // not safe if space is a shared instance
      _         <- space.reset(Blake2b256Hash.fromByteString(stateHash))
      data      <- space.getData(sortedPar)
    } yield data
```

tgrospic (approved):
LGTM

nzpr (approved):
LGTM

tgrospic:
bors merge

PR #3549, opened 2021-12-03
Check that bootstrap node and the current node has the same shard name

### LINKED ISSUE #3547: Validate shard name when approved block is received
## Overview

When node is first time connected to bootstrap node to download LFS state, it will download _approved_ block first.

We need to add a check that `shardId` of a received block and configuration parameter `shard-name` are equal meaning they are part of the same shard.

Network-id is already validated as part of lower level protocol code.
https://github.com/rchain/rchain/blob/7de591c89dc120abdbee7b93bfa791d003b4818b/comm/src/main/scala/coop/rchain/comm/transport/SslSessionServerInterceptor.scala#L70-L71

## Solution

Add shard check in `Initializing` engine when approved block is received and throw exception if it's not correct, together with existing validation.

https://github.com/rchain/rchain/blob/db4becad9f4d73e9bdd84c477012dcc7295c989f/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala#L125-L127

To validate the solution LFS (Last Finalized State) feature should be used. Two nodes should be connected, one as a standalone which should be used as bootstrap for the second. Second node after start will _ask_ the bootstrap node for _approved block_ and after it's received the new checking of the shard should happen.

@stanislavlyalin as part of this task please use this repo to test LFS sync with the node on test net. The timings for the synching should be updated on the issue #3402.

stanislavlyalin:
## Overview
Issue #3547. When ApprovedBlock is received from bootstrap node, checking that both nodes have the same shard names is performed.

### Notes
Sample output for same shard names
```
ApprovedBlock is self-signed by ceremony master.
Received approved block from bootstrap node.
Valid approved block Block #0 (918c70e135...) with empty parents (supposedly genesis) received. Restoring approved state.
Block Block #0 (918c70e135...) with empty parents (supposedly genesis) sender is empty.
Sending StoreItemsRequest to bootstrap
Stream collected.
...
```

Sample output for different shard names
```
Invalid LastFinalizedBlock received; refusing to add.
Connected to the wrong shard. Approved block received from bootstrap is in shard 'root' but expected is 'another-shard'. Check configuration option shard-name.
```

bors try

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
@@ -131,6 +131,13 @@ class Initializing[F[_]
 
       _ <- Log[F].info("Invalid LastFinalizedBlock received; refusing to add.").whenA(!isValid)
 
+      _ <- Log[F]
+            .info(
+              s"Bootstrap's node shard name '${approvedBlock.candidate.block.shardId}' " +
+                s"doesn't equals to current node shard name '${casperShardConf.shardName}'."
+            )
+            .whenA(!shardNameIsValid)
+
       // Start only once, when state is true and approved block is valid
       start <- startRequester.modify {
                 case true if isValid  => (false, true)
--- a/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
@@ -89,6 +89,7 @@ class Initializing[F[_]
       disableStateExporter: Boolean
   ): F[Unit] = {
     val senderIsBootstrap = RPConfAsk[F].ask.map(_.bootstrap.exists(_ == sender))
+    val shardNameIsValid  = approvedBlock.candidate.block.shardId == casperShardConf.shardName
 
     def handleApprovedBlock = {
       val block = approvedBlock.candidate.block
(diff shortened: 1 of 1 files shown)


tgrospic (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
```diff
@@ -124,12 +127,19 @@ class Initializing[F[_]
     for {
       // TODO resolve validation of approved block - we should be sure that bootstrap is not lying
       // Might be Validate.approvedBlock is enough but have to check
-      isValid <- senderIsBootstrap &&^ Validate.approvedBlock[F](approvedBlock)
+      isValid <- senderIsBootstrap &&^ Validate.approvedBlock[F](approvedBlock) &&^ shardNameIsValid.pure
```
nzpr: I believe here its better to put `shardNameIsValid.pure` before `Vaidate`, as validation is something expensive.

nzpr (approved):
LGTM, left minor comment

casper/src/main/scala/coop/rchain/casper/engine/Initializing.scala
tgrospic: Agree, shard check can be on the first place to check because the value is already computed (only wrapped in pure F).

stanislavlyalin:
bors merge

PR #3556, opened 2021-12-08
Parameterize minimum Phlo price

### LINKED ISSUE #3314: Parameterize minimum Phlo price
## Overview

Currently phlo price is hard coded and set to 1 (10^-8 REV) by default. This does not generate sufficient transaction fees to cover the network+hardware cost. The minimum needs to be suitably increased, to allow sufficient cost recovery to allow and encourage decentralization.  

### Impact

1. Changing this will impact wallets, dAppy and other projects - we need to inform them of intended changes so they can update their software accordingly.
2. Currently the network runs a lot of dummy transactions so that when a user deploy comes along, it can be included in a block. Raising the minimum phlo will cause more transaction costs for the coop.  So the timing of this change needs to take this into account.

### Design

From technical perspective, RNode needs to provide a configuration parameter for `min-phlo-price` with default value 1 as it's now hard-coded.
Deploy with phlo price with lower value will be rejected with an error just like now with phlo price `0`.

Additionally, RNode should provide API for external users to query the validators minimum phlo price.

### Solution

- [x] Add `minPhloPrice` as configuration parameter
  - [default.conf](https://github.com/rchain/rchain/blob/e0a833ea05/node/src/main/resources/defaults.conf#L184) in _casper_ section with default value `1`
  - mapped to object [CasperShardConf](https://github.com/rchain/rchain/blob/db4becad9f4d/casper/src/main/scala/coop/rchain/casper/Casper.scala#L113-L131)
- [x] Add `minPhloPrice` to gRPC and WebAPI `/status` method
- [x] Use min phlo price in BlockAPI to check if deploy is correct
  https://github.com/rchain/rchain/blob/a8ae9399ccf3eef7219f029bb5d231307993a4af/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L79-L84
- [x] Add validation of phlo price in [MultiParentCasperImpl](https://github.com/rchain/rchain/blob/c3fae859afd6f6/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L274) in block validation, all deploys (BlockMessage#body#deploys) must have phlo price at least minimum

stanislavlyalin:
## Overview
Added parametrization of minimal phlogiston price. To do this, you need to set the `--min-phlo-price` CLI parameter when running the node. Default value defined in [defaults.conf](https://github.com/rchain/rchain/blob/530ec0ea124e477a9219175ba09f4ef34fc9e0ce/node/src/main/resources/defaults.conf#L325) and equals `1`. For each deploy in block performs check that it's phlo price is greater or equal than the minimal phlogiston price. If at least one of the deploys has lower phloPrice, propose will be rejected. More info about `phloPrice` and usage [here](https://github.com/rchain/rchain/wiki/Cost-accounting-in-RChain).

The motivation is to provide sufficient transaction fees to cover the cost of the network and equipment.

### How to check validation is performs

1. Run two nodes in a bundle: bootstrap and validator (read-only node). See schema below:
```
(1)(2)
+------------------+                +-------------------+
|    bootstrap     |  (5) validate  |     validator     |----+
|                  |--------------->|    (read-only)    |    | (6) Log warning: One or more deploys...
| minPhloPrice = 1 |                | minPhloPrice = 10 | <--+
+------------------+                +-------------------+
     ^      ^
     |      |
     |      |
(3) deploy[phloPrice=1]
            |
      (4) propose
```
2. Bootstrap node should have lower `min-phlo-price`. For example, default value `1`. Validator node should have greater `min-phlo-price`. For example, `10`.
3. Make deploys on bootstrap node with appropriate phloPrice, for example `1`, but lower than needed for validator node.
4. Propose on bootstrap node
5. Validator will start validation
6. Warning log message `One or more deploys has phloPrice lower than 10` will be printed to stdout.

**Command to running bootstrap node**
```shell
./rnode run -s --validator-private-key <VALIDATOR_PRIVATE_KEY> --wallets-file wallets.txt --bonds-file bonds.txt --host 127.0.0.1 --allow-private-addresses
```
**Note**. The bootstrap node should be started from a "clean" state.

**Command to running validator node**
```shell
run -b "<BOOTSTRAP_NODE_ADDRESS>" --host 127.0.0.1 --allow-private-addresses --api-port-grpc-external 40411 --api-port-grpc-internal 40412 --api-port-admin-http 40415 --api-port-http 40413 --protocol-port 40410 --discovery-port 40414 --min-phlo-price 10 --data-dir .
```
**Note**. The `--data-dir` for validator node should be different from the bootstrap node, so this parameter should be provided.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -128,7 +128,7 @@ final case class CasperShardConf(
     bondMaximum: Long,
     epochLength: Int,
     quarantineLength: Int,
-    minPhloPrice: Int
+    minPhloPrice: Long
 )
 
 sealed abstract class MultiParentCasperInstances {
--- a/casper/src/main/scala/coop/rchain/casper/Casper.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Casper.scala
@@ -127,7 +127,8 @@ final case class CasperShardConf(
     bondMinimum: Long,
     bondMaximum: Long,
     epochLength: Int,
-    quarantineLength: Int
+    quarantineLength: Int,
+    minPhloPrice: Int
 )
 
 sealed abstract class MultiParentCasperInstances {
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -303,8 +303,16 @@ class MultiParentCasperImpl[F[_]
 
         // This validation is only to punish validator which accepted lower price deploys.
         // And this can happen if not configured correctly.
-        // _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
-        // _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
+        minPhloPrice = casperShardConf.minPhloPrice
+        _ <- EitherT(Validate.phloPrice(b, minPhloPrice)).recoverWith {
+              case _ =>
+                val warnToLog = EitherT.liftF[F, BlockError, Unit](
+                  Log[F].warn(s"One or more deploys has phloPrice lower than $minPhloPrice")
+                )
+                val asValid = EitherT.rightT[F, BlockError](BlockStatus.valid)
+                warnToLog *> asValid
+            }
+        _ <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
 
         depDag <- EitherT.liftF(CasperBufferStorage[F].toDoublyLinkedDag)
         status <- EitherT(EquivocationDetector.checkEquivocations(depDag, b, s.dag))
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -300,6 +300,8 @@ class MultiParentCasperImpl[F[_]
               EquivocationDetector.checkNeglectedEquivocationsWithUpdate(b, s.dag, approvedBlock)
             )
         _      <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+        _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
+        _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
         depDag <- EitherT.liftF(CasperBufferStorage[F].toDoublyLinkedDag)
         status <- EitherT(EquivocationDetector.checkEquivocations(depDag, b, s.dag))
         _      <- EitherT.liftF(Span[F].mark("equivocation-validated"))
--- a/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
@@ -299,9 +299,13 @@ class MultiParentCasperImpl[F[_]
         _ <- EitherT(
               EquivocationDetector.checkNeglectedEquivocationsWithUpdate(b, s.dag, approvedBlock)
             )
-        _      <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
-        _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
-        _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
+        _ <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+
+        // This validation is only to punish validator which accepted lower price deploys.
+        // And this can happen if not configured correctly.
+        // _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
+        // _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
+
         depDag <- EitherT.liftF(CasperBufferStorage[F].toDoublyLinkedDag)
         status <- EitherT(EquivocationDetector.checkEquivocations(depDag, b, s.dag))
         _      <- EitherT.liftF(Span[F].mark("equivocation-validated"))
--- a/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
@@ -534,7 +534,6 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments) 
     // Similar value `casper.min-phlo-price` defined in defaults.conf and should be moved to Scala based config
     val minPhloPrice = opt[Long](
       descr = "MinPhloPrice",
-      default = Some(1),
       validate = _ >= 0
     )
 
--- a/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
+++ b/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
@@ -531,6 +531,13 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments) 
       descr = "Private key for dummy deploys."
     )
 
+    // Similar value `casper.min-phlo-price` defined in defaults.conf and should be moved to Scala based config
+    val minPhloPrice = opt[Long](
+      descr = "MinPhloPrice",
+      default = Some(1),
+      validate = _ >= 0
+    )
+
   }
   addSubcommand(run)
 
--- a/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
+++ b/casper/src/main/scala/coop/rchain/casper/BlockStatus.scala
@@ -30,6 +30,7 @@ object BlockStatus {
   def containsExpiredDeploy: BlockError    = InvalidBlock.ContainsExpiredDeploy
   def containsFutureDeploy: BlockError     = InvalidBlock.ContainsFutureDeploy
   def notOfInterest: BlockError            = InvalidBlock.NotOfInterest
+  def lowDeployCost: BlockError            = InvalidBlock.LowDeployCost
 
   def isInDag(blockStatus: BlockStatus): Boolean =
     blockStatus match {
@@ -86,6 +87,7 @@ object InvalidBlock {
   case object ContainsExpiredDeploy   extends InvalidBlock
   case object ContainsFutureDeploy    extends InvalidBlock
   case object NotOfInterest           extends InvalidBlock
+  case object LowDeployCost           extends InvalidBlock
 
   val slashableOffenses: Set[InvalidBlock] =
     Set(
--- a/casper/src/main/scala/coop/rchain/casper/CasperConf.scala
+++ b/casper/src/main/scala/coop/rchain/casper/CasperConf.scala
@@ -22,7 +22,8 @@ final case class CasperConf(
     heightConstraintThreshold: Long,
     roundRobinDispatcher: RoundRobinDispatcher,
     genesisBlockData: GenesisBlockData,
-    genesisCeremony: GenesisCeremonyConf
+    genesisCeremony: GenesisCeremonyConf,
+    minPhloPrice: Int
 )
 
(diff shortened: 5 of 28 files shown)


stanislavlyalin:
bors try

casper/src/main/scala/coop/rchain/casper/Casper.scala
```diff
@@ -127,7 +127,8 @@ final case class CasperShardConf(
     bondMinimum: Long,
     bondMaximum: Long,
     epochLength: Int,
-    quarantineLength: Int
+    quarantineLength: Int,
+    minPhloPrice: Int
```
tgrospic: PhloPrice on DeployData is Long so better to have Long here also. Sorry, my mistake telling you Int.

stanislavlyalin: Done. I couldn't find a way, how to change `int32` to `int64` in [DeployServiceCommon.proto](https://github.com/stanislavlyalin/rchain/blob/1953cd52c4b8247be7b9ddf77b5fa774a742a93a/models/src/main/protobuf/DeployServiceCommon.proto). Changing type leads to error in generated code. So additional converting in gRPC `status` method [was provided](https://github.com/stanislavlyalin/rchain/blob/82ce6fb49eb168e423962714b4ec580c87fbb36f/node/src/main/scala/coop/rchain/node/api/DeployGrpcServiceV1.scala#L312).

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -300,6 +300,8 @@ class MultiParentCasperImpl[F[_]
               EquivocationDetector.checkNeglectedEquivocationsWithUpdate(b, s.dag, approvedBlock)
             )
         _      <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+        _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
+        _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
```
zsluedem: I am thinking maybe we can exclude validating here because the coop is controlling all the nodes.

stanislavlyalin: @zsluedem, thanks for your careful consideration! This is the initial requirement. See the last checkbox in [issue description](https://github.com/rchain/rchain/issues/3314#issue-799496971). Also, as Tomislav answered me in Discord:
> ... you can run your own network.

tgrospic: @zsluedem This is a good point. This validation is only to punish validator which accepted lower price deploys. And this can happen if not configured correctly.

Let's comment out this validation for now.

stanislavlyalin: Done

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -299,7 +299,9 @@ class MultiParentCasperImpl[F[_]
         _ <- EitherT(
               EquivocationDetector.checkNeglectedEquivocationsWithUpdate(b, s.dag, approvedBlock)
             )
-        _      <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+        _ <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+        // _      <- EitherT(Validate.phloPrice(b, casperShardConf.minPhloPrice))
+        // _      <- EitherT.liftF(Span[F].mark("phlogiston-price-validated"))
```
zsluedem: @stanislavlyalin Sorry Stas. One more minor request. Could you put a comment in codes why these lines are commeted?

zsluedem: Maybe it would be even better to have an issue to keep track on it.

stanislavlyalin: @zsluedem, I added Tomislav's comment as explanation :grinning:

stanislavlyalin:
bors try

stanislavlyalin:
bors try

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -299,7 +299,13 @@ class MultiParentCasperImpl[F[_]
         _ <- EitherT(
               EquivocationDetector.checkNeglectedEquivocationsWithUpdate(b, s.dag, approvedBlock)
             )
-        _      <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+        _ <- EitherT.liftF(Span[F].mark("neglected-equivocation-validated"))
+
+        // This validation is only to punish validator which accepted lower price deploys.
+        // And this scan happen if not configured correctly.
```
tgrospic: It's a small typo here, you meant _can happen_ right?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: @zsluedem I've added a task to check this validation on our secure list for HF3.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
stanislavlyalin: I just copied your [comment](https://github.com/rchain/rchain/pull/3556#discussion_r765490942) as explanation :relaxed: Fix from *scan* to *can*?

tgrospic: Yes, sorry my typo.  :) I corrected it in the comment.

stanislavlyalin: It's Ok :) I fixed typo in code comment

zsluedem:
@stanislavlyalin 
In order to test minimam phlo price right, you have to write a new tests in https://github.com/rchain/rchain/blob/dev/integration-tests/test/test_propose.py.You can imitate the tests already in the file.

1. The key point is that you add another option to run a node and you have to pass the added param in https://github.com/rchain/rchain/blob/dev/integration-tests/test/rnode.py#L527-L539 and start the node with the added param.  
2. Then just normal deploy with lower phlo price than the node config and expect some errors.https://github.com/rchain/rchain/blob/dev/integration-tests/test/rnode.py#L253

zsluedem:
@tgrospic @stanislavlyalin 
Well. For scala tests, I don't really have very good ideas to test. The best way seems like test https://github.com/rchain/rchain/pull/3556/files#diff-2379b0158e88d31fd00dc8e83ac032d15e3f712a420ae6e5e0adae1ebb48c8f3R646-R654 
and mock a block with bigger phlo price deploy

tgrospic:
> @tgrospic @stanislavlyalin Well. For scala tests, I don't really have very good ideas to test. The best way seems like test https://github.com/rchain/rchain/pull/3556/files#diff-2379b0158e88d31fd00dc8e83ac032d15e3f712a420ae6e5e0adae1ebb48c8f3R646-R654 and mock a block with bigger phlo price deploy

Test can call BlockAPI.deploy function.
https://github.com/rchain/rchain/blob/fbb93fedaa7103987cf2e10d0bde38b66119a10f/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L80-L84

stanislavlyalin:
> @stanislavlyalin In order to test minimam phlo price right, you have to write a new tests in https://github.com/rchain/rchain/blob/dev/integration-tests/test/test_propose.py.You can imitate the tests already in the file.
> 
>     1. The key point is that you add another option to run a node and you have to pass the added param in https://github.com/rchain/rchain/blob/dev/integration-tests/test/rnode.py#L527-L539 and start the node with the added param.
> 
>     2. Then just normal deploy with lower phlo price than the node config and expect some errors.https://github.com/rchain/rchain/blob/dev/integration-tests/test/rnode.py#L253

@zsluedem, @tgrospic, parameter added just in `default.conf` as pointed in [description](https://github.com/rchain/rchain/issues/3314). Is there a way to set corresponded parameter from CLI? Should integration test runs two nodes, master and slave, where slave makes deploys with different `phloPrice` values? ~And what about commented validation?~ (I think `BlockAPI.deploy` has corresponded check)

https://github.com/rchain/rchain/blob/28725a4af5df7727955efc50e411b3b0b7c3e801/casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala#L304-L307

tgrospic:
@stanislavlyalin Yes, we need CLI option also to be added as part of `run` command. CLI options should by in sync with default.conf.
https://github.com/rchain/rchain/blob/8ba22c07f1166f2d94ecc910c3f56ca686d964e2/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala#L145-L147

For now tests should only validate if BlockAPI received deploy with invalid phloPrice. When we uncomment validation of phloPrice in deploys inside block we will need additional tests for that.

For integration test only one node is enough and request to push a new deploy which should be rejected if phloPrice is too low.

casper/src/main/scala/coop/rchain/casper/Casper.scala
tgrospic: Fix is in this commit https://github.com/rchain/rchain/pull/3556/commits/6d4a87a8e022253cb8fbc9de51a22f63156b8137. The reason for the error was in custom implementation for object hashCode which must enumerate all types that are mentioned in generated code with [StacksafeScalapbGenerator](https://github.com/rchain/rchain/blob/af703b79cd27/project/StacksafeScalapbGenerator.scala).

zsluedem:
>  Should integration test runs two nodes, master and slave, where slave makes deploys with different phloPrice values?

one validator node is enough. FYI, there are no master and slave roles in the RChain network. (It is not the same as other distributed systems but similar.)  But there are indeed two kinds of nodes in RChain [observer and validator](https://gist.github.com/zsluedem/b2828d95681d4d0883f010316deb4fba#mainnet-servers)

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -531,6 +531,13 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
       descr = "Private key for dummy deploys."
     )
 
+    // Similar value `casper.min-phlo-price` defined in defaults.conf and should be moved to Scala based config
+    val minPhloPrice = opt[Long](
+      descr = "MinPhloPrice",
+      default = Some(1),
```
tgrospic: Do we need default here, it should be populated from Hocon file default.conf when merged with CLI options?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: Instead of completely comment out this validation let's create _warning_ log message. This way we can see in the logs when we have this situation.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
stanislavlyalin: Yes, you are right. Default value merges to NodeConf from defaults.conf. Fixed in [commit](https://github.com/rchain/rchain/pull/3556/commits/1864881c2e01aed1c025bf31dcddfa01b9bf30a5).

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
stanislavlyalin: @zsluedem, @tgrospic, fixed in [commit](https://github.com/rchain/rchain/pull/3556/commits/e3d85586f57ac3144a8b0e625cff827a52dd83cb)

tgrospic (approved):
LGTM

Very good work!

zsluedem (approved):
LGTM

stanislavlyalin:
bors merge

PR #3557, opened 2021-12-08
Cleanup HotStore in-memory implementation

### LINKED ISSUE #3553: Cleanup HotStore in-memory implementation
## Overview

HotStore is fixed for race condition causing hash mismatch error in PR #3526. Part of the code can be simplified and be written in more functional style.

## Cleanup HotStore and related tests

- [x] Remove unnecessary parentheses and braces
  - when using higher-order functions (functions as arguments)
    e.g. `f(x => {...})` -> `f { x => ...}`
  - when functions return F values
  - wrapping single values
- [x] Remove custom `zip` functions for Map (used only in HotStore)
 https://github.com/rchain/rchain/blob/b66a704886bd96249fc2db115377f3e5dde6bc2a/rspace/src/main/scala/coop/rchain/rspace/HotStore.scala#L311
- [x] Use functions instead of manually destructuring (getOrElse, map, ...)
- [x] Create intermediate variables for long expressions
- [x] Create intermediate type aliases for long expressions
- [x] Use `apply` definition for HotStore (smart) constructors
- [x] Rewrite HotStoreState smart constructor with all default arguments
  - update HotStoreSpec to only specify non default arguments
- [x] Replace Ref.of with short version in HotStoreSpec, ReplayRSpaceTests, StorageTestsBase
  `Ref.of[Task, HotStoreState[C, P, A, K]]` => `Ref[Task].of`
- [x] HotStoreSpec
    - remove additional parameter `M[_]`
    - make Ref parameter in class History explicit
- [x] After finishing all refactorings, use long running test to verify hash mismatch error (@zsluedem)

stanislavlyalin:
## Overview
Refactored `HotStore` code and related code in steps listed in the task description.

### Notes
I could not find how to globally add aliases for types, so this work has to be done in every trait. 


rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -152,21 +152,11 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
...
-                   // just return what is in hot store already
-                   case Some(data) => (state, data)
-                 }
-               })
-    } yield (result)
+      result <- hotStoreState.modify[Seq[Datum[A]]] { state =>
+                 val data = state.data.getOrElse(channel, fromHistoryStore)
+                 (state.copy(data = state.data.updated(channel, data)), data)
+               }
+    } yield result
```
tgrospic, line 173: This is not quite the same, in case when data is found on the channel (in _state_) we don't need to update the state with the same data.

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
```diff
@@ -132,27 +118,15 @@ trait HotStoreSpec[F[_], M[_]] extends FlatSpec with Matchers with GeneratorDriv
...
-                      ),
-                      ()
-                    )
-                )
-            _     <- hotStore.installContinuation(channels, installedContinuation)
-            res   <- hotStore.getContinuations(channels)
-            cache <- state.get
+            _ <- state.modify { _ =>
+                  (HotStoreState(continuations = Map(channels -> cachedContinuations)), ())
+                }
```
tgrospic: If previous state is not used when modified then it's discarded, so `set` method on Ref can be used.

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
```diff
@@ -1066,12 +866,12 @@ trait HotStoreSpec[F[_], M[_]] extends FlatSpec with Matchers with GeneratorDriv
     (channel: String, datumValue: String) =>
       fixture { (_, _, hotStore) =>
         val key   = channel
-        val datum = Datum.create(channel, datumValue, false)
+        val datum = Datum.create(channel, datumValue, persist = false)
```
tgrospic, line 822: :+1:

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: Fixed with using `Option.fold` method

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
stanislavlyalin: Done

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -211,22 +221,19 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
...
+                   state.copy(joins = state.joins.updated(channel, fromHistoryStore)),
+                   state.installedJoins.getOrElse(channel, Seq.empty) ++: fromHistoryStore
+                 )
+                 // ...or just return what is in hot store already
+                 state.joins
+                   .get(channel)
+                   .fold(historyStoreIfEmpty)(
+                     joins => (state, state.installedJoins.getOrElse(channel, Seq.empty) ++: joins)
+                   )
+               }
```
tgrospic, line 230: Try to rewrite this by using `state.joins.get(channel).map(...)`. We don't want to execute any operation on a _state_ if not necessary.
We can define `def historyStoreIfEmpty` but I'd say it makes things less clear.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -4,29 +4,35 @@ import cats.syntax.all._
...
 trait HotStore[F[_], C, P, A, K] {
-  def getContinuations(channels: Seq[C]): F[Seq[WaitingContinuation[P, K]]]
-  def putContinuation(channels: Seq[C], wc: WaitingContinuation[P, K]): F[Unit]
-  def installContinuation(channels: Seq[C], wc: WaitingContinuation[P, K]): F[Unit]
+  type Continuation    = WaitingContinuation[P, K]
+  type Continuations   = Seq[Continuation]
+  type Data            = Datum[A]
+  type HotStoreActions = Seq[HotStoreAction]
+  type RowMap          = Map[Seq[C], Row[P, A, K]]
+  type State           = HotStoreState[C, P, A, K]
```
tgrospic: Type aliases are useful to make the code more readable when types are very long. But they must be used sparingly and only when really necessary because when reading the code you must have it in the head and recalculate.

So I'd leave `Seq[_]` and `Datum[_]` (or any pattern _[_]) as is because it's not complicated type.

`RowMap` is only used in trait definition and once in implementation, so it's not very beneficial to have it as alias and force reader to unpack it.

Because aliases are defined with trait type parameters (C, P, A, K) they cannot be used outside of trait scope, e.g. for `State` this means it cannot be used on many places which makes it not very helpful.

My intention for aliases was more towards this `Deferred[F, Seq[WaitingContinuation[P, K]]]` which is used multiple times in a scope of a function. But also the question is if it's necessary to use it when calling _modify_.

---

For alias to be justified it must satisfy these criteria:
- drastically reduce the length of type definition
- used in smallest possible scope or be a global.

These are not very precise rules but it can serve as guidelines. Aliases are great language feature but it can make code much harder to reason about if not used very carefully.

stanislavlyalin: @tgrospic, yes, I agree. Reasonable reasons. I tried to fix it.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: Here is the example using Functor#map.
```scala
state.data
 .get(channel)
 .map((state, _))
 .getOrElse(
   state.copy(data = state.data.updated(channel, fromHistoryStore)),
   fromHistoryStore
 )
```

stanislavlyalin: @tgrospic, a very elegant solution! I didn't think of that. I updated the code

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: Done

stanislavlyalin:
bors try

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -256,112 +261,118 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
...
+      r <- historyStoreCache.modify { cache =>
+            cache.datums
+              .get(channel)
+              .map(v => (cache, (v, true)))
+              .getOrElse(cache.copy(datums = cache.datums + (channel -> d)), (d, false))
+          }
       (deferred, isComplete) = r
-      _                      <- (historyReaderBase.getData(channel) >>= deferred.complete).whenA(!isComplete)
+      data                   <- historyReaderBase.getData(channel)
+      _                      <- deferred.complete(data) whenA !isComplete
```
tgrospic: This change changed the behavior from previous code. `historyReaderBase.getData(channel)` was called only when deferred is not complete.

It would be interesting if this has influence on PoS test failure.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -4,15 +4,16 @@ import cats.syntax.all._
...
-import coop.rchain.shared.MapOps._
 import coop.rchain.rspace.internal._
 
 import scala.collection.immutable.Map
 
 trait HotStore[F[_], C, P, A, K] {
-  def getContinuations(channels: Seq[C]): F[Seq[WaitingContinuation[P, K]]]
-  def putContinuation(channels: Seq[C], wc: WaitingContinuation[P, K]): F[Unit]
-  def installContinuation(channels: Seq[C], wc: WaitingContinuation[P, K]): F[Unit]
+  type Continuation = WaitingContinuation[P, K]
```
tgrospic: @nzpr, @zsluedem Should we rename class `WaitingContinuation` to `Continuation` or `Kont` or ...?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -256,112 +261,118 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
...
-              cache.datums.get(channel) match {
-                case Some(v) => (cache, (v, true))
-                case None    => (cache.copy(datums = cache.datums + (channel -> d)), (d, false))
-              }
-            })
+      r <- historyStoreCache.modify { cache =>
+            cache.datums
+              .get(channel)
+              .map(v => (cache, (v, true)))
+              .getOrElse(cache.copy(datums = cache.datums + (channel -> d)), (d, false))
```
tgrospic: Please change tuple definition with `->` as other tuples. Both are ok but mixing it in the same function (and in the same line) looks confusing.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
```diff
@@ -256,112 +261,118 @@ private class InMemHotStore[F[_]: Concurrent, C, P, A, K](
...
-      mapped        = zipped.mapValues { case (d, k) => Row(d, k) }
+      data          = cache.data.map(_.leftMap(Seq(_)))
+      continuations = cache.continuations ++ cache.installedContinuations.mapValues(Seq(_))
+
+      dataAt = (i: Seq[C]) => data.getOrElse(i, Seq.empty[Datum[A]])
+      continuationAt = (i: Seq[C]) =>
+        continuations.getOrElse(i, Seq.empty[WaitingContinuation[P, K]])
+      mapped = (data.keySet ++ continuations.keySet).map { i =>
+        (i, Row(dataAt(i), continuationAt(i)))
+      }.toMap
```
tgrospic: Try to rewrite this _zipping_ with the use of `padZip` and `mapValues`, it can be little more succinct.

Also remove helper _zip_ function with its tests.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: @tgrospic, probably no. I found that one of the [previous commit](https://github.com/rchain/rchain/pull/3557/commits/04e0ddc0c8ae962ffbf145523acc7d88195ae667) already leads to the problem. I ran an automated check to test previous commits to detect when the problem happens. If the change you pointed me above changed behavior, I'll revert it.

tgrospic: OK, but even if it's not related we don't need to load data from underlying DB when data is already loaded in the HotStore state (Deferred is completed).

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
```diff
@@ -1377,31 +1130,25 @@ trait InMemHotStoreSpec extends HotStoreSpec[Task, Task.Par] {
...
   ) =
     (for {
-      historyState <- Ref.of[F, HotStoreState[String, Pattern, String, StringsCaptor]](
-                       HotStoreState[String, Pattern, String, StringsCaptor]()
-                     )
+      historyState <- Ref[F].of(HotStoreState[String, Pattern, String, StringsCaptor]())
       history = {
-        implicit val hs = historyState
-        new History[F, String, Pattern, String, StringsCaptor]
+        new History[F, String, Pattern, String, StringsCaptor](historyState)
```
tgrospic: Curly braces are not necessary now.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: @tgrospic, should it also be replaced in HotStoreSpec?

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: @tgrospic, reverted to the initial implementation

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: @tgrospic, @nzpr, this type was eliminated in the last implementation.

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
stanislavlyalin: @tgrospic, done in [commit](https://github.com/rchain/rchain/pull/3557/commits/d19b2d88dc85e99ceee0bd8ff24c9ef7de0f55f8). Also removed corresponding files.

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
stanislavlyalin: @tgrospic, [done](https://github.com/rchain/rchain/pull/3557/commits/a395749d5017a6e535a7eccc129fb72bf38e69d0).

rspace/src/test/scala/coop/rchain/rspace/HotStoreSpec.scala
stanislavlyalin: @tgrospic, is empty parentheses, in this case, is the previous state? I thought it is what should `modify` returns in an atomic way.

zsluedem (approved):
LGTM

rspace/src/main/scala/coop/rchain/rspace/HotStore.scala
tgrospic: It's fine.

tgrospic (approved):
Very good work and exercise!

stanislavlyalin:
bors merge

PR #3559, opened 2021-12-10
Create PoS vault based on unforgeable name (PoS contract)

### LINKED ISSUE #3513: REV vaults (number channel) merging

### LINKED ISSUE #3514: Modify PoS contract to use single PoS address to collect rewards
## Overview

The capability to merge special number channels (#3513) gives a way to improve the operations of REV vaults. Transactions that includes the same source or target REV address can be executed in parallel and inside merged block balance will be recalculated.

This is important for [PoS contract](https://github.com/rchain/rchain/blob/3371c8a400f4b82a7af833f25b9a0b6fc71709db/casper/src/main/resources/PoS.rhox#L27) which now uses per validator vaults created just for the purpose to prevent conflicts in each block when rewards are written into single PoS address. This solution still has the problem to create conflicts when merging multiple blocks from the same validator.
Similar situation is with _pending_ rewards state. It's also updated based on validator public key which creates conflicts in merging.
To resolve both problems, PoS contract can be simplified with the use of _mergeable_ REV vaults.

Another issue is rewards distribution calculation which now happens after each deploy. It's expensive and grows with the number of validators.

### Single PoS vault

Create single PoS vault and use it for all validators to transfer rewords in PoS (pre-charge, refund, bond) operations. Existing per validator vaults can be completely removed.
Now code which involves REV vault transfers can be written without limitations for deploy to be rejected if source or target address runs in parallel.

PoS vault should be created as multi-sig vault which will be govern similar to existing Coop address.

### Lazy rewards distribution

Rewards distribution should be implemented as lazily as possible. Rewards are distributed proportionally to the stake of validators and can be calculated on the basis of each calculated reward or on accumulated sum.
PoS address can collect all rewards and distribution calculation can be done only on epoch change when validators can be replaced.
Slashing can possibly be another time when this is necessary. This can be prevented if stake of slashed validator is moved to coop address immediately when slashed and its rewards is left to other validators.

### Steps to solution

- [x] fix rewards in POS. Pre-charge and refund are using **ONE** PoS address to collect stakes and remove perValidatorVaults
- [x] Create POS vault as ~~multi-sig~~ unforgeable vault.
- [x] implement function to calculate rewards for each validator and re-organize reward process in epoch

### LINKED ISSUE #3571: Add genesis-pos-vault-pub-key parameter in genesis config.
## Overview

After https://github.com/rchain/rchain/pull/3559, the node still needs a genesis parameter to config the pubkey.

### Impact

In genesis, we need to config this parameter. 

### Design

1. add one more `genesis-pos-vault-pubkey`(maybe some other names) to https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/node/src/main/resources/defaults.conf#L243-L290
2. add another option named `genesisPosVaultPubKey` in https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala#L454-L456
3. The new config `genesis-pos-vault-pubkey`should be similar the config as [minimumBond](https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala#L25) and is finally rendered in https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala#L10-L30
4.  The goal is to replace the value `posPubKey` in https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/casper/src/main/resources/PoS.rhox#L150 to the configure value

**Understand tests in https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/integration-tests/test/test_wallets.py is very important**

The goal is to add a new `test_transfer_from_pos_vault` in https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/integration-tests/test/test_wallets.py

1. add the `genesis-pos-vault-pub-key`command args in  https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/integration-tests/test/rnode.py#L531-L542
2. create a bootstrap node with the `genesis-pos-vault-pub-key` with a new private key
3. try to transfer the money inside the pos contract use deploy -> https://github.com/rchain/rchain/blob/11d7d732e70d42cc273613a6cefd24467d46930c/integration-tests/test/rnode.py#L255-L266 
4. check if the balance is changed correctly

In step 3 , the contract should be like below
```scala
 new return, rl(`rho:registry:lookup`), poSCh, stdout(`rho:io:stdout`), 
     deployerId(`rho:rchain:deployerId`), transferRetCh in {
   rl!(`rho:rchain:pos`, *poSCh) |
   for(@(_, PoS) <- poSCh) {
     @PoS!("posVaultTransfer", "$targetAddr", $amount, deployerId, *transferRetCh) |
     for (@(true, _)<- transferRetCh){
       stdout!("transfer successfully!")
     }
   }
 }
```

replace `$targetAddr`and `$amount` with the provided value.

### Resource estimation

3 days maximum?

zsluedem:
## Overview

Update POS contract.

## Design Detail

Life circle in POS

POS internal state:
"allBonds"        : Map[Validator, bonds],
"activeValidators": Set[Validator],
"withdrawers"     : Map[Validator, (bonds, quantinue)],
"comittedRewards" : Map[Validator, rewards], 
"pendingWithdrawers": Map[Validator, quantinue]

**There are two vaults `posVault` and `coopMultiSigVault` while pos contract have no controll on `coopMultiSigVault` but `posVault`**

Several crucial calculate points:

### UserDeploy

* precharge -> user transfer the precharge funding to posVault
* refund -> posVault transfer the refund amount back to userVault

### Bonding

User transfer the bonding amount directly to posVault. `allBonds` would update with {Validator: bondAmount}.

### Withdraw(Unbond)

`pendingWithdrawers` would update with {Validator: quantinue}

### EpochChange(CloseBlock)

1. Calculate all the validator current epoch rewards based on active validator bonding proportion
2. Accumulate `comittedRewards` with the result of step 1
4. Update `withdrawers` according to `allBonds` and `pendingWithdrawers` when pendingWithdrawers is not empty, remove the validator in `allBonds`
5. When the block number reaches the quantinue of `withdrawers` , transfer the corresponding bonds+ rewards to the corresponding validator from posVault. 
6. update `pendingWithdrawers` to empty Map {}.
5. Pick new `activeValidators`

### Slashing
1. Transfer the bonding amount of the slashed validator based on `allBonds` from posVault to coopMultiSigVault.
2. Set the bonding amount in `allBonds` of the slashed validator to 0
3. remove the slashed validator in `activeValidators`

### Notes

After this pr, we need to implement test in another pr. Details see https://github.com/rchain/rchain/issues/3571.

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -430,7 +430,7 @@ in {
                   }
                 }
               } |
-              new commitRewards, moveFunds in {
+              new commitRewards in {
                 // Private method which transfers all of a validator's rewards + bond to the Coop vault.
                 // Slash expects (Boolean, Either[Nil, String]) on returnCh.
                 contract PoS(@"slash", @deployerId, @blockHash, @sysAuthToken, returnCh) = {
@@ -607,26 +607,6 @@ in {
                         }
                       } else {
                         ackCh!((false, "Invalid system auth token"))
-                      }
-                    }
-                  }
-                } |
-                // Moves the entire contents of a validator's vault to the PoS vault.
-                contract moveFunds(@(unf, revAddr, vault), returnCh) = {
-                  new balanceCh, unfAuthKeyCh, transferDoneCh in {
-                    @vault!("balance", *balanceCh) |
-                    for (@balance <- balanceCh) {
-                      if (balance > 0) {
-                        @RevVault!("unforgeableAuthKey", unf, *unfAuthKeyCh) |
-                        for (@authKey <- unfAuthKeyCh) {
-                          // Transfer funds from validator vault to the PoS vault.
-                          @vault!("transfer", posVaultAddr, balance, authKey, *transferDoneCh) |
-                          for (@(true, _) <- transferDoneCh) {
-                            returnCh!(true)
-                          }
-                        }
-                      } else {
-                        returnCh!(true)
                       }
                     }
                   }
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -117,9 +117,8 @@ in {
           )
         }
       } |
-      // create unforgeable name vault with the provided pub key
       // ret: the return channel of the create
-      // pubKey: the pubkey which controll the unforgeable name
+      // pubKey: the pubkey which control the unforgeable name
       contract createUnfVault(ret, @pubkey) = {
         new unf, unfAddrRet, createRetCh, pubKeyChainUnfVault, unfVaultHandle, stdout(`rho:io:stdout`) in {
           revAddressOps!("fromUnforgeable", *unf, *unfAddrRet) |
@@ -149,7 +148,7 @@ in {
         }
       }|
       createUnfVault!(*posVaultCh, "$$posPubKey$$") |
-      for (@(posVaultUnf, posVaultAddr, posVault, humanControllHandle) <- posVaultCh){
+      for (@(posVaultUnf, posVaultAddr, posVault, humanControlHandle) <- posVaultCh){
       @RevVault!("unforgeableAuthKey", posVaultUnf, *posAuthKeyCh) |
       for (@posDeployerRevAddress <- posDeployerRevAddressCh;
            @posDeployerAuthKey    <- posDeployerAuthKeyCh;
@@ -529,13 +528,13 @@ in {
                                   stateProcessCh, paymentDoneCh,
                                   accDepositsDoneCh, commitRewardsCh,
                                   newValidatorsCh, removeQuarantinedCh, currentEpochRewardCh,
-                                  accumulateRewards, updatedRewardsCh, movePendingWithdrawCh, movePendingWithdrawer
+                                  commitCurrentEpochRewards, updatedRewardsCh, movePendingWithdrawCh, movePendingWithdrawer
                               in {
                                 PoS!("getCurrentEpochRewards", * currentEpochRewardCh) |
                                 for (@currentRewards <- currentEpochRewardCh) {
                                   runMVar!(*stateCh, *stateProcessCh, *ackCh) |
                                   for (@state, stateUpdateCh <- stateProcessCh) {
-                                    accumulateRewards!(state, currentRewards, *updatedRewardsCh) |
+                                    commitCurrentEpochRewards!(state, currentRewards, *updatedRewardsCh) |
                                     for (@newCommittedState <- updatedRewardsCh) {
                                       movePendingWithdrawer!(newCommittedState, *movePendingWithdrawCh) |
                                       for (@newMovedPendingState <- movePendingWithdrawCh){
@@ -559,7 +558,7 @@ in {
                                     }
                                   }
                                 } |
-                                contract accumulateRewards(@state, @currentRewards, returnCh) = {
+                                contract commitCurrentEpochRewards(@state, @currentRewards, returnCh) = {
                                   new accumulate in {
                                     @ListOps!("fold", currentRewards.toList(), state, *accumulate, *returnCh) |
                                     contract accumulate(@(pk, reward), @updatedState, resultCh) = {
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -4,6 +4,51 @@
 // initialBonds - the initial bonds map
 // epochLength - the length of the validation epoch in blocks
 // quarantineLength - the length of the quarantine time in blocks
+/*
+
+Life circle in POS
+
+POS internal state:
+"allBonds"        : Map[Validator, bonds],
+"activeValidators": Set[Validator],
+"withdrawers"     : Map[Validator, (bonds, quantinue)],
+"comittedRewards" : Map[Validator, rewards], 
+"pendingWithdrawers": Map[Validator, quantinue]
+
+**There are two vaults `posVault` and `coopMultiSigVault` while pos contract have no controll on `coopMultiSigVault` but `posVault`**
+
+Several crucial calculate points:
+
+## UserDeploy
+
+* precharge -> user transfer the precharge funding to posVault
+* refund -> posVault transfer the refund amount back to userVault
+
+## Bonding
+
+User transfer the bonding amount directly to posVault. `allBonds` would update with {Validator: bondAmount}.
+
+## Withdraw(Unbond)
+
+`pendingWithdrawers` would update with {Validator: quantinue}
+
+
+## EpochChange(CloseBlock)
+
+1. Calculate all the validator current epoch rewards based on active validator bonding proportion
+2. Accumulate `comittedRewards` with the result of step 1
+4. When the block number reaches the quantinue of `withdrawers` , transfer the corresponding bonds+ rewards to the corresponding validator from posVault. 
+5. Update `withdrawers` according to `allBonds` and `pendingWithdrawers` when pendingWithdrawers is not empty, remove the validator in `allBonds`
+6. update `pendingWithdrawers` to empty Map {}.
+5. Pick new `activeValidators`
+
+
+## Slashing
+1. Transfer the bonding amount of the slashed validator based on `allBonds` from posVault to coopMultiSigVault.
+2. Set the bonding amount in `allBonds` of the slashed validator to 0
+3. remove the slashed validator in `activeValidators`
+
+*/
 
 /*
  The table below describes the required computations and their dependencies
@@ -37,7 +82,7 @@ new PoS,
     PendingRewards, MergeIntMaps, treeHashMapCh,
     sysAuthTokenOps(`sys:authToken:ops`),
     multiSigRevVaultCh, coopVaultCh,
-    uriOut, sumFromPair
+    uriOut, sumFromPair, stdlog(`rho:io:stdout`)
 in {
   // Simultaneously retrieves RevVault, ListOps, TreeHashMap, and MultiSigRevVault contracts from the registry.
   registryLookup!(`rho:rchain:revVault`, *revVaultCh) |
@@ -122,39 +167,44 @@ in {
             }
           } |
           // Selects the initial active validators.
-          pickActiveValidators!($$initialBonds$$, {}, *initialActiveCh) |
+          pickActiveValidators!($$initialBonds$$.toList(), *initialActiveCh) |
           for (@initialActive     <- initialActiveCh; 
                @(true, _) <- moveInitialBondCh) {
-            new stateCh, pendingRewardsInitializedCh, rewardsInfo, totalActiveBonds, totalBonds, totalWithdraw in {
+            new stateCh, pendingRewardsInitializedCh, rewardsInfo, calTotalActiveBonds, calTotalBonds, calTotalWithdraw, calTotalCommittedRewards in {
               // Initializes PoS state structure.
+              stdlog!(("initial active", initialActive))|
               stateCh!({
                 // Map[PublicKey, Int] - each validator stake
                 "allBonds"        : $$initialBonds$$,
                 // Set[PublicKey]     - the active validators
                 "activeValidators": initialActive,
-                // Map[PublicKey, (Int, Int, Int)] - the validators wishing to withdraw and the block number when their quarantine ends
+                // Map[PublicKey, (Int, Int)] - the validators wishing to withdraw and the block number when their quarantine ends
                 //                              This begins at the start of the next epoch + quarantineLength.
-                //                              Value(Int, Int) represents (original bond, reward, quantinue length) 
+                //                              Value(Int, Int) represents (original bond, reward, quantinue length)
                 "withdrawers"     : {},
+                // Map[PublicKey, Int]
+                "comittedRewards" : {}, 
                 // Map[PublicKey, Int] - the validators wishing to withdraw and the block number when their quarantine ends 
                 "pendingWithdrawers": {},
                 "randomImages"    : {},
                 "randomNumbers"   : {},
               }) |
               contract rewardsInfo(retCh) = {
-                new posBalanceCh, totalActiveBondCh, totalBondCh, totalWithdrawCh in {
+                new posBalanceCh, totalActiveBondCh, totalBondCh, totalWithdrawCh, committedRewardsCh in {
                   @posVault!("balance", *posBalanceCh) |
-                  totalBonds!(*totalBondCh) |
-                  totalActiveBonds!(*totalActiveBondCh) |
-                  totalWithdraw!(*totalWithdrawCh) |
-                  for (@posBalance <- posBalanceCh; @totalBond <- totalBondCh; @activeBonds <- totalActiveBondCh; @totalWithdraw <- totalWithdrawCh){
+                  calTotalBonds!(*totalBondCh) |
+                  calTotalActiveBonds!(*totalActiveBondCh) |
+                  calTotalWithdraw!(*totalWithdrawCh) |
+                  calTotalCommittedRewards!(*committedRewardsCh) |
+                  for (@posBalance <- posBalanceCh; @totalBond <- totalBondCh; @activeBonds <- totalActiveBondCh; 
+                       @totalWithdraw <- totalWithdrawCh; @totalCommittedRewards <- committedRewardsCh){
                     for (@state <<- stateCh) {
-                      retCh!((posBalance , totalBond , activeBonds, totalWithdraw, state.get("allBonds")))
+                      retCh!((posBalance , totalBond , activeBonds, totalWithdraw, totalCommittedRewards, state.get("allBonds")))
                     }
                   }
                 }
               } |
-              contract totalActiveBonds(returnCh) = {
+              contract calTotalActiveBonds(returnCh) = {
                 new computeSum, computeDelta in {
                   for (@state <<- stateCh){
                     @ListOps!("fold", state.get("allBonds").toList(), 0, *computeSum, *returnCh) |
@@ -168,7 +218,7 @@ in {
                   }
                 }
               } |
-              contract totalBonds(returnCh) = {
+              contract calTotalBonds(returnCh) = {
                 new computeSum, computeDelta in {
                   for (@state <<- stateCh){
                     @ListOps!("fold", state.get("allBonds").toList(), 0, *computeSum, *returnCh) |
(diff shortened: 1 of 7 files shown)


casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,167 @@ in {
           )
         }
       } |
-      for (@posRevAddress <- posRevAddressCh;
-           @posAuthKey    <- posAuthKeyCh) {
-        for (@(true, coopVault) <<- coopVaultCh) {
+      // create unforgeable name vault with the provided pub key
+      // ret: the return channel of the create
+      // pubKey: the pubkey which controll the unforgeable name
+      contract createUnfVault(ret, @pubkey) = {
```
zsluedem, line 122: ~~Actually, I want to move this to `RevVault.rho`. Does that sound OK?~~ @tgrospic

zsluedem: NVM.

integration-tests/test/test_unbond.py
```diff
@@ -32,7 +32,9 @@ def get_total_cost_from_block(block_info: BlockInfo) -> int:
         total_cost += deploy.cost
     return total_cost
 
-
+# TODO https://github.com/rchain/rchain/issues/3572
+# pylint: disable=too-many-locals
+@pytest.mark.skip
```
zsluedem, line 37: https://github.com/rchain/rchain/issues/3572
There are something wrong in unbonding.

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,165 @@ in {
...
+          pickActiveValidators!($$initialBonds$$.toList(), *initialActiveCh) |
+          for (@initialActive     <- initialActiveCh; 
+               @(true, _) <- moveInitialBondCh) {
+            new stateCh, pendingRewardsInitializedCh, rewardsInfo, calTotalActiveBonds, calTotalBonds, calTotalWithdraw, calTotalCommittedRewards in {
               // Initializes PoS state structure.
               stateCh!({
                 // Map[PublicKey, Int] - each validator stake
                 "allBonds"        : $$initialBonds$$,
-                // List[PublicKey]     - the active validators
+                // Set[PublicKey]     - the active validators
```
zsluedem, line 178: I changed the dataType to `Set` because I want to delete the active validator immediately when slashing happen.

zsluedem: It is easier to just use the `delete` method in `Set`

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,165 @@ in {
           )
         }
       } |
-      for (@posRevAddress <- posRevAddressCh;
-           @posAuthKey    <- posAuthKeyCh) {
-        for (@(true, coopVault) <<- coopVaultCh) {
+      // create unforgeable name vault with the provided pub key
```
tgrospic: This comment is not clear. Public key is not related to created vault backed by unforgeable name.

casper/src/test/resources/PoSTest.rho
```diff
@@ -242,12 +242,12 @@ match (
...
-                                  ((finalBonds, "==", initialBonds.set(userPubKey, bondAmt)), "user is correctly added to the bond map"),
-                                  ((finalRewards, "==", initialRewards), "rewards don't change for bonding"),
+                                  ((finalBonds, "==", initialBonds), "user is not in the bond map after withdraw"),
+                                  ((finalRewards, "==", initialRewards.set(userPubKey, 0)), "rewards don't change for bonding"),
                                   ((finalUserBalance, "==", initialUserBalance - bondAmt), "user balance should change by bond amount"),
                                   ((finalPosBalance, "==", initialPosBalance + bondAmt), "pos balance should change by bond amount"),
                                   ((true, "==", withdrawers.contains(userPubKey)), "withdrawer is moved to withdrawers map"),
-                                  ((quarantinePeriod, "==", withdrawers.get(userPubKey)), "withdrawer has correct quarantine finish block")
+                                  (((bondAmt, quarantinePeriod), "==", withdrawers.get(userPubKey)), "withdrawer has correct quarantine and bonds finish block")
```
tgrospic: Did you mean `... correct bond amount and quarantine finish block`?

casper/src/test/resources/PoSTest.rho
```diff
@@ -287,6 +287,7 @@ match (
                     (bondAmt, chargeAmt, refundAmt, blockNumber) => {
                       @PoS!("bond", validatorDeployerId, bondAmt, *bondCh) |
                       for (@(true, _) <- bondCh) {
+                        stdlog!("info", ("successfully bond bbbb... ")) |
```
tgrospic: Is `bbbb` reference to a bonded validator?

casper/src/test/resources/PoSTest.rho
```diff
@@ -446,9 +454,14 @@ match (
...
                         @PoS!("getRewards", *initialRewardsCh) |
+                        @PoS!("getBonds", *initialBondsCh) |
+                        @PoS!("getActiveValidators", *initialActiveCh) |
                         for (@initialPosBalance   <- initialPosBalanceCh;
                              @initialUser1Balance <- initialUser1BalanceCh;
-                             @initialRewards      <- initialRewardsCh) {
+                             @initialRewards      <- initialRewardsCh;
+                             @initialBonds <- initialBondsCh;
+                             @initialActives <- initialActiveCh) {
+                          stdlog!("info", ("bonding distribution and actives before prechard and refund", initialBonds, initialActives, initialRewards)) |
```
tgrospic: Typo on _prechard_.

casper/src/test/resources/PoSTest.rho
```diff
@@ -299,27 +300,33 @@ match (
                               for (_ <- closeBlockCh) {
                                 @PoS!("getRewards", *rewardsCh) |
                                 for (@rewards <- rewardsCh) {
+                                  stdlog!("info", ("After prechar and refund of the rewards", rewards)) |
```
tgrospic: Typo on _prechar_.

casper/src/main/resources/PoS.rhox
```diff
@@ -366,104 +525,89 @@ in {
...
-                                      }
+                                contract payWithdrawer(@(pk, amount), returnCh) = {
+                                  new vaultCh, revAddressCh in {
+                                    revAddressOps!("fromPublicKey", pk, *revAddressCh) |
+                                    for (@toRevAddress <- revAddressCh) {
+                                      @posVault!("transfer", toRevAddress, amount, posAuthKey, *returnCh)
+                                    }
+                                  }
+                                } |
+                                contract accumulateRewards(@state, @currentRewards, returnCh) = {
```
tgrospic: Maybe we can call this contract `commitCurrentEpochRewards` or `commitEpochRewards`?

integration-tests/test/test_unbond.py
tgrospic: Is the missing block from slashed validator?

casper/src/test/resources/PoSTest.rho
zsluedem: Sorry. It is some left over.

integration-tests/test/test_unbond.py
zsluedem: The missing block is from unbonded validator.

casper/src/main/resources/PoS.rhox
zsluedem: ~~Oh. Sorry. Some old history commit left over.~~

zsluedem: not sure what you mean here actually. Could you please clarify a little bit?

tgrospic: Comment can be misunderstood that the vault is created with provided public key.

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,165 @@ in {
...
+              contract calTotalCommittedRewards(returnCh) = {
+                new computeSum, computeDelta in {
+                  for (@state <<- stateCh){
+                    @ListOps!("fold", state.get("comittedRewards").toList(), 0, *computeSum, *returnCh) |
+                    contract computeSum(@(pk, reward), @acc, resultCh) = {
+                      resultCh!(acc + reward)
+                    }
+                  }
+                }
+              } |
```
tgrospic: These contracts are very similar, it would be nice if we they can be extracted to common one.

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,165 @@ in {
           )
         }
       } |
-      for (@posRevAddress <- posRevAddressCh;
-           @posAuthKey    <- posAuthKeyCh) {
-        for (@(true, coopVault) <<- coopVaultCh) {
+      // create unforgeable name vault with the provided pub key
+      // ret: the return channel of the create
+      // pubKey: the pubkey which controll the unforgeable name
```
tgrospic: Typo in _controll_.

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,165 @@ in {
...
+                  }
+                }
+              } |
+              ret!((*unf, unfAddr, vault ,*unfVaultHandle))
+            }
+          }
+        }
+      }|
+      createUnfVault!(*posVaultCh, "$$posPubKey$$") |
+      for (@(posVaultUnf, posVaultAddr, posVault, humanControllHandle) <- posVaultCh){
```
tgrospic: Typo in _Controll_.

casper/src/main/resources/PoS.rhox
zsluedem: Deleted

casper/src/main/resources/PoS.rhox
```diff
@@ -4,6 +4,51 @@
 // initialBonds - the initial bonds map
 // epochLength - the length of the validation epoch in blocks
 // quarantineLength - the length of the quarantine time in blocks
+/*
+
+Life circle in POS
```
tgrospic: Did you meant _life cycle_?

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,149 @@ in {
...
-                // List[PublicKey]     - the active validators
+                // Set[PublicKey]     - the active validators
                 "activeValidators": initialActive,
-                // Map[PublicKey, Int] - are moved from pendingRewards on epoch change or slashing
-                "committedRewards": {},
-                // Map[PublicKey, Int] - the validators wishing to withdraw and the block number when their quarantine ends
-                //                       This begins at the start of the next epoch + quarantineLength.
+                // Map[PublicKey, (Int, Int)] - the validators wishing to withdraw and the block number when their quarantine ends
+                //                              This begins at the start of the next epoch + quarantineLength.
+                //                              Value(Int, Int) represents (original bond, reward, quantinue length)
```
tgrospic: Is it the case that the first number will not always be the _original bond_ e.g. in the case of slashed validator?

casper/src/main/resources/PoS.rhox
tgrospic: `accumulateTotalBondsAlg` and `accumulateCommittedRewardsAlg` are completely the same.

`accumulateTotalWithdrawAlg` has nothing to do with _withdraw_, it's just sum up _n_ `(_, (n, _))` from first argument with the second argument.

casper/src/main/resources/PoS.rhox
```diff
@@ -150,6 +301,10 @@ in {
               contract PoS(@"getQuarantineLength", returnCh) = {
                 returnCh!($$quarantineLength$$)
               } |
+              contract PoS(@"humanTransfer", @targetAddress, @amount, @deployerID, @transferRet) = {
+                // human transfer would just use this method
+                @humanControlHandle!("transfer", targetAddress, amount, deployerID, transferRet)
```
tgrospic, line 304: Let's rename this to `posVaultTransfer`, the meaning of _human_ can be difficult to interpret.

casper/src/main/resources/PoS.rhox
```diff
@@ -607,24 +712,19 @@ in {
         }
       } |
       // Selects set of active validators from the set of available validators.
-      contract pickActiveValidators(@allBonds, @withdrawers, returnCh) = {
+      contract pickActiveValidators(@allBonds, returnCh) = {
```
tgrospic, line 707: Why withdrawers are not checked anymore?

zsluedem:
@tgrospic Just curious . How would you like to solve the `ListOp` performance problem right now? Do you want me to dig in or you already got something in mind?

I think it would be better to merge this one first and solve the performance problem later if there are no big problem currently.

tgrospic:
> @tgrospic Just curious . How would you like to solve the `ListOp` performance problem right now? Do you want me to dig in or you already got something in mind?
> 
> I think it would be better to merge this one first and solve the performance problem later if there are no big problem currently.

I agree, this should not be part of this PR.

casper/src/main/resources/PoS.rhox
zsluedem: https://github.com/rchain/rchain/pull/3559/files#diff-bdd5f44530e16460d88220a323332c78c93f7b69585cbf1fd816d032aa611ff2R521-R530
Here is the sequence of process in epoch. And that's the comment explanation  https://github.com/rchain/rchain/pull/3559/files#diff-bdd5f44530e16460d88220a323332c78c93f7b69585cbf1fd816d032aa611ff2R38-R43.

Currently `pickActiveValidators` would only happen in genesis and epoch. In epoch, currently it is almost doing everything in sequence as the explanation above. Before the step 5 to pick up `activeValidators`, the `allBonds` are already updated according to the the `pendingWithdrawers` in step 4 which exclude `withdrawers`. That's why withdrawers is not needed here any more.

casper/src/main/resources/PoS.rhox
zsluedem: Sorry. The comments don't make sense to me, either. I might change it inconsistently.

casper/src/main/resources/PoS.rhox
```diff
@@ -170,16 +170,17 @@ in {
...
                 "allBonds"        : $$initialBonds$$,
                 // Set[PublicKey]     - the active validators
                 "activeValidators": initialActive,
-                // Map[PublicKey, (Int, Int)] - the validators wishing to withdraw and the block number when their quarantine ends
-                //                              This begins at the start of the next epoch + quarantineLength.
-                //                              Value(Int, Int) represents (original bond, reward, quantinue length)
+                // Map[PublicKey, (Int, Int)] - The validators who issued a withdraw request , finished validating jobs and not active.
+                //                              When the validator reachs the quantinue length, it would be removed in the map and 
+                //                              pos would transfer back all the original bond and rewards back to validator vault. 
+                //                              Value(Int, Int) represents (original bond + reward, quantinue length)
```
tgrospic, line 183: Slashed validator will have 0 stake not the original amount?

casper/src/main/resources/PoS.rhox
```diff
@@ -311,39 +462,29 @@ in {
...
-                                                "withdrawers"     : state.get("withdrawers")
-                                                                          .set(validator,
-                                                                            $$quarantineLength$$ + $$epochLength$$ * (1 + blockNumber / $$epochLength$$)),
-                                                "randomImages"    : state.get("randomImages"),
-                                                "randomNumbers"   : state.get("randomNumbers")
-                                              },
-                                              (true, Nil))
-                                            }
-                                          }
+                                        for (@(true, _) <- transferDoneCh) {
```
tgrospic: It's definitely better to slash only if transfer is successful but now we don't handle failure. In case of transfer error slash should return error also.

casper/src/main/resources/PoS.rhox
```diff
@@ -72,35 +117,147 @@ in {
...
-                "committedRewards": {},
-                // Map[PublicKey, Int] - the validators wishing to withdraw and the block number when their quarantine ends
-                //                       This begins at the start of the next epoch + quarantineLength.
+                // Map[PublicKey, (Int, Int)] - The validators who issued a withdraw request , finished validating jobs and not active.
+                //                              When the validator reachs the quantinue length, it would be removed in the map and 
+                //                              pos would transfer back all the original bond and rewards back to validator vault. 
+                //                              Value(Int, Int) represents (original bond + reward, quantinue length)
                 "withdrawers"     : {},
+                // Map[PublicKey, Int]
+                "comittedRewards" : {},
```
tgrospic: Typo in _committed_.

casper/src/main/resources/PoS.rhox
```diff
@@ -366,104 +507,97 @@ in {
...
-                                                }
-                                              }
-                                            }
+                                    @ListOps!("filter", state.get("withdrawers").toList(), *isQuarantineFinished, *quarantinedValidatorsCh) |
+                                    for (@quarantinedValidators <- quarantinedValidatorsCh) {
+                                      @ListOps!("fold", quarantinedValidators, state, *computeRemove, *returnCh) |
+                                      contract computeRemove(@(pk, (bonds, quantine)), @updatedState, resultCh) = {
+                                        new payReturnCh in {
+                                          payWithdrawer!((pk, bonds + updatedState.get("comittedRewards").getOrElse(pk, 0)), *payReturnCh) |
+                                          for (@(true, _) <- payReturnCh) {
```
tgrospic: In case of transfer error this code will hang.

The next question is, if we handle the error what to do. Should we continue or rollback everything.
Because state is updated in steps, rollback is not easy. For rollback it would be better to prepare all state changes and update the state as a final step.

Transfer can fail if PoS has not enough REVs, so we can just log the error and continue. I'm not sure what else can we do without heavier changes.

casper/src/main/resources/PoS.rhox
```diff
@@ -366,104 +507,97 @@ in {
...
+                                    }
+                                  }
+                                } |
+                                contract commitCurrentEpochRewards(@state, @currentRewards, returnCh) = {
+                                  new accumulate in {
+                                    @ListOps!("fold", currentRewards.toList(), state, *accumulate, *returnCh) |
+                                    contract accumulate(@(pk, reward), @updatedState, resultCh) = {
+                                      resultCh!(updatedState.set("comittedRewards",
+                                                                  updatedState.get("comittedRewards").set(pk, updatedState.get("comittedRewards").getOrElse(pk, 0) + reward)))
+                                    }
```
tgrospic, line 562: ~Should we just remove all committed reward map?~

casper/src/main/resources/PoS.rhox
```diff
@@ -366,104 +507,97 @@ in {
...
-                                    new vaultCh, revAddressCh in {
-                                      revAddressOps!("fromPublicKey", pk, *revAddressCh) |
-                                      for (@toRevAddress <- revAddressCh) {
-                                        @posVault!("transfer", toRevAddress, amount, posAuthKey, *returnCh)
-                                      }
+                                contract payWithdrawer(@(pk, amount), returnCh) = {
+                                  new vaultCh, revAddressCh in {
+                                    revAddressOps!("fromPublicKey", pk, *revAddressCh) |
+                                    for (@toRevAddress <- revAddressCh) {
+                                      @posVault!("transfer", toRevAddress, amount, posAuthKey, *returnCh)
```
tgrospic: It would be nice to create target vault so we don't ends up hanging if not created before.

casper/src/main/resources/PoS.rhox
```diff
@@ -486,7 +620,7 @@ in {
                         @RevVault!("unforgeableAuthKey", unf, *unfAuthKeyCh) |
                         for (@authKey <- unfAuthKeyCh) {
                           // Transfer funds from validator vault to the PoS vault.
-                          @vault!("transfer", posRevAddress, balance, authKey, *transferDoneCh) |
+                          @vault!("transfer", posVaultAddr, balance, authKey, *transferDoneCh) |
```
tgrospic: _moveFunds_ is not used anymore. I missed that when removing per validator vaults.

casper/src/main/resources/PoS.rhox
```diff
@@ -366,104 +507,97 @@ in {
...
-                                                  resultCh!(not validatorsToWithdrawSet.contains(pk))
-                                                }
-                                              }
-                                            }
+                                    @ListOps!("filter", state.get("withdrawers").toList(), *isQuarantineFinished, *quarantinedValidatorsCh) |
+                                    for (@quarantinedValidators <- quarantinedValidatorsCh) {
+                                      @ListOps!("fold", quarantinedValidators, state, *computeRemove, *returnCh) |
+                                      contract computeRemove(@(pk, (bonds, quantine)), @updatedState, resultCh) = {
+                                        new payReturnCh in {
+                                          payWithdrawer!((pk, bonds + updatedState.get("comittedRewards").getOrElse(pk, 0)), *payReturnCh) |
```
tgrospic: We can skip transfers if amount is 0, in case of slashed validator.

casper/src/main/resources/PoS.rhox
tgrospic: Makes sense. We need to take into account if withdraw fails and we decide to skip it. In that case we still need to handle it here.

casper/src/main/resources/PoS.rhox
zsluedem: Oh. Slashed validator would not be in `withdrawer` now. Currently, the code would just transfer all bonds to coopMultiSigVault

casper/src/main/resources/PoS.rhox
zsluedem: Good catch. I am not sure what is the best approach. Actually, because some of the properties(rewards) are derived from some stored value, if PoS balance is changed(human transfer) , all values can be wrong. My original ultimate goal is to make sure every step correct and precise which all situation should be covered but it seems like difficult if we don't review very carefully.  Hanging should be a fatal situation in which people should jump .

But I agree with you. Once hanging happens, we can hardly do much to make things correct. I think the difficult part of rollback is not state change. It is the transfer which already sends out can not return back.  So `log the error and continue` is something that can make things look normal.

Anyway, I agree with you ` just log the error and continue.`

casper/src/main/resources/PoS.rhox
zsluedem: Make sense.

casper/src/main/resources/PoS.rhox
zsluedem: if we just `just log the error and continue`, we can use the old `parMap` way

casper/src/main/resources/PoS.rhox
zsluedem: slashed validator won't be withdrawers now. Why slashed validators have to be a withdrawer?

casper/src/main/resources/PoS.rhox
zsluedem: We could remove committed reward map if we return rewards to validators at epoch instead of keeping it which I think it is more economic friendly.

zsluedem: Or, we can accumulate the rewards into bonds which is more like coop's current staking plan.

tgrospic: I was thinking about the place where committed rewards can be cleared.

Let's keep existing logic and accumulate rewards until the epoch when validator will withdraw his stake + rewards.

casper/src/main/resources/PoS.rhox
tgrospic: I agree with this change. Later we can adapt to partial slashing where some amount of stake will be left.

tgrospic (approved):
Very good work! Thanks for addressing the comments.

stanislavlyalin (approved):
I'm not familiar with the Rholang code. Is PoS logic really written in Rholang? I'm surprised :slightly_smiling_face: I'll just approve the PR so it can be merged because I can not add something more competent.

zsluedem:
bors merge

PR #3564, opened 2021-12-14
Make 40402 the default internal grpc port to create Repl/eval and propose clients

### LINKED ISSUE #3562: Create repl client for CLI on internal port 40402
## Overview

Using RNode CLI to call gRPC methods of running instance of RNode doesn't correspond to default ports for _repl_ and _eval_ calls.

For example if RNode is running with default ports, gRPC Repl service is exposed on private port 40402 and repl client created by CLI methods _repl_ and _eval_ is created with default port 40401.
This means that the user of CLI is forced to always set gRPC port even when using default ports in running node.

Example how to use _eval_ method with Docker it is described in this [comment](https://github.com/rchain/rchip-proposals/issues/48#issuecomment-991793435) which needs `--grpc-port` to be specified although standalone node is started with default ports.

### Solution

Use by default internal (private) port 40402 for CLI repl client.

https://github.com/rchain/rchain/blob/8ba22c07f1166f2d94ecc910c3f56ca686d964e2/node/src/main/scala/coop/rchain/node/Main.scala#L109-L114

stanislavlyalin:
## Overview
Allows to use *repl/eval* mode without explicit passing `--grpc-port` parameter (see [comment](https://github.com/rchain/rchip-proposals/issues/48#issuecomment-991793435)). Now this mode can be used as shown below
```shell
./rnode --grpc-internal-port 40412 eval hello.rho
or
./rnode -i 40412 eval hello.rho
or simply (with default value 40402)
./rnode eval hello.rho
```
**Note**: In the first two cases master rnode should be run with `--api-port-grpc-internal 40412`.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -124,7 +124,7 @@ object Main {
     implicit val proposeServiceClient: GrpcProposeService[F] =
       new GrpcProposeService[F](
         options.grpcHost(),
-        options.grpcPort(),
+        grpcPort,
         options.grpcMaxRecvMessageSize()
       )
 
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -109,7 +109,7 @@ object Main {
     implicit val replServiceClient: GrpcReplClient[F] =
       new GrpcReplClient[F](
         options.grpcHost(),
-        options.grpcPort(),
+        options.grpcInternalPort(),
         options.grpcMaxRecvMessageSize()
       )
     implicit val deployServiceClient: GrpcDeployService[F] =
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -105,14 +105,14 @@ object Main {
   private def runCLI[F[_]: Sync: Monixable: ConsoleIO: Timer](
       options: commandline.Options
   ): F[Unit] = {
-    val internalPort =
-      if (options.grpcPort.isSupplied) options.grpcPort() else options.grpcInternalPort()
+    val grpcPort =
+      if (options.grpcPort.isSupplied) options.grpcPort() else commandline.Options.GrpcInternalPort
 
     // Clients for executing gRPC calls on remote RNode instance
     implicit val replServiceClient: GrpcReplClient[F] =
       new GrpcReplClient[F](
         options.grpcHost(),
-        internalPort,
+        grpcPort,
         options.grpcMaxRecvMessageSize()
       )
     implicit val deployServiceClient: GrpcDeployService[F] =
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -109,7 +109,7 @@ object Main {
     implicit val replServiceClient: GrpcReplClient[F] =
       new GrpcReplClient[F](
         options.grpcHost(),
-        options.grpcInternalPort(),
+        replPort(options),
         options.grpcMaxRecvMessageSize()
       )
     implicit val deployServiceClient: GrpcDeployService[F] =
@@ -181,6 +181,16 @@ object Main {
       }
     ) >> program
   }
+
+  /**
+    * If --grpc-port option specified, returns it's value. Otherwise returns port-grpc-internal default value
+    * @param options command line options
+    * @return port for repl/eval mode
+    */
+  private def replPort(options: commandline.Options): Int =
+    if (options.grpcPort.isSupplied) options.grpcPort()
+    else
+      Configuration.defaultConf().at("api-server.port-grpc-internal").load[Int].getOrElse(40402)
 
   private def subcommand(options: commandline.Options): Command =
     options.subcommand match {
--- a/node/src/main/scala/coop/rchain/node/Main.scala
+++ b/node/src/main/scala/coop/rchain/node/Main.scala
@@ -105,11 +105,14 @@ object Main {
   private def runCLI[F[_]: Sync: Monixable: ConsoleIO: Timer](
       options: commandline.Options
   ): F[Unit] = {
+    val internalPort =
+      if (options.grpcPort.isSupplied) options.grpcPort() else options.grpcInternalPort()
+
     // Clients for executing gRPC calls on remote RNode instance
     implicit val replServiceClient: GrpcReplClient[F] =
       new GrpcReplClient[F](
         options.grpcHost(),
-        replPort(options),
+        internalPort,
         options.grpcMaxRecvMessageSize()
       )
     implicit val deployServiceClient: GrpcDeployService[F] =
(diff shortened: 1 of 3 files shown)


node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -182,6 +183,27 @@ object Main {
...
+      options.grpcPort()
+    } else {
+      final case class Port(portGrpcInternal: Int)
+      Configuration
+        .defaultConf()
+        .at("api-server")
+        .load[Port]
+        .right
+        .getOrElse(Port(options.grpcPort()))
+        .portGrpcInternal
```
nzpr: Can you have the same without intriducing case class?
```scala
Configuration
        .defaultConf()
        .at("api-server.port-grpc-internal")
        .load[Int]
        .right
        .getOrElse(options.grpcPort())
```

stanislavlyalin: @nzpr, yes, it works. Thanks for the tip!

nzpr (approved):
LGTM

node/src/main/scala/coop/rchain/node/Main.scala
```diff
@@ -109,7 +110,7 @@ object Main {
     implicit val replServiceClient: GrpcReplClient[F] =
       new GrpcReplClient[F](
         options.grpcHost(),
-        options.grpcPort(),
+        replPort(options),
```
tgrospic: Sorry for misled you on the path to get `api-server.port-grpc-internal`. ~~Now I see it makes more sense to just change default port number in `grpc-port` config used for repl client.~~ But grpc-port is also used for DeployServiceClient client. So let's create a new config parameter for `grpc-internal-port` which will be used for ReplServiceClient and ProposeServiceClient with default 40402.

https://github.com/rchain/rchain/blob/8ba22c07f1166f2d94ecc910c3f56ca686d964e2/node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala#L128-L132

In the future we will unify configuration with Scala types which will contain defaults and parsing libs will provide overrides. Now _hocon_ configuration (default.conf) serves as the main source which is not what we want.

stanislavlyalin: @tgrospic, I changed default value for grpcPort to `40402`and reverted previous irrelevant commits.

nzpr: I don't think we want to change `grpcPort` default. Its a public API port, so when binary is in client mode, e.g. `rnode deploy` it should work.
I'm not sure why previous changes were not OK? The goal was to use `rnode eval` internal port by default, and this is what we want, right? Because eval server is running on internal port.

tgrospic: @nzpr Yes, I notice that and corrected my comment. We need separate parameter for repl client.

> I'm not sure why previous changes were not OK? The goal was to use rnode eval internal port by default, and this is what we want, right? Because eval server is running on internal port.

It should be possible to set internal port separate from external, so all commands like `deploy` and `eval` work with the default values.

Btw. it's confusing for users what it means to call `rnode eval my-file.rho`, it's not obvious that this is gRPC call.
I'd rather have something like `rnode grpc eval my-file.rho`.

tgrospic: For example, this is not helpful. Where am I connecting?
```
$ rnode eval --help
  -p, --print-unmatched-sends-only
  -h, --help                         Show help message

 trailing arguments:
  file-names (required)
```

stanislavlyalin: @nzpr, @tgrospic, [resolved](https://github.com/rchain/rchain/pull/3564/commits/0a24a8d8ceaa63d4c53e672581e470715fb73d78)

nzpr: @tgrospic @stanislavlyalin  I'm not sure I get this. Rnode binary in server mode (rnode run) has two API servers - external and internal (accessing restricted resources), one 40401, another 40402. Eval API endpoint is available only on internal port.
If you use rnode in client mode, you have --grpc-port and --grpc-host params in action. Then you can choose what client call you want to make. Eval is one of client calls.

So `rnode --grpc-port 40402 eval` calls eval on localhost:40402.

I don't get why to add another `--grpc-internal-port`. It just duplicates `--grpc-port` which is telling which port to use making client call, isn't it? I was thinking that by default we want eval use internal port, and this is the purpose of the PR.

tgrospic (approved):
LGTM

node/src/main/scala/coop/rchain/node/Main.scala
tgrospic: @nzpr It's confusing that you need to specify default port to call eval and repl.

nzpr: @tgrospic hm.. I don't get why initial version was not ok? Why not to use default internal port instead of default for `--grpc-port` CLI param for eval and repl client calls without adding one more param? Why to add param for this? Now user have to know that he should use `--grpc-port` for deploy and `--grpc-internal-port` for eval and repl. 
And `rnode --grpc-port 40402 repl` won't work

tgrospic: @nzpr I agree that the new parameter is not the real solution. Port and host are the main problem here which are not sub-commands of grpc client calls so it's confusing for which commands they are used.

You are right, initial solution makes more sense. It's only changing behavior when port is not specified which can hardly break any existing code.

We need to move definition of config with defaults in Scala anyway, so for now it's much easier to just define default for `run.apiPortGrpcInternal` option and use it as alternative value instead of reading default.conf.
```scala
options.grpcPort.orElse(options.run.apiPortGrpcInternal.toOption)()
```
And this also applies to ProposeService client.

nzpr: Do you mean it is more beneficial to have `--grpc-port --grpc-host` under `deploy`, `eval`, `propose`, `repl` - all of these subcommands?
So not make them rnode param, but param of subcommand? 
I'm ok with that, but imo now its fine as well, rnode binary by default can be treated as a client, so you tell which server endpoint to use, and then you specify the API you want to call. Or you can start the rnode server with `rnode run`.

tgrospic: > Do you mean it is more beneficial to have --grpc-port --grpc-host under deploy, eval, propose, repl - all of these subcommands?

Yes.

> rnode binary by default can be treated as a client

RNode is a client but with deploy, eval, ... commands its purpose is to relay messages to running client. And why would gRPC protocol be something special.
Something like this `rnode eval --http --port 443 deploy ...`. 

---

We should see configuration as as small VM which can have many interfaces which all can work in a similar way.
Each time command is called user must wait on initial loading of JVM which makes the whole command look slow. So for that we can create interactive console using the same API in the background.

```
$ rnode console
> RNode version v0.12.4 relay console
> set protocol http
> OK
> set external-port 443
> OK
> status
> {
  "address":"rnode://4a9...67f6@node2.testnet.rchain.coop?protocol=42400&discovery=42404",
  "version":"RChain Node 0.12.4 (f9ebcbd93ef308ca1175ccff7aeaca8a81da34cc)",
  "peers":7,
  "nodes":34
}
> set private-key "0x1234..."
> OK
> deploy --phlo-limit 200000 ...
> OK signature 1234...
> eval my-program.rho
> OK state-hash 0xabcd...
> set state-hash 0x1234...
```

tgrospic (approved):
LGTM

With the latest solution we have hard-coded constants defined in Scala for external and internal gRPC ports. This is quick fix for now but goes into direction where the whole configuration will be in Scala with all defaults. From Scala we can generate Hocon and CLI options but also support other config formats.

nzpr (approved):
LGTM

stanislavlyalin:
bors merge

PR #3574, opened 2021-12-28
Refactor graphz

### LINKED ISSUE #3567: Refactor Graphz to use Ref instead of MonadState
## Overview

Exporter of the DAG to visual representation uses Graphviz format which is implemented in `graphz` project. `GraphGenerator` uses graphz to generate textual representation from BlockDagRepresentation object and with some additional loading from BlockStorage.

## Problem

There are unnecessary complications and strange types defined (like `DagInfo[G[_]]` which is not using `G`, or `F[G[Graphz[G]]]`) which uses multiple higher-kinded types.
https://github.com/rchain/rchain/blob/e64a5f0d4719c4cbbdc0e340263cc05be54658c2/casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala#L26-L39

The source of these problems is in implementation of `GraphSerializer` which uses `MonadState` to keep internal progress.
https://github.com/rchain/rchain/blob/76f14c516bac3d6be2f91318c9b202b4239465bb/graphz/src/main/scala/coop/rchain/graphz/Graphz.scala#L14-L29

## Solution

Change `GraphSerializer` implementation to use `Ref` instead of MonadState and unify all related code to use only one HKT (`F[_]`).

We also need similar replacement on MonadState in cost accounting. Cost state should be held in `Ref` which eliminates the use of multiple HKT types. Here is the reference to almost finished PR of this change [CostStateRef](
https://github.com/tgrospic/rchain/pull/12/files#diff-07923aa0328bc0a5d353933d197c3e90cf500ac3b7a3a2a7eed32beecda14615R31).

stanislavlyalin:
## Overview
Refactoring Graphz: changing `MonadState` to `Ref` and related changes.
This PR addresses the issues identified in https://github.com/rchain/rchain/issues/3567. Changes allow eliminating using unnecessary data types, such as `G`. In general, changes make code more explicit.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -379,18 +379,17 @@ object BlockAPI {
 
   def visualizeDag[
       F[_]: Monad: Sync: EngineCell: Log: SafetyOracle: BlockStore,
-      G[_]: Monad,
       R
   ](
       depth: Int,
       maxDepthLimit: Int,
       startBlockNumber: Int,
-      visualizer: (Vector[Vector[BlockHash]], String) => F[G[Graphz[G]]],
-      serialize: G[Graphz[G]] => R
+      visualizer: (Vector[Vector[BlockHash]], String) => F[Graphz[F]],
+      serialize: F[Graphz[F]] => R
   ): F[ApiErr[R]] = {
     val errorMessage = "visual dag failed"
-    def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[R]] =
-      for {
+    def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[R]] = {
+      val graph = for {
         dag <- MultiParentCasper[F].blockDag
         // the default startBlockNumber is 0
         // if the startBlockNumber is 0 , it would use the latestBlockNumber for backward compatible
@@ -400,9 +399,12 @@ object BlockAPI {
                         startBlockNum - depth,
                         Some(startBlockNum)
                       )
-        lfb   <- casper.lastFinalizedBlock
-        graph <- visualizer(topoSortDag, PrettyPrinter.buildString(lfb.blockHash))
-      } yield serialize(graph).asRight[Error]
+        lfb <- casper.lastFinalizedBlock
+        g   <- visualizer(topoSortDag, PrettyPrinter.buildString(lfb.blockHash))
+      } yield g
+      Sync[F].delay(serialize(graph).asRight[Error])
+    }
+
     EngineCell[F].read >>= (_.withCasper[ApiErr[R]](
       casperResponse(_),
       Log[F]
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -385,11 +385,11 @@ object BlockAPI {
       maxDepthLimit: Int,
       startBlockNumber: Int,
       visualizer: (Vector[Vector[BlockHash]], String) => F[Graphz[F]],
-      serialize: F[Graphz[F]] => R
+      serialize: Graphz[F] => R
   ): F[ApiErr[R]] = {
     val errorMessage = "visual dag failed"
-    def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[R]] = {
-      val graph = for {
+    def casperResponse(implicit casper: MultiParentCasper[F]): F[ApiErr[R]] =
+      for {
         dag <- MultiParentCasper[F].blockDag
         // the default startBlockNumber is 0
         // if the startBlockNumber is 0 , it would use the latestBlockNumber for backward compatible
@@ -399,11 +399,9 @@ object BlockAPI {
                         startBlockNum - depth,
                         Some(startBlockNum)
                       )
-        lfb <- casper.lastFinalizedBlock
-        g   <- visualizer(topoSortDag, PrettyPrinter.buildString(lfb.blockHash))
-      } yield g
-      Sync[F].delay(serialize(graph).asRight[Error])
-    }
+        lfb   <- casper.lastFinalizedBlock
+        graph <- visualizer(topoSortDag, PrettyPrinter.buildString(lfb.blockHash))
+      } yield serialize(graph).asRight[Error]
 
     EngineCell[F].read >>= (_.withCasper[ApiErr[R]](
       casperResponse(_),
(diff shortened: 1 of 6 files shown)


zsluedem (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
```diff
@@ -23,90 +24,87 @@ object GraphzGenerator {
...
 
-  final case class DagInfo[G[_]](
+  final case class DagInfo(
       validators: Map[String, ValidatorsBlocks] = Map.empty,
       timeseries: List[Long] = List.empty
   )
 
   object DagInfo {
-    def empty[G[_]]: DagInfo[G] = DagInfo[G]()
+    def empty: DagInfo = DagInfo()
```
tgrospic: Defining default values directly on a class makes it harder to use and read when instance is created. So we should avoid doing that and instead create separate functions like _empty_ here which will create instance with defaults or define their own default parameters.

```scala
def empty: DagInfo = DagInfo(validators = Map.empty, timeseries = List.empty)
```

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -379,18 +379,17 @@ object BlockAPI {
...
-      G[_]: Monad: GraphSerializer,
       R
   ](
       depth: Int,
       maxDepthLimit: Int,
       startBlockNumber: Int,
-      visualizer: (Vector[Vector[BlockHash]], String) => F[G[Graphz[G]]],
-      serialize: G[Graphz[G]] => R
+      visualizer: (Vector[Vector[BlockHash]], String) => F[Graphz[F]],
+      serialize: F[Graphz[F]] => R
```
tgrospic: I think we can simplify serialize function to `serialize: Graphz[F] => R`.

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -1,104 +1,104 @@
...
 import cats._
-import cats.effect.Sync
+import cats.effect.{Concurrent, Sync}
+import cats.effect.concurrent.Ref
 import cats.syntax.all._
-import cats.mtl._
 
 trait GraphSerializer[F[_]] {
   def push(str: String, suffix: String = "\n"): F[Unit]
+  def show: F[String]
```
tgrospic: The point of this trait is to abstract over logic how final result is constructed by calling `push` method.
This means that we don't want to have methods like `show` because whoever is creating the instance will have full control to collect all and get the final result.

This definition of `show` also implies that we must have String as a result which is doesn't have to be the case when writing to as file or list of string lines.

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -1,104 +1,104 @@
...
+case object Dotted extends GraphStyle
+case object Dashed extends GraphStyle
 
 sealed trait GraphArrowType
-final case object NormalArrow extends GraphArrowType
-final case object InvArrow    extends GraphArrowType
-final case object NoneArrow   extends GraphArrowType
+case object NormalArrow extends GraphArrowType
+case object InvArrow    extends GraphArrowType
+case object NoneArrow   extends GraphArrowType
```
tgrospic, line 64: Why to remove _final_ for case classes?

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
```diff
@@ -1,104 +1,104 @@
...
-      case Msquare      => "Msquare"
-      case Record       => "record"
-    }
+  implicit val showShape: Show[GraphShape] = {
+    case Circle       => "circle"
+    case DoubleCircle => "doublecircle"
+    case Box          => "box"
+    case PlainText    => "plaintext"
+    case Msquare      => "Msquare"
+    case Record       => "record"
```
tgrospic: Although this is simplified definition it has one downside. IntelliJ will not find this implementation of `Show[GraphShape]` when searching for instances.

casper/src/main/scala/coop/rchain/casper/api/GraphGenerator.scala
stanislavlyalin: Resolved in [commit](https://github.com/rchain/rchain/pull/3574/commits/280fdf4f0422d1a3c443b9eb27b675842fc6a472). @tgrospic, can you explain please why it is harder to use and read when the defaults are provided directly in class definition?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
stanislavlyalin: Resolved in [commit](https://github.com/rchain/rchain/pull/3574/commits/3731710108b1717020f2b884491db78dfe7bc0ad)

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
stanislavlyalin: Reverted in [commit](https://github.com/rchain/rchain/pull/3574/commits/a175818472f857fbeb97f59e15ce527b7c89ad92). It was IDE warning `'final' modifier is redundant for toplevel objects`

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
stanislavlyalin: Reverted in [commit](https://github.com/rchain/rchain/pull/3574/commits/a175818472f857fbeb97f59e15ce527b7c89ad92)

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
stanislavlyalin: Fixed in [commit](https://github.com/rchain/rchain/pull/3574/commits/fac58f356148e6c47d8188344c1b6364a931c3f5).

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
tgrospic: It seems that `final` modifier is added by default for all `object` definitions and possibly forbidden in the future versions.
https://github.com/scala/bug/issues/11094

I'm OK that we are not using final for objects and remove it on existing code.

It makes definitions with objects and classes little strange but this is Scala. :)
```scala
sealed trait A
case object B extends A
final class C(a: Int) extends A
```

tgrospic (approved):
I think we can finalize this iteration of changes for Graphz. Very good work!

graphz/src/main/scala/coop/rchain/graphz/Graphz.scala
stanislavlyalin: `final` keyword was removed in [commit](https://github.com/rchain/rchain/pull/3574/commits/dc1aff6d36a7fe2140b1795dbd8d0ca256fc68f2) to make the code clear.

stanislavlyalin:
bors merge

stanislavlyalin:
bors merge

PR #3577, opened 2022-01-05
Use unforgeable name as mergeable channel tag

### LINKED ISSUE #3513: REV vaults (number channel) merging

### LINKED ISSUE #3575: Implement the safe way to tag mergeable number channels
## Overview

@tgrospic introduce a new way to resolving conflicts with mergeable channels on numbers in https://github.com/rchain/rchain/pull/3492. But currently the codes is using a "__MERGEABLE__" string to 
tag these mergeable channels which is not safe at all

The issue is intended to explore a proper way to fix that.

1. replace __MERGEABLE__ with an unforgeable name
2. add config for mergeable tag to runtime

### Related issue
- https://github.com/rchain/rchain/issues/3513

zsluedem:
## Overview

Create an unforgeable name to tag number channels. Now the code is using fixes string to do that which is not safe at all and it is exposed to the user.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
+++ b/node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
@@ -59,7 +59,7 @@ object StateBalanceMain {
   // The way to get this Unforgeable name needs reporting casper to get all the concrete comm events.
   // Anyway, as long as RevVault.rho and genesis doesn't change, this value would be fixed.
   val genesisVaultMapPar: Par = GPrivate(
-    "2dea7213bab34c96d4d3271f9b9dcb0b50a242de7de2177da615d9de23106afd".unsafeHexToByteString
+    "5e8200d74ba8689013cf86f075775a18a42f1bd3a5ea122e82dfca9ba485a924".unsafeHexToByteString
   )
 
   @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements"))
--- a/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
@@ -440,7 +440,8 @@ object RhoRuntime {
       invalidBlocks: InvalidBlocks[F],
       extraSystemProcesses: Seq[Definition[F]],
       urnMap: Map[String, Par],
-      mergeChs: Ref[F, Set[Par]]
+      mergeChs: Ref[F, Set[Par]],
+      mergeableTagName: Par
   ): Reduce[F] = {
     lazy val replayDispatchTable: RhoDispatchMap[F] =
       dispatchTableCreator(
@@ -456,7 +457,8 @@ object RhoRuntime {
         chargingRSpace,
         replayDispatchTable,
         urnMap,
-        mergeChs
+        mergeChs,
+        mergeableTagName
       )
     replayReducer
   }
@@ -478,7 +480,8 @@ object RhoRuntime {
   def createRhoEnv[F[_]: Concurrent: Parallel: _cost: Log: Metrics: Span](
       rspace: RhoISpace[F],
       mergeChs: Ref[F, Set[Par]],
-      extraSystemProcesses: Seq[Definition[F]] = Seq.empty
+      extraSystemProcesses: Seq[Definition[F]] = Seq.empty,
+      mergeableTagName: Par
   ): F[(Reduce[F], Ref[F, BlockData], InvalidBlocks[F])] =
     for {
       mapsAndRefs                                     <- setupMapsAndRefs(extraSystemProcesses)
@@ -489,7 +492,8 @@ object RhoRuntime {
         invalidBlocks,
         extraSystemProcesses,
         urnMap,
-        mergeChs
+        mergeChs,
+        mergeableTagName
       )
       res <- introduceSystemProcesses(rspace :: Nil, procDefs.toList)
       _   = assert(res.forall(_.isEmpty))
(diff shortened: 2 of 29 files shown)


rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
```diff
@@ -67,7 +68,7 @@ object RholangAndScalaDispatcher {
       urnMap: Map[String, Par]
   ): (Dispatch[M, ListParWithRandom, TaggedContinuation], Reduce[M]) = {
     val emptyMergeableRef = Ref.unsafe[M, Set[Par]](Set.empty)
-
-    apply(tuplespace, dispatchTable, urnMap, emptyMergeableRef)
+    val dummyMergeableTag = Par()
```
tgrospic: Why to use dummy mergeable tag name here?

zsluedem: For some tests who don't care about mergeable tag name. Is it better to move it to test package?

node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
```diff
@@ -59,7 +59,7 @@ object StateBalanceMain {
   // The way to get this Unforgeable name needs reporting casper to get all the concrete comm events.
   // Anyway, as long as RevVault.rho and genesis doesn't change, this value would be fixed.
   val genesisVaultMapPar: Par = GPrivate(
-    "2dea7213bab34c96d4d3271f9b9dcb0b50a242de7de2177da615d9de23106afd".unsafeHexToByteString
+    "5e8200d74ba8689013cf86f075775a18a42f1bd3a5ea122e82dfca9ba485a924".unsafeHexToByteString
```
tgrospic, line 62: Do we have a test to check this hard coded value?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntime.scala
```diff
@@ -478,7 +480,8 @@ object RhoRuntime {
   def createRhoEnv[F[_]: Concurrent: Parallel: _cost: Log: Metrics: Span](
       rspace: RhoISpace[F],
       mergeChs: Ref[F, Set[Par]],
-      extraSystemProcesses: Seq[Definition[F]] = Seq.empty
+      extraSystemProcesses: Seq[Definition[F]] = Seq.empty,
+      mergeableTagName: Par
```
tgrospic: It's more flexible to put parameters with defaults at the end.

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
tgrospic: Yes, that would be better.

Also keep in mind that reducer will detect any tuple with this form `(MergeableTag, _)`, so with `Par()` as a mergeable tag this means any tuple `(Nil, _)` will be considered mergeable.
What I mean is that dummy tag must be something not possible to construct in testing Rholang code.

zsluedem: Good point. How about `Par(unforgeables=something)`?

node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
zsluedem: Yes. https://github.com/rchain/rchain/blob/dev/node/src/test/scala/coop/rchain/node/revvaultexport/VaultBalanceGetterTest.scala

rholang/src/main/scala/coop/rchain/rholang/interpreter/dispatch.scala
zsluedem: I moved this `apply` to test package since it is only used in tests.

tgrospic: > Good point. How about `Par(unforgeables=something)`?

Yes, in that way we can create name that are not possible to be created in any other way. Like in definition of system processes.
https://github.com/rchain/rchain/blob/76f14c516bac3d6be2f91318c9b202b4239465bb/rholang/src/main/scala/coop/rchain/rholang/interpreter/SystemProcesses.scala#L76-L81

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala
```diff
@@ -22,24 +22,25 @@ final class RhoRuntimeOps[F[_]: Sync](
...
 
   def evaluate(
       term: String,
       initialPhlo: Cost,
-      normalizerEnv: Map[String, Par]
+      normalizerEnv: Map[String, Par],
+      randSeed: Blake2b512Random
   ): F[EvaluateResult] = {
-    implicit val rand: Blake2b512Random = Blake2b512Random(128)
+    implicit val rand: Blake2b512Random = randSeed
```
tgrospic: Do we need this change here? We can make it on dev branch to reduce conflict on merging feature branch.

zsluedem: https://github.com/rchain/rchain/blob/dev/crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala#L137-L141
This is not a pure function. I can not predict unforgeable name in test without this change.

zsluedem: Well. My bad. This change is not neccessary .

tgrospic (approved):
Nice work!

PR #3578, opened 2022-01-10
Prevent read-only node to accept deploys

### LINKED ISSUE #3266: Read-only node should not accept deploys
Deploy to read-only should return an error.

### What did you see instead?
Read-only node accepted deploy as it was the validator.

Deploy to test net read-only node.
```
curl https://observer.testnet.rchain.coop/api/deploy \
  --data-binary '{"data":{"term":"Nil","timestamp":1607687297250,"phloPrice":1,"phloLimit":250000,"validAfterBlockNumber":18},"sigAlgorithm":"secp256k1","signature":"3045022100b8a9a6176fd9b8f394bafcad987524996d25102eaa83560e1da578ff444d40db022047e66f7d074740d51df429e84a16cfb9a4223eee225fa83c053c03ed70207d12","deployer":"04eb79a155592889626968233a30d3cae7c1d9ce664017ccc5ebf627b053d28f44158ad47b2c47bad4a49d88c568951a9dd7597b2aa612a2374070400c5499e22c"}'
```
Returned result is successful.
```
"Success!\nDeployId is: 3045022100b8a9a6176fd9b8f394bafcad987524996d25102eaa83560e1da578ff444d40db022047e66f7d074740d51df429e84a16cfb9a4223eee225fa83c053c03ed70207d12"
```

:heavy_plus_sign: It would be nice for checking if node is a validator is more precise than just checking if private key is provided in the configuration.

## Solution

If private key is supplied when running a node it's more complicated to check if this private key is bonded and node can produce blocks, so this task should be done in two steps.

- [x] When private key is NOT supplied and node operates as read-only node. Throw an error when deploy is received.
- [ ] When private key is supplied, check if private key is bonded and node is a validator which can create blocks and accept deploys. Throw an error if not bonded when deploy is received.
  - There is also edge case when a new validator _waiting_ for the next epoch where it will be bonded and start producing blocks. In that case it would be useful that in some amount of block before epoch change it can accept deploys.
  - This situation can also happen when validator is slashed and still have private key but cannot produce new blocks.

Another option would be for read-only node to _relay_ deploys to a validator node. In that case deployer must have a way to check deploy results and status. FYI @zsluedem

stanislavlyalin:
## Overview
If the node is running in read-only mode (without validator private key) it's rejecting deploys. Exception with text below are thrown
```
Deploy was rejected because node is running in read-only mode.```
This text is visible for client and shown in the log.

### Checking

1. Run node with command
```shell
./rnode run -s
```
Without `--validator-private-key` rnode runs in read-only mode.
2. Make deploy to this node. Deploy will be rejected with the message (see above).

### Notes
* Is it a beautiful solution to pass additional parameter `isNodeReadOnly`? Maybe better to pass the whole `conf` instance?
* Should `isNodeReadOnly` parameter have default value?

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -53,7 +53,8 @@ object BlockAPI {
   def deploy[F[_]: Concurrent: EngineCell: Log: Span](
       d: Signed[DeployData],
       triggerPropose: Option[ProposeFunction[F]],
-      minPhloPrice: Long
+      minPhloPrice: Long,
+      isNodeReadOnly: Boolean
   ): F[ApiErr[String]] = Span[F].trace(DeploySource) {
 
     def casperDeploy(casper: MultiParentCasper[F]): F[ApiErr[String]] =
@@ -69,6 +70,13 @@ object BlockAPI {
         // call a propose if proposer defined
         _ <- triggerPropose.traverse(_(casper, true))
       } yield r
+
+    // Check if node is read-only
+    val readOnlyError = new RuntimeException(
+      s"Deploy was rejected because node is running in read-only mode. " +
+        s"To be a validator node should be run with a validator private key."
+    ).raiseError[F, ApiErr[String]]
+    val readOnlyCheck = readOnlyError.whenA(isNodeReadOnly)
 
     // Check if deploy is signed with system keys
     val isForbiddenKey = StandardDeploys.systemPublicKeys.contains(d.pk)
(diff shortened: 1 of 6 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -70,6 +71,13 @@ object BlockAPI {
         _ <- triggerPropose.traverse(_(casper, true))
       } yield r
 
+    // Check if node is read-only
+    val readOnlyError = new RuntimeException(
+      s"Deploy was rejected because node is running in read-only mode. " +
+        s"To be a validator node should be run with a validator private key."
```
tgrospic: This error message is also returned to the client via API so only the first sentence makes sense.

With API rewrite we will distinguish debug messages for the log and friendly messages returned to the end user.

stanislavlyalin: Yes, I agree. Fixed.

tgrospic (approved):
Very nice work!

zsluedem (approved):
LGTM

stanislavlyalin:
bors merge

PR #2888, opened 2020-02-18
RCHAIN-4032: disable receives on the same channel

tgrospic:
## Overview
Because current implementation of [SpaceMatcher](https://github.com/rchain/rchain/blob/dcd5ee6/rspace/src/main/scala/coop/rchain/rspace/SpaceMatcher.scala#L72-L104) does not work well with receives on the same channels, we need to disallow these kind of joins `for (<- @2; <- @2)`.

### JIRA ticket:
https://rchain.atlassian.net/browse/RCHAIN-4032

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3579, opened 2022-01-12
Replaying cost match tests

zsluedem:
## Overview

Previously we disable replay cost tests because we  got some problems on consume on the same channel https://github.com/rchain/rchain/pull/3579/files#diff-a8950e05083ebb606328a9eb162794f4d131f122e08a886f90c911e77969fd82L252 . But we fixed that in https://github.com/rchain/rchain/pull/2888.
So it would be better to re-enable these tests.

### DIFF
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
@@ -253,7 +253,7 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
   "replay the same term with same event logs" should "get the same cost" in {
     val r = scala.util.Random
     for (_ <- 1 to 100) {
-      val long     = ((r.nextLong % 0X144000000L) + 0X144000000L) % 0X144000000L
+      val long     = math.abs(r.nextLong)
       val contract = fromLong(long)
       if (contract != "") {
         val result = evaluateAndReplay(Cost(Integer.MAX_VALUE), contract)
@@ -267,7 +267,7 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
   "replay the same term with not enough phlo" should "get the same result" in {
     val r = scala.util.Random
     for (_ <- 1 to 100) {
-      val long        = ((r.nextLong % 0X144000000L) + 0X144000000L) % 0X144000000L
+      val long        = math.abs(r.nextLong)
       val contract    = fromLong(long)
       val parsingCost = accounting.parsingCost(contract)
       if (contract != "") {
--- a/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
+++ b/rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
@@ -171,7 +171,7 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
             remainder /= 4
             joins += f"_ ${arrow} @${ch}"
           }
-          val joinStr = joins.mkString("; ")
+          val joinStr = joins.toSet.mkString("; ")
           result += f"for (${joinStr}) { 0 }"
           nonlinearRecv ||= (arrow == "<=")
         }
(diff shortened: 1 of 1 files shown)


rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -263,6 +264,22 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
     }
   }
 
+  "replay the same term with not enough phlo" should "get the same result" in {
+    val r = scala.util.Random
+    for (_ <- 1 to 100) {
+      val long        = ((r.nextLong % 0X144000000L) + 0X144000000L) % 0X144000000L
```
nzpr: I'm curious what this is doing?

nzpr (approved):
LGTM, minor question

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: To be very honest, I don't know. :) Maybe it is trying to generate a number with some kind of range.

zsluedem: Hm. I think it is useless. I would just remove it.

zsluedem: After https://github.com/rchain/rchain/runs/4932655784?check_suite_focus=true, I think these codes are for the `r` is negative situation.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -249,24 +250,10 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
       }
     }
 
-  // TODO: Remove ignore when bug RCHAIN-3917 is fixed.
-  it should "be repeatable when generated" ignore {
+  "replay the same term with same event logs" should "get the same cost" in {
```
tgrospic, line 253: We don't want to remove existing test, it's testing bigger scope of Rholang terms.

Please add this as an additional test which is testing much smaller scope.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -171,7 +172,7 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
             remainder /= 4
             joins += f"_ ${arrow} @${ch}"
           }
-          val joinStr = joins.mkString("; ")
+          val joinStr = joins.toSet.mkString("; ")
```
tgrospic, line 175: This function is generator of Rholang terms. Each index number corresponds to different terms so conversion to Set means to make some of the indexes equal.

What is the reason for this change?

zsluedem: https://rchain.atlassian.net/browse/RCHAIN-3917?
Rholang doesn't support it.
Same joins means `consuming two messages from the same channel`

tgrospic: > Every number gets decoded into a unique term

If we change internal behavior of this function, every number will not produce a unique term.
So to get terms without joins on the same channel these numbers must be skipped.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: No tests are removed . Why did you say that? I just rephrased it.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: Could you elaborate?
What specific do you want?

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
```diff
@@ -277,6 +264,22 @@ class CostAccountingSpec extends FlatSpec with Matchers with PropertyChecks with
     }
   }
 
+  "replay the same term with not enough phlo" should "get the same result" in {
+    val r = scala.util.Random
+    for (_ <- 1 to 100) {
```
tgrospic, line 269: Here are many tests removed because only 100 terms are considered. Each number represents a unique term so reducing the number of iterations means skipping tests for all these Rholang terms.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
tgrospic: If you want to skip some Rholang terms it's better to find the range of numbers which generates these terms and not used them in tests.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
tgrospic: I've added comment in the next test to explain.

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: Is it realistic to do that? Is it necessary? 
What is the shortcoming of the current solution? 
And how to get those numbers? Iterate all numbers and hardcode exclusions here?

rholang/src/test/scala/coop/rchain/rholang/interpreter/accounting/CostAccountingSpec.scala
zsluedem: If you look at the generated terms, they are almost identical with little changes. :)

zsluedem: Not neccessary to do all for 10000 times to make the test run much longer.

tgrospic: Yes, that's the point of this test, to check all possible combinations with very small deviations. So I'd say that running first 100 is not necessary at all because these basic terms are already covered by Rholang existing tests.

Maybe we can add another generator with terms without same channels in joins. But in a way to not just collapse joins on the same channels because that means we will generate and run the same terms for different input numbers.

PR #3584, opened 2022-01-20
Implemented HTTP binding for `getDataAtPar` method

### LINKED ISSUE #3538: Implement replace API for listenForDataAtName

stanislavlyalin:
## Overview
Added support for method `getDataAtPar` from WebAPI. This method allows getting deploy results more effectively than `getDataAtName` because doesn't use the `depth` parameter.

### How to check solution locally

1. Run RNode locally as validator (with `--validator-private-key` and other necessary parameters)
2. Make deploy with code below and store deploy signature
```
new return(`rho:rchain:deployId`) in { return!("data-at-par") }
```
3. Make propose and store returned block hash
4. Make POST request on `http://localhost:40403/api/data-at-name-by-block-hash` with JSON data
```json
{
    "blockHash": "<ADDED_IN_PROPOSE_BLOCK_HASH>",
    "name": {
        "UnforgDeploy": {
            "data": "<DEPLOY_SIGNATURE>"
            }
        },
    "usePreStateHash": false
}
```

The result of this request should be like this
```json
{
    "expr": [
        {
            "ExprString": {
                "data": "data-at-par"
            }
        }
    ],
    "block": {
        "blockHash": "2a0ef3da0b3d929fb9dada1a94a4bebd216dcdf1e474e28c831e417a6a957fb5",
        "sender": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
        "seqNum": 17,
        "sig": "3045022100f2c2e6f44075ae5097f51d2fae62dbfea76869f66bd0de8404df56fd0c46fc40022063b105ef8eddd86c5669d5215be2d688955de68bb205cf3d4466ac5566247cb8",
        "sigAlgorithm": "secp256k1",
        "shardId": "root",
        "extraBytes": "",
        "version": 1,
        "timestamp": 1643023442303,
        "headerExtraBytes": "",
        "parentsHashList": [
            "7a5024910e00c581c692ae52fdb9fe56407dad2989f7a14d7d2fb42abf4f8615"
        ],
        "blockNumber": 17,
        "preStateHash": "7fbb7ad62521ab64be0e54285d64f9dd4329d0b34a889052347bded047a271e5",
        "postStateHash": "4060c2837da16b736d5ca6ebba95fe13a78424c06f0b6be143fce765fa122530",
        "bodyExtraBytes": "",
        "bonds": [
            {
                "validator": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
                "stake": 150500000000000
            }
        ],
        "blockSize": "127859",
        "deployCount": 1,
        "faultTolerance": 1.0,
        "justifications": [
            {
                "validator": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
                "latestBlockHash": "7a5024910e00c581c692ae52fdb9fe56407dad2989f7a14d7d2fb42abf4f8615"
            }
        ],
        "rejectedDeploys": []
    }
}
```

**The sample output of `data-at-name` for compare**
```json
{
    "exprs": [
        {
            "expr": {
                "ExprString": {
                    "data": "data-at-name"
                }
            },
            "block": {
                "blockHash": "8c3ca0d7a0f500873408591df94e3fc872350f64aa94ebb34341faa2a3b39716",
                "sender": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
                "seqNum": 14,
                "sig": "304402204d1cc30afbfb3c4add72edee0d150fe2681271e1b63c729253e5eb9ef3ab5bd2022067a34bfc5e2e26a75caf243270aa1bcf8f603f8bfa58708fa99ffe83c33b63cf",
                "sigAlgorithm": "secp256k1",
                "shardId": "root",
                "extraBytes": "",
                "version": 1,
                "timestamp": 1642606004650,
                "headerExtraBytes": "",
                "parentsHashList": [
                    "dd4a3d1a0cb2571aa5cb86b4eded05125afb11c921e255185fd9ef1109ac57b9"
                ],
                "blockNumber": 14,
                "preStateHash": "594f4ee2bb7cb943f6e4ca66e26e8934bf98313929077442256efe2fdbc76c01",
                "postStateHash": "ccbcf5bee2c76731789dfed322ad11df2633ca5fadc33e8eef7497d0544436cb",
                "bodyExtraBytes": "",
                "bonds": [
                    {
                        "validator": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
                        "stake": 150500000000000
                    }
                ],
                "blockSize": "127861",
                "deployCount": 1,
                "faultTolerance": 1.0,
                "justifications": [
                    {
                        "validator": "04415d35d799325f7784d7fbae15a3c4d0a24ff97429c7512b877b34bf3b7c801d220c38877e9a8a0736d168000ac5139cf505d9c030d1b7714dce4128ca3771b3",
                        "latestBlockHash": "dd4a3d1a0cb2571aa5cb86b4eded05125afb11c921e255185fd9ef1109ac57b9"
                    }
                ],
                "rejectedDeploys": []
            }
        }
    ],
    "length": 1
}
```

### Notes
1. Is there way to combine functions https://github.com/rchain/rchain/blob/729c72dacc6e0f9855ffb5ab94fcb668fbc0d55d/node/src/main/scala/coop/rchain/node/api/WebApi.scala#L373-L377 with providing generic type with type bounds?
2. Is it necessary requirement?
> Response should be the same as for exploratory deploy (Seq[Par], LightBlockInfo).

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -801,17 +801,28 @@ object BlockAPI {
     )
   }
 
-  def getDataAtPar[F[_]: Concurrent: RuntimeManager: BlockStore](
+  def getDataAtPar[F[_]: Concurrent: EngineCell: Log: SafetyOracle: BlockStore](
       blockHash: String,
       par: Par,
       usePreStateHash: Boolean
-  ): F[Seq[Par]] =
-    for {
-      block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
-      stateHash = if (usePreStateHash) block.body.state.preStateHash
-      else block.body.state.postStateHash
-      sortedPar <- parSortable.sortMatch[F](par).map(_.term)
-      data      <- RuntimeManager[F].getData(stateHash)(sortedPar)
-    } yield data
+  ): F[(Seq[Par], LightBlockInfo)] = {
+
+    def casperResponse(
+        implicit casper: MultiParentCasper[F]
+    ): F[(Seq[Par], LightBlockInfo)] =
+      for {
+        block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
+        stateHash = if (usePreStateHash) block.body.state.preStateHash
+        else block.body.state.postStateHash
+        sortedPar      <- parSortable.sortMatch[F](par).map(_.term)
+        runtimeManager <- casper.getRuntimeManager
+        data           <- getDataWithBlockInfo(runtimeManager, sortedPar, block).map(_.get)
+      } yield (data.postBlockData, data.getBlock)
+
+    EngineCell[F].read >>= (_.withCasper[(Seq[Par], LightBlockInfo)](
+      casperResponse(_),
+      (Seq.empty[Par], LightBlockInfo()).pure
+    ))
+  }
 
 }
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -811,9 +811,7 @@ object BlockAPI {
         implicit casper: MultiParentCasper[F]
     ): F[ApiErr[(Seq[Par], LightBlockInfo)]] =
       for {
-        block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
-        stateHash = if (usePreStateHash) block.body.state.preStateHash
-        else block.body.state.postStateHash
+        block          <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
         sortedPar      <- parSortable.sortMatch[F](par).map(_.term)
         runtimeManager <- casper.getRuntimeManager
         data           <- getDataWithBlockInfo(runtimeManager, sortedPar, block).map(_.get)
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -802,14 +802,14 @@ object BlockAPI {
   }
 
   def getDataAtPar[F[_]: Concurrent: EngineCell: Log: SafetyOracle: BlockStore](
-      blockHash: String,
       par: Par,
+      blockHash: String,
       usePreStateHash: Boolean
-  ): F[(Seq[Par], LightBlockInfo)] = {
+  ): F[ApiErr[(Seq[Par], LightBlockInfo)]] = {
 
     def casperResponse(
         implicit casper: MultiParentCasper[F]
-    ): F[(Seq[Par], LightBlockInfo)] =
+    ): F[ApiErr[(Seq[Par], LightBlockInfo)]] =
       for {
         block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
         stateHash = if (usePreStateHash) block.body.state.preStateHash
@@ -817,12 +817,17 @@ object BlockAPI {
         sortedPar      <- parSortable.sortMatch[F](par).map(_.term)
         runtimeManager <- casper.getRuntimeManager
         data           <- getDataWithBlockInfo(runtimeManager, sortedPar, block).map(_.get)
-      } yield (data.postBlockData, data.getBlock)
+      } yield (data.postBlockData, data.getBlock).asRight[Error]
 
-    EngineCell[F].read >>= (_.withCasper[(Seq[Par], LightBlockInfo)](
-      casperResponse(_),
-      (Seq.empty[Par], LightBlockInfo()).pure
-    ))
+    val errorMessage =
+      "Could not get data at par, casper instance was not available yet."
+
+    EngineCell[F].read >>= (
+      _.withCasper[ApiErr[(Seq[Par], LightBlockInfo)]](
+        casperResponse(_),
+        Log[F].warn(errorMessage).as(s"Error: $errorMessage".asLeft)
+      )
+    )
   }
 
 }
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -11,6 +11,7 @@ import coop.rchain.casper.api.BlockAPI.LatestBlockMessageError
 import coop.rchain.casper.blocks.proposer.ProposerResult
 import coop.rchain.casper.engine.EngineCell.EngineCell
 import coop.rchain.casper.protocol.{BlockInfo, DataWithBlockInfo, DeployData, LightBlockInfo}
+import coop.rchain.casper.util.rholang.RuntimeManager
 import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.{SignaturesAlg, Signed}
 import coop.rchain.metrics.Span
@@ -36,7 +37,9 @@ trait WebApi[F[_]] {
   def deploy(request: DeployRequest): F[String]
 
   // Read data (listen)
-  def listenForDataAtName(request: DataRequest): F[DataResponse]
+  def listenForDataAtName(request: DataAtNameRequest): F[DataAtNameResponse]
+
+  def listenForDataAtPar(request: DataAtParRequest): F[DataAtParResponse]
 
   // Blocks info
 
@@ -63,7 +66,7 @@ trait WebApi[F[_]] {
 
 object WebApi {
 
-  class WebApiImpl[F[_]: Sync: RPConfAsk: ConnectionsCell: NodeDiscovery: Concurrent: EngineCell: Log: Span: SafetyOracle: BlockStore](
+  class WebApiImpl[F[_]: Sync: RPConfAsk: ConnectionsCell: NodeDiscovery: Concurrent: RuntimeManager: EngineCell: Log: Span: SafetyOracle: BlockStore](
       apiMaxBlocksLimit: Int,
       devMode: Boolean = false,
       cacheTransactionAPI: CacheTransactionAPI[F],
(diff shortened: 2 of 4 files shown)


node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -318,9 +318,10 @@ object Setup {
       )
       cacheTransactionAPI <- Transaction.cacheTransactionAPI(transactionAPI, rnodeStoreManager)
       webApi = {
-        implicit val (ec, bs, or, sp) = (engineCell, blockStore, oracle, span)
-        implicit val (ra, rc)         = (rpConfAsk, rpConnections)
-        val isNodeReadOnly            = conf.casper.validatorPrivateKey.isEmpty
+        implicit val (ec, bs, or, sp)      = (engineCell, blockStore, oracle, span)
+        implicit val (ra, rc)              = (rpConfAsk, rpConnections)
+        val isNodeReadOnly                 = conf.casper.validatorPrivateKey.isEmpty
+        implicit val rm: RuntimeManager[F] = runtimeManager
```
zsluedem: I am ok with both implicit or explicit parameter here.(Personally, I am more fond of explicit here)

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -193,19 +201,29 @@ object WebApi {
...
+  )
+
+  final case class DataAtNameResponse(
       exprs: List[RhoExprWithBlock],
       length: Int
   )
 
+  final case class DataAtParResponse(
+      pars: Seq[Par]
+  )
```
zsluedem: I am ok to return without `BlockInfo` here.

zsluedem (approved):
LGTM

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -193,19 +201,29 @@ object WebApi {
...
+      usePreStateHash: Boolean
+  )
+
+  final case class DataAtNameResponse(
       exprs: List[RhoExprWithBlock],
       length: Int
   )
 
+  final case class DataAtParResponse(
+      pars: Seq[Par]
```
tgrospic: Existing WebAPI methods returns its own representation of `Par` type which is `RhoExpr`. So new method should return `Seq[RhoExpr]` and `LightBlockInfo` object for the block on which data is returned.

If user didn't specify block hash as input, last finalized block will be used and LightBlockInfo will give user info for what state he is getting the data.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
tgrospic: If user didn't specify specific block he will not know from which state is the returned data. But I'd agree that in the future we can reduce amount data about block.

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
```diff
@@ -139,7 +141,10 @@ object WebApiRoutes {
       // Get data
 
       case req @ POST -> Root / "data-at-name" =>
-        req.handle[DataRequest, DataResponse](webApi.listenForDataAtName)
+        req.handle[DataAtNameRequest, DataAtNameResponse](webApi.listenForDataAtName)
+
+      case req @ POST -> Root / "data-at-par" =>
```
tgrospic: I don't like the name `data-at-par`. It's wrong because we accept only unforgeable name (RhoUnforg) and Par has no direct connection to WebAPI types.

Some suggestion I have in mind:
- data-at-name-block
- data-at-name-v2
- data-at-name-direct
- get-name-data
- get-channel-data
- data-at-channel

@zsluedem, @nzpr

_We will unify API interface and types for gRPC and HTTP so I don't consider this name as final for API v2._

node/src/main/scala/coop/rchain/node/api/WebApi.scala
tgrospic: We should reuse `ExploratoryDeployResponse` type which has the same structure. But renaming it to `RhoDataResponse`.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -36,7 +37,9 @@ trait WebApi[F[_]] {
   def deploy(request: DeployRequest): F[String]
 
   // Read data (listen)
-  def listenForDataAtName(request: DataRequest): F[DataResponse]
+  def listenForDataAtName(request: DataAtNameRequest): F[DataAtNameResponse]
+
+  def listenForDataAtPar(request: DataAtParRequest): F[DataAtParResponse]
```
tgrospic: Let's just call it `getDataAtPar` as in BlockAPI to easily distinguish from old method (which is misleading with prefix _listen_).

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
stanislavlyalin: Resolved in [commit](https://github.com/rchain/rchain/pull/3584/commits/0912026303b68c5a6c203b33e6a2cb9151c05d5c).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: I introduced this case class for response as duplicate of `ExploratoryDeployResponse`
https://github.com/rchain/rchain/blob/02edbaa462e7ee0f1c1dda1200e95163be4497d6/node/src/main/scala/coop/rchain/node/api/WebApi.scala#L245-L248

I couldn't find a way, how to add an alias for `ExploratoryDeployResponse`.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: Done in [commit](https://github.com/rchain/rchain/pull/3584/commits/02edbaa462e7ee0f1c1dda1200e95163be4497d6).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -97,11 +101,27 @@ object WebApi {
...
+        .map(toDataAtNameResponse)
+
+    def getDataAtPar(req: DataAtParRequest): F[RhoDataResponse] = {
+      implicit val rm: RuntimeManager[F] = runtimeManager
+
+      val deployId = req.name match {
+        case unforg: UnforgPrivate  => unforg.data
+        case unforg: UnforgDeploy   => unforg.data
+        case unforg: UnforgDeployer => unforg.data
+      }
```
tgrospic: `UnforgDeploy` is a name (channel) that holds deploy signature and it's a special name different from `UnforgDeploy` and `UnforgPrivate`.
Here we are searching data that is stored on each of this names, so they must be supplied to `getDataAtPar` as received from WebAPI. BlockAPI should return the block info in the same way as `BlockAPI.getListeningNameDataResponse`.

stanislavlyalin: Resolved in [commit](https://github.com/rchain/rchain/pull/3584/commits/e06ff930cb4847f5b07ad3dbf54928028a3881db)

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -801,17 +801,28 @@ object BlockAPI {
...
       usePreStateHash: Boolean
-  ): F[Seq[Par]] =
-    for {
-      block <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
-      stateHash = if (usePreStateHash) block.body.state.preStateHash
-      else block.body.state.postStateHash
-      sortedPar <- parSortable.sortMatch[F](par).map(_.term)
-      data      <- RuntimeManager[F].getData(stateHash)(sortedPar)
-    } yield data
+  ): F[(Seq[Par], LightBlockInfo)] = {
```
tgrospic: Let `Par` parameter be at the first place, it's the main data required for this method.

Also when Casper is not available we should return (and log) an error as in other methods.

stanislavlyalin: Done [here](https://github.com/rchain/rchain/pull/3584/commits/77d380dce27161918dced6a81956c0e6ec0354bb)

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -107,9 +107,12 @@ object WebApi {
...
 
     def getDataAtPar(req: DataAtParRequest): F[RhoDataResponse] =
-      BlockAPI.getDataAtPar(req.blockHash, toPar(req), req.usePreStateHash).map {
-        case (pars, block) => RhoDataResponse(pars.flatMap(exprFromParProto), block)
-      }
+      BlockAPI
+        .getDataAtPar(toPar(req), req.blockHash, req.usePreStateHash)
+        .flatMap(_.liftToBlockApiErr)
+        .map {
+          case (pars, block) => RhoDataResponse(pars.flatMap(exprFromParProto), block)
```
tgrospic: Please use `toExploratoryResponse` (renamed to `toRhoDataResponse`) because it has the same signature.

stanislavlyalin: Yes, thanks for your attentiveness! Done in [commit](https://github.com/rchain/rchain/pull/3584/commits/14d1f75f5a7c150b7a864319b7939e71d2c8d7fa).

tgrospic:
bors try

tgrospic:
bors ping

tgrospic:
bors try

node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala
tgrospic: Let's use the same pattern as exploratory deploy and call the method `data-at-name-by-block-hash`. And the parameters are the same except term/name field.

stanislavlyalin: @tgrospic, done in [commit](https://github.com/rchain/rchain/pull/3584/commits/a243fb16cf1305804a6530f57653831fc4fca122). Is parameter `usePreStateHash` needed? It doesn't used in `getDataAtPar`.

tgrospic (approved):
Very nice work!

stanislavlyalin:
bors merge

PR #3591, opened 2022-01-28
New reject rule for overflow and negative merged results

### LINKED ISSUE #3513: REV vaults (number channel) merging

zsluedem:
## Overview

 Implement rejections for merged number channels which are negative or overflow. as what https://github.com/rchain/rchain/issues/3513 asked for

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
@@ -67,24 +67,31 @@ object ConflictSetMerger {
       }
     }
 
+    def foldRejection(baseBalance: Map[Blake2b256Hash, Long], branches: Set[Branch]) = {
+      val (_, rejected) = branches.foldLeft((baseBalance, Set.empty[Branch])) {
+        case ((balances, rejected), deploy) =>
+          try {
+            (calMergedResult(deploy, balances), rejected)
+          } catch {
+            case _: ArithmeticException => (balances, rejected + deploy)
+          }
+      }
+      rejected
+    }
+
     def getMergedResultRejection(
         branches: Set[Branch],
         rejectOptions: Set[Set[Branch]],
         base: Map[Blake2b256Hash, Long]
     ): Set[Set[Branch]] =
-      rejectOptions.map {
-        case normalRejectOptions =>
-          val (_, rejected) = (branches diff normalRejectOptions)
-            .foldLeft((base, Set.empty[Branch])) {
-              case ((balances, rejected), deploy) =>
-                try {
-                  (calMergedResult(deploy, balances), rejected)
-                } catch {
-                  case _: ArithmeticException => (balances, rejected + deploy)
-                }
-
-            }
-          rejected ++ normalRejectOptions
+      if (rejectOptions.isEmpty) {
+        Set(foldRejection(base, branches))
+      } else {
+        rejectOptions.map {
+          case normalRejectOptions =>
+            val rejected = foldRejection(base, branches diff normalRejectOptions)
+            rejected ++ normalRejectOptions
+        }
       }
 
     val (rejectedAsDependents, mergeSet) =
--- a/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
@@ -17,7 +17,6 @@ import coop.rchain.rspace.internal.Datum
 object ConflictSetMerger {
 
   /** R is a type for minimal rejection unit */
-  @SuppressWarnings(Array("org.wartremover.warts.NonUnitStatements", "org.wartremover.warts.Throw"))
   def merge[F[_]: Concurrent: Log, R: Ordering](
       actualSet: Set[R],
       lateSet: Set[R],
@@ -59,11 +58,8 @@ object ConflictSetMerger {
       diff.foldLeft(originResult) {
         case (ba, br) =>
           val result = Math.addExact(ba.getOrElse(br._1, 0L), br._2)
-          if (result < 0) {
-            throw new ArithmeticException("merged result negative")
-          } else {
-            ba.updated(br._1, result)
-          }
+          assert(result >= 0, "merged result negative")
+          ba.updated(br._1, result)
       }
     }
 
@@ -81,6 +77,7 @@ object ConflictSetMerger {
             (calMergedResult(deploy, balances), rejected)
           } catch {
             case _: ArithmeticException => (balances, rejected + deploy)
+            case _: AssertionError      => (balances, rejected + deploy)
           }
       }
       rejected
--- a/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
+++ b/casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
@@ -52,10 +52,10 @@ object ConflictSetMerger {
     }
 
     def calMergedResult(
-        branch: R,
+        branch: Branch,
         originResult: Map[Blake2b256Hash, Long]
     ): Map[Blake2b256Hash, Long] = {
-      val diff = mergeableChannels.apply(branch)
+      val diff = branch.map(mergeableChannels).toList.combineAll
       diff.foldLeft(originResult) {
         case (ba, br) =>
           val result = Math.addExact(ba.getOrElse(br._1, 0L), br._2)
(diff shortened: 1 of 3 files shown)


casper/src/test/scala/coop/rchain/casper/merging/MergeNumberChannelSpec.scala
```diff
@@ -236,6 +237,334 @@ class MergeNumberChannelSpec extends FlatSpec {
   implicit val logEff  = Log.log[Task]
   implicit val spanEff = Span.noop[Task]
 
+  def testNegativeMergeadResult[F[_]: Concurrent: ContextShift: Parallel: Span: Log] = {
```
zsluedem: Highly similar codes for test now. I would clean up later.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
```diff
@@ -65,15 +107,32 @@ object ConflictSetMerger {
     val rejectionTargetF = (dc: Branch) => dc.map(cost).sum
     val optimalRejection = getOptimalRejection(rejectionOptions, rejectionTargetF)
     val toMerge          = branches diff optimalRejection
-    val rejected         = lateSet ++ rejectedAsDependents ++ optimalRejection.flatten
 
     for {
-      r                               <- Stopwatch.duration(toMerge.toList.flatten.traverse(stateChanges).map(_.combineAll))
+      baseMergeableChRes <- toMerge.flatten
```
tgrospic: `toMerge` is already with excluded `optimalRejection`. Did you want to use `branches` instead?

casper/src/main/scala/coop/rchain/casper/merging/DagMerger.scala
```diff
@@ -47,28 +47,24 @@ object DagMerger {
...
-                      changes,
-                      baseReader,
-                      mergeableChs,
-                      overrideTrieAction
-                    )
-        } yield actions
-      }
+      historyReader <- historyRepository.getHistoryReader(lfbPostState)
+      baseReader    = historyReader.readerBinary
+      baseGetData   = baseReader.getData(_: Blake2b256Hash).map(_.map(_.decoded))
```
tgrospic: To simplify my own code. Without binary part getData can be used from historyReader directly.
`baseGetData  = historyReader.getData _`

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
```diff
@@ -65,15 +107,32 @@ object ConflictSetMerger {
...
+                                   .apply(channelHash)
+                                   .map(res => (channelHash, res.getOrElse(0L)))
+                             )
+                             .map(_.toMap)
+      rejectedOverflowNegative = getMergedResultRejection(
+        toMerge,
+        baseMergeableChRes,
+        cost
+      )
+      ultimateMerge                   = branches.flatten diff optimalRejection.flatten diff rejectedOverflowNegative
```
tgrospic: We need little more logic here. `optimalRejection` is from channel conflicts and must be combined with rejected by negative/overflow before remove from all branches.
The reason is that rejection in one group can influence in rejection in other. This means to again calculate similar combinations like in `computeRejectionOptions` for channel conflicts.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
zsluedem: Well. After some thought, I think it would make calculations simple to use toMerge. Otherwise, if `optimalRejection` is conflicted with `negativeOrOverflowRejection`, we have to recalculate `overflowOrNegativeRejections` from `toMerge`(or some other thing which I am not certain).

Let's say the most basic structure
```
          D2 (transfer 10 from A to B)                           D4 for(_ <- @"X") {Nil} | (transfer 20 from A to B)
          D1 (transfer 20 from A to B) | @"X"!(1)          D3 (transfer 60 from A to B)
              \                                                                       /
                       Base
```
User A,B,C,D got 100 balance at `Base`.
D1 is short for deploy1.
D1 and D2 are in the same block1(B1). And D3 and D4 are in the same block2(B2). And we are going to merge B1 and B2. 

Based on the comm event conflicting rejection rules, D4 and D1 are conflicted. Let's say `optimalRejection` is D4.
And based on the negative rejection rules, if we look at `branches`, D2 can be  be the `negativeRejctsion` which we can actually reject nothing at all here.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
```diff
@@ -44,6 +51,41 @@ object ConflictSetMerger {
...
+    def calMergedResult(
+        branch: R,
+        originResult: Map[Blake2b256Hash, Long]
+    ): Map[Blake2b256Hash, Long] = {
+      val diff = mergeableChannels.apply(branch)
+      diff.foldLeft(originResult) {
+        case (ba, br) =>
+          val result = Math.addExact(ba.getOrElse(br._1, 0L), br._2)
+          if (result < 0) {
+            throw new ArithmeticException("merged result negative")
```
zsluedem: I am in a dilemma whether I should introduce effect F here to introduce `ApplicativeError` and raiseError or just throw error.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
zsluedem: Could you elaborate  little bit how to `combine optimalRejection  with rejected by negative/overflow`?

zsluedem: I am thinking maybe we don't need `optimalRejection ` here but `rejectionOptions` here.

zsluedem: We might get better options in rejectOptions. But the combine logic is still something need to be polished.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
tgrospic: We don't want to add more WartRemover annotations and use them only to existing code.

Because this error is a _platform fatal_ error you can use assertions which will throw exceptions.

casper/src/test/scala/coop/rchain/casper/merging/MergeNumberChannelSpec.scala
```diff
@@ -71,28 +73,37 @@ class MergeNumberChannelSpec extends FlatSpec {
...
             evalResults <- terms.toList.traverse {
-                            case (term, _, _) =>
+                            case deploy =>
                               for {
-                                evalResult <- runtime.evaluate(term)
+                                evalResult <- runtime.evaluate(deploy.term)
                                 _ = assert(
                                   evalResult.errors.isEmpty,
-                                  s"${evalResult.errors}\n$term"
+                                  s"${evalResult.errors}\n$deploy.term"
```
tgrospic: Should this print `${deploy.term}`?

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
zsluedem: why do you consider that as `platform fatal error`? I am very curious.

tgrospic: It's fatal in a sense that is not recoverable. Node must exit and not handle the error.

tgrospic: Oh, I see you are handle the error to detect negative or overflow. Let's better to encode this in result type and return Option or Either.

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
```diff
@@ -44,6 +50,54 @@ object ConflictSetMerger {
...
 
+    def calMergedResult(
+        branch: Branch,
+        originResult: Map[Blake2b256Hash, Long]
+    ): Map[Blake2b256Hash, Long] = {
+      val diff = branch.map(mergeableChannels).toList.combineAll
+      diff.foldLeft(originResult) {
+        case (ba, br) =>
+          val result = Math.addExact(ba.getOrElse(br._1, 0L), br._2)
+          assert(result >= 0, "merged result negative")
```
tgrospic: Better to catch error inside this function and return result in Option.

zsluedem: I made some modifications. Please take a look

casper/src/main/scala/coop/rchain/casper/merging/ConflictSetMerger.scala
```diff
@@ -44,11 +50,65 @@ object ConflictSetMerger {
...
+    def foldRejection(baseBalance: Map[Blake2b256Hash, Long], branches: Set[Branch]) = {
+      val (_, rejected) = branches.foldLeft((baseBalance, Set.empty[Branch])) {
+        case ((balances, rejected), deploy) =>
+          // TODO come up with a better algorithm to solve below case
+          // currently we are accumulating result from some order and reject the deploy once negative result happens
+          // which doesn't seem perfect cases below
+          //
+          // base result 10 and folding the result from order like [-10, -1, 20]
+          // which on the second case `-1`, the calculation currently would reject it because the result turns
+          // into negative.However, if you look at the all the item view 10 - 10 -1 + 20 is not negative
```
tgrospic, line 86: As a simple solution can we sort by diffs, e.g. `20, -1, -10` so that increasing of balance is calculated first?

zsluedem: That sounds better!! Overflow should happen fewer than negative!

zsluedem: After some digging, I found for a branch the `NumberChannelsDiff` would always be `A: -10, B:+10`because of the nature of transferring which the sorting doesn't actually help too much.

tgrospic: Yes, there can be multiple transfers in a branch so logic of sorting is more complex. Let's investigate this as part of another PR.

tgrospic (approved):
LGTM

PR #3568, opened 2021-12-21
Radix history (scala implementation)

hilltracer:
## Overview
New Scala implementation of History. Goal: get a faster version than the current implementation.

### Description of Radix tree
   We have chosen a simple structure to implement our tree. We only have one type of node. Each of the nodes can contain from 1 to 256 refs. Each Ref can be of two types: Leaf is element for storing data; NodePtr is pointer to the next child node.
RadixValue stored in Leafs; RadixKey stored implicitly. The radixKey is determined by the location of the Leaf in the tree. The value is formed by connecting the Prefix and Refs assigned to the elements of the graph. These elements run from the root to the given Leaf.
Radix tree is the basic structure for new History
image

Benefits of the new Scala implementation: 
 - Parallel traversing algorithm, which allows to use several CPU at the same time;
 - Given the new tree structure less nodes have to be modified per each insert or delete;
 - A new serialization algorithm that is 15 times faster than current scodec serializer.

### Tests
Test conditions
 - system parameters: CPU - AMD Ryzen™ 7 3700U, RAM - 16GB DDR4, SSD - M.2 PCIe 3.0x2
 - Java heap size 2048MB
 - Average - 10 experiences
 - WarmUp - 10 first experiences
 - Data sets - random (Identical in all experiments)

image

image

image

image

image

image

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3602, opened 2022-02-15
Eager Casper

nzpr:
## Overview
Initial implementation of Casper operating on a local view of the message.


shared/src/main/scala/coop/rchain/casper/lcasper/FinalityView.scala
```diff
@@ -0,0 +1,80 @@
...
+    //  therefore the highest one should be chosen.
+    val highestCompleteLevelIdx = targetViews.map(_.completeLevelIdx).max
+    val views = targetViews.map {
+      case v @ FinalityView(levels) =>
+        // align all views to start from the same level
+        val diff = highestCompleteLevelIdx - v.completeLevelIdx
+        FinalityView(levels.drop(diff.toInt))
+    }
+    assert(
+      views.map(_.completeLevelIdx).toSet.size == 1,
```
tgrospic: Use `distinct` if you don't need Set instance.

It faster see https://github.com/rchain/rchain/pull/3568#discussion_r802092553.

shared/src/main/scala/coop/rchain/casper/lcasper/FinalityView.scala
```diff
@@ -0,0 +1,80 @@
...
+    )
+    // align views bu adding dummy records on top
+    val maxHeight = views.map(_.levelsViews.size).max
+    val aligned = views.map { v =>
+      val toFill  = maxHeight - v.levelsViews.size
+      val padding = (1 to toFill).map(_ => LevelView(-1, Vector.empty[Fringe[M, S]]))
+      v.copy(levelsViews = v.levelsViews ++ padding)
+    }
+
+    // reconcile top down
```
tgrospic: What top down means here? Should we reconcile from common fringe?

shared/src/main/scala/coop/rchain/casper/lcasper/LevelView.scala
```diff
@@ -0,0 +1,78 @@
+package coop.rchain.casper.lcasper
+import coop.rchain.casper.lcasper.Casper.{Fringe, LevelReconciliation}
+
+import scala.annotation.tailrec
+
+// View on DAG level
```
tgrospic: What do you mean by _DAG level_?

shared/src/main/scala/coop/rchain/casper/lcasper/LevelView.scala
```diff
@@ -0,0 +1,78 @@
...
+      bondsMap: Map[S, Long]
+  ): LevelReconciliation[M, S] = {
+    require(
+      views.map(_.height).distinct.size == 1,
+      "Unexpected error: reconciliation across different levels is detected."
+    )
+
+    final case class PastureState(
+        amoeba: Vector[(Fringe[M, S], Int)],          // indexed legs of amoeba
+        pasture: Vector[Vector[(Fringe[M, S], Int)]], // indexed pieces of food in the pasture
```
tgrospic: What is the meaning of index? Vector has index of elements and Int is in tuple, which index you are referring to?

shared/src/main/scala/coop/rchain/casper/lcasper/FinalityView.scala
nzpr: I think it is not relevant for this particular piece. Here we just collapse two view.
```scala
final case class LevelView[M, S](height: Long, partitions: Vector[Fringe[M, S]]) 
```
Level view consists of non overlapping partitions (partial fringes), inside which conflicts are already resolved. 
The outcome of reconciliation of two levels can be just placing one `partitions` next to another, or, if some partitions are  overlapping - placing one non overlapping  `partitions` next to another PLUS placing next to this new bigger partitions. 

After this, conflict resolution should be done from bottom up, as you've said.
New rejections done by lower level should be respected when doing resolution in higher levels. 
So `FinalityReconciliation` computed here should be input for conflict resolver, which then should go from the common fringe.

shared/src/main/scala/coop/rchain/casper/lcasper/LevelView.scala
nzpr: Block height (blockNum)

shared/src/main/scala/coop/rchain/casper/lcasper/LevelView.scala
nzpr: Input for this algo is list of level views, which are list of non overlapping partitions. 
So when partition Psmall from view V1 is absorbed by Pbig in V2 - we should remove absorbed partition from V1 and replace partition with the aggregate one in V2. To be able to do this we need to be able to remove item with some index from a vector and replace item by its index. So this index is used for this purpose.

shared/src/main/scala/coop/rchain/casper/lcasper/FinalityView.scala
nzpr: But I just thought that you are right. If we go bottom up, at some level we can decide that LevelView's cannot overlap anymore, and apply some special case to combine remaining views as non overlapping.

nzpr:
Closed in favour of https://github.com/rchain/rchain/pull/3606

PR #3128, opened 2020-09-15
document http api swagger experiment

zsluedem:
## Overview
autogenerated doc swagger

When the users access http://host:port/api/swagger-ui/, the user can see an page like https://developer.rchain.coop/rnode-api but only the swagger

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3612, opened 2022-02-24
Web API v1 (with OpenAPI schema)

### LINKED ISSUE #3536: Use another DSL package for http API
## Overview

In https://github.com/rchain/rchain/issues/2974 and https://github.com/rchain/rchain/pull/3128, we explore to use another [scala http DSL package](https://github.com/http4s/rho) which would generate open API automatically for http API.

However, we got [some problems](https://github.com/rchain/rchain/pull/3128#pullrequestreview-518783753) to generate the open API correctly. 

This is the example https://developer.rchain.coop/rnode-api#/ and hopefully we only need the API part for each node.

### Impact

Should not get any impact to the core part.

### Dependencies

https://github.com/http4s/rho

### Design

To goal is to implement new api with new dsl language in a new file and replace https://github.com/rchain/rchain/blob/dev/node/src/main/scala/coop/rchain/node/web/WebApiRoutes.scala in the future.

### LINKED ISSUE #3542: Develop new version of API (v2)
## Overview

Currently RNode API provides gRPC and HTTP protocol access. gRPC offers complete functionality and HTTP has main support for deploy and propose. The goal of the new `v2` is to provide more unified interface and improve user experience.

This issue here is aimed to gathering the information about implementing the new API.

- [x] #3532
  - #3543
- [x] #3538
  - #3544
  - #3584
  - #3592
- [x] #3536
  - #3612
- [ ] #3537
- [ ] #3541
- [ ] #3539

### Impact

For those clients who use the old API should migrate to the new API. Our release steps should work like below

```
             v2 API release
                   |--------------------------------------------------------->
                   |<----  both APIs exist  ------- |                             
Old API --------------------------------------------| Remove old API

                                   Timeline
```

### Related issues
#2974 - Open API compatibility

tgrospic:
## Overview

This PR adds a new Web API implementation with the support to generate OpenAPI schema. Circe serializer is also generated from JsonSchema and it has different format then the current implementation.

**Existing API which is accessible on `/api` endpoint will not be changed but may not be updated with the new features.**

Current format has field names as a discriminator for types.
```js
{ ExprInt: 42 }
```
New format which is more JS friendly.
```js
{ type: "ExprInt", data: 42 }
```
By default the OpenAPI schema is accessible on
http://localhost:40403/api/v1/openapi.json
and for admin API (+ propose)
http://localhost:40405/api/v1/openapi.json.

NOTE: This PR contains the whole solution but it's mission definition for all methods on Web API with corresponding derivations of JsonSchema especially for Rholang types generated from protobuf (Par). Preparation and helper functions are added and the rest should be done in a new PR.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/api/json/JsonSchemaDerivation.scala
+++ b/node/src/main/scala/coop/rchain/node/api/json/JsonSchemaDerivation.scala
@@ -1,0 +1,81 @@
+package coop.rchain.node.api.json
+
+import com.google.protobuf.ByteString
+import coop.rchain.casper.PrettyPrinter
+import coop.rchain.casper.protocol.{
+  BlockInfo,
+  BondInfo,
+  DeployData,
+  DeployInfo,
+  JustificationInfo,
+  LightBlockInfo,
+  RejectedDeployInfo
+}
+import coop.rchain.models.Par
+import coop.rchain.models.syntax._
+import coop.rchain.node.api.WebApi._
+import coop.rchain.node.web.TransactionInfo
+import endpoints4s.{Invalid, Valid}
+
+/**
+  * JsonSchema derivations for Web API request/response objects (Scala sealed traits and case classes)
+  */
+trait JsonSchemaDerivations extends JsonSchemaDerivationsBase {
+
+  // format: off
+
+  implicit lazy val deployDataSchema      : JsonSchema[DeployData]                   = schemaRecord
+  implicit lazy val deployRequestSchema   : JsonSchema[DeployRequest]                = schemaRecord
+  implicit lazy val versionInfoSchema     : JsonSchema[VersionInfo]                  = schemaRecord
+  implicit lazy val apiStatusSchema       : JsonSchema[ApiStatus]                    = schemaRecord
+  implicit lazy val exploreDeployReqSchema: JsonSchema[ExploreDeployRequest]         = schemaRecord
+  implicit lazy val dataAtNameReqSchema   : JsonSchema[DataAtNameByBlockHashRequest] = schemaRecord
+  implicit lazy val rhoResponseSchema     : JsonSchema[RhoDataResponse]              = schemaRecord
+  implicit lazy val bondInfoSchema        : JsonSchema[BondInfo]                     = schemaRecord
+  implicit lazy val justInfoSchema        : JsonSchema[JustificationInfo]            = schemaRecord
+  implicit lazy val rejectedInfoSchema    : JsonSchema[RejectedDeployInfo]           = schemaRecord
+  implicit lazy val lightBlockInfoSchema  : JsonSchema[LightBlockInfo]               = schemaRecord
+  implicit lazy val deployInfoSchema      : JsonSchema[DeployInfo]                   = schemaRecord
+  implicit lazy val blockInfoSchema       : JsonSchema[BlockInfo]                    = schemaRecord
+//  implicit lazy val transactionInfoSchema : JsonSchema[TransactionInfo]              = schemaRecord
+
+  // Web API Rholang types (subset of protobuf generated types)
+  implicit lazy val rhoExprSchema: JsonSchema[RhoExpr] =
+    lazySchema("RhoExprRef")(schemaTagged[RhoExpr]).withDescription("Rholang expression (Par type)")
+  // TODO: add ExprXXX types (not necessary for derivation but for short type name without a namespace)
+  implicit lazy val rhoUnforgSchema     : JsonSchema[RhoUnforg]      = schemaTagged
+  implicit lazy val unforgPrivateSchema : JsonSchema[UnforgPrivate]  = schemaRecord
+  implicit lazy val unforgDeploySchema  : JsonSchema[UnforgDeploy]   = schemaRecord
+  implicit lazy val unforgDeployerSchema: JsonSchema[UnforgDeployer] = schemaRecord
+
+  // Protobuf generated Rholang types (from models project)
+//  implicit lazy val parSchema: JsonSchema[Par] = lazySchema("ParRef", schemaRecord[Par])
+  /** TODO: add all referenced types in Par type, see [[coop.rchain.node.encode.JsonEncoder]] as example for circe derivations */
+
+  // format: on
+
+  // Json Schema encoding for ByteString as String base 16
+  implicit lazy val byteStringSchema: JsonSchema[ByteString] =
+    defaultStringJsonSchema.xmapPartial { str =>
+      val bytesOpt = str.hexToByteString
+      bytesOpt.map(Valid(_)).getOrElse(Invalid(s"Invalid hex format '$str'"))
+    }(PrettyPrinter.buildStringNoLimit)
+}
+
+/**
+  * Helpers JSON schema derivation functions with short type name without a namespace.
+  */
+trait JsonSchemaDerivationsBase extends endpoints4s.generic.JsonSchemas {
+  import scala.reflect.runtime.universe
+  import scala.reflect.runtime.universe._
+
+  def schemaRecord[T: TypeTag: GenericJsonSchema.GenericRecord]: JsonSchema[T] = {
+    val tpeName = universe.typeOf[T].typeSymbol.name.decodedName.toString
+    genericRecord[T].named(tpeName)
+  }
+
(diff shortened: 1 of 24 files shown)


node/src/main/scala/coop/rchain/node/api/json/JsonSchemaDerivation.scala
```diff
@@ -0,0 +1,81 @@
...
+    lazySchema("RhoExprRef")(schemaTagged[RhoExpr]).withDescription("Rholang expression (Par type)")
+  // TODO: add ExprXXX types (not necessary for derivation but for short type name without a namespace)
+  implicit lazy val rhoUnforgSchema     : JsonSchema[RhoUnforg]      = schemaTagged
+  implicit lazy val unforgPrivateSchema : JsonSchema[UnforgPrivate]  = schemaRecord
+  implicit lazy val unforgDeploySchema  : JsonSchema[UnforgDeploy]   = schemaRecord
+  implicit lazy val unforgDeployerSchema: JsonSchema[UnforgDeployer] = schemaRecord
+
+  // Protobuf generated Rholang types (from models project)
+//  implicit lazy val parSchema: JsonSchema[Par] = lazySchema("ParRef", schemaRecord[Par])
+  /** TODO: add all referenced types in Par type, see [[coop.rchain.node.encode.JsonEncoder]] as example for circe derivations */
```
tgrospic, line 53: @stanislavlyalin Here are missing types that needs to be defined so that the whole Web API can be implemented.

node/src/main/scala/coop/rchain/node/web/package.scala
```diff
@@ -1,71 +1,81 @@
...
-        "/status"    -> CORS(StatusInfo.service[F]),
-        "/ws/events" -> CORS(event),
-        "/api"       -> CORS(WebApiRoutes.service[F](webApiRoutes) <+> reportingRoutesOpt)
+        "/metrics"   -> corsPolicy(NewPrometheusReporter.service[F](prometheusReporter)),
+        "/version"   -> corsPolicy(VersionInfo.service[F]),
+        "/status"    -> corsPolicy(StatusInfo.service[F]),
+        "/ws/events" -> corsPolicy(event),
+        "/api"       -> corsPolicy(WebApiRoutes.service[F](webApi) <+> reportingRoutesOpt),
+        // Web API v1 with OpenAPI schema
+        "/api/v1" -> corsPolicy(WebApiRoutesV1.create[F](webApi))
```
tgrospic, line 46: I'm using prefix `v1` for this version instead of v2 because this is the first web API implementation which is done properly with schema generation.

@zsluedem, @dckc Does this make sense to you?

dckc: No. I am not aware of any requirement to use schema generation for an OpenAPI. And there are certainly clients that use v1.

dckc:
Is this in addition to the existing webAPI? If not, be sure to signal it as a BREAKING CHANGE and open issues on clients such as rvote.

dckc:
What does "more JS friendly" mean?

node/src/main/scala/coop/rchain/node/api/json/JsonSchemaDerivation.scala
stanislavlyalin: @tgrospic, can you explain, please. Do you mean JSON serialization/deserialization implicit vals for messages used in `Par`? E.g. `Send`, `Receive`, `New`... `JsonEncoder.scala` already has it.

tgrospic: Yes, we need definitions for JsonSchema for all types used in `Par`. JsonEncoder can be an example what types are necessary to define but definitions there cannot be reused for this version of API because circe encoders/decorders are generated from JsonSchema and have different format.

node/src/main/scala/coop/rchain/node/web/package.scala
tgrospic: @dckc Sorry, I was not clear enough about my question.

Existing API which is accessible on `/api` endpoint will not be changed, but may not be updated with the new features/methods.

The new API will have all the same methods at the start, but with auto generated OpenAPI schema accessible via API also.
So there will be no breaking change on existing API (except what will be necessary for block-merge, but this is not the subject of this PR).

The new version of API, because it's basically the same, it's exposed as `/api/v1` and not *v2* because later we will unify API schema for gRPC and HTTP and this will be **version 2** of API.

Does this make sense?

node/src/main/scala/coop/rchain/node/web/WebApiDocsV1.scala
```diff
@@ -0,0 +1,47 @@
...
+/**
+  * OpenAPI endpoint definition (GET /openapi.json).
+  */
+final case class WebApiDocServer[F[_]: Sync]()
+    extends Endpoints[F]
+    with server.JsonEntitiesFromEncodersAndDecoders {
+  implicit val jCodec: endpoints4s.Encoder[OpenApi, String] = OpenApi.stringEncoder
+
+  val publicRoutes = endpoint(get(path / "openapi.json"), ok(jsonResponse[OpenApi]))
+    .implementedBy(const(WebApiDocs.publicApi))
```
zsluedem, line 43: just curious about this `const`. Is it just marker or it can be used by compiler to optimize?

tgrospic:
> What does "more JS friendly" mean?

@dckc Types are encoded with discriminator field `"type"` which is more common in JS (maybe this is just my view!) then using type name as the name of the field.
E.g. in this way it's not possible to have situation with multiple _type_ fields defined in the same object.
`{ "ExprInt": 42, "ExprString": "42" }`

This is also default encoding for JSON schema implemented in libraries _endpoint4s_ (used in this PR) and _http4s/rho_ used in previous PR (https://github.com/rchain/rchain/pull/3128#pullrequestreview-518783753).

node/src/main/scala/coop/rchain/node/web/package.scala
zsluedem: Do you want to put the function of `v2` in `/api` directly in the future ?

node/src/main/scala/coop/rchain/node/web/https4s/RouterFix.scala
```diff
@@ -0,0 +1,61 @@
...
+  /**
+    * Defines an [[HttpRoutes]] based on list of mappings and
+    * a default Service to be used when none in the list match incoming requests.
+    *
+    * The mappings are processed in descending order (longest first) of prefix length.
+    */
+  def define[F[_]: Monad](
+      mappings: (String, HttpRoutes[F])*
+  )(default: HttpRoutes[F]): HttpRoutes[F] =
+    mappings.sortBy(_._1.length).foldLeft(default) {
```
zsluedem, line 35: Just curious. Why sort here?

tgrospic: This is direct copy from http4s source.
https://github.com/http4s/http4s/blob/519b72f3/server/shared/src/main/scala/org/http4s/server/Router.scala#L40

Mapping are sorted to match the longest first e.g. if mappings are defined as `/api` and `/api/v1`, request to `/api/v1/blocks` should not match the first (short) segment `/api` and fail to find rest of the route in this mapping. In other words mappings are like sub-roots so we want to find the longest that can be matched and then go deeper.

node/src/main/scala/coop/rchain/node/web/package.scala
tgrospic: No, we should slowly obsolete and remove existing `/api` and for version 2 create `/api/v2`. In this way users can prepare for new version of API at their own pace and we can maintain compatibility as long as possible.

node/src/main/scala/coop/rchain/node/web/WebApiDocsV1.scala
tgrospic: `Function.const` is just a built-in helper function to ignore the first argument. It's the same as:
```scala
implementedBy(_ => WebApiDocs.publicApi)
```
I'm not aware that compiler is doing some optimization.

zsluedem (approved):
LGTM

stanislavlyalin (approved):
Nice commits ordering and comments in code! LGTM

PR #3621, opened 2022-03-07
Update Scala to version 2.12.15

tgrospic:
## Overview

This PR is another update for Scala as preparation for Scala 2.13 and finally to Scala 3.


casper/src/main/scala/coop/rchain/casper/util/comm/ListenAtName.scala
```diff
@@ -25,21 +25,21 @@ object ListenAtName {
 
   private def buildPar[G[_]: BuildPar](name: G[Name]) = implicitly[BuildPar[G]].build(name)
 
-  implicit def buildParF[F[_]: MonadError[?[_], Throwable]] = new BuildPar[λ[A => F[Id[A]]]] {
+  implicit def buildParF[F[_]: MonadError[*[_], Throwable]] = new BuildPar[λ[A => F[Id[A]]]] {
```
zsluedem, line 28: What is the difference between `?` and `*`? 

I thought that it doesn't care which charactor it used.

project/Dependencies.scala
```diff
@@ -106,30 +108,29 @@ object Dependencies {
...
+    "com.github.jnr"         % "jnr-ffi"     % "2.1.15",
+    "com.lihaoyi"            %% "geny"       % "0.6.10",
+    "com.lihaoyi"            %% "sourcecode" % "0.2.1",
+    "org.scala-lang.modules" %% "scala-xml"  % "1.3.0",
+    "com.typesafe"           % "config"      % "1.4.0"
   )
 
-  private val kindProjector = compilerPlugin("org.typelevel" %% "kind-projector" % "0.10.3")
+  private val kindProjector = compilerPlugin(
+    "org.typelevel" %% "kind-projector" % "0.13.2" cross CrossVersion.full
```
stanislavlyalin, line 132: What does `cross CrossVersion.full` means? It is not clear for me from sbt documentation.

casper/src/main/scala/coop/rchain/casper/util/comm/ListenAtName.scala
stanislavlyalin: I guess, here is the [answer](https://github.com/scala/scala/pull/9560). But it is not clear for me. Look like `_` replaces with `?`, `?` replaces with `*` for a while, `*` replaces back to `_` again according to [the article](http://dotty.epfl.ch/docs/reference/changed-features/wildcards.html). What is the reason?

project/Dependencies.scala
```diff
@@ -8,6 +8,7 @@ object Dependencies {
   val catsEffectVersion = "2.5.4"
   val catsMtlVersion    = "0.7.1"
   val fs2Version        = "2.5.10"
+  val monixVersion      = "3.4.0"
```
stanislavlyalin, line 11: What is the reason to extract `monixVersion`? It is used only in one place.

zsluedem (approved):
LGTM. Left some questions.

Just curious. Maybe it would also make sense to upgrade to `2.13`? I think someone tries to do that before but failed.

models/src/main/scala/coop/rchain/models/ParMapTypeMapper.scala
```diff
@@ -11,7 +11,10 @@ object ParMapTypeMapper {
 
   private[models] def parMapToEMap(parMap: ParMap): EMap =
     EMap(
-      parMap.ps.sortedList.map(t => zip(t._1, t._2)),
+      // Convert to Vector because with Scala 0.12.5 deserialized empty List (Nil)
```
stanislavlyalin: Did you mean `2.12.5`? It is similar to RNode version :slightly_smiling_face:

tgrospic: Thanks! Yes, I meant Scala version 2.12.15. :joy:

project/Dependencies.scala
tgrospic: Just to make it more explicit as it's important library for us and version is related to cats effect version.

project/Dependencies.scala
tgrospic: Sbt will use this information to correctly download (create download urls) artifacts from maven repository.

tgrospic:
> LGTM. Left some questions.
> 
> Just curious. Maybe it would also make sense to upgrade to `2.13`? I think someone tries to do that before but failed.

@zsluedem Yes, multiple times. :) This https://github.com/tgrospic/rchain/pull/20 is my last attempt and I think I'm much closer now.

casper/src/main/scala/coop/rchain/casper/util/comm/ListenAtName.scala
tgrospic: I think Scala folks just want to make thing more interesting. :smiley:

I think the main reason is to support partial application of type level functions.
```scala
F[_, B](A)  =>  F[A, B]
F[A, _](B)  =>  F[A, B]
```

stanislavlyalin (approved):
LGTM :+1:

tgrospic:
bors merge

PR #3630, opened 2022-03-15
SDK project with initial Dag Manager abstraction

tgrospic:
## Overview

This PR adds `sdk` project which will serve as a base for domain logic for RNode with corresponding data types.

Initially Dag Manager is added with related interfaces and extensions (syntax). Implementation is added as a starting point for further work to wrap existing BlockDagStorage and BlockRetriever.
Dag Manager will serve as a main point to control block indexes and caching. It will support all that is needed for Casper and finalization logic or API.


rspace-bench/src/test/scala/coop/rchain/lmdb/LMDBStore.scala
```diff
@@ -1,13 +1,12 @@
...
-import java.nio.ByteBuffer
-
 import cats.effect.Sync
 import coop.rchain.shared.Resources.withResource
 import org.lmdbjava.{CursorIterable, Dbi, Env, Txn}
 import scodec.bits.BitVector
 
-import scala.collection.JavaConverters._
+import java.nio.ByteBuffer
+import scala.jdk.CollectionConverters._
```
tgrospic, line 9: This change is required in Scala 2.13.

zsluedem (approved):
LGTM

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagManager.scala
```diff
@@ -0,0 +1,43 @@
...
+  /**
+    * Thread safe function to update finalized nodes of the DAG seen by specific message.
+    */
+  def finalize(by: MId, finalized: Set[MId]): F[Unit]
+
+  /* Basic loading operations */
+
+  def loadMessage(mid: MId): F[M]
+
+  def loadSender(sid: SId): F[M]
```
nzpr: Should this return `F[S]`?

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagManager.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  /**
+    * Thread safe function to insert new message to the DAG.
+    */
+  def insert(msg: M): F[Unit]
+
+  /**
+    * Thread safe function to update finalized nodes of the DAG seen by specific message.
+    */
+  def finalize(by: MId, finalized: Set[MId]): F[Unit]
```
nzpr: I'm not sure I understand this. Do I understand it right that this is not a call to trigger finalization, but just to record finalization computed before?

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagView.scala
```diff
@@ -0,0 +1,32 @@
...
+  /**
+    * Traversal through DAG starting from latest messages.
+    */
+  def messages: Stream[F, Chunk[(M, List[M])]]
+
+  /* Basic loading operations */
+
+  def loadMessage(mid: MId): F[M]
+
+  def loadSender(sid: SId): F[M]
```
nzpr: Should this return F[S]?

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  def justifications(m: M): Set[MId]
+
+  def sender(m: M): SId
+
+  def bondsMap(m: M): Map[SId, Long]
+
+  /* Sender */
+
+  def stake(s: S): Long
```
nzpr: I'm thinking that this might be a bit misleading, as there is no notion of a stake for sender in general - sender has a stake only inside some state, and sender can have different stake in different states. 
And also you already have `def bondsMap(m: M): Map[SId, Long]`, which gives the stake of sender as per some message.

nzpr (changes requested):

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagManager.scala
tgrospic: I will change this to:
`def insert(msg: M, finalized: Set[MId], provisionallyFinalized: Boolean): F[Unit]`
So with the insert of a new message, new finalization can be recorded.

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagData.scala
tgrospic: Good point. I'll remove it.

sdk/src/main/scala/coop/rchain/sdk/dag/data/DagManager.scala
```diff
@@ -0,0 +1,38 @@
...
+  /**
+    * Thread safe function to insert new message to the DAG with corresponding finalized messages.
+    */
+  def insert(msg: M, finalized: Set[MId], provisionallyFinalized: Boolean): F[Unit]
+
+  /* Basic loading operations */
+
+  def loadMessage(mid: MId): F[M]
+
+  def loadSender(sid: SId): F[S]
```
tgrospic, line 32: Maybe we can be even more restrictive and say that you need to get the DagView to load M and S and remove these two methods from manager.

nzpr (approved):
LGTM

PR #3631, opened 2022-03-15
Add shard id to deploy data

### LINKED ISSUE #3628: Add shard ID to deploy data
## Overview

To prevent replay of deploys between shards we need to add shard identifier to deploy data signed by the deployer.

**NOTE: This change is incompatible with the validator block creation/validation and client libraries/dApps which are using deploys.**

https://github.com/rchain/rchain/blob/93d8907f28506b87306bb16c296d496cd9fe6c7b/models/src/main/protobuf/CasperMessage.proto#L139-L148

## Releases

This feature should be first applied to as an patch to `v0.12.5` used by RPC and later added as part of `v0.13.x` version. Existing client libraries (wallets) must be taken into consideration.

Also we need to decide how we want to roll out this change in regard to current main net. Should we continue with existing deploy data structure until Hard Fork 2 or apply it immediately.

## Solution

The new `shardId` field in DeployData type should be validated when deploy is submitted to the node and also when block is replayed.
`shardId` must be signed together with existing fields signed fields in DeployData type.

Validation of shardId should be done in BlockAPI and block validation.

https://github.com/rchain/rchain/blob/0367c41a2b2645ad4ec6d8faaed7d7486870f175/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L53-L105

In block validation shardId validation should be moved before `repeatDeploy` and before deploys shardId validation.

https://github.com/rchain/rchain/blob/45b83dfe614f30e4705fc1649a6266da3f4f9e95/casper/src/main/scala/coop/rchain/casper/Validate.scala#L160-L190

- [x] Create feature branch `feature/v0.12-rpc` based on `v0.12.5` tag to track special release for RPC
  https://github.com/rchain/rchain/tree/feature/v0.12-rpc
- [x] Create PR to `dev` branch to be ready for `0.13.x` version
  https://github.com/rchain/rchain/pull/3631
- [x] Create PR to `feature/v0.12-rpc` branch
  https://github.com/rchain/rchain/pull/3645

stanislavlyalin:
## Overview

This pull request implements a shard id check when creating and validating a block. To do this, the `shardId` field has been added to the deploydata message.

### Prerequirements

1. Run node as Validator from **clean state**
```
./rnode run -s --validator-private-key <VALIDATOR_PRIVATE_KEY> --wallets-file wallets.txt --bonds-file bonds.txt --shard-name banana
```
Be notified that `shard-name` CLI parameter is provided.

2. [pyrchain](https://github.com/rchain/pyrchain) library should be updated according to new `shardId` field in DeployData.

### From WebAPI

Make POST request to http://localhost:40403/api/deploy with JSON data. For example, from Python code
```python
# create_deploy_data is function from pyrchain library (https://github.com/rchain/pyrchain)
deploy_data = create_deploy_data(
    key=private,
    term=contract,
    phlo_price=phlo_price,
    phlo_limit=100000,
    valid_after_block_no=valid_after_block_no,
    timestamp_millis=int(time.time() * 1000),
    shard_id=shard_id)

json = {
    'data': {
        'term': rholang_program_as_string,
        'timestamp': deploy_data.timestamp,
        'phloPrice': deploy_data.phloPrice,
        'phloLimit': deploy_data.phloLimit,
        'validAfterBlockNumber': deploy_data.validAfterBlockNumber,
        'shardId': deploy_data.shardId
    },
    'sigAlgorithm': deploy_data.sigAlgorithm,
    'signature': base64.b16encode(deploy_data.sig),
    'deployer': base64.b16encode(deploy_data.deployer)
}
requests.post('http://localhost:40403/api/deploy', json=json)
```

If deploy's `shardId` is equals to Validator's `shardId` the result will be
```
Success!
DeployId is: 304402201d7fe1c3dd6c3294973f245710a9ff72f1cbace23b75f58c340f5be5f1c7b0be0220441973f374933017ee3cd41925aac223fe0afb8b33942d1a39a66d014bb1cec9
```
Otherwise
```
Deploy shardId 'test' is not as expected network shard 'banana'.
```

### From gRPC

Call `deploy` method, for example via pyrchain library
```python
from rchain.crypto import PrivateKey
from rchain.client import RClient

private = PrivateKey.from_hex('<HEX_STRING>')
client = RClient(host, 40401)
client.deploy(private, contract, 1, 100000, valid_after_block_no, int(time.time() * 1000), shard_id)
```

If deploy's `shardId` is equals to Validator's `shardId` the result will be
```
3045022100b3f75f466af51666c3480e0dfd2f9e091dd3e649f6d586312099964f8a905ef40220689671ae55b26e9e493d9b58a8f9bdd0ce45585acf73cc0b8616d733c670e65f
```
Otherwise
```
Deploy shardId 'test' is not as expected network shard 'banana'.
```

### From another node

Run another instance of RNode with command
```shell
./rnode deploy --phlo-limit 100000000 --phlo-price 1 --valid-after-block-number 10 --private-key <VALIDATOR_PRIVATE_KEY> --shard-id banana nil.rho
```
where `nil.rho` is the path to the file with rholang code to deploy.

If deploy's `shardId` is equals to Validator's `shardId` the result will be
```
Response: Success!
DeployId is: 304402200c2d024bbc4cf5b26e323ac4131462f22c3bb303dc7511294351db01966d5d5702203fedaeecbb1a3aa6388c865ddd3c1dc9e45c9a7348977d25775ebfcb45f08ef1
```
Otherwise
```
Deploy shardId 'test' is not as expected network shard 'banana'.
```

### Notes

If everything is OK, we should release a new stable version of [pyrchain](https://github.com/rchain/pyrchain) library (from `0.3.13a0` to `0.3.13`) and update the dependency in integration tests.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -54,7 +54,8 @@ object BlockAPI {
       d: Signed[DeployData],
       triggerPropose: Option[ProposeFunction[F]],
       minPhloPrice: Long,
-      isNodeReadOnly: Boolean
+      isNodeReadOnly: Boolean,
+      shardId: String
   ): F[ApiErr[String]] = Span[F].trace(DeploySource) {
 
     def casperDeploy(casper: MultiParentCasper[F]): F[ApiErr[String]] =
@@ -77,6 +78,12 @@ object BlockAPI {
     ).raiseError[F, ApiErr[String]]
     val readOnlyCheck = readOnlyError.whenA(isNodeReadOnly)
 
+    // Check if deploy's shardId equals to node shardId
+    val shardIdError = new RuntimeException(
+      "Deploy was rejected because its shard is not equal node shard"
+    ).raiseError[F, ApiErr[String]]
+    val shardIdCheck = shardIdError.whenA((d.data.shardId != shardId))
+
     // Check if deploy is signed with system keys
     val isForbiddenKey = StandardDeploys.systemPublicKeys.contains(d.pk)
     val forbiddenKeyError = new RuntimeException(
@@ -96,7 +103,7 @@ object BlockAPI {
       .warn(errorMessage)
       .as(s"Error: $errorMessage".asLeft[String])
 
-    readOnlyCheck >> forbiddenKeyCheck >> minPhloPriceCheck >> EngineCell[F].read >>= (_.withCasper[
+    readOnlyCheck >> shardIdCheck >> forbiddenKeyCheck >> minPhloPriceCheck >> EngineCell[F].read >>= (_.withCasper[
       ApiErr[String]
     ](
       casperDeploy,
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
@@ -80,9 +80,9 @@ object BlockAPI {
 
     // Check if deploy's shardId equals to node shardId
     val shardIdError = new RuntimeException(
-      "Deploy was rejected because its shard is not equal node shard"
+      s"Deploy shardId '${d.data.shardId}' is not as expected network shard '$shardId'."
     ).raiseError[F, ApiErr[String]]
-    val shardIdCheck = shardIdError.whenA((d.data.shardId != shardId))
+    val shardIdCheck = shardIdError.whenA(d.data.shardId != shardId)
 
     // Check if deploy is signed with system keys
     val isForbiddenKey = StandardDeploys.systemPublicKeys.contains(d.pk)
--- a/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
+++ b/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
@@ -446,7 +446,7 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
   /**
     * Evaluates exploratory (read-only) deploy
     */
-  def playExploratoryDeploy(term: String, hash: StateHash): F[Seq[Par]] = {
+  def playExploratoryDeploy(term: String, hash: StateHash, shardId: String): F[Seq[Par]] = {
     // Create a deploy with newly created private key
     val (privKey, _) = Secp256k1.newKeyPair
 
@@ -456,7 +456,8 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       timestamp = System.currentTimeMillis,
       // Hardcoded phlogiston limit / 1 REV if phloPrice=1
       phloLimit = 100 * 1000 * 1000,
-      sec = privKey
+      sec = privKey,
+      shardId = shardId
     )
 
     // Create return channel as first private name created in deploy term
@@ -558,8 +559,8 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
 
   /* Read only Rholang evaluator helpers */
 
-  def getActiveValidators(startHash: StateHash): F[Seq[Validator]] =
-    playExploratoryDeploy(activateValidatorQuerySource, startHash)
+  def getActiveValidators(startHash: StateHash, shardId: String): F[Seq[Validator]] =
+    playExploratoryDeploy(activateValidatorQuerySource, startHash, shardId)
       .ensureOr(
         validatorsPar =>
           new IllegalArgumentException(
@@ -569,9 +570,9 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       )(validatorsPar => validatorsPar.size == 1)
       .map(validatorsPar => toValidatorSeq(validatorsPar.head))
 
-  def computeBonds(hash: StateHash): F[Seq[Bond]] =
+  def computeBonds(hash: StateHash, shardId: String): F[Seq[Bond]] =
     // Create a deploy with newly created private key
-    playExploratoryDeploy(bondsQuerySource, hash)
+    playExploratoryDeploy(bondsQuerySource, hash, shardId)
       .ensureOr(
         bondsPar =>
           new IllegalArgumentException(
--- a/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
+++ b/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
@@ -456,8 +456,7 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       timestamp = System.currentTimeMillis,
       // Hardcoded phlogiston limit / 1 REV if phloPrice=1
       phloLimit = 100 * 1000 * 1000,
-      sec = privKey,
-      shardId = ""
+      sec = privKey
     )
 
     // Create return channel as first private name created in deploy term
--- a/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
+++ b/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
@@ -446,7 +446,7 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
   /**
     * Evaluates exploratory (read-only) deploy
     */
-  def playExploratoryDeploy(term: String, hash: StateHash, shardId: String): F[Seq[Par]] = {
+  def playExploratoryDeploy(term: String, hash: StateHash): F[Seq[Par]] = {
     // Create a deploy with newly created private key
     val (privKey, _) = Secp256k1.newKeyPair
 
@@ -457,7 +457,7 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       // Hardcoded phlogiston limit / 1 REV if phloPrice=1
       phloLimit = 100 * 1000 * 1000,
       sec = privKey,
-      shardId = shardId
+      shardId = ""
     )
 
     // Create return channel as first private name created in deploy term
@@ -559,8 +559,8 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
 
   /* Read only Rholang evaluator helpers */
 
-  def getActiveValidators(startHash: StateHash, shardId: String): F[Seq[Validator]] =
-    playExploratoryDeploy(activateValidatorQuerySource, startHash, shardId)
+  def getActiveValidators(startHash: StateHash): F[Seq[Validator]] =
+    playExploratoryDeploy(activateValidatorQuerySource, startHash)
       .ensureOr(
         validatorsPar =>
           new IllegalArgumentException(
@@ -570,9 +570,9 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       )(validatorsPar => validatorsPar.size == 1)
       .map(validatorsPar => toValidatorSeq(validatorsPar.head))
 
-  def computeBonds(hash: StateHash, shardId: String): F[Seq[Bond]] =
+  def computeBonds(hash: StateHash): F[Seq[Bond]] =
     // Create a deploy with newly created private key
-    playExploratoryDeploy(bondsQuerySource, hash, shardId)
+    playExploratoryDeploy(bondsQuerySource, hash)
       .ensureOr(
         bondsPar =>
           new IllegalArgumentException(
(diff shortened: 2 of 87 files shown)


casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
```diff
@@ -73,11 +74,11 @@ class MultiParentCasperAddBlockSpec extends FlatSpec with Matchers with Inspecto
 //  }
 
   it should "accept signed blocks" in effectTest {
-    TestNode.standaloneEff(genesis).use { node =>
+    TestNode.standaloneEff(genesis, shardId = SHARD_ID).use { node =>
```
tgrospic: Can TestNode read shardId from supplied `genesis` context (genesisBlock.shardId), instead of adding additional parameter?

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
```diff
@@ -569,9 +570,9 @@ final class RuntimeOps[F[_]: Sync: Span: Log](
       )(validatorsPar => validatorsPar.size == 1)
       .map(validatorsPar => toValidatorSeq(validatorsPar.head))
 
-  def computeBonds(hash: StateHash): F[Seq[Bond]] =
+  def computeBonds(hash: StateHash, shardId: String): F[Seq[Bond]] =
     // Create a deploy with newly created private key
-    playExploratoryDeploy(bondsQuerySource, hash)
+    playExploratoryDeploy(bondsQuerySource, hash, shardId)
```
tgrospic: When we are executing exploratory deploys and read only Rholang we don't really care about deploys, so you can just set shardId to empty string.

tgrospic: If this is the only reason to add parameter to RuntimeManager than we don't need it.

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
stanislavlyalin: > Can TestNode read shardId from supplied `genesis` context (genesisBlock.shardId), instead of adding additional parameter?

Yes, exactly. Thanks! Fixed in [commit](https://github.com/rchain/rchain/pull/3631/commits/ed13326b836563eb10384d9ad17b52ef2d00f878).

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeSyntax.scala
stanislavlyalin: > When we are executing exploratory deploys and read only Rholang we don't really care about deploys, so you can just set shardId to empty string.

Yes, I agree. I fixed it and related code in [commit](https://github.com/rchain/rchain/pull/3631/commits/6b3323cd4d0558d4385132b75632a09a1bcaad60).

stanislavlyalin: Exactly. I removed parameter in [commit](https://github.com/rchain/rchain/pull/3631/commits/6b3323cd4d0558d4385132b75632a09a1bcaad60).

stanislavlyalin:
@tgrospic, @zsluedem, to successfully execute integration tests [pyrchain](https://github.com/rchain/pyrchain) library should be updated because a new `DeployData` structure is needed. But it can be updated just after merging this PR into `dev` which is impossible of failed integration tests. How to better resolve this deadlock?

```
   integration tests     ->        pyrchain       ->       deploy to RNode

(proto-file updated   )      (installed via pip  )   (built with new proto-file     )
(but not merged to dev)      (with old proto-file)   (incompatible with old pyrchain)
```

**Upd**. I'll update `pyrchain` lib with `CasperMessages.proto` from this PR source. And after the release new version of `pyrchain` it will be possible to successfully execute integration tests.

models/src/test/scala/coop/rchain/casper/protocol/DeployDataSpec.scala
```diff
@@ -18,7 +20,8 @@ class DeployDataSpec extends PropSpec with GeneratorDrivenPropertyChecks with Ma
       timestamp,
       phloPrice,
       phloLimit,
-      validAfterBlockNumber
+      validAfterBlockNumber,
+      shardId = SHARD_ID
```
tgrospic: Property based tests uses generated data when running tests. So shard name should also define its arbitrary instance from which random strings will be generated.

node/src/test/scala/coop/rchain/node/mergeablity/MergeabilityRules.scala
```diff
@@ -296,4 +299,6 @@ trait BasicMergeabilityRules extends ComputeMerge {
         new TestFailedException(e, failedCodeStackDepth = 5).severedAtStackDepth
     }
   }
+
+  def shardId: String
```
tgrospic: Did you see this necessary? In mergeability tests we only care testing conflicts so shard name should be irrelevant.

models/src/test/scala/coop/rchain/casper/protocol/DeployDataSpec.scala
stanislavlyalin: Yes, thanks for the tip. I didn't know about arbitrary values for tests. [Fixed](https://github.com/rchain/rchain/pull/3631/commits/4d7828d8ba3c470b334682cd2eaacd29f5a855a7).

node/src/test/scala/coop/rchain/node/mergeablity/MergeabilityRules.scala
stanislavlyalin: I tried to provide a maximum configurable environment. I [replaced](https://github.com/rchain/rchain/pull/3631/commits/02b423807c0d09c741f18f593cf583d6989c5b59) the method with just an empty string since this is not important.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -77,6 +78,12 @@ object BlockAPI {
     ).raiseError[F, ApiErr[String]]
     val readOnlyCheck = readOnlyError.whenA(isNodeReadOnly)
 
+    // Check if deploy's shardId equals to node shardId
+    val shardIdError = new RuntimeException(
+      "Deploy was rejected because its shard is not equal node shard"
```
tgrospic: Please change the error message to include more info what it's expected and what is seen.

Something like
`Deploy shardId '${d.data.shardId}' is not as expected network shard '$shardId'.`

casper/src/main/scala/coop/rchain/casper/util/ConstructDeploy.scala
```diff
@@ -28,15 +28,17 @@ object ConstructDeploy {
       phloLimit: Long = 90000,
       phloPrice: Long = 1L,
       sec: PrivateKey = defaultSec,
-      vabn: Long = 0
+      vabn: Long = 0,
+      shardId: String
```
tgrospic: I think we can safely add empty string as default for shardId to reduce amount of changes in this PR. It will go along with other parameters which have default values.

This should apply for other functions in this file.

casper/src/slowcooker/scala/coop.rchain.casper/HashSetCasperSpecification.scala
```diff
@@ -65,8 +65,13 @@ object HashSetCasperActions {
       node.addBlock(signed)
     )
 
-  def deployment(i: Int, ts: Long = System.currentTimeMillis()): Signed[DeployData] =
-    ConstructDeploy.sourceDeploy(s"new x in { x!(0) }", ts)
+  def deployment(
+      i: Int,
+      ts: Long = System.currentTimeMillis()
+  ): Signed[DeployData] =
+    ConstructDeploy.sourceDeploy(s"new x in { x!(0) }", ts, shardId = SHARD_ID)
```
tgrospic: With default shardId in `ConstructDeploy.sourceDeploy` we don't these kind of changes, which is mostly in tests.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
stanislavlyalin: Yes, I agree. Resolved in [commit](https://github.com/rchain/rchain/pull/3631/commits/2e69f99787d9e3143f44082b9c4e0bfe2f9382bb).

casper/src/main/scala/coop/rchain/casper/util/ConstructDeploy.scala
stanislavlyalin: [Resolved](https://github.com/rchain/rchain/pull/3631/commits/ea6d84a5249a8431d0675caa816940daac7abffc).

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -145,6 +145,7 @@ message DeployDataProto {
   int64 phloPrice             = 7; //phlo price
   int64 phloLimit             = 8; //phlo limit for the deployment
   int64 validAfterBlockNumber = 10;
+  string shardId              = 11;//shard ID to prevent replay of deploys between shards
```
zsluedem, line 148: If we could use int64, we could possibly save some space.

zsluedem: nvm

integration-tests/test/rnode.py
```diff
@@ -512,7 +512,8 @@ def make_bootstrap_node(
     number_of_active_validators: int = 10,
     epoch_length: int = 10000,
     quarantine_length: int = 50000,
-    min_phlo_price: int = 1
+    min_phlo_price: int = 1,
+    shard_id: str = ''
```
tgrospic: It should not be possible to start the node with empty shardId. Do we have this check?

integration-tests/test/rnode.py
```diff
@@ -251,11 +251,11 @@ def eval(self, rho_file_path: str) -> str:
         return self.rnode_command('eval', rho_file_path)
 
     def deploy(self, rho_file_path: str, private_key: PrivateKey, phlo_limit:int = DEFAULT_PHLO_LIMIT,
-               phlo_price: int = DEFAULT_PHLO_PRICE, valid_after_block_no:int=0) -> str:
+               phlo_price: int = DEFAULT_PHLO_PRICE, valid_after_block_no:int=0, shard_id: str = '') -> str:
```
tgrospic: We should define some global constant for default shardId used in tests and just override when needed.

integration-tests/test/rnode.py
stanislavlyalin: @tgrospic, yes, we have this check https://github.com/rchain/rchain/blob/45b83dfe614f30e4705fc1649a6266da3f4f9e95/casper/src/main/scala/coop/rchain/casper/Validate.scala#L131

This is the reason why I provided `shard_id = 'test` in integration tests. I can remove the default empty string if needed.

integration-tests/test/rnode.py
stanislavlyalin: Done in [3b3c313](https://github.com/rchain/rchain/pull/3631/commits/3b3c313b9b064892f0463cbbaa89c454408ae07e) and [bb228f4](https://github.com/rchain/rchain/pull/3631/commits/bb228f475fdde33ccf63aba97b72f10096906322).

models/src/test/scala/coop/rchain/models/blockImplicits.scala
```diff
@@ -44,7 +44,7 @@ object blockImplicits {
     validator       = ByteString.copyFrom(byteArray.toArray)
   } yield Justification(validator, latestBlockHash)
 
-  val signedDeployDataGen: Gen[Signed[DeployData]] =
+  def signedDeployDataGen(shardId: String): Gen[Signed[DeployData]] =
```
tgrospic: ShardId should be also generated as other fields. It can be overwritten by the caller.
`shardId <- arbitrary[String]`

The same for other Gen functions in this file.

stanislavlyalin: [Fixed](https://github.com/rchain/rchain/pull/3631/commits/c384e9ba83a54e7e7bc910661c667c56f31e887c).

tgrospic (approved):
Excellent work!

zsluedem (approved):
LGTM.

stanislavlyalin:
bors merge

PR #3639, opened 2022-03-24
Fix multi-sig vault with public key authentication

### LINKED ISSUE #3625: Create pubKey configuration for POS multi-sig wallet
## Overview

Proof of Stake contract has hard-coded value for multi-sig vault used for slashed stake. Instead of hard-coded public keys and quorum size they should be supplied as configuration parameters when genesis block is created. 

Currently rnode use [pureconfig](https://pureconfig.github.io/docs/index.html) to read config.

Pureconfig support `HOCON`, json, java`.properties` format. They all support array config.

#### Example
HOCON
```
pubKeys = [
   pubKey1, pubKey2, pubKey3
]
```

Json
```
{"pubKeys":[pubKey1, pubKey2, pubKey3]}
```

[Jave .properties](https://help.mulesoft.com/s/question/0D52T00004mXUpxSAG/how-to-declare-array-in-properties-file)(Doesn't care actually, no one uses this)

### Design

We should introduce this config like [bond-maximum](https://github.com/rchain/rchain/blob/dev/node/src/main/resources/defaults.conf#L270) which would be rendered in the [POS contract](https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala#L10-L32)

- [x] Introduce POS multi-sig pub keys (pos-multi-sig-public-keys) in https://github.com/rchain/rchain/blob/dev/node/src/main/resources/defaults.conf#L250-L290
- [x] Pass these `pubKeys` configs in https://github.com/rchain/rchain/blob/dev/casper/src/main/scala/coop/rchain/casper/genesis/contracts/ProofOfStake.scala#L10-L32
- [x] Introduce `posMultiSigQuorum` (pos-multi-sig-quorum) as a config parameter which defines how many confirmations are necessary to use multi-sig vault
  - this number should be checked to be less or equal the number of supplied public keys
- [x] Replace the hard-coded https://github.com/rchain/rchain/blob/888c8040892/casper/src/main/resources/PoS.rhox#L110-L112 pub keys and quorum data into parsing config with `$$` like https://github.com/rchain/rchain/blob/888c8040892/casper/src/main/resources/PoS.rhox#L240
- [ ] Add CLI options, it seems that rogach has support for lists e.g. --pub abc def.
  https://github.com/scallop/scallop/wiki/Option-types-and-definitions#trailing-arguments

### Verification

1. Start fresh node with 2 public keys for PoS multi-sig vault and quorum 1
2. Run exploratory deploy to get multi-sig REV address
    ```scala
    new return, lookup(`rho:registry:lookup`) in {
      new posCh, multiSigVaultCh in {
        lookup!(`rho:rchain:pos`, *posCh) |
        for(@(_, pos) <- posCh) {
          @pos!("getCoopVault", *multiSigVaultCh) |
          for(@(_, revAaddress, _) <- multiSigVaultCh) {
            return!(revAaddress)
          }
        }
      }
    }
    ```
3. Transfer 100 REVs to multi-sig REV address
    - needed to transfer REVs from multi-sig
5. Transfer 100 REVs to the first multi-sig public key (REV address)
    - needed to pay for deploy
6. Create deploy to transfer from multi-sig address to any other address and check that transfer is successful
    - variables are `${toAddr}` for target address (String) and `${amount}` (Number)
    ```scala
      new out(\`rho:io:stdout\`)
        , deployId(\`rho:rchain:deployId\`), deployerId(\`rho:rchain:deployerId\`)
        , lookup(\`rho:registry:lookup\`)
        , posCh, msRevVaultCh, revVaultCh, posMultiSigVaultCh, deployerKeyCh, toRevVaultCh, transferRetCh
      in {
        lookup!(\`rho:rchain:pos\`, *posCh) |
        lookup!(\`rho:rchain:multiSigRevVault\`, *msRevVaultCh) |
        lookup!(\`rho:rchain:revVault\`, *revVaultCh) |
        for(@(_, Pos) <- posCh ; @(_, MultiSigVault) <- msRevVaultCh ; @(_, RevVault) <- revVaultCh) {
          @Pos!("getCoopVault", *posMultiSigVaultCh) |
          @MultiSigVault!("deployerAuthKey", *deployerId, *deployerKeyCh) |
          @RevVault!("findOrCreate", "${toAddr}", *toRevVaultCh) |
          for(@(msRevVault, msRevAaddress, _) <- posMultiSigVaultCh ; _ <- toRevVaultCh ; @deployerKey <- deployerKeyCh) {
            @msRevVault!("transfer", "${toAddr}", ${amount}, deployerKey, *transferRetCh) |
            for(@transferRet <- transferRetCh) {
              deployId!(("Transfer result", transferRet))
            }
          }
        }
      }
    ```

tgrospic:
## Overview

Currently _auth token_ is checked directly for equality which is not how AuthKey contract should be used making multi-sig unavailable for all defined public keys.
Tests of multi-sig with the use of public keys are not written.

This PR contains a fix to use `AuthKey("check")` method to verify and find public key used for authentication.

**TODO: Add tests for multi-sig with the use of public keys.**

### DIFF
--- a/casper/src/main/resources/MultiSigRevVault.rho
+++ b/casper/src/main/resources/MultiSigRevVault.rho
@@ -18,7 +18,6 @@
 new MultiSigRevVault,
     rs(`rho:registry:insertSigned:secp256k1`), uriOut,
     rl(`rho:registry:lookup`),
-    _fromPublicKey,
     _multiSigRevVault,
     RevVaultCh,
     AuthKeyCh,
@@ -56,21 +55,11 @@ in {
     } |
 
     contract MultiSigRevVault(@"deployerAuthKey", deployerId, ret) = {
-      new DeployerIdOps(`rho:rchain:deployerId:ops`),
-          revAddrCh,
-          deployerPubKeyBytesCh in {
+      new DeployerIdOps(`rho:rchain:deployerId:ops`), deployerPubKeyBytesCh in {
         DeployerIdOps!("pubKeyBytes", *deployerId, *deployerPubKeyBytesCh) |
-        for (@deployerPubKeyBytes <- deployerPubKeyBytesCh) {
-          _fromPublicKey!(deployerPubKeyBytes, *ret)
-        }
-      }
-    } |
-
-    contract _fromPublicKey(@publicKey, ret) = {
-      new RevAddress(`rho:rev:address`), revAddrCh in {
-        RevAddress!("fromPublicKey", publicKey, *revAddrCh)|
-        for (@revAddress <- revAddrCh) {
-          @AuthKey!("make", (*_multiSigRevVault, revAddress), *ret)
+        // Create auth token with deployer public key
+        for(@deployerPubKeyBytes <- deployerPubKeyBytesCh) {
+          @AuthKey!("make", (*_multiSigRevVault, deployerPubKeyBytes), *ret)
         }
       }
     } |
@@ -82,11 +71,10 @@ in {
           quorumSize <= 0) {
         ret!((false, "invalid quorum size"))
       } else {
-        new unf, authKeyCh, authPubKeysCh, RevAddress(`rho:rev:address`), revAddrCh in {
+        new unf, authKeyCh, RevAddress(`rho:rev:address`), revAddrCh in {
           RevAddress!("fromUnforgeable", *unf, *revAddrCh) |
           @RevVault!("unforgeableAuthKey", *unf, *authKeyCh) |
-          @ListOps!("map", publicKeys, *_fromPublicKey, *authPubKeysCh) |
-          for (@revVaultAuthKey <- authKeyCh & @authPubKeys <- authPubKeysCh & @revAddr <- revAddrCh) {
+          for (@revVaultAuthKey <- authKeyCh & @revAddr <- revAddrCh) {
             new multiSig, nonceStore, nonceMapStore, revVaultCh, verify in {
               @RevVault!("findOrCreate", revAddr, *revVaultCh) |
               for (@maybeRevVault <- revVaultCh) {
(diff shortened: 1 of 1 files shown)


casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -107,9 +95,27 @@ in {
                     } |
 
                     contract verify(@auth, idxAuthPubKeysCh, unsealedCh) = {
-                      new unsealerTest in {
-                        // Check whether auth is in the list of authPubKeys
-                        @ListOps!("indexOf", authPubKeys, auth, *idxAuthPubKeysCh) |
```
zsluedem, line 112: Just a question. The `equality` you mean is from this `indexOf` compare, right?

zsluedem (approved):
LGTM

nzpr (approved):
LGTM

casper/src/main/resources/MultiSigRevVault.rho
tgrospic: Yes, exactly. Auth token are generated from public keys inside multi-sig contract for this comparison. Because this is not how they work they will always be different (unforgeable name created inside AuthKey contract) and authentication for multi-sig will never succeed.

tgrospic:
bors merge

casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -56,21 +55,11 @@ in {
...
-          _fromPublicKey!(deployerPubKeyBytes, *ret)
-        }
-      }
-    } |
-
-    contract _fromPublicKey(@publicKey, ret) = {
-      new RevAddress(`rho:rev:address`), revAddrCh in {
-        RevAddress!("fromPublicKey", publicKey, *revAddrCh)|
-        for (@revAddress <- revAddrCh) {
-          @AuthKey!("make", (*_multiSigRevVault, revAddress), *ret)
```
nzpr, line 73: This was a bug as well, right? revAddress were used instead of pubkey

tgrospic: No, this was not part of the bug. Here we need to create secret for auth key that is checked to execute transaction. It can be REV address but it's not necessary so using public key is also valid.

PR #3643, opened 2022-03-25
Fix nonce confirmation to use public key

tgrospic:
## Overview

This PR contains additional commits to the fix of multi-sig vault to use public key in nonce confirmation instead of auth token.

- #3639

### DIFF
--- a/casper/src/main/resources/MultiSigRevVault.rho
+++ b/casper/src/main/resources/MultiSigRevVault.rho
@@ -94,10 +94,10 @@ in {
                       @revVault!("balance", *ret)
                     } |
 
-                    contract verify(@auth, idxAuthPubKeysCh, unsealedCh) = {
+                    contract verify(@auth, pubKeysCh, unsealedCh) = {
                       new unsealerTest, loopFindAny in {
                         // Check whether auth is valid to any public key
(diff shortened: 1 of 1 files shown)


casper/src/main/resources/MultiSigRevVault.rho
```diff
@@ -199,7 +199,7 @@ in {
                                     @(*multiSig, *nonceMapStore)!(nonceMap) |
                                     ret!((false, "given data does not match stored data"))
                                   } else {
-                                    if (((not unsealed) and nonceConfirmers.contains(auth)) or
```
zsluedem, line 202: Just curious. What is the side effect of using `auth`? 
Is it because one person could create several different auth token and make the proposal pass?

zsluedem (approved):
LGTM

casper/src/main/resources/MultiSigRevVault.rho
tgrospic: Auth key is created with the secret which is related to specific public key. So if the same deployer creates multiple auth keys they are all valid to create/confirm transfer, although they will be different unforgeable names.

nzpr (approved):
LGTM

tgrospic:
bors merge

PR #3648, opened 2022-03-29
Fix MultiSigVault boolean bug and add test cases

### LINKED ISSUE #3647: Multi-sig contract not working with public keys

When multi-sig contract is created with public keys it should be possible to use these keys to create/confirm transfers.

### What did you see instead?

Multi-sig contract returns an error with `invalid auth`̣ when used by the deployer which has its public key defined for multi-sig. 

1. Create multi-sig with at least one public key defined and at least minimum quorum size.
2. Create a deploy with the private key defined for multi-sig.
3. Observe an error `invalid auth` instead of transfer result.

More detailed steps now multi-sig can be used are specified as part of the issue for PoS config parameters.
- #3625.

### What version of RNode are you using?

v.0.12.5 (but it should be any version of the node)

### Problem

The issue is that AuthKey contract used in multi-sig contract is used in a wrong way so it's not possible to authenticate the deployer with corresponding public key defined when multi-sig vault is created.
Tests for multi-sig with public keys are not written so this issues is never detected. All multi-sig vaults created with public keys access are locked like the one defined in PoS for slashed funds owned by Coop.
https://github.com/rchain/rchain/blob/888c80408925d8c471928c920644959e43f52752/casper/src/main/resources/PoS.rhox#L105-L117

After solving the problem with AuthKey another problem showed up with multi-sig vault. When checking boolean expression which is executed eagerly exception is thrown for accessing index out of range because code is expecting for second argument of boolean operator not be executed (`unsealers.nth(info.nth(0))`).

https://github.com/rchain/rchain/blob/942c6ef4323edca016bd0c8a304a2e5c022d8563/casper/src/main/resources/MultiSigRevVault.rho#L202-L203

- #3362

### Solution

The fix for the use of AuthKey is done in these two PRs.
- #3639
- #3643

We need this fix immediately to start the RPC main net so PRs with patches should also be created for `v0.12.5` version.
- #3641

The epic to track all changes for RPC is tracked in this issue.
- #3642

To solve the issue with boolean operators Rholang bug should be fixed, but because we want to have RPC version ASAP Greg suggested that the solution should be creation of the special method which will handle the boolean logic that is needed for this purpose.

zsluedem:
## Overview

1. fix boolean bug in MultiSigVault
2. add pubkey multisig test cases
3. add pos multisig vault test case

### DIFF
--- a/rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
+++ b/rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
@@ -59,8 +59,8 @@ object CompiledRholangTemplate {
       case (content, (name, value)) => content.replace(s"""$$$$$name$$$$""", value.toString)
     }
 
-    s"""//Loaded from resource file <<$classpath>>
-       #$finalContent
+    s"""$finalContent
+       # //Loaded from resource file <<$classpath>>
        #""".stripMargin('#')
   }
 }
--- a/casper/src/main/resources/MultiSigRevVault.rho
+++ b/casper/src/main/resources/MultiSigRevVault.rho
@@ -200,7 +200,7 @@ in {
                                     ret!((false, "given data does not match stored data"))
                                   } else {
                                     if (((not unsealed) and nonceConfirmers.contains(pubKey)) or
-                                        (unsealed and nonceConfirmers.contains(unsealers.nth(info.nth(0))))) {
+                                        (unsealed && nonceConfirmers.contains(unsealers.nth(info.nth(0))))) {
                                       @(*multiSig, *nonceMapStore)!(nonceMap) |
                                       ret!((false, "already confirmed"))
                                     } else {
--- a/casper/src/main/scala/coop/rchain/casper/util/ConstructDeploy.scala
+++ b/casper/src/main/scala/coop/rchain/casper/util/ConstructDeploy.scala
@@ -47,12 +47,14 @@ object ConstructDeploy {
   def sourceDeployNow(
       source: String,
       sec: PrivateKey = defaultSec,
+      phloLimit: Long = 90000L,
       vabn: Long = 0,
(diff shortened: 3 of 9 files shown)


rholang/src/main/scala/coop/rchain/rholang/build/CompiledRholangSource.scala
```diff
@@ -59,8 +59,8 @@ object CompiledRholangTemplate {
       case (content, (name, value)) => content.replace(s"""$$$$$name$$$$""", value.toString)
     }
 
-    s"""//Loaded from resource file <<$classpath>>
-       #$finalContent
+    s"""$finalContent
+       # //Loaded from resource file <<$classpath>>
        #""".stripMargin('#')
```
zsluedem, line 64: put the added comment in the end because it would trouble on debugging the content lines.

tgrospic: What kind of troubles you had? Can you add additional comment at the end like:
`// End loaded from resource file <<$classpath>>`

zsluedem:
```scala
val finalContent = "@a!(!)"

val originalOutcome = """//Loaded from resource file <<$classpath>>
@a!(1)"""

val newOutcome =  """@a!(1)
//Loaded from resource file <<$classpath>>
"""
```
The above `originalOutcome` would give error `Top level free variables are not allowed: a at 2:2.` while the actual error line is `1:2` by `finalContent`

tgrospic (approved):
Excellent work!

hilltracer (approved):
I've tried to understand rho scripts (because it needs for a good review) but didn't have this understanding. So I simply run your tests with `and` and `&&` in MultiSigRevVault.rho. And it works fine.

zsluedem:
bors merge

PR #3650, opened 2022-03-31
BlockReceiverState implementation

stanislavlyalin:
## Overview

Implementation of thread-safe block receiver state. The state is the Map with block hashes as the key and statuses as the value. Thread-safety is provided by using [Ref](https://typelevel.org/cats-effect/docs/std/ref). This implementation does not store blocks in memory or on disk. Block can be in one of the following statuses:
* *Receiving* when we got a hash of a block we don't have.
* *Received* when the block is fully received.
* *Unvalidated* when the block is stored in BlockStore.
* *Validated* when the block is removed from CasperBuffer and ready to be added into DAG (BlockMetadataStore).

When the node is started, block hashes from the BlockMetadataStore with the *Validated* status and from the BlockStore with the *Unvalidated* status are added to the state.

### Notes

Tests for the new feature are not implemented.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
@@ -32,15 +32,13 @@ final case class BlockReceiverState[MId](state: Map[MId, RecvStatus]) {
 }
 
 object BlockReceiver {
-  def streams[F[_]: Concurrent: Sync: RuntimeManager: RequestedBlocks: RPConfAsk: TransportLayer: BlockStore: BlockDagStorage: CommUtil: Log: Timer: Time: Metrics: Span /*: Concurrent: Sync: RuntimeManager: BlockDagStorage: BlockStore: RequestedBlocks: Timer: Time: RPConfAsk: TransportLayer: CommUtil: Metrics: Log: Span*/ ](
-      storeManager: KeyValueStoreManager[F],
+  def streams[F[_]: Concurrent: Sync: RuntimeManager: RequestedBlocks: RPConfAsk: TransportLayer: BlockStore: BlockDagStorage: CasperBufferStorage: BlockRetriever: CommUtil: Log: Timer: Time: Metrics: Span](
       incomingBlocks: Queue[F, BlockMessage],
       receiverOutputQueue: Queue[F, BlockHash],
       finishedProcessingStream: Stream[F, BlockHash],
       casperShardConf: CasperShardConf,
       state: Ref[F, BlockReceiverState[BlockHash]]
   ): (Stream[F, BlockMessage], Stream[F, Unit]) = {
-    val casperBufferF = CasperBufferKeyValueStorage.create[F](storeManager)
 
     def blockStr(b: BlockMessage)       = PrettyPrinter.buildString(b, short = true)
     def logMissingDeps(b: BlockMessage) = Log[F].info(s"Block ${blockStr(b)} missing dependencies.")
@@ -53,57 +51,37 @@ object BlockReceiver {
       incomingBlocks.dequeue
         .parEvalMapUnorderedProcBounded[F, BlockMessage](_.pure)
         .filter(block => block.shardId == casperShardConf.shardName)
-        .evalFilter(
-          block =>
-            casperBufferF >>= { casperBuffer =>
-              implicit val cb: CasperBufferStorage[F] = casperBuffer
-              checkIfOfInterest(block)
-            } >>= { r =>
-              logNotOfInterest(block).unlessA(r).as(r)
-            }
-        )
-        .evalFilter(
-          block =>
-            checkIfWellFormedAndStore(block) >>= { r =>
-              logMalformed(block).unlessA(r).as(r)
-            }
-        )
-        .evalMap(
-          block =>
-            casperBufferF >>= { casperBuffer =>
-              {
-                implicit val cb = casperBuffer
-                commitToBuffer(block, None)
-              } *> block.pure
-            }
-        )
+        .evalFilter { block =>
+          checkIfOfInterest(block) >>= { r =>
+            logNotOfInterest(block).unlessA(r).as(r)
+          }
+        }
+        .evalFilter { block =>
+          checkIfWellFormedAndStore(block) >>= { r =>
+            logMalformed(block).unlessA(r).as(r)
+          }
+        }
+        .evalTap(commitToBuffer(_, None))
         .evalMap(block => state.modify(st => (st.add(block.blockHash), block)))
 
     def processesBlocksStream =
-      finishedProcessingStream
-        .parEvalMapUnorderedProcBounded(
-          _ =>
-            for {
-              casperBuffer   <- casperBufferF
-              blockRetriever = BlockRetriever.of[F]
-              // TODO: Checking and updating state not atomic
-              hashesInReceive <- state.get.map(_.state.keySet)
-              blocksInReceive <- hashesInReceive.toList.traverse(BlockStore[F].getUnsafe)
-              blocksResolved <- {
-                implicit val (br, cb) = (blockRetriever, casperBuffer)
-                blocksInReceive
-                  .filterA(
-                    block =>
-                      checkDependenciesWithEffects[F](block) >>= { r =>
-                        logMissingDeps(block).unlessA(r).as(r)
-                      }
-                  )
-                  .map(_.toSet)
-              }
-              hashesResolved   = blocksResolved.map(_.blockHash)
-              hashesUnresolved = hashesInReceive.diff(hashesResolved)
-              removedHashes    <- hashesResolved.toList.traverse(casperBuffer.remove)
-              _ = state.update { st =>
+      finishedProcessingStream.parEvalMapUnorderedProcBounded { hash =>
+        for {
+          // TODO: Checking and updating state not atomic
+          hashesInReceive <- state.get.map(_.state.keySet)
+          blocksInReceive <- hashesInReceive.toList.traverse(BlockStore[F].getUnsafe)
+          blocksResolved <- blocksInReceive
+                             .filterA { block =>
+                               checkDependenciesWithEffects[F](block) >>= { r =>
+                                 logMissingDeps(block).unlessA(r).as(r)
+                               }
+                             }
+                             .map(_.toSet)
+
+          hashesResolved   = blocksResolved.map(_.blockHash)
+          hashesUnresolved = hashesInReceive.diff(hashesResolved)
+          removedHashes    <- hashesResolved.toList.traverse(CasperBufferStorage[F].remove)
+          _ <- state.update { st =>
                 // TODO: The pattern (removedHashes.contains) not guarantee atomically removing hash
                 //  from CasperBuffer and state because incomingBlocksStream can push hashes in parallel.
                 val newSt = hashesResolved.filter(removedHashes.contains).foldLeft(st) {
@@ -111,12 +89,12 @@ object BlockReceiver {
                 }
                 newSt
               }
-              _ = hashesResolved
+          _ <- hashesResolved.toList
                 .filter(removedHashes.contains)
-                .foreach(receiverOutputQueue.enqueue1)
-              // TODO: Should we request dependencies for hashesUnresolved?
-            } yield ()
-        )
+                .traverse_(receiverOutputQueue.enqueue1)
+          // TODO: Should we request dependencies for hashesUnresolved?
+        } yield ()
+      }
     (incomingBlocksStream, processesBlocksStream)
   }
 
--- a/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
@@ -72,6 +72,15 @@ object BlockReceiver {
               logMalformed(block).unlessA(r).as(r)
             }
         )
+        .evalMap(
+          block =>
+            casperBufferF >>= { casperBuffer =>
+              {
+                implicit val cb = casperBuffer
+                commitToBuffer(block, None)
+              } *> block.pure
+            }
+        )
         .evalMap(block => state.modify(st => (st.add(block.blockHash), block)))
 
     def processesBlocksStream =
@@ -97,15 +106,18 @@ object BlockReceiver {
               }
               hashesResolved   = blocksResolved.map(_.blockHash)
               hashesUnresolved = hashesInReceive.diff(hashesResolved)
+              removedHashes    <- hashesResolved.toList.traverse(casperBuffer.remove)
               _ = state.update { st =>
-                val newSt = hashesResolved.foldLeft(st) {
-                  case (st, hash) =>
-                    st.remove(hash)
+                // TODO: The pattern (removedHashes.contains) not guarantee atomically removing hash
+                //  from CasperBuffer and state because incomingBlocksStream can push hashes in parallel.
+                val newSt = hashesResolved.filter(removedHashes.contains).foldLeft(st) {
+                  case (st, hash) => st.remove(hash)
                 }
                 newSt
               }
-              // TODO: Should resolved hashes be pushed atomically or not?
-              _ = hashesResolved.foreach(receiverOutputQueue.enqueue1)
+              _ = hashesResolved
+                .filter(removedHashes.contains)
+                .foreach(receiverOutputQueue.enqueue1)
               // TODO: Should we request dependencies for hashesUnresolved?
             } yield ()
         )
--- a/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
+++ b/casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
@@ -24,14 +24,10 @@ import fs2.Stream
 import fs2.concurrent.Queue
 
 sealed trait RecvStatus
-case object Receiving   extends RecvStatus // Block receiving started
-case object Received    extends RecvStatus // Block fully received
 case object Unvalidated extends RecvStatus // Block received but not validated (put in BlockStore)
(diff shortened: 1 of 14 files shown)


casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
```diff
@@ -0,0 +1,18 @@
+package coop.rchain.casper.engine
+
+import cats.effect.Sync
+import cats.effect.concurrent.Ref
+
+sealed trait RecvStatus
+case object RecvStarted  extends RecvStatus
+case object RecvFinished extends RecvStatus
+
+final case class BlockReceiverState[F[_], MId](state: Map[MId, RecvStatus]) {
```
tgrospic: F is not used.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
```diff
@@ -0,0 +1,18 @@
...
+  def contains(m: MId): Boolean       = state.contains(m)
+  def set(m: MId, status: RecvStatus) = new BlockReceiverState[F, MId](state + (m -> status))
```
tgrospic: `new` is redundant for case classes.

To create a new state use `copy(state + (m -> status))` so that all other fields (if defined) are copied also.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
```diff
@@ -0,0 +1,18 @@
...
+}
+
+object BlockReceiverState {
+  def create[F[_]: Sync, MId]: F[Ref[F, BlockReceiverState[F, MId]]] =
+    Ref.of(BlockReceiverState[F, MId](Map.empty[MId, RecvStatus]))
```
tgrospic: Instance of Ref which holds the state should be internal to the component (block receiver in this case) and function should be defined to access/modify internal state.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -304,6 +315,9 @@ class Running[F[_]
                   s"from ${peer.endpoint.host}"
               )
             )
+        _ <- blockReceiverState.update { state =>
+              state.set(b.blockHash, RecvFinished)
+            }
```
tgrospic: So here function defined on block receiver component should be called, not accessing the internal state directly.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
```diff
@@ -0,0 +1,18 @@
...
+}
```
tgrospic: Please take a look how LfsBlockRequester is defined with its internal state inside Ref which is created in a function which wraps the whole component and returns Stream of received block hashes.

https://github.com/rchain/rchain/blob/dd46a6b59f0d7585ae27148f72b2c236197b0bb3/casper/src/main/scala/coop/rchain/casper/engine/LfsBlockRequester.scala#L139-L351

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
stanislavlyalin: [Resolved](https://github.com/rchain/rchain/pull/3650/commits/7f04c4c6225e5a57913a46fa8322c6c90d9c3ceb).

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
stanislavlyalin: [Resolved](https://github.com/rchain/rchain/pull/3650/commits/0268c670a40d29a01e1f475cfd1ec75f4439e1f5).

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
stanislavlyalin: @tgrospic, I [moved](https://github.com/rchain/rchain/pull/3650/commits/0268c670a40d29a01e1f475cfd1ec75f4439e1f5) `Ref` inside `BlockReceiveState` but solution is differ you mentioned below ([LfsBlockRequester](https://github.com/rchain/rchain/blob/dd46a6b59f0d7585ae27148f72b2c236197b0bb3/casper/src/main/scala/coop/rchain/casper/engine/LfsBlockRequester.scala#L139-L351)). Should I remove `Ref` from state and wrap it into `Stream`?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
stanislavlyalin: @tgrospic, I fixed the usage of `Ref` in both places of `Running.scala`. Here is the first one
```scala
val _ = blockReceiverState
      .contains(h)
      .map { contains =>
        if (contains) blockReceiverState.set(h, RecvStarted)
      }
```
and the second
```scala
_ <- blockReceiverState.set(b.blockHash, RecvFinished)
```
In the first case checking the hash and modifying the state is not atomic so I think we should some synchronization mechanism for that.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
```diff
@@ -2,17 +2,18 @@ package coop.rchain.casper.engine
...
 sealed trait RecvStatus
 case object RecvStarted  extends RecvStatus
 case object RecvFinished extends RecvStatus
 
-final case class BlockReceiverState[MId](state: Map[MId, RecvStatus]) {
-  def contains(m: MId): Boolean                                = state.contains(m)
-  def set(m: MId, status: RecvStatus): BlockReceiverState[MId] = copy(state + (m -> status))
+final case class BlockReceiverState[F[_]: Sync, MId](ref: Ref[F, Map[MId, RecvStatus]]) {
+  def contains(m: MId): F[Boolean]             = ref.get.map(_.contains(m))
+  def set(m: MId, status: RecvStatus): F[Unit] = ref.update(_ + (m -> status))
```
tgrospic: We can see this class as `BlockReceiver` component which has the state `Map[MId, RecvStatus]`.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
tgrospic: What you mean by _wrap it into `Stream`_?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
```diff
@@ -109,6 +110,13 @@ object Running {
     val processHash =
       BlockRetriever[F].admitHash(h, peer.some, BlockRetriever.HashBroadcastRecieved)
 
+    // TODO: [sl] We should also check hash in Validation and DAG
+    val _ = blockReceiverState
+      .contains(h)
+      .map { contains =>
+        if (contains) blockReceiverState.set(h, Receiving)
+      }
```
tgrospic: This is a problem for concurrency if two threads read `blockReceiverState.contains(h)` before `blockReceiverState.set(h, Receiving)`. This transition of state must be done inside Ref so it's atomic.

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
tgrospic: Just add operation on the state which will do both, checking is contains the hash and also setting the status.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
stanislavlyalin: The [stream](https://github.com/rchain/rchain/blob/dd46a6b59f0d7585ae27148f72b2c236197b0bb3/casper/src/main/scala/coop/rchain/casper/engine/LfsBlockRequester.scala#L152) function returns `Stream` which is wrapped Ref over ST (state) if my understanding is correct.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiverState.scala
stanislavlyalin: @tgrospic, can you explain please what do you mean? I just moved this class into BlockReceiver object and created singleton object for class. Is [it](https://github.com/rchain/rchain/pull/3650/commits/0268c670a40d29a01e1f475cfd1ec75f4439e1f5) correct?

casper/src/main/scala/coop/rchain/casper/engine/Running.scala
stanislavlyalin: @tgrospic, we cannot use Ref inside `BlockReceiver` because we should also check the hash in the DAG so the checking should be on the top level of logic. And I don't have an idea how to use one more Ref in that case. I found the simplest solution as
```scala
val lock = new Object
def add(i: Int) = lock.synchronized { x += i }
```
but I guess it can slow down performance

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -106,6 +106,19 @@ object Setup {
...
       // Block storage
       blockStore <- KeyValueBlockStore(rnodeStoreManager)
+      blockKeyValueStore <- rnodeStoreManager
+                             .store("blocks")
+                             .map(
+                               _.toTypedStore[BlockHash, BlockMessage](
+                                 codecBlockHash,
+                                 codecBlockMessage
+                               )
+                             )
```
tgrospic: Block store uses compression for BlockMessage serialization.
https://github.com/rchain/rchain/blob/326155661e4be70cde0bfe4043ae08af66ebeba4/block-storage/src/main/scala/coop/rchain/blockstorage/KeyValueBlockStore.scala#L94-L106

Now that DAG is used to find message by block hash (#3521) we can reuse KeyValueTypedStore directly as a block store.
Please create another PR for that with special scodec for BlockMessage which uses compression as it's now defined in `KeyValueBlockStore`. This serializer should wrap BlockMessage protobuf serialization (as you did with `codecBlockMessage`) with addition for compression.

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -106,6 +106,19 @@ object Setup {
...
+      blockMetadataStore <- rnodeStoreManager.database[BlockHash, BlockMetadata](
+                             "block-metadata",
+                             codecBlockHash,
+                             codecBlockMetadata
+                           )
```
tgrospic: Instead of directly working with underlying store used in BlockMetadataStore define function(s) which can access data you need from BlockMetadataStore class and supply these function to BlockReceiver. in that way only BlockReceiver should now know internal implementation for BlockMetadataStore.

stanislavlyalin: I [used](https://github.com/rchain/rchain/pull/3650/commits/7981e75498697da43ad7426abd00fde5b075baaf) little bit different approach. I just pass unvalidated and validated set of hashes.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
```diff
@@ -8,10 +8,10 @@ import coop.rchain.shared.syntax._
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 import scodec.codecs._
 
-object ApprovedStore {
+object approvedStore {
```
nzpr: I'm not sure this is a good change, per Scala [naming conventions](https://docs.scala-lang.org/style/naming-conventions.html#objects) object name should start with uppercase with rare exceptions

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -173,10 +180,19 @@ object Setup {
...
       // block processing state - set of items currently in processing
       blockProcessorStateRef <- Ref.of(Set.empty[BlockHash])
+      unvalidated            <- blockStore.toMap
+      validated              <- blockMetadataStore.toMap
+      blockReceiverState <- BlockReceiver.create[F](
+                             blockRetriever,
+                             casperBufferStorage,
+                             unvalidated.keySet,
+                             validated.keySet
+                           )
```
tgrospic: We cannot read the whole metadata store. It can take minutes .

Please supply functions from Dag storage which are necessary to check if block is already in the dag.

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
```diff
@@ -0,0 +1,86 @@
...
+import coop.rchain.models.BlockMetadata
+import coop.rchain.shared.Log
+import coop.rchain.store.KeyValueTypedStore
+
+sealed trait RecvStatus
+case object Receiving   extends RecvStatus // Block receiving started
+case object Received    extends RecvStatus // Block fully received
+case object Unvalidated extends RecvStatus // Block received but not validated (put in BlockStore)
+case object Validated   extends RecvStatus // Block received and validated (ready to put in BlockMetadataStore/DAG)
+case object Invalid     extends RecvStatus // Block received but invalid
```
nzpr: I have some questions related to the scope of BlockReceiver and these statuses

Does Receiving here means that node became aware of some new hash and tries to retrieve full block from the network/disk/torrent/etc? It looks to me so, so maybe better to call this `Wanted`?

What is the difference between `Received` and `Unvalidated`? Why to distinguish between Validated and Invalid here?

stanislavlyalin: > Does Receiving here means that node became aware of some new hash and tries to retrieve full block from the network/disk/torrent/etc? It looks to me so, so maybe better to call this Wanted?

Yes, make sense. Maybe better `Waited`?

> What is the difference between Received and Unvalidated?

Block `Received` after getting `BlockMessage` message in `Running` and `Unvalidated` when it stored in `BlockStore` after some initial validation (format, signature, ...).

> Why to distinguish between Validated and Invalid here?

A block is considered `Validated` when it is removed from the `CasperBuffer`. The `Invalid` status is for invalid blocks, but it is not currently in use, so it can be removed.

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
stanislavlyalin: Probably [done](https://github.com/rchain/rchain/pull/3650/commits/ecdff2c36e5773de6768b112b7542a066b85b4e7).

casper/src/main/scala/coop/rchain/casper/engine/BlockReceiver.scala
```diff
@@ -0,0 +1,87 @@
...
+    } yield BlockReceiver(ref, blockRetriever, casperBuffer)
+
+  def create[F[_]: Sync /*: Log*/ ](
+      blockRetriever: BlockRetriever[F],
+      casperBuffer: CasperBufferStorage[F],
+      blockStore: BlockStore[F],
+      blockDagRepresentation: BlockDagRepresentation[F]
+  ): F[BlockReceiver[F]] =
+    for {
+      blockHashes <- blockStore.toMap >>= { _.keySet.pure }
```
tgrospic: `toMap` on the block store is very heavy operation, it will load all blocks which can be tens of GBs.

We don't need to load all blocks at the beginning, BlockReceiver just needs to resolve what to do with newly recived block hash and block. Is it ready for the next stage (ready to validation/DAG), missing dependencies or it should be ignored.

On the beginning casper buffer must be processed because it contains blocks that are downloaded but not yet in the DAG.

stanislavlyalin: @tgrospic, I decided to use hashes from `CasperBuffer` as `Unvalidated` hashes. `CasperBufferStorage` hasn't got functions to collect all stored hashes so I added [`nodes`](https://github.com/rchain/rchain/pull/3650/commits/75b710614497bc4444b9e66662bc1be14cb8c71a) function for that. Or maybe it should be named `blocks` or `messages`?
**Upd** `CasperBufferStorage` trait can be unchanged if `toDoublyLinkedDag` function will be used.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
tgrospic: Fixed in https://github.com/rchain/rchain/pull/3673/commits/97886a864d6cf0329c539a83f1ed053969e53d24 for block store. Approved block will be removed for HF2 anyway.

casper/src/main/scala/coop/rchain/casper/engine/NodeSyncing.scala
```diff
@@ -116,7 +119,8 @@ class NodeSyncing[F[_]
     case b: BlockMessage =>
       Log[F]
         .info(s"BlockMessage received ${PrettyPrinter.buildString(b, short = true)} from $peer.") *>
-        blockMessageQueue.enqueue1(b)
+        blockMessageQueue.enqueue1(b) *>
+        incomingBlocks.enqueue1(b.blockHash)
```
tgrospic: Why is this needed? When node is in LFS syncing mode it will receive blocks with LfsBlockRequester and this blockMessageQueue is just for that. After LFS sync is done NodeRunning is responsible to handle network mesages.

stanislavlyalin: Please take a look at [commit](https://github.com/rchain/rchain/pull/3650/commits/20171946cdf4b0e439d3808b51ee8d059b1d1df1). This `blockProcessorQueue` fills only when node running. Are we consider that LFS blocks should not be handled with BlockReceiver?

tgrospic:
Superseded by:
- #3722.

PR #3655, opened 2022-04-08
BlockStore implemented as KeyValueTypedStore

stanislavlyalin:
## Overview

`BlocKStore` is a key-value store for both `BlockMessage` and `ApprovedBlock`. For the first one the key is the hash of the block and the value is the `BlockMessage`. For the second - the key is a fixed value of `42`, and the value is `ApprovedBlock`, because he can only be one. We have `KeyValueTypedStore` for this purpose. This PR splits `BlockStore` into `BlockStore` for `BlockMessage` and `ApprovedStore` for `ApprovedBlock`.

`KeyValueTypedStore` requires codecs to encode/decode values. `BlockMessage` uses compression encoding using the [lz4](https://github.com/lz4/lz4-java) library. Appropriate codecs have been added for the `ApprovedBlock` key and value.

The `finality` directory and all attached files have been removed as they are not being used.


block-storage/src/main/scala/coop/rchain/blockstorage/LastFinalizedStore.scala
```diff
@@ -0,0 +1,14 @@
...
+import coop.rchain.blockstorage.dag.codecs.{codecBlockHash, codecSeqNum}
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.shared.syntax.sharedSyntaxKeyValueStore
+import coop.rchain.store.{KeyValueStore, KeyValueTypedStore}
+
+object LastFinalizedStore {
+  type LastFinalizedStore[F[_]] = KeyValueTypedStore[F, Int, BlockHash]
+
+  def apply[F[_]: Sync](keyValueStore: KeyValueStore[F]): LastFinalizedStore[F] =
+    keyValueStore.toTypedStore(codecSeqNum, codecBlockHash)
```
tgrospic: LastFinalizedStore can be completely removed, it's not used anymore.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
```diff
@@ -0,0 +1,56 @@
...
+    // ApprovedStore extensions / syntax
+    private val approvedStore: ApprovedStore[F]
+) {
+  def getApprovedBlock: F[Option[ApprovedBlock]] =
+    for {
+      // Optional block message from the store
+      approvedBlockHash <- approvedBlockKeyAsHash
+      proto             <- approvedStore.get1(approvedBlockHash).map(_.map(_.toProto))
+      // Decode protobuf message / throw if fail
+      block <- proto.traverse(ApprovedBlock.from(_).leftMap(errorApprovedBlock).liftTo[F])
```
tgrospic: Here is unnecessary serailization to proto and then back to Approved block.
`approvedStore.get1(approvedBlockHash)` should return ApprovedBlock, right?

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
```diff
@@ -0,0 +1,20 @@
...
+
+object ApprovedStore {
+  type ApprovedStore[F[_]] = KeyValueTypedStore[F, BlockHash, ApprovedBlock]
+
+  def apply[F[_]: Sync](
+      kvm: KeyValueStoreManager[F]
+  ): F[KeyValueTypedStore[F, BlockHash, ApprovedBlock]] =
+    kvm
+      .store("blocks-approved")
+      .map(_.toTypedStore[BlockHash, ApprovedBlock](codecBlockHash, codecApprovedBlock))
```
tgrospic: Use `bytes` scodec as a key. No need to involve ByteString when we know the key is `Array[Byte]` which can easily be converted to scodec.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
```diff
@@ -0,0 +1,56 @@
...
+    } yield block
+
+  def putApprovedBlock(block: ApprovedBlock): F[Unit] =
+    approvedBlockKeyAsHash.flatMap(approvedStore.put(_, block))
+
+  val approvedBlockKey: Array[Byte] = Array[Byte](42)
```
tgrospic: Create key here directly with`bytes` scodec.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
```diff
@@ -20,11 +21,26 @@ object codecs {
...
+
   val codecBlockMetadata = variableSizeBytes(uint16, bytes).xmap[BlockMetadata](
     byteVector => BlockMetadata.fromBytes(byteVector.toArray),
     blockMetadata => ByteVector(blockMetadata.toByteString.toByteArray)
   )
 
+  val codecBlockMessage = variableSizeBytesLong(uint32, bytes).xmap[BlockMessage](
+    byteVector => bytesToBlockMessage(byteVector.toArray),
+    blockMessage => ByteVector(blockMessageToBytes(blockMessage))
+  )
```
tgrospic, line 34: You can use `bytes` scodec directly. Variable size is only needed when combining raw _bytes_ in a list.

The same for approved block codec.

casper/src/main/scala/coop/rchain/casper/blocks/BlockProcessor.scala
```diff
@@ -115,15 +116,15 @@ object BlockProcessor {
...
   def apply[F[_]
   /* Execution */   : Concurrent
-  /* Storage */     : BlockStore: BlockDagStorage: CasperBufferStorage 
+  /* Storage */     : BlockDagStorage: CasperBufferStorage 
   /* Diagnostics */ : Log
   /* Comm */        : CommUtil: BlockRetriever
   ] // format: on
-  (
+  (blockStore: BlockStore[F])(
       implicit casperBuffer: CasperBufferStorage[F]
```
tgrospic: Please remove duplicated implicit `CasperBufferStorage`.

block-storage/src/main/scala/coop/rchain/blockstorage/LastFinalizedStore.scala
stanislavlyalin: Sure. I forgot to push my latest changes :slightly_smiling_face: So now it is removed.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
stanislavlyalin: @tgrospic, I [simplified](https://github.com/rchain/rchain/pull/3655/commits/a5176fc33d155fce6725a71e09987e076ffd5339) ApprovedStore key to just 1 byte. Is it ok?

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
```diff
@@ -0,0 +1,20 @@
...
+
+object ApprovedStore {
+  type ApprovedStore[F[_]] = KeyValueTypedStore[F, Byte, ApprovedBlock]
+
+  def apply[F[_]: Sync](
+      kvm: KeyValueStoreManager[F]
+  ): F[KeyValueTypedStore[F, Byte, ApprovedBlock]] =
+    kvm
+      .store("blocks-approved")
+      .map(_.toTypedStore[Byte, ApprovedBlock](byte, codecApprovedBlock))
```
tgrospic, line 20: Nice! Even better then bytes. :)

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
```diff
@@ -1,31 +1,39 @@
...
+import coop.rchain.shared.syntax._
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import net.jpountz.lz4.{LZ4CompressorWithLength, LZ4DecompressorWithLength}
 
-trait BlockStore[F[_]] {
-  def get(blockHash: BlockHash): F[Option[BlockMessage]]
-
-  def put(f: => (BlockHash, BlockMessage)): F[Unit]
+object BlockStore {
+  type BlockStore[F[_]] = KeyValueTypedStore[F, BlockHash, BlockMessage]
```
tgrospic, line 14: Although it makes sense to use BlockStore explicitly, for now better to minimize changes. I suggest this definition which will enable us to use BlockStore implicitly as now.

Object name start with lowercase so it's not shadowing `BlockStore[F]` constructor for implicit use as we have now.
```scala
object blockStore {
  type BlockStore[F[_]] = KeyValueTypedStore[F, BlockHash, BlockMessage]

  // Simulates usual apply definition to get the instance `BlockStore[F]` from implicit value
  def BlockStore[F[_]](implicit instance: BlockStore[F]): instance.type = instance
  ...
}
```

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
stanislavlyalin: Yes, exactly. I [simplified](https://github.com/rchain/rchain/pull/3655/commits/ed10be1e029e455e2a83486984442db276201d83) implementation of `getApprovedBlock`.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3655/commits/a5176fc33d155fce6725a71e09987e076ffd5339).

casper/src/main/scala/coop/rchain/casper/blocks/BlockProcessor.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3655/commits/23edd3d800648c5615666c4b37c6b2255c18a916).

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
stanislavlyalin: Yes, I agree. I [reverted](https://github.com/rchain/rchain/pull/3655/commits/65d20ccf0b97bf3f24df18d50ee9c30c8ce0f6e5) changes and now PR has much less changed lines.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
stanislavlyalin: @tgrospic, do we have functions to convert ByteVector to ApprovedBlock and back?

**Upd**: We can use intermediate ApprovedBlockProto but it is very similar to existing codecApprovedBlock. Should I move converting to ApprovedStoreOps?

casper/src/main/scala/coop/rchain/casper/blocks/proposer/Proposer.scala
```diff
@@ -197,7 +198,7 @@ object Proposer {
 
     val proposeEffect = (c: Casper[F], b: BlockMessage) =>
       // store block
-      BlockStore[F].put(b) >>
+      BlockStore[F].put(b.blockHash, b) >>
```
tgrospic: Please create extension for `put(b: BlockMessage)` as it was defined on old BlockStore.

casper/src/test/scala/coop/rchain/casper/api/BlocksResponseAPITest.scala
```diff
@@ -48,7 +49,7 @@ class BlocksResponseAPITest
     for {
       genesis <- createGenesis[Task](bonds = bonds)
       _       <- dagstore.insert(genesis, invalid = false, approved = true)
-      _       <- blockstore.put(genesis)
+      _       <- blockstore.put(genesis.blockHash, genesis)
```
tgrospic: Note to use put with one argument here also.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
```diff
@@ -46,3 +46,5 @@ final class KeyValueTypedStoreOps[F[_], K, V](
   def getOrElse(key: K, elseValue: V)(implicit f: Functor[F]): F[V] =
     get1(key).map(_.getOrElse(elseValue))
 }
+
+object KeyValueTypedStoreSyntaxObj extends KeyValueTypedStoreSyntax
```
tgrospic: Why is this object definition required. Syntax trait should be extended in the project level package file.

https://github.com/rchain/rchain/blob/606c87d944ec7e4080387d0b8bd7b961e122992e/shared/src/main/scala/coop/rchain/shared/package.scala#L12-L17

So when used one import is needed for project extensions.
```scala
import coop.rchain.shared.syntax._
```

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
```diff
@@ -25,6 +26,19 @@ object codecs {
...
+
+  val codecApprovedBlock = variableSizeBytesLong(uint32, bytes).xmap[ApprovedBlock](
+    byteVector =>
+      ApprovedBlock
+        .from(ApprovedBlockProto.parseFrom(byteVector.toArray))
+        .getOrElse(ApprovedBlock.empty),
```
tgrospic: We don't want to silently swallow errors. You can use `exmap` which supports errors. Empty definitions fro BlockMessaage and ApprovedBlock are not needed.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
tgrospic: BlockMessaage and ApprovedBlock are defined with protobuf so the only correct way is to use proto variants with protobuf encoding/decoding via ByteString.
Scodec definition is just transparent wrapper which uses binary directly from protobuf, that's the reason why `bytes` scodec is enough without size prefix (from `variableSizeBytesLong` scodec).

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStore.scala
```diff
@@ -1,31 +1,40 @@
...
+      kvm: KeyValueStoreManager[F]
+  ): F[KeyValueTypedStore[F, BlockHash, BlockMessage]] =
+    kvm
+      .store("blocks")
+      .map(_.toTypedStore[BlockHash, BlockMessage](codecBlockHash, codecBlockMessage))
+
+  def bytesToBlockMessage(bytes: Array[Byte]): BlockMessage =
+    BlockMessage
+      .from(BlockMessageProto.parseFrom(decompressBytes(bytes)))
+      .getOrElse(BlockMessage.empty)
```
tgrospic: `bytesToBlockMessage` should return Either directly so error can be used in scodec definition.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
stanislavlyalin: Yes, I tried this way but it leads to errors such a
```
BlockProcessor.scala:51:56: value ||^ is not a member of type parameter F[Boolean]
[error]     val alreadyProcessed  = c.dagContains(b.blockHash) ||^ c.bufferContains(b.blockHash)
```
I guess the problem with this
```scala
trait AllSyntaxShared
    extends KeyValueStoreSyntax
    with KeyValueTypedStoreSyntax
    with KeyValueStoreManagerSyntax
    with MonixableSyntax
    with Fs2StreamSyntax
    with catscontrib.ToBooleanF  // <--
```

casper/src/main/scala/coop/rchain/casper/blocks/proposer/Proposer.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3655/commits/7072a5297e0ed0d6872f9ca130d313295078a0ec).

casper/src/test/scala/coop/rchain/casper/api/BlocksResponseAPITest.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3655/commits/7072a5297e0ed0d6872f9ca130d313295078a0ec).

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
stanislavlyalin: Probably [resolved](https://github.com/rchain/rchain/pull/3655/commits/9b9c0bded63f00fea1ea6363ab798c693d223f7a).

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
```diff
@@ -0,0 +1,20 @@
...
+
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.shared.syntax._
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import scodec.bits.ByteVector
+import scodec.codecs._
+
+object approvedStore {
+  type ApprovedStore[F[_]] = KeyValueTypedStore[F, Byte, ByteVector]
```
tgrospic: Type of value should not be raw ByteVector but ApprovedBlock so all encoding should be done in scodec definition. In the same way as BlockStore with BlockMessage.

shared/src/main/scala/coop/rchain/store/KeyValueTypedStoreSyntax.scala
tgrospic: You are right, `ToBooleanF` should not be added (yet) to shared syntax because it can cause duplicate import with (FYI @nzpr):
https://github.com/rchain/rchain/blob/94ac5f79beb51445586e86bd9c8c7854c1ef5118/shared/src/main/scala/coop/rchain/catscontrib/Catscontrib.scala#L1-L5

I have small commit to fix that.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
stanislavlyalin: Probably [done](https://github.com/rchain/rchain/pull/3655/commits/099836b2cf7c151f85b598f7b004eed7b7c8b848).

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
```diff
@@ -0,0 +1,31 @@
...
+  val approvedBlockKey: Byte = 42.toByte
+
+  def getApprovedBlock: F[Option[ApprovedBlock]] = approvedStore.get1(approvedBlockKey)
+
+  def putApprovedBlock(block: ApprovedBlock): F[Unit] = approvedStore.put(approvedBlockKey, block)
+
+  /**
+    * This is fatal error from which Approved Store can not be recovered.
+    */
+  case class ApprovedStoreFatalError(message: String) extends Exception(message)
```
tgrospic: This is left over.

stanislavlyalin: Yes, you are right. This is my carelessness. I [removed](https://github.com/rchain/rchain/pull/3655/commits/bda8ca1f0b83e68a48c75cda0af64c2e5bcb637c) it.

tgrospic (approved):
Excellent work! Thanks for addressing all comments.

tgrospic:
Reference for the whole cleanup:
- #3655
- #3495
- #3342 
- #3205

File based storage introduced:
- #2076

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
```diff
@@ -0,0 +1,26 @@
+package coop.rchain.blockstorage
+
+import cats.effect.Sync
+import cats.implicits._
```
zsluedem: recommend `cats.syntax.all._`instead of `cats.implicits._`

reason https://discord.com/channels/391435113415573504/410182072054054912/714082912181551104

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import cats.effect.Sync
+import cats.syntax.all._
+import coop.rchain.blockstorage.dag.codecs.codecApprovedBlock
+import coop.rchain.casper.protocol.{ApprovedBlock, ApprovedBlockProto}
+import coop.rchain.shared.syntax._
+import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import scodec.codecs._
+
+object approvedStore {
```
zsluedem, line 11: Another nitpick.
`ApprovedStore` instead of `approvedStore`. Just some warning I saw in intellJ:)

zsluedem (approved):
LGTM. Left some nitpicks.

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStoreSyntax.scala
stanislavlyalin: Yes, reasonable. [Fixed](https://github.com/rchain/rchain/pull/3655/commits/61a9b18a9873ce548ae9b3f0c33a2ac488071fe8).

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
stanislavlyalin: Yes, it looks strange but here is [reason](https://github.com/rchain/rchain/pull/3655#discussion_r848253129) of lower-case object name.

stanislavlyalin:
bors merge

block-storage/src/main/scala/coop/rchain/blockstorage/ApprovedStore.scala
tgrospic: @zsluedem This object definition has the purpose like a package which starting with lower case makes sense (and to avoid hiding alias and implicit instance helper).
https://docs.scala-lang.org/style/naming-conventions.html#objects

zsluedem: Hm.... The link says `Object names are like class names (upper camel case).`

While you could `mimicking a package or function which is not common` with lower case. What a confusion!

tgrospic: I just tested with upper case and it can work also (I think I misunderstood the error in syntax file when I was refactoring).
```scala
object BlockStore {
  type BlockStore[F[_]] = KeyValueTypedStore[F, BlockHash, BlockMessage]

  def apply[F[_]](implicit instance: BlockStore[F]): instance.type = instance
}
```
This changes a lot of files so better to do it in a separate PR to minimize conflicts.

PR #3661, opened 2022-04-14
Remove deprecated code (estimator, safety oracle)

nzpr:
## Overview
This is the first part in series of changes (`finalization` feature branch) aiming to replace current finalizer with the new one that supports block merge. 
in this PR no new code is introduced, only several tests are adjusted that were relying on finalization + dummy values are placed where finalizer output is required. PTAL at TODO comments next to adjustments.

Changes list:
1. Remove storage for last finalized block. This should be reimplemented as part of integration of new finalizer.
2. Remove obsolete finalizer.
3. Remove Estimator and safety oracle. As new version of finalization does not rely on main parent computation, this code is obsolete.
4. Adjust tests. Tests that do not make sense anymore are ignored. Tests that assumes some specific finalizer output are adjusted to not make such assumption.  

### Notes
One tests remaining failing is integration test for finalization. It does not make sense to ignore it and as this PR is based on a feature branch, it can and should be fixed later when finalization code is merged.

### Please make sure that this PR:
- [x] is at most 200 lines of code (excluding tests),
- [x] meets [RChain development coding standards](https://rchain.atlassian.net/wiki/spaces/DOC/pages/28082177/Coding+Standards),
- [x] includes tests for all added features,
- [x] has a reviewer assigned,
- [x] has [all commits signed](https://rchain.atlassian.net/wiki/spaces/DOC/pages/498630673/How+to+sign+commits+to+rchain+rchain).

### [Bors](https://bors.tech/) cheat-sheet:

- `bors r+` runs integration tests and merges the PR (if it's approved),
- `bors try` runs integration tests for the PR,
- `bors delegate+` enables non-maintainer PR authors to run the above.

PR #3662, opened 2022-04-15
Refactor BlockAPI

### LINKED ISSUE #3659: Refactor BlockAPI
## Overview

To include new multi-parent finalization we need to refactor/delete a lot of existing code.

As part of this change we need to simplify current `BlockAPI` implementation.

## Design

1. Create `BlockAPI` trait with existing methods currently defined on the BlockAPI object
    - for now it can be in models project (because we have generated protobuf there!) in `coop.rchain.api`
1. Create implementation which should only accept RuntimeManager and BlockDagStorage (BlockDagRepresentation)
    - for all other dependencies just add TODO comment with fake data
1. Instantiate new BlockAPI implementation in `Setup` and use where is needed like gRPC servers and WebAPI

## Additional changes

Currently, `addDeploy` method is defined on MultiParentCasperImpl which is used in BlockAPI. This should be moved directly to new BlockAPI implementation and DeloyStorage should be accessible in BlockDagStorage.

stanislavlyalin:
## Overview

Introduced BlockAPI as a trait and its implementation. Eliminated some dependencies such as EngineCell, SafetyOracle and partially MultiParentCasper. In the future the new implementation will replace current BlockAPI object.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
@@ -40,7 +40,7 @@ import coop.rchain.shared.Log
 
 import scala.collection.immutable
 
-class BlockAPIImpl[F[_]: Concurrent: Log: Span: SafetyOracle: BlockStore: DeployStorage: MultiParentCasper](
+class BlockAPIImpl[F[_]: Concurrent: Log: Span: BlockStore: DeployStorage: MultiParentCasper](
     runtimeManager: RuntimeManager[F],
     dagStorage: BlockDagStorage[F],
     validatorPrivateKey: Option[String]
@@ -445,8 +445,7 @@ class BlockAPIImpl[F[_]: Concurrent: Log: Span: SafetyOracle: BlockStore: Deploy
                                    dag
                                      .isFinalized(block.blockHash)
                                      .map(isFinalized => if (isFinalized) 1f else -1f),
-                                   SafetyOracle[F]
-                                     .normalizedFaultTolerance(dag, block.blockHash)
+                                   Float.MinValue.pure[F]
                                  )
       initialFault   <- normalizedInitialFault(ProtoUtil.weightMap(block))
       faultTolerance = normalizedFaultTolerance - initialFault
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
@@ -1,0 +1,759 @@
+package coop.rchain.casper.api
+
+import cats.effect.concurrent.Ref
+import cats.effect.{Concurrent, Sync}
+import cats.syntax.all._
+import com.google.protobuf.ByteString
+import coop.rchain.api.BlockAPI_v2
+import coop.rchain.blockstorage.BlockStore
+import coop.rchain.blockstorage.dag.BlockDagRepresentation
+import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
+import coop.rchain.casper.DeployError._
+import coop.rchain.casper._
+import coop.rchain.casper.api.BlockAPI.{
+  BlockRetrievalError,
+  NoBlockMessageError,
+  ValidatorReadOnlyError
+}
+import coop.rchain.casper.blocks.proposer.ProposeResult._
+import coop.rchain.casper.blocks.proposer._
+import coop.rchain.casper.engine.EngineCell._
+import coop.rchain.casper.engine._
+import coop.rchain.casper.genesis.contracts.StandardDeploys
+import coop.rchain.casper.protocol._
+import coop.rchain.casper.state.instances.ProposerState
+import coop.rchain.casper.syntax._
+import coop.rchain.casper.util._
+import coop.rchain.casper.util.rholang.{RuntimeManager, Tools}
+import coop.rchain.crypto.PublicKey
+import coop.rchain.crypto.signatures.Signed
+import coop.rchain.graphz._
+import coop.rchain.metrics.{Metrics, Span}
+import coop.rchain.models.BlockHash.BlockHash
+import coop.rchain.models.rholang.sorter.Sortable._
+import coop.rchain.models.serialization.implicits.mkProtobufInstance
+import coop.rchain.models.syntax._
+import coop.rchain.models.{BlockMetadata, Par}
+import coop.rchain.rspace.hashing.StableHashProvider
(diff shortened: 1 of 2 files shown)


casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
```diff
@@ -0,0 +1,629 @@
...
+import coop.rchain.rspace.hashing.StableHashProvider
+import coop.rchain.rspace.trace._
+import coop.rchain.shared.Log
+
+import scala.collection.immutable
+
+class BlockAPIImpl[F[_]: Concurrent: Log: Span: BlockStore: DeployStorage: MultiParentCasper](
+    runtimeManager: RuntimeManager[F],
+    dagStorage: BlockDagStorage[F],
+    validatorPrivateKey: Option[String]
```
tgrospic: We can use ValidatorIdentity already created in Setup.
https://github.com/rchain/rchain/blob/8603d7a643504a17843526f890e4b45a81f8fcd3/node/src/main/scala/coop/rchain/node/runtime/Setup.scala#L184-L187

stanislavlyalin: Yes, you are right. [Done](https://github.com/rchain/rchain/pull/3662/commits/099e6fa7593df90414e61c31f12a11b7e6256541).

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
```diff
@@ -0,0 +1,626 @@
...
+                                        block
+                                      )
+                                    }
+        blocksWithActiveName = maybeBlocksWithActiveName.flatten
+      } yield (blocksWithActiveName, blocksWithActiveName.length).asRight
+
+  private def getMainChainFromTip(depth: Int): F[IndexedSeq[BlockMessage]] =
+    for {
+      tipHashes <- estimator
+      tipHash   = tipHashes.head
```
nzpr: This will crush node when executed, as `estimator` is empty, and .head on empty list is called. I'm not sure that this API call even should be here, as we don't have main parents. But maybe it is better to make it aligned with <https://github.com/rchain/rchain/pull/3661/commits/c156fa1581077a2dd27cbbcd2106f4dfd7462e83> @tgrospic wdy think?

nzpr, line 222: This will crush node when executed, as `estimator` is empty, and .head on empty list is called. I'm not sure that this API call even should be here, as we don't have main parents. But maybe it is better to make it aligned with <https://github.com/rchain/rchain/pull/3661/commits/c156fa1581077a2dd27cbbcd2106f4dfd7462e83> @tgrospic wdy think?

tgrospic: `getMainChainFromTip` is already removed in dev so this code needs to be sync with latest changes in BlockAPI.
https://github.com/rchain/rchain/blob/33597197cff60dd3a1afe93576569170ce638b47/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L203-L285

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
```diff
@@ -0,0 +1,626 @@
...
+import scala.collection.immutable
+
+class BlockAPIImpl[F[_]: Concurrent: Log: Span: BlockStore: DeployStorage: MultiParentCasper](
+    runtimeManager: RuntimeManager[F],
+    dagStorage: BlockDagStorage[F],
+    validatorOpt: Option[ValidatorIdentity]
+) extends BlockAPI_v2[F] {
+
+  val BlockAPIMetricsSource: Metrics.Source = Metrics.Source(Metrics.BaseSource, "block-api")
+  val DeploySource: Metrics.Source          = Metrics.Source(BlockAPIMetricsSource, "deploy")
```
tgrospic: Please start with lowercase for value names (inside class).
https://docs.scala-lang.org/style/naming-conventions.html#constants-values-and-variables

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
```diff
@@ -0,0 +1,626 @@
...
+      dag <- dagStorage.getRepresentation
+      blockInfo <- dag
+                    .contains(block.blockHash)
+                    .ifM(
+                      getFullBlockInfo(block),
+                      BlockRetrievalError(
+                        s"Error: Block with hash $hash received but not added yet"
+                      ).raiseError[F, BlockInfo]
+                    )
+    } yield blockInfo.asRight).map(_.asInstanceOf[ApiErr[BlockInfo]]).handleError {
```
tgrospic: Try to avoid `asInstanceOf` because it can be unsafe coercion and throw runtime exception.

You probably need to specify left type when wrapping `blockInfo.asRight[String]`.

Another example can be with nested types like `F[List[_]]` which may not be coerced automatically to `F[Seq[_]]`. To make Scala compiler happy I've specified type of `response`.
https://github.com/rchain/rchain/blob/c5629bc6753a22d06d9760f6afcb3a6e419f8c32/casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala#L203-L231

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
stanislavlyalin: I [synchronized](https://github.com/rchain/rchain/pull/3662/commits/929cdf2905dd3a67aabac134695d4edfc55c7508) the code with latest `dev`.

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3662/commits/e7aeb4416e72d2b82bcc839b87259411d7b66d3b).

casper/src/main/scala/coop/rchain/casper/api/BlockAPIImpl.scala
stanislavlyalin: Resolved after [synchronization](https://github.com/rchain/rchain/pull/3662/commits/929cdf2905dd3a67aabac134695d4edfc55c7508).

tgrospic (approved):
Nice!

nzpr (approved):
LGTM

stanislavlyalin:
bors merge

PR #3663, opened 2022-04-18
Fix pos authKey global reused problem

zsluedem:
## Overview

When exploring #3485 changing the registry of AuthKey and found that the POS contract keeps reusing the global pos auth key which would make update authkey not compatible.

The authkey should be `make` in every needed place and use different generated authkey.

The pr is about to change the way how pos use AuthKey.rho and make sure authkey is properly created.

**Shortcoming of making a lot of authkey is that the rspace would full of these persistent consume which is no use or possible exploit bug.**

### DIFF
--- a/casper/src/main/resources/PoS.rhox
+++ b/casper/src/main/resources/PoS.rhox
@@ -97,7 +97,7 @@ in {
        TreeHashMap            <- treeHashMapCh &
        @(_, MultiSigRevVault) <- multiSigRevVaultCh) {
     new posDeployerRevAddressCh, posDeployerAuthKeyCh, posDeployerVaultCh, posVaultCh, 
-        initialActiveCh, pendingRewardsMapCh, createUnfVault, moveInitialBondCh, posAuthKeyCh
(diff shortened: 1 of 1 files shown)


tgrospic:
> Shortcoming of making a lot of authkey is that the rspace would full of these persistent consume which is no use or possible exploit bug.

I agree that every use of AuthKey creates persistent consume which is probably only used once when auth token is verified. Maybe we don't need persistent consume because receiver of the auth token only needs to verify it once?

I don't see how auth token which is not accessible by anyone can be exploited. Unforgeable names are the very means to prevent exploits of that kind.

tgrospic (approved):
LGTM

zsluedem:
> Maybe we don't need persistent consume because receiver of the auth token only needs to verify it once?

I agree with you.

> I don't see how auth token which is not accessible by anyone can be exploited. Unforgeable names are the very means to prevent exploits of that kind.

Yes .You are right.

hilltracer (approved):
LGTM

tgrospic:
bors merge

PR #3664, opened 2022-04-19
Casper (equivocation) cleanup (BlockAPI/BlockReportAPI)

tgrospic:
## Overview

This PR is part of cleanup of DAG management and storage in preparation for multi-parent finalization.

It removes equivocation detector/tracker with its storage, removes dependency to Casper (EngineCell) in BlockAPI and BlockReportAPI, replaces EngineCell with BlockDagStorage when applicable right now (depends on BlockAPI refactoring #3662).
As part of the change is the move of BlockDagKeyValueStorage to casper project which is preparation to complete removal of block-storage project.


casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
```diff
@@ -352,26 +337,26 @@ class MultiParentCasperImpl[F[_]
...
+//                        val newEquivocationRecord =
+//                          EquivocationRecord(
+//                            block.sender,
+//                            baseEquivocationBlockSeqNum,
+//                            Set.empty[BlockHash]
+//                          )
+//                        tracker.insertEquivocationRecord(newEquivocationRecord)
+//                      }
+//                } yield ()
+//              }
```
tgrospic, line 359: For now, I commented old equivocation code just as a reference for easier tracking where code needs to be updated.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -556,37 +547,19 @@ object BlockAPI {
...
-                                     .map(isFinalized => if (isFinalized) 1f else -1f),
-                                   Float.MinValue.pure[F]
-                                 )
-      initialFault   <- casper.normalizedInitialFault(ProtoUtil.weightMap(block))
-      faultTolerance = normalizedFaultTolerance - initialFault
-
-      blockInfo = constructor(block, faultTolerance)
-    } yield blockInfo
+      constructor: (BlockMessage, Float) => A
+  ): F[A] = constructor(block, -1f).pure[F]
```
nzpr, line 487: Might be its better to put here `Float.minValue` to emphasise this is a plug and not some normal value expected from deprecated SafetyOracle.

tgrospic: This is to satisfy what's expected in tests. But we will remove fault tolerance completely so I don't consider it's important.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
```diff
@@ -19,11 +19,11 @@ object codecs {
       byteString => ByteVector(byteString.toByteArray)
     )
 
-  val codecDeployId = xmapToByteString(variableSizeBytes(uint8, bytes))
+  val codecByteString = xmapToByteString(bytes)
```
nzpr, line 22: Could you elaborate why specifying size is not necessary? Is this change compatible with current node version or clean database is required?

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStoreSyntax.scala
```diff
@@ -31,19 +31,9 @@ final class BlockStoreOps[F[_]: Sync](
...
-    * https://github.com/typelevel/cats-effect/pull/854
-    * So extra source parameters are a desperate measure to indicate who is the caller.
     */
-  def getUnsafe(hash: BlockHash)(
-      implicit line: sourcecode.Line,
-      file: sourcecode.File,
-      enclosing: sourcecode.Enclosing
-  ): F[BlockMessage] = {
-    def source = s"${file.value}:${line.value} ${enclosing.value}"
-    def errMsg = s"BlockStore is missing hash ${PrettyPrinter.buildString(hash)}\n  $source"
```
stanislavlyalin, line 46: Why it is removed? It is not used or not important information?

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
stanislavlyalin: Maybe make sense to add `TODO` comment here?

block-storage/src/main/scala/coop/rchain/blockstorage/dag/codecs.scala
tgrospic: It's not compatible.

Size prefix is only needed if _bytes_ are combined together, like in a list or with multiple fields.

E.g. if bytes are `aaaa` and `bbbb` (pseudo representation) then decoding each will be correct, but just combine them `aaaabbbb` there is no way to distinguish what is the first part what is the second so something like `4aaaaa4bbbb` is needed for decoder to correctly recognize each piece.
This means we don't need to defined size prefix on individual codec but only when they are combined.

stanislavlyalin: > This means we don't need to defined size prefix on individual codec but only when they are combined.

Hm.. Maxim meet similar problem yesterday. I'll address him your comment.

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -200,89 +200,52 @@ object BlockAPI {
...
-        //  e.g. depth=1 should always return latest block created by the node
-        blocksWithActiveName <- blockDataStream
-                                 .map(_.flatten)
-                                 .take(depthWithLimit)
-                                 .compile
-                                 .toList
-                                 .map(_.flatten)
-      } yield (blocksWithActiveName, blocksWithActiveName.length).asRight
-
-    if (depth > maxBlocksLimit)
```
stanislavlyalin, line 231: Why this check was removed?

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
```diff
@@ -200,89 +200,52 @@ object BlockAPI {
...
```
stanislavlyalin, line 273: And here too.

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
```diff
@@ -16,7 +15,7 @@ import coop.rchain.shared.syntax._
 
 import scala.collection.concurrent.TrieMap
 
-class BlockReportAPI[F[_]: Concurrent: Metrics: EngineCell: Log: BlockStore](
+class BlockReportAPI[F[_]: Concurrent: BlockStore: Metrics: Log](
```
stanislavlyalin, line 18: Is there any rule to sort implicits here?

stanislavlyalin (approved):
LGTM. Nice commits splitting! Just added some questions.

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/api/BlockAPI.scala
tgrospic: Good catch! My mistake.

casper/src/main/scala/coop/rchain/casper/MultiParentCasperImpl.scala
tgrospic: Makes sense but the whole file will be removed.

block-storage/src/main/scala/coop/rchain/blockstorage/BlockStoreSyntax.scala
tgrospic: We had this to get source trace on errors because before monix v3 this was not supported so async code was impossible to trace.
https://github.com/monix/monix.io/blob/main/_docs/3x/eval/stacktraces.md

casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
tgrospic: I'm using the rule to first specify execution dependencies like Monad, Sync, Concurrent, Time, ... and at the end less important dependencies like Metrics, Log, ... And all else in the middle.

And with more arguments I found it's useful to disable formatting and add additional comments.
https://github.com/rchain/rchain/blob/65d20ccf0b97bf3f24df18d50ee9c30c8ce0f6e5/casper/src/main/scala/coop/rchain/casper/engine/Running.scala#L249-L255

tgrospic:
bors merge

PR #3677, opened 2022-04-29
Remove signature in insertsigned

### LINKED ISSUE #3669: Remove unneccessary signature in insertSigned
## Overview

Currently when people want to use [insertSigned](https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/Registry.rho#L395) in rholang, they need to provide signature of the object they want to insert. That's not really neccessary because the deploy is already include the signature of the privatekey. This create additional trouble for user to insert things in the registry.

So to insert signed data to the Registry user needs to provide only the data itself with the nonce + DeployerId (public key of the deployer).
https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/casper/src/main/resources/Registry.rho#L393
In this way URI is generated from DeployerId (and stays the same) and can be updated of course with the new nonce (increased).

### Impact

For those who use `insertSigned` should change the arguments they put.

### Dependencies

No.

### Design

1. Remove the sig in  https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/Registry.rho#L395 
2. remove signature verify in https://github.com/rchain/rchain/blob/dev/casper/src/main/resources/Registry.rho#L401

zsluedem:
## Overview

The pr is intended to address #3669 .

After this pr the `insertSigned` would be
https://github.com/zsluedem/rchain/blob/c7d7c945afe4dd633e99f7c694b1025281c0e714/casper/src/main/resources/Either.rho#L264-L269

Previously it is like 

https://github.com/rchain/rchain/blob/ced5a7251bcb8c2f9a995c843d82b12d79845ebd/casper/src/main/resources/Either.rho#L264-L269

### Notes

Breaking changes which people use `insertSigned` should change their codes.


zsluedem:
https://github.com/rchain/rchain/pull/3687 new pr based on dev

PR #3682, opened 2022-05-03
Find deploy method

### LINKED ISSUE #3537: Implement new findDeploy API method
## Overview

Current **limitation** of `/api/deploy/{deployId}` API
1. You can not know whether the deploy is received by the validator if there are some network hiccups happen in sending deploy.
2. You can not know why your deploy is not packed in a block even the sending process is successfully.(valid after block number is not valid)

In short, users find no way to know the real status of a deploy that is not in a block. And **the users want to know that.**

### Impact

No impact.

### Design

Provide a `Status` field in the return of the api.
```scala
final case class FindDeployResponse(
  status: InTransactionPools || ValidAfterNumberOutAge || WaitingForFinalization ||Finalized || NotFinalized(Orphaned)
  data: {
    "deployInfo": deployDataTye,
    "blockInfo": blockData,  // exist only the deploy is packed
  }
)
```

## Solution

- [x] The first part is to expose deploy status method on the API.
  - #3682
  - #3695

- [ ] What's left is to add _Expired_, _Finalized_  and _Rejected_ statuses in which deploy can be.

nzpr:
## Overview
Implementation of logic for <https://github.com/rchain/rchain/issues/3537>
Makes `BlockDagKeyValueStorage` implementation providing status of particular deploy through `deployStatus` method.
To track deploy execution status `ExecutionTracker` trait is supplied to `RuntimeManagerImpl` and connected with  new API methods. `BlockDagKeyValueStorage`

### Notes
Connecting `BlockDagKeyValueStorage.deployStatus` method to external RPC and web API should be done in the followup PR.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
+++ b/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
@@ -23,6 +23,8 @@ import coop.rchain.models.{BlockHash, BlockMetadata, Validator}
 import coop.rchain.shared.syntax._
 import coop.rchain.blockstorage.syntax._
 import coop.rchain.casper.dag.BlockDagKeyValueStorage._
+import coop.rchain.casper.rholang.RuntimeManager.BlockExecutionTracker
+import coop.rchain.rholang.interpreter.EvaluateResult
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 import fs2.Stream
@@ -292,6 +294,16 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 object BlockDagKeyValueStorage {
   implicit private val BlockDagKeyValueStorage_FromFileMetricsSource: Source =
     Metrics.Source(BlockStorageMetricsSource, "dag-key-value-store")
+
+  def executionTracker[F[_]](blockDagStorage: BlockDagKeyValueStorage[F]) =
+    new BlockExecutionTracker[F] {
+      override def execStarted(d: DeployId): F[Unit] =
+        blockDagStorage.commitExecutionStarted(d)
+      override def execComplete(d: DeployId, res: EvaluateResult): F[Unit] = {
+        val err = res.errors.map(_.getMessage).mkString("\n")
+        blockDagStorage.commitExecutionComplete(d, if (err.isBlank) "Success" else err)
+      }
+    }
 
   private final case class DagStores[F[_]](
       metadata: BlockMetadataStore[F],
--- a/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
+++ b/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
@@ -1,6 +1,7 @@
 package coop.rchain.casper.dag
 
-import cats.effect.Concurrent
+import cats.{Alternative, Applicative, Monad}
+import cats.effect.{Concurrent, Sync}
 import cats.effect.concurrent.{Ref, Semaphore}
 import cats.syntax.all._
 import com.google.protobuf.ByteString
@@ -11,7 +12,7 @@ import coop.rchain.blockstorage.dag.codecs._
 import coop.rchain.blockstorage.dag.{BlockDagStorage, BlockMetadataStore, DagRepresentation}
 import coop.rchain.blockstorage.syntax._
 import coop.rchain.blockstorage.util.BlockMessageUtil._
-import coop.rchain.casper.PrettyPrinter
+import coop.rchain.casper.{MultiParentCasper, PrettyPrinter}
 import coop.rchain.casper.protocol.{BlockMessage, DeployData}
 import coop.rchain.crypto.signatures.Signed
 import coop.rchain.metrics.Metrics.Source
@@ -21,8 +22,12 @@ import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, Validator}
 import coop.rchain.shared.syntax._
 import coop.rchain.blockstorage.syntax._
+import coop.rchain.casper.dag.BlockDagKeyValueStorage._
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
+import fs2.Stream
+
+import scala.collection.concurrent.TrieMap
 
 final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
     representationState: Ref[F, DagRepresentation],
@@ -186,6 +191,9 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
                       newDagMsgState
                     )
                 }
+          _ <- removeExpiredFromPool(deployStore, dag).map(
+                _.map((_, ())).map((expiredMap.update _).tupled)
+              )
         } yield dag
       }
     )
@@ -236,6 +244,20 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
 
   // Map of deploys being executed and execution results
   private val execMap    = TrieMap.empty[DeployId, Option[String]]
+  private val expiredMap = TrieMap.empty[DeployId, Unit]
+  def deployStatus(d: DeployId): F[String] =
+    for {
+      finSet <- blockMetadataIndex.finalizedBlockSet
+      r <- BlockDagKeyValueStorage.deployStatus(
+            d,
+            deployStore,
+            deployIndex,
+            invalidBlocksIndex,
+            finSet,
+            execMap.toMap,
+            expiredMap.toMap.keySet
+          )
+    } yield r
 
   def commitExecutionStarted(
       d: DeployId
@@ -304,7 +326,9 @@ object BlockDagKeyValueStorage {
     )
   }
 
-  def create[F[_]: Concurrent: Log: Metrics](kvm: KeyValueStoreManager[F]): F[BlockDagStorage[F]] =
+  def create[F[_]: Concurrent: Log: Metrics](
+      kvm: KeyValueStoreManager[F]
+  ): F[BlockDagKeyValueStorage[F]] =
     for {
       lock   <- MetricsSemaphore.single[F]
       stores <- createStores(kvm)
@@ -339,4 +363,62 @@ object BlockDagKeyValueStorage {
       stores.deployPool,
       stores.invalidBlocks
     )
+
+  def deployStatus[F[_]: Sync](
+      d: DeployId,
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
+      invalidBlocksIndex: KeyValueTypedStore[F, DeployId, BlockMetadata],
+      finalizedSet: Set[BlockHash],
+      executedMap: Map[DeployId, Option[String]],
+      expiredSet: Set[DeployId]
+  ): F[String] = {
+    val pooledF   = deployStore.collect { case (dId, _) => dId }.map(_.contains(d))
+    val expiredF  = Sync[F].delay(expiredSet.contains(d))
+    val includedF = deployIndex.collect { case (dId, _) => dId }.map(_.contains(d))
+    val orphanedF = invalidBlocksIndex.collect { case (dId, _) => dId }.map(_.contains(d))
+    val finalizedF = deployIndex.get1(d).map { hashOpt =>
+      hashOpt.exists(h => finalizedSet.contains(h))
+    }
+    val inProgress = if (executedMap.contains(d)) {
+      val done = executedMap(d).nonEmpty
+      val r =
+        if (done) s"Executed, result: ${executedMap(d).get}"
+        else "Execution is in progress."
+      r.some
+    } else none[String]
+
+    Stream(
+      inProgress.pure,
+      pooledF.map(v => v.guard[Option].as("Pooled")),
+      expiredF.map(v => v.guard[Option].as("Expired")),
+      orphanedF.map(v => v.guard[Option].as("Orphaned")),
+      includedF.map(v => v.guard[Option].as("Included")),
+      finalizedF.map(v => v.guard[Option].as("Finalized"))
+    ).covary[F]
+      .map(Stream.eval)
+      .flatten
+      .collectFirst { case Some(status) => status }
+      .compile
+      .last
+      .map(_.getOrElse("Unknown"))
+  }
+
+  def removeExpiredFromPool[F[_]: Monad](
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      dag: DagRepresentation
+  ): F[List[DeployId]] = {
+    val expiredF = deployStore
+      .collect {
+        case (_, v) =>
+          val d       = v()
+          val expired = dag.latestBlockNumber - d.data.validAfterBlockNumber > MultiParentCasper.deployLifespan
+          expired.guard[Option].as(d)
+      }
+      .map(_.flatten.toList)
+    expiredF.flatMap { v =>
+      val sigs = v.map(_.sig)
+      deployStore.delete(sigs).as(sigs)
+    }
+  }
 }
--- a/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
+++ b/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
@@ -233,6 +233,18 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def addDeploy(d: Signed[DeployData]): F[Unit] = deployStore.put(d.sig, d)
 
   override def pooledDeploys: F[Map[DeployId, Signed[DeployData]]] = deployStore.toMap
+
+  // Map of deploys being executed and execution results
+  private val execMap    = TrieMap.empty[DeployId, Option[String]]
+
+  def commitExecutionStarted(
+      d: DeployId
+  ): F[Unit] = Sync[F].delay { execMap.update(d, none[String]) }
+
+  def commitExecutionComplete(
+      d: DeployId,
+      status: String
+  ): F[Unit] = Sync[F].delay { execMap.update(d, status.some) }
 }
 
 object BlockDagKeyValueStorage {
--- a/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
+++ b/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
@@ -245,19 +245,39 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   // Map of deploys being executed and execution results
   private val execMap    = TrieMap.empty[DeployId, Option[String]]
   private val expiredMap = TrieMap.empty[DeployId, Unit]
-  def deployStatus(d: DeployId): F[String] =
-    for {
-      finSet <- blockMetadataIndex.finalizedBlockSet
-      r <- BlockDagKeyValueStorage.deployStatus(
-            d,
-            deployStore,
-            deployIndex,
-            invalidBlocksIndex,
-            finSet,
-            execMap.toMap,
-            expiredMap.toMap.keySet
-          )
-    } yield r
+
+  def deployStatus(d: DeployId): F[String] = blockMetadataIndex.finalizedBlockSet.flatMap {
+    finalizedSet =>
+      val pooledF   = deployStore.contains(d)
+      val expiredF  = Sync[F].delay(execMap.contains(d))
+      val includedF = deployIndex.contains(d)
+      val orphanedF = invalidBlocksIndex.contains(d)
+      val finalizedF = deployIndex.get1(d).map { hashOpt =>
+        hashOpt.exists(h => finalizedSet.contains(h))
+      }
+      val inProgress = if (execMap.contains(d)) {
+        val done = execMap(d).nonEmpty
+        val r =
+          if (done) s"Executed, result: ${execMap(d).get}"
+          else "Execution is in progress."
+        r.some
+      } else none[String]
+
+      Stream(
+        inProgress.pure,
+        pooledF.map(v => v.guard[Option].as("Pooled")),
+        expiredF.map(v => v.guard[Option].as("Expired")),
+        orphanedF.map(v => v.guard[Option].as("Orphaned")),
+        includedF.map(v => v.guard[Option].as("Included")),
+        finalizedF.map(v => v.guard[Option].as("Finalized"))
+      ).covary[F]
+        .map(Stream.eval)
+        .flatten
+        .collectFirst { case Some(status) => status }
+        .compile
+        .last
+        .map(_.getOrElse("Unknown"))
+  }
 
   def commitExecutionStarted(
       d: DeployId
@@ -363,46 +383,6 @@ object BlockDagKeyValueStorage {
       stores.deployPool,
       stores.invalidBlocks
     )
-
-  def deployStatus[F[_]: Sync](
-      d: DeployId,
-      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
-      deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
-      invalidBlocksIndex: KeyValueTypedStore[F, DeployId, BlockMetadata],
-      finalizedSet: Set[BlockHash],
-      executedMap: Map[DeployId, Option[String]],
-      expiredSet: Set[DeployId]
-  ): F[String] = {
-    val pooledF   = deployStore.collect { case (dId, _) => dId }.map(_.contains(d))
-    val expiredF  = Sync[F].delay(expiredSet.contains(d))
-    val includedF = deployIndex.collect { case (dId, _) => dId }.map(_.contains(d))
-    val orphanedF = invalidBlocksIndex.collect { case (dId, _) => dId }.map(_.contains(d))
-    val finalizedF = deployIndex.get1(d).map { hashOpt =>
-      hashOpt.exists(h => finalizedSet.contains(h))
-    }
-    val inProgress = if (executedMap.contains(d)) {
-      val done = executedMap(d).nonEmpty
-      val r =
-        if (done) s"Executed, result: ${executedMap(d).get}"
-        else "Execution is in progress."
-      r.some
-    } else none[String]
-
-    Stream(
-      inProgress.pure,
-      pooledF.map(v => v.guard[Option].as("Pooled")),
-      expiredF.map(v => v.guard[Option].as("Expired")),
-      orphanedF.map(v => v.guard[Option].as("Orphaned")),
-      includedF.map(v => v.guard[Option].as("Included")),
-      finalizedF.map(v => v.guard[Option].as("Finalized"))
-    ).covary[F]
-      .map(Stream.eval)
-      .flatten
-      .collectFirst { case Some(status) => status }
-      .compile
-      .last
-      .map(_.getOrElse("Unknown"))
-  }
 
   def removeExpiredFromPool[F[_]: Monad](
       deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
--- a/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
+++ b/casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
@@ -8,8 +8,8 @@ import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.dag.BlockDagStorage
 import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
 import coop.rchain.casper.protocol._
-import coop.rchain.casper.rholang.RuntimeManager.{ExecutionTracker, MergeableStore, StateHash}
-import coop.rchain.casper.rholang.syntax.RuntimeSyntax.{SystemTransition, UserTransition}
+import coop.rchain.casper.rholang.RuntimeDeployResult._
+import coop.rchain.casper.rholang.RuntimeManager.{BlockExecutionTracker, MergeableStore, StateHash}
 import coop.rchain.casper.rholang.types.{ReplayFailure, SystemDeploy}
 import coop.rchain.casper.syntax._
 import coop.rchain.crypto.signatures.Signed
@@ -79,7 +79,7 @@ final case class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log: Contex
     historyRepo: RhoHistoryRepository[F],
     mergeableStore: MergeableStore[F],
     mergeableTagName: Par,
-    executionTracker: ExecutionTracker[F]
+    executionTracker: BlockExecutionTracker[F]
 ) extends RuntimeManager[F] {
 
   def spawnRuntime: F[RhoRuntime[F]] =
@@ -113,18 +113,18 @@ final case class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log: Contex
   ): F[(StateHash, Seq[ProcessedDeploy], Seq[ProcessedSystemDeploy])] =
     for {
       runtime  <- spawnRuntime
-      _        <- terms.map(_.sig).toList.traverse(executionTracker.callbackExecutionStarted)
+      _        <- terms.map(_.sig).toList.traverse(executionTracker.execStarted)
       computed <- runtime.computeState(startHash, terms, systemDeploys, blockData)
       _ <- {
         val v = computed._2.map(tx => (tx.deploy.deploy.sig, tx.evalResult))
-        v.toList.traverse((executionTracker.callbackExecutionComplete _).tupled)
+        v.toList.traverse((executionTracker.execComplete _).tupled)
       }
       (stateHash, usrDeployRes, sysDeployRes) = computed
       (usrProcessed, usrMergeable, _) = usrDeployRes
-        .map(UserTransition.unapply(_).get)
+        .map(UserDeployRuntimeResult.unapply(_).get)
         .unzip3
       (sysProcessed, sysMergeable) = sysDeployRes
-        .map(SystemTransition.unapply(_).get)
+        .map(SystemDeployRuntimeResult.unapply(_).get)
         .unzip
 
       // Concat user and system deploys mergeable channel maps
@@ -149,7 +149,7 @@ final case class RuntimeManagerImpl[F[_]: Concurrent: Metrics: Span: Log: Contex
       .flatMap {
         case (preState, stateHash, processed) =>
           val (processedDeploys, mergeableChs, _) =
-            processed.map(UserTransition.unapply(_).get).unzip3
+            processed.map(UserDeployRuntimeResult.unapply(_).get).unzip3
 
           // Convert from final to diff values and persist mergeable (number) channels for post-state hash
           val preStateHash  = preState.toBlake2b256Hash
@@ -262,7 +262,7 @@ object RuntimeManager {
       historyRepo: RhoHistoryRepository[F],
       mergeableStore: MergeableStore[F],
       mergeableTagName: Par,
-      executionTracker: ExecutionTracker[F]
+      executionTracker: BlockExecutionTracker[F]
   ): F[RuntimeManagerImpl[F]] =
     Sync[F].delay(
       RuntimeManagerImpl(
@@ -279,7 +279,7 @@ object RuntimeManager {
       store: RSpaceStore[F],
       mergeableStore: MergeableStore[F],
       mergeableTagName: Par,
-      executionTracker: ExecutionTracker[F]
+      executionTracker: BlockExecutionTracker[F]
   )(
       implicit ec: ExecutionContext
   ): F[RuntimeManagerImpl[F]] =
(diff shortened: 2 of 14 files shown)


casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -200,9 +202,23 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
     deployIndex.get1(deployId)
 
-  override def addDeploy(d: Signed[DeployData]): F[Unit] = deployStore.put(d.sig, d)
+  override def addDeploy(d: Signed[DeployData]): F[Unit] =
+    deployStore.put(d.sig, d).as(vabnMap.update(d.sig, d.data.validAfterBlockNumber))
 
   override def pooled: F[Map[DeployId, Signed[DeployData]]] = deployStore.toMap
+
+  import coop.rchain.models.syntax._
+  val vabnMap = mutable.TreeMap.empty[DeployId, Long]
```
tgrospic: TreeMap is not thread safe, TrieMap can be used instead. Please also define member variables on the top of the class.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +313,34 @@ object BlockDagKeyValueStorage {
       stores.deployPool,
       stores.invalidBlocks
     )
+
+  def deployStatus[F[_]: Sync](
+      d: DeployId,
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      getRepresentation: F[DagRepresentation],
```
tgrospic: Don't accept F values as input to functions.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +313,34 @@ object BlockDagKeyValueStorage {
...
+      deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
+      vabnOpt: DeployId => Option[Long],
```
tgrospic: What does it mean optional result for vabn?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +313,34 @@ object BlockDagKeyValueStorage {
...
+      invalidBlocksIndex: KeyValueTypedStore[F, DeployId, BlockMetadata],
+      finalizedSet: F[Set[BlockHash]],
```
tgrospic: Same here for F input values.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +313,34 @@ object BlockDagKeyValueStorage {
...
+      executedMap: Map[DeployId, Seq[DeployId]]
```
tgrospic: Is this map used?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: I used inmem map to track VABN, so to not read deploy store each time.
I adjusted logic to not use it bit read after each block is added, so this is removed <https://github.com/rchain/rchain/pull/3682/commits/51f96ca8ba170ca4f4c2c8d6ac9aae41949b8cf3>.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: Yes it is 
https://github.com/rchain/rchain/blob/51f96ca8ba170ca4f4c2c8d6ac9aae41949b8cf3/casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala#L349-L355

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -21,9 +22,13 @@ import coop.rchain.models.Validator.Validator
 import coop.rchain.models.{BlockHash, BlockMetadata, Validator}
 import coop.rchain.shared.syntax._
 import coop.rchain.blockstorage.syntax._
+import coop.rchain.casper.dag.BlockDagKeyValueStorage._
 import coop.rchain.shared.{Log, LogSource}
 import coop.rchain.store.{KeyValueStoreManager, KeyValueTypedStore}
 
+import scala.collection.concurrent.TrieMap
+import scala.collection.mutable
```
stanislavlyalin: It looks outdated.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -203,6 +213,28 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def addDeploy(d: Signed[DeployData]): F[Unit] = deployStore.put(d.sig, d)
 
   override def pooledDeploys: F[Map[DeployId, Signed[DeployData]]] = deployStore.toMap
+
+  val executedMap = TrieMap.empty[DeployId, (Seq[DeployId], Option[String])]
+  val expiredMap  = TrieMap.empty[DeployId, Unit]
```
stanislavlyalin: Consider to make it private.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
```diff
@@ -22,6 +22,8 @@ trait BlockDagStorage[F[_]] {
   def lookupByDeployId(blockHash: DeployId): F[Option[BlockHash]]
   def addDeploy(d: Signed[DeployData]): F[Unit]
   def pooledDeploys: F[Map[DeployId, Signed[DeployData]]]
+  def deployStatus(d: DeployId): F[String]
+  def commitDeployReplayed(d: DeployId, pack: Seq[DeployId], err: Option[String]): F[Unit]
```
stanislavlyalin: What are the purpose of this method? Can you add descriptions to new functions introduced in the PR please. It is not clear for me even after small reading the code.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +329,52 @@ object BlockDagKeyValueStorage {
...
+
+  def removeExpiredFromPool[F[_]: Monad](
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      dag: DagRepresentation
+  ): F[List[DeployId]] = {
+    val expiredF = deployStore.toMap
+      .map(_.filter {
+        case (_, d) =>
+          dag.latestBlockNumber - d.data.validAfterBlockNumber > MultiParentCasper.deployLifespan
+      }.map { case (_, d) => d }.toList)
```
stanislavlyalin: I just want to clarify. Is this code in `F` context? It looks outside but value wrapped into `F` is like a function which can be a part of a formula?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +329,52 @@ object BlockDagKeyValueStorage {
...
+        if (errOpt.isDefined) s"erred with ${errOpt.get}" else "is executed succesfully."
+      s"Deploy ${statusStr}. Block execution is in progress, ${remaining} deploys remaining to pack in a block.".some
+    } else none[String]
+    for {
+      _ <- inProgress.pure
+      _ <- pooledF.map(v => v.guard[Option].as("Pooled"))
+      _ <- expiredF.map(v => v.guard[Option].as("Expired"))
+      _ <- orphanedF.map(v => v.guard[Option].as("Orphaned"))
+      _ <- includedF.map(v => v.guard[Option].as("Included"))
+      r <- finalizedF.map(v => v.guard[Option].as("Finalized"))
```
stanislavlyalin: Why this statuses not combined but just selected last of them?

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -46,8 +48,9 @@ trait RuntimeManager[F[_]] {
       systemDeploys: Seq[ProcessedSystemDeploy],
       blockData: BlockData,
       invalidBlocks: Map[BlockHash, Validator],
+      logResult: Option[(BlockDagStorage.DeployId, EvaluateResult) => F[Unit]] = None,
       isGenesis: Boolean
-  ): F[Either[ReplayFailure, StateHash]]
+  )(implicit deployExecutionTracker: DeployExecutionTracker[F]): F[Either[ReplayFailure, StateHash]]
```
stanislavlyalin: Why implicit? To minimize changes when a function is called?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: As `deployStore.toMap` is F[Map], the whole function is inside F

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: Thanks for the question, actually I did it wrong here.

What I wanted to do here is to flatten multiple Options to return first Some and short circuit. But I did 2 mistakes here: 1) these options are inside F context, so I'm flattening F[_], not Option[_]. So I had to wrap values inside OptionT, like `OptionT(inProgress.pure)`2) But Even if options would be not inside F[_], or I use OptionT - still flattening of multiple Options will return None or latest Some in the chain.
```scala
Some(1).flatMap(_ => None.flatMap(_ => Some(3))) = None: Option[Int]
```

As for comprehension is just a syntax sugar for `flatMap`, you can use `for {}` for any monad. Option and Either are Monads, so you can combine results (flatten) them this way. Another example of using such a trick is <https://github.com/rchain/rchain/blob/97886a864d6cf0329c539a83f1ed053969e53d24/casper/src/main/scala/coop/rchain/casper/Validate.scala#L141-L158>.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
nzpr: Yes, in RuntimeManagerTest there are bunch of calls, so with creating a single implicit at the beginning of the file you can call it a day.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3682/commits/2a738a4aed9c67693797bbc305671839f57d7e58

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3682/commits/2a738a4aed9c67693797bbc305671839f57d7e58

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -297,4 +329,58 @@ object BlockDagKeyValueStorage {
...
+      .compile
+      .last
+      .map(_.getOrElse("Unknown"))
+  }
+
+  def removeExpiredFromPool[F[_]: Monad](
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      dag: DagRepresentation
+  ): F[List[DeployId]] = {
+    val expiredF = deployStore.toMap
```
tgrospic: There is collect function defined for KVTS which don't require the whole store be loaded immediately but for each key.
https://github.com/rchain/rchain/blob/674009b65ab6488bbc101aa85c6b7a87df6c305d/shared/src/main/scala/coop/rchain/store/KeyValueTypedStore.scala#L12-L17

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeReplaySyntax.scala
```diff
@@ -92,28 +97,39 @@ final class RuntimeReplayOps[F[_]: Sync: Span: Log](
...
-      replaySystemDeploy: ProcessedSystemDeploy => F[Either[ReplayFailure, NumberChannelsEndVal]]
+      replayDeploy: ProcessedDeploy => F[
+        Either[ReplayFailure, (EvaluateResult, NumberChannelsEndVal)]
+      ],
+      replaySystemDeploy: ProcessedSystemDeploy => F[Either[ReplayFailure, NumberChannelsEndVal]],
+      deployExecutionTracker: DeployExecutionTracker[F]
   ): F[Either[ReplayFailure, (Blake2b256Hash, Vector[NumberChannelsEndVal])]] = {
-    type Params[D] = (Seq[D], Vector[NumberChannelsEndVal])
+    type Params[D] = (List[D], Vector[NumberChannelsEndVal])
+    import deployExecutionTracker._
```
tgrospic: Please don't use this pattern to import all fields from input arguments. It makes harder to read the code because you just using functions available in the scope of the function but it's not obvious immediately.

If you want to use import at lest don't use wildcard.

casper/src/main/scala/coop/rchain/casper/rholang/InterpreterUtil.scala
```diff
@@ -124,13 +126,32 @@ object InterpreterUtil {
...
+        deployExecutionTracker = new DeployExecutionTracker[F] {
+          // TODO add started callback
+          override def callbackReplayStarted(d: DeployId): F[Unit] = ().pure[F]
+          override def callbackReplayFinished(d: DeployId, res: EvaluateResult): F[Unit] = {
+            val errOpt =
+              if (res.errors.isEmpty) none[String]
+              else res.errors.map(_.getMessage).mkString("\n").some
+            BlockDagStorage[F].commitDeployReplayed(d, internalDeploys.map(_.deploy.sig), errOpt)
+          }
+        }
```
tgrospic: DeployExecutionTracker doesn't depend on it's internal state and uses global state inside, so it can be created once and supplied to RuntimeManager implementation.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -63,7 +63,15 @@ trait RuntimeSyntax {
...
   type SysEvalResult[S <: SystemDeploy] = (Either[SystemDeployUserError, S#Result], EvaluateResult)
-}
+  final case class UserTransition(
+      deploy: ProcessedDeploy,
+      mergeable: NumberChannelsEndVal,
+      evalResult: EvaluateResult
+  )
+  final case class SystemTransition(
+      deploy: ProcessedSystemDeploy,
+      mergeable: NumberChannelsEndVal
```
nzpr: These case classes serve as wrappers for output of deploy execution. As system deploys do not provide evaluation result, there are two of them. Would be good to unify this in future, making processed deploy a single thing.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeReplaySyntax.scala
```diff
@@ -41,19 +41,20 @@ import coop.rchain.rspace.hashing.Blake2b256Hash
...
-  ): RuntimeReplayOps[F] =
-    new RuntimeReplayOps[F](runtime)
+  ): RuntimeReplayOps[F] = new RuntimeReplayOps[F](runtime)
 }
 
-final class RuntimeReplayOps[F[_]: Sync: Span: Log](
-    private val runtime: ReplayRhoRuntime[F]
-) extends RuntimeSyntax {
-
+object RuntimeReplaySyntax {
```
tgrospic, line 53: Why is this object needed?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -52,33 +53,42 @@ import coop.rchain.rspace.hashing.{Blake2b256Hash, StableHashProvider}
...
+  )
```
tgrospic: Please move these definition in a separate file. Better to keep extension definitions clean.

Also you are now introducing _transition_ in naming which is not mention anywhere else so it's difficult to understand what is transitioning. How _user_ is transitioning?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -291,17 +298,17 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
         _ <- (runtime.revertToSoftCheckpoint(fallback) *>
               printDeployErrors(deploy.sig, evaluateResult.errors)).whenA(!evalSucceeded)
 
-      } yield (deployResult, evaluateResult.mergeable)
+      } yield (deployResult, evaluateResult)
     }
```
nzpr, line 295: We had the whole result previously but were returning only mergeable channels. So making evaluator return result is only this change actually.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -285,4 +324,16 @@ object RuntimeManager {
       scodec.codecs.bytes,
       deployMergeableDataSeqCodec
     )
+
+  // TODO add more callbacks (system deploys, maybe something else)
+  trait ExecutionTracker[F[_]] {
+    def callbackExecutionStarted(d: DeployId): F[Unit]
+    def callbackExecutionComplete(d: DeployId, res: EvaluateResult): F[Unit]
```
tgrospic: Not necessary to prefix these functions with _callback_ or _execution_. Also they can accept list of deploy ids.

Please rename this interface to `BlockExecutionTracker` so it's more specific for its purpose.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -186,6 +191,9 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
                       newDagMsgState
                     )
                 }
+          _ <- removeExpiredFromPool(deployStore, dag).map(
+                _.map((_, ())).map((expiredMap.update _).tupled)
+              )
```
tgrospic, line 198: Maybe we can remove expired deploys when deploy storage is read when block is created. In that way logic for adding block and expired deploys will not be mixed.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -327,4 +363,62 @@ object BlockDagKeyValueStorage {
...
+      deployStore: KeyValueTypedStore[F, DeployId, Signed[DeployData]],
+      deployIndex: KeyValueTypedStore[F, DeployId, BlockHash],
+      invalidBlocksIndex: KeyValueTypedStore[F, DeployId, BlockMetadata],
+      finalizedSet: Set[BlockHash],
+      executedMap: Map[DeployId, Option[String]],
+      expiredSet: Set[DeployId]
+  ): F[String] = {
+    val pooledF   = deployStore.collect { case (dId, _) => dId }.map(_.contains(d))
+    val expiredF  = Sync[F].delay(expiredSet.contains(d))
+    val includedF = deployIndex.collect { case (dId, _) => dId }.map(_.contains(d))
```
tgrospic: Use get by key not iterate over all records.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -327,4 +363,62 @@ object BlockDagKeyValueStorage {
...
+    val orphanedF = invalidBlocksIndex.collect { case (dId, _) => dId }.map(_.contains(d))
```
tgrospic: The same hare.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -327,4 +363,62 @@ object BlockDagKeyValueStorage {
...
+    val finalizedF = deployIndex.get1(d).map { hashOpt =>
```
tgrospic: This is the same as _includedF_.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -327,4 +363,62 @@ object BlockDagKeyValueStorage {
       stores.deployPool,
       stores.invalidBlocks
     )
+
+  def deployStatus[F[_]: Sync](
```
tgrospic: Why is this method defined here on companion object and not directly as member on BlockDagKeyValueStorage?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -233,6 +241,32 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def addDeploy(d: Signed[DeployData]): F[Unit] = deployStore.put(d.sig, d)
 
   override def pooledDeploys: F[Map[DeployId, Signed[DeployData]]] = deployStore.toMap
+
+  // Map of deploys being executed and execution results
+  private val execMap    = TrieMap.empty[DeployId, Option[String]]
+  private val expiredMap = TrieMap.empty[DeployId, Unit]
+  def deployStatus(d: DeployId): F[String] =
```
stanislavlyalin: Why is `deployStatus` not part of the `BlockDagStorage` trait?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
stanislavlyalin: `deployStatus` is only used in `BlockDagKeyValueStorage`. Why not make it a private member of this class? Then there is no need to pass `execMap` and `expiredMap`. In fact, many function parameters are fields of the `BlockDagKeyValueStorage` class.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -327,4 +363,62 @@ object BlockDagKeyValueStorage {
...
+    } else none[String]
+
+    Stream(
+      inProgress.pure,
+      pooledF.map(v => v.guard[Option].as("Pooled")),
+      expiredF.map(v => v.guard[Option].as("Expired")),
+      orphanedF.map(v => v.guard[Option].as("Orphaned")),
+      includedF.map(v => v.guard[Option].as("Included")),
+      finalizedF.map(v => v.guard[Option].as("Finalized"))
+    ).covary[F]
```
stanislavlyalin: Could you explain this piece of code, please?

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -285,4 +324,16 @@ object RuntimeManager {
       scodec.codecs.bytes,
       deployMergeableDataSeqCodec
     )
+
+  // TODO add more callbacks (system deploys, maybe something else)
+  trait ExecutionTracker[F[_]] {
```
stanislavlyalin: Why are `ExecutionTracker` and `noOpExecutionTracker` placed in the `RuntimeManager`? Is it his responsibility? Its responsibility is not clear from the name and set of methods. Many things for tests are initialized in `engine/Setup.scala`. Maybe move `noOpExecutionTracker` there?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -233,6 +241,32 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
...
+    } yield r
+
+  def commitExecutionStarted(
+      d: DeployId
+  ): F[Unit] = Sync[F].delay { execMap.update(d, none[String]) }
+
+  def commitExecutionComplete(
+      d: DeployId,
+      status: String
+  ): F[Unit] = Sync[F].delay { execMap.update(d, status.some) }
```
stanislavlyalin, line 291: I would prefer to put the `commitExecutionStarted` and `commitExecutionComplete` functions into an trait. This will make the code more consistent and make it easy to put another implementation of `BlockDagStorage`, such as mock.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeReplaySyntax.scala
nzpr: You cannot define `implicit val RuntimeMetricsSource` in class `RuntimeReplayOps` if it extends AnyVal.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
nzpr: Do you have a suggestion how to name this case class? It is a data about state transition done by user, as it contain processed user deploy and metadata about this processed deploy.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: You still have to leave here update of expired map, to be able to detect that your deploy is expired before the next block is created (which can be a long time). So I'm not sure why to remove action on deploy storage here.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: I replaced with `contains` directly on KVStore, not sure how I overlooked this API

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: Not sure wdy mean - it is already `get`, isn't it?

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: My thinking is that this is easier to test. But ok, I moved it.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: As @tgrospic specified, we don't want this to be on a trait but just a part of one implementation (BlockDagKeyValueStore). I think it is because providing deploy statuses is not necessary feature of all block dag storages.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3682#discussion_r869618823, but ok I'll move it

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: There are several checks to be performed - each check returns whether it is found that deploy is in some status.
`pooledF` returns F[Boolean], 
`v.guard[Option].as("Pooled")` returns Some("Pooled") if v is true and None if v is false. 

`guard[F[_]]` is a syntax of an `Alternative` type class, and `cats` library has implementation of an `Alternative[Option]`

So each check returns F[None] or F[Some(*deploy status string*)].

Then all these executions are packed as a Steam that represent a sequence of computations. What I want to do is evaluate all these computations (see `.map(Stream.eval)`) until first Some is found (see `.collectFirst { case Some(status) => status }`). This ensures that computation short circuits, so Stream execution stops and if `inProgress.pure` returns `Some(something)` nothing else is executed.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
stanislavlyalin: I guess in both cases (includedF, finalizedF) you can use similar code
```scala
val includedF = deployIndex.get1(d).map(_.isDefined)
```

```scala
val finalizedF = deployIndex.get1(d).map ...
```

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
nzpr: If it is residing on Runtime manager, why would Block be mentioned here? Should then I move trait into another place? Also ptal https://github.com/rchain/rchain/pull/3682#discussion_r869580232

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
nzpr: My logic was that runtime manager is responsible for controlling runtimes and executions done by these runtimes, and this trait is about tracking executions, providing callbacks that runtime manager can use to speak with outside world.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: https://github.com/rchain/rchain/pull/3682#discussion_r869619896
I think these are functions just for this particular implementation.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
nzpr: I've adjusted in https://github.com/rchain/rchain/pull/3682/commits/41051c4d46479ac1a4a5fda23f7f68102d5f66e2
 this to `val includedF = deployIndex.contains(d)`

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -285,4 +324,16 @@ object RuntimeManager {
       scodec.codecs.bytes,
       deployMergeableDataSeqCodec
     )
+
+  // TODO add more callbacks (system deploys, maybe something else)
```
tgrospic: What is the purpose of this comment or for who? We have issues and RCHIPs to track new features. Please don't overload TODO for everything.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
tgrospic: Deploy is already mentioned in each function which is part of the block.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
tgrospic: Please keep in mind distinction between RuntimeManager trait and its implementation. @nzpr you are describing the interface and implementation can have connection to whatever is needed.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
tgrospic: `ProcessedDeployResult` or `RuntimeDeployResult`.

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
```diff
@@ -124,11 +126,24 @@ object Setup {
...
       runtimeManagerWithHistory <- {
         implicit val sp = span
+        val executionTracker = new ExecutionTracker[F] {
+          override def callbackExecutionStarted(d: DeployId): F[Unit] =
+            blockDagStorage.commitExecutionStarted(d)
+          override def callbackExecutionComplete(d: DeployId, res: EvaluateResult): F[Unit] = {
+            val err = res.errors.map(_.getMessage).mkString("\n")
+            blockDagStorage.commitExecutionComplete(d, if (err.isBlank) "Success" else err)
+          }
+        }
```
tgrospic: Please extract creation of ExecutionTracker to a separate method. Implementation objects for BlockDagStorage or RuntimeManager seems like a good place.

node/src/test/scala/coop/rchain/node/mergeablity/ComputeMerge.scala
```diff
@@ -61,7 +62,9 @@ trait ComputeMerge {
             baseDeploysRes <- baseDeploySources.toList.traverse(
                                runtime.processDeployWithMergeableData
                              )
-            (baseDeploys, baseMergeChs) = baseDeploysRes.unzip
+            (baseDeploys, baseMergeChs, _) = baseDeploysRes
+              .map(UserTransition.unapply(_).get)
```
tgrospic: Note: Not a good practice to extract value from Option with get call. In this case we know it will always succeed but it makes bad example.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
nzpr: Deploy is a signed state transition that user wants to execute, so imo it's natural for runtime manager.

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
stanislavlyalin: @tgrospic, why deployStatus is a part of implementation? Now I am binding deployStatus to WebAPI and I'm forced to perform an illogical conversion in BlockAPI
```scala
override def deployStatus(deployId: DeployId): F[String] =
  BlockDagStorage[F].asInstanceOf[BlockDagKeyValueStorage[F]].deployStatus(deployId)
```

node/src/test/scala/coop/rchain/node/mergeablity/ComputeMerge.scala
nzpr: I know that but what is another way to unzip sequence of tuples wrapped in a case class?

node/src/main/scala/coop/rchain/node/runtime/Setup.scala
nzpr: https://github.com/rchain/rchain/pull/3682/commits/64b0ae190c4a5d6d2c1b0c13d74c75a10566b076#diff-ebd696254b3fec517fbfe8e1cec6930aeb0652d74d79edcc5bff57c2262d5118R132

node/src/test/scala/coop/rchain/node/mergeablity/ComputeMerge.scala
tgrospic: I don't know anything better then explicitly deconstruct.
```scala
(baseDeploys, baseMergeChs, _) = baseDeploysRes
  .map { case UserDeployRuntimeResult(deploy, mergeable, evalResult) => (deploy, mergeable, evalResult) }
  .unzip3
```

tgrospic (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
```diff
@@ -233,12 +243,68 @@ final class BlockDagKeyValueStorage[F[_]: Concurrent: Log] private (
   override def addDeploy(d: Signed[DeployData]): F[Unit] = deployStore.put(d.sig, d)
 
   override def pooledDeploys: F[Map[DeployId, Signed[DeployData]]] = deployStore.toMap
+
+  // Map of deploys being executed and execution results
+  private val execMap    = TrieMap.empty[DeployId, Option[String]]
+  private val expiredMap = TrieMap.empty[DeployId, Unit]
+
+  def deployStatus(d: DeployId): F[String] = blockMetadataIndex.finalizedBlockSet.flatMap {
```
zsluedem, line 251: Could we return `F[someType]` instead of `F[String]`? String is too wide.

stanislavlyalin: This is a purpose of my next PR so don't worry about it.

zsluedem (approved):
LGTM, then

nzpr:
bors merge

nzpr:
bors merge

casper/src/main/scala/coop/rchain/casper/dag/BlockDagKeyValueStorage.scala
tgrospic: `deployStatus` which returns string is just temporary functions and should not be on block dag storage at all.

We need to see what is needed for BlockApi to return deploy status and add just that to block dag storage.

PR #3684, opened 2022-05-04
Block metadata refactoring + removal of invalid blocks from runtime

tgrospic:
## Overview

This PR is another preparation for the new multi-parent finalizer. I discovered another issue with block validation of justification regression which is much easier to implement with the new finalizer state (DagMessageState) so that's why it's included in this PR but not yet fully used.

Slashing is simplified by directly using validator public key so invalid blocks are removed from runtime. Existing slashing mechanism will use creator public key which means node will slash itself if calculation is wrong (WTF?).

Leftover from approved protocol related messages are removed, still approved block is used to get LFB but it will be replaced with fringe request/response messages in subsequent PRs.


models/src/main/protobuf/CasperMessage.proto
```diff
@@ -269,3 +251,21 @@ message StoreItemsMessageProto {
...
+  int64 blockNum                             = 2;
+  bytes sender                               = 3;
+  int32 seqNum                               = 4;
+  repeated JustificationProto justifications = 5 [(scalapb.field).collection_type="collection.immutable.List"];
+  repeated BondProto bonds                   = 6 [(scalapb.field).collection_type="collection.immutable.List"];
+  bool invalid                               = 7; // whether the block was marked as invalid
+
+  bool directlyFinalized                     = 20; // whether the block has been last finalized block (LFB)
+  bool finalized                             = 21; // whether the block is finalized
+  repeated bytes parents                     = 22 [(scalapb.field).collection_type="collection.immutable.List"];
```
tgrospic, line 243: These will be removed in subsequent PRs and fringe will be added.

models/src/main/protobuf/CasperMessage.proto
```diff
@@ -107,23 +107,6 @@ message BlockHashMessageProto {
   bytes  blockCreator = 2;
 }
 
-message BlockMetadataInternal {
```
nzpr, line 110: Would be nice at some point to leave in CasperMessage only really messages related to Casper, and move everything else to another file.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/DagMessageState.scala
```diff
@@ -0,0 +1,188 @@
+package coop.rchain.blockstorage.dag
+
+import cats.syntax.all._
+
+// DagMessageState represents state of one validator in the network
+final case class DagMessageState[M: Ordering, S: Ordering](
```
nzpr, line 6: Will there be other states that describes local copy of a DAG? I'm trying to understand why the is not just DagState

block-storage/src/main/scala/coop/rchain/blockstorage/dag/DagMessageState.scala
```diff
@@ -0,0 +1,188 @@
...
+    }
+    /* End Debug log */
+
+    newMsgView
+  }
+
+  /**
+    * Inserts a message to sender's state
+    */
+  def insertMsgView(msgView: Message[M, S]): DagMessageState[M, S] =
```
nzpr: Might be it would be better to name methods as `createMsg`, `insertMsg`, like `createMsgAndUpdateSender` below. As you work with `Message` case class.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/Finalizer.scala
```diff
@@ -0,0 +1,183 @@
...
+    LazyList.unfold(f(init)) { iter =>
+      if (!iter.hasNext) none
+      else {
+        val x = iter.next()
+        Some((x, iter ++ f(x)))
+      }
+    }
+
+  // Iterate self parent messages
+  def selfParents(mv: Message[M, S], finalized: Set[Message[M, S]]): Seq[Message[M, S]] =
```
nzpr, line 54: Just to make sure I understand correctly - this produces self parents only inside conflict set? So this finalized set along with filtering in `unfold` fuction works as a traversal stopper?

casper/src/main/resources/PoS.rhox
```diff
@@ -431,60 +431,50 @@ in {
...
-                                            "randomNumbers"   : state.get("randomNumbers")
-                                          },
-                                          (true, Nil))
-                                        }
-                                      }
-                                    }
+                          for (@state, stateUpdateCh <- stateProcessCh) {
+                            new valBondCh, valRewardCh, posAuthKeyCh in {
+                              // Flow of logic:
+                              // 0. check if blockHash is the hash of an invalid block
```
nzpr: Given blockHash is removed should this comments be adjusted as well?

casper/src/main/resources/PoS.rhox
```diff
@@ -431,60 +431,50 @@ in {
               new commitRewards in {
                 // Private method which transfers all of a validator's rewards + bond to the Coop vault.
                 // Slash expects (Boolean, Either[Nil, String]) on returnCh.
-                contract PoS(@"slash", @deployerId, @blockHash, @sysAuthToken, returnCh) = {
```
nzpr, line 434: Do you have an idea why these two params removed were introduced here previously?

models/src/main/protobuf/CasperMessage.proto
tgrospic: I tried that, but BlockMetadata has dependency on types from CasperMessage. If multiple generators are specified they will include all dependency and generate duplicates.
https://github.com/rchain/rchain/blob/069b4093a47d9545528224b4e025817896ce27c0/build.sbt#L243-L246

We should not have internal types defined in the same way as protocol or external types.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/DagMessageState.scala
tgrospic: I renamed everywhere to just use a message instead of message view.

block-storage/src/main/scala/coop/rchain/blockstorage/dag/Finalizer.scala
tgrospic: Yes, all calculation for the next fringe should never go below existing fringe. As opposite what we have now where we need to download approved block justifications + self parent for each of them (which I also plan to remove).
https://github.com/rchain/rchain/blob/ecddeba9694184da11acf6cf4c7f1fba60ee6d86/casper/src/main/scala/coop/rchain/casper/engine/LfsBlockRequester.scala#L78-L89

casper/src/main/resources/PoS.rhox
tgrospic: I don't know. But I can add this to another chapter to my book _How to not build the blockchain_. :smile:

I'm guessing it's to validate slashing in replay in some way but that was not implemented at all.

What is even worse, if invalid blocks are not set correctly, node will use it own public key and slash itself. (WTF?)

nzpr (approved):
LGTM

tgrospic:
bors ping

tgrospic:
bors try

tgrospic:
bors try

tgrospic:
bors try

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -422,67 +381,36 @@ object Validate {
...
-                BlockStatus.justificationRegression.asLeft.asRight.pure,
-                Applicative[F].pure(Left(tail))
-              )
-            }
+  def checkJustificationRegression[F[_]: Monad: BlockDagStorage](
+      b: BlockMessage
+  ): F[Option[Boolean]] =
+    for {
+      msgMap <- BlockDagStorage[F].getRepresentation.map(_.dagMessageState.msgMap)
+//      justifications = b.justifications.map(_.latestBlockHash).map(msgMap)
```
zsluedem, line 398: Leftover?

zsluedem (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/Validate.scala
tgrospic: It's related to below TODO comment. All justification should be present in `msgMap` to check justification regression. This code should be used to get justifications instead.

tgrospic:
bors merge

PR #3687, opened 2022-05-09
Remove signature in insertSigned

### LINKED ISSUE #3669: Remove unneccessary signature in insertSigned

zsluedem:
## Overview

The pr is intended to address #3669 .

After this pr the `insertSigned` would be
https://github.com/zsluedem/rchain/blob/c7d7c945afe4dd633e99f7c694b1025281c0e714/casper/src/main/resources/Either.rho#L264-L269

Previously it is like 

https://github.com/rchain/rchain/blob/ced5a7251bcb8c2f9a995c843d82b12d79845ebd/casper/src/main/resources/Either.rho#L264-L269

Currently we only need:
```
          rs!(
            (9223372036854775807, bundle+{*RevVault}),
            *deployerId,
            *uriOut
          )
```

### Notes

Breaking changes which people use `insertSigned` should change their codes.

### DIFF
--- a/casper/src/main/resources/AuthKey.rho
+++ b/casper/src/main/resources/AuthKey.rho
@@ -19,7 +19,8 @@ new
   rs(`rho:registry:insertSigned:secp256k1`),
   uriOut,
   _authKey,
-  stdout(`rho:io:stdout`)
+  stdout(`rho:io:stdout`),
+  deployerId(`rho:rchain:deployerId`)
 in {
 
   // An `AuthKey` is a means for creating a safe authentication token based on a process / unfograble name
@@ -68,9 +69,8 @@ in {
   } |
 
   rs!(
-    "04f4b4417f930e6fab5765ac0defcf9fce169982acfd046e7c27f9b14c0804014623c0439e5c8035e9607599a549303b5b6b90cd9685e6965278bddca65dac7510".hexToBytes(),
     (9223372036854775807, bundle+{*AuthKey}),
-    "304402205d2dc163a40804a445237ef6bd181249a68598d146e66d0c96484c6cfeaaaffd02205e537f1e7bae82b99abbdc25c8472c44f5848085a3a3f23357507c05524be4a7".hexToBytes(),
+    *deployerId,
     *uriOut
   )
 }
--- a/casper/src/main/resources/Registry.rho
+++ b/casper/src/main/resources/Registry.rho
@@ -7,8 +7,6 @@ new _registryStore,
     bootstrapInsertSigned(`rho:registry:insertSigned:secp256k1`),
     buildUri,
     ops(`rho:registry:ops`),
-    secpVerify(`rho:crypto:secp256k1Verify`),
-    blake2b256(`rho:crypto:blake2b256Hash`),
     TreeHashMap
 in {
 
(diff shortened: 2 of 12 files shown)


casper/src/main/resources/AuthKey.rho
```diff
@@ -68,9 +69,8 @@ in {
   } |
 
   rs!(
-    "04f4b4417f930e6fab5765ac0defcf9fce169982acfd046e7c27f9b14c0804014623c0439e5c8035e9607599a549303b5b6b90cd9685e6965278bddca65dac7510".hexToBytes(),
     (9223372036854775807, bundle+{*AuthKey}),
-    "304402205d2dc163a40804a445237ef6bd181249a68598d146e66d0c96484c6cfeaaaffd02205e537f1e7bae82b99abbdc25c8472c44f5848085a3a3f23357507c05524be4a7".hexToBytes(),
+    *deployerId,
```
tgrospic, line 73: Nice improvement!

---

Additional task can be to simplify deployers with only one private key for all blessed contract. I'm not sure why are keys created for each contract?!
https://github.com/rchain/rchain/blob/643fd5f3e277cd29d069c263002526c52b3ad248/casper/src/main/scala/coop/rchain/casper/genesis/contracts/StandardDeploys.scala#L32-L55

casper/src/main/resources/AuthKey.rho
```diff
@@ -19,7 +19,8 @@ new
   rs(`rho:registry:insertSigned:secp256k1`),
   uriOut,
   _authKey,
-  stdout(`rho:io:stdout`)
+  stdout(`rho:io:stdout`),
```
tgrospic, line 22: `stdout` can be removed because it's not used.

casper/src/test/resources/RegistryTest.rho
```diff
@@ -42,7 +43,6 @@ match (
           ("InsertArbitrary and lookup should work", *test_insertArbitraryAndLookup),
           ("InsertArbitrary should work with increased nonce", *test_insertSignedValidNonce),
           ("InsertArbitrary should return Nil for invalid nonce", *test_insertSignedInvalidNonce),
```
tgrospic: Please correct text description for these two tests. They are testing Insert signed not arbitrary.

---

Please also add test for nonce with `Long.MaxValue + 1` which should still be incorrect so we have the same situation tested as with blessed contracts.

casper/src/main/resources/AuthKey.rho
zsluedem: InsertSigned would generate an `URI` based on the pubkey. Use **one key** only would result in the same `URI`.

zsluedem: What's your idea with that? Could this be another pr? I want to simplify changes so that I don't have to deal with heavy conflict in https://github.com/rchain/rchain/pull/3485

casper/src/main/resources/AuthKey.rho
zsluedem: It is removed in https://github.com/rchain/rchain/pull/3485. In order to minimize conflicts. I would just leave it.

casper/src/main/resources/AuthKey.rho
tgrospic: NVM, we need separate keys for blessed deploys.

casper/src/main/resources/Registry.rho
```diff
@@ -392,49 +392,38 @@ in {
   bootstrapInsertSigned!(*insertSignedCh) | // this will work only once
   for (insertSigned <- insertSignedCh) {
 
-    contract insertSigned(@pubKeyBytes, @value, @sig, ret) = {
-      match value {
-        (nonce, data) => {
-          new uriCh, hashCh, verifyCh in {
-            blake2b256!((nonce, data).toByteArray(), *hashCh) |
-            for (@hash <- hashCh) {
-              secpVerify!(hash, sig, pubKeyBytes, *verifyCh) |
```
tgrospic, line 401: Reference to _blake2b256_ and _secpVerify_ can be removed as not used anymore.

tgrospic (approved):
Excellent!

nzpr (approved):
LGTM

zsluedem:
bors merge

PR #3693, opened 2022-05-11
Block message version 2 (hash & validation)

### LINKED ISSUE #3788: Block hashing doesn't include justifications
## Overview

Block integrity is ensured by hashing of block data and signing the hash by the validator which produce a block.

The problem with [current (v0.12.6) implementation of block hashing](https://github.com/rchain/rchain/blob/v0.12.6/casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala#L286-L295) is that doesn't include justification field as part of hashed data.

```scala
def hashBlock(blockMessage: BlockMessage): BlockHash =
  ProtoUtil.hashByteArrays(
    blockMessage.header.toProto.toByteArray,
    blockMessage.body.toProto.toByteArray,
    blockMessage.sender.toByteArray,
    StringValue.of(blockMessage.sigAlgorithm).toByteArray,
    Int32Value.of(blockMessage.seqNum).toByteArray,
    StringValue.of(blockMessage.shardId).toByteArray,
    blockMessage.extraBytes.toByteArray
  )
```

It seems that the problem has roots even in the very first commit where input for block signature is calculated:
`hash(justification) + blockHash` :question:.
https://github.com/rchain/rchain/blob/b1e7c9a4f2d2da4b53b58d04452ea77d6d0e2439/casper/src/main/protobuf/coop/rchain/casper/protocol/CasperMessage.proto#L87-L93

Because of the sensitivity of the bug, this issue is created retroactively after fix PR is merged and a new release published on main net.
- https://github.com/rchain/rchain/pull/3693

## Consequence

This bug enables attackers to use original block proposed by the validator to make changes to justifications and send it to other validators effectively causing block creator to be slashed, probably choosing from a few slashing reasons like justification regression or invalid sequence number.

tgrospic:
## Overview

This PR fixes the major bug in block hashing function. To support compatibility to replay previous blocks, block version field is used mark _version 1_ with existing hashing and _version 2_ with the fix.

Details about the bug are specified in the linked issue.

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/Validate.scala
+++ b/casper/src/main/scala/coop/rchain/casper/Validate.scala
@@ -8,6 +8,7 @@ import com.google.protobuf.ByteString
 import coop.rchain.blockstorage.BlockStore
 import coop.rchain.blockstorage.dag.BlockDagRepresentation
 import coop.rchain.blockstorage.syntax._
+import coop.rchain.casper.genesis.Genesis
 import coop.rchain.casper.protocol.{ApprovedBlock, BlockMessage, Justification}
 import coop.rchain.casper.util.ProtoUtil.bonds
 import coop.rchain.casper.util.rholang.RuntimeManager
@@ -17,7 +18,7 @@ import coop.rchain.crypto.signatures.Secp256k1
 import coop.rchain.dag.DagOps
 import coop.rchain.metrics.{Metrics, Span}
 import coop.rchain.models.BlockHash.BlockHash
-import coop.rchain.models.BlockMetadata
+import coop.rchain.models.{BlockMetadata, BlockVersion}
 import coop.rchain.models.Validator.Validator
 import coop.rchain.shared.{Base16, _}
 
@@ -138,16 +139,17 @@ object Validate {
       true.pure
     }
 
-  def version[F[_]: Monad: Log](b: BlockMessage, version: Long): F[Boolean] = {
+  def version[F[_]: Monad: Log](b: BlockMessage): F[Boolean] = {
     val blockVersion = b.header.version
-    if (blockVersion == version) {
+    if (BlockVersion.Supported.contains(blockVersion)) {
       true.pure
     } else {
+      val versionsStr = BlockVersion.Supported.mkString(" or ")
       Log[F]
         .warn(
           ignore(
             b,
-            s"received block version $blockVersion is the expected version $version."
+            s"received block version $blockVersion is not the expected version $versionsStr."
           )
         )
         .as(false)
@@ -165,8 +167,6 @@ object Validate {
       expirationThreshold: Int
   ): F[ValidBlockProcessing] =
     (for {
-      _ <- EitherT.liftF(Span[F].mark("before-block-hash-validation"))
-      _ <- EitherT(Validate.blockHash(block))
       _ <- EitherT.liftF(Span[F].mark("before-timestamp-validation"))
       _ <- EitherT(Validate.timestamp(block))
       _ <- EitherT.liftF(Span[F].mark("before-repeat-deploy-validation"))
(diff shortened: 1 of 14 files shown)


casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -165,8 +167,6 @@ object Validate {
       expirationThreshold: Int
   ): F[ValidBlockProcessing] =
     (for {
-      _ <- EitherT.liftF(Span[F].mark("before-block-hash-validation"))
-      _ <- EitherT(Validate.blockHash(block))
```
tgrospic, line 169: Wrong block hash is now ignored.

integration-tests/test/test_slash.py
```diff
@@ -32,6 +33,34 @@
 BONDED_VALIDATOR_KEY_3 = PrivateKey.from_hex("af47862137d4e772f540029ae73ee01443c61288f3df9307a13d681de6ad2de4")
 
 
+def blake2b_32(data: bytes = b'') -> hashlib.blake2b:
+  return hashlib.blake2b(data, digest_size=32)
+
+def gen_block_hash_from_block(block: BlockMessage) -> bytes:
```
tgrospic, line 40: @zsluedem This function can be removed from PyRChain or kept in sync for every node version.

tgrospic:
bors try

tgrospic:
bors try

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
```diff
@@ -293,6 +305,19 @@ object ProtoUtil {
...
     )
+
+  /**
+    * Create hash of a BlockMessage, all fields must be included except signature
+    */
+  def hashBlockVersion2(blockMessage: BlockMessage): BlockHash = {
+    val emptyBytes = ByteString.EMPTY
+    val blockClearSigData = blockMessage.copy(
+      blockHash = emptyBytes,
+      sig = emptyBytes
```
nzpr, line 316: Sig algorithm is still inside signed body? Did we want to make it out? I'm ok if we do this later, but maybe now is a good time.

casper/src/test/scala/coop/rchain/casper/util/BlockHashSignatureSpec.scala
```diff
@@ -2,25 +2,74 @@ package coop.rchain.casper.util
...
+        val hashModified  = ProtoUtil.hashBlock(blockModified)
 
-      hashModified shouldBe hash
+        // Check hashes are different for modified block
+        hashModified !== hash ?= isModified
+      }
     }
   }
 
+  "block hash" should "not match for changed sequence number" in {
```
nzpr, line 29: Do we need more of these for all fields in block?

casper/src/main/scala/coop/rchain/casper/Validate.scala
nzpr: This should be done before saving to the block store, right?

nzpr (approved):
LGTM, minor comments

casper/src/main/scala/coop/rchain/casper/Validate.scala
tgrospic: Yes, block hash is checked in `checkIfWellFormedAndStore` together with the signature.

casper/src/main/scala/coop/rchain/casper/util/ProtoUtil.scala
tgrospic: We can do it as part of 0.13 version. I was trying to keep minimal changes here.

casper/src/test/scala/coop/rchain/casper/util/BlockHashSignatureSpec.scala
tgrospic: I'd like to have it. It's basically the same pattern for each field so good candidate for some macro to simplify.

PR #3695, opened 2022-05-12
Binding of BlockAPI deployStatus to gRPC, WebAPI and CLI

### LINKED ISSUE #3537: Implement new findDeploy API method

stanislavlyalin:
## Overview

Implemented binding of the BlockAPI `deployStatus` method to the gRPC and WebAPI interfaces and CLI.

### Using gRPC/CLI
```console
$ ./rnode deploy --phlo-limit 100000000 --phlo-price 1 --private-key <PRIVATE_KEY> --shard-id root <PATH_TO_FILE>/contract.rho
Response: Success!
DeployId is: 3044022053ca277d0991a6c4e3c51518e6b3387e1a984317c67e97adcd238f9fa29e1b86022006667a90b52d4c7f6f5a2dc50949179a24b803a08c4caae53b20cbecab41fb83

$ ./rnode deploy-status --deploy-signature 3044022053ca277d0991a6c4e3c51518e6b3387e1a984317c67e97adcd238f9fa29e1b86022006667a90b52d4c7f6f5a2dc50949179a24b803a08c4caae53b20cbecab41fb83
Deploy status: notProcessed {
  status: "Pooled"
}
```

### Using WebAPI

GET-request on http://localhost:40403/api/v1/deploy-status/3044022053ca277d0991a6c4e3c51518e6b3387e1a984317c67e97adcd238f9fa29e1b86022006667a90b52d4c7f6f5a2dc50949179a24b803a08c4caae53b20cbecab41fb83

```json
{"type":"NotProcessed","status":"Pooled"}
```


casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -174,11 +179,63 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+      for {
+        blockHash <- deployBlockHash
+        blockInfo <- getBlockInfo(blockHash)
+        seqPar    <- getSeqPar(blockHash)
+      } yield f(blockInfo, seqPar)
+
+    def notProcessed(message: String) = DeployExecStatus().withNotProcessed(NotProcessed(message))
+
+    for {
+      status      <- BlockDagStorage[F].asInstanceOf[BlockDagKeyValueStorage[F]].deployStatus(deployId)
```
tgrospic: This is not safe cast. Better to accept `BlockDagKeyValueStorage` or just deploy status function.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -174,11 +179,63 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
-      .asInstanceOf[BlockDagKeyValueStorage[F]]
-      .deployStatus(deployId)
-      .map(_.asRight[Error])
+  override def deployStatus(deployId: DeployId): F[ApiErr[DeployExecStatus]] = {
+
+    def deployBlockHash: F[String] =
+      findDeploy(deployId).map(_.getOrElse(LightBlockInfo.defaultInstance)).map(_.blockHash)
+
+    def getBlockInfo(blockHash: String): F[BlockInfo] =
+      getBlock(blockHash).map(_.getOrElse(BlockInfo.defaultInstance))
```
tgrospic: Default block here doesn't make sense. If user receive it in any case it cannot reason abut what happened. If block must  be found raise an error.
Also using BlockApi public methods with already wrapped errors can create another point of confusion. It's wrong that errors are converted to string in BlockApi in the first place. Better to use BlockStore.getUnsafe and BlockDagStorage[F].lookupByDeployId directly.

For these kind of errors user is not responsible so admin of the node should see it in the node log. Ensure that this is the case.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -174,11 +179,63 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+                       case (blockInfo, _) =>
+                         val message =
+                           s"""Deploy $deployIdStr was stored in deploy-index but not finalized."""
+                         val error = ProcessedWithError(message, blockInfo)
+                         DeployExecStatus().withProcessedWithError(error)
+                     }
+                   case "Pooled" =>
+                     notProcessed(s"""Deploy "$deployIdStr" stored but not processed.""").pure
+                   case "Expired"     => notProcessed(s"""Deploy "$deployIdStr" was expired.""").pure
+                   case "Unknown" | _ => notProcessed(s"""Deploy "$deployIdStr" not found.""").pure
```
tgrospic: Don't wrap enumerated string. This status will be used in client code to detect status so it must be simple constant not the sentence that needs to be parsed.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
stanislavlyalin: @tgrospic, the simplest solution is to move `deployStatus` to `BlockDagStorage` trait as I mentioned [before](https://github.com/rchain/rchain/pull/3682#discussion_r871296148).

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +178,55 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
       )
   }
 
+  override def deployStatus(deployId: DeployId): F[ApiErr[DeployExecStatus]] = {
+
+    def getSeqPar(blockHash: String): F[Seq[Par]] =
+      for {
+        par       <- Sync[F].delay(GDeployId(deployId).asInstanceOf[Par])
```
tgrospic: Avoid using `asInstanceOf` because it's not type safe.

In this case you need implicit conversions defined for Rholang types.
```scala
val par: Par = GDeployId(deployId)  // here Par is constructed with implicit conversions
```

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +178,55 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+      for {
+        blockHash <- BlockDagStorage[F].lookupByDeployId(deployId).map(_.get.toHexString)
+        blockInfo <- getBlock(blockHash) >>= liftToBlockApiErr
+        seqPar    <- getSeqPar(blockHash)
+      } yield f(blockInfo, seqPar)
+
+    def notProcessed(message: String) = DeployExecStatus().withNotProcessed(NotProcessed(message))
+
+    for {
+      status <- BlockDagStorage[F].deployStatus(deployId)
```
tgrospic: `deployStatus` that Anton created was just as an template how to get specific data and not to be used in its current form. Now access to storage is in `deployStatus`, converted to simple string and here storage is queried again.

Please take the logic from `deployStatus` here or prepare more structured data there and use storage access only once.

models/src/main/protobuf/DeployServiceCommon.proto
```diff
@@ -27,6 +27,10 @@ message FindDeployQuery {
   bytes deployId = 1;
 }
 
+message DeployStatusQuery {
+  bytes deployId = 1;
+}
```
tgrospic: We already have this type `FindDeployQuery`.

models/src/main/protobuf/DeployServiceV1.proto
```diff
@@ -34,6 +34,7 @@ option (scalapb.options) = {
 service DeployService {
   // Queue deployment of Rholang code (or fail to parse).
   rpc doDeploy(DeployDataProto) returns (DeployResponse) {}
+  rpc deployStatus(DeployStatusQuery) returns (DeployStatusResponse) {}
```
tgrospic: Please add short description in comments.

casper/src/main/scala/coop/rchain/casper/api/BlockApi.scala
```diff
@@ -82,6 +85,8 @@ object BlockApi {
   type Error     = String
   type ApiErr[A] = Either[Error, A]
 
+  final class BlockApiException(message: String) extends Exception(message)
+
```
stanislavlyalin: This definition already presented in WebApi.scala but it makes more sense to place it here.

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
```diff
@@ -617,6 +617,19 @@ final case class Options(arguments: Seq[String]) extends ScallopConf(arguments)
...
   addSubcommand(deploy)
 
+  val deployStatus = new Subcommand("deploy-status") {
+    descr(
+      "Returns the status of the deploy with provided id."
+    )
+    helpWidth(width)
+
+    val deployId = opt[Array[Byte]](
+      descr = "Id of the deploy.",
```
tgrospic: The current code is not consistent and somewhere is used deploy `signature` and somewhere deploy `Id`. Please use in description and response to the user as `signature`. It will make more sense to users.

casper/src/main/scala/coop/rchain/casper/api/BlockApi.scala
```diff
@@ -82,6 +85,8 @@ object BlockApi {
   type Error     = String
   type ApiErr[A] = Either[Error, A]
 
+  final class BlockApiException(message: String) extends Exception(message)
```
tgrospic: Is there a reason to define specialized exception type?

I'm not against it but before we create clear strategy how to classify errors it can just create confusion that it's defined for some handling reasons.

stanislavlyalin: This error class is declared in WebApi.scala, as I wrote in the comment [above](https://github.com/rchain/rchain/pull/3695#discussion_r880596172), but in my opinion it is more logical to move it here.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3695/commits/98cbd1fedfd81b0eb248f8b1ea4a74351ad15eb9).

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
stanislavlyalin: The disadvantage of the previous implementation was string parsing. I got away from this by [adding a trait and derived classes](https://github.com/rchain/rchain/pull/3695/commits/f2682ce89b8c9d22de7d3a599c0e2cd2c5c8b3b1). Also, an unnecessary call to `BlockDagStorage` is not performed, and the hash of the block corresponding to the deployment is passed to the field of derived classes. I'm sure the code could be written more beautifully, but I couldn't figure out how.

models/src/main/protobuf/DeployServiceCommon.proto
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3695/commits/9eb938cbcc9dfa4e3902dd4efac2d1ed0dd96d43).

models/src/main/protobuf/DeployServiceV1.proto
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3695/commits/d035a9ce61b501cbcea69ce579f1d24435d35c52).

node/src/main/scala/coop/rchain/node/configuration/commandline/Options.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3695/commits/613226c16065ce7a0928490bc387ad8af1ec1526).

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
tgrospic: Repeating my comment form PR you mentioned.

`deployStatus` which returns string is just temporary functions and should not be on block dag storage at all.

We need to see what is needed for BlockApi to return deploy status and add just that to block dag storage.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
stanislavlyalin: Link to commit in comment above is outdated after force push. In the [current version](https://github.com/rchain/rchain/pull/3695/commits/448443e668227ef6185cf33f08d35a8424a7bc8d) implementation moved to BlockApiImpl.

casper/src/test/scala/coop/rchain/casper/api/BlockQueryResponseAPITest.scala
```diff
@@ -69,14 +71,17 @@ class BlockQueryResponseAPITest
...
+
   // TODO: Test tsCheckpoint:
   // we should be able to stub in a tuplespace dump but there is currently no way to do that.
   "getBlock" should "return successful block info response" in withStorage {
     implicit blockStore => implicit blockDagStorage =>
       runtimeManagerResource.use { implicit runtimeManager =>
         for {
           _                  <- prepareDagStorage[Task]
-          blockApi           <- createBlockApi[Task]("", 1)
+          blockApi           <- createBlockApi[Task]("", 1, etState)
```
tgrospic: Why not create Ref inside `createBlockApi` as other default values?

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +181,60 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
       )
   }
 
+  override def deployStatus(deployId: DeployId): F[ApiErr[DeployExecStatus]] = {
+    def withSeqParAndBlockInfo(
+        blockHashOpt: Option[BlockHash]
```
tgrospic: Don't use Option in input arguments, except in cases where it carries information.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +181,60 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+    )(f: (Seq[Par], BlockInfo) => DeployExecStatus): F[DeployExecStatus] = {
+      import coop.rchain.models.rholang.implicits._
+      for {
+        blockHash <- blockHashOpt.map(_.toHexString).get.pure
```
tgrospic: Don't use `get` directly on Option.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +181,60 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+                          DeployExecStatus().withProcessedWithSuccess(success)
+                        case DeployStatusError(s) =>
+                          DeployExecStatus().withProcessedWithError(
+                            ProcessedWithError(s, blockInfo)
+                          )
+                      }
+                  }
+                } else notProcessed("In Progress")
+              } else notProcessed(unknown)
+            } else notProcessed(if (isDeployInPool) "Pooled" else unknown)
```
tgrospic: Don't create deeply nested if/match expressions which are prone to errors, hard to read and almost impossible to change later.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +181,60 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+
+    def notProcessed(status: String): F[DeployExecStatus] =
+      DeployExecStatus().withNotProcessed(NotProcessed(status)).pure
+
+    def unknown = "Unknown"
+
+    def statusF: F[DeployExecStatus] =
+      for {
+        blockHashOpt    <- BlockDagStorage[F].lookupByDeployId(deployId)
+        isDeployInBlock = blockHashOpt.isDefined
```
tgrospic: Don't use `isDefined` on Option because it's implementation detail, try to use map/flatMap/orElse.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +181,60 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+        isDeployInPool  <- BlockDagStorage[F].pooledDeploys.map(_.contains(deployId))
```
tgrospic: Why to read from another storage if deploy is found in the block?

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -92,7 +99,8 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
     devMode: Boolean,
     triggerProposeOpt: Option[ProposeFunction[F]],
     proposerStateRefOpt: Option[Ref[F, ProposerState[F]]],
-    autoPropose: Boolean
+    autoPropose: Boolean,
+    etState: Ref[F, Map[DeployId, Option[DeployStatus]]]
```
tgrospic: BlockApi should not be manage internal state of components it uses. It should accept execution tracker.

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
```diff
@@ -0,0 +1,30 @@
...
+  override def execComplete(d: DeployId, res: EvaluateResult): F[Unit] = {
+    val err    = res.errors.map(_.getMessage).mkString("\n")
+    val status = if (err.isBlank) DeployStatusSuccess else DeployStatusError(err)
+    state.update(_ + (d -> status.some))
+  }
+}
+
+object ExecutionTracker {
+  def apply[F[_]](state: Ref[F, Map[DeployId, Option[DeployStatus]]]) =
+    new ExecutionTracker[F](state)
```
tgrospic: This additional constructor can e.g. create default state. In this form it's just the identical as class constructor so no real point of it.

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
```diff
@@ -0,0 +1,30 @@
...
+trait BlockExecutionTracker[F[_]] {
+  def execStarted(d: DeployId): F[Unit]
+  def execComplete(d: DeployId, res: EvaluateResult): F[Unit]
+}
+
+sealed trait DeployStatus
+case object DeployStatusSuccess                    extends DeployStatus
+final case class DeployStatusError(status: String) extends DeployStatus
+
+final class ExecutionTracker[F[_]](state: Ref[F, Map[DeployId, Option[DeployStatus]]])
```
tgrospic: The name of implementation should not be subset of the name of the interface. It has more functionality the the interface so it must have some clue about that in the name.

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
```diff
@@ -0,0 +1,30 @@
...
+    extends BlockExecutionTracker[F] {
+  override def execStarted(d: DeployId): F[Unit] = state.update(_ + (d -> none[DeployStatus]))
+  override def execComplete(d: DeployId, res: EvaluateResult): F[Unit] = {
+    val err    = res.errors.map(_.getMessage).mkString("\n")
+    val status = if (err.isBlank) DeployStatusSuccess else DeployStatusError(err)
```
tgrospic: What will happen here is `errors = Seq(InterpreterError(""))`? EvaluateResult has a field to check if it has errors.

casper/src/main/scala/coop/rchain/casper/api/BlockApi.scala
tgrospic: This doesn't explain why this definition is needed. At least it's not used in BlockApi and it's defined in this file.

Please add WebApi changes to this PR also.

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
stanislavlyalin: Good catch!

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
stanislavlyalin: If the deployment is not found in the DAG (not included in the block), it may be in processing (pooled). This flag is needed to inform the user of the appropriate status.

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +183,81 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
       )
   }
 
+  override def deployStatus(deployId: DeployId): F[ApiErr[DeployExecStatus]] = {
```
stanislavlyalin, line 186: Very elegant implementation! It is easy to read and understand but difficult to write for me :slightly_smiling_face:
It is interesting to use `collect` and `as` functions. I always use `map` but your solution with this functions more beautiful :+1:

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +183,81 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+        // Deploy found in the pool, waiting to be executed and added to a block
+        notProcessed("Pooled")
+      }
+    }
+
+    def findCurrentlyExecutedDeploy: OptionT[F, DeployExecStatus] = {
+      val deployStatusOptT = OptionT(executionTracker.findDeploy(deployId))
+      deployStatusOptT.map { _ =>
+        // Status found, block creation in progress, deploy execution started
+        notProcessed("Running")
```
stanislavlyalin: Maybe we can use `as` function here as above in `findPooledDeploy`?
```scala
deployStatusOptT.as {
  // Status found, block creation in progress, deploy execution started
  notProcessed("Running")
}
```

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
```diff
@@ -0,0 +1,39 @@
...
+
+sealed trait DeployStatus
+case object DeployStatusStarted                    extends DeployStatus
+final case class DeployStatusError(status: String) extends DeployStatus
+
+object StatefulExecutionTracker {
+  def apply[F[_]: Sync]: F[StatefulExecutionTracker[F]] =
+    for {
+      ref <- Ref.of(Map.empty[DeployId, DeployStatus])
+    } yield new StatefulExecutionTracker(ref)
```
stanislavlyalin, line 23: Why it is needed? The only reason in my opinion is not to write `new` when instantiating.

stanislavlyalin (commented):
I didn't add enum/constants for string statuses as they are not duplicated. @tgrospic, is it necessary?
I believe this PR can be used as a guide to using FP in Scala :slightly_smiling_face:

casper/src/main/scala/coop/rchain/casper/BlockExecutionTracker.scala
tgrospic: There are two reasons:
1. make creation lazy so any computation in constructor will not be executed on call but when F value is executed
2. in this case creation of Ref requires F context

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
tgrospic: Yes, you are right. Good catch!
It can be on liner.
`deployStatusOptT.as(notProcessed("Running"))`

casper/src/test/scala/coop/rchain/casper/api/BlocksResponseAPITest.scala
```diff
@@ -1,9 +1,12 @@
 package coop.rchain.casper.api
 
 import cats.effect.Resource
+import cats.effect.concurrent.Ref
 import coop.rchain.blockstorage.BlockStore.BlockStore
 import coop.rchain.blockstorage.dag.BlockDagStorage
+import coop.rchain.blockstorage.dag.BlockDagStorage.DeployId
 import coop.rchain.blockstorage.syntax._
+import coop.rchain.casper.DeployStatus
```
tgrospic: Is this leftover? The same for BlockQueryResponseAPITest.scala?

tgrospic (approved):
LGTM

zsluedem (commented):
What about `Rejected` status , `Expired` status and `Finalized`status? Is it future plan?

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -175,6 +183,77 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+    def findPooledDeploy: OptionT[F, DeployExecStatus] = {
+      val deployPooled = BlockDagStorage[F].containsDeployInPool(deployId).map(_.guard[Option])
+      // Deploy found in the pool, waiting to be executed and added to a block
+      OptionT(deployPooled).as(notProcessed("Pooled"))
+    }
+
+    def findCurrentlyExecutedDeploy: OptionT[F, DeployExecStatus] = {
+      val deployStatusOptT = OptionT(executionTracker.findDeploy(deployId))
+      // Status found, block creation in progress, deploy execution started
+      deployStatusOptT.as(notProcessed("Running"))
```
zsluedem, line 236: ~~The `Running`  status is useless from the user perspective.I suggest remove it.~~

stanislavlyalin:
> What about Rejected status , Expired status and Finalizedstatus? Is it future plan?

@tgrospic, wdy think?

tgrospic:
> What about `Rejected` status , `Expired` status and `Finalized`status? Is it future plan?

@zsluedem Yes, we can add this later.

zsluedem (approved):
LGTM . Please keep https://github.com/rchain/rchain/issues/3537 open until we got `finalized`, `rejected` status.

stanislavlyalin:
bors merge

PR #3696, opened 2022-05-13
Added class KeySegment, replace KeyPath to KeySegment in code

### LINKED ISSUE #3670: Refactoring of History and RadixTree data types
## Overview
For better readability and code organization, we need to use common history types in these data structures:
https://github.com/rchain/rchain/blob/04bc32552441531fa49dc7e78e03039b1550e9ce/rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala#L25-L33

### Impact
Many files are in Rspace because `KeyPath` is used in many places.

1. Fully remove old implementation of History (also tests, and related files).
https://github.com/rchain/rchain/blob/04bc32552441531fa49dc7e78e03039b1550e9ce/rspace/src/main/scala/coop/rchain/rspace/history/HistoryMergingInstances.scala#L47-L51
1. Change `value: ByteVector` and `ptr: ByteVector` to `value: Blake2b256Hash` and `ptr: Blake2b256Hash`. Adaptation RadixTree code for new data types.
1. Create trait `KeyPath`. This trait should be founded on ByteVector. Need to create basic operations: append, equal, common, ... etc. Change `prefix: ByteVector` to `prefix: KeyPath` in RadixTree, adapt RadixTree code.
1. Use a new type in history instead of this data type:
https://github.com/rchain/rchain/blob/04bc32552441531fa49dc7e78e03039b1550e9ce/rspace/src/main/scala/coop/rchain/rspace/history/History.scala#L80
Adopting history implementation and these data types for the new KeyPath:
https://github.com/rchain/rchain/blob/04bc32552441531fa49dc7e78e03039b1550e9ce/rspace/src/main/scala/coop/rchain/rspace/history/HistoryAction.scala#L10-L11

After each step needed to validate, that time performance doesn't slower than the old implementation. With the help [tests](https://github.com/rchain/rchain/blob/04bc32552441531fa49dc7e78e03039b1550e9ce/node/src/test/scala/coop/rchain/node/perf/HistoryGenKeySpec.scala). Need to increase the size of data to 100000 records in tests temporally.

maximus19910112:
## Overview
Third step of Refactoring of History and RadixTree according to the [issue](https://github.com/rchain/rchain/issues/3670)

### Notes
1. Added trait KeySegmentBase and child class KeySegment (that should replace KeyPath in History.scala)
2. Remove type KeyPath, adaptation of code to KeySegment implementation
3. Removed MergingHistory artefacts
4. Made performance measurement before and after the recfactoring. The measurement results are shown in the table below. 

|                    	|  timeI(sec) 	|  timeR(sec) 	|  timeD(sec) 	|  timeExp(sec) 	|  timeValid(sec) 	|     numNode 	|
|--------------------	|-------------	|-------------	|-------------	|---------------	|-----------------	|-------------	|
| Before refactoring 	| 0.723       	| 0.555       	| 0.283       	| 0.478         	| 0.356           	| 29990       	|
| After refactoring  	| 0.626       	| 0.584       	| 0.264       	| 0.488         	| 0.369           	| 29990       	|

Note: numIRD = 100000, typeStore = Lmdb, typeHistory = Radix


rspace/src/main/scala/coop/rchain/rspace/history/KeySegmentBase.scala
```diff
@@ -0,0 +1,23 @@
+package coop.rchain.rspace.history
+import scodec.bits.ByteVector
+import scala.annotation.tailrec
+
+trait KeySegmentBase[A] {
+  val value: A
+  def size: Long
+  def isEmpty: Boolean
+  def nonEmpty: Boolean
+  def head: Byte
```
nzpr: What are other possible implementations for KeySegmentBase then ByteVector?
Here you are making a type parameter A, but assume that head is a `Byte`, and tail is `KeySegmentBase[A]`. Does not this mean that all implementations of `KeySegmentBase[A]` are sequences of bytes?

rspace/src/main/scala/coop/rchain/rspace/history/HistoryAction.scala
```diff
@@ -1,11 +1,10 @@
...
-import coop.rchain.rspace.history.History.KeyPath
 
 sealed trait HistoryAction {
-  def key: KeyPath
+  def key: KeySegment
 }
 
-final case class InsertAction(key: KeyPath, hash: Blake2b256Hash) extends HistoryAction
-final case class DeleteAction(key: KeyPath)                       extends HistoryAction
+final case class InsertAction(key: KeySegment, hash: Blake2b256Hash) extends HistoryAction
```
nzpr, line 9: It looks like base trait is not used anywhere except when defining `KeySegment` implementation. Why to define trait then at all?

stanislavlyalin (commented):
It looks like unnecessary complications. The implementation of KeySegment class almost fully repeates ByteVector methods. I think the idea is to make code more type explicit so please consider to use `type KeyPath = ByteVector` with some useful methods.
Also, please don't forget to update PR description after code updating.

node/src/test/scala/coop/rchain/node/perf/HistoryGenKeySpec.scala
```diff
@@ -281,13 +282,15 @@ class HistoryGenKeySpec extends FlatSpec with Matchers with BeforeAndAfterAll {
...
+                initInsertActions = initHashes.map(x => InsertAction(KeySegment(x.bytes), x))
 
                 insReadDelHashes = genTasks.toList
-                insertActions    = insReadDelHashes.map(x => InsertAction(x.bytes.toArray.toList, x))
-                deleteActions    = insReadDelHashes.map(x => DeleteAction(x.bytes.toArray.toList))
-                historyInitW     <- getHistory(v0)
-                historyInit      <- getHistory(v0)
+                insertActions = insReadDelHashes.map(
+                  x => InsertAction(KeySegment(x.bytes), x)
+                )
```
stanislavlyalin: Looks like unnecessary code line splitting :smiley: I use a rule: minimize changes and save only those that are relevant to the task.

rspace/src/main/scala/coop/rchain/rspace/history/History.scala
```diff
@@ -11,7 +11,7 @@ import scodec.bits.ByteVector
 /**
   * History definition represents key-value API for RSpace tuple space
   *
-  * [[History]] contains only references to data stored on keys ([[KeyPath]]).
+  * [[History]] contains only references to data stored on keys ([[KeySegment]]).
```
stanislavlyalin, line 13: What is the need to change KeyPath to KeySegment? It leads to many changes in code but not make sense for me :thinking:

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
```diff
@@ -72,6 +72,8 @@ final case class HistoryRepositoryImpl[F[_]: Concurrent: Parallel: Log: Span, C,
 
   private def calculateStorageActions(action: HotStoreTrieAction): Result = {
     import HistoryRepositoryInstances.{PREFIX_DATUM, PREFIX_JOINS, PREFIX_KONT}
+    implicit def fromSeqToKeySegment(prefixSeq: Seq[Byte]): KeySegment =
+      KeySegment(prefixSeq.toArray)
```
stanislavlyalin: Hm :thinking: Why not to make this constructing **explicit** like other constructors in KeySegment companion object? Or use this approach with **implicit** function in other similar places?

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,54 @@
+package coop.rchain.rspace.history
+
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+final case class KeySegment(val value: ByteVector) {
```
stanislavlyalin: @nzpr mentioned that it can be implemented as [value class](https://docs.scala-lang.org/overviews/core/value-classes.html) to eliminate unnecessary wrappers and increase performance. Maybe it make sense?

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,54 @@
...
+        val lHead = l.head
+        val rHead = r.head
+        if (lHead == rHead) go(common :+ lHead, l.tail, r.tail)
+        else (common, l, r)
+      }
+
+    go(KeySegment.empty, this, b)
+  }
+
+  require(value.size <= 127)
```
stanislavlyalin: Usually this definition places on the top of the class.

rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala
```diff
@@ -201,8 +201,9 @@ object RadixTree {
 
           // Decoding type of non-empty item
           val item =
-            if (isLeaf(secondByte)) Leaf(ByteVector(prefix), Blake2b256Hash.fromByteArray(valOrPtr))
-            else NodePtr(ByteVector(prefix), Blake2b256Hash.fromByteArray(valOrPtr))
+            if (isLeaf(secondByte))
+              Leaf(KeySegment(prefix), Blake2b256Hash.fromByteArray(valOrPtr))
```
stanislavlyalin: Looks like unnecessary code line splitting. It is not so important but it is more difficult to detect actual change on code review.

rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala
```diff
@@ -768,22 +755,26 @@ object RadixTree {
...
           (childItemIdx, childInsPrefix) = (byteToInt(insPrefix.head), insPrefix.tail)
           childItemOpt                   <- update(childNode(childItemIdx), childInsPrefix, insValue) // Deeper
           returnedItem = childItemOpt.map { childItem =>
             val updatedChildNode = childNode.updated(childItemIdx, childItem)
-            saveNodeAndCreateItem(updatedChildNode, childPrefix, compaction = false)
+            saveNodeAndCreateItem(
+              updatedChildNode,
+              childPrefix,
+              compaction = false
+            )
```
stanislavlyalin: The same as above (unnecessary code line splitting)

rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala
```diff
@@ -804,29 +795,23 @@ object RadixTree {
             else {
               // Create child node, insert existing and new leaf in this node.
               // Intentionally not recursive for speed up.
-              val (commPrefix, insPrefixRest, leafPrefixRest) = commonPrefix(insPrefix, leafPrefix)
+              val (commPrefix, insPrefixRest, leafPrefixRest) = insPrefix.commonPrefix(leafPrefix)
```
stanislavlyalin: If we make `commonPrefix` a class method, the name `commonPrefixWith` will look more natural (in my subjective opinion). Or even `commonWith` if all related variables has `prefix` in the name.
```scala
insPrefix.commonPrefixWith(leafPrefix)
// or
insPrefix.commonWith(leafPrefix)
```

rspace/src/main/scala/coop/rchain/rspace/history/History.scala
hilltracer: Path means a whole way to data, but we also are storing part of this way.

rspace/src/main/scala/coop/rchain/rspace/history/RadixTree.scala
hilltracer: @maximus19910112 just select lines in Sublime Merge and push "revert lines" =)

rspace/src/main/scala/coop/rchain/rspace/history/History.scala
```diff
@@ -11,7 +11,7 @@ import scodec.bits.ByteVector
 /**
```
hilltracer, line 10: Please remove unused imports: for example `import scodec.bits.ByteVector`. Use `optimize imports` in Idea for changed files.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
hilltracer: I agree, it will be cool to validate this thing.

hilltracer: image

rspace/src/main/scala/coop/rchain/rspace/history/HistoryMergingInstances.scala
```diff
@@ -96,11 +96,11 @@ final case class NodePointer(hash: Blake2b256Hash) extends ValuePointer
...
-  def append(affix: KeyPath, t: Trie): TriePath =
+final case class TriePath(nodes: Vector[Trie], conflicting: Option[Trie], edges: KeySegment) {
+  def append(affix: KeySegment, t: Trie): TriePath =
     this.copy(nodes = this.nodes :+ t, edges = this.edges ++ affix)
 }
 
 object TriePath {
-  def empty: TriePath = TriePath(Vector(), None, Nil)
+  def empty: TriePath = TriePath(Vector(), None, KeySegment.empty)
 }
```
hilltracer: This file is also a MergingHistory artifact, and it will be good to remove it. Also, it will be good to remove dependence classes.

rspace/src/main/scala/coop/rchain/rspace/history/HistoryRepositoryImpl.scala
hilltracer: ` InsertAction(PREFIX_DATUM +: i.hash.bytes.toSeq.toList, dataHash)`
`InsertAction(PREFIX_JOINS +: i.hash.bytes.toSeq.toList, joinsHash)`, etc.
It is that what I asked @maximus19910112 to change - it is wrong to convert Bytevector(`i.hash.bytes`) to List, after that convert again to ByteVector in apply. Please use direct ByteVector creation: `KeySegment(PREFIX_DATUM +: i.hash.bytes)`

rspace/src/test/scala/coop/rchain/rspace/history/HistoryActionTests.scala
```diff
@@ -317,13 +316,16 @@ class HistoryActionTests extends FlatSpec with Matchers with InMemoryHistoryTest
 
 object TestData {
 
-  implicit def toByteVector(bytes: KeyPath): ByteVector = ByteVector(bytes)
+  // implicit def toByteVector(bytes: KeyPath): ByteVector = ByteVector(bytes)
```
hilltracer: Please remove commented code.

rspace/src/test/scala/coop/rchain/rspace/history/HistoryActionTests.scala
```diff
@@ -317,13 +316,16 @@ class HistoryActionTests extends FlatSpec with Matchers with InMemoryHistoryTest
...
+  implicit def toKeySegment(seq: Seq[Byte]): KeySegment =
+    KeySegment(seq.toArray)
```
hilltracer: Please remove this implicit and change _zeros vals. Also please change `def hexKey(s: String): Seq[Byte] ` to `def hexKey(s: String):  KeySegment`. It will be good to add `def apply(ab: Seq[Byte]): KeySegment` in KeySegment class

hilltracer (changes requested):
After changing please unite commites in one. Good luck!

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,45 @@
+package coop.rchain.rspace.history
+
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+final case class KeySegment(value: ByteVector) extends AnyVal {
```
stanislavlyalin, line 7: Why is the value size check missing? Before that it was?

rspace/src/test/scala/coop/rchain/rspace/history/HistoryActionTests.scala
```diff
@@ -317,26 +311,25 @@ class HistoryActionTests extends FlatSpec with Matchers with InMemoryHistoryTest
...
-  val _31zeros: KeyPath         = List.fill(31)(0).map(_.toByte)
-  def zerosAnd(i: Int): KeyPath = _31zeros :+ i.toByte
-  def prefixWithZeros(s: String): KeyPath = {
+  def zerosAnd(i: Int): KeySegment = _31zeros :+ i.toByte
+  def prefixWithZeros(s: String): KeySegment = {
     val a = List.fill(32 - s.length)(0).map(_.toByte)
     val b = s.toCharArray.toList.map(_.asDigit).map(_.toByte)
-    a ++ b
+    KeySegment(a ++ b)
   }
```
hilltracer: Better to use direct bytes filling instead of `map(._toByte)` conversion: `List.fill(31)(0.toByte)`

hilltracer (changes requested):
Please revert value size check in KeySegment class, and create simple unit test for this in HistoryActionTests.scala after that we can approve PR.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
maximus19910112: I returned size check

stanislavlyalin: Why you reverted `extends AnyVal`? :thinking:

maximus19910112: I did it because I didn't find any benefit from the using value class instead of usual case class
Screenshot from 2022-05-30 13-29-27

rspace/src/test/scala/coop/rchain/rspace/history/RadixTreeSpec.scala
```diff
@@ -157,8 +157,12 @@ class RadixTreeSpec extends FlatSpec with Matchers with OptionValues with InMemo
       val referenceErrorMessage =
         s"assertion failed: The length of all prefixes in the subtree must be the same."
       for {
-        leafItemOpt <- impl.update(RadixTree.EmptyItem, initialKVPair.rKey, initialKVPair.rValue)
-        err         <- impl.update(leafItemOpt.get, wrongKVPair.rKey, wrongKVPair.rValue).attempt
+        leafItemOpt <- impl.update(
+                        RadixTree.EmptyItem,
+                        initialKVPair.rKey,
+                        initialKVPair.rValue
+                      )
```
stanislavlyalin: Unnecessary line splitting?

hilltracer (approved):

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,55 @@
+package coop.rchain.rspace.history
+
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+final case class KeySegment(value: ByteVector) {
+  require(value.size <= 127)
```
nzpr: Please always specify for such checks (assert and require) the second parameter - message to show to a user in case requirement is not met. Otherwise if require fails, it fails silently, it makes debugging hard.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  def size: Long               = value.size
+  def nonEmpty: Boolean        = value.nonEmpty
```
nzpr, line 9: Make this !isEmpty, no need to access value again. Main methods which define the trait should always be as minimal as possible, and everything else is a syntax on top if the trait. Here it's ok because it's a shortcut solution, so this is just a reminder that you should not define unnecessary methods on classes/traits.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  override def equals(obj: Any): Boolean =
+    obj match {
+      case c: KeySegment => c.value == value
+      case _             => false
+    }
+
+  override def hashCode(): Int = value.hashCode
+
+  def commonPrefixWith(b: KeySegment): (KeySegment, KeySegment, KeySegment) = {
```
nzpr: It is better to define functions that implements the logic in an companion object. So please define function `def commonPrefixWith(a:KeySegment, b: KeySegment): ` on an object and use here this function `commonPrefixWith(this, b)`. Also try to minimize number of lines of the code - here arguments of `go` function can all be on on line. Please check other places as well.

nzpr: Also this should be called just `commonPrefix`

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,55 @@
+package coop.rchain.rspace.history
+
+import scodec.bits.ByteVector
+
+import scala.annotation.tailrec
+
+final case class KeySegment(value: ByteVector) {
```
nzpr: Better to make this a value class 
```final case class KeySegment(val value: ByteVector) extends AnyVal ```
You don't need to define equals and hashCode in that case, and prevent boxing.

**Review by nzpr — APPROVED** [review]
LGTM

File: rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala [hunk 22]
```diff
@@ -0,0 +1,44 @@
...
+object KeySegment {
+  def apply(bv: ByteVector): KeySegment = {
+    require(bv.size <= 127, "Size of key segment is more than 127")
+    new KeySegment(bv)
+  }
+  def apply(ab: Array[Byte]): KeySegment = KeySegment(ByteVector(ab))
+  def apply(sb: Seq[Byte]): KeySegment   = KeySegment(ByteVector(sb))
+  val empty: KeySegment                  = KeySegment(ByteVector.empty)
+
+  def commonPrefix(a: KeySegment, b: KeySegment): (KeySegment, KeySegment, KeySegment) = {
```
tgrospic, line 31: Why did we lost documentation for this function?

PR #3701, opened 2022-05-17
Trait for key segment

nzpr:

### DIFF
--- a/rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
+++ b/rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
@@ -1,0 +1,40 @@
+package coop.rchain.rspace.history
+import coop.rchain.rspace.syntax._
+
+import scala.annotation.tailrec
+
+/**
+  * Segment of a History key
+  * @tparam V type of a key value
+  * @tparam W type of a word in key alphabet
+  */
+trait KeySegment[K, W] {
+  def empty: K
+  def size(ks: K): Int
+  def isEmpty(ks: K): Boolean
+  def headOption(ks: K): Option[W]
+  def tailOption(ks: K): Option[K]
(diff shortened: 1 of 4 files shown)


rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+  * Segment of a History key
+  * @tparam V type of a key value
+  * @tparam I type of a word in key alphabet
+  */
+trait KeySegment[V, I] {
+  def size: Long
+  def isEmpty: Boolean
+  def nonEmpty: Boolean
```
tgrospic: One of empty/nonEmpty should be on a trait, other can be derived by negation.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,47 @@
...
+  def head: I
```
tgrospic: What's behavior of _head_? Should it throw exception? The same is for _tail_.

rspace/src/main/scala/coop/rchain/rspace/history/KeySegment.scala
```diff
@@ -0,0 +1,47 @@
...
+  def headOption: Option[I]
+  def toHex: String
+  def :+(item: I): V
+  def tail: V
+  def ++(other: V): V
```
tgrospic: What this trait represents? A type-class or the data itself.

If it's data itself tail should return `KeySegment` as the rest without head.

If it's a type-class each operation should accept data similar to static functions.

E.g. `toHex` looks like transforming KeySegment to a string so it's like KeySegment as data, but `++` accepts V so it's not combining two segments KeySegments.

PR #3716, opened 2022-05-25
Fix getDataAtPar API method

### LINKED ISSUE #3715: Fix dataAtPar method without checking block deploys
## Overview

Block API method to get data at name works incorrectly. It uses checking of produces/consumes from old listen for data at name which must not be done on the new data at name method.

With this check dataAtName method only returns result if blockHash is specified for exact block where deploy is included in the block.

## Solution

Use code without this change to get data:
`data <- RuntimeManager[F].getData(stateHash)(sortedPar)`
https://github.com/rchain/rchain/pull/3584/files#diff-c0f46ab075076a0def47a8f71ac02093384d87074f65fada8a895fd9262ba116L814

nzpr:
## Overview
Closes https://github.com/rchain/rchain/issues/3715

### DIFF
--- a/casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
+++ b/casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
@@ -598,12 +598,24 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
 
   override def getDataAtPar(
       par: Par,
-      blockHash: String,
+      blockHashStr: String,
       usePreStateHash: Boolean
-  ): F[ApiErr[(Seq[Par], LightBlockInfo)]] =
-    for {
-      block     <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
-      sortedPar <- parSortable.sortMatch[F](par).map(_.term)
-      data      <- getDataWithBlockInfo(sortedPar, block).map(_.get)
-    } yield (data.postBlockData, data.block).asRight[Error]
+  ): F[ApiErr[(Seq[Par], LightBlockInfo)]] = {
+    val blockHashOpt = blockHashStr.hexToByteString
+    val dataF = for {
+      hashDecoded <- blockHashOpt.liftTo(new Exception(s"$blockHashStr is not a valid block hash."))
+      blockOpt    <- BlockStore[F].get1(hashDecoded)
+      block       <- blockOpt.liftTo(new Exception(s"Block $blockHashStr not found."))
+      sortedPar   <- parSortable.sortMatch[F](par).map(_.term)
+      stateHash = if (usePreStateHash) block.body.state.preStateHash
+      else block.body.state.postStateHash
+      data <- RuntimeManager[F].getData(stateHash)(sortedPar)
+      lbi  = getLightBlockInfo(block)
+    } yield (data, lbi)
+
+    dataF
+      .map(_.asRight[Error])
+      .handleErrorWith(_.getMessage.asLeft[(Seq[Par], LightBlockInfo)].pure)
+  }
+
 }


tgrospic (approved):
Awesome!

stanislavlyalin (approved):

nzpr:
bors merge

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -600,10 +600,16 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
+    val loadData = (for {
+      block     <- OptionT(BlockStore[F].get1(blockHash.unsafeHexToByteString))
+      sortedPar <- OptionT.liftF(parSortable.sortMatch[F](par).map(_.term))
+      stateHash = if (usePreStateHash) block.body.state.preStateHash
+      else block.body.state.postStateHash
+      data <- OptionT.liftF(RuntimeManager[F].getData(stateHash)(sortedPar))
+      lbi  = getLightBlockInfo(block)
+    } yield (data, lbi))
+    if (blockHash.isEmpty) "Invalid block hash".asLeft[(Seq[Par], LightBlockInfo)].pure[F]
+    else loadData.value.map(_.toRight[Error](s"No block with $blockHash."))
```
tgrospic: You can just raise exceptions for each wrong data and catch them at the end to convert to Left. Now that conversion to Either and string errors is done here (on the wrong place too early). 
```scala
val data = for {
  a <- aOpt.liftTo(new Exception(...))
  b <- bOpt.liftTo(new Exception(...))
  c <- cOpt.liftTo(new Exception(...))
} yield ...
data.map(_.toRight).handleErrorWith { case ex => ex.toLeft }
```

models/src/main/scala/coop/rchain/models/BlockHash.scala
```diff
@@ -5,5 +5,6 @@ import com.google.protobuf.ByteString
 object BlockHash {
   type BlockHash = ByteString
 
-  val Length = 32
+  val Length    = 32 // length in bytes
+  val HexLength = 64 // each byte is 2 chars
```
tgrospic: Your comment is correct but code duplicate the information about length.
`val HexLength = Length * 2`

casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala
```diff
@@ -600,10 +601,22 @@ class BlockApiImpl[F[_]: Concurrent: RuntimeManager: BlockDagStorage: BlockStore
...
       blockHash: String,
       usePreStateHash: Boolean
-  ): F[ApiErr[(Seq[Par], LightBlockInfo)]] =
-    for {
-      block     <- BlockStore[F].getUnsafe(blockHash.unsafeHexToByteString)
+  ): F[ApiErr[(Seq[Par], LightBlockInfo)]] = {
+    val hashIsInValid = blockHash.length != BlockHash.HexLength
+    val dataF = for {
+      _         <- new Exception(s"$blockHash is not a valid block hash.").raiseError.whenA(hashIsInValid)
+      blockOpt  <- BlockStore[F].get1(blockHash.unsafeHexToByteString)
```
tgrospic: Conversion from hex can fail e.g. `"0011mm".unsafeHexToByteString`.

Also what with hex `0x112233`?

tgrospic: Bigger question for me is how we ended up using string for binary in protobuf in the first place.

tgrospic: Take a look how it's done here.
https://github.com/rchain/rchain/blob/fa6a47b248b6fae131d9071171c4e4b1c195b22e/casper/src/main/scala/coop/rchain/casper/api/BlockApiImpl.scala#L475-L490

nzpr: `Bigger question for me is how we ended up using string for binary in protobuf in the first place.` I'm not sure wdy mean. Are you talking about hash being a string here and not bytes?
https://github.com/rchain/rchain/blob/cdae0514e4b0f58b756bd57ee07d82e85f1e91e8/models/src/main/protobuf/DeployServiceCommon.proto#L53-L57

nzpr:
bors merge

PR #3718, opened 2022-05-30
Changed Web API data-at-name-by-block-hash to accept Par object

### LINKED ISSUE #3712: Change Web API data-at-name-by-block-hash to accept Par object
## Overview

gRPC method to get data at name (`getDataAtName`) accepts `Par` type but Web API only accepts unforgeable name.

## Solution

Change `name` field in `DataAtNameByBlockHashRequest` to be `RhoExpr` which is representation of `Par` in Web API.

https://github.com/rchain/rchain/blob/ace32ac2aa31d82b7bb2c8e76483428d5a39a263/node/src/main/scala/coop/rchain/node/api/WebApi.scala#L190-L194

stanislavlyalin:
## Overview

Changed the `name` field type in the `DataAtNameByBlockHashRequest` class from `RhoUnforg` to `RhoExpr` to match the request with `getDataAtPar` in gRPC.

### Note

We don't have integration test for request `data-at-name-by-block-hash`. Lets consider to add it.

### DIFF
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -354,7 +354,7 @@ object WebApi {
       // Binary data is encoded as base16 string
       case ExprBytes(data) =>
         exprToPar(Expr().withGByteArray(data.unsafeHexToByteString))
-      case ExprUnforg(data) => unforgToPar(data)
+      case ExprUnforg(data) => unforgToParProto(data)
     }
   }
 
@@ -364,12 +364,12 @@ object WebApi {
     case UnforgDeployer(name) => GDeployerIdBody(GDeployerId(name.unsafeHexToByteString))
   }
 
-  private def unforgToPar(unforg: RhoUnforg): Par =
+  private def unforgToParProto(unforg: RhoUnforg): Par =
     Par(unforgeables = Seq(GUnforgeable(unforgToUnforgProto(unforg))))
 
   // Data request/response protobuf wrappers
 
-  private def toPar(req: DataAtNameRequest): Par = unforgToPar(req.name)
+  private def toPar(req: DataAtNameRequest): Par = unforgToParProto(req.name)
 
   private def toPar(req: DataAtNameByBlockHashRequest): Par = rhoExprToParProto(req.name)
 
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -339,9 +339,9 @@ object WebApi {
     // Nested expressions (Par, Tuple, List and Set are converted to JSON list)
     case data: ExprPar   => exprParToParProto(data)
     case ExprTuple(data) => Expression(Expr().withETupleBody(ETuple(data.map(rhoExprToParProto))))
-    case ExprList(data)  => RhoList(data.map(rhoExprToParProto))
-    case ExprSet(data)   => RhoSet(data.map(rhoExprToParProto))
-    case ExprMap(data)   => RhoMap(data.map { case (k, v) => (String(k), rhoExprToParProto(v)) })
+    case ExprList(data)  => List(data.map(rhoExprToParProto))
+    case ExprSet(data)   => Set(data.map(rhoExprToParProto))
+    case ExprMap(data)   => Map(data.map { case (k, v) => (String(k), rhoExprToParProto(v)) })
     // Terminal expressions (here is the data)
     case ExprBool(data)   => Boolean(data)
     case ExprInt(data)    => Number(data)
@@ -383,12 +383,13 @@ object WebApi {
 
   private def toDataAtNameResponse(req: (Seq[DataWithBlockInfo], Int)): DataAtNameResponse = {
     val (dbs, length) = req
-    val exprsWithBlock = dbs.foldLeft(List[RhoExprWithBlock]()) { (acc, data) =>
-      val exprs = data.postBlockData.flatMap(exprFromParProto)
-      // Implements semantic of Par with Unit: P | Nil ==> P
-      val expr  = if (exprs.size == 1) exprs.head else ExprPar(exprs.toList)
-      val block = data.block
-      RhoExprWithBlock(expr, block) +: acc
+    val exprsWithBlock = dbs.foldLeft(collection.immutable.List[RhoExprWithBlock]()) {
+      (acc, data) =>
+        val exprs = data.postBlockData.flatMap(exprFromParProto)
+        // Implements semantic of Par with Unit: P | Nil ==> P
+        val expr  = if (exprs.size == 1) exprs.head else ExprPar(exprs.toList)
+        val block = data.block
+        RhoExprWithBlock(expr, block) +: acc
     }
     DataAtNameResponse(exprsWithBlock, length)
   }
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -262,7 +262,7 @@ object WebApi {
 
   // RhoExpr from protobuf
 
-  private def exprFromParProto(par: Par): Option[RhoExpr] = {
+  def exprFromParProto(par: Par): Option[RhoExpr] = {
     val exprs =
       par.exprs.flatMap(exprFromExprProto) ++
         par.unforgeables.flatMap(unforgFromProto) ++
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -373,7 +373,7 @@ object WebApi {
   }
 
   private def unforgToParProto(unforg: RhoUnforg): Par =
-    Par(unforgeables = Seq(GUnforgeable(unforgToUnforgProto(unforg))))
+    Unforgeable(GUnforgeable(unforgToUnforgProto(unforg)))
 
   // Data request/response protobuf wrappers
 
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -1,16 +1,16 @@
 package coop.rchain.node.api
 
 import cats.effect.Sync
+import cats.kernel.Monoid
 import cats.syntax.all._
 import com.google.protobuf.ByteString
 import coop.rchain.casper.api.BlockApi
 import coop.rchain.casper.protocol.{BlockInfo, DataWithBlockInfo, DeployData, LightBlockInfo}
 import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.{SignaturesAlg, Signed}
-import coop.rchain.models.GUnforgeable.UnfInstance.{GDeployIdBody, GDeployerIdBody, GPrivateBody}
-import coop.rchain.models._
 import coop.rchain.models.rholang.implicits.VectorPar
 import coop.rchain.models.syntax._
+import coop.rchain.models.{Bundle, ETuple, Expr, GUnforgeable, Par}
 import coop.rchain.node.api.WebApi._
 import coop.rchain.node.web.{CacheTransactionAPI, TransactionResponse}
 import coop.rchain.rholang.interpreter.RhoType._
@@ -335,10 +335,33 @@ object WebApi {
 
   // RhoExpr to protobuf
 
-  private def rhoExprToParProto(exp: RhoExpr): Par = exp match {
+  /**
+    * Private definition of [[Monoid]] for [[Par]] with the implementation of
+    * combining fields used by WebApi representation of Par [[RhoExpr]].
+    */
+  implicit private object ParMonoid extends Monoid[Par] {
+    override def empty: Par = VectorPar()
+
+    override def combine(x: Par, y: Par): Par =
+      x.copy(
+        x.sends ++ y.sends,
+        x.receives ++ y.receives,
+        x.news ++ y.news,
+        x.exprs ++ y.exprs,
+        x.matches ++ y.matches,
+        x.unforgeables ++ y.unforgeables,
+        x.bundles ++ y.bundles,
+        x.connectives ++ y.connectives
+      )
+  }
+
+  /**
+    * Converts [[RhoExpr]] WebApi representation to protobuf [[Par]] type.
+    */
+  def rhoExprToParProto(exp: RhoExpr): Par = exp match {
     // Nested expressions (Par, Tuple, List and Set are converted to JSON list)
-    case data: ExprPar   => exprParToParProto(data)
-    case ExprTuple(data) => Expression(Expr().withETupleBody(ETuple(data.map(rhoExprToParProto))))
+    case ExprPar(data)   => data.map(rhoExprToParProto).combineAll
+    case ExprTuple(data) => TupleN(data.map(rhoExprToParProto))
     case ExprList(data)  => List(data.map(rhoExprToParProto))
     case ExprSet(data)   => Set(data.map(rhoExprToParProto))
     case ExprMap(data)   => Map(data.map { case (k, v) => (String(k), rhoExprToParProto(v)) })
@@ -352,28 +375,11 @@ object WebApi {
     case ExprUnforg(data) => unforgToParProto(data)
   }
 
-  private def exprParToParProto(expr: ExprPar): Par = {
-    val exprs = expr.data.map(rhoExprToParProto)
-    VectorPar().copy(
-      exprs.flatMap(_.sends),
-      exprs.flatMap(_.receives),
-      exprs.flatMap(_.news),
-      exprs.flatMap(_.exprs),
-      exprs.flatMap(_.matches),
-      exprs.flatMap(_.unforgeables),
-      exprs.flatMap(_.bundles),
-      exprs.flatMap(_.connectives)
-    )
+  private def unforgToParProto(unforg: RhoUnforg): Par = unforg match {
+    case UnforgPrivate(name)  => Name(name.unsafeHexToByteString)
+    case UnforgDeploy(name)   => DeployId(name.unsafeDecodeHex)
+    case UnforgDeployer(name) => DeployerId(name.unsafeDecodeHex)
   }
-
-  private def unforgToUnforgProto(unforg: RhoUnforg): GUnforgeable.UnfInstance = unforg match {
-    case UnforgPrivate(name)  => GPrivateBody(GPrivate(name.unsafeHexToByteString))
-    case UnforgDeploy(name)   => GDeployIdBody(GDeployId(name.unsafeHexToByteString))
-    case UnforgDeployer(name) => GDeployerIdBody(GDeployerId(name.unsafeHexToByteString))
-  }
-
-  private def unforgToParProto(unforg: RhoUnforg): Par =
-    Unforgeable(GUnforgeable(unforgToUnforgProto(unforg)))
 
   // Data request/response protobuf wrappers
 
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -9,6 +9,7 @@ import coop.rchain.crypto.PublicKey
 import coop.rchain.crypto.signatures.{SignaturesAlg, Signed}
 import coop.rchain.models.GUnforgeable.UnfInstance.{GDeployIdBody, GDeployerIdBody, GPrivateBody}
 import coop.rchain.models._
+import coop.rchain.models.rholang.implicits.VectorPar
 import coop.rchain.models.syntax._
 import coop.rchain.node.api.WebApi._
 import coop.rchain.node.web.{CacheTransactionAPI, TransactionResponse}
@@ -353,22 +354,15 @@ object WebApi {
 
   private def exprParToParProto(expr: ExprPar): Par = {
     val exprs = expr.data.map(rhoExprToParProto)
-
-    def merge[A](f: Par => Seq[A]): Seq[A] = exprs.foldLeft(Seq.empty[A]) {
-      case (acc, p) => acc ++ f(p)
-    }
-
-    Par(
-      merge(_.sends),
-      merge(_.receives),
-      merge(_.news),
-      merge(_.exprs),
-      merge(_.matches),
-      merge(_.unforgeables),
-      merge(_.bundles),
-      merge(_.connectives),
-      exprs.headOption.getOrElse(Par()).locallyFree,
-      exprs.headOption.getOrElse(Par()).connectiveUsed
+    VectorPar().copy(
+      exprs.flatMap(_.sends),
+      exprs.flatMap(_.receives),
+      exprs.flatMap(_.news),
+      exprs.flatMap(_.exprs),
+      exprs.flatMap(_.matches),
+      exprs.flatMap(_.unforgeables),
+      exprs.flatMap(_.bundles),
+      exprs.flatMap(_.connectives)
     )
   }
 
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -355,7 +355,7 @@ object WebApi {
       case ExprInt(data)    => exprProtoToParProto(Expr().withGInt(data))
       case ExprString(data) => exprProtoToParProto(Expr().withGString(data))
       case ExprUri(data)    => exprProtoToParProto(Expr().withGUri(data))
-      // Binary data is encoded as base16 string
+      // Binary data is decoded from base16 string
       case ExprBytes(data) =>
         exprProtoToParProto(Expr().withGByteArray(data.unsafeHexToByteString))
       case ExprUnforg(data) => unforgToParProto(data)
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -12,7 +12,7 @@ import coop.rchain.models._
 import coop.rchain.models.syntax._
 import coop.rchain.node.api.WebApi._
 import coop.rchain.node.web.{CacheTransactionAPI, TransactionResponse}
-import coop.rchain.rholang.interpreter.RhoType.Expression
+import coop.rchain.rholang.interpreter.RhoType.{Boolean, ByteArray, Expression, Number, String, Uri}
 
 trait WebApi[F[_]] {
   def status: F[ApiStatus]
@@ -345,12 +345,12 @@ object WebApi {
         case (k, v) => (rhoExprToParProto(ExprString(k)), rhoExprToParProto(v))
       }.toList)))
     // Terminal expressions (here is the data)
-    case ExprBool(data)   => Expression(Expr().withGBool(data))
-    case ExprInt(data)    => Expression(Expr().withGInt(data))
-    case ExprString(data) => Expression(Expr().withGString(data))
-    case ExprUri(data)    => Expression(Expr().withGUri(data))
+    case ExprBool(data)   => Boolean(data)
+    case ExprInt(data)    => Number(data)
+    case ExprString(data) => String(data)
+    case ExprUri(data)    => Uri(data)
     // Binary data is decoded from base16 string
-    case ExprBytes(data)  => Expression(Expr().withGByteArray(data.unsafeHexToByteString))
+    case ExprBytes(data)  => ByteArray(data.unsafeHexToByteString.toByteArray)
     case ExprUnforg(data) => unforgToParProto(data)
   }
 
--- a/node/src/main/scala/coop/rchain/node/api/WebApi.scala
+++ b/node/src/main/scala/coop/rchain/node/api/WebApi.scala
@@ -12,7 +12,7 @@ import coop.rchain.models._
 import coop.rchain.models.syntax._
 import coop.rchain.node.api.WebApi._
 import coop.rchain.node.web.{CacheTransactionAPI, TransactionResponse}
-import coop.rchain.rholang.interpreter.RhoType.{Boolean, ByteArray, Expression, Number, String, Uri}
+import coop.rchain.rholang.interpreter.RhoType._
 
 trait WebApi[F[_]] {
   def status: F[ApiStatus]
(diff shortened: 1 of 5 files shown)


node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -248,6 +244,7 @@ object WebApi {
   // Binary converters - protobuf uses ByteString and in JSON is base16 string
 
   private def toHex(bs: ByteString) = Base16.encode(bs.toByteArray)
+  private def fromHex(s: String)    = s.unsafeDecodeHex.toByteString
```
tgrospic: We can now remove both of these functions because they are defined as extensions.
https://github.com/rchain/rchain/blob/44b6febd85277df1a1e34c83752b84d342f62327/models/src/main/scala/coop/rchain/models/ByteStringSyntax.scala#L36
https://github.com/rchain/rchain/blob/a8ae9399ccf3eef7219f029bb5d231307993a4af/models/src/main/scala/coop/rchain/models/StringSyntax.scala#L13-L14

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -308,6 +305,30 @@ object WebApi {
       ExprMap(fields.toMap).some
     } else none
 
+  private def parProtoFromExpr(exp: RhoExpr): Par = {
```
tgrospic: This function should go just below `RhoExpr to protobuf` comment and should be named in the same way as others `exprToParProto`.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -330,13 +351,14 @@ object WebApi {
     case UnforgDeployer(name) => GDeployerIdBody(GDeployerId(name.unsafeHexToByteString))
   }
 
+  private def unforgToPar(unforg: RhoUnforg): Par =
```
tgrospic: This name should have suffix `Proto`.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -308,6 +305,30 @@ object WebApi {
...
+      case ExprMap(data) =>
+        exprToPar(Expr().withEMapBody(ParMap(data.map {
+          case (k, v) => (parProtoFromExpr(ExprString(k)), parProtoFromExpr(v))
+        }.toList)))
+      // Terminal expressions (here is the data)
+      case ExprBool(data)   => exprToPar(Expr().withGBool(data))
+      case ExprInt(data)    => exprToPar(Expr().withGInt(data))
+      case ExprString(data) => exprToPar(Expr().withGString(data))
+      case ExprUri(data)    => exprToPar(Expr().withGUri(data))
+      // Binary data is encoded as base16 string
```
tgrospic: Maybe it would be more clear to say that data is decoded from base 16 string?

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -308,6 +305,30 @@ object WebApi {
       ExprMap(fields.toMap).some
     } else none
 
+  private def parProtoFromExpr(exp: RhoExpr): Par = {
+    def exprToPar(exp: Expr): Par = Par(exprs = Seq(exp))
```
tgrospic: To be more precise `exprProtoToParProto` would mean it's all about proto types.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3718/commits/3559da9d0b91d948558d0c6765379790b93d4392).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3718/commits/a8e52338ef1efd7498ecaf3a37665857609e922f).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3718/commits/d9f613486bb84a973be0fa79e11fd61a3a770066).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: Yes, make sense. I [changed](https://github.com/rchain/rchain/pull/3718/commits/1dc98e2df15208b1428b042939442f6e8d1e5e5d) comment.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: This leads to ugly line wrapping but [done](https://github.com/rchain/rchain/pull/3718/commits/449c161268389042e235aa0430ea50c61d647ddf) :slightly_smiling_face:

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +329,53 @@ object WebApi {
   private def exprFromBundleProto(b: Bundle): Option[RhoExpr] = exprFromParProto(b.body)
 
   private def mkUnforgExpr(f: String => RhoUnforg, bs: ByteString): ExprUnforg =
-    ExprUnforg(f(toHex(bs)))
+    ExprUnforg(f(bs.toHexString))
 
   // RhoExpr to protobuf
 
+  private def rhoExprToParProto(exp: RhoExpr): Par = {
+    def exprProtoToParProto(exp: Expr): Par = Par(exprs = Seq(exp))
```
tgrospic: I always forgot that we have nice constructors for proto in RhoType.
https://github.com/rchain/rchain/blob/0bfc2aa13e73a27c1f21200dee7336bd252ea17a/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala#L130-L134

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +329,53 @@ object WebApi {
...
+      case ExprList(data) =>
+        exprProtoToParProto(Expr().withEListBody(EList(data.map(rhoExprToParProto))))
+      case ExprSet(data) =>
+        exprProtoToParProto(Expr().withESetBody(ParSet(data.map(rhoExprToParProto))))
+      case ExprMap(data) =>
+        exprProtoToParProto(Expr().withEMapBody(ParMap(data.map {
+          case (k, v) => (rhoExprToParProto(ExprString(k)), rhoExprToParProto(v))
+        }.toList)))
+      // Terminal expressions (here is the data)
+      case ExprBool(data)   => exprProtoToParProto(Expr().withGBool(data))
```
tgrospic: This can then be `Boolean(data)`.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +329,53 @@ object WebApi {
...
+      case ExprTuple(data) =>
+        exprProtoToParProto(Expr().withETupleBody(ETuple(data.map(rhoExprToParProto))))
+      case ExprList(data) =>
+        exprProtoToParProto(Expr().withEListBody(EList(data.map(rhoExprToParProto))))
+      case ExprSet(data) =>
+        exprProtoToParProto(Expr().withESetBody(ParSet(data.map(rhoExprToParProto))))
+      case ExprMap(data) =>
+        exprProtoToParProto(Expr().withEMapBody(ParMap(data.map {
+          case (k, v) => (rhoExprToParProto(ExprString(k)), rhoExprToParProto(v))
+        }.toList)))
```
tgrospic: You can extend RhoType in the same way for List, Set and Map.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +329,53 @@ object WebApi {
...
 
   // RhoExpr to protobuf
 
+  private def rhoExprToParProto(exp: RhoExpr): Par = {
+    def exprProtoToParProto(exp: Expr): Par = Par(exprs = Seq(exp))
+
+    exp match {
+      // Nested expressions (Par, Tuple, List and Set are converted to JSON list)
+      case ExprPar(data) =>
+        exprProtoToParProto(Expr().withEListBody(EList(data.map(rhoExprToParProto))))
```
tgrospic: This is not correct, ExprPar represents Par in proto, it's not the same as Expr List.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: Can you explain please? Do you mean this change?
```scala
case Boolean(data) => Expression(Expr().withGBool(data))
```
Why only for `Boolean`? What about `Long` and `String`? Why it is necessary?

stanislavlyalin: Oh, I see what you mean
https://github.com/rchain/rchain/blob/0bfc2aa13e73a27c1f21200dee7336bd252ea17a/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala#L42-L50

tgrospic: Here are a few examples.
```scala
import ...RhoType._

val b = Boolean(true) // : Par
val n = Number(42L) // : Par
val s = String("hey!") // : Par
```

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: @tgrospic, I'm not sure how to correctly fix it. As far as I remember `Par` can contain nested `Par`s in `Bundle` field. So I'm trying to reqursively build `Par` from nested `Par`s as sequence of Bundles.
```scala
case ExprPar(data) => Par(bundles = data.map(expr => Bundle(rhoExprToParProto(expr))))
```

tgrospic: Something like this where all fields are merged together.
```scala
def exprParToParProto(expr: ExprPar) = {
  val exprs = expr.data.map(rhoExprToParProto)
  exprs.foldLeft(Par()) {
    case (acc, p) =>
      acc.copy(
        sends = acc.sends ++ p.sends,
        exprs = acc.exprs ++ p.exprs
        // ...
      )
  }
}
```
But this is not the right way because we first create Par for each RhoExpr case and then merge them together. So better approach would be to do the _merging_ without creating intermediate Par's.

And for Bundles, it's not possible to create them with input from WebAPI. Only subset of Par types are available in RhoExpr types.

stanislavlyalin: Please take a look [8f55e85](https://github.com/rchain/rchain/pull/3718/commits/8f55e8563a403c128564d58d788c8aaba630ccd7).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +330,62 @@ object WebApi {
...
+      merge(_.sends),
+      merge(_.receives),
+      merge(_.news),
+      merge(_.exprs),
+      merge(_.matches),
+      merge(_.unforgeables),
+      merge(_.bundles),
+      merge(_.connectives),
+      exprs.headOption.getOrElse(Par()).locallyFree,
+      exprs.headOption.getOrElse(Par()).connectiveUsed
```
tgrospic: Information about locallyFree and connectiveUsed is not present in RhoExpr types so it cannot be populated to Par object.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +330,62 @@ object WebApi {
...
+    // Binary data is decoded from base16 string
+    case ExprBytes(data)  => ByteArray(data.unsafeHexToByteString.toByteArray)
+    case ExprUnforg(data) => unforgToParProto(data)
+  }
+
+  private def exprParToParProto(expr: ExprPar): Par = {
+    val exprs = expr.data.map(rhoExprToParProto)
+
+    def merge[A](f: Par => Seq[A]): Seq[A] = exprs.foldLeft(Seq.empty[A]) {
+      case (acc, p) => acc ++ f(p)
```
tgrospic: There is a lot of concatenation involved here. What is the performance characteristic of this code in regard to this documentation?
https://docs.scala-lang.org/overviews/collections-2.13/performance-characteristics.html

The goal is to iterate as minimal as possible and use structure/operations which is best suited for specific case.
Hint: special constructor for Par which uses another structure for collections.
https://github.com/rchain/rchain/blob/e4bcf60be2971a6f619b3b0ec8a1ad1cb1d82c57/models/src/main/scala/coop/rchain/models/rholang/implicits.scala#L207-L218

---

Please create tests for this function so we can be sure it's correctly transforms data.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +330,62 @@ object WebApi {
...
+  }
+
   private def unforgToUnforgProto(unforg: RhoUnforg): GUnforgeable.UnfInstance = unforg match {
     case UnforgPrivate(name)  => GPrivateBody(GPrivate(name.unsafeHexToByteString))
     case UnforgDeploy(name)   => GDeployIdBody(GDeployId(name.unsafeHexToByteString))
     case UnforgDeployer(name) => GDeployerIdBody(GDeployerId(name.unsafeHexToByteString))
   }
 
+  private def unforgToParProto(unforg: RhoUnforg): Par =
+    Par(unforgeables = Seq(GUnforgeable(unforgToUnforgProto(unforg))))
```
tgrospic: We can use RhoType here also.
https://github.com/rchain/rchain/blob/8f55e8563a403c128564d58d788c8aaba630ccd7/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala#L142-L147

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: > What is the performance characteristic of this code in regard to this documentation?

`n * L`, where `n = 8` - number of fields, and `L` is length `exprs`

> Hint: special constructor for Par which uses another structure for collections.

I'm trying this solution
```scala
VectorPar().copy(
  exprs.flatMap(_.sends),
  exprs.flatMap(_.receives),
  exprs.flatMap(_.news),
  exprs.flatMap(_.exprs),
  exprs.flatMap(_.matches),
  exprs.flatMap(_.unforgeables),
  exprs.flatMap(_.bundles),
  exprs.flatMap(_.connectives)
)
```
Is it looks correct? I will add test for it.

stanislavlyalin: I [changed](https://github.com/rchain/rchain/pull/3718/commits/d51e04ef716004eb08c16eb10592e2d31c5d322e) creating `Par` instance with `VectorPar` object and [added](https://github.com/rchain/rchain/pull/3718/commits/7287f5db7c7c9bfa16289b2da12af61b8679157a) corresponding test.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3718/commits/aaf616747e9ef78b0e5afa0ec204d581c4473ab5).

models/src/test/scala/coop/rchain/models/VectorParSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+    constructedViaCopy.sends shouldBe sends
+    constructedViaCopy.receives shouldBe receives
+    constructedViaCopy.news shouldBe news
+    constructedViaCopy.exprs shouldBe exprs
+    constructedViaCopy.matches shouldBe matches
+    constructedViaCopy.unforgeables shouldBe unforgeables
+    constructedViaCopy.bundles shouldBe bundles
+    constructedViaCopy.connectives shouldBe connectives
+  }
```
tgrospic: What is tested here?

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,23 +331,55 @@ object WebApi {
...
 
   private def mkUnforgExpr(f: String => RhoUnforg, bs: ByteString): ExprUnforg =
-    ExprUnforg(f(toHex(bs)))
+    ExprUnforg(f(bs.toHexString))
 
   // RhoExpr to protobuf
 
+  private def rhoExprToParProto(exp: RhoExpr): Par = exp match {
+    // Nested expressions (Par, Tuple, List and Set are converted to JSON list)
+    case data: ExprPar   => exprParToParProto(data)
```
tgrospic: Is there a reason to use different _case_ style for this case?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
```diff
@@ -79,6 +79,36 @@ object RhoType {
...
+
+  type RhoMap = Map.type
+  object RhoMap {
+    def unapply(p: Par): Option[Map[Par, Par]] =
+      p.singleExpr().collect {
+        case Expr(EMapBody(ParMap(s, _, _, _))) => s.toMap
+      }
+
+    def apply(s: Map[Par, Par]): Par = EMapBody(ParMap(s.toSeq))
+  }
```
tgrospic, line 134: Existing methods use different naming convention. Can you please follow the same logic. I'm not saying what's wrong or right just to be consistent in all definitions.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: With `case ExprPar(data)` data has type `List[RhoExpr]`. I little bit [confused](https://github.com/rchain/rchain/pull/3718#discussion_r892721910) :thinking: As I understood we should not process `ExprPar` as list of Expr.

models/src/test/scala/coop/rchain/models/VectorParSpec.scala
stanislavlyalin: It is checked that the fields packed in `Par` exactly match the original ones. What do you mean by the [test](https://github.com/rchain/rchain/pull/3718#discussion_r902409364) for the VectorPar constructor?

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala
stanislavlyalin: The problem was the ambiguity between the regular `List` and the RhoType `List`. I brought the name in line with the previous definitions and [tried to solve](https://github.com/rchain/rchain/pull/3718/commits/d42243d33e46177a938812798e499198e861bc2b) the problem of ambiguity.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
tgrospic: Sorry, I was not precise enough. The function I have in mind to test is `rhoExprToParProto`.

We need to be sure that data from WebApi is correctly transformed to Par type for further processing.

VectorPar has all collection fields initialized to Vector so any operation of concatenation will not introduce overhead of iteration of all elements like List.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,32 +331,71 @@ object WebApi {
...
+        x.connectives ++ y.connectives
+      )
+  }
+
+  /**
+    * Converts [[RhoExpr]] WebApi representation to protobuf [[Par]] type.
+    */
+  def rhoExprToParProto(exp: RhoExpr): Par = exp match {
+    // Nested expressions (Par, Tuple, List and Set are converted to JSON list)
+    case ExprPar(data)   => data.map(rhoExprToParProto).combineAll
```
tgrospic, line 363: `combineAll` is Monoid `combine` which `fold` over list with `empty` as a start accumulator value (also from Monoid instance).

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,32 +331,71 @@ object WebApi {
...
+        x.unforgeables ++ y.unforgeables,
+        x.bundles ++ y.bundles,
+        x.connectives ++ y.connectives
+      )
+  }
+
+  /**
+    * Converts [[RhoExpr]] WebApi representation to protobuf [[Par]] type.
+    */
+  def rhoExprToParProto(exp: RhoExpr): Par = exp match {
```
tgrospic, line 361: ~~Generative tests can be written as a round trip from `Par => Par` composition of `exprFromParProto andThen rhoExprToParProto`.~~

Better to test `RhoExpr => RhoExpr` (composition of `rhoExprToParProto andThen exprFromParProto`) then Par because conversion of Map is not isomorphic. Conversion from Par to RhoExpr _forgets_ types of keys.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,32 +331,71 @@ object WebApi {
...
 
-  private def unforgToUnforgProto(unforg: RhoUnforg): GUnforgeable.UnfInstance = unforg match {
-    case UnforgPrivate(name)  => GPrivateBody(GPrivate(name.unsafeHexToByteString))
-    case UnforgDeploy(name)   => GDeployIdBody(GDeployId(name.unsafeHexToByteString))
-    case UnforgDeployer(name) => GDeployerIdBody(GDeployerId(name.unsafeHexToByteString))
+  /**
+    * Private definition of [[Monoid]] for [[Par]] with the implementation of
+    * combining fields used by WebApi representation of Par [[RhoExpr]].
+    */
+  implicit private object ParMonoid extends Monoid[Par] {
```
stanislavlyalin, line 342: Why not `implicit private val parMonoid: Monoid[Par] = new Monoid[Par] ...`? I think this is the more common way.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,32 +331,71 @@ object WebApi {
...
+        x.sends ++ y.sends,
+        x.receives ++ y.receives,
+        x.news ++ y.news,
+        x.exprs ++ y.exprs,
+        x.matches ++ y.matches,
+        x.unforgeables ++ y.unforgeables,
+        x.bundles ++ y.bundles,
+        x.connectives ++ y.connectives
+      )
+  }
```
stanislavlyalin, line 356: Is it simpler than just `foldLeft`? I guess it has the same complexity but has more specific implementation.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
tgrospic: I don't have strong preference, but in this way we have a type level entity which we can reference and compare, instead of anonymous type.

node/src/main/scala/coop/rchain/node/api/WebApi.scala
tgrospic: The main advantage is to be compositional e.g. to get `Monoid[List[Int]]` we can _reuse_ `Monoid[Int]`.
```scala
import cats.syntax.all._

val r2 = 4 |+| 2 // 6

val r1 = List(4) |+| List(2) // List(4, 2)
```
Here is a very good article with concise explanation what problems can solve Semigroup and Monoid.
https://blog.rockthejvm.com/semigroups-and-monoids-in-scala/

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+      roundTripTest(ExprPar(listExpr)).collect { case ExprPar(data)     => data } shouldBe listExpr.some
+      roundTripTest(ExprTuple(listExpr)).collect { case ExprTuple(data) => data } shouldBe listExpr.some
+      roundTripTest(ExprList(listExpr)).collect { case ExprList(data)   => data } shouldBe listExpr.some
+      roundTripTest(ExprSet(listExpr)).collect { case ExprSet(data)     => data } shouldBe listExpr.some
+
+      val mapExpr =
+        Map[String, RhoExpr]("a" -> ExprBool(true), "b" -> ExprInt(42), "c" -> ExprList(listExpr))
+      roundTripTest(ExprMap(mapExpr)).collect { case ExprMap(data) => data } shouldBe mapExpr.some
+    }
+  }
```
tgrospic, line 22: This kind of test is ideal to be generative using ScalaCheck.

Very simple example is check for DeployData serializer. Arbitrary instance will generate many different instances and encoding and then decoding must get to the same data. If failed case is found ScalaCheck will try to find shrinked version of data which fails.

Going from RhoExpr to RhoExpr should get the same object.
https://github.com/rchain/rchain/blob/29e96c59d650dfff54a5dc1ffd370c80a25a9724/models/src/test/scala/coop/rchain/casper/protocol/DeployDataSpec.scala#L10-L32

stanislavlyalin: [Done](https://github.com/rchain/rchain/pull/3718/commits/94cb8789063fdfc5d221aef19fd0a007537302d7).

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -3,38 +3,122 @@ package coop.rchain.node
...
 
-  "WebAPI RhoExpr <-> Par conversion" should {
-    "work correct with simple types" in {
-      roundTripTest(ExprBool(true)).collect { case ExprBool(data)            => data } shouldBe true.some
-      roundTripTest(ExprInt(42)).collect { case ExprInt(data)                => data } shouldBe 42.some
-      roundTripTest(ExprString("abc")).collect { case ExprString(data)       => data } shouldBe "abc".some
-      roundTripTest(ExprUri("http://test.com")).collect { case ExprUri(data) => data } shouldBe "http://test.com".some
+  // Simple types
+
+  property("WebAPI ExprBool <-> Par conversion work correct") {
```
tgrospic: You don't need to do each case manually. Leave this to ScalaCheck to do. Just define Arbitrary[RhoExpr] and check `RhoExpr <=> RhoExpr`.

HINT: https://github.com/alexarchambault/scalacheck-shapeless#usage

node/src/main/scala/coop/rchain/node/api/WebApi.scala
```diff
@@ -335,32 +331,71 @@ object WebApi {
...
 
   private def toDataAtNameResponse(req: (Seq[DataWithBlockInfo], Int)): DataAtNameResponse = {
     val (dbs, length) = req
-    val exprsWithBlock = dbs.foldLeft(List[RhoExprWithBlock]()) { (acc, data) =>
-      val exprs = data.postBlockData.flatMap(exprFromParProto)
-      // Implements semantic of Par with Unit: P | Nil ==> P
-      val expr  = if (exprs.size == 1) exprs.head else ExprPar(exprs.toList)
-      val block = data.block
-      RhoExprWithBlock(expr, block) +: acc
+    val exprsWithBlock = dbs.foldLeft(collection.immutable.List[RhoExprWithBlock]()) {
```
tgrospic, line 392: Does the full namespace for List is needed? It seems to work fine on other places.

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+          genExprString,
+          genExprUri,
+          genExprBytes,
+          genExprUnforg
+        )
+      )
+
+    // Helpers
+    def withList[A <: RhoExpr](f: List[RhoExpr] => A): Gen[A] =
+      Gen.listOfN(numOfElements, genRhoExpr).map(data => f(data))
```
tgrospic: This is just `.map(f)`.

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+import coop.rchain.models.syntax._
+import coop.rchain.node.api.WebApi._
+import org.scalacheck.Arbitrary._
+import org.scalacheck.{Arbitrary, Gen}
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.propspec.AnyPropSpec
+import org.scalatestplus.scalacheck.ScalaCheckDrivenPropertyChecks
+
+class RhoExprToParSpec extends AnyPropSpec with ScalaCheckDrivenPropertyChecks with Matchers {
+  property("WebAPI RhoExpr <-> Par conversion work correct") {
```
tgrospic: This should say `RhoExpr => Par =>  RhoExpr`, right?

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -0,0 +1,71 @@
...
+    forAll { rhoExpr: RhoExpr =>
+      // TODO: Remove after finishing debugging
+      // println(expected)
+      // println(s"  $actual")
+
+      exprFromParProto(rhoExprToParProto(rhoExpr)) shouldBe rhoExpr.some
```
tgrospic: You can even define this function `RhoExpr => RhoExpr` with function composition.
```scala
val rhoExprToRhoExpr = rhoExprToParProto >>> exprFromParProto
// or
val rhoExprToRhoExpr = rhoExprToParProto andThen exprFromParProto
```

node/src/main/scala/coop/rchain/node/api/WebApi.scala
stanislavlyalin: It is conflicted with `List` from RhoType.scala

https://github.com/rchain/rchain/blob/6ad914c91e5a9c499c5736f651ceffb3ed7ac147/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala#L107

But I don't know why it is works in other cases e.g. here

https://github.com/rchain/rchain/blob/28db3868ff83703c9dad18f69db8ce9baeba8c8b/node/src/main/scala/coop/rchain/node/api/WebApi.scala#L37

tgrospic: Interesting. We should definitely plan to change RhoType with names not in conflict with built-in types.

node/src/test/scala/coop/rchain/node/RhoExprToParSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  // Converts RhoExpr to Par and then back to RhoExpr
+  private val rhoExprToRhoExpr = rhoExprToParProto _ andThen exprFromParProto
+
+  implicit private def aArb: Arbitrary[RhoExpr] = {
+
+    val numOfElements = 2 // StackOverflowError with n >= 3
+
+    // def genExprPar   = withList(ExprPar)
```
stanislavlyalin: Converting to `Par` removes nesting, as noted in [message](<https://discord.com/channels/391435113415573504/410182072054054912/992439986412584980>). Therefore, comparing the original `ExprPar` and the received one leads to a test error. We need to decide if we can ignore nesting in this case.

tgrospic:
Let's for now just skip generating ExprPar in test. Until we have basic stuff done right this will be just waste of time.
For ExprPar we can create some simpler test which will not generate nested ExprPar's. It will be good enough for now.

The main problem is that `ExprPar(List(x, y))` should be the same as `ExprPar(List(y, x))`. To really fix this we should use MultiSet instead of List to hold x and y. Scala doesn't have MultiSet (only in 2.13). I tried to use simple version of multi set `Map[RhoExpr, Int]` with conversions to List[RhoExpr] but it was not possible to make JsonEncoder in WebAPI to automatically generate derivations for all types.

tgrospic (approved):
LGTM :tada:

zsluedem (approved):
LGTM. It is more applicable although some strange `Par` is still not possible.

stanislavlyalin:
bors merge

Issue #3685, opened 2022-05-05
Configure sepearete chain specific unforgeable parameter


zsluedem:
After #3720 , the unforgeable can not be predicted anymore. And some of the chain-specific parameters are no more deterministic. Different chains like testnet, mainnet and other private networks would get totally different unforgeable names for these. We should provide a way to config this.

1. NonNegativeMergeableTag(neccessary for running the node) https://github.com/rchain/rchain/blob/b89d6ec56ce437727aafcdf3c4dc21c3262f36ec/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala#L26-L34
2. transferUnforgeable (needed only for observer which enable transaction api)https://github.com/rchain/rchain/blob/b89d6ec56ce437727aafcdf3c4dc21c3262f36ec/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala#L26-L34
3. storeTokenUnforgeable (needed only for taking a snapshot)https://github.com/rchain/rchain/blob/b89d6ec56ce437727aafcdf3c4dc21c3262f36ec/node/src/main/scala/coop/rchain/node/revvaultexport/RhoTrieTraverser.scala#L78-L83
4. getGenesisVaultMapPar(needed only for taking a snapshot)https://github.com/rchain/rchain/blob/b89d6ec56ce437727aafcdf3c4dc21c3262f36ec/node/src/main/scala/coop/rchain/node/revvaultexport/StateBalances.scala#L42-L60

The parameters above should be configured based on the chain.**It would be great to have some default values for mainnet, testnet**

PR #3720, opened 2022-05-31
Introduce new random seed for rholang deploy

### LINKED ISSUE #3691: Introduce new random seed for rholang deploy
## Overview

Currently we use deployer and timestamp for rholang execution random seed which is used to generate unforgeable names.
https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/casper/src/main/scala/coop/rchain/casper/rholang/Tools.scala#L9-L13

This is not secure enough because we cannot ensure that the same user will not repeat the same timestamp in the future when validator compacted its state and removed old data.

After https://github.com/rchain/rchain/pull/3687, RNode doesn't need predictable unforgeable name for `InsertSigned` at all. So we should change our random seed generating algorithm to have more secure way to generate unforgeable names.

Releated issues:
- https://rchain.atlassian.net/browse/RCHAIN-1048
- https://github.com/rchain-test/rchain-secure/issues/3

## Design

The random seed should generated by:

1. [shardId](https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/models/src/main/protobuf/CasperMessage.proto#L75)
1. [blockNumber](https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/models/src/main/protobuf/CasperMessage.proto#L161)
1. [sender](https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/models/src/main/protobuf/CasperMessage.proto#L71)
1. [preStateHash](https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/models/src/main/protobuf/CasperMessage.proto#L155)

## Algorithm

The algorithm is to serialize `shardId`, `blockNumber`, `sender` (public key) and `preStateHash` into a byte array and blake hash the byte array.
The generated random seed above should be used for the block. Let's supposed that there are 3 deploys in the block.
Then the random seed for each deploy would be 
```
val randomseed = Blake2b512Random(hash(shardId:blockNumber:sender:preStateHash))
val deploy1seed = randomseed.split(1)
val deploy2seed = randomseed.split(2)
val deploy3seed = randomseed.split(3)
```

### Explore deploy

For the explore deploy we should use randomly generated seed to ensure there is no conflict.
https://github.com/rchain/rchain/blob/0ec5f6035909078b0fd108e9e7e8bf7f47bd71cd/casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala#L468-L477

In the same way how is done in evaluate function:
`val rand: Blake2b512Random = Blake2b512Random(128)`.
https://github.com/rchain/rchain/blob/6dd06305020f02f9fed99c7eda6cbd980fdb0b7d/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala#L37-L54

## Solution

Create case class with needed fields with corresponding `Codec[BlockRandomSeed]` instance. Companion object should have functions to get the `Blake2b512Random` instance.

```scala
case class BlockRandomSeed(
  shardId: String,
  blockNumber: Long,
  sender: PublicKey,
  preStateHash: Blake2b256Hash
)
```

Changes should be implemented in two PRs:

- [ ] Creation of a case class, codec, serialization and random seed function
- [ ] Use of new random seed for a block in play and replay

zsluedem:
## Overview

Bloked by https://github.com/rchain/rchain/pull/3729

1. use a new case class `BlockRandomSeed` to generate seed for a block
2. explore deploy use a random seed instead of combination of  deployer and timestamp
3. play and replay use new random seed

### Notes

Some constant number
```
  val PreChargeSplitIndex: Byte  = 1.toByte
  val UserDeploySplitIndex: Byte = 2.toByte
  val RefundSplitIndex: Byte     = 3.toByte
```

The deploys in Genesis block would use random seed like:
`rand.splitByte(index.toByte).splitByte(UserDeploySplitIndex)`


casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
```diff
@@ -1,3 +1,7 @@
+package coop.rchain.casper
```
zsluedem, line 1: The changes in this file should not be the final result. Currently the changes here is to make the code compile and tests pass. The final result should be taken in https://github.com/rchain/rchain/pull/3729 . This pr would be merged after #3729 .

tgrospic: We can put this file inside `rholang` folder because it's dealing with Rholang state.

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
```diff
@@ -102,15 +103,16 @@ object ReportingCasper {
           _         <- reportingRuntime.setBlockData(blockdata)
 
           // Reset runtime (in-memory) state
-          _ <- reportingRuntime.reset(Blake2b256Hash.fromByteString(preStateHash))
-
+          _    <- reportingRuntime.reset(Blake2b256Hash.fromByteString(preStateHash))
+          rand = BlockRandomSeed.fromBlock(block)
```
tgrospic: Please move this line below the comment and leave one empty line to make it more clear where rand variable is part of.

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
```diff
@@ -180,7 +188,12 @@ object ReportingRuntime {
       mergeChs <- Ref.of(Set[Par]())
       rhoEnv <- {
         implicit val c = cost
-        createRhoEnv(reporting, mergeChs, Genesis.NonNegativeMergeableTagName, extraSystemProcesses)
+        createRhoEnv(
+          reporting,
+          mergeChs,
+          Genesis.MainnetNonNegativeMergeableTagName,
+          extraSystemProcesses
+        )
```
tgrospic, line 197: Is this reformat needed?

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -39,25 +40,24 @@ import scala.concurrent.ExecutionContext
...
-      startHash: StateHash,
-      deploy: Signed[DeployData]
-  ): F[Seq[Par]]
   def replayComputeState(startHash: StateHash)(
       terms: Seq[ProcessedDeploy],
       systemDeploys: Seq[ProcessedSystemDeploy],
       blockData: BlockData,
-      withCostAccounting: Boolean
+      withCostAccounting: Boolean,
+      rand: Blake2b512Random
```
tgrospic: Maybe it makes sense to put `rand` immediately after deploys because it's connected how deploys are executed. On other similar methods also.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -457,12 +469,12 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
     )
 
     // Create return channel as first private name created in deploy term
-    val rand = Tools.unforgeableNameRng(deploy.pk, deploy.data.timestamp)
+    val rand = Blake2b512Random.apply(10)
```
tgrospic: Why length 10? We now have 128 defaults. And why to explicitly write `apply`?
https://github.com/rchain/rchain/blob/6dd06305020f02f9fed99c7eda6cbd980fdb0b7d/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala#L42

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -457,12 +469,12 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
...
     import coop.rchain.models.rholang.implicits._
-    val returnName: Par = GPrivate(ByteString.copyFrom(rand.next()))
+    val returnName: Par = GPrivate(ByteString.copyFrom(rand.copy().next()))
```
tgrospic: Why copy is needed? The original rand is not used for any other purposes.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -489,16 +501,21 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
       deploy: Signed[DeployData]
   )(implicit s: Sync[F]): F[Seq[Par]] = {
     // Create return channel as first unforgeable name created in deploy term
-    val rand = Tools.unforgeableNameRng(deploy.pk, deploy.data.timestamp)
+    val rand = Blake2b512Random.apply(10)
```
tgrospic: Length 10 again?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -489,16 +501,21 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
...
     import coop.rchain.models.rholang.implicits._
-    val returnName: Par = GPrivate(ByteString.copyFrom(rand.next()))
-    captureResults(start, deploy, returnName)
+    val returnName: Par = GPrivate(ByteString.copyFrom(rand.copy().next()))
```
tgrospic: The same question for copy here?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -508,30 +525,28 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
   def captureResultsWithErrors(
       start: StateHash,
       deploy: Signed[DeployData],
-      name: Par
+      name: Par,
+      rand: Blake2b512Random
```
tgrospic: `rand` should be just after deploy.

casper/src/test/scala/coop/rchain/casper/rholang/RuntimeManagerTest.scala
```diff
@@ -73,20 +74,39 @@ class RuntimeManagerTest extends AnyFlatSpec with Matchers {
       runtimeManager: RuntimeManager[F],
       deploy: Signed[DeployData],
       stateHash: StateHash
-  ): F[(StateHash, ProcessedDeploy)] =
+  ): F[(StateHash, ProcessedDeploy)] = {
+    val rand = BlockRandomSeed(
+      genesis.shardId,
+      0L,
```
tgrospic: Literal arguments should be named, on other places as well.

casper/src/test/scala/coop/rchain/casper/rholang/RuntimeManagerTest.scala
```diff
@@ -439,7 +455,7 @@ class RuntimeManagerTest extends AnyFlatSpec with Matchers {
         deploy1 <- ConstructDeploy.sourceDeployNowF(
                     s"""new return in { for(nn <- @"nn"){ nn!("value", *return) } } """
                   )
-        result1 <- mgr.captureResults(hash, deploy1)
+        result1 <- mgr.spawnRuntime >>= { _.captureResults(hash, deploy1) }
```
tgrospic, line 459: Why is now a new runtime instance needed?

node/src/main/scala/coop/rchain/node/revvaultexport/mainnet1/StateBalanceMain.scala
```diff
@@ -42,6 +43,15 @@ final case class StateOptions(arguments: Seq[String]) extends ScallopConf(argume
...
   )
+  val shardId = opt[String](
+    descr = "ShardId of the node",
+    required = true
+  )
+
+  val genesisSender = opt[String](
+    descr = "The sender of the genesis",
+    required = true
+  )
```
tgrospic: Why genesisSender is needed?

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -245,13 +249,23 @@ object Transaction {
...
+    val unfogeableBytes = Iterator.continually(rand.next()).drop(10).next()
+    import coop.rchain.models.rholang.implicits._
+    GPrivate(ByteString.copyFrom(unfogeableBytes))
+  }
+
+  // TODO make a hard-coded mainnet unforgeable name after the config of the hard-fork 2 is launched
+  def MainnetTransferUnforgeable: Par = {
+    val rand = Blake2b512Random(10)
+    import coop.rchain.models.rholang.implicits._
+    GPrivate(ByteString.copyFrom(rand.next()))
```
tgrospic: Extract this to helper function because it's duplicated code.

Even better way to create Par is by using RhoType e.g.
`RhoType.Name(GPrivate(ByteString.copyFrom(rand.next())))`.

node/src/main/scala/coop/rchain/node/web/Transaction.scala
```diff
@@ -245,13 +249,23 @@ object Transaction {
...
-    )
-    val unfogeableBytes = Iterator.continually(seedForRevVault.next()).drop(10).next()
-    GUnforgeable(GPrivateBody(GPrivate(ByteString.copyFrom(unfogeableBytes))))
+  def transferUnforgeable(shardId: String, validatorKey: PublicKey): Par = {
+    val rand = BlockRandomSeed(
+      shardId,
+      0,
+      validatorKey,
+      Blake2b256Hash.fromByteString(emptyStateHashFixed)
+    ).generateRandomNumber.splitByte(6.toByte).splitByte(BlockRandomSeed.UserDeploySplitIndex)
```
tgrospic: This pattern repeats, extract it to helper function.

rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoRuntimeSyntax.scala
```diff
@@ -39,9 +39,9 @@ final class RhoRuntimeOps[F[_]: Sync](
       initialPhlo: Cost,
       normalizerEnv: Map[String, Par]
   ): F[EvaluateResult] = {
-    implicit val rand: Blake2b512Random = Blake2b512Random(128)
+    val rand: Blake2b512Random = Blake2b512Random(128)
```
tgrospic: Construction of random Blake2b512Random is now repeated on many places. Please define it on Blake2b512Random itself like `Blake2b512Random.genDefault` or `Blake2b512Random.defaultRandom`.

casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
```diff
@@ -8,21 +12,32 @@ import scodec.{Codec, TransformSyntax}
 final case class BlockRandomSeed(
     shardId: String,
     blockNumber: Long,
-    sender: PublicKey,
     validatorPublicKey: PublicKey,
     preStateHash: Blake2b256Hash
-)
+) {
+  def generateRandomNumber: Blake2b512Random =
+    Blake2b512Random(encode(this))
```
tgrospic: Let's make default constructor private and add constructor on companion object. In that way we can add validation for parameters like shardId to be only ASCII characters.

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -1,152 +1,8 @@
...
-      50,
-      1,
-      1,
-      genesisParams.proofOfStake.minimumBond,
-      genesisParams.proofOfStake.maximumBond,
-      genesisParams.proofOfStake.epochLength,
-      genesisParams.proofOfStake.quarantineLength,
-      1
-    )
-  }
```
tgrospic: How is this become unused?

@nzpr Do you have a clue?

nzpr: I checked out release `v0.9.24` and this is used in only in tests for Engine types (Running, CeremonyMaster and so on), and you replaced these with fs2.Stream based.

tgrospic: Wow, I missed that. It's surprising to see all that code to support this Engine thing. Thanks for finding this out.

casper/src/test/scala/coop/rchain/casper/engine/Setup.scala
```diff
@@ -1,152 +1,8 @@
...
   private def endpoint(port: Int): Endpoint = Endpoint("host", port, port)
 
   def peerNode(name: String, port: Int): PeerNode =
```
tgrospic: You can now remove this whole file. What's left it's only used on BlockRetrieverSpec.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
zsluedem: It is needed. Because `rand.next()` change rand internal state. `rand.next()` is not a pure function

casper/src/test/scala/coop/rchain/casper/rholang/RuntimeManagerTest.scala
zsluedem: If you look at most the methods in runtimeManager, it `new runtime` in most methods. Here is to remove only test used method `captureResults` in `RuntimeManager` trait

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
zsluedem: Too long line

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -21,15 +26,24 @@ final case class Genesis(
...
     val unforgeableByte = Iterator.continually(rand.next()).drop(1).next()
-    GPrivate(ByteString.copyFrom(unforgeableByte))
+    unforgeableByte.toParUnforgeableName
+  }
+
+  // TODO make a hard-coded mainnet unforgeable name after the config of the hard-fork 2 is launched
+  def MainnetNonNegativeMergeableTagName: Par = {
+    val rand = Blake2b512Random.defaultRandom
+    import coop.rchain.models.rholang.implicits._
+    rand.next().toParUnforgeableName
```
tgrospic: Why is this needed specifically for main net?

zsluedem: it is a place holder right now. It is needed in the future.

tgrospic: Why is needed in the future?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
tgrospic: Oh, I see now random is shared and previously copy was created here.

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
```diff
@@ -90,9 +90,16 @@ class MultiParentCasperAddBlockSpec extends AnyFlatSpec with Matchers with Inspe
...
         signedBlock2 <- node.addBlock(deploy2)
+        blockData    = BlockData.fromBlock(signedBlock2)
+        rand         = BlockRandomSeed.fromBlock(signedBlock2)
+
         data <- getDataAtPrivateChannel[Effect](
                  signedBlock2,
-                 Base16.encode(Tools.unforgeableNameRng(deploy2.pk, deploy2.data.timestamp).next())
+                 rand
+                   .splitByte(0.toByte)
+                   .splitByte(BlockRandomSeed.UserDeploySplitIndex)
```
tgrospic, line 99: We need good comments for every split of random generator, otherwise it will be difficult for understand the code later or for new developers.

casper/src/test/scala/coop/rchain/casper/batch1/MultiParentCasperRholangSpec.scala
```diff
@@ -66,18 +70,22 @@ class MultiParentCasperRholangSpec extends AnyFlatSpec with Matchers with Inspec
            |}
          """.stripMargin
 
-      def calculateUnforgeableName(timeStamp: Long): String =
-        Base16.encode(
-          Tools.unforgeableNameRng(Secp256k1.toPublic(ConstructDeploy.defaultSec), timeStamp).next()
-        )
+      def calculateUnforgeableName(block: BlockMessage): String = {
```
tgrospic: This calculation is for deploy so maybe rename it to `calculateDeployUnforgeableName`.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -39,25 +40,24 @@ import scala.concurrent.ExecutionContext
...
-      deploy: Signed[DeployData]
-  ): F[Seq[Par]]
   def replayComputeState(startHash: StateHash)(
       terms: Seq[ProcessedDeploy],
-      systemDeploys: Seq[ProcessedSystemDeploy],
+      rand: Blake2b512Random,
       blockData: BlockData,
-      withCostAccounting: Boolean
+      withCostAccounting: Boolean,
+      systemDeploys: Seq[ProcessedSystemDeploy]
```
tgrospic: Why did you move system deploys? They are executed in a similar way as user deploys and use random generator.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
zsluedem: How would you like to detect mergeable conflicts in the future? I am not sure what you mean

zsluedem: It is just a place holder now. I want to replace it in https://github.com/rchain/rchain/issues/3685

casper/src/test/scala/coop/rchain/casper/addblock/MultiParentCasperAddBlockSpec.scala
zsluedem: I use some named values. Do you think that's enough?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -21,15 +26,30 @@ final case class Genesis(
...
-      StandardDeploys.nonNegativeNumberTimestamp
-    )
-    import coop.rchain.models.rholang.implicits._
+  def NonNegativeMergeableTagName(shardId: String, validatorKey: PublicKey): Par = {
+    val nonNegativeContractIndex: Byte = 3
+    val rand = BlockRandomSeed
+      .generateSplitRandomNumber(
+        BlockRandomSeed(
+          shardId,
+          0,
```
nzpr: should this be `genesis-block-number` from config?

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -21,15 +26,30 @@ final case class Genesis(
...
+  val genesisPubKey = PublicKey(Array[Byte]())
 
-  val NonNegativeMergeableTagName: Par = {
-    val rand = Tools.unforgeableNameRng(
-      StandardDeploys.nonNegativeNumberPubKey,
-      StandardDeploys.nonNegativeNumberTimestamp
-    )
-    import coop.rchain.models.rholang.implicits._
+  def NonNegativeMergeableTagName(shardId: String, validatorKey: PublicKey): Par = {
+    val nonNegativeContractIndex: Byte = 3
```
nzpr: would be good to say why it is 3.
Also there is `BlockRandomSeed.UserDeploySplitIndex` - maybe make this value also a val on some class, e.g. `Genesis.NonNegativeContractSplitIndex `?

nzpr (approved):
LGTM

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
```diff
@@ -119,25 +122,31 @@ object ReportingCasper {
           terms: Seq[ProcessedDeploy],
           systemDeploys: Seq[ProcessedSystemDeploy],
           blockData: BlockData,
-          withCostAccounting: Boolean
+          withCostAccounting: Boolean,
+          rand: Blake2b512Random
```
tgrospic: Please put _random_ parameter after systemDeploys because they are dependent for Rholang execution.

casper/src/main/scala/coop/rchain/casper/rholang/InterpreterUtil.scala
```diff
@@ -216,7 +218,8 @@ object InterpreterUtil {
       systemDeploys: Seq[SystemDeploy],
       runtimeManager: RuntimeManager[F],
       blockData: BlockData,
-      computedParentsInfo: (StateHash, Seq[ByteString])
+      computedParentsInfo: (StateHash, Seq[ByteString]),
+      rand: Blake2b512Random
```
tgrospic: The same here, random generator parameter should be after systemDeploys because they are dependent for Rholang execution.

casper/src/main/scala/coop/rchain/casper/rholang/RuntimeManager.scala
```diff
@@ -39,25 +40,24 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.duration.FiniteDuration
 
 trait RuntimeManager[F[_]] {
-  def captureResults(
-      startHash: StateHash,
-      deploy: Signed[DeployData]
-  ): F[Seq[Par]]
   def replayComputeState(startHash: StateHash)(
+      rand: Blake2b512Random,
```
tgrospic: Same here and in all other methods. Rholang terms should be first because they are main parameters which are executed and random generator immediately after.

I'm OK if you want to put random generator first but be consistent on all methods.

models/src/main/scala/coop/rchain/models/ByteArraySyntax.scala
```diff
@@ -12,4 +13,6 @@ trait ByteArraySyntax {
 class ByteArrayOps(private val ba: Array[Byte]) extends AnyVal {
   def toByteString: ByteString = ByteString.copyFrom(ba)
   def toHexString: String      = Base16.encode(ba)
+  def toParUnforgeableName: Par =
+    Par(unforgeables = Seq(GUnforgeable(GPrivateBody(GPrivate(ByteString.copyFrom(ba))))))
```
tgrospic: RhoType can be used instead.
https://github.com/rchain/rchain/blob/0bfc2aa13e73a27c1f21200dee7336bd252ea17a/rholang/src/main/scala/coop/rchain/rholang/interpreter/RhoType.scala#L102-L110

You can also add overload for `Name`.
```scala
def apply(gprivateBytes: Array[Byte]): Par = apply(GPrivate(ByteString.copyFrom(gprivateBytes)))

// With that unforgeable name can be created simply without the need for extension
val unforgName = Name(<array_byte>)
```

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
tgrospic: Hm, basically we have dependency on data from genesis block to create random seed.

I don't see a problem that we use only part of the data that will not change on the lifetime of the shard in which case we don't need to know anything about genesis block, like sender or block number.

What I mean is to create random generator with only shard name:
```scala
val genesisRandom(shardId: String) =
  NonNegativeMergeableTagName(shardId, PublicKey(Array[Byte]), 0L)
```

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -22,15 +27,36 @@ final case class Genesis(
...
 object Genesis {
+  val genesisPubKey = PublicKey(Array[Byte]())
 
-  val NonNegativeMergeableTagName: Par = {
-    val rand = Tools.unforgeableNameRng(
-      StandardDeploys.nonNegativeNumberPubKey,
-      StandardDeploys.nonNegativeNumberTimestamp
-    )
-    import coop.rchain.models.rholang.implicits._
+  def NonNegativeMergeableTagName(
```
tgrospic: Functions should start with lower case, only object constant should start with capital letter.

models/src/main/scala/coop/rchain/models/ByteArraySyntax.scala
zsluedem: Dependency impossible to introduce `rholang` here.

tgrospic: I'm not saying to change the extension method, but to remove it and use `Name(<array_byte>)` constructor instead.

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
```diff
@@ -65,28 +73,31 @@ object BlockCreator {
...
-                                  .info(
-                                    s"Issuing slashing deploy justified by block ${PrettyPrinter.buildString(invalidBlock)}"
-                                  )
-                                  .as(sd)
-                            }
-        } yield slashingDeploys
+        }
+
+        def needSlashing: Boolean = ilmFromBonded.nonEmpty
+      }
```
tgrospic, line 94: Do we really need this additional inner class? Can this just be a function?

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeReplaySyntax.scala
```diff
@@ -257,28 +261,26 @@ final class RuntimeReplayOps[F[_]](private val runtime: ReplayRhoRuntime[F]) ext
...
+      replayBlockSystemDeploy(processedSystemDeploy, rand).value
     )
 
-  def replayBlockSystemDeploy(blockData: BlockData)(
-      processedSysDeploy: ProcessedSystemDeploy
+  def replayBlockSystemDeploy(
+      processedSysDeploy: ProcessedSystemDeploy,
+      rand: Blake2b512Random
   )(implicit s: Sync[F], span: Span[F]): EitherT[F, ReplayFailure, NumberChannelsEndVal] = {
     import processedSysDeploy._
```
tgrospic: Now when you are refactoring this code, please remove this import. It just makes code less readable.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -134,17 +135,27 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
     */
   def computeGenesis(
       terms: Seq[Signed[DeployData]],
-      blockNumber: Long
+      blockNumber: Long,
+      shardId: String
```
tgrospic: Because shard ID is use only for random seed, it make sense to receive random generator directly instead, like in other methods like `computeState`.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -485,16 +501,20 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
...
+    val rand       = Blake2b512Random.defaultRandom
+    val returnName = rand.copy().next().toParUnforgeableName
+    captureResults(start, rand, deploy, returnName)
   }
 
-  def captureResults(start: StateHash, deploy: Signed[DeployData], name: Par)(
+  def captureResults(
+      start: StateHash,
+      rand: Blake2b512Random,
+      deploy: Signed[DeployData],
```
tgrospic, line 500: Please keep first two parameters the same so it can be just as an overloaded version of captureResults with additional parameters.

casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala
```diff
@@ -503,31 +523,29 @@ final class RuntimeOps[F[_]](private val runtime: RhoRuntime[F]) extends AnyVal
 
   def captureResultsWithErrors(
       start: StateHash,
+      rand: Blake2b512Random,
```
tgrospic: Here is similar, let's keep order of parameters similar to captureResults.

crypto/src/main/scala/coop/rchain/crypto/hash/Blake2b512Random.scala
```diff
@@ -138,6 +138,8 @@ object Blake2b512Random {
     apply(bytes)
   }
 
+  def defaultRandom: Blake2b512Random = Blake2b512Random(128)
```
tgrospic: Please keep _apply_ definitions grouped together as they _specially_ named constructors.

casper/src/main/scala/coop/rchain/casper/blocks/proposer/BlockCreator.scala
zsluedem: I was thinking the cache `ilmFromBonded`. Let me switch back.

node/src/test/scala/coop/rchain/node/mergeablity/MergingBranchMergerSpec.scala
```diff
@@ -87,11 +92,17 @@ class MergingBranchMergerSpec extends AnyFlatSpec with Matchers {
...
-      ) :: Nil
-      blockData = BlockData(blockNum, validator, seqNum)
-      r         <- runtimeManager.computeState(baseState)(userDeploys, systemDeploys, blockData)
+      blockData   = BlockData(blockNum, validator, seqNum)
+      rand = BlockRandomSeed.generateRandomNumber(
+        BlockRandomSeed(
+          genesis.shardId,
+          blockNum,
+          validator,
+          Blake2b256Hash.fromByteString(baseState)
```
tgrospic: We now have all conversions defined as extensions between ByteString <=> Array[Byte] <=> Blake2b256Hash.
https://github.com/rchain/rchain/blob/44b6febd85277df1a1e34c83752b84d342f62327/models/src/main/scala/coop/rchain/models/ByteStringSyntax.scala#L40

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -23,15 +31,29 @@ final case class Genesis(
 )
 
 object Genesis {
+  val genesisPubKey = PublicKey(Array[Byte]())
```
tgrospic: We need to create concise logic how random seed is created in different situations.

For this purpose we created specialized type `BlockRandomSeed` so it make sense to put the logic there.

`genesisPubKey`:
1. is not really genesis public key so it's misleading
2. is not used for any other purpose that to be part of data to create a seed so no need to be available outside of the scope where it's used.

On many places where seed is needed the same code is duplicated and it should be moved to `BlockRandomSeed`. Create helper functions for each scenario which will serve as a reference to show how many cases we have to handle.

Don't try to create everything in one expression which then becomes stretched through many lines.
```scala
val seed = BlockRandomSeed(shardId, 0, genesisPubKey, emptyStateHashFixed.toBlake2b256Hash)
val rand = generateSplitRandomNumber(seed, nonNegativeContractIndex, splitIdx)
```

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -23,15 +31,29 @@ final case class Genesis(
...
+  ): Par = {
+    // NonNegative contract is the 4th contract deployed in the genesis, start from 0. Index should be 3
+    val nonNegativeContractIndex: Byte = 3
+    val rand = BlockRandomSeed
+      .generateSplitRandomNumber(
+        BlockRandomSeed(
+          shardId,
+          genesisRandomSeedBlockNumber,
+          genesisPubKey, // using fixed pubkey to make sure unforgeable name is predictable without configuring sender
+          Blake2b256Hash.fromByteString(emptyStateHashFixed)
```
tgrospic: For genesis all fields can be empty except shard ID. In that case we don't need to track empty state hash to be compatible for specific network.

models/src/main/scala/coop/rchain/casper/protocol/CasperMessage.scala
```diff
@@ -147,10 +148,6 @@ object BlockMessage {
     )
 
   def toProto(bm: BlockMessage): BlockMessageProto = {
-    // To guarantee the same hash lists and maps must be sorted
-    // - can be defined on common place
-    implicit val byteStringOrdering: Ordering[ByteString] =
-      Ordering.by[ByteString, Iterable[Byte]](_.toByteArray)(Ordering.Iterable[Byte])
```
zsluedem, line 153: This is duplicated from https://github.com/rchain/rchain/blob/885def3ae05786339ab5e83adc88730f8844386a/models/src/main/scala/coop/rchain/models/ByteStringSyntax.scala#L15-L16 .
I remove this one because it would cause conflict.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
```diff
@@ -72,9 +70,10 @@ object Genesis {
   ): F[BlockMessage] = {
     val blessedTerms =
       defaultBlessedTerms(genesis.proofOfStake, genesis.registry, genesis.vaults, genesis.shardId)
-
+    val blockData = BlockData(genesis.blockNumber, BlockRandomSeed.GenesisRandomSeedPubKey, 0)
```
tgrospic: This should be real genesis sender in BlockData.

casper/src/main/scala/coop/rchain/casper/engine/LfsBlockRequester.scala
```diff
@@ -220,9 +220,7 @@ object LfsBlockRequester {
                                  .whenA(isLastLatest)
 
                            // Update dependencies for requesting
-                           requestDependencies = Sync[F].delay(
-                             ProtoUtil.dependenciesHashesOf(block)
-                           ) >>= (deps => st.update(_.add(deps.toSet)))
+                           requestDependencies = st.update(_.add(block.justifications.toSet))
```
tgrospic: This change is already on dev branch. Can you please rebase/merge latest dev.

casper/src/main/scala/coop/rchain/casper/Validate.scala
```diff
@@ -263,18 +264,20 @@ object Validate {
...
     } yield status
 
-  // Validator should only process deploys from its own shard
+  // Validator should only process deploys from its own shard with shard names in ASCII characters only
   def deploysShardIdentifier[F[_]: Monad: Log](
       b: BlockMessage,
       shardId: String
-  ): F[ValidBlockProcessing] =
+  ): F[ValidBlockProcessing] = {
+    assert(shardId.onlyAscii, "Shard name should contain only ASCII characters")
```
tgrospic: These changes are also present on latest dev.

casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
```diff
@@ -34,4 +56,114 @@ object BlockRandomSeed {
...
+      shardId: String
+  ): Par = {
+    // NonNegative contract is the 4th contract deployed in the genesis, start from 0. Index should be 3
+    val nonNegativeContractIndex: Byte = 3
+    val seed = BlockRandomSeed(
+      shardId,
+      GenesisRandomSeedBlockNumber,
+      GenesisRandomSeedPubKey,
+      EmptyBytesBlakeHash
+    )
```
tgrospic: Please extract this part in a separate constructor function. I don't think we need to have these three constants at all in that case.

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
```diff
@@ -84,14 +83,18 @@ object ReportingCasper {
...
     new ReportingCasper[F] {
       override def trace(block: BlockMessage): F[ReplayResult] =
         for {
-          reportingRspace  <- ReportingRuntime.createReportingRSpace(rspaceStore)
-          reportingRuntime <- ReportingRuntime.createReportingRuntime(reportingRspace)
-          preStateHash     = block.preStateHash
+          reportingRspace <- ReportingRuntime.createReportingRSpace(rspaceStore)
+          reportingRuntime <- ReportingRuntime.createReportingRuntime(
+                               reportingRspace,
+                               shardId = shardId
```
tgrospic: No need to use named arguments when variable name makes it clear.

casper/src/main/scala/coop/rchain/casper/reporting/ReportingCasper.scala
```diff
@@ -172,14 +181,20 @@ object ReportingRuntime {
 
   def createReportingRuntime[F[_]: Concurrent: Log: Metrics: Span: Parallel](
       reporting: RhoReportingRspace[F],
-      extraSystemProcesses: Seq[Definition[F]] = Seq.empty
+      extraSystemProcesses: Seq[Definition[F]] = Seq.empty,
+      shardId: String
```
tgrospic: Parameter with default values should be at last position.

casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala
zsluedem: What is ` real genesis sender`?

tgrospic: Validator who created a genesis block. With the introduction of the new multi-parent finalizer we don't have special logic for genesis ceremony based on signatures but on stakes and block sender is populated in the same way as any other block created by the validator.
When genesis block is created, sender field is populated with the validator who created the block.
https://github.com/rchain/rchain/blob/9e6e585480638cc60ab2dda73fad3bea41b03a24/casper/src/main/scala/coop/rchain/casper/genesis/Genesis.scala#L98-L125

zsluedem: It was empty by default? 
https://github.com/rchain/rchain/blob/4cf4d64f2d094f2cc512a8ccd53431c392f73e47/casper/src/main/scala/coop/rchain/casper/rholang/syntax/RuntimeSyntax.scala#L145

Do you want to change that?

casper/src/main/scala/coop/rchain/casper/rholang/InterpreterUtil.scala
```diff
@@ -83,6 +85,8 @@ object InterpreterUtil {
                      maxSeqNums = Map[Validator, Long](block.sender -> 0L)
                    ).pure[F]
                  }
+      rand = if (parentsSet.nonEmpty) BlockRandomSeed.fromBlock(block)
+      else BlockRandomSeed.fromGenesis(block)
```
tgrospic: Why not do this check inside `BlockRandomSeed.fromBlock`?

models/src/main/scala/coop/rchain/models/ByteArraySyntax.scala
tgrospic: Please remove this unused extension method.

casper/src/test/scala/coop/rchain/casper/rholang/RuntimeManagerTest.scala
```diff
@@ -58,39 +61,71 @@ class RuntimeManagerTest extends AnyFlatSpec with Matchers {
   implicit val metricsEff: Metrics[Task] = new metrics.Metrics.MetricsNOP[Task]
   implicit val noopSpan: Span[Task]      = NoopSpan[Task]()
 
-  val genesisContext = GenesisBuilder.buildGenesis()
-  val genesis        = genesisContext.genesisBlock
+  val genesisContext  = GenesisBuilder.buildGenesis()
+  val genesis         = genesisContext.genesisBlock
+  val genesisBlockNum = 0L
+  val genesisSeqNum   = 0L
```
tgrospic: These numbers can be read from genesis block itself.
```scala
val genesisBlockNum = genesis.blockNumber
val genesisSeqNum   = genesis.seqNum
```

casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
```diff
@@ -38,4 +57,72 @@ object BlockRandomSeed {
...
+
+  def fromGenesis(block: BlockMessage): Blake2b512Random =
+    fromGenesis(block.shardId)
+
+  def fromGenesis(shardId: String): Blake2b512Random = {
+    val seed = BlockRandomSeed(
+      shardId,
+      GenesisRandomSeedBlockNumber,
+      GenesisRandomSeedPubKey,
+      EmptyBytesBlakeHash
```
tgrospic: `GenesisRandomSeedPubKey`, `GenesisRandomSeedBlockNumber` and `EmptyBytesBlakeHash` are used only here for this purpose so they don't even deserve to have their own name. :)

casper/src/main/scala/coop/rchain/casper/rholang/InterpreterUtil.scala
zsluedem: Genesis is different.

casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
zsluedem: What name do you perfer?

zsluedem: I surrender completely to your paranoia on names.

tgrospic: No name, just add the values here. These names are not used anywhere else except in this method.

casper/src/main/scala/coop/rchain/casper/BlockRandomSeed.scala
```diff
@@ -38,4 +50,76 @@ object BlockRandomSeed {
...
+        PublicKey(block.sender),
+        block.preStateHash.toBlake2b256Hash
+      )
+      generateRandomNumber(seed)
+    } else {
+      fromGenesis(block)
+    }
+
+  def fromGenesis(block: BlockMessage): Blake2b512Random =
+    fromGenesis(block.shardId)
```
tgrospic: This function can be removed as it's part of unified fromBlock.

tgrospic (approved):
Very good work! :tada:

nzpr (approved):
LGTM

tgrospic:
bors merge

zsluedem:
bors merge

